PR #8, opened 2019-04-30
Update theme.styl

NoahZinsmeister:
Remove dashes from intro block `<li>`s for clarity, since only the header is clickable.

image


alancwoo:
Good call, thanks! The links are now the full block on homepage: 99c9e0d841114b9a55b7a7eaf41f3e94137b8f56

PR #10, opened 2019-04-30
Added Totle to 'use' page.

flygoing:


flygoing:
Given the fact that "shill"ish additions like this are likely to happen a ton, might it make sense to build a queue into this app example page? Just have a large array of them and either randomize it on load or have the page shift through the array every week or so?

alancwoo:
@jjmstark mentioned that Vitalik had a similar idea to add dynamic rotation for some of these lists. What we could easily do is start collecting things into a json file and have them rotate/randomize accordingly - could also combine with a collapsing box so a few could appear up top and the rest could reveal below.

@jjmstark should we think about getting something like that going?

seichris:
@alancwoo @jjmstark Wouldn't it be better to have a curated list of (subjectively) good links/projects, instead of a huge randomized list? I'd guess the community would be fine with the EF judging who comes first.

alancwoo:
@seichris Yes for sure, I imagine the JSON object would still be curated, as I suppose the goal is to present a concise list and not the 'encyclopedia'.

PR #11, opened 2019-04-30
added curated list of defi products

sneg55:


jjmstark:
hey @sneg55 - looking through this I noticed that defiprime says it includes apps built on other platforms like Bitcoin, EOS, Tron, etc. There's nothing wrong with that in principle, but a user who came to defiprime through ethereum.org might be surprised to find themselves looking at a list of applications built on other platforms.

For now, I'll reject this PR. If you added a feature that let us link directly to a subset of the list ("defiprime.com/ethereum-apps-only") we would re-evaluate.

sneg55:
I see. I'd just point that while defiprime.com is platform agnostic, 99% of listed products are built on Ethereum and marked  with Eth logo: 
image

Also, most of the categories consist of Ethereum only products, i.e.: 
https://defiprime.com/assets-tokenization 
https://defiprime.com/derivatives 
https://defiprime.com/decentralized_prediction_markets

PR #15, opened 2019-04-30
more descriptive titles for pages

sneg55:


sneg55:
@jjmstark what do you think about it?

jjmstark:
My initial reaction is that this probably isn't necessary - mainly these changes just add "Ethereum" to the title, but I think the reader will understand that on ethereum.org/learn, the site is for "learning about ethereum". 

What do others think?  cc @ethjoe @EvanVanNessETH @bmann ?

@sneg55 you might also be interested in this Issue I opened recently, which is related: https://github.com/ethereum/ethereum-org-website/issues/72

sneg55:
That's how existing titles looks like at google:  
image

EvanVanNessEth:
I agree with @sneg55.   We don't want to overdo it, but a descriptive title that looks good in google search results plus ranks better in results is worthwhile

something simple like "The Ethereum blockchain is an open-source decentralized application platform"

bmann:
Yeah. Ethereum will Show up in search differently than optimizing for EthereumDotOrg. Ethereum should be the publicly visible phrase. 

Social metadata per page really should be optimized— my PR was a first step. SEO is actually accessibility if you think about it.

Issue #6, opened 2019-04-30
missing OGP tags in HEAD


sneg55:
Please add http://ogp.me/ tags so when links shared at social media they will show with a proper cover image and title/description

bmann:
I'm new to Vuepress, so just digging through this. Here's an article with some tips https://alligator.io/vuejs/vue-seo-tips/

There is currently one meta header in the config.js file.

bmann:
OK, found some examples I can get started with https://github.com/vuejs/vuepress/issues/777

bmann:
Looks like this plugin is specifically for Vuepress and can help https://github.com/lorisleiva/vuepress-plugin-seo

bmann:
Merged for now

PR #20, opened 2019-04-30
First pass at adding social headers

bmann:
Addresses #6 

Adds basic OGP / Twitter metadata. Uses the image on the front page, which maybe doesn't read the best.

Probably should install this plugin and configure defaults https://github.com/lorisleiva/vuepress-plugin-seo

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -12,20 +12,14 @@ module.exports = {
   },
   head: [
     ['meta', {name: 'viewport', content: 'width=device-width,initial-scale=1,maximum-scale=1,maximum-scale=1'}],
-    ['link', {rel: 'icon', type: 'image/png', href: 'favicon.png'}]
+    ['link', {rel: 'icon', type: 'image/png', href: 'favicon.png'}],
     ['meta', { name: 'twitter:site', content: '@Ethereum' }],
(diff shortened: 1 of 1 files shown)


docs/.vuepress/config.js
```diff
@@ -12,7 +12,14 @@ module.exports = {
   },
   head: [
     ['meta', {name: 'viewport', content: 'width=device-width,initial-scale=1,maximum-scale=1,maximum-scale=1'}],
-    ['link', {rel: 'icon', type: 'image/png', href: 'favicon.png'}]
+    ['link', {rel: 'icon', type: 'image/png', href: 'favicon.png'}],
+    ['meta', { name: 'twitter:site', content: '@Ethereum' }],
+    ['meta', { name: 'twitter:creator', content: '@Ethereum' }],
+    ['meta', { property: 'og:type', content: 'article' }],
```
mirshko: this should be set to "website" not "article"

alancwoo:
Thanks @bmann ! I imagine down the road if we want more heavy SEO we could go with something like https://github.com/nuxt/vue-meta which would work well with vuepress as well!

bmann:
@alancwoo I found a vuepress specific one https://github.com/lorisleiva/vuepress-plugin-seo

Likely each page / section needs proper metadata, which looks like it can be added in Markdown front matter. I’ll come back to this later and experiment.

PR #38, opened 2019-05-01
Added stuff from the Consensys developer portal

virgil:
https://ethereum.consensys.net/ethereum-dev-portal

I think we should probably add their images as well, but I'll leave that decision to the designer.  I added two of the images just to get a sense for how it would look if we did this naively.


virgil:
Closing this request.  Will submit a new revised one.

PR #42, opened 2019-05-01
Add Buidler to list of dev tools in build section.

fzeoli:


maurelian:
Happy to include buidler, but please make the formatting consistent with the rest of the listings.

image

docs/build/index.md
```diff
@@ -64,6 +64,12 @@ Ethereum has a large and growing number of tools to help developers build, test,
 ### Looking for other options?
 - [Ethereum Developer Tools List #Frameworks](https://github.com/ConsenSys/ethereum-developer-tools-list#frameworks)
 
+**Workflow automation**
+
+### Buidler *A task runner for Ethereum smart contract developers*
+- [buidler.dev](https://buidler.dev)
+- [Github](https://github.com/nomiclabs/buidler)
+
```
maurelian: Lines 67 to 71 should be moved above line 64, or is Buidler not a developer tool?

fzeoli: It is, and the way I see it this is the structure I proposed:
```
Developer tools
    Frameworks
        Truffle
        Embark
        Waffle

    Workflow automation
        Buidler
```

Buidler is a task runner, not a development or testing framework, so I added a new category under `Developer tools`, same hierarchy level as `Frameworks`. I wasn't sure where else to put it.

What do you suggest?

fzeoli:
> Happy to include buidler, but please make the formatting consistent with the rest of the listings.
> 
> image

I think it is the same format as the other listings. `Workflow automation` is formatted the same as `Frameworks`, and the Buidler line is the same as the other tools. Maybe the `Workflow automation` bit was confusing out of context? or am I missing something?

image

docs/build/index.md
maurelian: You're right. Actually the issue is that the `Frameworks` formatting is wrong. 
image

IMO 'frameworks' which is probably one of the vaguest words in all of software should just be removed, and Buidler inserted under Developer tools so that we have:

```
Developer tools
        Truffle
        Embark
        Waffle
        Buidler
```

What do you think?

fzeoli: I agree 100% about "frameworks", but I think the original intention was to have "Developer tools" as the top level category for all the developer tools, and then have more detailed categories inside. Maybe we could repeat the tools across several categories to communicate more precisely what they're good for? like

```
Developer tools
    Testing libraries
        Waffle
        Truffle
        Embark
    Dapp development
        Embark
    Workflow automation
        Buidler
    Deployment
        Etherlime
        Truffle
        Embark
        Buidler
     Linters
          Solhint
```
I know that with the huge amount of features some of these tools have they'll be repeated a lot, but while redundant this feels like much better orientation for someone who's never used any of them.

What do you think? I'm also not against doing just one list without categories but I don't think that will last very long as all the dev tools are added

maurelian: At this point, I don't want to list things multiple times, or worry about how to design for another level of headings. 

Going with the simplest solution for now.

PR #45, opened 2019-05-02
Adding etherlime in the frameworks section

Perseverance:


docs/build/index.md
```diff
@@ -57,6 +57,10 @@ Ethereum has a large and growing number of tools to help developers build, test,
 - [Github](https://github.com/embark-framework/embark)
 - [Documentation](https://embark.status.im/docs/)
 
+### Etherlime *Ethers.js based framework for dapp development (Solidity & Vyper), deployment, debugging, testing and more.*
```
maurelian: This would be easier to merge if it weren't arbitarily placed above another item which is already included. 

I'm not favoring waffle, as I've used neither. I just want to default to 'first come first served' without other strong arguments to go on.

Perseverance: I've no particular preference where to put it, I've randomly put it there for chronological reason, but if you want to default to FIFO I am fine to move it downwards

PR #47, opened 2019-05-02
Add Kauri under "Learn about Ethereum"

kcole16:
Add kauri.io in the "Learn" section


kcole16:
Makes sense, updated!

PR #58, opened 2019-05-07
"Build" page doesn't list tools to run a dev client

vrde:
In the *Build* page I see information on how to testnet and how to run your own node, but it doesn't mention any tool on how to run a local node for development and testing, and it's not obvious to a newcomer that they can run `geth` or `parity` in *single player mode* to just practice interacting with a real node.

I've added a link to [ethnode](https://github.com/vrde/ethnode), a simple tool to locally run **Geth** or **Parity** in development mode.


maurelian:
+1. More devs should be running tests against a production node like geth/parity.

vrde:
> +1. More devs should be running tests against a production node like geth/parity.

@maurelian cannot agree more :)

Thank you for merging.

PR #60, opened 2019-05-08
Add a link to explain the term "blockchain"

tasdienes:
My parents reviewed this page and one comment was that the term "blockchain" is dropped without explaining what it means.  I added a link to one of the most simple and low-tech explanations that I could find.

### DIFF
--- a/docs/beginners/index.md
+++ b/docs/beginners/index.md
@@ -13,7 +13,7 @@ New to Ethereum? You’re in the right place. Let’s start with the big picture
 - An internet where everyone has access to an open financial system.
 - An internet built on neutral, open-access infrastructure, controlled by no company or person.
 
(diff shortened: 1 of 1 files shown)


maurelian:
I'm curious why that particular link was chosen?
Linking to wikipedia seems like the more obvious default option: https://www2.deloitte.com/ch/en/pages/strategy-operations/articles/blockchain-explained.html

tasdienes:
Wikipedia is good, but I was looking for something that my parents could understand, and I think they would have a hard time with the Wikipedia definition.  Multiple references might be even better.

jjmstark:
@maurelian agree that wikipedia would be a better link.

maurelian:
Feel free to reopen if the link is changed.

PR #61, opened 2019-05-08
Whiteblock Genesis Under Testing Tools

zscole:
Added Whiteblock Genesis (open source software) under testing tools section.

### DIFF
--- a/docs/build/index.md
+++ b/docs/build/index.md
@@ -124,6 +124,11 @@ Ethereum has a large and growing number of tools to help developers build, test,
 ### hevm *Implementation of the EVM made specifically for unit testing and debugging smart contracts.*
 - [Github](https://github.com/dapphub/dapptools/tree/master/src/hevm)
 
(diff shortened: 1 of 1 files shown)


maurelian:
Seems like whiteblock would be a valuable tool for consortia etc. operating private/permissioned networks. 

@jjmstark is that in scope?

@zscole am I wrong?

jjmstark:
@maurelian Yes, I think that is in scope.

If in the future there are many tools aimed at the "consortia" use-case, we might want to create a specific section for them in /Build.

zscole:
> Seems like whiteblock would be a valuable tool for consortia etc. operating private/permissioned networks.
> 
> @jjmstark is that in scope?
> 
> @zscole am I wrong?

Yes, I believe that's a potential use case as well, but we provide explicit support for a multitude of Ethereum-based clients and protocols. It's really just a blockchain specific development and testing framework and quite a few Ethereum projects are using it now.

PR #68, opened 2019-05-10
Color and type changes

alexvandesande:
Let's be honest, the site isn't winning everyone's hearts. And while I really like the simplicity aspect of it, having very few pages and most of them pointing outward, I think the typography and colors could be improved. This PR isn't an overall redesign, just small changes that in my opinion are a good starting point for the website

### Before

Screen Shot 2019-05-10 at 5 12 56 PM

The blue "ethereum.org" homepage makes it look like a parked domain. The type is small and has no contrast. Everything is spans and ul's under the hood

### After

Screen Shot 2019-05-10 at 5 13 02 PM

Changed the title to just show ethereum. Still need to work on a better color. Each header is a different color, following the foundation's palette. Added H3 and H2s to the code and made them contrast more, with variations of the text including a condensed version.

-----

Is it the best thing ever? No.
Is it an improvement on the current thing? Yes (IMHO).
If we get this merged I'd like to work on a better hero header for the image as well as thinking on how to do small UI improvements like linking to the actual file of the GitHub that needs to change.


jjmstark:
@alancwoo what do you think?

I'm not sold on these specific colour changes - one reason to keep the colour profile of the landing page simple is that we will cycle through HERO images with wildly different designs & colour schemes. Keeping the rest of the page relatively simple might make it less likely to clash with a given colour scheme. 

Another option would be that each time we change the HERO image, the designer who submitted it can tweak the colour of the links on ethereum.org. This means there wouldn't be any "clash" and would be an indicator on the rest of the site that something has changed.

FYI @alexvandesande a few hours ago we uploaded a [design spec](https://github.com/ethereum/ethereum-org-website/blob/master/ethereum.org-hero-image-specs.pdf) for the HERO image submissions - if you're interested in submitting one, please take a look! We are actively talking to several designers right now to get a HERO image to replace the current one in the near future.

alexvandesande:
I think both are possible. Seems to me you're trying to do a "google doodle" style, accepting code and video, but that's an example where you can have both a changing main image and consistent pallete. Of course, we could change these colors with the hero image too.

Still, I still think the type is an improvement, even if you take out the color.

alancwoo:
@alexvandesande Thank you for the suggestions so far as the purpose was to start with an essentially 'blank slate' and move forward with community suggestions. I would agree with a number of the suggestions, in particular simplifying the title to `Ethereum` - @jjmstark what are your thoughts on that point?

Regarding other changes, I'm not entirely sure how I feel about how Plex Light is rendering, especially on a non-HiDPI screen, but do see the strength of increasing contrast in headers. I could also see it being interesting that the 'Accent' colour being dictated by the hero image being interesting as well (as long as it functions in dark mode as well).

@alexvandesande is your primary concern that scanning pages is difficult due to a lack of visibly bold headers/sections? Is there also some reason behind the all-caps below but not above?

@jjmstark is it possible to even simplify further by somehow combining the bullet points and the hero image navigation?

alancwoo:
image

The contrast is strong in the headlines but a bit problematic in the copy text

jjmstark:
@alancwoo I think changing from `ethereum.org` to `Ethereum` makes sense for the reasons @alexvandesande gives! @alancwoo Can you implement that change?

@alancwoo pending the larger idea of "changing colours with the HERO image", changing the 4 subpage colours to match the colour scheme used for Devcon as @alexvandesande makes sense. What do others think? If there are no objections, we should also implement that change.

alexvandesande:
@alancwoo " is your primary concern that scanning pages is difficult due to a lack of visibly bold headers/sections? Is there also some reason behind the all-caps below but not above? @jjmstark is it possible to even simplify further by somehow combining the bullet points and the hero image navigation?"

I definitely need it needs more contrast in the text. I feel the current design really looks too bare as if there's no CSS at all. Regarding the contrast in internal pages, the MD also needs modifying to change from only using STRONG to also using headers. Here's how the "beginners" page looks for instance:

Screen Shot 2019-05-16 at 4 52 43 PM

I didn't change much on the hero layout partly because I agree that it's content's feel redundant with the section below it, but I didn't want to change too much in this PR. I think that we could probably move the text below each header to the sections below it and have the hero image be just an image (maybe with some text internally depending on the hero). It allows more flexibility for hero images.

We can discuss personal choice in types forever, I picked Flex Sans Condensed because it had a large family of weights and widths, and because I like preferring open source fonts, by the spirit of the thing.. It's the same reason I previously used Source Sans Pro. I like the contrast mainly, and it allows us to use very large and bold, or very thin and condensed.

Also notice I did also make a few under the hood adjustments like adding more classes and different types of tags, because it usually helps the robots to read it (let's be inclusive to robots too!).

oovg:
Being such a text-driven website, great usability improvements can be made by simply applying better typographical hierarchy. This PR by @alexvandesande would be a **great** start and enable us to keep iterating incrementally. 

@alancwoo any issues here that need resolved before merging?

samajammin:
@alexvandesande - thank you for this! You surface some key opportunities for design improvements:

- Typography
  - [Establish consistent font hierarchy](https://github.com/ethereum/ethereum-org-website/issues/203)
  - Use appropriate HTML elements (e.g. headers vs. list items for homepage section headers)
- Color palette
  - [Establish primary, secondary, tertiary colors to use across the site](https://github.com/ethereum/ethereum-org-website/issues/204)

I think the best approach here is to address each of these in individual issues/PRs. I'm going to close this PR in favor of these newly created issues. 

Feel free to re-open this if you feel there's anything I've overlooked. I also encourage you to continue to surface ideas! Your effort is much appreciated. For future ideas/work, I recommend isolating independent changes into separate PRs. This will make it easier to gather rough consensus & achieve incremental improvements.

PR #71, opened 2019-05-13
Nethermind

tkstanczak:
Added Nethermind to the clients list


CPSTL:
At this point in time, Nethermind is not ready to be added. Getting the client formally audited would be a good start.

Issue #27, opened 2019-05-01
The contrast of the Ethereum graphic is too low


ricburton:
The image is really light against the white and black background.

Is there a raw version of this file somewhere? Happy to play with some options.

jjmstark:
@alancwoo should be able to share it with you!

If you come up with something you like, shoot an email to website@ethereum.org. We want to rotate out the current image in the near-term and are collecting submissions from the community.

alancwoo:
@ricburton Sorry for the delay, this is the raw image!

![hero-bg](https://user-images.githubusercontent.com/1481890/57944674-0a255a80-78d8-11e9-9010-b147e38d9c76.png)

oovg:
@ricburton maybe an animated gradient overlay gif could be cool.

Here's a pass for review.
- removed background color (now transparent .png)
- now only one image is needed (currently two diff images are being served for light/dark)
- added a gradient overlay that maintains similar contrast on light and dark (want to tweak a little more though)
- added a bit more padding on top and bottom for breathability of content, and allows the image to be a **bit** larger

Screenshots of light and dark mode:
![Screen Shot 2019-05-17 at 12 21 55 PM](https://user-images.githubusercontent.com/23664663/57951446-eeea2900-789e-11e9-88ab-3e29f3d35f95.png)

![Screen Shot 2019-05-17 at 12 22 08 PM](https://user-images.githubusercontent.com/23664663/57951464-fad5eb00-789e-11e9-8944-dcaf7ba9eea7.png)

oovg:
@alancwoo PR => https://github.com/ethereum/ethereum-org-website/pull/81

ethjoe:
Wholeheartedly endorse this, @oovg. Small but beautiful change. @jjmstark and others?

alancwoo:
@oovg I think this is wonderful! @jjmstark what do you think? @ricburton also think animating this would be really great.

jjmstark:
@alancwoo No opposition from me! This is a great iterative improvement.

@oovg I should note that we are likely to swap in an entirely different image relatively soon, but this is a really nice change for the meantime :)

ricburton:
Definitely like this direction. What's the entirely new image looking like?

Richard

alancwoo:
@ricburton I believe there are submissions and also some illustrators/designers on board incorporating the eth branding/logo in more deliberate fashion.

alancwoo:
#87

PR #81, opened 2019-05-17
Replaces HERO image with one that creates more contrast in light and …

oovg:
…dark mode. Removes requirement of showing a diff image for light/dark modes


alancwoo:
@oovg Thanks so much, this has been merged into production which I believe is synced to master weekly.

alancwoo:
Close #27

oovg:
@alancwoo what happened to the new HERO image with better contrast? Doesn't look like it made it to production/master.

PR #87, opened 2019-05-20
Hero image

alancwoo:
Remove PNG, adjust crop/compression


oovg:
@alancwoo What happened to the hero image with better contrast (gradient overlay)?

PR #90, opened 2019-05-21
feat: add ctrl n/p to searchbox

shanejonas:


alancwoo:
@shanejonas Just wondering, on Linux `ctrl+n` will typically open a new window and `ctrl+p` will trigger a print, so not sure how useful it is? (You're on mac I imagine?)

PR #94, opened 2019-05-23
Add Pantheon to list of Ethereum Clients

timbeiko:
Add Pantheon to list of Ethereum Clients


timbeiko:
@djrtwo thanks for approving! What is the process of getting this merged?

PR #98, opened 2019-05-28
Added Manticore and Oyente.

virgil:
I especially want Manticore added.


maurelian:
Is anyone still using Oyente? No commits in almost a year, so it won't support create2 for instance.

virgil:
Okay drop Oyente

PR #101, opened 2019-05-28
Added Storage section

virgil:
Seems a sensible thing to mention.


jjmstark:
makes sense to me - thoughts @maurelian @CPSTL?

maurelian:
If this is targetting developers coming to ethereum... a sentence with context as to _why_ they need storage might be useful?

virgil:
Suggest whatever you'd like to better explain it and the merger can add it in.

maurelian:
Sorry for blocking this. I don't have time right now to modify it, and think it's better included imperfectly than not at all. I've merged locally to fix a conflict and pushed these changes to staging.

Issue #63, opened 2019-05-08
Create a feature that randomly chooses 5 of 10 "suggested apps" to display on /Use


jjmstark:
We would like to have a feature where:

- We have a pre-determined list of 10 applications built on ethereum
- Every time a user loads the page, 5 of those applications are chosen at random
- Those 5 are displayed in a random order on https://ethereum.org/use/, under "Use an application built on Ethereum"

maurelian:
Looking at #23, I can see how this will be a regular item for consideration. Is there someway of deferring the curation to another source? There's a lot of dapp listing site out there, maybe one has an API we can pull from, or we can just link out to them. 

If we go the direction of "just include everything that's not an obvious scam", I'd be in favor of modifying this feature to: 

* We have a huge list of applications built on ethereum
* Every time a user loads the page, 5 of those applications are chosen at random
* Those 5 are displayed in a random order on https://ethereum.org/use/, under "Use an application built on Ethereum"
* There's a link to "view more apps", which has the full list.

jjmstark:
I propose we keep the list small and tightly curated for the immediate future, for the following reasons:

- Committing to "everything that's not a scam" still creates a very large surface area for review (there are ~1,900 listed on stateofthedapps.com). 
- If we ever make a mistake and do list a scam, we've created a horrible experience for some person who used ethereum.org. 
- The purpose of linking to these apps on /Use is to actually show the user something interesting they can "do with Ethereum". If that's the goal, a curated list of high-quality applications - with good landing pages, good UX, known to be non-scams - seems to achieve it better. 
- We already achieve part of the "be comprehensive" goal by linking to other people's lists further down this page.

jjmstark:
One valuable thing to do here would be to define the criteria we're using to put applications on this list. A few starting points:

- Is the app a "user" application? There are many applications or services "built on ethereum" that are narrowly targeted at developers or sophisticated users. For instance, we would not list Infura here (though it is, and should be, on the /Build page)
- Does the app have a good user on-boarding process? Can a user follow our link, and find all the instructions they need to get started on the applications website?
- Does the application "round out" the list by adding a new *kind* of application not already present?

maurelian:
Both of the comments are great. 
I think the only criterion I would add is:

* Is it a popular / well-established app?

That's subjective, but cuts out the incentive for a long tail of apps trying to get a free listing.

maurelian:
The only criterion I'd add is: 

* Should be a popular and established application. 

[...]

PR #106, opened 2019-06-04
Italian translation

maxxflyer:
I just translated to italian only the welcome page so far. I guess we need a way in the home page to reach any kind of language translations.  How are we going to manage this and the website for translations?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,33 +3,33 @@
   <p>👋 Benvenuti a ethereum.org!</p>
 </h1>
 
-[Ethereum.org](https://ethereum.org) is a primary online resource for the Ethereum community. It is improved and changed over time through the contributions of community members who submit content, give feedback, or volunteer their time to managing its evolution.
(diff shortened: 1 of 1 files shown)


nihal9ns:
I believe it would be a good option to have a translation service integrated into the website, so we can simply add the strings in `key: value` pairs and then on the web, use the keys. The benefit of this is that we can have the strings translated to any language that we require.

alancwoo:
@maxxflyer This is wonderful, thank you! Maybe we can work out a plan to structure translations sitewide so any contributor can indeed PR translations via https://vuepress.vuejs.org/guide/i18n.html#site-level-i18n-config

My only concern is that links/items could fall out of sync between languages causing quite a fragmented experience.

However, maybe if we pull out 'lists' into JSON objects as proposed in #63 we could avoid this issue.

jjmstark:
Hey @maxxflyer - this is awesome! We are currently in the midst of a large translation effort, which will be announced soon. Would you like to be involved in the Italian translation program? Send us an email at website [at] ethereum.org, and we'll put you in touch with the team managing this project.

jjmstark:
Closing since this is now already underway on Crowdin

maxxflyer:
I found the crowdin translation. Not sure how I can help, I see some work is already done for the website

PR #111, opened 2019-06-07
Add mythril to list of security tools

s0b0lev:


CPSTL:
> An impartial reviewer would wonder why Slither was pushed down in the list. I'd accept if there was only the addition without reordering.

I agree.

s0b0lev:
Hi! Looks like I remove the fork from which create a PR, may I submit another PR and close this one?

maurelian:
certainly!

PR #123, opened 2019-06-26
Tutorial is updated regularly

maheshmurthy:


jjmstark:
Looks like this blog post was last updated in February 2019 - I would be OK moving up the "last updated" date to February, but I don't think it would qualify for "Updated often".

maheshmurthy:
Sure, I just updated to February as suggested. Thanks!

PR #130, opened 2019-07-10
Added scaling overview article

awallendahl:


jjmstark:
The Ethereum 101 blog series is already linked on /Developers, so I don't think it makes sense to include twice. If you remove that, should be no issue merging to add the Scalability article!

awallendahl:
just saw that - will update. thanks

PR #131, opened 2019-07-10
Added kauri to list of developer resources

awallendahl:


jjmstark:
Hey @awallendahl - Kauri is already listed as a general resource on /learn, and we link to several Kauri articles on /developers. I don't think its necessary to add an additional link to /Developers at this time, although if there are specific articles on Kauri that you think should be added to any particular section, please do submit a PR!

PR #139, opened 2019-07-23
Add beginner introduction to Ethereum

kvhnuke:


ethjoe:
Incredible work! It's very thorough, well rounded, and not overly product-focused. 

It may be a better fit under the "Use" section than "Beginner" after a quick read, with a first half very much focused on wallets and how to obtain ETH. That said, there should be a place for it on the site.

jjmstark:
Great! I would say let's even add it to both /Use and /Learn, as it's a very comprehensive article that covers topics relevant on both pages. I've created PRs #145 and #146 to add it to each page, in a section that I thought made sense.

PR #163, opened 2019-08-09
Update README.md

Z3N0:
should be "has" instead of "it"


jjmstark:
Not actually a typo - "is" is used because we're referring to a continuous process in the present tense ("the website is improved by..."), rather than "has" which would be past-tense ("the website has been improved by...").

PR #166, opened 2019-08-11
Replace WorkSans with system font for readability. Remove link alike underline header's decoration

rumkin:
Current font has bad readability due to huge letter spacing which makes strings to fall apart. It's good for title, but not for text. Apple's San Francisco and Microsoft's Segoe UI fonts have better readability and adapted for long continuous reading and big blocks of text. It's not so fancy, but extremely functional.

Also I fixed headers' 1px underlines which made them look like links, what can confuse user.

## Before
Screen Shot 2019-08-11 at 14 05 01

## After
Screen Shot 2019-08-11 at 14 04 53

As a bonus it freed around 1 Mb of network traffic for font loading.


samajammin:
Thank you for this! I'm going to close this for now, mainly on the basis that (1) we're not convinced the current font has major readability issues, and (2) not wanting to use the default Apple system font.

Open to feedback on this if others want to way in - this is currently a "weakly held" opinion.

PR #167, opened 2019-08-11
What is a smart contract - EthHub Link

jamesconnolly93:
As mentioned previously, I've started beginners definition of Smart Contracts/DApps on EthHub. Planning to continue to build on it, but thought the link would belong here.


jjmstark:
Also, does EthHub have section-anchor-links? would be better to link directly to the bit about smart contracts, the user might be confused by clicking on something with the title "What are smart contracts?" and ending up at a page that starts "What is Etheruem?"

jamesconnolly93:
Good catches - thanks! I've updated the link to fix the typo and add in the anchor link.

Issue #276, opened 2019-10-01
Style the 404 page


samajammin:
## Expected Behavior
Helpful/whimsical 404 page.

## Current Behavior
Page doesn't look great
![Image 2019-10-01 at 2 23 22 PM](https://user-images.githubusercontent.com/8097623/66001762-0a715d00-e457-11e9-857b-75aec3be3448.png)

## Possible Solution
https://github.com/vuejs/vuepress/issues/1178

amanchande:
Can I try this?

samajammin:
@amanchande yes, please! Thanks for asking. Assigning this your way.

amanchande:
I am currently new to github and vue so can you please give some guides fore the same.

samajammin:
Github:
https://guides.github.com

Vue:
https://vuejs.org

This site uses VuePress specifically:
https://vuepress.vuejs.org

This VuePress issue may help you with the custom 404 page:

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 50.0 DAI (50.0 USD @ $1.0/DAI)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/ethereum-org-website/276/3537).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/ethereum-org-website/276/3537).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $152,365.32 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 1 month from now.
Please review their action plans below:


**1) [iower](https://gitcoin.co/profile/iower) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/ethereum/ethereum-org-website/276/3537?mutate_worker_action=approve&worker=iower) | [reject worker](https://gitcoin.co/issue/ethereum/ethereum-org-website/276/3537?mutate_worker_action=reject&worker=iower))_.**

For the 404 page:
[...]

PR #310, opened 2019-10-09
Using a simple, translatable 404 page that uses the sites global theme

ScyDev:
## Description

- If no matching route was found, redirect to the 404 page.
- The 404 page has its own MD file which is handles the same as every other page, therefore using the global theme. And it's translatable.

Please advise what different text should be used for the 404 page. 
An alternative solution would be to use a separate `NotFound.vue` file, but I found the separate 404.md file to be the cleaner way.

## Related Issue

#276

Screenshot from 2019-10-09 16-32-58

## Checklist:

- [x] I have read the contributing guidelines in the project README.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.


NeoSPU:
> have been fixed the errors from previews pell-request
## Screenshot:
[https://drive.google.com/drive/folders/1heQYMgMj3t7dthN6SfK3lFAfM-yahZ4y?usp=sharing](url)

docs/404.md
```diff
@@ -0,0 +1,15 @@
+---
+title: 404
```
samajammin: ```title: Not Found```

docs/404.md
```diff
@@ -0,0 +1,15 @@
...
+meta:
```
samajammin: Delete this `meta` section.

docs/de/404.md
```diff
@@ -0,0 +1,8 @@
+---
```
samajammin: Please delete this & find a solution that uses the same 404 page for every language. I'd prefer not to add the burden of translating the 404 page into every language.

Even in you're current solution, this page is never reached. E.g. when I navigate to https://deploy-preview-310--ethereumorg.netlify.com/de/asdfasdf it redirects to the English 404.

docs/404.md
```diff
@@ -0,0 +1,15 @@
...
+  - name: description
+    content: Ethereum is a global, decentralized platform for money and new kinds of applications. On Ethereum, you can write code that controls money, and build applications accessible anywhere in the world.
+  - property: og:title
+    content: Home | Ethereum
+  - property: og:description
+    content: Ethereum is a global, decentralized platform for money and new kinds of applications. On Ethereum, you can write code that controls money, and build applications accessible anywhere in the world.
+lang: en-US
+---
+
+# Oh noes! It's a 4-oh-4!
```
samajammin: ```# Not Found```

docs/404.md
```diff
@@ -0,0 +1,15 @@
...
+
+Please use the search box above to find what you're looking for.
```
samajammin: ```Please use the search box above to find what you're looking for or [return home](/).```

docs/de/404.md
ScyDev: Seeing that the rest of the site is translated, why not translate the 404 page? It's not mandatory though, if no specific `404.md` is placed in a language subdir, then the english one from the root dir will be used.

It worked reliably on my local dev env. I'll check again and see what's going wrong here about the 404 page not being reached.

samajammin: Mmm ok, I can support that! Use the language-specific 404, otherwise fallback to English if it doesn't exist.

Again, I think it's also important to retain the URL that the user typed in (i.e. maintain, that current UX):
> Can you also find a solution that retains the URL the user attempted to reach (vs. redirecting them to 404.html)? That's typically a best practice so the user can see the URL that was broken.

samajammin: In other words, if the `NotFound.vue` maintains the URL, I'd rather go that route. We can sort out the language-specific copy in a similar fashion to the HomePage component, by using translations.js.

ScyDev: I'll check if I can maintain the URL with `404.md`.

IIRC with the `NotFound.vue` approach there was no easy support for the sites global theme. 

Will check all the points you mentioned in the next few days and update you.

ScyDev: @samajammin I got it working with `NotFound.vue`, keeping the original URL. Put the script from `Layout.vue` into a separate file, because it's also needed in `NotFound.vue` when using the sites global theme. I hope that's ok.

For some reason now the "Not Found" page is always in english, including the menu. Since you didn't want it translatable earlier, should I put more effort into fixing this?

ScyDev:
@samajammin btw can you please approve me on Gitcoin? Thanks.

samajammin:
Thank you for this @ScyDev! I didn't create the Gitcoin bounty so I can't approve you immediately but I'm working with the folks who did & we'll make sure you get approved & paid out ASAP.

ScyDev:
@samajammin Thanks, I enjoyed it.

samajammin:
@ScyDev glad to hear! Hope you continue to contribute 😄

ScyDev:
@samajammin Gladly, if you put the issues on Gitcoin I certainly will :) Speaking of which, the bounty hasn't been released yet and the funder hasn't reacted to my email. Can you ping him perhaps?

ceresstation:
@ScyDev Sorry about that, releasing the funds now!

PR #623, opened 2019-12-04
Remove eth-utils from the Rust projects list

ksato9700:
## Description
Because [eth-utils](https://github.com/ethereum/eth-utils/) is not a Rust project but a Python project, I would suggest to remove it from the Rust projects list.


samajammin:
@all-contributors please add @ksato9700 for content

PR #520, opened 2019-11-14
Add 'Get Involved' page to help people contribute

wschwab:
See #489 : This is a rough draft (intended to start discussion) for adding resources to help people find what they can do to get involved with the Ethereum ecosystem.

## Description

This PR adds a new page: 'Get Involved', and also adds a some links about running a node to the developer resources (this had been originally included in the Get Involved page, and was moved on suggestion).

## Related Issue
#489 

A note: I don't know Vue, so I hope I managed to copy effectively from the patterns other pages followed, my apologies if this creates extra work.

Issue #489, opened 2019-11-12
Add 'Get Involved' page to help people find what they can do


wschwab:
(This idea has been discussed by the [Ethereum Cat Herders](https://www.ethereumcatherders.com/) and a member of the EF, which is why I'll be using 'we' instead of 'me'. Just to be clear - I wrote the content and the code, and the ECH is looking it over.)

**Is your feature request related to a problem? Please describe.**

We think it would be good to have a page dedicated to helping people find what they can do to be involved in the ecosystem, coders and non-coders alike.

**Describe the solution you'd like**

We propose adding a new page to the menu on the home page title 'Get Involved' with links and short descriptions of what newcomers can contribute.

**Describe alternatives you've considered**

I'm not sure if this is an alternative, but the ECH is in general trying to ramp up its ability to help people find their place, and while I don't think this is a stated goal of the Magicians, I believe to be well within their purview, too.

I'd like to assign this to myself, as I have the material pretty much ready (as a rough draft for feedback).

wschwab:
I'd add a `wip` label (re the contributing guidelines), but I don't think I have the permissions.

samajammin:
Thanks for raising this @wschwab! I think this is a great idea & I know [our team](https://github.com/ethereum/ethereum-org-website#-how-are-decisions-about-the-site-made) has had conversations with many community members around this as well. 

I feel like this page may be very similar to this "community page" idea in #269 (which I believe is in the works, @jjmstark may be able to help you get involved). Or do you think this is a separate idea that needs a distinct page?

wschwab:
I think this is a separate idea. I'm looking for a good way to describe why, so apologies if this isn't too clear.

The community page idea (which is awesome, btw) focuses on community resources for interaction about the platform - you can ask questions or post answers on Stack, have a conversation on Reddit, go to a meetup, etc. The page we're proposing comes to answer one question: How can I contribute something to the ecosystem? As such the focus is less on where to interact/consume content, but rather what can be done. Examples of resources mentioned are [Ethereum Challenges](https://challenges.ethereum.org) and MolochDAO.

One concrete example of the difference might be that the Community page talks about attending a meetup, whereas the 'Get Involved' page would talk about hosting one.

@samajammin Is that clear? I have a rough draft that I could push if it would make things clearer, but I figured that I shouldn't PR out of the blue.

samajammin:
[...]

Issue #269, opened 2019-09-29
Add a "Community" page


samajammin:
## Expected Behavior

Create a page that displays community resources (i.e. ethereum.org/community) to provide tangible ways for visitors to connect with others & get involved in the Ethereum community.

## Current Behavior
Page does not exist

## Possible Solution
Something similar to [React's community page](https://reactjs.org/community/support.html)
Possible areas to cover:
- Online communities (forums, chat servers, StackOverflow topics)
- Events
- Meetups
- Media (podcasts, news outlets, social accounts)

## Context

Discussed with https://github.com/ethereum/ethereum-org-website/issues/260

## [UPDATE: 11/25]
Page draft here: https://paper.dropbox.com/doc/Ethereum.org-Community-Subpage--ApSbLBst88Orfx1kqLOJx2IuAg-T7PZSn1jFvEnSVwE4co3h

pr0toshi:
Really hope this gets through. The .org is the reference for anyone new and would be the first place many go to find out about the project. ethereum being about open and collaborative community it's off to not promote that community.

jamesconnolly93:
I can take a crack at putting this together this week if there isn't someone working on it already.

pr0toshi:
Got one that I was thinking of just converting that was going to be for a knowledge base project on eth but could just generalize to the space. Got optional channels and all that up. 

https://invite.gg/branch

samajammin:
I started a basic sheet to organize info on various Ethereum communities, feel free to add to it! https://docs.google.com/spreadsheets/d/1fRpwpbAUYgbIGjC_rqmYR9TYW5sTuRdACTb2RopjLko/edit#gid=0

jamesconnolly93:
Awesome. One thought - just looking at the list of events, should we maybe just point people to Meetup as a place to search for events rather than trying to vet this list? Could be hard to manage over time as people tag new events with 'Ethereum' on Meetup. 

Just on a cursory glance there are EOS events etc. that are on there but not really related to Ethereum. Trying to curate this list seems like a challenging ongoing task.

samajammin:
Completely agree. I suppose I wasn't clear about that 😄  

At least to start, I'm thinking we focus on the "communities" tab, which references [Ethereum Meetups](https://www.meetup.com/topics/ethereum/) as one of them. I scraped that Meetup list mostly out of curiosity. Feel free to add other categories you're interested in, although I can't promise they'll be added.

pr0toshi:
[...]

PR #625, opened 2019-12-05
Add challenges.ethereum.org

ethers:
Add challenges.ethereum.org


jjmstark:
@ethers doesn't look like this is in the right place, as you have added it to the "Layer 2" section but `challenges` isn't particular to Layer 2 tech. 

Is there another place where you think it would make sense on this page?

ethers:
@jjmstark The section is "Improving Ethereum’s Scalability" which has L1 and L2.  It looked like a section on L2, but that's only because all links for the section are at the end: it probably isn't helpful to intersperse the L1 links between the L1 and L2 descriptions?
The link to Making Sense of L2 could be brought down after the other articles on scalability to help emphasize that the section is about scalability?  Or is there a "how to contribute" or "call to action" section somewhere else on the site that I missed?

samajammin:
> Or is there a "how to contribute" or "call to action" section somewhere else on the site that I missed?

That sounds relevant to #489 (PR: #520), which I believe is planned to be incorporated into the Community page (#269). @jjmstark any update on the progress of that? If it's far off, should we consider merging #520 & adding this there in the meantime?

samajammin:
Going to accept & merge this for now. I believe adding this to the Community page (#269) would be beneficial as bounties are a great way for folks to get involved with the ecosystem. Thanks folks!

samajammin:
@all-contributors please add @ethers for content

PR #624, opened 2019-12-05
Add XL-friendly layout.

Sesamestrong:
## Description

This includes centering content Medium-style and keeping the footer always at least at the bottom of the page. See #207 for information on exactly what this implements.
Uses flexbox and ```margin:0px auto;``` to center things. Uses ```max-width``` to set the maximum width of the main content. This creates an effect similar to what Medium uses. The main content is set to be ```min-height:100vh;```, and is structured with Flexbox around a) a ```div``` holding the main content and b) the ```footer```. The ```div``` has property ```flex-grow:1;```, allowing it to grow to satisfy the ```min-height:100vh;``` requirement. This places the footer aligned at the bottom.

I used ```@media``` queries to limit width of items only when the viewport width is greater than ```1280px```. The max content width limit under XL circumstances is ```1140px```.

## Related Issue

#207

### ```1244px``` viewport width (Original site, my changes):

Original site, width <1280px
Changed site, same width

### ```>1280px``` viewport width (Original site, my changes):

Original site, width >1280px

Changed site, same width

Issue #207, opened 2019-09-15
Adjust site layout properly for XL screens


samajammin:
### Prize Bounty
50 DAI

### Challenge Description
Our current layout doesn't adjust properly for XL screens (see screenshots in comments). Update the site layout to adjust properly for XL screens.

### Submission Requirements
- [ ] Create centered fixed-width container (`1140px`) for all content (including nav & footer), similar to [Medium example below](https://github.com/ethereum/ethereum-org-website/issues/207#issuecomment-533814710) that applies at XL screen breakpoint (`$breakL = 1280px`)
- [ ]  If the content does not take up the full viewport height, position the footer at the bottom of the viewable screen. To be clear, do NOT make the footer sticky via an absolute/fixed position. Merely ensure it displays at the bottom of the screen in situations where the page content is not the full height of the screen (e.g. on the homepage for most XL screens, [see comment below](https://github.com/ethereum/ethereum-org-website/issues/207#issuecomment-558480470)).

### Judging Criteria
The prize will be awarded to the first submission of sufficient quality. "Sufficient quality" is an admittedly subjective evaluation. We're seeking a flexible solution that will adapt to any XL screen size, e.g. by using a flexbox pattern for alignment or viewport height rules rather than applying fixed pixel dimensions.

### Winner Announcement Date
We will review work within a week of submission. Payouts will occur within a week of submission acceptance.

sneg55:
I think expected is something like medium.com layout with equal sides and top menu with fixed-width after breakpoint:
![image](https://user-images.githubusercontent.com/7769371/65376723-4edf4a80-dc58-11e9-9d62-7fe8cd977549.png)

actual behavior :

![image](https://user-images.githubusercontent.com/7769371/65376745-85b56080-dc58-11e9-8386-0d49978d0da3.png)

samajammin:
Still an issue.

e.g. ethereum.org/
![wide-screen-1](https://user-images.githubusercontent.com/8097623/69602040-fbc4a380-0fca-11ea-89d9-cbda2af0e4f8.png)

e.g. ethereum.org/enterpise
![wide-screen-2](https://user-images.githubusercontent.com/8097623/69602046-0121ee00-0fcb-11ea-82fc-9028b68684cb.png)

samsuh:
![image](https://user-images.githubusercontent.com/18237703/69604460-662d1200-0fd2-11ea-920e-8bfdd324eeca.png)
[...]

PR #636, opened 2019-12-11
Add XL-friendly, configurable layout.

Sesamestrong:
## Description

This includes centering content Medium-style and keeping the footer always at least at the bottom of the page. See #207 for information on exactly what this implements.
Uses flexbox and ```margin:0px auto;``` to center things. Uses ```max-width``` to set the maximum width of the main content. This creates an effect similar to what Medium uses. The main content is set to be ```min-height:100vh;```, and is structured with Flexbox around a) a ```div``` holding the main content and b) the ```footer```. The ```div``` has property ```flex-grow:1;```, allowing it to grow to satisfy the ```min-height:100vh;``` requirement. This places the footer aligned at the bottom.

I used ```@media``` queries to limit width of items only when the viewport width is greater than ```1280px```. The max content width limit under XL circumstances is ```1140px```.

As the screen size cutoff point is ```$breakL```, I used that, and made a new config variable ```$contentWidthL```, which is equal to ```1140px```.

NOTE: This is an updated version of pull request #624.

## Related Issue

#207
#624 

### ```1244px``` viewport width (Original site, my changes):

Original site, width <1280px
Changed site, same width

### ```>1280px``` viewport width (Original site, my changes):

Original site, width >1280px

Changed site, same width


samajammin:
Thanks @Sesamestrong!

@all-contributors please add @Sesamestrong for code.

PR #728, opened 2020-01-28
Add community event

pi0neerpat:
Smart contract workshop on #defi. 

We've had regular monthly events on Ethereum dev since 2018


pi0neerpat:
You're right! I didn't even see that section. I made the change.

pi0neerpat:
No thank you!

PR #730, opened 2020-01-29
Update Remix info

ryestew:
## Description

- Updating Remix info on the /developers page ( Developer Resources)
- Updating Remix info on the /build page (Get Started)
- Adding a link to remix.ethereum.org in the navbar under Developers


yann300:
@samajammin could we get a review for this?
The main goal of the PR is to modify the `build` section to reflect that Remix online is also quite used in the community.
We can push some update if you feel that should be displayed differently, opened to any suggestions ;)

samajammin:
@CPSTL @jjmstark tagging you folks for review. Thanks!

jjmstark:
Hey @ryestew & @yann300, 

Thanks for submitting this PR! We need to reject parts of it, but other parts we think can be incorporated into the site (and in one case we've already gone ahead and done so). This PR has also given us new ideas for ways to better on-board users of ethereum.org to Remix in an appropriate way, including through Studio.

In particular, one goal we have for Studio is that it should lead users towards more full-featured tools like Remix and others. We think Studio is the best "5-minute introduction to the Ethereum stack" experience available, since it’s a very stripped-down tool and doesn’t have a lot of features. But once a user has played with studio for a few minutes, we want them to continue their exploration of Ethereum through other tools.

Adding a better “Next steps” feature to Studio has been on our roadmap, but we’re going to raise the priority of that update. The immediate change is we're going to add a “Next Steps” section at the bottom of each tutorial template on Studio, which directs users to Remix and other tools. 

We’ve [created an issue](https://github.com/SuperblocksHQ/ethereum-studio/issues/203) to make this change, and we'll be working on this over the next few days to get something shipped. We're interested in your input here, on how best to link people to specific parts of Remix (e.g. is it possible to link directly to the tutorial?). 

Back to this PR, responding to the specific changes:

**Adds Remix as a top-level item in the navigation menu under “Developers”**

- Unfortunately we need to reject this change
- The top-level navigation menu’s purpose is to help users navigate to the different pages on Ethereum.org. Those pages themselves link out to resources and tools like Remix. 
- The top-level nav shouldn’t be used for specific projects. If we did begin adding projects to the top-level nav, the menu would quickly become less functional. 
- The reason we include Studio in the top-level nav is because it was specifically commissioned for this purpose - to be an interactive component of ethereum.org that lets developers get started with a very simple interface.
- Because it has a very narrow scope, is primarily maintained by the website team, and is more properly thought of as a page on ethereum.org than an independent project, it belongs in the top-level navigation.

**Expands the “Build” page to include a large focus on Remix**

While we again have to reject this as-is, we would like to incorporate some of these ideas into an upcoming PR that improves the /build page and includes more information about Remix (and other tools):

- The “Build” page is specifically targeted at beginners
- Studio will remain the primary focus of this page because we believe it is the best “total beginner” experience that lets the user immediately start playing with the Ethereum stack
- But Studio’s scope is very limited - after studio provides a gentle introduction, we want to push users to other tools or resources appropriate for them, like coding tutorials or more full-featured IDEs like Remix
- What we’d like to do is expand the “below the fold” section to show more information about the different beginner-oriented tools on the page
    - Is there a way to link directly to the Remix “Discover Solidity” tutorial? This is probably the ideal place to point new users towards, but can be difficult to find when a user opens the default Remix interface
- We’re currently updating the design & layout of several pages, and will include these changes in an upcoming release, likely borrowing from this PR in how it describes/presents Remix

**Reorganizes the “IDE” section of ethereum.org/developers to put Remix at  the top**

- We do agree that moving the “IDE” section above the “Developer Tools” section makes logical sense, and accept that change
- In general we reject PRs that move a given project to the top of the list, vs where it exists in the list currently
- However, we agree that for the /Developer page, Studio shouldn’t be at the top of this list - since it is a narrowly targeted tool aimed at beginners. In a separate PR we’ve already re-organized this list to move Studio to the bottom of it (#769)
- As noted earlier, we’re currently in the middle of a major design update that will improve the design of this page, which may mean more re-organization and a less strictly-vertical arrangement of tools

Closing this out, but let's please continue this conversation in the issues linked above (or in new PRs/Issues that you want to raise). Again, we appreciate you taking the time to contribute here!

PR #770, opened 2020-02-20
Updated all icons to reference a single component [fixes #758]

### LINKED ISSUE #758: Update iconography

**Additional context**

))

carlfairclough:
Updated all icons to reference a single icon component

## Description

A single `<icon name="..." /> component can now be used to generate icons detailed in the style guide. These have been implemented in the header.

They accept the following props:
```
lightModeColor: String,
size: String, Number
```

`darkModeColor` has not been implemented as it is reliant on a class which exists higher in the DOM tree. To target dark mode, use the following in CSS:
`#wrapper.dark-mode .unique-class svg path { fill: pink }`

image
image
image


docs/.vuepress/components/Icon.vue
```diff
@@ -0,0 +1,121 @@
+<template>
+  <svg
```
samajammin: Where'd you get all these SVG files? Do you think this is the best approach vs. using an icon library, e.g. [Font Awesome](https://github.com/FortAwesome/vue-fontawesome)? This SVG approach feels like it could get a bit unwieldy if we add more icons down the line.

docs/.vuepress/theme/components/Header.vue
```diff
@@ -22,53 +22,61 @@
...
-          alt="Switch to Dark Mode"
-          class="hide-dark"
-          src="../images/icon-sun.svg"
-        />
-        <img
-          alt="Switch to Light Mode"
-          class="show-dark"
-          src="../images/icon-moon.svg"
-        />
+        <icon name="lightmode" class="hide-dark" />
```
samajammin: Rather than having 2 DOM elements & additional classes (`hide-dark` & `show-dark`) to handle the display logic, I'd rather just pass `darkMode` property from the Layout component into Header, then update the icon `name` based on that to present the correct icon. Make sense? You can see an example in Layout where we pass that property to the Hero component.

carlfairclough: Ah I'd missed that, I wasn't entirely sure how to handle it & I prefer what you suggest. Ty!

carlfairclough:
@samajammin I've updated the icon to handle isDarkMode, and I'm using `v-if` to display the correct icon vs hiding with CSS.

I also updated dark/darkmode props to be a type-checked isDarkMode prop (Bool). This means there have been some minor updates which reach into components like the Hero.

I've removed all unused icons, with the caveat that the list icons are now base64 encoded directly in the CSS, as it's likely those styles will removed/updated soon anyway.

docs/.vuepress/components/Icon.vue
carlfairclough: (Commenting for visibility)
These are directly from the figma file, they all share size & visual weight. It might be unwieldy in future I agree but imagine there will be minor updates now that the bulk of this is done (i.e. darkmode handling, passing colors, etc)

docs/.vuepress/components/Icon.vue
```diff
@@ -0,0 +1,128 @@
...
+      type: String,
+      required: true
+    },
+    size: {
+      type: [Number, String],
+      default: 24
+    },
+    darkModeColor: {
+      type: String,
+      default: '#FFF'
```
samajammin: Instead of the hex color `#FFF`, I'd like to see us import variables from config.styl, e.g. `$colorWhite`

docs/.vuepress/components/Icon.vue
```diff
@@ -0,0 +1,128 @@
...
+    },
+    lightModeColor: {
+      type: String,
+      default: '#000'
```
samajammin: Same here.

docs/.vuepress/theme/components/Header.vue
```diff
@@ -22,53 +23,57 @@
...
+    },
+    isDarkMode: {
+      type: Boolean,
+      default: false
+    }
+  }
 }
 </script>
 
+// unscoped css to pass styles to navigations
```
samajammin: Yay comments!

docs/.vuepress/theme/components/SearchBox.vue
```diff
@@ -153,10 +154,24 @@ export default {
 <style lang="stylus">
 @import '../styles/config.styl'
 
+#wrapper.dark-mode .search-box svg path
+  fill: $subduedColor
```
samajammin: Nice use of variable 👍

docs/.vuepress/theme/components/SidebarButton.vue
```diff
@@ -1,44 +1,47 @@
...
-    </svg>
-    <router-link
-      :to="$localePath"
-      class="home-link"
-    >
+    <a href="#" @click="$emit('toggle-sidebar')" class="flex">
+      <icon
+        name="menu"
+        :isDarkMode="isDarkMode"
+        lightModeColor="#000"
```
samajammin: Same here - would prefer variables.

docs/.vuepress/theme/styles/dark.styl
```diff
@@ -35,17 +35,15 @@
...
           color $textColorDark
           &.highlight
-            background-image url(../images/highlight-dark.svg)
+            background-image url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHhtbG5zOnhsaW5rPSJodHRwOi8vd3d3LnczLm9yZy8xOTk5L3hsaW5rIiB3aWR0aD0iMjgwIiBoZWlnaHQ9IjI4IiB2ZXJzaW9uPSIxLjEiIHZpZXdCb3g9IjAgMCAyODAgMjgiPjx0aXRsZT5GaWxsIDE8L3RpdGxlPjxkZXNjPkNyZWF0ZWQgd2l0aCBTa2V0Y2guPC9kZXNjPjxnIGlkPSJQYWdlLTEiIGZpbGw9Im5vbmUiIGZpbGwtcnVsZT0iZXZlbm9kZCIgc3Ryb2tlPSJub25lIiBzdHJva2Utd2lkdGg9IjEiPjxnIGlkPSJMYW5kaW5nLUFsdC1BLUNvcHkiIGZpbGw9IiM0RjRGNEYiIHRyYW5zZm9ybT0idHJhbnNsYXRlKC0yODUuMDAwMDAwLCAtODgxLjAwMDAwMCkiPjxnIGlkPSJHcm91cC0zIiB0cmFuc2Zvcm09InRyYW5zbGF0ZSgyNzUuMDAwMDAwLCA4MjUuMDAwMDAwKSI+PHBhdGggaWQ9IkZpbGwtMSIgZD0iTTIwLjYzNDY3MzEsODMuOTY2MzA3NSBDMTA2Ljc2MDc1Nyw3Ny44NTI1Njg0IDE5My4wNTUwMzgsNzYuNTA1MjQ4NyAyNzkuMzI3MTg5LDc5LjkzMzE4MzQgQzI5My41NzA4NzksODAuNDk4NjE1OSAyOTMuNTQ0MzIyLDU4LjQxMTQwODMgMjc5LjMyNzE4OSw1Ny44NDU5NzU4IEMxOTMuMDU1MDM4LDU0LjQxODA0MTEgMTA2Ljc2MDc1Nyw1NS43NjUzNjA4IDIwLjYzNDY3MzEsNjEuODc5MDk5OSBDNi41MTA0OTExMyw2Mi44ODE4NTkxIDYuMzk5ODM0NzMsODQuOTc3OTAxNiAyMC42MzQ2NzMxLDgzLjk2NjMwNzUiLz48L2c+PC9nPjwvZz48L3N2Zz4=")
```
samajammin: Why are we doing this?

docs/.vuepress/components/Icon.vue
carlfairclough: We can't pass a stylus variable into this but we *can* pass CSS variables. I'm happy to go with CSS vars, but it's a departure from the current system

PR #794, opened 2020-03-04
Update Navigation & Search

carlfairclough:
Updated navigation & search components

## Description

A pretty in-depth update of navigation and search. Involves modal-style overlays and navigation within the menu sidebar. It also involves extras
- Removed navigation duplication from the `Outline`
- No separate sidebar component
- All navigation display is managed within CSS, with visibility props being managed via CSS
- Search edge states catered for
  - Blank
  - No results

## Related Issue
#764 #755

#### Nav Items
image
image

#### Search
Results
image
image

Blank States
image
image
image

### Mobile
image
image
image
image
image
image


carlfairclough:
@samajammin here's a draft PR on this. It's not ready to be merged IMO because the way nav links are being styled is pretty messy and should be done inside the `NavLink.vue` component.

I'm pausing here because it's a huge change already I would appreciate review to catch any glaring errors before I go and make that CSS manageable.

I'm in particular looking for attention around how I'm managing properties/data with $emit

docs/.vuepress/theme/components/Navigation.vue
```diff
@@ -0,0 +1,113 @@
...
+      isMobileNavVisible: false
+    }
+  },
+  computed: {
+    darkOrLightModeIcon() {
+      return this.isDarkMode ? 'darkmode' : 'lightmode'
+    }
+  },
+  methods: {
+    handleNavToggle(value) {
```
samajammin: I believe this can be simplified. Since this is merely toggling a boolean, we don't need to pass in a value. Test changing this to:
```
    handleNavToggle() {
      this.isMobileNavVisible = !this.isMobileNavVisible
    }
```

samajammin: Then any instances of emitting the toggle can be updated from:
```$emit('nav-toggle', false)```
To
```$emit('nav-toggle')```

File: docs/.vuepress/theme/Layout.vue [hunk 2]
```diff
@@ -16,7 +14,7 @@
           </p>
           <Content />
         </main>
-        <Sidebar :items="sidebarItems" @close-sidebar="closeSidebar" />
+        <Sidebar :items="sidebarItems" />
```
samajammin: Is the `Sidebar` component even used / needed now? It appears not (removing it keeps all functionality, from what I can tell). If so, let's delete it.

samajammin: I believe that would also mean we could delete `SidebarGroup`, `SidebarLink` & `SidebarButton` components.

docs/.vuepress/theme/styles/config.styl
```diff
@@ -96,6 +96,10 @@ $lhRegular = 1.2em
 $lhMedium = 1.5em
 $lhLarge = 1.8em
 
+
+$fsl3 = 1.5rem
```
samajammin: Let's make these variables a bit more descriptive

docs/.vuepress/theme/styles/config.styl
```diff
@@ -114,3 +118,11 @@ $contentWidthXL = 1440px;
 // code
 $lineNumbersWrapperWidth = 3.5rem
 $codeLang = js ts html md vue css sass scss less stylus go java c sh yaml py
+
+.reset-type
```
samajammin: This config.styl is really just for variables - I'd say put this class in theme.styl if you need a global class. It looks like this class is only used in SearchBox.vue though, why not just place it there?

docs/.vuepress/theme/NotFound.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div id="wrapper" :class="pageClasses">
-    <Header @toggle-mode="toggleMode" />
+    <Navigation @toggle-mode="toggleMode" />
```
samajammin: Let's change this to `dark-mode-toggle` to match your changes to Layout.vue - I think this is broken right now.

docs/.vuepress/theme/components/Navigation.vue
```diff
@@ -0,0 +1,113 @@
...
+        class="icon-menu"
+        @click.native="handleNavToggle(true)"
+      />
+    </span>
+  </div>
+</template>
+
+<script>
+import NavLinks from './NavLinks.vue'
+import SearchBox from './SearchBox.vue'
```
samajammin: Don't think this import is needed - you don't use the `SearchBox` component in this file?

docs/.vuepress/theme/components/SearchBox.vue
```diff
@@ -40,17 +59,61 @@
...
-      focused: false,
       focusIndex: 0
     }
   },
 
+  watch: {
+    isSearchFocused: function() {
+      // watch it
+      this.isSearchFocused &&
+        document.getElementById('main-search-field').focus()
```
samajammin: Could we solve this with CSS instead? E.g. adding a conditional class :class="{ focused: isSearchFocused }"

docs/.vuepress/theme/components/SearchBox.vue
```diff
@@ -1,35 +1,54 @@
...
+        @blur="$emit('focus-toggle', false)"
+        @keyup.enter="go(focusIndex)"
+        @keyup.up="onUp"
+        @keyup.down="onDown"
+        placeholder="Search"
+      />
+      <icon name="search" class="icon-search-field" />
+    </div>
+
+    <div v-if="blankState && isSearchFocused" class="blank-state">
```
samajammin: Should this only display if search is focused? Seems we could display this whether or not the search input is focused.

docs/.vuepress/theme/components/NavLinks.vue
```diff
@@ -35,35 +135,222 @@ export default {
     }
   },
   methods: {
-    isActive
+    isActive,
+    handleSearchToggle(value) {
```
samajammin: Similar to `handleNavToggle`, I think this could be simplified to just toggle a boolean without passing a value.

docs/.vuepress/theme/components/NavLinks.vue
```diff
@@ -35,35 +135,222 @@ export default {
...
+      this.isSearchVisible = value
+    },
+    handleFocusToggle(value) {
```
samajammin: Similar to `handleNavToggle`, I think this could be simplified to just toggle a boolean without passing a value.

docs/.vuepress/theme/components/SearchBox.vue
```diff
@@ -1,35 +1,54 @@
...
+      v-if="!blankState && isSearchFocused"
       class="suggestions"
-      v-if="showSuggestions"
       :class="{ 'align-right': alignRight }"
-      @mouseleave="unfocus"
     >
       <li
         class="suggestion"
         v-for="(s, i) in suggestions"
         :class="{ focused: i === focusIndex }"
```
samajammin: You deleted the `.focused` class, so you can't see which search result is focused.

Current:
Image 2020-03-05 at 7 45 44 AM

Your version:
Image 2020-03-05 at 7 45 24 AM

Let's bring this back :)

samajammin:
Another styling note - let's ensure the search doesn't overlap the sidebar (i.e. keep the current functionality).

Current:
Image 2020-03-05 at 7 49 57 AM

Your PR:
Image 2020-03-05 at 7 49 51 AM

samajammin:
A couple style notes on the desktop nav. 
1) The nav icon is now too small & no longer center aligned
2) The links feel awkwardly far apart

Style guide:
Image 2020-03-05 at 7 52 43 AM

Your PR:
Image 2020-03-05 at 7 53 58 AM

samajammin:
@carlfairclough what is the `Outline` you're referring to in the PR description?

docs/.vuepress/theme/components/Navigation.vue
```diff
@@ -0,0 +1,113 @@
...
+      :handleNavToggle="handleNavToggle"
+      @nav-toggle="handleNavToggle"
+      @dark-mode-toggle="$emit('dark-mode-toggle')"
+    />
+
+    <span class="icon-link icon-menu" tabindex="0">
+      <icon
+        name="menu"
+        class="icon-menu"
+        @click.native="handleNavToggle(true)"
```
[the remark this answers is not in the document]
carlfairclough: I'm not sure this would make a difference because of the z-index of the overlay. I'd be more inclined to hide the menu icon when the menu is open

docs/.vuepress/theme/components/NavLink.vue
```diff
@@ -1,16 +1,20 @@
...
+    @click.native="$emit('nav-toggle', false)"
     >{{ item.text }}</router-link
   >
   <a
-    v-else
+    v-else-if="!item.hideMobile"
     :href="link"
-    class="nav-link external"
+    :class="childClass"
+    v-on:click.native="clickedItem(), $emit('nav-toggle', false)"
```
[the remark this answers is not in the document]
carlfairclough: I think it was a hangover from me trying `v-on:click` and `v-on:click.native`

docs/.vuepress/theme/components/Navigation.vue
```diff
@@ -0,0 +1,113 @@
...
+      <img
+        class="header-logo"
+        src="../images/ethereum-logo-wireframe.png"
+        alt="Ethereum Logo"
+      />
+    </router-link>
+    <NavLinks
+      :isDarkMode="isDarkMode"
+      :isMobileNavVisible="isMobileNavVisible"
+      :handleNavToggle="handleNavToggle"
```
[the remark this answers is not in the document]
carlfairclough: `:handleNavToggle` is used to toggle the nav visibility
`:isMobileNavVisible` is used to add or remove the classes which show the nav
`:isDarkMode` is used to show the correct darkmode/lightmode icon

docs/.vuepress/theme/components/Navigation.vue
carlfairclough: Certain buttons only do specific things, so I'm averse to toggling the value. We'd have to then likely have to monitor the viewport width  to prevent anything clashing as a result of the nav toggling
- When a nav link is clicked, the nav should always hide
- When the close icon is clicked, the nav should always hide
- When the menu icon is clicked, the nav should always open

docs/.vuepress/theme/styles/config.styl
carlfairclough: Removed it completely :)

carlfairclough:
> Another styling note - let's ensure the search doesn't overlap the sidebar (i.e. keep the current functionality).

I'm not sure I agree with this, i don't think this negatively impacts anything and is impacted by the icons/text to the right

On the focus state & classes: I'm exploring this, it focuses/tabs fine with the `@blur` function is removed, but then escaping the view becomes difficult (either detect a click outside the element, which would introduce complexity or a slight restructure of the search which matches the dropdown, relying on `:focus-within` with CSS. I think `:focus-within` is the best option cause we can then remove some of the JS (i.e. isSearchFocused)

docs/.vuepress/theme/components/SearchBox.vue
carlfairclough: This is resolved — now using :focus-within instead of JS

docs/.vuepress/theme/components/SearchBox.vue
carlfairclough: This isn't a css thing, this targeting the input field:
1. User taps "search" on mobile
2. Search opens, with the input focused and as a result their keyboard opens.

I've had to switch to a `display: none` toggle for the search, which introduces issues with this anyway as it prevents the field from being focused. As a result I've stripped all of this out.

On a related note, something like `:class="{ focused: isSearchFocused }"` is now being used. The css `focus-within` selector isn't incredibly well supported so JS is handling this again (without significant changes):
```
searchClasses() {
      return {
        'search-box': true,
        'search-hidden': !this.isSearchVisible,
        'focus-within': this.focused
      }
    },
```

docs/.vuepress/theme/components/SearchBox.vue
carlfairclough: This was a desktop-first approach and has been changed to be hidden with CSS instead when the search isn't focused

docs/.vuepress/theme/components/NavLinks.vue
carlfairclough: I'd prefer to keep this, it allows us to clear the search in certain scenarios i.e. when a link is clicked. imo it's much better to be explicit when there's a desired behaviour

docs/.vuepress/theme/components/Navigation.vue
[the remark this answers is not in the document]
carlfairclough: This has been updated to hide when the nav is open

carlfairclough:
I believe this is ready for merge @samajammin. I've addressed most of your comments & replied to the ones which I haven't actioned with explanation

carlfairclough:
I've been testing the search in Safari (desktop) and come across an issue where blur prevents the `@click`, `@mousedown`, `@mouseup` events from firing, whether that's on the links themselves (`@blur`) or the wrapper (`@focusout`). I'm at a loss as to the cause here, and every solution seems to surface another problem with it.

The last resort would be to only show results when the input is focused (existing behaviour) but imo that is less than ideal, so any help would be massively appreciated.

docs/.vuepress/theme/components/NavLinks.vue
samajammin: I don't follow - `it allows us to clear the search in certain scenarios i.e. when a link is clicked` - where are we doing this? From what I understand this merely sets `isSearchVisible` to true or false.

docs/.vuepress/theme/components/Navigation.vue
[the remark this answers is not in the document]
samajammin: Right but your passing `handleNavToggle` to the `NavLinks` component, which doesn't do anything with that prop. I removed this line locally & it doesn't seem to change any 
functionality.

docs/.vuepress/theme/components/Navigation.vue
samajammin: Right but the nav links & close icon are only visible when the nav is visible & the menu icon is only visible when the nav is closed. So the toggling boolean logic works, no viewport monitoring necessary 😃

samajammin:
@carlfairclough approved! I'll merge this after you rebase against the latest `dev` branch. Holler if you want some quick assistance on that.

samajammin:
@carlfairclough - regarding your comment:

> I've been testing the search in Safari (desktop) and come across an issue where blur prevents the @click, @mousedown, @mouseup events from firing, whether that's on the links themselves (@blur) or the wrapper (@focusout). I'm at a loss as to the cause here, and every solution seems to surface an issue in behavior (e.g. focus not firing).

So what's the impact from the perspective of a user here?

docs/.vuepress/theme/components/Navigation.vue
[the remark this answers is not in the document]
carlfairclough: It closes the mobile nav when a link is clicked

docs/.vuepress/theme/components/NavLinks.vue
carlfairclough: Switched this out for the toggle like you mentioned — I honestly prefer being super explicit but I'll go with your judgement

carlfairclough:
I think we're good to go unless there are any blunders @samajammin

docs/.vuepress/theme/components/Navigation.vue
carlfairclough: Just to return to this conversation, not being explicit here toggles the overflow on desktop too, and to keep things simple I've reintroduced the bool. Not having this would prevent the user from scrolling

carlfairclough:
@samajammin resolved the Safari issue & can tab (tab key) through search results

samajammin:
@carlfairclough awesome!!! Please resolve the merge conflicts & I'll work to get this merged ASAP.

PR #800, opened 2020-03-06
Tidied translations directory structure

carlfairclough:
Updated translation structure. All translations now live in the `translations` directory *without* affecting the permalink.

## Description

This is not a change in functionality, it is instead a change in how the content directory is structured, with all content living in `translations`. The reason for doing this is to clearly delineate english-language content from other content, allowing us to clearly separate concerns given the ongoing work with content transposition.

This is enabled with a small addition to `.vuepress/config.js`, which removes the `translations` part of the path
```
extendPageData($page) {
    if ($page.path.split('/')[1] == 'translations') {
      $page.path = $page.path.replace('/translations/', '/')
      console.log('translated', $page.path)
    }
  },
```


samajammin:
I dig! Thanks for doing this!

PR #814, opened 2020-03-10
Remove unused classes

samajammin:
## Description
This `home` class is no longer used

## Related Issue
N/A


samajammin:
@carlfairclough please check me on this but I don't believe this class affects any styles

carlfairclough:
I don't see any problems with this 👍

PR #842, opened 2020-03-18
fix firefox page padding [fixes #833]

### LINKED ISSUE #833: Pages with Sidebar present lost their margin on Firefox
**Describe the bug**

Pages with Sidebar present lost their margin - see /developers & /learn

**To Reproduce**
Navigate to /developers or /learn

**Expected behavior**

Should have standard page margins

**Screenshots**

Image 2020-03-17 at 11 00 12 AM

 - OS: MacOS
 - Browser: Firefox
 - Version: 73.0.1

**Additional context**

I believe introduced in #794, as it's not yet a bug on production.

carlfairclough:
This provides a fallback for browsers which do not read min() max() css funtions.

## Description

A stylus function has been created which will allow for linear scaling between two values.

34px @ 320px viewport → 72px @ 768px viewport
`cssGrad(320, 34, 768, 72)`

This function must be followed by a media query at the upper and lower limits, to prevent runaway scaling:

```stylus
.el
  width cssGrad(320, 34, 768, 72)
  @media screen and (min-width: 768px)
    width: 72px
```

A fixed-px fallback has been added for the list blank states, as that was previously using min().

## Related Issue
#833


docs/.vuepress/theme/styles/responsive.styl
```diff
@@ -6,10 +6,16 @@
   .page 
     width 100%
     max-width 1280px
+    padding-left cssGrad(320, 12, 1280, 96)
```
samajammin: Seems this function would be more useful if it didn't require the media query. Since you have to add the @media query anyway, is there any difference in functionality than having a `padding-left 12px` here?

samajammin: If so, I'd rather just put that vs. obscuring what's going on.

carlfairclough: I'd add that as a fallback, as the vw tag helps to keep things fluid (imo).
I have a much stronger preference for fluid sizing vs things being fixed and resizing at breakpoints

carlfairclough: Just added a fallback

carlfairclough:
@samajammin This doesn't happen when in dev so I'm assuming it's a part of the stylus build process, perhaps ensuring that there is only one of each property per class?

I've admitted defeat — this is using media queries now

PR #839, opened 2020-03-18
Lists [fixes #767]

carlfairclough:
Added lists & documentation for them Implemented simple list on `/developers` page

## Description

Added multiple list variations based on the Figma Style Guide. These include lists with descriptions, and list cards in multiple layouts.

An array of objects should be passed to leverage the card format, and a class should be added to the standard lists to enable their modification.

## Screenshots (if appropriate):
Style Guide Documentation — Lists   Ethereum org
Style Guide Documentation — Lists   Ethereum org

PR #843, opened 2020-03-18
Card components [fixes #765]

### LINKED ISSUE #765: Create card component

**Additional context**

))

carlfairclough:
Adds a card and card list component

### Cards

Cards can now be created which accept the following:
- `level: optional, Number, default: 3` defines the `h#` tag for the card

- `header: optional, String` pass either a link to an image, or an emoji here
- `hero: optional, Boolean, default: false` if an image is passed to the card, this will display it as a hero
- `leftimg: optional, Boolean, default: false`  if an image is passed to the card, this will display it in a left column
- `title: required, String` the title
- `content: required, String` the text under the subtitle
- `link: optional, Object`
  - `to` the url
  - `text` the text to display
- `button: optional, Object or Array of 2 items`
  - `to` the url
  - `text` the text to display
-  `clickable: optional, Boolean, default: true` This toggles whether the card is clickable

`link` and `button` are mutually exclusive
`leftimg` and `hero` are mutually exclusive

### Card List
- `items: required, Array` an array of objects with card properties
- `level: optional, Number, default: 3` defines the `h#` tag for the cards
- `desktopColumns: optional, default: 3` the number of columns on desktop
- `tabletColumns: optional, default: 3` the number of columns on tablet
- `smallTabletColumns: optional, default: 2` the number of columns on small tablets/large phones
For small mobile devices, one column is always displayed.

## Related Issue

#765 

Start Building   Ethereum org
Start Building   Ethereum org-dark


docs/.vuepress/components/Card.vue
```diff
@@ -0,0 +1,106 @@
...
+      type: [Object, Array]
+    },
+    clickable: {
+      required: false,
+      type: Boolean,
+      default: true
+    }
+  },
+  computed: {
+    isExternal() {
```
samajammin: Similar to my comments on #839 - let's use the existing `isExternal` helper function in util.js instead of this. That way, we don't need to pass a boolean around.

docs/.vuepress/components/Card.vue
```diff
@@ -0,0 +1,106 @@
+<template>
+  <a v-if="isExternal && clickable" :href="getLink" class="hide-icon flex">
```
samajammin: Let's always add `target="_blank" rel="noopener noreferrer"` to external links.

docs/.vuepress/components/Card.vue
```diff
@@ -0,0 +1,106 @@
...
+    },
+    content: {
+      required: true,
+      type: String
+    },
+    link: {
+      required: false,
+      type: Object
+    },
+    button: {
```
samajammin: Can we just pass a boolean, i.e. `isButton` to determine if the link should be a button or not? Seems simpler vs. passing this object. Or am I missing something?

carlfairclough: yea you're right, good catch

Issue #844, opened 2020-03-18
Language card external links display multiple arrows


samajammin:
**Describe the bug**

Language card external links display multiple arrows

**To Reproduce**
See /languages page

**Expected behavior**

Should only display 1 arrow

**Screenshots**

![Image 2020-03-18 at 8 35 09 AM](https://user-images.githubusercontent.com/8097623/76978371-d6e69a00-68f3-11ea-9bc6-176dc26a8baf.png)

**Desktop (please complete the following information):**
All

**Smartphone (please complete the following information):**
All

**Additional context**

I believe introduced in #813

samajammin:
FYI @carlfairclough

samajammin:
Fixed in #846

PR #846, opened 2020-03-19
Hide 2nd external link icon [fixes 844]

carlfairclough:
Added hide icon class to language card

## Description

The nested `a` tags with external hrefs were causing two external link icons to be displayed. The `hide-icon` class has been added to the wrapper to hide the 2nd.

## Related Issue

#844


carlfairclough:
I can't test this locally but it *should* be fixed.

PR #849, opened 2020-03-19
added keyboard up + down arrow nav for navigation and search

carlfairclough:
You can now use arrows to navigate through the menu drop downs and the search results

## Description

Added `up()` and `down()` functions to `SearchBox.vue` and `NavDropdown.vue` which ensure that keypresses & focus are only triggered when relevant.

### DIFF
--- a/docs/.vuepress/components/CardInner.vue
+++ b/docs/.vuepress/components/CardInner.vue
@@ -118,7 +118,7 @@ export default {
 }
 </script>
 
-<style lang="stylus">
+<style lang="stylus" scoped>
 @import '../theme/styles/config.styl';
 
 .header
--- a/docs/.vuepress/components/CardList.vue
+++ b/docs/.vuepress/components/CardList.vue
@@ -51,7 +51,7 @@ export default {
 }
 </script>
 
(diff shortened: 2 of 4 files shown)


carlfairclough:
Managed to reproduce. I think multiple `keydown.enter`s were interfering with each other. Using just one caused some secondary issues resolved by moving to `keyup.enter`, but then search was being cleared on keypress. `key___.enter` `$emit` functions have been moved into `forceUnFocus()` to ensure that they aren't fired when the `input` is focused.

samajammin:
@carlfairclough nice! Seems to be working smoothly now.

I have no idea how this PR may effect it but I'm noticing some weird wrapping/spacing issues on the Sidebar, which don't appear in lastest `dev`: https://share.getcloudapp.com/ApurvpLP

Please look into fixing this before we merge. Thanks!

carlfairclough:
The cards merge was clashing as those styles weren't scoped and things have been out together with the recent merges. Cards are now scoped so you shouldn't see any issues
https://5e73c43950446c0008fe4d45--ethereumorg.netlify.com/developers/

Issue #853, opened 2020-03-20
Add dapp explanation on /dapps page


samajammin:
**Is your feature request related to a problem? Please describe.**

We've gotten feedback from users that it'd be valuable to add a brief explanation on the "what" & "why" of the /dapps page. Currently we just list some example dapps to use.

**Describe the solution you'd like**

Add a brief intro paragraph explaining what dapps are & how they're different.

**Describe alternatives you've considered**

**Additional context**

VoltronDeveloper1:
You all are so far away from the true power its unbelievable. Its elementary my dear Watson

samajammin:
Solved in #854

PR #854, opened 2020-03-20
Add dapp explanation

samajammin:
## Description
Add a brief "What is a dapp?" intro on the /dapps page explaining what dapps are & how they're different.

## Related Issue
#853

dapp-content

## Additional context

Adds citation to OpenZeppelin's doc where most of the content is from. See relevant discussion here: https://github.com/OpenZeppelin/docs.openzeppelin.com/issues/155#issuecomment-597881924

### DIFF
--- a/docs/dapps/index.md
+++ b/docs/dapps/index.md
@@ -14,6 +14,10 @@ lang: en-US
 
 **Want to get started using Ethereum?** There are hundreds of decentralized applications (dapps) built on Ethereum that you can use today. {.l4 .featured}
 
+## What is a dapp? {#what-is-a-dapp}
+
+How do Ethereum applications differ from traditional web applications? Decentralized applications (dapps) on Ethereum are web applications backed by Ethereum smart contracts. Instead of using a centralized server or database, these applications rely on the blockchain as a storage backend. This leads to potentially unstoppable applications: anyone can spin up a copy of the frontend, and freely connect it to the public Ethereum network.
+
 ## How to use an application built on Ethereum {#how-to-use-an-application-built-on-ethereum}
 
 The best way to get started with Ethereum is to jump right in!
@@ -33,3 +37,5 @@ Interested in looking at more applications built on Ethereum?
 Some Ethereum applications require a wallet - [learn more about Ethereum wallets here](/wallets/).
 
 Some Ethereum applications will require ETH - [learn more about ETH here](/eth/).
+
+<small>Page content provided in part by [OpenZeppelin](https://docs.openzeppelin.com/learn/building-a-dapp).</small>
--- a/docs/dapps/index.md
+++ b/docs/dapps/index.md
@@ -10,7 +10,7 @@ meta:
 lang: en-US
 ---
 
-# Using Ethereum {#using-ethereum}
+# Use Ethereum Applications {#use-ethereum-applications}
 
 **Want to get started using Ethereum?** There are hundreds of decentralized applications (dapps) built on Ethereum that you can use today. {.l4 .featured}
 
--- a/docs/dapps/index.md
+++ b/docs/dapps/index.md
@@ -38,4 +38,4 @@ Some Ethereum applications require a wallet - [learn more about Ethereum wallets
 
 Some Ethereum applications will require ETH - [learn more about ETH here](/eth/).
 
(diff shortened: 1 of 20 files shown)


jjmstark:
I think the idea of "quoting" other people's explanations of key ideas / concepts is a really useful one:

- It's a good middle ground between "we want to be a portal that pushes the user out into the ecosystem" and "it's most convenient if the user can get what they need on the page they are already on"
- it's a valuable way to signal to the user "hey, you really should check out the source for this insightful quote you just read"

In this case though, I'm not sure this is the right quote to use as our definition of a dapp:

- This definition would include many dapps, but doesn't really fit most DeFi applications, where the point of using Ethereum is not as a "storage backend".
- While I know that people's definitions of "dapp" differs (and some would probably argue that dapps refers to non-financial use-cases of Ethereum), this page is really about "Applications built on Ethereum" than any particular subset. For instance, we obviously already list DAI on this page and will continue to include DeFi apps
- One way we could still use this quote is to include it further down on the page, as one of a few different definitions of what being "built on ethereum" actually means

samajammin:
@jjmstark thanks for the feedback! I've tweaked the definition to include application logic in addition to storage, please let me know what you think. To be clear, this was never a _direct_ quote from the OpenZeppelin docs but it does borrow heavily from their wording, so thought it was fitting to credit them with a citation.

I also like the idea of having multiple definitions of Ethereum dapps, as it's a hard concept to nail down with a single definition but I feel like this is a good first swing at it that is better than the existing content (of no definition).

samajammin:
@carlfairclough I'd appreciate your input on a citation format. Thoughts on how 

From this PR (ignoring the emoji arrow that's been fixed):
Image 2020-03-30 at 1 41 44 PM
It's currently just a <small>, italicized text at the bottom of the markdown page:
```
_<small>Page content provided in part by [OpenZeppelin](https://docs.openzeppelin.com/learn/building-a-dapp).</small>_
```

carlfairclough:
Would it be useful to have a clear callout section here for OpenZeppelin (without it being an ad)?

However for now I think this works, unless there is an original source in which case I would explicitly cite that. Two things I'd look at are [the <cite> element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/cite) and [the creative commons attribution guide](https://wiki.creativecommons.org/wiki/best_practices_for_attribution). The CC attributions are terse but are a good outline.

image

image

samajammin:
Thanks @carlfairclough - I updated to use `<cite>` & reference the article's title & author. I think that's a solid pattern to adopt for now.

jjmstark:
Reviewed and reflected on this more, and I'm not sure we should ship this yet. A few concerns:

- The user persona for this page leans towards "beginners", or at least people who do not have much pre-existing knowledge of Ethereum. 
- However, this definition relies on the reader understanding various technical concepts to make sense of: "frontend", "backend", "program logic", "storage", "spin up a copy", etc. 
- I think this is a quite good explanation of the technical difference for a developer audience, but I think it's very opaque to anyone who is not a developer. 

I don't want us to stop iterating on this though - here's a draft of a different version that I think addresses these concerns:

--

## What makes Ethereum apps different?

Applications built on Ethereum ("dapps") can do things regular apps can't:

- Create new kinds of money and digital assets
- Web apps that are unstoppable and uncensorable
- Build decentralized organizations, property, or virtual worlds that are governed collectively

-- 

If we wanted to keep the paragraph you've written @samajammin, we could insert it after those bullet points (after all, while this page is aimed at beginners, devs are obviously part of that audience). However we could also save it for a part of the site that is targeted more squarely at developers.

samajammin:
Good points @jjmstark. I've updated the copy with your changes added. I agree the explanation is a bit technical but I don't have a better alternative at the moment. I feel like explaining the differences between Ethereum dapps & traditional web apps ultimately requires a baseline technical understanding of how a traditional web app functions.

I'm going to merge this. It seems like we're good with the copy additions & attribution approach but I'm always open for additional changes. Thanks!

VoltronDeveloper1:
I Algorithm C  Proccessor. O  Devbug. White  A  Papers  I  Roadmap. Send A Billion  C  Silo.   ORoyal I  Flush. N Team. Iz wide shut. IHIVE. Coin.

Issue #832, opened 2020-03-17
Headers are hidden when navigating to anchor links


samajammin:
**Describe the bug**

Looks to be a regression of #292 

When navigating to a page section via an anchor link, the nav covers the header.

**To Reproduce**
- e.g. site search: https://share.getcloudapp.com/5zuyPRWG
- e.g. sidebar navigation: https://share.getcloudapp.com/KouW1RGj

**Expected behavior**

Header of section should be visible

**Screenshots**

- e.g. site search: https://share.getcloudapp.com/5zuyPRWG
- e.g. sidebar navigation: https://share.getcloudapp.com/KouW1RGj

**Desktop (please complete the following information):**
All browsers

**Smartphone (please complete the following information):**
All browsers

**Additional context**

Not sure if this was introduced in #801 or #794

samajammin:
FYI @carlfairclough

samajammin:
@carlfairclough looks like there was a regression. I'm seeing this issue again in latest `dev`: https://share.getcloudapp.com/GGuN5boJ

samajammin:
This was solved in #859

PR #859, opened 2020-03-21
updating markdown text handling, isolated text styles [fixes #858 #832]

### LINKED ISSUE #858: Header permalinks do not always work
It appears that some permalinks are not working.

**To Reproduce**
1. Visit https://ethereum.org/wallets/
2. Click the `What is an Ethereum wallet, and which one should I use?` permalink icon
3. See that nothing happens — the url is followed by `%3F` which is not in the heading id

**Expected behavior**
The link should go to the heading

carlfairclough:
Update markdown handling & reduce specificity of markdown styles.

## Description

This replaces the way that headings are being handled in markdown, and moves to one layer of css specificity for markdown-generated content. In the current format, text styles are sometimes applied to child elements inside components.

This also means that the existing permalink handling has been replaced (resolving #832) and that the permalink targets have padding (resolving #858)

## Related Issue

#832 #858


docs/.vuepress/config.js
```diff
@@ -77,7 +77,38 @@ module.exports = {
...
+        'l' + tkn.tag.substr(-1),
+        anchor,
+        // Add classes here
+        'tc-text-500'
+      ].join(' ')
+      return buildTag(tkn, slf, classes)
+    }
+    r.paragraph_open = (tokens, idx, options, env, slf) =>
+      buildTag(tokens[idx], slf, 'l7 tc-text300')
+    r.paragraph_close = () => '</p>'
```
samajammin: Why do you need to close the paragraph but nothing else?

carlfairclough: For some reason , not including the close tag will mean it doesn't render. Super weird little bug, I should probably report it to markdown-it

PR #881, opened 2020-03-26
updated build page (wip)

carlfairclough:
Updated the build page. Translations aren't included here but it is a restructure of content which would be good to get feedback on, and advice on what necessary steps should be made to release

image
image

### DIFF
--- a/docs/.vuepress/components/BuildPage.vue
+++ b/docs/.vuepress/components/BuildPage.vue
@@ -1,122 +1,75 @@
 <template>
-  <div class="start-building center flex flex-column flex-center page">
-    <header class="center flex flex-column flex-center">
-      <div class="headline-text-container">
-        <h1 class="l2 tc-text500">
-          {{ translateString('page-build-title') }}
-        </h1>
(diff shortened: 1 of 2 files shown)


samajammin:
A few thoughts here:

1. Ethereum Studio gif
Image 2020-03-30 at 10 12 18 AM
I like this new layout but moving the gif to the bottom means that now there’s no color/image above the fold, which I think makes it worse. I think having the gif up top is the most helpful visual to get a quick sense of what Ethereum Studio is. 

2. Banner
Image 2020-03-30 at 10 15 39 AM
I like the look of this banner callout! I think it'd be more appropriate to link to the Developers resource page given that this page is targeted at devs.

3. Learning resources
Image 2020-03-30 at 10 17 18 AM
I think this sections makes the most sense near the bottom (below the gif & the templates) as a next step AFTER the user has checked out Ethereum Studio. e.g. 
> Now that you've gotten a handle on the Ethereum stack with Ethereum Studio, continue your learning with these online resources from the community.

samajammin:
@carlfairclough I'm closing this out. Feel free to re-open when the time comes.

PR #892, opened 2020-03-26
Wallets page copy changes

jjmstark:
Various changes in response to individuals content audit:

- Remove ledger blog post, as hardware wallets are already covered by other links here
- Remove Opera, since it is not strictly a wallet but a browser with wallet functionality. May be re-added in future when page is expanded to include a larger suite of options.
- Added Argent, Coinbase wallet, and Gnosis safe

### DIFF
--- a/docs/wallets/index.md
+++ b/docs/wallets/index.md
@@ -18,13 +18,15 @@ lang: en-US
 
 Wallets are applications that make it easy to hold and send [ETH](/eth/), as well as interact with [applications built on Ethereum](/dapps/).
 
-Want to install a wallet?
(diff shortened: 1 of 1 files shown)


jjmstark:
A bit more context about this PR:

- Recently we conducted an "audit" of all the individuals content on the website, and are now in the process of implementing those recommendations
- Some of these recommendations involve changing the list of wallets on the site, which means we are now adding a few options to this list as a quick fix
- This is the first step towards expanding the ability of community members to suggest wallets that should be listed on this page - will be more shared about this soon on how we're going to further change and improve the /wallets page.

PR #906, opened 2020-04-01
Added twemoji [fixes #776]

### LINKED ISSUE #776: Update emojis to Twitter's Twemoji

**Additional context**

))

carlfairclough:
This adds twemoji

## Description

To make this work the following had to happen
1) Parse text
2) Get the file name from the twemoji `url` key
3) Get the emoji from the twemoji gh-pages branch

It is done this way as there is no locally-hosted twemoji package which works out of the box. The `twemoji-parser` always points to an external CDN, which in turn mirrors the `twemoji#gh-pages` branch.

For the default markdown (`config.js`), the only way to access the svg was by using `fs.readFileSync()` to get the contents, which can then be manipulated.

When inside a component where strings are being used, a new markdown parser is created which parses inline (no wrapping tags) and searches for the emoji. This doesn't inline the SVG and automatically places the twemojis used in the `assets` directory.

All emojis *must* be called like so: `:wave:`. Using. a native emoji e.g. `👋` will **not** be parsed,

## Related Issue

#776

## Screenshots

image


carlfairclough:
Emojis have been updated, including rolling out parsing in card components.

Issue #824, opened 2020-03-15
Site Translation: Add Romanian (v1.1)


samajammin:
### Prize Bounty
75 DAI

### Challenge Description
Romanian translations (#222) have completed review & are ready to be added to the website!

### Submission Requirements

1. Download the [files from CrowdIn](https://crowdin.com/project/ethereumfoundation/ro#) for this version of the content (`Contents v.1.1`). You may need to sign up for CrowdIn & join our project in order to download the files - here's the [invite link](https://crowdin.com/project/ethereumfoundation/invite).
2. Create new directory: /docs/translations/ro/
3. Copy in the CrowdIn files (the strings.json & markdown files) into the new /docs/translations/ro/ directory, keeping the same directory structure as in CrowdIn (i.e. the subdirectories & index.md filenames)
4. In each newly added markdown file...
- Within the [front matter](https://vuepress.vuejs.org/guide/frontmatter.html):
  - Confirm each file has `description`, `og:title` & `og:description` meta properties - check out other translation markdown files for reference
  - Confirm each file has the correct `lang` property (it should match the `locales` key in config.js)
- Update all relative links to include the language path. e.g. a link to the /learn page within the /docs/translations/sv/what-is-ethereum/index.md file should link to ethereum.org/sv/learn, not ethereum.org/learn
5. Add the site navigation for the translation in config.js by adding a new `locales` key. It should match the structure of other v1.1 translations (e.g. Swedish, `/se/`). For reference, you can see the site version of each language in translations.js `version` key.
6. Update the /utils/translations.js file
- Import the strings.json file for this language, e.g.
```const id = require('../../../translations/sv/strings.json')```
- Add the imported file variable to the `stringTranslations` object for this language (the key should match the `locales` key from config.js)
- Add an entry to the `languageMetaData` object for this language, e.g.
```sk: { language: 'Slovenský', 'language-english': 'Slovak', path: '/sk/' },```

**Additional context**

For further guidance, see https://github.com/ethereum/ethereum-org-website/pull/779 as a recent translation addition.

### Judging Criteria
The prize will be awarded to the first valid submission.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 75.0 DAI (75.0 USD @ $1.0/DAI)  attached to it as part of the ethereum fund.__

[...]

PR #934, opened 2020-04-08
Romanian (v1.1)

antodp:
# Description
Add Romanian translations (v1.1)

# Related issue
#824


docs/.vuepress/config.js
```diff
@@ -414,6 +414,54 @@ module.exports = {
         }
       ]
     },
+    '/ro/': {
+      lang: 'ro-RO',
+      label: translate('language'),
+      nav: [
```
samajammin, line 420: Looks like were missing the other nav items here. There's only 2, there should be 4.

antodp: Totally missed it! Fixed in the last commit.
Thank you.

Issue #817, opened 2020-03-14
Create a script to update internal links of translated files


samajammin:
### Prize Bounty
75 DAI

### Challenge Description
Currently when adding a new language translation or updating an existing translation (e.g. https://github.com/ethereum/ethereum-org-website/pull/816), all internal links within the translated markdown files must be manually updated to point to the appropriate page (i.e. the translated page vs. the corresponding English page). For instance, a link to the /learn page within the /docs/sv/what-is-ethereum/index.md file must be updated from `/learn` to `/sv/learn` (so it points to the translated page, ethereum.org/sv/learn, not the English version at ethereum.org/learn). See [this commit](https://github.com/ethereum/ethereum-org-website/pull/816/commits/ed70dce695ec8b1399c5fddde09c3efabb84f34b) as an example.

Create a node script to automate these updates.

### Submission Requirements
The node script should take a language directory as an argument, e.g.:
```
yarn update-internal-links docs/translations/de/
```
This script should traverse all markdown files within the directory for internal links, e.g.
```
[some internal link](/learn/)
```
It should IGNORE all external links, e.g.
```
[some external link](https://someexternalsite.com/)
```

For each internal link:
- If the link does not contain the language directory passed in, add it (/learn/ —> /de/learn/)
- If the link does not have a trailing slash, add it (/de/learn → /de/learn/)
- If the link already contains the language directory, do nothing (/de/learn/ --> /de/learn/)

The script should check that the passed argument (the language directory) is in fact a valid language directory (e.g. by checking the directory names within /docs/translations/, or keys of the `locales` object in config.js).

**Describe alternatives you've considered**

**Additional context**

It may be useful to check out the existing script at generateHeadingID.js as an example to work off.

### Judging Criteria
The prize will be awarded to the first valid submission.

samajammin:
@think-in-universe you might like this bounty 😄

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 75.0 DAI (75.0 USD @ $1.0/DAI)  attached to it as part of the ethereum fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/ethereum-org-website/817/4186).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/ethereum-org-website/817/4186).
[...]

PR #939, opened 2020-04-08
Script to update internal links of translated files

antodp:
## Description
The script will change all the internal links as follows:

- If the link does not contain the language directory passed in, add it (/learn/ —> /__lang__/learn/)
- If the link does not have a trailing slash, add it (/__lang__/learn → /__lang__/learn/)

Example usage (in project root directory):
```
yarn update-internal-links de
```

~~There is room for improvement (in terms of string manipulation), but for now I think it can do the work.~~
__EDIT__ fixed a bug in last commit and improved string manipulation

It works like this:

- retrieve all the Markdown files
- find all occourrences of links and apply rules to them
- replace the old links with the new correct links in the file body
- write the file

## Related Issue
#817


samajammin:
@all-contributors please add @antodp for code

Issue #985, opened 2020-04-23
Developer benefits should include no single point of failure and no devops


ethers:
Developer benefits should include no single point of failure and no devops

As an example:
https://news.ycombinator.com/item?id=22847595

> I run my own website out of a VPS and even that is nontrivial: occasional outages due to SSL certificates expiring, the VPS getting shut down because my credit card expires so the auto-payments stop working, troubles with the server software itself, etc etc. Meanwhile, if you just dump stuff onto one of these decentralized layers or even just a blockchain, you publish once, it's easy and it's there forever.
> 
> I truly believe that for "fully decentralized" to be viable, it must mean "serverless" (as in no servers, meaning no single computers whose IP addresses or DNS names or equivalents are explicitly referenced in a line of client-side code, not meaning that AWS lambda stuff). That does mean you're making significant efficiency tradeoffs for the foreseeable future; for some types of applications it's worth it and for others it's not.

I'll link to a PR to give an example of such phrasing and how the developer benefits should be more concrete.

ethers:
Other references https://twitter.com/VitalikButerin/status/1247995507756216320  and the previous tweet.

PR #986, opened 2020-04-23
A little more developer centric benefits

ethers:
A little more developer centric benefits and an example for #985


samajammin:
Thanks for this!

PR #990, opened 2020-04-29
Update meetups, update twemoji to latest, add parsing to lists

carlfairclough:
## Description

This updates the meetup list:
- now using the list component with flags
- in alphabetical order by country
- listed in an object rather than markdown

This also resolves two previously unknown issues:
- mixed elements (e.g. image, text) in list columns would not be displayed inline. They now are.
- twemoji previously was using a `v1.0` collection of SVG assets, which doesn't include flags. This has been updated to `latest`

image
image

### DIFF
--- a/docs/.vuepress/components/ListCard.vue
+++ b/docs/.vuepress/components/ListCard.vue
@@ -10,16 +10,16 @@
         <article :class="{ 'has-meta': item.meta }">
           <!-- h3 is overridden by the level prop. Default is 3 (h3) -->
           <h3 class="l7 mb-0 mt-0 tc-text500 art-title" :is="itemTitleTag">
-            {{ item.title }}
+            <span v-html="inlineMd(item.title)" />
           </h3>
           <p v-if="item.subtitle" class="l7 mb-0 mt-0 tc-text100 art-subtitle">
-            {{ item.subtitle }}
+            <span v-html="inlineMd(item.subtitle)" />
           </p>
           <p
             v-if="item.meta"
             class="l7 mb-0 mt-0 tc-text100 art-meta flex flex-center"
           >
-            {{ item.meta }}
+            <span v-html="inlineMd(item.meta)" />
           </p>
         </article>
       </router-link>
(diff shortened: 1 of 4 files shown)


docs/.vuepress/components/ListCard.vue
```diff
@@ -10,16 +10,16 @@
         <article :class="{ 'has-meta': item.meta }">
           <!-- h3 is overridden by the level prop. Default is 3 (h3) -->
           <h3 class="l7 mb-0 mt-0 tc-text500 art-title" :is="itemTitleTag">
-            {{ item.title }}
+            <span v-html="inlineMd(item.title)" />
```
samajammin, line 13: Remind me why this `inlineMd ` is needed? This is in order to render the twemojis?

docs/.vuepress/components/ListCard.vue
```diff
@@ -108,20 +110,16 @@ export default {
       align-items center
       grid-column 3
       content '↗️'
+      grid-row: 1/999
```
samajammin, line 113: No `:` needed. What's the `1/999` do in effect? I tried toggling this in the browser & didn't notice a difference.

docs/.vuepress/components/ListCard.vue
```diff
@@ -108,20 +110,16 @@ export default {
...
       &.has-meta
         grid-template-columns auto 3fr 1fr auto
-        .art-title
-          grid-row 1
-        .art-subtitle
-          grid-row 2
         .art-meta
           grid-column 3
-          grid-row 1/999
+          grid-row: 1/999
```
samajammin, line 120: Same here - No `:` needed.

docs/.vuepress/components/ListCard.vue
carlfairclough: Yeah, inlineMd would render the emojis *and* render any inline markdown styles (italics, bold, etc)

docs/.vuepress/components/ListCard.vue
carlfairclough: 1/999 ensures that the content spans all available rows. Really, we could do 1/3 here, as it has a max of 3 rows, but 1/999 keeps things simple and will never have to be updated. It's sort of like setting `z-index` to `9999`

samajammin, line 113:
```suggestion
      grid-row 1/999
```

docs/.vuepress/components/ListCard.vue
samajammin, line 120:
```suggestion
          grid-row 1/999
```

PR #1017, opened 2020-05-11
Update homepage

carlfairclough:
This updates the homepage
## Description

This leverages a new illustration across languages, and includes an updated homepage layout for **english** language only, right now. It would require translation integration.

### English
image
image

### Other languages:
image


samajammin:
On a large screen (1280px), I'm seeing the persona sections wrap. I don't think they should until a smaller screen - perhaps at `$breakL` (1024px)?

The amount of space looks strange:
Image 2020-05-11 at 11 03 25 AM

samajammin:
I think the homepage header & subheader should be larger, perhaps `l2` & `l4` respectively? Currently both `l5`. Here's what I'm thinking:
Image 2020-05-11 at 11 06 35 AM

docs/.vuepress/components/UpdatedHomePage.vue
```diff
@@ -0,0 +1,152 @@
+<template>
+  <div class="pa-1 l-up-pl-2 l-up-pr-2">
+    <div class="flex flex-column">
+      <div class="flex space-between align-center">
+        <h1 class="l5 max-w-55ch">
+          Welcome to Ethereum, a community-driven internet built for you
```
samajammin, line 6: @jjmstark should confirm but I think to begin with, we want to keep the copy the same (i.e. `translateString('page-home-title')`) or "Ethereum is a global, open-source platform for decentralized applications."

docs/.vuepress/components/UpdatedHomePage.vue
```diff
@@ -0,0 +1,152 @@
...
+        </h1>
+        <span class="l5 hidden m-up-inline-block " id="morph" ref="morph"
+          >Ξ</span
+        >
+      </div>
+      <div class="tc-text200 max-w-55ch">
+        <p class="l5">
+          Using Ethereum, you can write code that controls digital value, runs
```
samajammin, line 14: Same here, this should remain `{{ translateString('page-home-subtitle') }}`

samajammin:
@jjmstark do you think we should keep that rotating translation of Ethereum (right side text in [this video](https://share.getcloudapp.com/z8uxGBDL))? I think it's kind of cool to keep but wanted to call out as it's not prominent. It's also a bit less clear than before that it's translations of "Ethereum". I feel like the current homepage design makes that more obvious.

samajammin:
I like the current homepage display for translations - I think keeping the 3 persona sections as-is makes sense given that's the copy we have translations for.

docs/.vuepress/components/UpdatedHomePage.vue
jjmstark: Yes, headline copy for english should remain the same for now. As discussed in Figma thread, we *should* update this but I want to disentangle that update from shipping the new homepage design. 

(but as noted, we can make content adjustments as needed before shipping - these initial PRs from @carlfairclough are just about getting the design elements finalized)

jjmstark:
> @jjmstark do you think we should keep that rotating translation of Ethereum (right side text in [this video](https://share.getcloudapp.com/z8uxGBDL))? I think it's kind of cool to keep but wanted to call out as it's not prominent. It's also a bit less clear than before that it's translations of "Ethereum". I feel like the current homepage design makes that more obvious.

I really like this element and would love to preserve it on the site even once the new art is up, but I agree it might be a bit confusing the way it's used here.

Maybe instead it could be incorporated into an update to the Languages page?

jjmstark:
> I like the current homepage display for translations - I think keeping the 3 persona sections as-is makes sense given that's the copy we have translations for. 

For all translated versions of the site, the "below the hero image" section should be unchanged, as it is in the screenshot that @carlfairclough posted above. Is that what you are saying @samajammin?

> Perhaps it still makes sense to add the images for each persona? @jjmstark - what do you think?

I don't feel too strongly about this - if it's easy to do, let's do it. But our goal in any case should be to get the translations "caught up" fast after shipping the new content updates.

jjmstark:
Only remaining comment carried over from Figma: On the footer there is a purple-grey eth glyph here, besides "Website last updated...". I've always thought that looked kind of weird - style of glyph isnt used elsewhere, and it looks misaligned. 

Can we just remove the glyph from the footer? (this can probably be a separate PR, maybe @samajammin this is easy enough for you to do w/ out requiring Sam?)

samajammin:
> For all translated versions of the site, the "below the hero image" section should be unchanged, as it is in the screenshot that @carlfairclough posted above. Is that what you are saying @samajammin?

@jjmstark correct.

samajammin:
> Only remaining comment carried over from Figma: On the footer there is a purple-grey eth glyph here, besides "Website last updated...". I've always thought that looked kind of weird - style of glyph isnt used elsewhere, and it looks misaligned.
> 
> Can we just remove the glyph from the footer? (this can probably be a separate PR, maybe @samajammin this is easy enough for you to do w/ out requiring Sam?)

@jjmstark removed in #1023

samajammin:
Closing this in favor of #1115

PR #1018, opened 2020-05-11
Update what is ethereum page

carlfairclough:
This updates the CardList component to cater for "flat" cards with large non-hero images.
It also creates a TextHero component, allowing text with an image.

## Description

`/what-is-ethereum` has been updated to contain card lists and a clear hero image, to guide people through an onboarding process. It retains the original content block, which should ideally be updated before this can be rolled out.

`CardLists` now accept a `large` boolean and a `flat` boolean. `large will increase the padding and title size, while flat will remove shadows and apply a muted background colour.

A hero can accept a `title: required`, an `intro` (small text appearing above the title) and `image`.

image
image
image


samajammin:
I feel like the padding for this page should keep the same layout as other pages.

E.g. the "What is Ether" page keeps consistent alignment w/ nav & footer content:
Image 2020-05-11 at 11 23 49 AM

This new "What is Ethereum" page no longer does & the lack of consistency is a bit jarring when navigating to other pages:

Image 2020-05-11 at 11 24 07 AM

Image 2020-05-11 at 11 24 21 AM

docs/what-is-ethereum/index.md
```diff
@@ -8,9 +8,53 @@ meta:
   - property: og:description
     content: Guides, resources, and tools for users completely new to Ethereum.
 lang: en-US
----
+layout: landingpage
```
samajammin, line 11: In general I'd prefer to keep this type of content within an e.g. `<WhatIsEthereumPage/>` component with the strings.json translation pattern vs. having all the content in the markdown. The primary reason is that it will lead to less duplication, e.g. imagine this is rolled out on 30 languages. If we want to change the image for one of these cards, now we have to update the card's `header` in 30 places vs. in 1 component file. Or if we decide to remove the card `title` for whatever reason or make the card clickable, it'd be best to consolidate that in one place to edit vs. across a bunch of markdown files.

docs/.vuepress/theme/Layout.vue
```diff
@@ -76,7 +76,8 @@ export default {
   },
   computed: {
     isLandingPage() {
-      return this.$page.frontmatter && this.$page.frontmatter.layout === 'home'
+      const layouts = ['home', 'landingpage']
```
samajammin, line 79: Do we want this? Why not just keep as a standard `page` class? See my other comments w/ images on why I think we should keep the padding the same as other pages.

jjmstark:
Only comment here: in my [earlier feedback on this page](https://www.notion.so/efdn/Design-feedback-5-4-2020-2dcbdc7bda084aa4b7818b39f0d6bc26#5ff449734ff943b69182ddf0ae86edb5), I suggested that we just want **3** "key concept" blocks in the top section of this page, rather than 4.

I get that we can just delete one of them, though in that case I'm not sure the current arrangement is ideal. @carlfairclough would be interested in a take on this that's designed for 3 instead of 4 cards.

carlfairclough:
https://github.com/ethereum/ethereum-org-website/pull/1018#issuecomment-626876150
> I feel like the padding for this page should keep the same layout as other pages.
> ...
> E.g. the "What is Ether" page keeps consistent alignment w/ nav & footer content:
> ...
> This new "What is Ethereum" page no longer does & the lack of consistency is a bit jarring when navigating to other pages:

These pages are different from "long-form" pages and should be treated differently. Keeping the existing page width/padding is something I'd avoid because:

- content max-width is currently ~768px, and alignment caters to that (apart from on the homepage)
- long-form content has a sidebar, the landing pages do not

Footer alignment is an issue that I agree with, however that had been made to work with the prior pages, and the implementation here references the existing style guide. The footer would also need to be updated eventually.

This also explains why the `['home', 'landingpage`] changes are involved, as template information should be contained in the `.md` file and components pulled into that accordingly. IMO it's a step towards a stronger set of traffic-control pages, with the text-only pages being displayed in a consumable way. Left-alignment of text-only pages should eventually be brought to match the landing pages (if necessary!) and not vice-versa. The site should have the strongest initial signals in the traffic control pages (visual design, communication) to allow people to discover content that's relevant.

carlfairclough:
https://github.com/ethereum/ethereum-org-website/pull/1018#issuecomment-627519400

@jjmstark Switching to 3 would be a very quick change but we went with this approach as the content currently leads with the following 4 bullets:
> Ethereum is the foundation for a new era of the internet:
> • An internet where money and payments are built in.
> • An internet where users can own their data, and your apps don’t spy and steal from you.
> • An internet where everyone has access to an open financial system.
> • An internet built on neutral, open-access infrastructure, controlled by no company or person.

docs/.vuepress/theme/Layout.vue
carlfairclough: The `page` class is very limiting because it's tailored to long-form content rather than rich landing pages. For example, setting full-width sections (e.g. to set a background color) is a huge hassle. `landingpage` removes those limitations, while avoiding pulling in the hero (which should be called in the homepage markdown anyway).

samajammin: Got it. In that case, if we plan to use this across these new landing pages, I recommend we create a new component for this custom layout. For instance, we could create a new `<LandingPage />` component & in these markdown files specify `layout: LandingPage` in the frontmatter. Then we apply specific classes / styling within that LandingPage.vue component.

More on how to accomplish that here:
https://vuepress.vuejs.org/theme/default-theme-config.html#custom-page-class

samajammin:
@carlfairclough good point. I agree w/ you on the need for different layout vs. the standard `page` for long-form content but the current layout here is still too wide & looks awkward, particularly for the sections when there IS long-form content:
Image 2020-05-13 at 12 44 12 PM

How about we remove the `max-width: 768px;` for these landing page layouts (as you have done) but update the padding to stay in line with the footer?

samajammin:
Closing this out in favor of #1134

Issue #1062, opened 2020-05-28
Site Translation: Add Malayalam (v1.1)


samajammin:
### Prize Bounty
50 DAI

### Challenge Description
Malayalam translations have completed review & are ready to be added to the website!

### Submission Requirements
1. Download the [files from CrowdIn](https://crowdin.com/project/ethereumfoundation/ml-IN#) for this version of the content (`Contents v.1.1`) - here's a [screenshot showing how](https://share.getcloudapp.com/ApujZjyn). You may need to sign up for CrowdIn & join our project in order to download the files - here's the [invite link](https://crowdin.com/project/ethereumfoundation/invite).
2. Create new directory: /docs/translations/ml/
3. Copy in the v1.1 CrowdIn files (the strings.json & markdown files) into the new /docs/translations/ml/ directory, keeping the same directory structure as in CrowdIn (i.e. the subdirectories & index.md filenames)
4. In each newly added markdown file...
- Within the [front matter](https://vuepress.vuejs.org/guide/frontmatter.html):
  - Confirm each file has `description`, `og:title` & `og:description` meta properties - check out other translation markdown files for reference
  - Confirm each file has the correct `lang` property (it should match the `locales` key in config.js)
- Update all relative links to include the language path by running our [script](https://github.com/ethereum/ethereum-org-website/blob/dev/scripts/updateInternalLinks.js), i.e. run `yarn update-internal-links ml` from the root of the project 
5. Add the site navigation for the translation in config.js by adding a new `locales` key. It should match the structure of other v1.1 translations (e.g. Swedish, `/se/`). For reference, you can see the site version of each language in translations.js `version` key.
6. Update the /utils/translations.js file
- Import the strings.json file for this language, e.g.
```const sk = require('../../../translations/sk/strings.json')```
- Add the imported file variable to the `stringTranslations` object for this language (the key should match the `locales` key from config.js)
- Add an entry to the `languageMetaData` object for this language, e.g.
```sk: { language: 'Slovenský', 'language-english': 'Slovak', path: '/sk/' },```

**Additional context**

For further guidance, see https://github.com/ethereum/ethereum-org-website/pull/964 as a recent translation addition.

### Judging Criteria
The prize will be awarded to the first valid submission.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

[...]

PR #1067, opened 2020-05-28
Fix frontmatter values

samajammin:
## Description
Fix frontmatter values - they weren't intended to be translated.

## Related Issue
#1060 #1061 #1062


samajammin:
@sebastiantf FYI, these frontmatter values (`sidebar`, `layout` & `hideHero`) are not intended to be translated. Not your fault, just letting you know for next time 😄 Thanks again for the PRs!

I'll work with our Translation team (cc @sorumfactory) to see if we can prevent these phrases from being translated moving forward.

sebastiantf:
I found some of those in the Malayalam translations and changed them. But didn't think it was there in other translations since it was the first time I saw them.

Will surely look out for them next time 😃

PR #1154, opened 2020-07-01
Added images for homepage sections

ryancreatescopy:
## Description

– Added new illustrations to site to finish homepage image updates
– Added appropriate image alt descriptions


ryancreatescopy:
@samajammin reckon we should make these illustrations a little larger? I think we lose a bit of the detail right now.

Issue #1040, opened 2020-05-22
/dapps list shuffles text, but not hrefs


seichris:
### Prize Bounty
20 DAI

### Challenge Description
There's a bug in the RandomAppList.js component on the [Dapps page](https://ethereum.org/en/dapps/) - the list shuffles text, but not hrefs.

Fix the RandomAppList.js component to shuffle both text & href of items.

### Submission Requirements
The current behavior works as expected in local development but not once built & deployed. The solution must work on the preview deploy of the PR in order to be considered valid.

### Judging Criteria
The prize will be awarded to the first valid submission.

## Original Bug Report:

**Describe the bug**
Reload https://ethereum.org/en/dapps/ to see that the lists of DApps is shuffling.
However, it only shuffles the text. But it keeps the href in place.

**Expected behavior**
It should also shuffle the href.

_**Edited on 7/6/2020 to reflect new bounty**_

seichris:
**Fix**
Problem maybe in here
https://github.com/ethereum/ethereum-org-website/blob/dev/docs/.vuepress/components/RandomAppList.vue or
https://github.com/ethereum/ethereum-org-website/blob/dev/docs/.vuepress/components/RandomList.vue

samajammin:
Yikes! Thanks for pointing this out @seichris!

samajammin:
@all-contributors please add @seichris for bug reports

samajammin:
Updated description for Gitcoin bounty.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 20.0 DAI (20.0 USD @ $1.0/DAI)  attached to it as part of the ethereum fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ethereum/ethereum-org-website/1040/4387).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ethereum/ethereum-org-website/1040/4387).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $113,198.42 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

[...]

PR #1163, opened 2020-07-06
chore: fixed list shuffles on RandomAppList 

developerfred:
## Description
Changing the logic of random numbers for correction or random reproduction
#1040

https://share.getcloudapp.com/2Nu5prrO

### DIFF
--- a/src/components/RandomAppList.js
+++ b/src/components/RandomAppList.js
@@ -1,4 +1,4 @@
-import React from "react"
+import React, { useEffect, useState } from "react"
 import Link from "./Link"
 import Translation from "./Translation"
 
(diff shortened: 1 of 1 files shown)


developerfred:
@samajammin I believe I discovered the problem, we need to use hooks. Every time the component is shuffled the Intl  loads. Putting the component to be the last one to load using userEffects. makes the component work normally.

src/components/RandomAppList.js
```diff
@@ -40,15 +40,22 @@ const appList = [
   },
 ]
 
-const RandomAppList = () => {
-  const list = appList.map((item) => {
-    item.randomNumber = Math.random()
-    return item
-  })
-  list.sort((a, b) => a.randomNumber - b.randomNumber)
+const RandomAppList = () =>
```
samajammin, line 43: Missing brackets here which is breaking the build:
```
const RandomAppList = () => {
```

src/components/RandomAppList.js
```diff
@@ -40,15 +40,22 @@ const appList = [
...
+const [setAppRandomList, appRandomList] = useState({ dapps: []});
```
samajammin, line 44: The set state function should be the second item in the array.

samajammin: I also don't think we need an object here (`useState({ dapps: []})`), I think we can just use an array `useState([])`

developerfred:
Nice! Thanks @samajammin

samajammin:
@all-contributors please add @developerfred for code.

PR #1204, opened 2020-07-14
Add color variables from design system

samajammin:
## Description
1st step towards cleaning up our color variables & applying uniform colors across the app.

## Related Issue
None


samajammin:
@ryancreatescopy FYI - here's a step towards rolling out the design system colors. This creates all the variables, now we just need to roll out the colors (mainly by replacing the one-off variables we've used).

PR #1237, opened 2020-07-21
change: moment.js to luxon

developerfred:
## Description

Proposal

as i added another lib of momentum intl i was wondering how long it takes to load and how bad it is, i saw a new lib that moment's own support team.

is Luxon has a [total of 20.4kb](https://bundlephobia.com/result?p=luxon@1.24.1)

and it's great for those who will enter the site by cell phone, and how it flows much better with gatsby


developerfred:
I just updated @samajammin

developerfred:
9aafd963...5f33b004 its working now, but I need to pass the timestamp.

PR #1250, opened 2020-07-22
Update CryptoZombies links

samajammin:
## Description
- Update links to more relevant resource (Solidity course)
- Update links to translated content if it exists

## Related Issue
None

### DIFF
--- a/src/content/developers/index.md
+++ b/src/content/developers/index.md
@@ -526,7 +526,7 @@ The Ethereum network is made up of many nodes who run compatible client software
 
 **Cryptozombies -** **_Learn to code games on ethereum._**
 
-- [Cryptozombies.io](https://cryptozombies.io/)
+- [Cryptozombies.io](https://cryptozombies.io/en/solidity)
 
 **Ethernaut -** **_Solidity based wargame where each level is a contract to be hacked._**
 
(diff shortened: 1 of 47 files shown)


samajammin:
@ryancreatescopy @ryancreatescopy @sorumfactory we should think through how we might be able to do this at scale for external resources we link to.

ryancreatescopy:
> @ryancreatescopy @ryancreatescopy @sorumfactory we should think through how we might be able to do this at scale for external resources we link to.

Do what at scale exactly? Let's chat this over...

sorumfactory:
Sam, do you mean linking the third party "translated" contents to the "translated Ethereum.org" pages? I'm not sure If I understand the context correctly. It would be great if you could elaborate on it more.

samajammin:
Yes, sorry for the lack of clarification.

CryptoZombies is a top resource that we link to. It also has translated version of its content in ~10 languages. I'm wondering what the best way is for us to recognize these type of instances & to ensure that our translated content links out to the corresponding translated content.

I created a card where we can brainstorm ideas if you have any: https://www.notion.so/efdn/Think-through-how-best-to-link-to-3rd-party-content-translations-4eaf6bef40f146cb9955618f1158acaf

Issue #1363, opened 2020-08-21
Adding MEW's absolute beginner's guide comic


gamalielhere:
We would like to add MEW's "What Is Ethereum?" animated comic to the "How Ethereum Works" category on https://ethereum.org/en/learn/

samajammin:
Hey @gamalielhere - thanks for the suggestion. Could you please share a link to the animated comic you're referring to?

PR #1364, opened 2020-08-21
Adding MEW's absolute beginner's guide comic

gamalielhere:
## Description

Adds MEW's animated comic "What is Ethereum?" as a resource for learning about Ethereum.

## Related Issue

 #1363


ryancreatescopy (commented):
Hi @gamalielhere thanks so much for the PR.

This comic is really cool but I don't think it covers anything new that's not already covered by existing items. This is something we consider when deciding what to put on the site. You can see the whole policy here: https://ethereum.org/en/contributing/adding-articles

It's also a little too promotional for Ethereum.org — we want to remain as unbiased as possible. 

Hope this makes sense and doesn't put you off making PRs in the future. 
Thanks!

Issue #1205, opened 2020-07-15
New Wallets page


ryancreatescopy:
**Is your feature request related to a problem? Please describe.**

A new page that provides a more detailed overview of Ethereum wallets.

Design goals:
1. Explain wallet clearly with minimal jargon and introduce the required concepts with jumping off points for more information (when available)
2. Help a user choose a wallet suited to what they want to do with it
3. Provide a clear next step for when a user has a wallet: dapps/eth
4. Allow users to suggest new wallets based on criteria (_will be considered in the find a wallet page_)
5. Help a user set up and interact with a wallet safely
    - Explain why this is necessary
6. Make the page more visually exciting than a "readme"

**Additional context**

Current page: 
https://ethereum.org/en/dapps/

Design:
https://www.figma.com/file/VYUlXIF1DSORx9xpwIx3eB/Wallets?node-id=81%3A2

samajammin:
Solved in #1206

PR #1386, opened 2020-08-31
Always show wallet filter tags

samajammin:
## Description
Always show wallet filter tags

## Related Issue
#1205

## Screenshots

Image 2020-08-30 at 8 10 03 PM


ryancreatescopy (changes requested):
This looks great! However I double checked the inactive tags and they don't pass AA accessibility. Have committed a fix which solves it. Had to up the opacity to `0.7` which in turn made it hard to distinguish between active and inactive tags so I also added a border and a shadow to better differentiate the two. Let me know what you think.

Issue #1394, opened 2020-09-02
Set up "What is Ethereum?" page for translations


samajammin:
**Is your feature request related to a problem? Please describe.**
We recently updated our ["What is Ethereum?" page](https://github.com/ethereum/ethereum-org-website/blob/dev/src/pages-conditional/what-is-ethereum.js) from a markdown file to a fully featured React component.

We need to set this content up for translation. 

**Describe the solution you'd like**
1.  Move the English copy into [/intl/en.json](https://github.com/ethereum/ethereum-org-website/blob/dev/src/intl/en.json) - see [our homepage](https://github.com/ethereum/ethereum-org-website/blob/dev/src/pages/index.js) for an example of how this is done.
2. Replace the English copy with `<Translation/>` and/or the `intl.formatMessage()` function (again, see the homepage for an example of how this is done)

**Describe alternatives you've considered**
None

**Additional context**
We the use `[gastby-plugin-intl](https://github.com/wiziple/gatsby-plugin-intl)` for our internationalization. That plugin's documentation may provide more context for how it works.

samajammin:
Solved in #1406

PR #1406, opened 2020-09-03
set up what is Ethereum? page for translations

ph5500:
fixes [#1394]

## Description

Moved English copy of What-is-Ethereum page to /intl/en.json file using gatsby-plugin-intl for internationalization


samajammin:
Thanks for the PR @ph5500! Couple things
1. Looks like the build is breaking - did running this locally work? Try running `yarn start:static` & ensure you're able to build locally without any errors.
2. In the future, please include the issue # directly in the commit, i.e. your commit message should be `set up what is Ethereum? page for translations [Fixes #1394]`. That helps git link the commit directly to the issue. More on that [here](https://github.com/ethereum/ethereum-org-website#submit-your-pr).

src/intl/en.json
```diff
@@ -155,5 +155,60 @@
   "language-swedish": "Swedish",
   "language-turkish": "Turkish",
   "language-chinese": "Chinese Simplified",
-  "website-last-updated": "Website last updated"
+  "website-last-updated": "Website last updated",
+  "whats-ethereum-meta-title": "What is Ethereum?",
```
samajammin, line 159: Can we continue with the `page-` pattern? e.g. so start all these with `page-what-is-ethereum-...`

src/pages-conditional/what-is-ethereum.js
```diff
@@ -216,101 +219,105 @@ const StyledCallout = styled(Callout)`
...
-    description:
-      "Better products and experiences are being built all the time because Ethereum products are compatible by default. Companies can build on each other's success.",
+    title: <Translation id="whats-ethereum-compatibility-card" />,
+    description: <Translation id="whats-ethereum-compatibility-card-desc" />,
   },
 ]
 
 const WhatIsEthereumPage = ({ data }) => {
+  const intl = useIntl()
+  const contentVersion = getLangContentVersion(intl.locale)
```
samajammin, line 255: I don't think you need to check content version at all on this page.

src/pages-conditional/what-is-ethereum.js
```diff
@@ -216,101 +219,105 @@ const StyledCallout = styled(Callout)`
...
   const actions = [
     {
-      title: "ETH",
+      title: "whats-ethereum-native-title",
+      shouldDisplay: contentVersion > 1.0,
```
samajammin, line 259: Like comment above, this shouldn't be needed.

src/pages-conditional/what-is-ethereum.js
```diff
@@ -216,101 +219,105 @@ const StyledCallout = styled(Callout)`
...
+              <Translation id="whats-ethereum-accessibility" />
+            </Subtitle>
+            <SubtitleTwo>
+              <Translation id="whats-ethereum-tools-needed" />
+            </SubtitleTwo>
           </Header>
           <Hero
             fluid={data.hero.childImageSharp.fluid}
-            alt="Illustration of a person peering into a bazaar, meant to represent Ethereum"
+            alt="whats-ethereum-alt-img-bazaar"
```
samajammin, line 312: Since, you're not using the `<Translation/>` component here, you'll need to use the `formatMessage` function - see here for an example: https://github.com/ethereum/ethereum-org-website/blob/dev/src/pages/index.js#L373

src/pages-conditional/what-is-ethereum.js
```diff
@@ -329,60 +336,76 @@ const WhatIsEthereumPage = ({ data }) => {
       <BannerContainer>
         <Banner
           fluid={data.banner.childImageSharp.fluid}
-          alt="An illustration of characters in a social space dedicated to Ethereum with a large ETH logo"
+          alt="whats-ethereum-alt-img-social"
```
samajammin, line 339: Same deal here (& all your other usages of `alt`).

src/pages-conditional/what-is-ethereum.js
```diff
@@ -216,101 +219,105 @@ const StyledCallout = styled(Callout)`
 const cards = [
   {
     emoji: ":bank:",
-    title: "Banking for everyone",
-    description:
-      "Not everyone has access to financial services. But all you need to access Ethereum and its lending, borrowing and savings products is an internet connection.",
+    title: <Translation id="whats-ethereum-banking-card" />,
```
samajammin, line 222: Did you try running the app? I don't think you can put a component in this field... I suspect that's why the build is broken. You could just put the id here (i.e. `"whats-ethereum-banking-card"`) & then translate using the `formatMessage` function when you pass it into the component.

samajammin (changes requested):
Good start but a few issues... I left some comments. LMK if you have any questions!

Thanks

src/components/ActionCard.js
```diff
@@ -70,8 +71,12 @@ const ActionCard = ({
         )}
       </ImageWrapper>
       <Content className="action-card-content">
-        <h3>{title}</h3>
-        <Description>{description}</Description>
+        <h3>
```
nicc938, line 74:
```suggestion
        <h3>
```

src/pages-conditional/what-is-ethereum.js
ph5500: The translation component seems to be working properly when ran locally. Please let me know if there are any issues.

nicc938 (changes requested):
😃

samajammin (approved):
Thanks for this!

PR #1407, opened 2020-09-04
Converted How to submit section to h2 tag so it appears in toc

gilbertginsberg:


samajammin (approved):
Thanks for the PR @gilbertginsberg!

FYI, displaying h3s in the TOC is doable as well, you'd just want to set `sidebarDepth: 2` in the frontmatter.

Display as h2:
Image 2020-09-03 at 5 44 53 PM


Display as h3:
Image 2020-09-03 at 5 44 36 PM

LMK which you'd prefer.

gilbertginsberg:
Ahh, okay okay! Very good. 
Then in that case, I actually prefer the latter (as `h3`) :P because _How to submit_ stands out more.

PR #1436, opened 2020-09-14
edn-mvp EVM/gas article

gichiba:
## Description
This is a working draft of the EVM and gas articles for the edn. 

I believe that although the current structure specifies that gas should be a second article, I'd like to merge the two into one page, since it's really difficult to discuss the concept of gas separate from the concept of EVM execution.

### DIFF
--- a/src/content/developers/docs/blockchain-basics/index.md
+++ b/src/content/developers/docs/blockchain-basics/index.md
@@ -61,7 +61,7 @@ The native cryptocurrency of Ethereum. Users pay ether to other users to have th
 
 ### EVM
 
-The Ethereum Virtual Computer is the global virtual computer who’s state every participant on the Ethereum network stores and agrees on. Any participant can request the execution of arbitrary code on the EVM; code execution changes the state of the EVM.
+The Ethereum Virtual Machine is the global virtual computer who’s state every participant on the Ethereum network stores and agrees on. Any participant can request the execution of arbitrary code on the EVM; code execution changes the state of the EVM.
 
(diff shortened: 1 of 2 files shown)


ryancreatescopy (commented):
This is great @gichiba! I only have one piece of feedback. The paragraph "EVM instructions". could do with a few explainer sentences. Things like "SHA-3-256" and even the blockchain-specific stack operations could do with a bit of elaboration in my opinion. Even if it is just a link where a user can find out more.

samajammin (approved):
This is excellent @gichiba - thanks for your contribution!

I agree with @ryancreatescopy's point re: explaining a few references but overall I think this is good to merge as-is. If you'd like to make any further edits, feel free to open a new PR.

I definitely hear you re: merging Gas & EVM into one page. I'm open to that but also think we could accomplish our goals with 2 distinct pages (w/ plenty of interlinks). There's certainly a UX tradeoff but part of the reasoning here is there's a ton of search engine volume for "eth gas" related terms & having a standalone page to explain what gas is & why it exists would us rank for these terms & address these specific user questions. I think it's possible to explain gas at a high level without necessarily explaining the EVM in depth - gas is used to pay for computation on the network & as a result, helps prevent spam. We can include links to this EVM page for people who want to learn more & understand the relationship between them.

What do you think about that? Again, I'm open to potentially going either way.

samajammin:
FYI @gichiba I'm closing this out in favor of #1451 (same commits from you & I fixed the merge conflict).

PR #1464, opened 2020-09-21
Update solidity example

hrkrshnn:
## Description

Updated the example since the older one uses syntax that is deprecated. The example is the same as the one here https://solidity.readthedocs.io/en/v0.7.1/introduction-to-smart-contracts.html#subcurrency-example


samajammin (approved):
Awesome! Thanks for this 😄

samajammin:
@all-contributors please add @hrkrshnn for content.

Issue #1263, opened 2020-07-24
Nav accessibility issues


samajammin:
**Describe the bug**

```
warning ESLintError:
/Users/samrichards/code/ef/ethereum-org-website/src/components/Nav/index.js
  285:9  warning  Visible, non-interactive elements with click handlers must have at least one keyboard listener  jsx-a11y/click-events-have-key-events
  285:9  warning  Static HTML elements with event handlers require a role                                         jsx-a11y/no-static-element-interactions
```

**To Reproduce**

1. Run app (`yarn start`)
2. View console warnings

**Expected behavior**

We should add keyboard listeners to the site navigation for accessibility.

notrixbe:
i can do a fix for this issue

notrixbe:
A possible patch.  

`From c6293d72e4503b348346512405a984e926589063 Mon Sep 17 00:00:00 2001
From: wim notredame <wim.notredame@healthconnect.be>
Date: Sat, 26 Sep 2020 01:29:55 +0200
Subject: [PATCH] fix: 1263 - warning navigation + remove outline when click

---
 src/components/Nav/index.js | 12 ++++++++++--
 1 file changed, 10 insertions(+), 2 deletions(-)

diff --git src/components/Nav/index.js src/components/Nav/index.js
index efa64060..c489320b 100644
--- src/components/Nav/index.js
+++ src/components/Nav/index.js
@@ -50,6 +50,9 @@ const NavContent = styled.div`
     justify-content: space-between;
   }
 `
+const NavMobileButton = styled.span`
+  outline:none;
+`
 
 const InnerContent = styled.div`
   display: flex;
@@ -364,9 +367,14 @@ const Nav = ({ handleThemeChange, isDarkTheme, path }) => {
             toggleTheme={handleThemeChange}
             linkSections={mobileLinkSections}
           />
-          <span onClick={handleMenuToggle}>
+          <NavMobileButton
+            onClick={handleMenuToggle}
+            onKeyDown={handleMenuToggle}
+            role="button"
+            tabIndex="0"
+          >
             <MenuIcon name="menu" />
-          </span>
+          </NavMobileButton>
         </NavContent>
       </StyledNav>
       {shouldShowSubNav && (
-- 
2.24.3 (Apple Git-128)

`

Luckylukee:
#1263

PR #1519, opened 2020-09-26
#1263 bugfix-navigation eslint error

notrixbe:
#1263 bugfix navigation.

## Description
remove outline because otherwise user see small border when click it.
also needed to be added on dropdown.

## Related Issue
none


notrixbe:
wrong git account . update soon

PR #1521, opened 2020-09-26
Added tooling for building public dataset addendum and clarified award amounts

gilbertginsberg:
## Description

Edited medalla data challenge flyer text

* Added tooling for building public dataset addendum 
* and clarified award amounts


samajammin (changes requested):
Hey @gilbertginsberg this looks good to me except for the changes within /assets/ - can you please remove those?

PR #1525, opened 2020-09-26
[ERC-721] Non-Fungible Token Standard Documentation

arturgontijo:
## Description

Adding the documentation for `ERC-721`.
@samajammin let me know if the section title `Popular NFTs` is ok for your suggestion.


samajammin (approved):
Thanks for this!

PR #1530, opened 2020-09-28
Update smart contract languages page

franzihei:
## Description

Updates the smart contract languages page in the docs section. 
Overview of changes and rationale: 
-  Rewording the intro from "huge advantage" to "great aspect" which I personally find less opinionated (but feel free to change).
- Fixing some capitalisation and punctuation issues for consistency.
- Updating Solidity link for "Solidity by Example" (the chosen link did not support Solidity v.0.7.).
- Biggest changes in the "How to choose" section: I don't think the prior chosen table view makes sense since the elements of the table did not really compare to each other. Hence, I rewrote it to consist of two lists. We can further work on these two lists; happy to adjust the content further there. Btw, Solidity is not a "JavaScript-like language" :) 

And as I outlined before: I don't think this "comparison" page makes much sense but I tried to still improve it. :)

### DIFF
--- a/src/content/developers/docs/smart-contracts/languages/index.md
+++ b/src/content/developers/docs/smart-contracts/languages/index.md
@@ -200,19 +200,19 @@ This example should give you a sense of what Vyper contract syntax is like. For 
 
 ## How to choose {#how-to-choose}
 
-So, how to pick the fitting smart contract language for your purpose? As with any other programming language, it's mostly about choosing the right tool for the right job as well as personal preferences.
+As with any other programming language, it's mostly about choosing the right tool for the right job as well as personal preferences.
 
 Here are a few things to consider if you haven't tried any of the languages yet: 
 
-What is great about Solidity? 
+### What is great about Solidity? 
 - If you are a beginner, there are many tutorials and learning tools out there. See more about that in the [Learn by Coding](https://ethereum.org/en/developers/learning-tools/) section.
 - Good developer tooling available.
-- Solidity is good for gas golfing (optimizing gas usage of your contracts).
-- Solidity has a quite big developer community, which means you'll most likely find answers to your questions quite quickly.
+- Solidity is good for gas golfing (optimizing the gas usage of your contracts).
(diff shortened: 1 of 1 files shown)


franzihei:
Two things that just came into my mind: 
- We could also add that there are several compiler maintained by different teams (for Solidity that's the solc compiler from the EF, SOLL from SecondState and Solang).
- We should also mention Yul somewhere.

ryancreatescopy (approved):

samajammin (approved):
Thanks for these updates @franzihei!

Approving & will merge. A few thoughts on your comments/changes.

> Updating Solidity link for "Solidity by Example" (the chosen link did not support Solidity v.0.7.).

That's true but should we remove that link? I think it's still a great resource & we already link to the solidity docs in the link above.

> We could also add that there are several compiler maintained by different teams (for Solidity that's the solc compiler from the EF, SOLL from SecondState and Solang).

I think that's a great idea! It'd. be most relevant to add that information on [our page about compiling](https://ethereum.org/en/developers/docs/smart-contracts/compiling/) - we'd welcome any PRs around that! See #1550 for more info.

> We should also mention Yul somewhere.

I don't know much about Yul. Seems like a pretty advanced topic but I'm open to additions where you think it makes sense. Perhaps on the compiling page also? I'd defer to you & your team on this.

Thanks again!

samajammin:
@all-contributors please add @franzihei for content.

franzihei:
I'll work on adding a Yul section because I think it is relevant in here. Yul is an intermediate language and Yul+ aims to be a low-level, highly efficient extension to it.
I'll open a new PR with all the resources and info linked to it. :) 

> That's true but should we remove that link? I think it's still a great resource & we already link to the solidity docs in the link above.

I kept the "Solidity by example" link in the further reading section. Since Solidity is still a language that is evolving at high speed it is important to encourage use of the latest version of Solidity so I don't want to encourage readings / tools that are not on the latest version yet. Once it has been updated to 0.7 I'm happy to switch back the links! :)

PR #1536, opened 2020-09-28
Add fallback redirect to /en/ [Fixes #1374]

### LINKED ISSUE #1374: Add fallback redirects for requests that don't include a language path

Requests to pages without a language path (e.g. https://ethereum.org/eth2) currently 404.

I imagine it's easy to forget the language path when linking to pages. I myself just did this (& I build the darn site):

- Come up with a way to automatically redirect pages to English equivalent if the requested page doesn't exist
- Ideally this accomplished using a redirect rule in [our netlify.toml](https://github.com/ethereum/ethereum-org-website/blob/dev/netlify.toml) ([docs here](https://docs.netlify.com/configure-builds/file-based-configuration/#redirects))

- Manually add a redirect from every base page (e.g. https://ethereum.org/eth2 --> https://ethereum.org/en/eth2) 👎 
- Add additional info on [our 404 page](https://github.com/ethereum/ethereum-org-website/blob/dev/src/pages/404.js) (e.g. mention something like "is your URL missing a language path?")

**Additional context**
I tried the automatic route using our intl plugin but it lead to this issue: https://github.com/ethereum/ethereum-org-website/issues/1347
**Note: this approach may become a valid solution again once we roll out every page for every language!**

Docs on redirects using netlify.toml: https://docs.netlify.com/configure-builds/file-based-configuration/#redirects

samajammin:
## Description
Redirect request URLs to add the English directory (`/en/`) if the requested URL doesn't exist.

## Related Issue
#1374

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -256,6 +256,14 @@
   from = "/en/delphi/"
   to = "/en/developers/docs/programming-languages/delphi/"
 
+
+# English fallback redirect
+# If no page exists at /request-url, redirect to /en/request-url
+[[redirects]]
+  from = "/*"
+  to = "/en/:splat"
+  force = false
+
 # A redirect rule with all the supported properties
 # [[redirects]]
 #   from = "/old-path"


samajammin:
@ryancreatescopy @jjmstark I'm wondering what you think of this potential solution to #1374. Things to consider:

It solves the core pain point we identified - loading the URL https://deploy-preview-1536--ethereumorg.netlify.app/eth2/ will correctly redirect a user to https://deploy-preview-1536--ethereumorg.netlify.app/en/eth2/

There may be drawbacks, however. It potentially creates confusion when requesting URLs that don't exist & the issue wasn't that a language path was missing.
- e.g. a request to https://deploy-preview-1536--ethereumorg.netlify.app/nothing-here/ redirects to another page that doesn't exist: https://deploy-preview-1536--ethereumorg.netlify.app/en/nothing-here/
- This might be particularly confusing for translations, e.g. a request to https://deploy-preview-1536--ethereumorg.netlify.app/de/eth2/ redirects to https://deploy-preview-1536--ethereumorg.netlify.app/en/de/eth2/
- Perhaps other negative side effects that I'm not thinking of?

I'm hoping I can find a way to make this logic smarter, e.g. by only redirecting if the request URL contains a language path but I'm not yet sure how. My question to you: **if this is the best we can do, is this worth doing?**

Thanks!

jjmstark:
> e.g. a request to https://deploy-preview-1536--ethereumorg.netlify.app/nothing-here/ redirects to another page that doesn't exist: https://deploy-preview-1536--ethereumorg.netlify.app/en/nothing-here/

I think this is fine - the end result is the same, so it's not like a user is getting less use from this.


> This might be particularly confusing for translations, e.g. a request to https://deploy-preview-1536--ethereumorg.netlify.app/de/eth2/ redirects to https://deploy-preview-1536--ethereumorg.netlify.app/en/de/eth2/

Good point - ideally what we want is that if a user requests a page that exists in english, but not in the language directory they try and use, it reverts back to the english page (e.g. user requests de/eth2, they get en/eth2). Is it easy to implement that kind of logic?

samajammin:
Thanks for the input @jjmstark - I can look into that.

In the meantime, we're blocked on merging this. While it works in the preview deploy environment, this change unfortunately breaks local development. I've contacted the Netlify support team & @overlordofmu created this bug report (thanks @overlordofmu!): https://github.com/netlify/cli/issues/1299

I'll keep an eye on it.

samajammin:
I'm going to merge this now & just comment this section out when developing with `netlify dev` locally.

Issue #969, opened 2020-04-17
Create an "Ethereum nodes/clients" page


samajammin:
**Is your feature request related to a problem? Please describe.**

Current information on the site is pretty limited:
- https://ethereum.org/learn/#clients-and-nodes
- https://ethereum.org/developers/#clients--running-your-own-node

Why improve it?
- There's significant search volume for "ethereum node" related terms (~15K queries/month)
- It should be easy for visitors to find info on
    - What is an Ethereum node? High level overview of how one fits into the big picture
    - The different ETH clients available & how to follow progress/contribute to each project
    - How to run a node
    - The state of the network (# of nodes online, client diversity)
    - Different kinds of nodes (e.g. light client vs. full node)?

**Describe the solution you'd like**

A dedicated page to Ethereum nodes

**Describe alternatives you've considered**

Build off existing sections

**Additional context**

Some resources to consider
- https://kauri.io/ethereum-101-part-2-understanding-nodes/48d5098292fd4f11b251d1b1814f0bba/a
- [https://ethernodes.org/](https://ethernodes.org/)
- [https://ethstats.net/](https://ethstats.net/)
- [https://medium.com/@JustinMLeroux/running-ethereum-full-nodes-a-guide-for-the-barely-motivated-a8a13e7a0d31](https://medium.com/@JustinMLeroux/running-ethereum-full-nodes-a-guide-for-the-barely-motivated-a8a13e7a0d31)
- [https://docs.ethhub.io/using-ethereum/running-an-ethereum-node/](https://docs.ethhub.io/using-ethereum/running-an-ethereum-node/)

MariusVanDerWijden:
I also see a lot of questions regarding state size, dag size, time to sync a full node and most commonly used parameters for a node. Maybe this could also be part of this site.

taxmeifyoucan:
I am thrilled to see this PR! I see this as a crucial in questions of client diversity and network grow. 

So I was thinking how should high level content of the page look like.

1. What's a node? 
    1. Intro description
    2. Visualization [Tomo could make this]
2. Why should I run a node?
    1. Benefits for an individual, privacy, trustless, self sufficiency / web3
    2. Run a service (examples / infura / ...) 
    3. Provide infrastructure (VIP node / community space (trusted info)) 
3. Which client to pick? 
    1. Geth 
    2. OpenEthereum
    3. Turbo-Geth
    4. Besu
    5. Nethermind
[...]

PR #1541, opened 2020-09-29
Added info about each client

taxmeifyoucan:
Added paragraphs about each implementation after talking to client implementer teams. I think this should be part of the page, feel free to edit.

## Related Issue
#969


samajammin (approved):
Great - thanks for this addition!

PR #1584, opened 2020-10-06
Added Sydney's most active Ethereum dev community

jzaki:
The previous non-Ethereum link (replaced) was the product of a distinct shift away from the local Ethereum community.

### DIFF
--- a/src/components/MeetupList.js
+++ b/src/components/MeetupList.js
@@ -6,6 +6,18 @@ import Link from "./Link"
 
 const meetups = [
   {
+    title: "Bokky's Ethereum Workshop",
+    emoji: ":australia:",
+    location: "Sydney",
+    link: "https://www.meetup.com/BokkyPooBahs-Ethereum-Workshop/",
+  },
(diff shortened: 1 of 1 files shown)


samajammin (changes requested):
Thanks for the PR @jzaki! BokkyPooBah's Ethereum Workshop looks like a great group 👍 

My concern here is around replacing Web3 Sydney vs. just adding the new group. While I agree Web3 Sydney does not appear to be _specific_ to Ethereum, it does mention Ethereum (as the first protocol) & seems like a mission-aligned community. Do you have evidence that it's not an Ethereum community?

jzaki:
The incidents around the replaced group being spun out were, in a nutshell, controversial. But to avoid digging up past dirt, I will amend my pr to leave it there, and add the other meetup too. People will figure out where the conversation is happening 👍

samajammin (approved):

samajammin:
@all-contributors please add @jzaki for content.

PR #1607, opened 2020-10-09
Removed "no transaction fees" from wallet features [Fixes #1412]

### LINKED ISSUE #1412: Issue with "Find a wallet" on ethereum.org
**Describe the bug**

Argent no longer subsidizes transfer fees, but the ethereum.org website still shows Argent as having gas subsidization.

**To Reproduce**

2. Click on "No transaction fees"
3. Click on "Search chosen features"

**Expected behavior**

Argent should not show up when "No transaction fees" is selected.

ryancreatescopy:
## Description

Due to high gas prices and the decision for the majority of wallet teams to reinstate transaction fees, we've removed this as a filterable feature for finding a wallet. 

We will aim to replace this with another feature that would be useful to users – if you're reading this and have any ideas about what feature we should list, raise an issue!

## Related Issue

#1412


samajammin (approved):
👏

PR #1608, opened 2020-10-09
fixed typo?

mustafawm:
Hello,

I think there is a typo under Further Applications section of the white paper:

_2. Crop insurance. One can easily make a financial derivatives
contract **but**  using a data feed of the weather instead of any price
index. If a farmer in Iowa purchases a derivative that pays out_

should be **by**


samajammin (approved):
Nice catch! Thanks 😄

samajammin:
@all-contributors please add @mustafawm for content.

PR #1639, opened 2020-10-15
EthOnline hack

samajammin:
## Description
WIP "wallet wizard" to onboard users by creating a wallet for them & sending them testnet ETH to play with.

## Related Issue
None, EthOnline hackthon submission: https://hack.ethglobal.co/showcase/wallet-wizard-recJTSmudhaUcShjw


src/components/Button.js
```diff
@@ -41,16 +49,41 @@ const Secondary = styled(StyledButton)`
   }
 `
 
-const Button = ({ to, isSecondary, children, className }) => {
+const Button = ({
+  to,
+  isSecondary,
+  children,
+  className,
+  marginLeft = 0,
```
samajammin, line 57: Not sure this is a pattern we want to start rolling out on components 🤔 

Note to self to find a better approach.

samajammin:
FYI @ryancreatescopy we have the happy path working for the ETH faucet (thanks to the benevolent @rauljordan)!

Note: this will be broken for you. The frontend currently expects a local server to be running:
https://github.com/samajammin/goerli-faucet

I need to:
1) set up the faucet as a remote server
2) add loading, success & error states to the frontend
but in the meantime we can start planning next steps around the flow we want to provide & sort out what's doable by EOD Thursday.

src/components/Button.js
ryancreatescopy: I gotta say I love the flexibility of being able to affect positioning of components without having to create a `StyledButton` but if there's a better way, I'm in!

samajammin:
FYI @ryancreatescopy closing this out for now. I won't delete the branch so we can return to this later.

PR #1657, opened 2020-10-21
Adds clearTimeout() to Morpher Component [Fixes #1317]

### LINKED ISSUE #1317: Morpher component memory leak
**Describe the bug**

The Morpher.js component doesn't `clearTimeout` when the component is unmounted.

Console error:
```
index.js:2177 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
    in Morpher (at pages/index.js:337)
    in h3 (created by pages__H3)
    in pages__H3 (at pages/index.js:336)
    in div (created by pages__Title)
    in pages__Title (at pages/index.js:332)
    in header (created by pages__Header)
    in pages__Header (at pages/index.js:331)
    in div (created by pages__Content)
    in pages__Content (at pages/index.js:328)
    in div (created by pages__Page)
    in pages__Page (at pages/index.js:312)
    in HomePage (created by HotExportedHomePage)
    in AppContainer (created by HotExportedHomePage)
    in HotExportedHomePage (created by PageRenderer)
```

**To Reproduce**

1. Run project locally
2. Got to http://localhost:8888/en/
3. Navigate to a different page
4. View console error

**Expected behavior**

Update component to set a variable on the `setTimeout` call so we can call `clearTimeout` on the unmount of `useEffect`

wackerow:
### /src/components/Morpher.js
- Established variable `morphTimeout` in root scope of component
- Assigned `setTimeout()` to `morphTimeout` when called inside `update()` function
- Called `clearTimeout(morphTimeout)` in return function of `useEffect` to clear on component unmount, fixing memory leaks
- Added empty array set as second parameter to `useEffect` to prevent unnecessary re-rendering

#### Note:
Also adjusted words list order, moving `"Ethereum"` from first to last. The page initially renders the value in `state.text` which initializes to `"Ethereum"` already. Before this adjustment, `Ethereum` would render twice in a row when component loads, and cycle would end on `ఇథిరియూమ్`. A known bug is that this continues to reload the last word on the list without cycling back around. This can be fully addressed in a separate PR, but this briefly patches the glitchy appearance of this so `Ethereum` does not load twice in a row, and also so it ends on `Ethereum` then stops any further re-renders by calling `clearInterval(morphInterval)` in the `else` block of the `setInterval(() => {...}, 3000)`

## Related Issue #1317


samajammin (approved):
Awesome! Thanks for this.

samajammin:
@wackerow feel free to create an issue & PR for the other bug you mentioned (that it fails to cycle back through the list). Would appreciate it! 😁

Issue #1650, opened 2020-10-19
Add "Previous" & "Next" links to Developer Docs pages


samajammin:
**Is your feature request related to a problem? Please describe.**

Currently all documentation pages are "dead ends" unless we manually add links at the end of the markdown files.

**Describe the solution you'd like**

Add a component to the [docs.js template](https://github.com/ethereum/ethereum-org-website/blob/dev/src/templates/docs.js) that automatically generates "Previous" & "Next" links at the bottom of every page.

Design screenshot:
![Image 2020-10-19 at 2 17 52 PM](https://user-images.githubusercontent.com/8097623/96512811-e6291f80-1215-11eb-978a-9f9aa5b2ca4f.png)
(on mobile, these buttons should stack)

Figma design file:
https://www.figma.com/file/dclyxiMIYeAGUSgJoV3xdc/Sketches?node-id=34%3A1934

The logic for how the links should be generated should follow the sequential structure of [/data/developer-docs-links.yaml](https://github.com/ethereum/ethereum-org-website/blob/dev/src/data/developer-docs-links.yaml). Specifically:
- If current page is the first item of a section (e.g. /developers/docs/intro-to-ethereum/), hide "Previous" link
- If current page is within a section, link "Previous" and "Next" to its siblings
   - e.g. if the page is /developers/docs/smart-contracts/anatomy/
      - "Previous" should be /developers/docs/smart-contracts/languages/
      - "Next" should be /developers/docs/smart-contracts/libraries/
  - e.g if the page is /developers/docs/smart-contracts/
      - "Previous" should be /developers/docs/ethereum-stack/
      - "Next" should be /developers/docs/development-networks/ (since it's the next sibling, even though this page has children pages)
- If current page is the last item of a section, link "Next" to first item of next section
   - e.g. /developers/docs/mining/ should link "Next" to "/developers/docs/ethereum-stack/"
- If current page is the last item of last section (i.e. /developers/docs/layer-2-scaling/), hide "Next" link

Ideally the implemented solution doesn't hardcode any paths into the logic (e.g. don't check if location.pathname === "/developers/docs/layer-2-scaling/" in order to hide the "Next" link, instead can you find a solution to traverse the tree & identify if it's the last node?)

**Describe alternatives you've considered**
- Manually set next & previous of each page 👎 

**Additional context**
When implementing your new component, use our `<Link />` component to perform the linking.

wackerow:
Taking a look at this now if you want to assign me!

PR #1670, opened 2020-10-23
Adds "Previous" & "Next" links to Developer Docs pages [Addresses #1650]

wackerow:
## Description
Added custom component `src/components/DocsNav.js`

Per feature requested:
- [x] Add a component to the docs.js template that automatically generates "Previous" & "Next" links at the bottom of every page
- [x] on mobile, these buttons should stack
- [x] Figma design file: https://www.figma.com/file/dclyxiMIYeAGUSgJoV3xdc/Sketches?node-id=34%3A1934
The logic for how the links should be generated should follow the sequential structure of /data/developer-docs-links.yaml. Specifically:
- [ ] If current page is the first item of a section (e.g. /developers/docs/intro-to-ethereum/), hide "Previous" link
_(not exactly...)_
- [ ] If current page is within a section, link "Previous" and "Next" to its siblings
_(not exactly...)_
- [x] If current page is the last item of a section, link "Next" to first item of next section
e.g. /developers/docs/mining/ should link "Next" to "/developers/docs/ethereum-stack/"
- [x] If current page is the last item of last section (i.e. /developers/docs/layer-2-scaling/), hide "Next" link
- [x] Ideally the implemented solution doesn't hardcode any paths into the logic (e.g. don't check if location.pathname === "/developers/docs/layer-2-scaling/" in order to hide the "Next" link, instead can you find a solution to traverse the tree & identify if it's the last node?)
- [x] When implementing your new component, use our <Link /> component to perform the linking.

### Approach taken
This is a slightly different, simpler approach to the ordering of the links, using a more linear manner (can obviously be modified further). Items returned from `developer-docs-links.yaml` are iterated through and added to an array in order recursively (going deep first), producing a list of all possible documents that can be linked to, _in order exactly as shown in the left_ `SideNav`.
We find the current index of this array by matching the "to" field to the current doc page path, then the `currentIndex - 1` will provide data for `Previous` button, and `currentIndex + 1` provides data for `Next` button, with logic to hide button if out of range (no previous button if we're on very first page, and no next button if we're on the very last)

#### Beginning of stack*:
image
(*updated after 75a8ec4)
#### End of stack:
image
#### Also in light mode:
image
#### Handling longer title names:
image
#### Smaller screens:
image
#### Even smaller screens
image

## Related Issue #1650 
@samajammin As noted above, not _exactly_ what was requested, but I thought this approach was slightly more direct, and could always be adjusted with another commit. 
And it works, and is better than nothing =)

Let me know what you think!


ryancreatescopy:
Hey! This looks like a great start. I'm not sure about omitting "Prev" or "Next" if it's the first/last doc in a section now that I see it. I think we want to encourage continued exploration into a different section. The design of this component was also more a sketch than final design so I think either way I'll want to keep the PR open until we can tweak the UI for this component as it's not quite right now. Thanks for jumping in!

wackerow:
Awesome... yeah please feel free to drop any suggestions, updates to the sketch, or just take over the code =)
Was a little tricky dealing with longer document title names, like "Integrated Development Environments (IDES)" (hence why I made note of that in the screenshot above), and looking at it closely again, there are some alignment issues I'll try to straight out.

wackerow:
### Old (13b5114)
image
image
### New (ebe2f3e overlay to show alignment)
image
#### 1-line
image
#### 2-line wrap
image
#### 3-line wrap
image

This change evenly breaks up the vertical height (82px - 2px for border = 80px) into 5 even rows of height 16px (1rem). The top row is always empty as a margin. Second row is always "previous/next" text (aligning top with top of emoji). Short text will display on row 4 to align bottom with bottom of emoji. Text longer than 1 line will start on row 3 (directly below "previous/next") and wrap to row 4... and if it's even longer it will then spill into the bottom margin (row 5). 
The emoji is given `size={3}` to be 3rem tall, with 1rem of margin on all sides, keeping the an even grid of 1rem spacing, and giving it a fitting total height of 5rem.

samajammin (approved):
This looks great @wackerow - thanks again!

We can make further design tweaks if needed in future PRs. I think this is good to go.

PR #1678, opened 2020-10-25
Fix GWEI <-> ETH Unit Conversion

joncursi:
In the simple transaction example, there appears to be a unit conversion error between GWEI <-> ETH.

## Description

```
21000 GAS * 200 GWEI/GAS = 4200000 GWEI

4200000 GWEI = 4200000000000000 WEI = 0.0042 ETH
```


samajammin (approved):
Thanks for catching this 😃

Issue #774, opened 2020-02-21
Developer resources for layer 2


samajammin:
**Is your feature request related to a problem? Please describe.**

There's currently no page or developer resources section on layer 2 solutions / options.

There is a section on "Scaling Ethereum", which provides an overview of solutions:
https://ethereum.org/learn/#improving-ethereums-scalability

But as a developer I'm not provided with any projects or tools to check out or build on.

**Describe the solution you'd like**

The use case I'd like to solve for is the Ethereum developer who is looking explore layer 2 options to build on, i.e. live projects that have tools / SDKs that I can incorporate into my project. Maybe I'm building a game & want to figure out how to use state channels or compare options.

**Describe alternatives you've considered**

1. Create a new dedicated page to this topic
2. Add a section to the /developers page

**Additional context**

Some resources we may be able to build off:

State of Layer 2 Development (3-part series)
- https://media.consensys.net/the-state-of-ethereum-layer-2-protocol-development-2-f22b2603abd6
- https://media.consensys.net/the-state-of-layer-2-protocol-development-2-39d45e98cc29
- https://media.consensys.net/the-state-of-layer-2-protocol-development-3-752f4742aef5

Awesome Layer 2
- https://github.com/Awesome-Layer-2/awesome-layer-2

Consensys tools list (Developer Guides for 2nd Layer Infrastructure section)
- https://github.com/ConsenSys/ethereum-developer-tools-list#developer-guides-for-2nd-layer-infrastructure

EthHub pages on Layer 2 Scaling
- https://docs.ethhub.io/ethereum-roadmap/layer-2-scaling/zk-rollups/

Block 123 List of Awesome Layer 2 Projects
- https://www.block123.com/en/feature/awesome-layer-2-list

samajammin:
Reopening.

samajammin:
This issue should be solved as part of #1395 - we welcome contributions on it!

samajammin:
We now have a live page: https://ethereum.org/en/developers/docs/layer-2-scaling/

I'm going to keep this issue open, as the page still needs help. We welcome any PRs!

samajammin:
Solved in #1683 & #1857 - thanks @ryancreatescopy!

PR #1683, opened 2020-10-26
Added Optimistic rollups info

ryancreatescopy:
## Description

Improved our layer 2 docs by adding more information about optimistic rollups. This is based on extra context provided by Ben Jones' ETHOnline talk (https://youtu.be/96UfJPYyFcs?t=18863)

## Related Issue

Relates to #774 but doesn't necessarily solve it.


src/content/developers/docs/layer-2-scaling/index.md
```diff
@@ -18,6 +18,53 @@ You should have a good understanding of all the foundational topics. Implementin
 - It can be unnecessarily expensive to use blockchain applications
 - Any updates to scalability should not be at the expense of decentralization of security – layer 2 builds on top of Ethereum.
 
+## Rollups {#rollups}
+
+Rollups are solutions that bundle or "roll up" sidechain transactions into a single transaction and generate a cryptographic proof, known as a SNARK (succinct non-interactive argument of knowledge). Only this proof is submitted to the main chain.
+
+_Sidechains are Ethereum-compatible, independent blockchains._
+
+In other words, rollups mean that all state and execution is handled in sidechains – signature verification, contract execution, etc. The main Ethereum chain only stores transaction data.
```
samajammin, line 27:
```suggestion
In other words, rollups mean that all state and execution is handled in sidechains – signature verification, contract execution, etc. The main Ethereum chain (layer 1) only stores transaction data.
```

samajammin: Perhaps more than just here, I feel like this page needs an explanation that layer 1 == the main chain.

src/content/developers/docs/layer-2-scaling/index.md
```diff
@@ -18,6 +18,53 @@ You should have a good understanding of all the foundational topics. Implementin
...
+**Useful for:**
+
+- reducing fees for users
+- open participation
+- fast transaction throughput
+
+There are two types of rollups:
+
+- Zero knowledge – can only handle simple transactions
+- Optimistic – can run anything Ethereum can using an optimistic version of the EVM, the Optimistic virtual machine
```
samajammin, line 40: I think the primary differentiation between ZK & optimistic rollups is the proof mechanism, i.e. validity proofs vs. fraud proofs: 
- https://support.deversifi.com/en/docs/77-what-is-the-difference-between-zk-rollup-and-optimistic-rollup
- https://medium.com/starkware/validity-proofs-vs-fraud-proofs-4ef8b4d3d87a
- https://medium.com/starkware/validity-proofs-vs-fraud-proofs-strike-back-4d0bf90eed15

From my understanding, it's these security models that creates the core trade-offs between these 2 approaches.

src/content/developers/docs/layer-2-scaling/index.md
```diff
@@ -18,6 +18,53 @@ You should have a good understanding of all the foundational topics. Implementin
...
+
+![Diagram showing what happens when a fraudulent transaction occurs in an Optimistic rollup in Ethereum](./optimistic-rollups.png)
+
+The gas you need to run the computation of the fraud proof is even reimbursed. Ben Jones from Optimism describes the bonding system in place:
+
+"_anyone who might be able to take an action that you would have to prove fraudulent to secure your funds requires that you post a bond. You basically take some ETH and lock it up and you say "Hey, I promise to tell the truth"... If I don't tell the truth and fraud is proven, this money will be slashed. Not only does some of this money get slashed but some of it will pay for the gas that people spent doing the fraud proof_"
+
+So you get reimbursed for proving fraud.
+
+#### The Optimism virtual machine (OVM)
```
samajammin, line 60: I think this is a decent overview of the OVM but for us to go into this, I think there needs to be an explanation that Optimism (& the OVM approach) is not the only implementation of optimistic rollups. e.g. Fuel Labs has an implementation of optimistic rollups that uses a [UTXO based transaction system](https://docs.fuel.sh/v1.0.0/Concepts/Fundamentals/Fuel%20Overview.html#architectureahighlevelview) so as a result, it doesn't have the OVM or EVM-like execution environment. It optimizes for a limited use case:  [ERC-20 transfers](https://docs.fuel.sh/v1.0.0/Introduction/Use%20Cases.html).

In short, I think we need to explain that within rollups, there are 2 approaches in terms of security model:
- ZK rollups (validity proofs)
- Optimistic rollups (fraud proofs)

Even within these 2 categories, there are still a range possible implementations, with multiple projects taking different approaches that also have their own set of tradeoffs.

samajammin (changes requested):
Solid start! Left a few comments that I think we should address before merging, primarily:
- Explaining the security models as the key differentiation between ZK & optimistic rollups
- Explaining that there's multiple implementation approaches within optimistic rollups

src/content/developers/docs/layer-2-scaling/index.md
ryancreatescopy: I didn't include content about ZK-rollups in the scope of this update as I only did the research on Optimistic. If we think it's important context for understanding Optimistic we can add more info on ZK-rollups before merging.

src/content/developers/docs/layer-2-scaling/index.md
ryancreatescopy: Ah, I didn't realise OVM was just one brand's implementation of Optimistic rollups. Perhaps we should remove this section rather than us have to commit to explaining every brand's implementation. We could replace it with links to OVM and UTXO etc...

src/content/developers/docs/layer-2-scaling/index.md
samajammin: I think we should at least explain the difference in terms of validity proof vs. fraud proof. As mentioned elsewhere, it's possible that optimistic rollup implementations `can only handle simple transactions` as well, so I don't think that's a good way to define the 2 types of rollups.

src/content/developers/docs/layer-2-scaling/index.md
samajammin: Ya I'm fine with just removing the section on OVM. Perhaps in the general rollups section just explain that there's different implementations that optimize for different sets of tradeoffs.

src/content/developers/docs/layer-2-scaling/index.md
```diff
@@ -18,6 +18,61 @@ You should have a good understanding of all the foundational topics. Implementin
...
+**Useful for:**
+
+- reducing fees for users
+- open participation
+- fast transaction throughput
+
+There are two types of rollups with different security models:
+
+- Zero knowledge: runs computation off-chain and submits a <b>validity proof</b> to the chain
+- Optimistic: assumes transactions are good and only runs computation, via a <b>fraud proof</b>, in the event of a challenge
```
samajammin, line 40:
```suggestion
- Optimistic: assumes transactions are valid by default and only runs computation, via a **fraud proof**, in the event of a challenge
```

src/content/developers/docs/layer-2-scaling/index.md
```diff
@@ -18,6 +18,61 @@ You should have a good understanding of all the foundational topics. Implementin
...
+
+**Useful for:**
+
+- reducing fees for users
+- open participation
+- fast transaction throughput
+
+There are two types of rollups with different security models:
+
+- Zero knowledge: runs computation off-chain and submits a <b>validity proof</b> to the chain
```
samajammin, line 39:
```suggestion
- Zero knowledge: runs computation off-chain and submits a **validity proof** to the chain
```

src/content/developers/docs/layer-2-scaling/index.md
```diff
@@ -18,6 +18,61 @@ You should have a good understanding of all the foundational topics. Implementin
...
+- Optimistic: assumes transactions are good and only runs computation, via a <b>fraud proof</b>, in the event of a challenge
+
+### Zero knowledge rollups {#zk-rollups}
+
+Zero knowledge rollups, also known as ZK-Rollups, bundle hundreds of transfers off-chain into a single transaction via a smart contract. From the data submitted, the smart contract can verify all of the transfers that are included. This is known as a validity proof.
+
+With a ZK rollup, validating a block is quicker and cheaper because less data is included. You don't need all the transaction data to verify the transaction, just the proof.
+
+The sidechain where ZK rollups happen have optimised a few things to reduce transaction size further. A user signs up and gets an index rather than an address... this reduces a transaction from 32 bytes to just 4 bytes.
```
samajammin, line 48:
```suggestion
The sidechain where ZK rollups happen can be optimised to reduce transaction size further. For instance, an account is represented by an index rather than an address, which reduces a transaction from 32 bytes to just 4 bytes.
```

src/content/developers/docs/layer-2-scaling/index.md
```diff
@@ -18,6 +18,61 @@ You should have a good understanding of all the foundational topics. Implementin
...
+
+### Optimistic rollups {#optimistic-rollups}
+
+Optimistic rollups use a side chain that sits in parallel to the main Ethereum chain. They can offer improvements in scalability because they don't do any computation by default. Instead, after a transaction they propose the new state to mainnet. Or "notarise" the transaction.
+
+As computation is the slow, expensive part of using Ethereum, this can offer up to 10-100x improvements in scalability dependent on the transaction. This number will increase even more with the introduction of the Eth2 upgrade: [shard chains](/en/eth2/#shard-chains). This is because there will be more data available in the event that a transaction is disputed.
```
samajammin, line 54:
```suggestion
As computation is the slow, expensive part of using Ethereum, this can offer up to 10-100x improvements in scalability dependent on the transaction. This number will increase even more with the introduction of the Eth2 upgrade: [shard chains](/en/eth2/shard-chains). This is because there will be more data available in the event that a transaction is disputed.
```

samajammin (approved):

samajammin (approved):

PR #1699, opened 2020-10-28
Fix internal 301 redirects

samajammin:
## Description
Add trailing slashes & removes capitalization of links to ensure they don't result in 301 redirects (which triggers a page refresh - bad UX).

## Related Issue
None

### DIFF
--- a/src/content/about/index.md
+++ b/src/content/about/index.md
@@ -49,7 +49,7 @@ This means the website needs to handle many different user journeys, from “a d
 
 ## Design principles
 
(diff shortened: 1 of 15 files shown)


samajammin:
FYI @ryancreatescopy something to be aware of - internal links are unfortunately quite finicky & should have:
- trailing slash
- correct capitalization (we should always use lower case for slugs)

ryancreatescopy (approved):
:+1:

PR #1711, opened 2020-10-30
Add styled-system to project

samajammin:
## Description

- Add styled-system to project: https://styled-system.com/
- Move src/components/Theme.js → src/theme.js (since it's not actually a component)

## Related Issue
None


src/components/ButtonLink.js
```diff
@@ -1,17 +1,18 @@
 import React from "react"
 import styled from "styled-components"
+import { margin } from "styled-system"
```
samajammin, line 3: FYI @ryancreatescopy - here's an example of how to add this to a component.

1) import from styled-system (e.g. `import { margin } from "styled-system"`)
2) add to a styled component (e.g. `${margin}`)
3) pass all props the component receives to the styled component (see `...props` code below)

src/components/ButtonLink.js
```diff
@@ -42,19 +43,23 @@ const Secondary = styled(StyledButton)`
...
   return (
-    <Primary to={to} hideArrow={true} className={className}>
+    <Primary to={to} hideArrow={true} className={className} {...props}>
       {children}
     </Primary>
   )
 }
 
+ButtonLink.defaultProps = {
+  mt: `0.5rem`,
```
samajammin, line 62: @ryancreatescopy curious for your thoughts on this. I added this to keep parity with the current designs (default margin-top of 0.5rem on all buttons) but I don't think this is actually needed/desired. Seems kind of strange to me to apply a margin to all buttons.

ryancreatescopy: The desired margin will probably differ dependent on context. So it'd probably be best to remove a default margin IMO as you'll need to add custom margin regardless.

src/components/ButtonLink.js
ryancreatescopy: Ok, I think this makes sense. Looks good.

samajammin:
@ryancreatescopy - I removed the default ButtonLink margin. This is ready for your review!

ryancreatescopy (approved):

Issue #1718, opened 2020-11-02
Add more entries to ethereum.org glossary


ryancreatescopy:
As the first iteration of the ethereum.org glossary is based on Mastering Ethereum there are plenty more terms that have entered the Ethereum lexicon since it was published. This PR is open to anyone who wants to add terms to the glossary. Some examples, I can think of straight away:

- ZK snarks
- ZK rollups
- Optimistic rollups
- Layer 2
- Sidechains
- Rollups
- Shard chains
- Staking
- defi
- stablecoin

This list is not exhaustive, there are plenty more, so feel free to open a PR with any terms you feel are important.

### Acceptance criteria
- [ ] Descriptions include anchor links to other glossary terms if mentioned
- [ ] Where there are specific ethereum.org pages that cover the term, include a link so the user can get more information
- [ ] The term is added under the correct letter
- [ ] The term title includes a `{#term-title}` so that it can be access via the table of contents and anchor links.
- [ ] The term isn't a product advertisement

samajammin:
Another glossary source: https://ethdocs.org/en/latest/glossary.html

License is also CC: https://github.com/ethereum/homestead-guide/blob/master/LICENSE-CC-BY-SA-4-POINT-0

samajammin:
Solved in #2015 #1728 #1725

Issue #1717, opened 2020-11-02
Add links to ethereum.org pages in glossary


ryancreatescopy:
**Is your feature request related to a problem? Please describe.**

The ethereum.org glossary references some key concepts that we also have pages for. For example "accounts". Where these pages exist we should provide links to them from the glossary so a user can get more information on a chosen topic.

wackerow:
Feel free to assign me =)

PR #1724, opened 2020-11-02
Glossary-to-docs linking [Fixed #1717]

wackerow:
## Description
- Added as many links as reasonably possible to other parts of ethereum.org website from terms in glossary (Most of these were to the developer docs pages for continued learning)
- Utilized 👉 emoji on links to other parts of ethereum.org website (differentiating from internal glossary # links, and external links)
- Standardized text for these links to "More about ____" for simplicity and cleaner UI/UX on mobile
- Also added term "token standard" and "gwei" with appropriate linking (ref: #1718)
- Refactored superscript text used in currency definition exponents. Removed unicode superscript, replaced with `<sup></sup>` for markdown interpretation. 

## Related Issue
Fixes #1717
Contributes to #1718

PR #1725, opened 2020-11-02
Adds glossary term content [contributes to #1718]

wackerow:
## Description
Added definitions for **defi**, **stablecoin**, **staking**.

_Note: submitted this as an independent PR from #1717 changes so terms could be reviewed separately, however, I did link to `#token-standard` in this which is dependent on the new definition slipped into #1717. Also utilized new emoji pattern suggested in #1717._

## Related Issue
#1718


src/content/glossary/index.md
```diff
@@ -92,6 +92,12 @@ Decentralized application. At a minimum, it is a [smart contract](#smart-contrac
 
 See [non-fungible token (NFT)](#nft)
 
+### defi {#defi}
+
+Short for "decentralized finance," a broad category of [Dapps](#dapp) with the intent of providing a wide range of financial services in a decentralized, peer-to-peer manner, without the need for a trusted third party, but instead utilizing [smart contracts](#smart-contract). How truly decentralized a given defi application may be involves many factors, and the truth often exists somewhere on a spectrum between completely centralized, and completely decentralized.
```
ryancreatescopy, line 97: Due to the long sentence structure and technical subject matter this is quite hard to read. Can we break this definition up into smaller sentences? We also probably don't need the sentence on "how decentralized a given defi application may be" as this is relevant for dapps in general.  Maybe we zero in on the fact that defi aims to offer the crypto equivalent of financial services without any of the gatekeeping, so anyone with an internet connection can benefit from financial services.

src/content/glossary/index.md
```diff
@@ -424,6 +430,18 @@ A procedural (imperative) programming language with syntax that is similar to Ja
 
 A [hard fork](#hard-fork) of the Ethereum blockchain, which occurred at block 2,675,000 to address more denial-of-service attack vectors and clear state (see [Tangerine Whistle](#tangerine-whistle)). Also, a replay attack protection mechanism.
 
+### stablecoin {#stablecoin}
+
+An [ERC-20 token](#token-standard), backed by a reserve asset, with price pegged to real-world asset value. Examples of uses include pegging an ERC-20 tokens value to the value of fiat currency, precious metals, or other cryptocurrencies.
```
ryancreatescopy, line 435: Suggested edit:

"An ERC-20 token with a value pegged to another asset's value. There are stablecoins backed by fiat currency like dollars, precious metals like gold, and other cryptocurrencies like Bitcoin."

src/content/glossary/index.md
```diff
@@ -424,6 +430,18 @@ A procedural (imperative) programming language with syntax that is similar to Ja
...
+
+<a href="/en/eth/#tokens">More about tokens<Emoji text=":point_right:" size={1} marginLeft={0.75} /></a>
+
+### staking {#staking}
+
+The act of locking up a quantity of [ether](#ether) (your stake in the network) to earn the right to validate [transactions](#transaction) and propose [blocks](#block) under a [proof-of-stake](#pos) consensus model. For [Ethereum 2.0](#serenity), each validator requires staking 32 ether by depositing to a [smart contract](#smart-contract), which enables each validator to attest to transactions and propose blocks. Being a validator carries responsibilities, and comes with rewards for fulfilling them, and varying degrees of penalties for not.
```
ryancreatescopy, line 441: I think we should capitalise "Ether". 

Also, I think this definition could be simplified somewhat – not sure we need to mention the smart contract or attestations. Also, we'll have dedicated Eth2 and staking pages to link users to soon, for more info.

A suggested edit:

"Depositing a quantity of Ether (your stake) to become a validator and secure the network. A validator checks transactions and proposes blocks under a proof-of-stake consensus model. Staking gives you an economic incentive to act in the best interests of the network. You'll get rewards for carrying out your validator duties, but lose varying amounts of ETH if you don't."

ryancreatescopy (changes requested):
Thanks for adding these. Some small content edits suggested.

src/content/glossary/index.md
wackerow: 👍 That sounds good to me, especially as long as we can link to more info.

The note about 'ether' being capitalized I would honestly argue against though. Per our own white paper:
image
Just like 'dollars' or 'bitcoin', I always use 'ether' (or all capital tickers 'USD' 'BTC' 'ETH') when referring to the currency personally. 

That being said, searching the repo for each spelling I see a _lot of both_. So, at the very least I do think it's worth clarifying this and maybe making an issue to standardize across the site.

src/content/glossary/index.md
wackerow: How does this sound?
`Short for "decentralized finance," a broad category of [Dapps](#dapp) aiming to provide financial services backed by the blockchain, without any gatekeepers, so anyone with an internet connection can participate.`

src/content/glossary/index.md
ryancreatescopy: Ok, let's go for `ether` and `ETH` – will open a separate PR to address this!

ryancreatescopy (approved):

src/content/glossary/index.md
```diff
@@ -92,6 +92,12 @@ Decentralized application. At a minimum, it is a [smart contract](#smart-contrac
 
 See [non-fungible token (NFT)](#nft)
 
+### defi {#defi}
+
+Short for "decentralized finance," a broad category of [Dapps](#dapp) aiming to provide financial services backed by the blockchain, without any intermediaries, so anyone with an internet connection can participate.
+
+<a href="/en/dapps/#how-to-use-an-application-built-on-ethereum">More about Dapp uses<Emoji text=":point_right:" size={1} marginLeft={0.75} /></a>
```
samajammin, line 99:
```suggestion
[More about Dapp uses <Emoji text=":point_right:" size={1} marginLeft={0.75} />](/dapps/#how-to-use-an-application-built-on-ethereum)
```

samajammin: We can & should continue to use markdown syntax where possible.

Note: we can & should also drop the language path (/en/ in this case). When we use markdown links, Gatsby uses [our Link component](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Link.js) under the hood, which automatically adds the language path. This makes it easier when translating pages - we don't have to update the links everywhere.

src/content/glossary/index.md
```diff
@@ -424,6 +430,18 @@ A procedural (imperative) programming language with syntax that is similar to Ja
 
 A [hard fork](#hard-fork) of the Ethereum blockchain, which occurred at block 2,675,000 to address more denial-of-service attack vectors and clear state (see [Tangerine Whistle](#tangerine-whistle)). Also, a replay attack protection mechanism.
 
+### stablecoin {#stablecoin}
+
+An [ERC-20 token](#token-standard) with a value pegged to another asset's value. There are stablecoins backed by fiat currency like dollars, precious metals like gold, and other cryptocurrencies like Bitcoin.
+
+<a href="/en/eth/#tokens">More about tokens<Emoji text=":point_right:" size={1} marginLeft={0.75} /></a>
```
samajammin, line 437: Same deal here.

src/content/glossary/index.md
```diff
@@ -424,6 +430,18 @@ A procedural (imperative) programming language with syntax that is similar to Ja
...
+
+### staking {#staking}
+
+Depositing a quantity of [ether](#ether) (your stake) to become a validator and secure the [network](#network). A validator checks [transactions](#transaction) and proposes [blocks](#block) under a [proof-of-stake](#pos) consensus model. Staking gives you an economic incentive to act in the best interests of the network. You'll get rewards for carrying out your validator duties, but lose varying amounts of ETH if you don't.
+
+<a href="/en/eth2/get-involved/">More about staking<Emoji text=":point_right:" size={1} marginLeft={0.75} /></a>
```
samajammin, line 443: Same deal here.

samajammin (changes requested):
Thanks for this! See comments - let's just update the links to markdown syntax.

wackerow:
@samajammin Only reason I didn't use markdown for those links was because doing that left an underlined space before the Emoji (the 'marginLeft' prop doesn't seem to be passing this way), and I didn't want to remove the emoji from the link.
image
...But I get your point about taking advantage of the MD parsing... Happy to refactor this back to markdown, but working on a new component suggested by @ryancreatescopy that will change this entirely (#1737). If this is causing linking issues with `/en/` baked in I can fix this first if preferred (same applies to the links added in #1724)

samajammin:
Hmm good point. Notice we did refactor the Emoji component. You should now pass a string for margin left, e.g. ml={'1 rem'}

src/content/glossary/index.md
wackerow: @samajammin 👍 Will definitely remove the /en/ language path. (Note though that all `static.js` pages will interpret the anchor tags into the custom `Link` component as well, not just when using markdown... custom component in the works will do the same):
image

I'll fix these now to get this pulled in first.

src/content/glossary/index.md
```diff
@@ -424,6 +430,18 @@ A procedural (imperative) programming language with syntax that is similar to Ja
...
+
+An [ERC-20 token](#token-standard) with a value pegged to another asset's value. There are stablecoins backed by fiat currency like dollars, precious metals like gold, and other cryptocurrencies like Bitcoin.
+
+[More about tokens<Emoji text=":point_right:" size={1} ml={`0.75rem`} />](/eth/#tokens)
+
+### staking {#staking}
+
+Depositing a quantity of [ether](#ether) (your stake) to become a validator and secure the [network](#network). A validator checks [transactions](#transaction) and proposes [blocks](#block) under a [proof-of-stake](#pos) consensus model. Staking gives you an economic incentive to act in the best interests of the network. You'll get rewards for carrying out your validator duties, but lose varying amounts of ETH if you don't.
+
+[More about staking<Emoji text=":point_right:" size={1} ml={`0.75rem`} />](/eth2/get-involved/)
```
samajammin, line 443:
```suggestion
[More about staking<Emoji text=":point_right:" size={1} ml={`0.75rem`} />](/eth2/staking/)
```

samajammin (approved):
🙌

samajammin (approved):

PR #1732, opened 2020-11-03
Replace EVMC with evmone in EVM implementations list

gumb0:
EVMC is not an EVM implementation itself, it's an interface between EVM implementations and EVM users.

evmone is such EVM implementation.


samajammin (approved):
Thanks for catching & updating!

samajammin:
@all-contributors please add @gumb0 for content.

PR #1733, opened 2020-11-03
Increase top to account for Nav

samajammin:
## Description
Tweak `top` of sticky items to account for Nav.

Not sure what caused it but current overlap is too close:
Image 2020-11-03 at 7 23 19 AM

## Related Issue
None


ryancreatescopy (approved):
:+1:

PR #1721, opened 2020-11-02
Updated nav structure and added new home icon

ryancreatescopy:
## Description

Reorganising the nav structure of ethereum.org to house future content.

### Todo: 
- [ ] update footer with same structure as nav bar
- [x] addd aria labels to new navbar dropdowns

PR #1738, opened 2020-11-04
Updated nav structure and added new home icon [Extends PR #1721]

wackerow:
## Description
 - [x] new home icon / favicon
 - [x] add aria labels to new navbar dropdowns
 - [x] cleaned up desktop/mobile 'developer' with new dropdown menu
 - [x] 'community' pulled out to own category
 - [x] update footer with same structure as nav bar (plus 'ecosystem' and 'about' sections)
 - [x] footer spacing accommodates new additional columns, and stacks short categories, using css grid
 - [x] footer conditionally shows categories based on `shouldDisplay` prop

## Related _PR_
_Extension of #1721 (but on a branch I have permission to work on 😉)_


ryancreatescopy (dismissed):

src/components/Footer.js
```diff
@@ -325,31 +357,33 @@ const Footer = () => {
...
-                          </FooterLink>
-                        </ListItem>
-                      )
-                    })}
-                </List>
-              </LinkSection>
-            )
-          })}
+          <LinkGrid>
+            {linkSections.map((section, idx) => {
```
samajammin, line 361: This logic needs to be changed a bit. We now also need a `shouldDisplay` prop for each section, then a filter here to check if that prop evaluates to true. Reason being, some of them we should hide for languages.

e.g. check out the footer for Czech (content v1.0): https://deploy-preview-1738--ethereumorg.netlify.app/cs/
We should hide both Community & Enterprise sections since there's no links to display

e.g. check out the footer for German (content v1.1: https://deploy-preview-1738--ethereumorg.netlify.app/de/
We should hide the Community section since there's no links to display

So Enterprise should display for v1.1 or higher. Community should display for versions higher than v1.1.

samajammin (changes requested):
This looks great! Just need the Footer updated to account for different language versions & conditionally display those sections. See my comment.

src/components/Footer.js
wackerow: Totally agree... see f3c1ae3 below, should fix it right up.

src/components/Nav/index.js
```diff
@@ -295,32 +392,9 @@ const Nav = ({ handleThemeChange, isDarkTheme, path }) => {
...
     },
-    {
-      text: "page-glossary",
-      to: "/glossary/",
-      shouldDisplay: contentVersion > 1.1,
-    },
   ]
   let mobileLinkSections = cloneDeep(linkSections)
 
-  // If contentVersion includes EDN (>1.1), strip out Developers links
```
samajammin, line 306: Do we want to remove this logic? I don't think so but @ryancreatescopy please confirm. My understanding is we want to keep the single "Developers" link in main nav, since there's a subnav. Currently Developers now has both the dropdown & the subnav, which feels awkward:
Image 2020-11-11 at 3 00 45 PM

src/components/Nav/index.js
```diff
@@ -262,11 +322,48 @@ const Nav = ({ handleThemeChange, isDarkTheme, path }) => {
         },
       ],
     },
+    {
```
samajammin, line 325: See comment below - not sure we actually want this.

src/components/Nav/index.js
```diff
@@ -188,10 +184,74 @@ const Nav = ({ handleThemeChange, isDarkTheme, path }) => {
...
+          to: "/wallets/",
+          shouldDisplay: contentVersion > 1,
+        },
+      ],
+    },
+    {
+      text: "page-home-section-learn-title",
+      ariaLabel: "nav-learn-aria-label",
+      shouldDisplay: contentVersion > 1.1,
+      items: [
```
samajammin, line 223: @ryancreatescopy what's your thoughts on the ordering of these links? My 2 gwei: I think "Guides and resources" should probably be at/near the top given that it's more intro-level vs. EIPs.

samajammin (changes requested):
This is looking great! Left a few comments which I'd like to see resolved. A couple we will want @ryancreatescopy to weigh in on.

You'll also want to rebase off `dev` to fix en.json merge conflicts.

samajammin:
@wackerow sorry to throw new changes at you but please note #1750 - now for English (content versions > 1.1) we have 2 enterprise pages that should be listed in the footer.

src/components/Footer.js
```diff
@@ -33,7 +33,22 @@ const LastUpdated = styled.div`
   color: ${(props) => props.theme.colors.text200};
 `
 
+const LinkGrid = styled.div`
```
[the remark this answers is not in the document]
wackerow: Any preference?
#### justify-content: space-around;
image
#### justify-content: space-between;
image

And did you mean 2 columns _up to `breakpoints.l`_ or _larger than `breakpoints.l`_?
In either case, I have a fix that does:
screen size < `breakpoints.m` = 1 column (2 columns easily get clipped in this range)
between `.m` and `.l` = 2 columns (see above images for options)
larger than `.l` = Full 6 columns (which... do okay, but window's just above that threshold get a little cramped... could maybe modify the breakpoint to require slightly larger before showing all 6.

src/components/Nav/index.js
wackerow: (I agree... Think it belong first, then whitepaper which started it all, then EIPs which evolved it, then move on to Eth2, and cap it with the glossary for reference... just my two wei... and note my two wei are 10^-9 as valuable as sam's 2 gwei hah)

src/components/Nav/index.js
wackerow: Yeah either way... not my call of course, but I'll just mention I've found it nice having that dropdown there while testing. 
I like being able to toggle that from any page and quickly see the sub-sections (just like with Use Ethereum and Learn)... otherwise it requires navigation to load the Developers page before you can see the Docs/Tutorials/etc options.
Example: I want to get to the tutorials from some random page, but I forget if it's part of Learn, or Developers. From any page (without leaving that page) I can peak at both dropdown menus to confirm before navigating to Developers.

Totally get your point in the screenshot though, makes it look a little weird if you're already on Developers and click that 🤷‍♂️

ryancreatescopy: @samajammin I added it back in. I don't think it's that harmful for users to have both menus while on `/developers/` and figured it's a better experience that allows users to dive in to a page like tutorials, for example, from anywhere on the site. Happy to discuss this further if you think it creates a bad experience.

src/components/Nav/index.js
ryancreatescopy: I'm happy with that ordering @wackerow – EIPs shouldn't be at the top, that's for sure.

wackerow:
### Summary of changes
- New favicon 
- Aria tags added
- Header sections: "Use Ethereum," "Learn," "Developers", "Enterprise" (All dropdown menus) and "Ethereum Cumminity"
- Footer sections: "Use Ethereum," "Learn," "Developers", "Ecosystem*," "About...*",  and "Ethereum Cumminity" and "Enterprise" (* these two are exclusive to footer)
- "Learn" category reordered to have "Guides and resources" first in both header/footer
- Footer to grid view to accommodate more sections, max 6 columns if screen larger than `breakpoints.l`
- Grid view collapses to 2 columns when screen smaller than `breakpoints.l`, and down to 1 column smaller than `breakpoints.m` (all justify-content: space-between)
- Developers header now has dropdown menu
- Added content version conditionals to footer to only display sections and items that are available in region

Think that's about it to this point. Could be ready for pull if it looks good =)

src/components/Footer.js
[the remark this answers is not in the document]
samajammin: Looks great! That's what I was thinking 😄

samajammin (approved):

samajammin:
Awesome job @wackerow! 🙌

PR #1739, opened 2020-11-04
Standardised spelling of ether (as per whitepaper)

ryancreatescopy:
## Description

The site currently uses `ether` and `Ether` interchangeably. This PR ensures we only use `ether` or `ETH` for short. Exceptions: when `ether` starts a sentence.

I've used `ether` as this is how it is spelled within the whitepaper. You also wouldn't capitalise dollars, cents or pounds when used in a sentence. `ETH` should be used when referring to an amount in a sentence like 32ETH. This is the equivalent of referring to a dollar amount like $32USD.

cc @wackerow


wackerow:
👍 Thanks for doing this! Especially on a primary resource page like this, important to remain consistent. Hah, we're all so proud of Ethereum that it's hard not to capitalize "Ether"!

samajammin (approved):

Issue #1737, opened 2020-11-03
UI/UX Glossary linking enhancement


wackerow:
**Is your feature request related to a problem? Please describe.**
Suggested by @ryancreatescopy -- This is a continuation of #1724 with goal to improve the UI/UX of the links on the glossary page by adding styling to further differentiate between different types of links beneath definitions (internal anchor links, vs links to external pages).

**Describe the solution you'd like**
![image](https://user-images.githubusercontent.com/40891631/97992150-34275100-1dda-11eb-9150-f44180609a03.png)
(pulling in the page title and meta description to populate the content)

**Describe alternatives you've considered**
- Leave internal anchor links with default anchor styling (colored, underlined) and 👉 emoji (as implemented in #1724)
- Leave internal anchor links with default anchor styling (colored, underlined) without emoji indicator

**Additional context**
May wait til other PRs get merged to avoid confusion/conflicts, but will take a look at this.

wackerow:
@ryancreatescopy Have a component started to address this, but I may need your help with writing the query to pull in the `title` and `description` from the document being linked to. 

My thought was to have a component called something like `<DocLink>` with all the templating and styling. Import into the `static.js` and `MDXProvider`. In the glossary `.md` file, this could then be used in the form of `<DocLink to="/en/developers/docs/accounts" />`...

..but I'm not quite sure how to then query using that path to get the correct `.md` document, and parse out the `title` and `description`. Using static data I've got a working version, but would obviously prefer to make it dynamically pull in what's needed with a query for many reasons. 

Lemme know what you think!

samajammin:
We might just have to bite the bullet & hardcode the title & description. Gatsby's GraphQL doesn't support dynamic queries (queries with variables) since it runs them on build: https://www.gatsbyjs.com/docs/static-vs-normal-queries/

samajammin:
(There also may be a clever way of accomplishing this that I just haven't thought of)

wackerow:
Okay, well I guess it wasn't just me then hah... I'll parse these out today

wackerow:
Closed with #1753

PR #1753, opened 2020-11-06
Internal link component [Issue #1737]

wackerow:
## Description
**First draft**: This adds a component, currently called `DocLink` that takes props `to`, `title` and `description` and renders a `styled(Link)` container with the title/description.
#### Screenshots with light/dark mode, and link hover-over
image
image
(View Netlify for this PR: [/en/glossary/](https://deploy-preview-1753--ethereumorg.netlify.app/en/glossary/))

## To discuss/decide/do
- Need to decide how long we're willing to let the description be. Could overshoot and then limit to number of lines and truncate, or could just manually decide how much and what info to put in each one (I erred on the side of putting too much in and in generally copies the first paragraph of each page).
- Some small design related tweaks I was playing around with and settled on this as a first suggestion. I tried the shadow-box approach first and it felt like the component overshadowed the actual glossary entires. Found that border of 2px without the shadow, with a color change and very slight shadow on hover-over felt nice, but this is obviously not my decision to make, just wanted to offer some thoughts and we can go from here.
- Consider changing name of component

_Note- This was created as a branch off dev prior to the other glossary update PRs #1724 and #1725 being merged._

Labeled this as **draft** til we sort out above details, and figured after the other glossary PRs merge in I can fix any conflicts that this causes, and fix any terms missing here. Wanted to get something up for you guys to be able to review though, let me know what you think.

## Related Issue
#1737


ryancreatescopy:
Tried to commit a design alternative but not sure it worked...

Bearing in mind, we want to make sure it's clear to users that they're navigating to another page within the current site but we don't want the design to dominate the glossary entries. 

image
image
image
image


Regarding the copy length... I'm against imposing strict character limits or truncation. Instead we may just need to be very careful to craft as short a description as we can that is still helpful. I can definitely help with this.

wackerow:
#### Screenshots as of b458d28
image
image
image
image

@samajammin @ryancreatescopy This may be ready for review if you're comfortable with the length of the descriptions being used (this could always be addressed later in another PR)

ryancreatescopy:
Just want to throw some alternatives into the mix. Do you think we need to actually show a page description? Or should the title be descriptive enough for these purposes? Have we overthought this and perhaps it'd just be simpler to do emoji + title + arrow...

Just wanted to raise these before we decide and merge something...

image
image
image

wackerow:
@ryancreatescopy Hm... 
So as I was putting this together I did feel that several link "descriptions" were... semi redundant? Especially in cases where the information was almost the same, it seemed like it would just cause extra reading without any new information, and/or take up that space without much benefit.

The description also adds some complexity in terms of the manual trimming required, and it's static so if the MD file it links to changes, the link will no longer match. And alternative here would be what we kinda already discussed, which would be truncating the description maybe to one line and add an ellipsis (...) at the end?

_That being said, I think my honest favorite of the options so far would be that last one:_
image

I think the one with two icons gives a feeling like each line links to something different... and I kinda like including the path with the name, it very clearly feels like an internal link that won't take them too far.

samajammin:
I tend to agree, I think adding a description w/ every link is a bit unnecessary, especially if it's something that must be added manually. IMO I think linking to the page w/ ("More on [topic]...") is sufficient. I don't think displaying the path adds much value, given that we already clearly delineate internal vs external links w/ the arrow icon.

wackerow:
@samajammin Would you suggest using this option then?
image

...or just ditching the component idea entirely and leaving the regular text links as they were?

image

samajammin:
Either works for me, happy to differ to you & @ryancreatescopy.

wackerow:
That means you @ryancreatescopy ! haha (I prefer the look of the component... "buttons" always seem so much more enticing to click hah)
Just lemme know and we can wrap it up one way or the other 👍

ryancreatescopy:
> Either works for me, happy to differ to you & @ryancreatescopy.

@samajammin @wackerow let's go with the page emoji and arrow option. I think this makes them more noticeable and would also work really well in the developer docs. To help advertise related pages, e.g. on the consensus mehanisms page. I'd imaging using DocLink component to advertise PoS and PoW pages.

wackerow:
Merged updated dev branch and fixed conflicts.

This component now takes the format `<DocLink title="" to="" />` to populate a clickable div with hover-over styling:
#### Light mode
image
#### Light mode - hover
image
#### Dark mode
image
#### Dark mode - hover
image
Responsive sizes on mobile, maintaining alignment with larger names
#### Mobile-size demo
image

Baked in margin is zero so this can more easily be reused elsewhere outside of Glossary. Currently this is the only implementation, and all internal glossary links have been refactored to use this.

src/components/DocLink.js
```diff
@@ -0,0 +1,68 @@
...
+const EmojiCell = styled.div`
+  display: flex;
+  align-items: center;
+`
+
+const DocLink = ({ to, title }) => {
+  return (
+    <Container to={to}>
+      <EmojiCell>
+        <Emoji size={1} text=":page_with_curl:" mr={3} />
```
samajammin, line 55:
```suggestion
        <Emoji size={1} text=":page_with_curl:" mr={`1 rem`} />
```

samajammin (approved):
Looks great! Thanks again @wackerow!

samajammin (approved):

samajammin:
@wackerow one minor issue I'm noticing - it appears the entire card is not clickable. See video:
https://share.getcloudapp.com/qGuv2NvD

Ideally the entire component would have the hover effect & link when someone mouses over it.

wackerow:
@samajammin Got it, good catch... I'll fix this in a separate PR 👍

PR #1770, opened 2020-11-11
Adds component import/exports to templates [Fixes #1768]

### LINKED ISSUE #1768: Eth2 Banner bug
**Describe the bug**

Occasionally the eth2 banner doesn't render correctly. I've had this error on both mobile and desktop, pre and pose v2.10.0 deploy. 

**To Reproduce**

1. Go to an /eth2 page
2. Look at banner 

**Expected behavior**

Banner should have a primary colour background and text should be centered.

**Screenshots**

image

wackerow:
## Description
Eth2 Staking banner was displaying incorrectly after loading on a markdown page as a result of the new eth2 template not yet being assigned in gatsby.

- Adds 'eth2' slug to gatsby template check to properly associate any eth2 markdown content files to this template
- Adjusted necessary component imports
- Fixes one small markdown bug (added missing key to a section header)
- Also made "Summary" box in 'eth2 template' render conditionally, only if data is available

## Screenshots
image
image

## Related Issue #1768

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -111,6 +111,8 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
       template = `tutorial`
     } else if (slug.includes(`/docs/`)) {
       template = `docs`
+    } else if (slug.includes(`/eth2/`)) {
+      template = `eth2`
     }
 
     createPage({
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -111,8 +111,6 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
       template = `tutorial`
     } else if (slug.includes(`/docs/`)) {
       template = `docs`
-    } else if (slug.includes(`/eth2/`)) {
-      template = `eth2`
     }
 
     createPage({
--- a/src/templates/eth2.js
+++ b/src/templates/eth2.js
@@ -30,7 +30,6 @@ import Translation from "../components/Translation"
 import TranslationsInProgress from "../components/TranslationsInProgress"
 import Warning from "../components/Warning"
 import SectionNav from "../components/SectionNav"
-import ExpandableCard from "../components/ExpandableCard"
 import { getLocaleTimestamp } from "../utils/time"
 import { isLangRightToLeft } from "../utils/translations"
 import {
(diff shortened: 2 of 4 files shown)


ryancreatescopy:
Hey @wackerow thanks for this! One small issue is that the community grants should use the static.js template — the eth2 template was designed to be a contained section of "chapters" related to the eth2 upgrades. The grants page is a more temporary traditional ethereum.org page

wackerow:
What about the others in the /contents/eth2/ folder?

wackerow:
I ask because it seems those still need this fix, and the `/staking-community-grants/` needs a modifier of some kind, or to be moved out of that folder (so it doesn't use that template that the others need)

gatsby-node.js
```diff
@@ -111,6 +111,8 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
       template = `tutorial`
     } else if (slug.includes(`/docs/`)) {
       template = `docs`
+    } else if (slug.includes(`/eth2/`)) {
```
samajammin, line 114: We don't actually want to apply this blanket rule for Eth2 pages. Some of theme should use the static.js template. That's why I added the frontmatter logic above (to check for the `template` field, which is set to `eth2` on the desired pages.

wackerow:
Okay.. so I reverted those gatsby changes to your points... 
I was getting console warnings about the `CardContainer` originally, and focused on fixing this at least, hoping maybe that will solve the banner issue reported. I imported/exported this in the `static` template since this is where our `staking-community-grants` is looking to. The warning has resolved on my end, and I am no longer able to replicate (so far) the banner issue originally reported.

src/templates/eth2.js
```diff
@@ -192,6 +193,7 @@ const components = {
   Eth2BeaconChainActions,
   Eth2ShardChainsList,
   Eth2DockingList,
+  CardContainer,
```
samajammin, line 196: This isn't needed on this template, at least not yet.

samajammin (changes requested):
Thanks for making this update. See my one comment, we don't need that component within the eth2 template.

src/templates/eth2.js
wackerow: hah, didn't realize this component was individually imported into the other eth2 pages like 'vision' page... so this dwindled down to just importing/exporting `CardContainer` in the static template, and fixing that one small `key` for the `##` header in the markdown file. 
Again, not getting banner issues at the moment 🤞

samajammin (approved):
These changes LGTM! Just a note, I don't believe this actually fixes #1768.

PR #1771, opened 2020-11-11
Eth2 bug fixes (nesting/margin)

wackerow:
## Description
When navigating to `/en/eth2/get-involved/` a console warning generates as a result of a `<ul>` tag being nested within a `<p>` tag. This is in `/pages/eth2/get-involved/index.js`. The `<p>` tag here is not needed and has been removed, eliminating the warning. 

#### Related Issue (none filed):
image

#### Update
Also added commit to update `marginLeft` prop for `<Emoji>` component to use new `ml` format


samajammin (approved):
Thanks!

Issue #1774, opened 2020-11-11
Mobile nav should include search


ryancreatescopy:
**Is your feature request related to a problem? Please describe.**

At the moment the interaction for searching on mobile is not obvious and a little cumbersome. Using a top right icon, followed by a bottom-left icon isn't comfortable and it's not clear where to find search.

**Describe the solution you'd like**

I believe search should be always available on mobile from the header. Navigation is already more difficult on mobile due to a lack of an omnipresent header. Search in a tap from any page would be a nice UX improvement for moving around the site.

**Describe alternatives you've considered**

Alternatively, we could move search to sit at the top of the mobile menu, but I think this is also suboptimal as it doesn't improve the discoverability of the feature.

**Additional context**

samajammin:
Solved in #1775

PR #1775, opened 2020-11-11
Adds search on mobile header [#1774]

ryancreatescopy:
## Description

- Added search icon to header.
- Changes animation of drawer (enters from left)
- Tweaked alignment of close icon on mobile search and mobile search from nav

**Todo**
- [x] Add modal background when pressing the search icon
- [ ] Ensure icons have aria-labels
- [ ] Ensure a user can tab to icons simply via keyboard

## Related Issue

#1774


ryancreatescopy:
Realised, I've removed the X so you can't cancel the mobile menu easily. Need to fix.

samajammin (changes requested):
Awesome! Love the idea to make search easier to access. 

The UI functionality looks great but I don't like the overall duplication of code (having the SearchContainer & it's logic/children in both Nav/index.js & Nav/mobile.js). We should refactor this so it all lives in one place. Happy to pair on this next week / do it myself when there's time available.

ryancreatescopy:
> Awesome! Love the idea to make search easier to access. 
> 
> 
> 
> The UI functionality looks great but I don't like the overall duplication of code (having the SearchContainer & it's logic/children in both Nav/index.js & Nav/mobile.js). We should refactor this so it all lives in one place. Happy to pair on this next week / do it myself when there's time available.

For sure. I had the same thought that it should live separately in one place and we reference it from index and mobile.

zech564 (dismissed):

samajammin:
@ryancreatescopy - refactored this so that:
- State is maintained by the parent Search/Index.js component
- Mobile search UI is consolidated into the Search/Mobile.js component

Please double-check the functionality is what's expected but I believe this achieves our goal.

samajammin (dismissed):

ryancreatescopy:
> @ryancreatescopy - refactored this so that:
> 
> * State is maintained by the parent Search/Index.js component
> * Mobile search UI is consolidated into the Search/Mobile.js component
> 
> Please double-check the functionality is what's expected but I believe this achieves our goal.

Awesome! This pretty much does everything I intended! Couple things:

1. Looks as though the changes have shifted the Search bar, dark mode icon, and Languages navItem to the left on desktop. Compare it with prod to see a small indent. 
2. When you press on the mobile search in the nav, it loads the menu first and then the search drawer slides over it. Could we do it without first loading the menu drawer?

src/components/Nav/Mobile.js
```diff
@@ -205,28 +193,28 @@ const BlankSearchState = styled.div`
...
     <>
       <MobileModal
         animate={isOpen ? "open" : "closed"}
         variants={mobileModalVariants}
         initial="closed"
         onClick={handleClose}
-      ></MobileModal>
+      />
       <MenuContainer
         animate={isOpen ? "open" : "closed"}
```
wackerow, line 219: Change `isOpen` to `isMenuOpen`

src/components/Nav/Mobile.js
```diff
@@ -303,22 +291,21 @@ const MobileNavMenu = ({
...
+            <CloseIconContainer onClick={() => toggleMenu("search")}>
+              <Icon name="close" />
+            </CloseIconContainer>
           </SearchHeader>
           <Search handleSearchSelect={handleClose} />
           <BlankSearchState>
             <Emoji text=":sailboat:" size={3} />
             <Translation id="search-box-blank-state-text" />
           </BlankSearchState>
         </SearchContainer>
```
wackerow, line 308: Move this code block just below closing `</MenuContainer>` tag

src/components/Nav/Mobile.js
```diff
@@ -303,22 +291,21 @@ const MobileNavMenu = ({
...
-        <CloseIconContainer onClick={handleClose}>
-          <Icon name="close" />
-        </CloseIconContainer>
```
wackerow, line 321: (Don't forget this... most mobile devices need this)

wackerow (commented):
These two changes will allow the Search menu to open independently of the menu (eliminating that animation glitch). 

Only small issue here relates to PR #1960 which I posted recently, related to the z-index of the `BottomMenu`. I overlooked that this was still in the works when I posted that... If you pull the `<SearchContainer>` out of the `<MenuContainer>` then a z-index on the `BottomMenu` would not work to force it on top.

That being said, with two dedicated buttons in the nav bar, both for search and for the menu, it may not be worth bothering to have the bar at the bottom while searching as I had previously suggested in #1959, and just keep that exclusively in the menu.

samajammin (approved):
Made the edits you suggested @ryancreatescopy @wackerow - merging in!

PR #1794, opened 2020-11-11
Removes color styling override of external links [Fixes #1788]

### LINKED ISSUE #1788: External warning link
**Describe the bug**

Warning link external icon should match the color of the warning link (e.g. `PrimaryDark800`), currently it displays as `primary`. You can find an example here: http://ethereum.org/en/eth2/shard-chains/

**Screenshots**

image

wackerow:
## Description
This is one possible fix to issue #1788. 
I may be missing something but when I remove the `color: ${(props) => props.theme.colors.primary}` styling that is placed on all external links, this allows all links to keep the color styling from its parent. I tried a few different pages, in light and dark mode, and they all seem to work just fine without that override. Not sure why we could want this, but again, still new so I may be missing something deeper.
image
image
image
image

## Related Issue #1788

### DIFF
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -15,7 +15,6 @@ const isHashLink = (to) => HASH_PATTERN.test(to)
 
 const ExternalLink = styled.a`
   &:after {
-    color: ${(props) => props.theme.colors.primary};
     margin-left: 0.125em;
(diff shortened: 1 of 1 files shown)


samajammin (approved):
Looks to do the trick - nice work!

PR #1808, opened 2020-11-13
Upgrades ProductCard component (adds subjects / GitHub info) [Closes #1782, Closes #1783]

### LINKED ISSUE #1782: Add Github stars to developer products we list

Throughout the developer section, we list developer products but we don't make it easy for users to pick one at a glance. For example: https://ethereum.org/en/developers/local-environment/

A bit of a vanity metric, but it would be useful for some users to see GitHub stars on these cards to help them make decisions quicker about which products to check out first. 

These cards should show the number of github stars the projects have. For any project that isn't open source, we should reconsider listing them.

### LINKED ISSUE #1783: Add programming language labels to developer products we list

Throughout the developer section, we list developer products but we don't make it easy for users to pick one at a glance. For example: https://ethereum.org/en/developers/local-environment/

It would be useful to include a tag or label that identifies the language the product uses, where relevant, to help users choose.

These cards should show a language label relevant to the product. For any project that isn't open source, we should reconsider listing them.

wackerow:
## Description
Adds pill-like component that pulls GitHub repo star count for projects listed on the site.
image
image

Only shows if `gitAccount` and `gitRepo` names are passed. Added these props to `<ProductCard>` components, and updated the data fed to them for sandboxes and frameworks. 

Still need to update all of the /eth2/get-involved listings. --Update: done

## Related Issue #1782


ryancreatescopy:
Should we add the GitHub icon to make it clearer we're pulling in github data?

wackerow:
@ryancreatescopy Any thoughts on where to put it? In the component? A separate indicator for each card? One larger indicator for these whole sections? My thought was squeezing more glyphs into a component like this would make it overly complicated (especially if you have both that and the star) and if someone doesn't know what the stars are, they probably don't care about them.

Just some thoughts, let me know what you think. Keep in mind I'm gonna work on a component with a horizontal/wrapping 'pill' list for each programming language... still trying to find the best way for it to all fit together nicely.

samajammin:
This is amazing!

**Re: the GH icon / star layout, a couple options:**
- Keep as-is visually but have the component be a link to the GH repo, which would make it obvious where the data is coming from. Only potential issue I see with this is a link nested within a link, which is not technically valid HTML. Perhaps we add a button to the cards which act as the link to the project.
-  Display the GH icon with the word "stars", like this common widget:
Image 2020-11-13 at 9 23 39 AM
https://ghbtns.com/

**Re: this approach to querying data:**
Overall I think this is a valid v1 solution. One thing to consider is that this creates a network request on page load for every project we display, which is far from optimal. I think the ideal solution would be to query these project stars via GraphQL at build time (we already have the GH API available). In this case, the data would already be in the static HTML the user loads so we could avoid the delay & added load of all these network requests. The downside with this approach is that stars wouldn't dynamically update but given how often we deploy the site, I don't see this being a concern.

wackerow:
@samajammin Agree with everything, and you've iterated some of the issues I came across in v1. I originally made this a link until I realized it was already nested (at least when used in a `ProductCard`). 
Had some small issues using the `Icons` component in the first draft because of the baked in hover-over coloring, but if we move this out of a nest link, it would no longer be an issue.... just not sure where to put it exactly, since it seems to belong right inside these product cards. 
I'll peak at the ghbtns.com page and mull it over, but if anyone has any thoughts, all ears

**Re: Querying**
Good call, and I agree that at the rate this site gets rebuilt it shouldn't be an issue (stars don't add up that fast). I had mimicked the approach taken in the `Roadmaps.js` component, but that seems to just be one API call, not one call _per_ item listed like this PR. Will look at this =)

wackerow:
image
@samajammin @ryancreatescopy Let me know what you think of this approach. The whole bar is clickable to go to github repo. 

Note, links work but there is a nesting issue still which I will address (which will also address the small issue of the background highlighting going away when hovering over the project image). Also, this has not been refactored to use graphql for api calls, and there may be issues with rate limiting for-the-moment.

wackerow:
image
Quick Demo: https://share.getcloudapp.com/OAuJYZlZ
No more REST api usage for this, all done with graphql queries on the pages each project is listed, and this is passed down to the GitBar component.

ryancreatescopy:
One concern I currently have is whether we're pulling in the languages correctly. If you look at the below screenshots, Vyper.fun isn't labelled Vyper and CryptoZombies isn't labelled Solidity. How can we ensure we pull in the most useful language label?

image

wackerow:
These programs are not necessarily written primarily in the language they're intending to teach (the primary language used in the vyper.fun repo _isn't_ vyper. I don't know a way to automatically pull the language they're intended to teach other than just hard coding that somehow.

wackerow:
@ryancreatescopy I definitely understand your point here though, I was just commenting that this is pulling the primary programming language of the specific git repo... but it does end up changing what we're displaying to the user, and if the goal is to emphasize the languages you can _practice_ by _using_ these apps, I would suggest just hard coding these in (which I'd be happy to do). 

That being said, I feel this changes what this "GitHub" bar represents, and perhaps in the end that's my bad for combining the two issues you posted, as it seems I may have turned "GitHub stars + programming languages you can learn" into "GitHub stars and primary GitHub repo languages", and the whole thing links to GitHub, not the game/app. 

My initial thoughts on this would be to leave the bar roughly as it is (fully representing GitHub repo), and perhaps just add another pill or two ie `[Solidity]` and `[Vyper]` _within the original portion of the ProductCard_, where the link still represents going to the product page. 

Happy to keep fine tuning, lemme know what you think =)

ryancreatescopy:
If we don't show the GitHub stars/button,  we should avoid having so much margin above the title: 
image

wackerow:
Updates:
- [x] No GitHub information on educational materials (which includes everything at /developers/learning-tools/)
- [x] Educational products contain hard-coded array of "subjects" they are designed to teach, placed as colored pills below product description
- [x] Other developer products (eth2 clients / local frameworks) do _not_ have a 'subject array' but do have GitHub information, including API pulls for the top 1 or 2 programming languages used for the repo, and pulling repo star count
- [x] Star count shown in small upper-right widget with GitHub icon and star emoji
- [x] Hover over button links to GitHub and highlights entire button contents including Icon
- [x] GitHub repo languages populate below description as colored pills
- [x] Eth2 Clients restricted to only show primary language
- [x] Title of product has conditional margin to shift if github widget is present
- [x] GitHub widget aligns with edge of bottom button
- [x] Bottom button _replaces_ the link previously placed on the entire product card (prevents nested links)
- [x] On card hover: Shadows reduced, and background highlighting removed
- [x] Fix mobile glitch where subject pills were collapsing and clipping

@samajammin @ryancreatescopy ... I think this is just about everything discussed. I'd say it's ready for review.

wackerow:
One last note:  I know it was commented about the use of 'web3' as a subject (ie. Consensys bootcamp). This can obviously be changed, but my reasoning for keeping it in at the moment is because, yes, this is truly javascript or typescript, but if someone knows javascript and they are looking to learn about Ethereum, they may not be interested in just seeing "JavaScript" since it's not obvious by that label alone that it's something new to them. 
I haven't taken this Concensys course, but my assumption is they are not teaching vanilla javascript, but instead are _teaching web3_ which utilizes JS/TS. 

Options: 
1- Switch it to `JavaScript` and/or `TypeScript`
2- Perhaps change to something like `web3 / JS`
3- Leave as `web3` subject matter and adjust with future PR if needed

src/components/GitStars.js
```diff
@@ -0,0 +1,67 @@
+import React from "react"
+import styled from "styled-components"
+import Emoji from "./Emoji"
+import Icon from "./Icon"
+import Link from "./Link"
+
+const Container = styled(Link)`
+  margin-right: -0.5rem;
```
samajammin, line 8: Generally I'm not a fan of negative margin - why is this needed here? If we want these components to be aligned further right, I suggest we reduct the padding of the card content.

src/components/GitStars.js
```diff
@@ -0,0 +1,67 @@
...
+  background: ${(props) => props.theme.colors.lightBorder};
+`
+
+const Pill = styled.div`
+  text-align: center;
+`
+
+const GlyphPill = styled(Pill)`
+  display: flex;
+  flex-direection: row;
```
samajammin, line 31: Typo here, but you can just remove this line (`flex-direction` defaults to `row`)

src/pages-conditional/developers/local-environment.js
```diff
@@ -605,30 +625,105 @@ export const query = graphql`
...
+        languages(orderBy: { field: SIZE, direction: DESC }, first: 2) {
+          ...devtoolLanguages
+        }
+      }
+    }
     openzeppelin: file(relativePath: { eq: "devtools/openzeppelin.png" }) {
       ...devtoolImage
     }
+    openzeppelinGitHub: github {
+      repository(owner: "OpenZeppelin", name: "openzeppelin-contracts") {
```
samajammin, line 665:
```suggestion
      repository(owner: "OpenZeppelin", name: "openzeppelin-sdk") {
```

src/pages-conditional/developers/local-environment.js
```diff
@@ -605,30 +625,105 @@ export const query = graphql`
         }
       }
     }
+    waffle: file(relativePath: { eq: "devtools/waffle.png" }) {
+      ...devtoolImage
+    }
+    waffleGitHub: github {
+      repository(owner: "EthWorks", name: "waffle") {
```
samajammin, line 632: Nice work pulling these in via GraphQL! FYI we could also save manually tracking the `gitHubUrl`s above by adding `url` key within these queries, e.g.
```
  waffleGitHub: github {
    repository(owner: "EthWorks", name: "waffle") {
      stargazerCount
      languages(orderBy: {field: SIZE, direction: DESC}, first: 2) {
        nodes {
          name
        }
      }
      url
    }
  }
```

src/pages-conditional/developers/local-environment.js
```diff
@@ -274,6 +274,7 @@ const frameworksList = [
     name: "Waffle",
     description:
       "The most advanced testing lib for smart contracts. Use alone or with Scafold-eth or Hardhat.",
+    gitHubUrl: "https://github.com/EthWorks/waffle",
```
samajammin, line 277: Since we're already using the GitHub API to pull in star counts, we could also use the API to pull in GitHub URL. See example code in my comment below.

src/pages-conditional/developers/local-environment.js
```diff
@@ -589,6 +601,14 @@ export const devtoolImage = graphql`
   }
 `
 
+export const devtoolLanguages = graphql`
```
samajammin, line 604: Can we wrap the entire GitHub query as a fragment, e.g. `repoInfo`?

```
export const repoInfo = graphql`
  fragment repoInfo on GitHub_Repository {
    stargazerCount
    languages(orderBy: { field: SIZE, direction: DESC }, first: 2) {
      nodes {
        name
      }
    }
    url
  }
`
```

That would reduce some of the duplication in our query.

samajammin (changes requested):
This looks really solid! Left a handful of comments, then I think this is ready to merge. Awesome work.

I think there's still some polishing we may want to do re: the programming language tags but I think we wait to tweak that later if needed in a separate PR.

@wackerow just a note for future PRs: whenever possible let's try to isolate changes into individual PRs. I can understand including the programming language tags in this PR since those are also fed from the GitHub API but I would've liked to see that separated. I think keeping this PR focused solely on the GitHub stars widget would have keep the conversation more focused & would have allowed this contribution to get merged faster.

Thanks again!

src/pages-conditional/developers/local-environment.js
samajammin: Wrong repo 😄

src/pages-conditional/developers/local-environment.js
wackerow: Thank you for this... this was the original goal but couldn't quite get it to work. This does! Just fixed.

wackerow:
Sorry for delay here, after rebasing the sizing broke. Found what had changed and cleaned up an unnecessary nested div that was causing issues, it now lays out the way it had been, with slight tweaks to align tags after talking with @ryancreatescopy 

image
image
image

src/components/ProductCard.js
```diff
@@ -24,37 +25,33 @@ const Image = styled(Img)`
...
-    box-shadow: 0px 8px 17px rgba(0, 0, 0, 0.15);
-    background: ${(props) => props.theme.colors.tableBackgroundHover};
     transition: transform 0.1s;
     transform: scale(1.02);
   }
 `
 
 const Content = styled.div`
   padding: 1.5rem;
+  padding-right: 1rem;
```
samajammin, line 46:
```suggestion
```
I don't see any reason to have different `padding-right`

wackerow: That was in lieu of a slight negative margin on the github component so it's right-sided alignment is in-line with the right-side of the bottom button

samajammin (approved):
This is great! Thanks for your persistence on this 💪

wackerow:
Hooray! haha no problem =)

PR #1834, opened 2020-11-14
Correction for typo "an" to "can"

Arnor1711:
## Description
Correction for "The beacon chain receives state information from shards and makes it available for other shards so that the network an stay in sync." to "The beacon chain receives state information from shards and makes it available for other shards so that the network can stay in sync."


ryancreatescopy (approved):

ryancreatescopy:
@all-contributors please add @Arnor1711 for bugs and content.

PR #1838, opened 2020-11-15
Make docs links full width on mobile

samajammin:
## Description
Make docs links full width on mobile

**Before:**
Image 2020-11-14 at 4 03 07 PM

**After:**
Image 2020-11-14 at 4 02 06 PM

## Related Issue
None

### DIFF
--- a/src/components/DocsNav.js
+++ b/src/components/DocsNav.js
@@ -5,7 +5,6 @@ import docLinks from "../data/developer-docs-links.yaml"
 import Link from "./Link"
 import Emoji from "./Emoji"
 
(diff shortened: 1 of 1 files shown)


ryancreatescopy (changes requested):
image

I'm not getting full-width on the preview. We may want to set the breakpoint at `m` instead of `s`

samajammin:
@ryancreatescopy nice catch - I've updated the media query to match the container breakpoint.

ryancreatescopy (approved):
LGTM now :+1:

PR #1841, opened 2020-11-15
Adds custom footer breakpoints [Fixes #1828]

### LINKED ISSUE #1828: Wrap footer into columns of 3 when needed

Footer is a bit scrunched for browser widths from L breakpoint (1024px) to ~1300px:

Image 2020-11-13 at 2 52 33 PM

For this width range, let's break out the footer into 3 columns.

Increase the breakpoint where we break it into 2 columns (currently L breakpoint)

wackerow:
##  Description
Standard content breakpoints are inflexible/insufficient for the needs of the footer columns. This updates `Footer.js` to use new breakpoints and collapse columns from 6 down to 1 as the screen sizes change, and should not cause any clipping.

Decided to just put these in the `theme.js` file, but as I'm looking at this I'm not sure how reusable these values would be since they're fairly tuned to the 6-column footer layout. If preferred I could revert that and just hardcode the pixel values into `Footer.js`. 

## Related Issue #1828


src/theme.js
```diff
@@ -277,6 +277,13 @@ const theme = {
     l: "1024px",
     xl: "1440px", // Used as the max-width
   },
+  footerBreakpoints: {
```
samajammin, line 280: Agree `footerBreakpoints` will likely be specific to the Footer component. Let's move these out of theme & keep in that component for now.

samajammin (changes requested):
This looks good! A couple changes to request:

1. Let's move the breakpoint variables into the Footer component
2. Let's adjust to create columns of 4 for the larger breakpoint
It looks awkward to me when only Community & Enterprise sections wrap:
Image 2020-11-15 at 2 34 02 PM
I'd rather see 2 rows of 4 items (4 on top, 3 on bottom).
It appears we DO break into 2 rows of 4 columns at some point but the layout is awkward:
- The section headers should be vertically aligned
- The blank section should be at the end
Image 2020-11-15 at 2 38 08 PM

3. Same deal when we break into 2 rows of 3 columns & into 3 rows of 2 columns the section headers should be aligned:
Image 2020-11-15 at 2 40 01 PM

wackerow:
Will take a look at that and try to find a way to align those better, as it's related to the css-grid. Otherwise I can move the breakpoints object no problem, and just remove the 5-column options.

samajammin:
Thanks! I'm not very familiar w/ CSS grid. Solving with flexbox, one approach might be to conditionally allow flex items to wrap & shrink/grow depending on the breakpoint.

wackerow:
Yeah I may have to rethink this (grid vs flex), I'll see if I can't iron this out soon

wackerow:
This simplifies the entire design by making each "row" only one section tall (instead of squeezing the two short sections, Enterprise and Community, on top of one another). Expanded to 7 columns to account for this on full size desktop windows, where everything will be in one row without clipping. Breakpoints customized to avoid any clipping as the rows collapse down to eventually be one column, and as it collapses the subsequent rows items all align along their top edges. 
Provides motivation to expand Enterprise/Community sections 😉

samajammin (changes requested):
This looks close @wackerow! But not quite there.

We don't want a situation where only 1 items wraps, e.g.
Image 2020-11-17 at 12 05 28 PM
At this point, I'd like to see the footer break into 2 rows of 4 items, like this:
Image 2020-11-17 at 12 06 45 PM

Everything else looks good to me 👍

samajammin:
@wackerow just had a quick offline chat w/ @ryancreatescopy & have an update request (apologies for the moving target here) - let's update the footer to just have 6 sections by bringing the "Community" link under "Ecosystem".

With that, we can update the footer to:
- display 1 row of 6 sections for large screens
- break into 2 rows of 3 sections for medium screens (at a breakpoint you deem appropriate) 
- break into 3 rows of 2 sections for small screens (again at whatever breakpoint you think makes sense)
- 6 rows of 1 sections for mobile / xs screens

Make sense? Let me know if you have any questions! Thanks again.

wackerow:
Check this one out. I found the bug that was causing the crazy malalignment (it was not filtering out shouldDisplay of false). Now the grid plays very nicely. I set breakpoints for 2/3/4/6 columns, and it's lookin nice on my end.

samajammin (approved):

PR #1875, opened 2020-11-17
Updated language around security/audit on product listing policy

GeorgeTrotter:
## Description

Audited - > security-tested, with some examples. *Please* change language as you see fit!


ryancreatescopy (approved):
🥳

PR #1885, opened 2020-11-18
Add tools to "Dapp Tools"

pi0neerpat:
I noticed that the new version of the site doesn't have the dapp tools list anymore (for example: https://github.com/ethereum/ethereum-org-website/blob/385a2334acd2410f1f02c31fced099b4c0fa7710/src/content/translations/de/developers/index.md#entwicklerwerkzeuge-developer-tools).

If this isn't the best place to put this, please let me know. 

Thanks yall!

### DIFF
--- a/src/content/developers/docs/dapps/index.md
+++ b/src/content/developers/docs/dapps/index.md
@@ -74,6 +74,16 @@ Tokens must be generated in order to prove the value nodes that contribute to th
 - [rimble.consensys.design](https://rimble.consensys.design)
 - [GitHub](https://github.com/ConsenSys/rimble-ui)
 
+**One Click Dapp** **_- FOSS tool for generating dapp frontends from an ABI._**
+
+- [oneclickdapp.com](https://oneclickdapp.com)
+- [GitHub](https://github.com/One-Click-Dapp/one-click-dApp)
+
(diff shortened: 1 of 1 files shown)


samajammin (approved):
Hey sorry we must have missed that during the migration - thanks for catching & adding!

pi0neerpat:
Thanks @samajammin. Is there some page that has all the tools together, like the one I linked to above?

samajammin:
> Thanks @samajammin. Is there some page that has all the tools together, like the one I linked to above?

@pi0neerpat not in our latest versions of the website. The overwhelming feedback we got from users was the our old Developers page (which you can see on translations of the site that aren't yet up to date, e.g. https://ethereum.org/de/developers/) was 1) too overwhelming & 2) didn't give enough context on the categories of tools. In our updates, we tried to expand on each area of tooling to provide context for developers in addition to listing the products/projects available (e.g. https://ethereum.org/en/developers/docs/apis/javascript/).

Curious to hear your thoughts - do you think there should be a page that has all the tools together? Perhaps there's a way we could provide both if you think it's useful to have a single page that lists all tools.

pi0neerpat:
Actually I like it this way much better. The organization is good too, on-par or better than Eth Hub.

I might steal this Gatsby layout for my own docs site haha

PR #1888, opened 2020-11-18
Add link to Medalla data challenge results

samajammin:
## Description

Add link to Medalla data challenge results

## Related Issue

None


ryancreatescopy (commented):
In general, let's avoid links that reference things like "here". Can we make the link simply "View the results" — this is more descriptive as the link copy.

ryancreatescopy (approved):

PR #1093, opened 2020-06-05
WIP - #979. TODO: remove repeated logic in components + review

loganbek:

PR #1899, opened 2020-11-20
Add interactive Eth2 diagram

ryancreatescopy:
## Description

Add an interactive diagram to the Eth2 page based on the very helpful: https://twitter.com/trent_vanepps/status/1321936246332104705?s=20

This will help users understand the relationship between upgrades.

Can't get z-index to work


src/components/Eth2Diagram.js
```diff
@@ -96,85 +199,175 @@ const PrePhase2 = styled.div`
...
+const Eth2Diagram = () => {
+  const [modalState, setModalState] = useState(false)
+
+  // Update modal state to upgrade props, else close modal
+  const handleClick = (upgrade) => {
+    const newState = upgrade || false
+    setModalState(newState)
+  }
+
+  const themeContext = useContext(ThemeContext)
```
samajammin, line 306: @ryancreatescopy I don't believe importing `ThemeContext` is needed here if all we're doing is changing colors based on the dark/light theme.

Instead, we should be creating new variables in the `lightColors` & `darkColors` of theme.js so that depending on dark/light mode, the variable displays the desired colors.

This will remove the needs to pass `isDarkTheme` to all the components & the need for any conditional logic within the styled components.

LMK if this doesn't make sense!

src/components/Eth2Diagram.js
```diff
@@ -1,82 +1,185 @@
...
 import ButtonLink from "./ButtonLink"
 import Link from "./Link"
+import { ThemeContext } from "styled-components"
+import Emoji from "./Emoji"
 
 const Eth1 = styled.div`
-  border: 1px solid ${(props) => props.theme.colors.black50};
-  background: ${(props) => props.theme.colors.success200};
+  cursor: pointer;
+  border: 1px solid
```
samajammin, line 11: Just to clarify my other point re: not passing in `isDarkTheme` as a prop... we should create a color variable in theme.js (e.g. `mainnetBorder` that is `success300` for dark mode & `black50` for light mode). That way we avoid the need of passing the props around & this conditional logic based off it.

samajammin (changes requested):
@ryancreatescopy I refactored this a bit to fix the z-index issue.

See my comments regarding the conditional styles based on light/dark mode. Once we address those, I think this is good to go!

samajammin:
FYI @ryancreatescopy we should try to get this merged as it will impact @ph5500's work on #1093.

If we need approval from @jjmstark we can always comment out the component in the eth2/index.js.

samajammin (commented):
@ryancreatescopy merging per Discord discussion.

samajammin (approved):

PR #1907, opened 2020-11-21
Morpher.js ESLintError: React Hook useEffect has missing dependencies [Fixes #1672]

### LINKED ISSUE #1672: ESLintError: React Hook useEffect has missing dependencies
**Describe the bug**
```warning  React Hook useEffect has missing dependencies: 'morphTimeout', 'morpher', 'state.text', and 'state.words'. Either include them or remove the dependency array```

Getting this since my `Morpher.js` update where I included an empty array to the `useEffect` dependencies argument.

**To Reproduce**
1. Run project locally with `yarn start`
2. See console warnings

**Expected behavior**
Morpher works properly without producing a linter warning. 

**Desktop:**
- macOS Big Sur

**Additional context**
Wanted to make note of this, will fix with PR shortly

mnewlin:
… [Fixes #1672]

### DIFF
--- a/src/components/Morpher.js
+++ b/src/components/Morpher.js
@@ -119,6 +119,7 @@ const Morpher = () => {
       if (count < Math.max(slen, rlen)) {
         // Only use a setTimeout if the frameRate is lower than 60FPS
         // Remove the setTimeout if the frameRate is equal to 60FPS
+        if (morphTimeout) clearTimeout(morphTimeout)
         morphTimeout = setTimeout(() => {
           window.requestAnimationFrame(update)
         }, 1000 / frameRate)
@@ -149,9 +150,9 @@ const Morpher = () => {
 
     return () => {
       clearInterval(morphInterval)
-      clearTimeout(morphTimeout)
+      //clearTimeout(morphTimeout)
     }
-  }, [])
+  })
 
   return (
     <NavLink to="/en/languages/">


wackerow (commented):
Hey, thanks for jumping in! Viewing the Netlify preview of this shows the morpher getting stuck on 以太坊 (which is the first item in the array after initially loading with "Ethereum"). This happened to me when I took out the empty array for dependencies previously.  The clearTimeout was intentionally placed in the return of useEffect to clear these only when leaving the page.

mnewlin:
I believe the useEffect is getting triggered everytime the state is updating to the transitionary text and then the clearTimeout was getting triggered before each useEffect trigger so that it was stopping the timeout before it got a chance to update the title text, and so it was never getting to the "end" text and getting stopped at gibberish.

wackerow:
Yeah, the way it's set up there is no need to trigger useEffect more than once. This is why I placed an empty array for dependencies, to explicitly tell it to run once on mounting then never again, and still allows us to unmount timeouts/intervals before leaving page. I have never had any issues using an empty array, but it's causing a console warning here, and it doesn't work properly if you give it dependencies or remove the array entirely.

samajammin (changes requested):
Thanks for looking into this @mnewlin!

As @wackerow mentioned, this gets rid of the ESLintError but it doesn't produce the desired effect of cycling through the array of "Ethereum" translations. Please try to find a solution that satisfies both conditions 😄

mnewlin:
Yeah, I see now it is getting stuck at the one end title now. It was cycling through when I commented out the clearTimeout , but I guess when I added the clearTimeout to the update() function it caused a problem again. I will update and submit again, double checking again, apologies.

mnewlin:
So, interesting enough, adding the [] keeps the useEffect cleanup function from getting called at all, so affectively it is blocking the call to clearTimeout altogether.

mnewlin:
Yes, I see what you mean now, no way I can find to get around it short of disabling the warning for that line but I know we don't want to do that. I will move on to something else for now. Thanks!

samajammin:
Thanks @mnewlin. I'll close this out for now but feel free to re-open when you find a valid solution!

PR #1911, opened 2020-11-21
Change from Mainnet to mainnet

strykerin:


ryancreatescopy (commented):
Both these instances should probably retain mainnet capitalisation.

In the table, it is first in a list so in effect starts the sentence. 
In the article, it adheres to the title case used by the article author.

Thanks for raising the PR but will close it for the above reasons.

PR #1913, opened 2020-11-22
Links to static image assets bypass intl and open in new tab [Closes #1807, Closes #1918]

### LINKED ISSUE #1807: Clicking on a markdown image leads to a 404
**Describe the bug**

**To Reproduce**

1. Go to https://ethereum.org/en/developers/docs/nodes-and-clients/
2. Click on one of the images
3. See 404 "not found" error

**Expected behavior**

- Should we prevent this images from being clickable?
- Should we expand / zoom in on the image?
- I *don't* think clicking on the image should navigate to a standalone page of the image - that doesn't seem useful

**Screenshots**

All

All

**Additional context**

### LINKED ISSUE #1918: Open images in new window at full size

When I click on an image I want the image to open in a new window at full size so that I can view the image properly.

Could open the image in a lightbox/modal but doesn't seem worth it.

**Additional context**

Related:
#1807 #1913

wackerow:
## Description
Images parsed by mdx renderer are being passed through Link to be wrapped in anchor tag, but these links result in 404. These links all involve the directory "static" which as far as I can tell we never want to link to. This PR places a conditional within the `Link.js` component to look for links including "static" and returns `{children}` wrapped in a `span` instead of returning a `<Link>` component (intentionally didn't use `div` as this needs to be compatible within `<p>` tags.) 

This prevents images within EDN from being rendered as links, and are no longer clickable. Remaining link types continue to work appropriately.

## Related Issue #1807

### DIFF
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -69,14 +69,10 @@ const Link = ({
   const isGlossary = to.includes("glossary")
   const isStatic = to.includes("static")
 
-  if (isStatic) {
-    return <span className={className}>{children}</span>
-  }
-
   // Must use <a> tags for anchor links
   // Otherwise <Link> functionality will navigate to homepage
   // See https://github.com/gatsbyjs/gatsby/issues/21909
-  if (isHash) {
+  if (isHash || isStatic) {
     return (
       <a className={className} href={to}>
         {children}
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -67,6 +67,11 @@ const Link = ({
   const isExternal = to.includes("http") || to.includes("mailto:")
   const isHash = isHashLink(to)
   const isGlossary = to.includes("glossary")
+  const isStatic = to.includes("static")
+
+  if (isStatic) {
+    return <span className={className}>{children}</span>
+  }
 
   // Must use <a> tags for anchor links
   // Otherwise <Link> functionality will navigate to homepage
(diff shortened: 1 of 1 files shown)


ryancreatescopy (approved):
Nice job on the fix. I think we should create a new issue to allow users to click on images to open a new window with the full size image. Some of the diagrams in the developers section might included pretty small text that a user may want to enlarge.

wackerow:
@ryancreatescopy Sorry to overwrite directly on this branch after you approved it, but once I noticed it, this was a fairly simple fix along the same lines as the first commit. 

Realized these links were breaking because they was being passed through `intl` which was appending `/en/` to the image links inappropriately. I simply switched to utilize `<a href` for these links with `target="_blank"`, which bypasses prepending the root region folder, and opens the image as you would expect in a new tab, without navigating away from the current page. Working well for me on desktop and mobile. 

[(Netlify preview example)](https://deploy-preview-1913--ethereumorg.netlify.app/en/developers/tutorials/getting-started-with-ethereum-development-using-alchemy/)

Hopefully this is what you were looking for and this PR can handle both issues at once without any other delays, otherwise I do apologize and I'll just revert these changes.

ryancreatescopy (approved):
This looks to do the job!

samajammin (approved):
Excellent! Nice fix 🏆

PR #1915, opened 2020-11-22
Adds Tutorial card to dev home [Fixes #1781]

### LINKED ISSUE #1781: Create tutorials card on developer homepage

Currently there's no card that links users to the tutorials section on our [developer homepage](https://ethereum.org/en/developers/), yet there are cards for docs and the other pages.

We should add a card to better surface tutorials and keep that page consistent with the others. This should not break the design of the page and should also include removing the callout banner that is currently used to advertise the tutorials.

wackerow:
## Description
Adds tutorials object to `paths` array in `/developers/index.js`. This creates a fourth card, so for symmetry, adjusted `StyledCard` styling to fit 4 cards until the breakpoint of 1120px, when it collapses to 2 (and eventually to 1). (This breakpoint falls between themed breakpoints, but given the min-width placed on this card, when screen is between `breakpoints.l` (1024px) and 1120px this caused it to layout 3 cards in the first row, and 1 full width below it.)

Also removes `CalloutBanner` from bottom of dev home.

## Related Issue #1781


ryancreatescopy (approved):
Looks good to me!

ryancreatescopy (approved):

samajammin (approved):
Thanks!

PR #1917, opened 2020-11-22
Fixes links within mdx h2/h3, adjusts `code` css [Fixes #1843]

### LINKED ISSUE #1843: Tutorial template h2s/h3s
**Describe the bug**

If you put a link or a `code`  in a h2/h3 title it doesn't display properly.

**To Reproduce**

1. Create a new tutorial in your local
2. Add `## Example h2 with [a link](https://google.com)` or `## Example h2 with `code` `
3. Preview in your local 

**Expected behavior**

Links should work/display in titles.
Code should inherit font size from header.

**Screenshots**

image
image

wackerow:
## Description
- This adds a css selector to the `Header2` and `Header3` components within `SharedStyledComponents.js`, selecting for `<code>` children, and adjusts height to be relative to its parent.
- Changed the `a` sub-styling on these two components to `a.header-anchor` to select _only_ for the link being generator when parsing mdx (the chain-link to the left of the header). This leaves intentionally placed links within a header to render normally.
- Note this does not actually create nested anchor tags (as I previously expressed concern for), because the actual header text is not a link (only the side chain-link icon), and the ToC only strips the text, and does not apply a second link or additional styling (like `code` styling for example)

## Related Issue #1843

image

### DIFF
--- a/src/components/SharedStyledComponents.js
+++ b/src/components/SharedStyledComponents.js
@@ -317,7 +317,7 @@ export const Header2 = styled.h2`
   }
 
   /* Anchor tag styles */
-  a {
+  a.header-anchor {
     position: relative;
     display: none;
     margin-left: -1.5em;
@@ -335,7 +335,7 @@ export const Header2 = styled.h2`
   }
 
   &:hover {
-    a {
+    a.header-anchor {
       display: initial;
       fill: ${(props) => props.theme.colors.primary};
     }
--- a/src/components/SharedStyledComponents.js
+++ b/src/components/SharedStyledComponents.js
@@ -330,6 +330,10 @@ export const Header2 = styled.h2`
     }
   }
 
+  code {
+    font-size: 0.8em;
+  }
+
   &:hover {
     a {
       display: initial;
@@ -370,6 +374,10 @@ export const Header3 = styled.h3`
       display: initial;
       fill: ${(props) => props.theme.colors.primary};
     }
+  }
+
+  code {
+    font-size: 0.85em;
   }
 
   &:hover {
--- a/src/components/SharedStyledComponents.js
+++ b/src/components/SharedStyledComponents.js
@@ -317,6 +317,7 @@ export const Header2 = styled.h2`
   }
 
   /* Anchor tag styles */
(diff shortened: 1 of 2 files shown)


wackerow:
image
image

Noticed a very slight alignment issue when rendering these `<code>` snippets, which seems to render some of them perfectly in line, and others nudged down (and seems to happen at any screen size).

Let me know if anyone has thoughts on why this is rendering differently on two different places on the screen at the same time.

@ryancreatescopy Btw, I chose the `em` units to be a little less than one because using 1 made this text look almost bolded and the characters became slightly taller than surrounding text (different font-family I assume)... Let me know what you think, or if you have a different thought on this entirely.

ryancreatescopy:
To me, `1em` looks better. It doesn't seem to bolden it at all. I think removing the padding from it might be useful too.
image


When it's in the middle of a sentence though, we get this weird space after the code which goes up to the following word. Do you think we should remove that?

wackerow:
@ryancreatescopy Ah... yeah ideally, hah... did most in dark mode which is a little harder to see the background.. I'll take a look, and also switch back to 1em

wackerow:
Previous commit adjusted this styling within "SharedStyledComponents", targeting H2/H3... but after reviewing, it seems more appropriate to blanket this across all code snippets from the `layout.css` file, since all code snippets are currently forced to `0.85rem`, despite surrounding context, and gap glitch following code snippets occurs everywhere, even if outside a header.

- Reverted `SharedStyledComponents` changes that related to the code snippets, and placed this logic in `layout.css` by adjusting `font-size: 0.85rem` to `font-size: 1em`. This will allow code snippets to match the font size of its surrounding text, instead of referencing the 'root em'.
- Trimmed vertical padding on code snippets by 25% (now `0.15em`)
- Removed `:after` styling to remove gap being placed after code snippets

### Screenshots as of a47177a:
image
image

ryancreatescopy:
This looks good to me – looks much better in the UI! I defer to @samajammin on the technical implementation.

ryancreatescopy (approved):

samajammin (commented):
LGTM! Nice work!

PR #1920, opened 2020-11-23
Adds gaming definition

ryancreatescopy:
## Description

Gaming copy was using the collectibles copy in error. This fix adds an intro to decentralized gaming.
Also, made the options inactive/active states clearer.


wackerow:
@ryancreatescopy Note #1940 adds language support for the dapps page... I noticed the gaming definition was a duplicate and put in my own filler, but I just replaced it with the text you wrote here. 

Just wanted to mention because right now there will be a conflict at line 1071, but I see you have a couple others changes in here too (that I did not mess with). Just be aware that my PR does fix the 'gaming definition' issue.

samajammin (approved):

Issue #1928, opened 2020-11-24
Card misalignment mobile


ryancreatescopy:
**Describe the bug**

Stacked cards on mobile have margin that should only be on desktop. This is causing some dodgy misalignment. 

This may be present elsewhere but I noticed it on the beacon chain page. 

**Expected behavior**

Cards should align and have the same margin on mobile. 

**Screenshots**

![image](https://user-images.githubusercontent.com/40891631/100063013-6996eb80-2e28-11eb-98e9-a6aaf2c2614c.png)

samajammin:
Solved in #1931

Issue #1927, opened 2020-11-24
Update eth2 copy that mentions dependency on reaching staking threshold


ryancreatescopy:
**Is your feature request related to a problem? Please describe.**

Some copy in the Eth2 section explains that the beacon chain will launch if the deposit contract got enough deposits.

We reached the threshold so this is no longer an "if"!

**Describe the solution you'd like**

Update the copy based on the new status.

samajammin:
Solved in #1931

PR #1931, opened 2020-11-24
Eth2 updates and InfoBanner update [#1928, #1927]

ryancreatescopy:
## Description

- Added new Eth2 articles
- Removed mention of beacon chain needing x deposits and replaced it with copy saying that we've now got enough deposits to launch
- Refactored InfoBanner which was displaying weirdly on mobile – aligned it with Warning design. Had to tweak a few instances of it to make sure it displayed correctly.
- Fixed dodgy alignment of cards on beacon chain page.
- While working on this PR I discovered I'd used the wrong string for the dapps page title in the nav. Have fixed that too... this caused the selectable tag on the find-wallet page to read incorrectly.

## Related Issues

#1928 
#1927


src/pages-conditional/wallets/find-wallet.js
```diff
@@ -87,6 +86,14 @@ const Header = styled.header`
   }
 `
 
+const InfoBannerContainer = styled.div`
```
samajammin, line 89: Why not just style `<InfoBanner/>` directly vs. creating this container?

src/pages-conditional/wallets/find-wallet.js
```diff
@@ -87,6 +86,14 @@ const Header = styled.header`
...
+  @media (max-width: ${(props) => props.theme.breakpoints.l}) {
+    margin-left: 2rem;
+    margin-right: 2rem;
+    margin-top: -2rem;
```
samajammin, line 93: Why negative margin? I suspect there's a better way to solve this, e.g. by reducing margin-top of the element above.

src/pages/get-eth.js
```diff
@@ -331,14 +331,15 @@ const GetETHPage = ({ data }) => {
               <Translation id="page-get-eth-." />
             </em>
           </p>
+          <br />
```
samajammin, line 334: Perhaps we add margin-top vs. a `<br/>`? They can act a bit unpredictably.

samajammin (changes requested):
Looks good for the most part but a couple questions/comments. Happy to pair on this tomorrow.

src/pages-conditional/wallets/find-wallet.js
ryancreatescopy: It makes the banner go full width as it's not in a parent container with any padding/margin set.

src/pages-conditional/wallets/find-wallet.js
ryancreatescopy: Changed margins of the container above.

src/pages/get-eth.js
ryancreatescopy: done.

samajammin (approved):

PR #1940, opened 2020-11-24
Dapps page set up to support multiple languages [Fixes #1936]

### LINKED ISSUE #1936: Set up Dapps component pages for translations
We recently updated most of our Dapps page from markdown files to a fully featured React component:
https://github.com/ethereum/ethereum-org-website/blob/dev/src/pages-conditional/dapps.js

We need to set this content up for translation. 

1.  Move the English copy into [/intl/en.json](https://github.com/ethereum/ethereum-org-website/blob/dev/src/intl/en.json) - see [our homepage](https://github.com/ethereum/ethereum-org-website/blob/dev/src/pages/index.js) for an example of how this is done.
2. Replace the English copy with our `<Translation/>` component and/or the `intl.formatMessage()` function (again, see the homepage for an example of how this is done)

None

**Additional context**
- We the use [`gastby-plugin-intl`](https://github.com/wiziple/gatsby-plugin-intl) for our internationalization. That plugin's documentation may provide more context for how it works.
- Let's not worry about translating brand/project names

wackerow:
## Description
Converts new dapps page to be compatible for translations utilizing `<Translation id="" />` and `intl.formatMessage`, and extracting english into `en.json`

## Related Issue #1936


samajammin (changes requested):
This looks awesome @wackerow! But I'm not loving all the duplicate `defaultMessage` declarations - I think we could create a utility function to make this a bit less verbose.

Would you be up for attempting a refactor here? I'm thinking we add an exportable function to src/utils/translations.js, e.g. `translateMessageId`. This function would receive to params, `intl` (the `intl` object from `gatsby-plugin-intl`) & `id` (the message ID).

That way instead of writing out e.g.
```
intl.formatMessage({
        id: "page-dapps-features-1-title",
        defaultMessage: getDefaultMessage("page-dapps-features-1-title"),
      })
```
We could just have
```
translateMessageId("page-dapps-features-1-title", intl)
```

This function could look something like:
```
export const translateMessageId = (id, intl) => {
  // some logic to check if `intl` is valid (has a formatMessage function)
  // && verify `id` is a kebab-case string. If not, throw error.
  intl.formatMessage({
    id,
    defaultMessage: getDefaultMessage(id),
  })
}
```
Let me know if that makes sense or not! Basically, I'm envisioning a wrapper function on intl.formatMessage to always return a default message.

wackerow:
@samajammin Refactored with the changes you mentioned, except an explicit kebob check. I had tried requiring a match with `/[\w-]+/` which worked for this specific page, but if this were to be applied to some IDs (ie ` "page-get-eth-DEX's-desc-3"`) it would not work because of the apostrophe. Instead this checks if the message returned by `getDefaultMessage` is the same as the ID (which is currently what is returned and rendered if there is no match in the language JSON for the ID passed) in which case it logs a `console.error` and returns an empty string.

This is an example console log of when an ID is misspelled:
image
Should make it pretty clear to anyone debugging where the problem is, and this seems to make a specific kebob check redundant. (If you still want it, let me know, but if we're going to add restrictions, we should first make sure we don't break our existing references if we chose to use this method elsewhere, and iron out the specific formatting rules we want)

Lemme know what you think!

samajammin (approved):
Bravo! Thanks for making that refactor 😄

PR #1968, opened 2020-11-27
Trilemma mobile [Fixes #1932]

### LINKED ISSUE #1932: Add trilemma to mobile

On the Eth2 vision page we have an interactive trilemma, so folks can explore the challenges of decentralized scaling: https://ethereum.org/en/eth2/vision/

Currently this isn't styled for mobile, so doesn't show on phones. We'd like to create a mobile implementation of this trilemma.

I think modals might be the best way to go as per this quick prototype: https://www.figma.com/proto/UP36ufDF6PaVPwi7Eqzzge/Mobile-trilemma?node-id=1%3A13&viewport=428%2C400%2C0.6768559217453003&scaling=min-zoom

On mobile having the information in a separate card above or below the SVG is a poor experience as the user must continuously scroll up and down to interact with the trilemma and read the info. It's also not clear that interactions do anything. Hence, wondering whether a modal would be a better bet. 

- Text should still be legible with a 16px minimum font size

**Screenshot**
Desktop implementation: 

image

wackerow:
## Description
Adds mobile compatibility for "Trilemma" section of Eth2 Vision. 
- Removes conditional from eth2 vision page that was hiding trilemma on small screens
- Refactored using @media CSS queries instead of useEffect for screen sizing
- Utilizes animated modal slide-up card when trilemma circles are tapped from screen size < `breakpoints.l` (when format switches to `flex-direction: column`.
- Automatically hides the desktop `ExplanationCard` when below this width breakpoint.
- Automatically switches between the two types of explanation cards responsively if screen size changes
- Modal background or `X` button clickable to leave slide-up card
- Card height depends on content size
- Triangle width set to 100% on mobile to allow shrinking and prevent clipping.

Note: The `text` is within the `svg`, and when it scales down the text does as well, despite their font-size declaration. So as a fix:
- Font size increased conditionally on small screens (under `breakpoints.s`, which caused some clipping on the right side)
- Shifted text left on screens under `breakpoints.s` to prevent clipping.

This should keep the font size reasonable, and without clipping, through all range of screen sizes.

Also as the SVG scaled down on smaller screens it resulted in a large vertical margin. Wasn't sure the best way to eliminate, and just used a negative vertical margin in the meantime. Let me know if you have any suggestions for this.

Demo: https://share.getcloudapp.com/xQuYZj6P
Netlify link: [Eth2 Vision](https://deploy-preview-1968--ethereumorg.netlify.app/en/eth2/vision/)
## Related Issue
#1932


AML-MAR5-MDH-sudo (approved):

samajammin (approved):
LGTM 🤩 - @ryancreatescopy please take a look.

wackerow:
Oops, sorry to "dismiss" the review... just cleared up the conflicts, so the new Translation changes will be included properly.

ryancreatescopy:
@wackerow this looks awesome. Just one thing: can we double check the copy above the trilemma. On my screen it's appearing like it's got too much margin left and right:
image

ryancreatescopy (changes requested):
See comment

ryancreatescopy (approved):
So so happy to see this on mobile! Thanks @wackerow

Issue #1866, opened 2020-11-17
Extra sidebar element on page load


samajammin:
**Describe the bug**

If I open up a docs page from a link (e.g. https://ethereum.org/en/developers/docs/layer-2-scaling/), I'm seeing a strange extra `Layout__MainContent` element.

**To Reproduce**

1. Go to https://ethereum.org/en/developers/docs/layer-2-scaling/
2. See strange white element

**Expected behavior**

**Screenshots**

![Image 2020-11-16 at 4 58 51 PM](https://user-images.githubusercontent.com/8097623/99326627-461ee000-282d-11eb-81b1-a0658aa9efb8.png)

![Image 2020-11-16 at 4 58 44 PM](https://user-images.githubusercontent.com/8097623/99326584-4323ef80-282d-11eb-852e-1e6dd5ee32a9.png)

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: Brave

wackerow:
(PR #1973 may fix this as well when it's ready... looking at it closer now)

wackerow:
Reiterating a tl;dr form of the PR #1973 thread comments that relate to this issue:
- Cached versions of sites seem to be affecting certain DOM elements in strange ways
- (The above white space is an improperly styled "SideNavMobile" component)
- Clearing browser cache seems to fix this, but remains an issue for the end user who should not have to do that

(this was _not fixed_ by PR #1973, as suggested in my last comment)

samajammin:
Thanks @wackerow, so are you still seeing this issue / have you been able to replicate it?

wackerow:
@samajammin No, not after clearing cache, everything has been looking good... just makes me wonder if any end users are experiencing anything similar, but not sure there is much more to add to this for now.

samajammin:
Seems like these might have something to do with toggling from light to dark mode 🤔 :
https://share.getcloudapp.com/Jruq6R8E

wackerow:
Hm.. I had toggling dark mode fix the issue once. Not able to replicate this at the moment.

PR #1973, opened 2020-11-29
Layout refactor using `sticky` [Closes #1848, Closes #1867]

### LINKED ISSUE #1848: Docs incomplete mobile banner bug
**Describe the bug**

The incomplete banner displays underneath the mobile table of contents for the docs 

**To Reproduce**

1. Go to https://ethereum.org/en/developers/docs/layer-2-scaling/ on mobile 
2. See banner 

**Expected behavior**

You should see the full banner sitting underneath the table of contents.

**Screenshots**

image

### LINKED ISSUE #1867: Layout margin issue when loading directly on a developers page
**Describe the bug**

When I land directly on a developer page (or hard refresh on the page), e.g. https://ethereum.org/en/developers/learning-tools/ our top margin is not set correctly.

**To Reproduce**

1. Go to https://ethereum.org/en/developers/learning-tools/
2. See margin-top issue above h1

**Expected behavior**

`Layout__MainContainer` element margin-top should be 134px (`props.theme.variables.navSubNavHeightDesktop`), not 4.75rem (`props.theme.variables.navHeight`)
https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Layout.js#L37

**Screenshots**

Landing on homepage, then navigating to dev page (looks good):
Image 2020-11-16 at 5 02 08 PM

Landing directly on dev page (margin looks off):
Image 2020-11-16 at 5 01 54 PM

- OS: MacOS
- Browser: Brave

**Additional context**

I suspect this has to do how we conditionally set margin-top in our Layout.js component:
https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Layout.js#L37
My hunch is that on initial page load, that prop is not getting properly set.

wackerow:
## Description
This refactor changes the top `Nav` component (which includes EDN navigation) and `SideNavMobile` components to be positioned `sticky` instead of `fixed`. This should fix issues with banners not being aligned properly, as they were not positioned relative to their fixed siblings. 

- `Nav/index.js` and `SideNavMobile.js` both changed to `position: sticky`
- Moved `<SideNavMobile>` outside of `<MainContainer>`
- Removed conditional `margin-top` styling to `MainContainer`, as this will now sit relative to nav components naturally
- Removed now unused `shouldShowSubNav` constant (and its dependencies)

Also, while in the `Layout.js` file, there are two deprecated methods being used (`.addListener(() => {})` and `.removeListener(() => {})`... these are now `.addEventListener("change", () => {})` and `.removeEventListener("change", () => {})`). I was going to update these, however, dark mode state is being loaded in `useEffect` already, and changed via `handleThemeChange()` directly which re-renders the view. I removed these calls, and the dark mode seems to work perfectly fine without it.

⚠️  Given that this is a layout adjustment that potentially influences every single page, any help testing this would be appreciated. Locally it is working exactly like the live site on dev pages, EDN, ETH2, and all the MDX pages (except with no banner issues on reloads or direct linking). 
Please let me know if you spot something I missed... (Note, this issue has been challenging to reproduce in dev, at least for me, but this should hopefully fix the banner issue, will test the Netlify link when this saves)


wackerow:
So this wasn't intended to address/involve issue #1866, but that issue now appears to be affecting this Netlify build. This is getting very confusing for forgive me if I post some screenshots to demonstrate and also think through what is going on here.

## Example of issue, using `/en/developers/docs`
### This PR branch: Netlify vs Localhost
This is a comparison of the netlify build for this PR (on the left) and the same page running on localhost (on the right):
image
Notice that highlighted elements _should_ match, but don't. In this case, the Netlify version has wrapped the `SideNavMobile` element inside a `MainContainer` element, which **completely** changes the way `SideNavMobile` is styled and renders (and it may not render at all depending on that styling).
### ethereum.org vs localhost dev-branch
Similarly, this compares the live page on the left to my local `dev` on the right:
image
`Layout_MainContent` is, similar to above, wrapping the intended element `SideNavMobile` when it shouldn't, causing this:
image
This giant space is actually the `SideNavMobile` component completely re-styled after being inadvertently wrapped in a `BannerNotification` which is wrapped in `MainContent`. You can see the term "README" nested inside this giant white space:
image
### Lastly, Netlify of this PR: From homepage vs reload on `/docs/` page
Navigating to docs _from homepage_:
image
After refresh:
image

@samajammin @ryancreatescopy Sorry for all the screenshots, but testing this has been a challenge and needed to try and organize what's happening here. Have either of you seen anything like this before where a component for some reason wraps its sibling? Especially with these discrepancies between environments, browsers, devices, and how you got to the page? Let me know if you have any thoughts, but switched this to draft-mode in the meantime.

On the plus side, the initial changes this proposed seem to be working nicely to fix the top alignment issues we've been having (when this other bug isn't getting in the way causing side-effects).

wackerow:
@samajammin Yes, actually... as I was checking that I noticed similarly it worked in a private browser window, and also seemed to have an instance that corrected after toggling light/dark mode.  This actually fixed all of it for me (at least in the meantime) for both netlify builds and the live site.

src/components/Layout.js
```diff
@@ -71,22 +63,6 @@ const Layout = (props) => {
     } else {
       setIsDarkTheme(window.matchMedia("(prefers-color-scheme: dark)").matches)
     }
-    window
```
samajammin, line 74: Agree that these can be removed.

They DO add functionality (they listen to if the browser preferences are changed) but I think that's such a rare edge case to adjust for that we should just remove this.

samajammin (approved):
This looks great! Nice simplification of the layout. Thanks for doing this ❤️

samajammin:
Great work! I'm comfortable merging this as I'm confident these changes are unrelated to #1866.

PR #1979, opened 2020-11-30
add 0.000000001 in scientific notation as well

MihirLuthra:
I am new to Ethereum right now. While reading Gas documentation, I felt it would be easier for new users to remember Gwei-ETH relation if it was specified in scientific notation as well. So, I just added a 10<sup>-9</sup> in brackets.


samajammin (approved):
Thanks!

samajammin:
@all-contributors please add @MihirLuthra for content.

PR #1988, opened 2020-11-30
Updating Solidity links to soliditylang.org domain

franzihei:
We moved most Solidity related resources (namely the blog and the documentation) to the soliditylang.org domain, so I'm updating all domains affected by it in the Smart Contract Languages page.


ryancreatescopy (approved):
awesome!

PR #1989, opened 2020-11-30
changed from isn't to is as that makes more sense.

tentodev:
I was reading through the Ethereum documentation and came across this sentence and felt that this is a mistake. I am not a blockchain expert so if I am wrong, an explanation would be appreciated.


ryancreatescopy:
Hi @tentodev `isn't` is actually correct. I've updated your PR to try and make it clearer that if Validator X isn't chosen to propose a new block, they must attest to validator Y's proposal.

sudo-vaibhav:
this makes a lot more sense.

ryancreatescopy (approved):

ryancreatescopy:
@all-contributors please add @tentodev for content and bugs.

PR #1958, opened 2020-11-26
Adds className prop to DocLink component [Fixes #1869]

wackerow:
## Description
Adds the `className` prop to `DocLink` custom component to allow styling with styled-components. 
(Also a small clean up with `<Arrow />`)

## Related Issue
#1869

PR #1998, opened 2020-11-30
Adds className prop to CardList component

wackerow:
## Description
Follow-up to #1958, only other component I found that was being styled with `styled-components` without the `className` prop being passed. This adds it to the component. 

## Related Issue (none new filed)
https://github.com/ethereum/ethereum-org-website/blob/689d9d6ecc462a27d79d94eecbabd44548632a93/src/pages/eth2/get-involved/index.js#L132

### DIFF
--- a/src/components/CardList.js
+++ b/src/components/CardList.js
@@ -56,9 +56,9 @@ const Image = styled(Img)`
   margin-top: 4px;
 `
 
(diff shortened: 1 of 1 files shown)


samajammin (approved):
Thanks for the catch & fix!

PR #2002, opened 2020-11-30
Removes trailing underlined space on internal links

wackerow:
## Description
Logic was adding a space to the end of internal links, causing it to be underlined. This was put in place to give a buffer to the new trailing glossary icon, so as an alternative I slightly adjusted the margin on this icon and removed the forced `{" "}`. 

Side note: This does _not_ fix the issue of this icon wrapping without the last word.

## Related Issue (none filed):
image


samajammin (approved):
Another solid catch & fix. Cheers!

Issue #2006, opened 2020-12-01
Set up remaining components & pages for translations


samajammin:
**Is your feature request related to a problem? Please describe.**

Catch-all issue to set up remaining components & pages for translations. Feel free to knock any/each component in individual PRs!

**Describe the solution you'd like**

Refactor components:
- [ ] AssetDownload
- [ ] BugBountyPoints
- [ ] CallToContribute
- [ ] Codeblock
- [ ] DocsNav
- [ ] EthExchanges
- [ ] FileContributors
- [ ] ProductList
- [ ] SectionNav
- [ ] TranslationsInProgress
- [ ] TutorialMetadata

Refactor page components:
- [ ] languages.js

Refactor template components:
- [ ] docs.js

For each component:
1.  Move the English copy into [/intl/en.json](https://github.com/ethereum/ethereum-org-website/blob/dev/src/intl/en.json)
2. Replace the English copy with our `<Translation/>` component and/or the `translateMessageId` function

**Describe alternatives you've considered**

**Additional context**

Similar work to #1937 & #1936 - see associated PRs for solution.

samajammin:
@ryancreatescopy @wackerow @ph5500 could definitely use your help with these if you have any capacity over the next few days 😄 - we're trying to finish these & upload our new content files for translations by the end of this week 😅 

Feel free to ping me here / message me in Discord if you have any questions! Happy to hop on a call & walk through how to do this.

wackerow:
I will start working on these from the bottom of the list upward

samajammin:
@ryancreatescopy I'm curious to hear your thoughts re: the TableOfContent.js component & docs.js template.

Given the current translation process (English copy --> Crowdin translations --> Github files), we only show the "Edit this page" button link to Github for English pages. I suppose we should do the same for the "This page is incomplete" banner (i.e. hide it unless it's an English page)? Or should we reconsider this flow?

ph5500:
I can start working from the top of the list down if that works.

wackerow:
@ph5500 Not sure where you're at, if you're working on this now, but the bottom ones didn't take terribly long and I'm currently on `EthExchanges` which is slightly more involved, so I'll stop after this one.

samajammin:
Amazing work @wackerow @ph5500 💪 - love the quick turnaround! Our global visitors will soon thank you 😄

PR #2008, opened 2020-12-01
Expands translation support

wackerow:
## Description
Refactored the following for translation support

- [x] EthExchanges
- [x] FileContributors
- [x] ProductList
- [x] SectionNav
- [x] TranslationsInProgress
- [x] TutorialMetadata
- [x] languages.js
- [x] docs.js

## Related Issue #2006


src/components/EthExchanges.js
```diff
@@ -439,11 +441,15 @@ const EthExchanges = () => {
         // Add state exceptions if Country is USA
         let description = null
         if (
-          state.selectedCountry.country === "United States of America (USA)"
+          state.selectedCountry.country ===
```
samajammin, line 444: Note: this actually doesn't need translation as it'd not a user-facing string 😄

src/components/EthExchanges.js
```diff
@@ -473,11 +479,15 @@ const EthExchanges = () => {
             // Add state exceptions if Country is USA
             let description = null
             if (
-              state.selectedCountry.country === "United States of America (USA)"
+              state.selectedCountry.country ===
```
samajammin, line 482: Same here.

src/intl/en.json
```diff
@@ -129,6 +129,29 @@
   "nav-use-aria-label": "Use Ethereum menu",
   "nav-learn-aria-label": "Learn menu",
   "on-this-page": "On this page",
+  "comp-tutorial-metadata-tip-author": "TIP AUTHOR",
```
samajammin, line 132: We should probably format this text to uppercase vs. storing here as uppercase.

samajammin (approved):
AMAZING! Thanks for this & for the quick turnaround ❤️ 🏆

src/intl/en.json
wackerow: Yeah not a bad thought... I bet there are some others in there that are duplicated only to change the case that could potentially be styled to reduce redundancies. Let me know, I could post it as an issue if you want and take a look

PR #2015, opened 2020-12-01
New eth2 terms for the glossary

ryancreatescopy:
## Description

With the launch of the genesis block🚀, thought we should add some technical glossary terms for the eth2 stuff as they get used a lot!

### DIFF
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -34,9 +34,19 @@ In [Solidity](#solidity), `assert(false)` compiles to `0xfe`, an invalid opcode,
 
 <DocLink to="/developers/docs/security/" title="Security" />
 
+### attestation {#attestation}
+
+A validator vote for a [Beacon Chain](#beacon-chain) or [shard](#shard) [block](#block). Validators must attest to blocks, signalling that they agree with the state proposed by the block.
+
 <Divider />
 
 ## B {#section-b}
+
+### Beacon Chain {#beacon-chain}
+
+An Eth2 upgrade that will become the coordinator for the Ethereum network. It introduces [proof-of-stake](#proof-of-stake) and [validators](#validator) to Ethereum. It will eventually be merged with [mainnet](#mainnet).
+
+<DocLink to="/eth2/beacon-chain/" title="Beacon Chain" />
 
 ### big-endian {#big-endian}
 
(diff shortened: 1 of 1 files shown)


ryancreatescopy:
@djrtwo would be great if you could do a quick accuracy check on this for me! Also, if there are any other common terms you think we could add, that'd be awesome.

src/content/glossary/index.md
```diff
@@ -34,10 +34,20 @@ In [Solidity](#solidity), `assert(false)` compiles to `0xfe`, an invalid opcode,
 
 <DocLink to="/developers/docs/security/" title="Security" />
 
+### attestation {#attestation}
+
+A validator vote for a [Beacon Chain](#beacon-chain) or [shard](#shard) [block](#block). Validators must attest to blocks, signalling that they agree with the state proposed by the block.
```
wackerow, line 39: If I am correct, the spelling of "signalling" vs "signaling" is regional. We only have "en" without any differentiation for British vs American, but correct me if I'm wrong, words like "organization" vs "organisation" on the site tend to be the American spelling?
For consistency would suggest changing to `signaling`

src/content/glossary/index.md
```diff
@@ -144,6 +164,12 @@ A short string of data a user produces for a document using a [private key](#pri
 
 A cryptographic algorithm used by Ethereum to ensure that funds can only be spent by their owners.
 
+### epoch {#epoch}
+
+A period of 32 [slots](#slot) (6.4 minutes) in the [Beacon Chain](#beacon-chain)-coordinated system. [Validator](#validator) [committees](#committee) are shuffled every epoch for security reasons. There's an opportunity at each epoch for the chain to be [finalised](#finality).
```
wackerow, line 169: Same as above here with "finalised"... would suggest changing to `finalized`

wackerow (commented):
👍

src/content/glossary/index.md
samajammin: That's a good point, there's quite a mix of English (British vs. American) across the site. Probably makes sense to pick one to go with.

samajammin: Not sure what the best way to decide, e.g. if [Google Trends data](https://trends.google.com/trends/explore?q=organization,organisation) is an accurate signal of understanding/demand:
Image 2020-12-01 at 9 17 24 AM

samajammin (dismissed):
LGTM

ryancreatescopy:
It pained me to do it, but changed to US English. We also go the. thumbs up from Lakshman so let's merge after review.

samajammin (approved):

PR #2019, opened 2020-12-01
Made all tutorial tags lowercase

ryancreatescopy:
## Description

All tutorial tags should be added as lowercase to prevent duplicates in the tag list.


samajammin (approved):
Thanks!

PR #2034, opened 2020-12-03
Sanitizes tutorial tags before grouping [Fixes #2020]

### LINKED ISSUE #2020: Tutorial tags shouldn't duplicate (even with different casing)

If I add a tutorial and I tag it "JavaScript" it shouldn't generate a new tag alongside "javascript". Instead our logic should recognise that they are the same and group it under the one tag which displays on the frontend as `JAVASCRIPT` anyway.

**Additional context**

This is mainly for the tutorials in the developers section.

wackerow:
## Description
Reduces `allTags` down to array of unique tags after sanitizing them with `.toLowerCase().trim()`. This array is then sorted by name, since the previously capitalized tags are at the top from the returned query data.

Now for example: If you list "JavaScript " when making a new tutorial markdown, it will lump with "javascript", sum up the totals for each, and display "JAVASCRIPT" with the proper sum on the UI. Clicking this will properly filter for all of them, as all tags are forced to lowercase and trimmed before being stored in the `allTutorials` constant.

Kept styling logic separate, leaving the CSS to style these as uppercase.

## Related Issue #2020

### DIFF
--- a/src/pages-conditional/developers/tutorials.js
+++ b/src/pages-conditional/developers/tutorials.js
@@ -192,7 +192,7 @@ const TutorialsPage = ({ data }) => {
       title: tutorial.frontmatter.title,
       description: tutorial.frontmatter.description,
       author: tutorial.frontmatter.author,
-      tags: tutorial.frontmatter.tags,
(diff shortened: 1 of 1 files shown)


samajammin (approved):
Nicely done!

PR #1938, opened 2020-11-24
Stablecoins page [WIP #1934]

ryancreatescopy:
## Description

New stablecoins page – feedback welcome.

## Related Issue

#1934

PR #2049, opened 2020-12-03
Sets up `alt` text on images for translation [Fixes #2044]

### LINKED ISSUE #2044: Set up image alt texts for translation

Similar to #2006 - we should update the `alt` text across our images to source the strings from en.json.

Use our `translateMessageId` function to refactor all `alt` attributes.

**Additional context**

Background info on alt text:
https://moz.com/learn/seo/alt-text

wackerow:
## Description
- Similar to previous translation-related PRs, extracted any English 'alt' fields into `en.json`, and utilized `translateMessageId(id, intl)` in their place
- Did not set up product names / brands for translation

Let me know if anyone finds any that I missed

## Related Issue #2044


src/components/CalloutBanner.js
```diff
@@ -71,11 +73,12 @@ const CalloutBanner = ({
   children,
   className,
 }) => {
+  const intl = useIntl()
   return (
     <StyledCard className={className}>
       <Image
         fluid={image}
-        alt={`${title} image`}
+        alt={`${title} ${translateMessageId("image", intl)}`}
```
ryancreatescopy, line 81: I think this is a candidate for moving the alt text to the page component rather than in `CalloutBanner.js`

`{title} image` is likely to be a poor alt tag and may not translate well.

src/components/ProductCard.js
```diff
@@ -121,10 +123,14 @@ const ProductCard = ({
...
   subjects,
 }) => {
+  const intl = useIntl()
   return (
     <Card>
       <ImageWrapper background={background}>
-        <Image fixed={image} alt={`${name} logo`} />
+        <Image
+          fixed={image}
+          alt={`${name} ${translateMessageId("logo", intl)}`}
```
ryancreatescopy, line 132: Something I've realised is that by just allowing a translator to translate "logo", these alt tags may not make sense in some languages.

Although Solidity logo in English... in Spanish "Solidity" would come after logo like "Logotipo de Solidity" (that's just a poor google translate).

I think we should use `alt={alt}` in this component and then wherever ProductCard is used we add description to the array and reference a full key in the JSON file, e.g. `"solidity-logo": "Solidity logo"`

I've done it for the stablecoins page here, for reference: https://github.com/ethereum/ethereum-org-website/pull/1938/commits/842088053f17f65d199ba884741d3cdc2c02811b

src/components/ProductList.js
```diff
@@ -83,7 +86,12 @@ const ProductList = ({ content, category }) => {
         return (
           <Item key={id || idx}>
             <ImageContainer>
-              {image && <Image fluid={image} alt={`The ${title} logo`} />}
+              {image && (
+                <Image
+                  fluid={image}
+                  alt={`${title} ${translateMessageId("logo", intl)}`}
```
ryancreatescopy, line 92: Same as ProductCard

src/components/WalletCard.js
```diff
@@ -63,12 +65,13 @@ const Description = styled.p`
...
 
 const WalletCard = ({ wallet }) => {
+  const intl = useIntl()
   return (
     <Card to={wallet.url} hideArrow={true}>
       <ImageWrapper background={wallet.brand_color}>
         <Image
           fixed={wallet.image.childImageSharp.fixed}
-          alt={`${wallet.name} logo`}
+          alt={`${wallet.name} ${translateMessageId("logo", intl)}`}
```
ryancreatescopy, line 74: Same as ProductCard

src/pages-conditional/dapps.js
```diff
@@ -993,7 +993,7 @@ const DappsPage = ({ data }) => {
       ),
       url: "https://uniswap.exchange/swap",
       image: data.uniswapec.childImageSharp.fixed,
-      alt: "Uniswap Logo",
+      alt: `Uniswap ${translateMessageId("logo", intl)}`,
```
ryancreatescopy, line 996: We should reference a `"uniswap-logo"` and include the full string `"Uniswap logo"` as in this commit for stablecoins: https://github.com/ethereum/ethereum-org-website/pull/1938/commits/842088053f17f65d199ba884741d3cdc2c02811b

I kept the strings to just `product-logo` so we can reuse them across pages.

src/pages-conditional/developers/learning-tools.js
```diff
@@ -175,6 +175,7 @@ const LearningToolsPage = ({ data }) => {
         <StyledCardGrid>
           {sandboxes.map((sandbox, idx) => {
             return (
+              /* Product Name: Do Not Translate name/alt */
```
ryancreatescopy, line 178: why are we not translating alt?

src/pages/build.js
```diff
@@ -281,6 +285,7 @@ const BuildPage = ({ data }) => {
       <ActionCardContainer>
         {resources.map((resource, idx) => {
           return (
+            /* Product Name: Do Not Translate title/alt */
```
ryancreatescopy, line 288: why are we not translating alt?

ryancreatescopy (changes requested):

src/components/CalloutBanner.js
samajammin: Agreed. I think we should just keep this component "dumb" in the sense that it has no knowledge of `intl`. Let's keep that logic in the parent components & explicitly pass the `altText` as a prop.

In general, I think our approach should be to minimize the number of components that must instantiate the `intl` object with `useIntl()`

src/components/EthVideo.js
```diff
@@ -14,7 +17,11 @@ const EthVideo = ({ className, videoSrc }) => {
     <div className={className}>
       <video
         id="hero-video"
-        alt={`ETH glyph video - ${isDarkTheme ? "Dark" : "Light"}`}
+        alt={`${translateMessageId("comp-eth-video-alt", intl)} - ${
```
samajammin, line 20: While this implementation looks correct to me, @ryancreatescopy what are you thoughts on this? Having a different alt text for dark vs. light seems overkill to me. It's not something I'd want to to roll out on every image variation.

src/components/ProductCard.js
samajammin: Agreed! Nice catch. Same with my comment as well, I think where possible we should keep components dumb & explicitly pass the alt text as a prop.

src/pages-conditional/dapps.js
samajammin: But in that case wouldn't we need different translation strings for every product? I don't think it's worth our translator's time to translate a bunch of those ("Uniswap logo", "Darkforest logo", "Pooltogether logo", etc.). In general I don't think proper nouns like company names should be put into translation files.

src/pages-conditional/developers/learning-tools.js
samajammin: I suspect because they're all product/company names?

samajammin (changes requested):
Left some additional comments. I agree with your points @ryancreatescopy except for including product/company names.

wackerow:
@ryancreatescopy @samajammin wow, lot of stuff here, thanks for going through this guys... let me try to summarize:

- The brand/product names weren't translated because they are proper names, and I agree they should be left out of translation files as much as possible. 
- Fragments won't translate well: ie. "___ logo" or "___ image", and in these two examples they seem to complicate things since they're often next to a brand name ("Uniswap logo")... this mixes a name that should not be translated, with english, causing increased complexity when honestly, "Uniswap" would do just fine imo (same with all of these examples... I think just dropping the word "logo", "image" or "- light/dark mode" would prevent unnecessary efforts by translators. 
- Along those lines, `ProductList`, `WalletCard`, `ProductCard` can be simplified to remove string interpolation logic from  alt={``}

Let me know how that sounds and I can patch this today

src/pages-conditional/dapps.js
ryancreatescopy: But this string isn't just the product name, it includes "logo"... this implements a structure and makes it a sentence which will differ in structure based on language. If we want to do localisation as properly as possible, I believe we should include these as strings. It's not a lot of extra work IMHO.

src/pages-conditional/developers/learning-tools.js
ryancreatescopy: An alt description that is just a product name isn't very descriptive. At the least it'd be a logo I imagine in which case I think it needs a string.

src/components/EthVideo.js
ryancreatescopy: I think this could just have a shared "Video of a transforming Eth glyph" alt description. No need to describe the mode.

src/pages-conditional/developers/learning-tools.js
wackerow: This page for example (`learning-tools.js`) has the "alt" text hard coded into the product data array which is simply the product name (ie. `alt: "Remix",`).. Others use string interpolation to add " logo", like the `dapps.js` page. I didn't change any of the existing terms, just tried to update what was already there as it was, which is why this did not have any translation for "alt"... because "alt" was just "Remix" or "Ethernauts" for example.

I do think some standardization one way or the other would help. And I kinda agree that, given the purpose of "alt", especially for visually impaired users who require text readers, having the word "logo" _does_ give more context for the user. This of course would come with the trade-off of more lines to translate, though I doubt it would be all that difficult to translate "logo" a few times (by having "Uniswap logo", "Ethernaut logo", "Remix logo" etc, as unfragmented language strings)

wackerow:
First some thoughts...
For the intended purpose of "alt" to be realized, I feel it should be a descriptor of the image being shown to aid in accessibility for those with disabilities. I would argue "Uniswap logo" is more helpful than "Uniswap". Yes, this adds several lines to the translation file, however, to batch edit these should be cake, and a translator can easily reorder these words if the syntax for their language requires it, without fragmented strings.
https://cliquestudios.com/alt-text/ - just one article talking about how descriptive alt texts can be really helpful for others

Secondly... I implemented those changes (of course any of this can be reverted if decided against)
- Gets rid of the fragmented strings which ended in "logo", "image" or "light/dark mode".
- Removes string interpolation logic in the `alt` tags to simplify to `alt={alt}` in CalloutBanner, ProductCard, ProductList
- Added a "logo" alt text to `en.json` for all product images being displayed

Again, I know it adds several en.json entries, but even if someone translating only knows how to copy/paste, this shouldn't be too difficult to translate in bulk. @samajammin Lemme know what you think, and if you find anything I missed, and I can adjust it further tomorrow

samajammin (approved):
Thanks for making these changes! I know we threw a lot at you 😅

samajammin:
@ryancreatescopy looks like we need you to approve as well.

ryancreatescopy (approved):

PR #2051, opened 2020-12-04
Added transparent logo.

OliverPetrovic:
I edited the logo used in the readme.md file and removed transparency, so it doesn't look bad in dark theme.
I also edited the global font size to 62.5%:

1rem (root font size) = 16px (but may vary)
1rem = 62.5% => 1rem = 10px (10px is easier to use than 16px)
And then I reverted it back in body {}, font-size: 1.6rem.

This means, that now anyone can use 1rem as 10px, but the font size stays the same.
Everything defined by rem in the css will be more responsible, even though I don't see lots of it there now.

## Description

Added eth-transparent.png
Changed eth.png to eth-transparent.png in the readme.md file
Added some CSS for custom scrollbar into the layout file.
It looks more modern now :+1: 

## Related Issue

image
image


samajammin (changes requested):
Hey @OliverPetrovic - thanks for this contribution!

The README image update looks great.

Re: the CSS changes around font sizes, the current updates won't work as-is. Check out the preview deploy: https://deploy-preview-2051--ethereumorg.netlify.app/en/

e.g. the site nav text is now enormous:
Image 2020-12-04 at 10 50 46 AM

While I'm not opposed to this base change, I'm not convinced by it. I can see the developer experience benefit to be able to think of rems in factors of 10px but in reality this would require us to update large amounts of CSS (doing a grep search in our codebase for `rem;`, there's over 1K results). There's also the fact that we tend to style layouts in multiples of 8px (something we should document more in our design system 😅), so working with `rem` as a factor of 16px hasn't really been an issue.

Let me know your thoughts on this. If we want to roll out a refactor of all our `rem` CSS use, I suggest we move that to a separate PR so we can merge the image change.

Thanks again for taking the time!

OliverPetrovic:
Hello,
thank you for letting me know. 
I removed the rem stuff, so it should be just the image right now.

Thank you!

samajammin (approved):
Thanks for this!

samajammin:
@all-contributors please add @OliverPetrovic for content.

samajammin:
@OliverPetrovic out of curiosity, what do you use to view Github in dark mode? 🌚

OliverPetrovic:
Hello,
I use this chrome extension:

https://chrome.google.com/webstore/detail/github-dark-theme/odkdlljoangmamjilkamahebpkgpeacp

Thanks :)
~ Oliver

wackerow:
Love the timing of this... GitHub just officially released Dark Mode haha.... I was using "Dark Reader" extension, but this is better

PR #2098, opened 2020-12-10
Light mode for code [Fixes 2096]

wackerow:
## Description
Adds light-mode to code blocks and reverts deep purple changes to inline code.

- Separated `code` and `tt` css selectors to style inline `code` the way it was previously
- Passed a `theme` prop to the `Highlight` prism component which allows altering the syntax highlighting color theme of code blocks
- New light mode theme created and passed conditionally when page is in light-mode
- The `defaultProps` contains a `theme` object which I also extracted in the event this deep purple theme wants to be adjusted (if there is no immediate interest I could revert the dark mode object to `...defaultProps.theme` to trim out a chunk of code)
- Adjusted coloring of copy/expand buttons to match

Light:
image
Dark:
image


wackerow:
Noticed this blended in completely on dev docs page in light mode, so lightened the background a shade and added 1px border:
image

ryancreatescopy (approved):

samajammin (approved):
Looks great, nice work!

PR #2102, opened 2020-12-11
Filters query param passed to /find-wallet/ page [Fixes #2087]

### LINKED ISSUE #2087: Add filters to URL params

Situations like this it'd be nice to be able to link directly to the wallets that have the given feature:

image

Investigate solutions to support URL, e.g. React router's useParam hook.

**Additional context**

Hat tip to @ph5500 for the suggestion.

wackerow:
## Description
- Passes `filters` queryParam when linking to `wallets/find-wallet/`
- Value passed is a comma-separated string of `id` values corresponding to the wallet filter categories desired
- Values are parsed and passed to the initial state of `selectedFeatureIds` if there are any
- Set this up for the callout card in `StablecoinAccordion` which links to `/find-wallets/`. This now loads with `has_card_deposits` and `has_dex_integrations` already checked.


samajammin (changes requested):
Awesome! The functionality looks good - the wallet results filters correctly. I notice when loading the page however that the check mark doesn't appear on the checkbox:

Image 2020-12-14 at 11 00 06 AM

The checkbox should display as checked:
Image 2020-12-14 at 11 00 12 AM

Mind investigating this?

samajammin:
@wackerow @ryancreatescopy another question (that could be solved in a separate PR) - should we also update the browser URL when the user applies filters? That would make filters more easily shareable.

wackerow:
This relocates the logic into `WalletCompare` under `useEffect`, passing the `location` prop down from `find-wallet.js`. If current filters are empty and there are query filters passed (like on first load), it will load these into state.

@samajammin This is fixing the above bug on my production build, but I have not yet addressed updating this query string when selecting filters to make this even more sharable. Assuming this works on Netlify build I can look at adding that if you'd like.

wackerow:
@samajammin That should do it... lemme know what you think!

wackerow:
Hah, oops, had forgotten to add this to `clearFilters()`... this last commit should restore that functionality

samajammin (changes requested):
Nice work! This look great.

The _only_ strange issue I see is when loading a URL with query params, e.g.
https://deploy-preview-2102--ethereumorg.netlify.app/en/wallets/find-wallet/?filters=has_high_volume_purchases,has_dex_integrations

When I click to "Clear filters", it doesn't actually clear the filters.

wackerow:
Hm, yeah I see that now... that's a little strange and seems to not happen every time for me, but will take a look at this to try and fix that issue before implementing it.

wackerow:
That seems to do the trick... @samajammin should be all set now

samajammin (approved):
Looks great! Nice work 😄

PR #2108, opened 2020-12-12
Fix renders twice on page load

esteticalVE:
## Description
 Created function for detecting and setting the right theme;
Resolved issue when component renders twice.
Layout was rendered twice because initial state in useState might be different from preferred by the user.
So the state changed and triggered second render;

### DIFF
--- a/src/components/Layout.js
+++ b/src/components/Layout.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from "react"
+import React, { useState, useEffect, useRef } from "react"
 import { ThemeProvider } from "styled-components"
 import { IntlProvider, IntlContextProvider } from "gatsby-plugin-intl"
 import styled from "styled-components"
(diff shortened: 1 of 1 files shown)


src/components/Layout.js
```diff
@@ -52,17 +52,25 @@ const StyledBannerNotification = styled(BannerNotification)`
   text-align: center;
 `
 
-// TODO `Layout` renders twice on page load - why?
+function isDarkThemeUtil(isUseEffect, setThemeCb) {
```
wackerow, line 55: consider refactoring to arrow syntax to match codebase convention
ie: `const isDarkThemeUtil = (isUseEffect, setThemeCb) => { ... }`

wackerow (changes requested):
Thanks for taking a look at this! I am no longer getting a duplicate render when `Layout` loads, however, this seems to be causing issues with the initial rendering of Dark vs Light mode:
image
In the above you see portions of the site styled for light mode, and others styled for dark mode. This does not seem to show in when running in a dev environment, likely because of access to the "window" value that is not available during server side building for production. (btw, you can test a local "production build" by running `yarn start:static` which will run on port 9000, and you will likely see this same thing)
Not sure best way to fix this, but hope this helps a little if you can find a way to adjust it to fix the initial rendering. Thanks again for the help!

samajammin (changes requested):
Thanks for looking into this @esteticalVE!

I'm seeing the same issue as @wackerow in the preview deploy:
https://deploy-preview-2108--ethereumorg.netlify.app/en/

Please see if you can solve for that, otherwise close the PR. As you pointed out, the useEffect is causing the re-render so perhaps there is no way around it in order to adjust for the user's browser settings & to set the mode accordingly. I'm not sure but I'd certainly welcome solutions here! Thanks again.

esteticalVE:
@wackerow @samajammin, thanks for feedback, i've closed this one, i will try to solve it once again, will open new PR

PR #2126, opened 2020-12-15
Adds translation banner

ryancreatescopy:
## Description

Adds translation banner component for our upcoming translations improvements.
Example banner is present on the homepage on the local build.


ryancreatescopy:
image

I want to try and ensure the mobile nav is the top layer when activated, not the translations banner. 

I tried to do this by removing the `z-index` from the banner but then for some reason the images display over the top of the banner. No other content seems to do this...

image

Issue #1921, opened 2020-11-23
Prevent glossary link wrapping


ryancreatescopy:
**Describe the bug**

The new glossary link with icon shouldn't wrap onto two lines. The icon should be treated as part of the last word, so instead of only the icon wrapping onto a new line "network" should wrap too (or the entire link if easier).

![image](https://user-images.githubusercontent.com/40891631/99969678-b8924180-2d92-11eb-82d0-d8cd3280cd5b.png)

samajammin:
I suspect this is the same issue as #980 - we solved by changing from `display: inline-block;` to `display: inline;` (see https://github.com/ethereum/ethereum-org-website/pull/1229/files) but unfortunately this may remove the hover effect 😕

santdeleon:
I think i've got a fix for this. Should have a PR shortly 👍

PR #2104, opened 2020-12-11
Fragmented strings cleanup [wip]

ryancreatescopy:
## Description

Removing fragmented strings (due to links) for localisation purposes.

## Related Issue

#2048

Issue #2047, opened 2020-12-03
Break out translation json files


samajammin:
**Is your feature request related to a problem? Please describe.**

Our app uses [`gatsby-plugin-intl`](https://github.com/wiziple/gatsby-plugin-intl) for internationalization. This plugin expects a single json file (located at /src/intl/${lang}.json) containing all strings for each language translation.

Here's our current en.json file (~1500 lines):
https://github.com/ethereum/ethereum-org-website/blob/dev/src/intl/en.json

This creates issues with our translation process in [Crowdin](https://crowdin.com/project/ethereumfoundation), where volunteers translate these files. Currently we wait until a file is 100% translated before passing it on to our professional translation review service. Waiting for the entire file to be translated is far from ideal... we want to be able to review & publish translation updates on a page-by-page or even component-by component-basis in order to publish translations ASAP.

**Describe the solution you'd like**

- Split out the monster en.json file into a set of distinct files (per page? per component? plus common strings?)
- Create a node script to merge these individual json files on build, so the app continues to work with the intl plugin

**Describe alternatives you've considered**
Still not sure how to best organize the json files...
- Break out the json files by end-user page?
- Break out the json files by component?
- Should we consolidate duplicate strings when possible (e.g. in a common.json)?

wackerow:
I've been thinking about something like this as I've been working with the translation set up. 

As I've been setting them up, I try to organize them by the `js` file referencing them, which is typically either a page layout , or a reusable component... I think breaking them out by component/page is most natural, that way a new component can have its own language json, and can be translated and approved individually. 

I really like the idea of a "common" json as well. I'm not sure how it would work behind the scenes in terms of approving each change however, given that this list would likely change/grow over time as more phrases collide.

@samajammin Side question regarding translations while we're on it: Should we be proactively preparing for syntax issues between languages?  Many of our sentences are broken into pieces because of things like links being embedded, which I feel could cause difficulties when translating to a language where "subjects" "verbs" and "adjectives", for example, don't go in the same order. 
ie. `The green car` = `El coche verde` in Spanish. If we translated each word as a separate fragment, our translation would yield "El verde coche"

ryancreatescopy:
This is a really good point @wackerow — ideally we shouldn't have links in the middle of sentences if they require a separate string. I've tried to do that as much as possible with links like "More on X" coming at the end of paragraphs rather than in the content itself, but have definitely slipped into bad habits in the last few pages. 

[...]

PR #2130, opened 2020-12-16
Fix line wrapping for Internal Links

santdeleon:
## Description

- Added css styling to internal links to remove line wrapping. 
- Broke the **page-eth2-staking-validators-desc** json value out into three parts/components such that one of the components can become a glossary link to the 51% attack url (**/glossary/#51-attack**).

**Eth2 Staking Validators description changes**
- The json value pair for **"page-eth2-staking-validators-desc"** has become three pairs.
1) **page-eth2-staking-validators-desc**
2) **page-eth2-staking-validators-desc-link**
3) **page-eth2-staking-validators-desc-2**
This allows us to break out and turn one portion of the description into a link.

**Link Changes**
- The `<InternalLink />` component was given the css style `white-space: nowrap;`. This keeps `<GlossaryIcon />` components from wrapping onto new lines when a user's screen is resized. It also keeps the entirety of all internal links from line wrapping. In short, internal links will always show the full length of the link at any given time.
## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/1921


src/components/Link.js
```diff
@@ -31,6 +31,7 @@ const ExternalLink = styled.a`
 `
 
 const InternalLink = styled(IntlLink)`
+  white-space: nowrap;
```
wackerow, line 34: Unfortunately this will affect all other internal links, and force them to not wrap at all when they should word-wrap.

wackerow (changes requested):
Hey @santdeleon thanks for taking a look at this! This would potentially fix that specific link commented on in the original issue, but changing `InternalLink` to `nowrap` will affect many other links, and ultimately these should word-wrap, with the issue being just the `[?]` icon that trails `/glossary/` links specifically should stick to the last word.

There has also been a push recently to limit the breaks within sentences that require translation, as this complicates releasing these pages in other languages with proper syntax.  Hope this helps add some guidance on what the goal is here, and I certainly appreciate the work. 

Btw- Took a look at this issue myself previously and it's tricky since the way it works on the 'External Links' is it uses a css pseudo-element `::after` with `content: "↗";` to add the arrow, which is text. In this situation we use an `Icon` which is a React component, and that cannot be passed into the css `content` property like `↗` can. Another attempt was to play with margin-right of the text for glossary links to make room for the Icon on the right, but this also does not work properly with word-wrapping on inline items. 🤷‍♂️

santdeleon:
@wackerow I see. I thought that might be a problem. This is a bit tricky but I think I have a better solution.

What about replacing the `whitespace: nowrap` with 
```
white-space: ${({ className }) =>
    className && className.includes("nowrap") && "nowrap"};
```
With this solution all you would need to do is pass a `"nowrap"` className to your Link component and then the css can conditionally render the apply `white-space` styling as necessary.

I'm sure there are many ways to solve this but this might have the least overhead?

I have it working locally and am happy to send it through if you think this could work.

wackerow:
Right, this would help make it more selective, but my concern would be that ultimately the `nowrap` value causes the entire inline item (the entire link text) to stay on one line, versus getting only the trailing `[?]` icon to "not wrap." I previously tried applying `nowrap` to just the glossary icon but it doesn't work that way since it's a whole new inline item. 

Ultimately not up to me though, feel free to push changes! =)

santdeleon:
Right right. Hmm. I think this may be our best bet at the moment. The only way I can think of that works for sure is using `content` like we do with ExternalLink, but since content can't take html/jsx we can't render the icon. 😣

I'll push my current fix for now and touch back on this pending team approval etc.

samajammin (changes requested):
Thanks for looking into this @santdeleon! Thank you @wackerow for the assist as well 😄 

@wackerow brought up a valid point - we've made a push to remove these type of fragmented links because they break up sentences, making them difficult to translate (more context in #2047 & #2104). As a result your changes to the Staking page are outdated & no longer necessary.

As for the styling changes, I think your solution is not ideal (since it forces the entire link to `nowrap`) BUT I think it's better than the current situation. However I believe there's a way to accomplish this without requiring the explicit class name for glossary links. Since we already have a `isGlossary` variable in the Link component, let's just pass that boolean as a prop to our `InternalLink` component. Then you can add your logic based on that, e.g. something like:
```
const InternalLink = styled(IntlLink)`
  white-space: ${(props) => (props.isGlossary ? `nowrap` : `normal`)};
  ...
```

Please make that update & remove the changes the Staking files & we can get this merged. Thanks again!

santdeleon:
Hey @samajammin! Upon your suggestions above, it seems this PR becomes invalid after implementing them.

So when I revert the staking page and wrap the description in a `Link` component (which I assume is what was expected) the whole description becomes a link. Then applying the style changes using the `isGlossary` prop like in the snippet you left above, we end up with one giant link that sticks to one line, never wraps, and overflows on the X-axis. 

I think my solution only works (at least for this specific issue) if the descriptions are broken up.

**So i'm wondering:**
1. If the changes in this PR to need to be abandoned entirely
2. Or if we actually just intend to remove the `Link` component entirely and leave the description "link-less" (like how it is in prod) - making the style changes unnecessary.

Sorry If I explained that in a confusing way! Please do let me know your thoughts. Thanks!

samajammin:
@santdeleon hmmm... not sure I follow entirely. I wouldn't worry too much about the specific example in the the issue. I think merely making that style change I mentioned above (& only that change) would suffice here. LMK if you have any other thoughts / questions. Thanks again!

samajammin (approved):
Thanks again!

samajammin:
@all-contributors please add @santdeleon for code.

Issue #2072, opened 2020-12-07
Roll out every page for every language


samajammin:
**Is your feature request related to a problem? Please describe.**

Currently we only display pages for content we've fully translated.

This results in poor UX for visitors where falling back to English may be a better option. e.g. someone visiting ethereum.org/ko will otherwise never know that ethereum.org/en/get-eth/ exists.

**Describe the solution you'd like**

Let's deploy every page for every language with banners notifying the user that this page needs to be translated.

![image](https://user-images.githubusercontent.com/8097623/101375934-d072c680-3864-11eb-9598-4177cc006844.png)

**Describe alternatives you've considered**
Wait until pages are fully translated until deploying the page.

**Additional context**

wackerow:
This is a really great idea... especially since there are also page translator extensions that can translate for the user as a backup... ie, the above page in Korean:
![image](https://user-images.githubusercontent.com/54227730/101381451-9d800100-386b-11eb-9a40-4beabca8e71e.png)
But this only works if the user can find that page.

samajammin:
Solved in #2129

PR #2161, opened 2020-12-21
Translate breadcrumbs

samajammin:
## Description

Add translation support to Breadcrumbs component & necessary strings to translate.

## Related Issue
#2072


samajammin:
FYI @ryancreatescopy to keep things simple, I've just added the needed slug fragments to the common.json file. Fetching the page title for each page would have added a bunch of complexity.

PR #2172, opened 2020-12-24
Translation styling adjustments

wackerow:
## Description
- Changed couple instances of translation text that was pre-styled as all-caps. Moved styling for all caps to css using `text-transform: uppercase;` 
- Cleaned up some stray extra spaces in some translation phrases that were not being rendered. Many of these were at the beginning of paragraphs... if the goal here was to indent the text, let me know, as we should use styling for that as well, since these leading spaces in the JSON string values do not render. 
- Also, while eyeing spaces, added a space to two instances of "32ETH" to make it "32 ETH" for consistency as this is what is used elsewhere on the site.

## Related Issue
(None filed, per conversation in #2047)


ryancreatescopy (approved):
lgtm

samajammin (approved):
Awesome - thanks!

PR #2187, opened 2020-12-27
Update index.md

anh193:
fix typo


ryancreatescopy (commented):
Thanks for PR but we refer to these libraries as convenience libraries (noun) as they remove the need for interacting directly with the JSON RPC. Your PR is still accurate but we don't need to make the change. Thanks again.

PR #2196, opened 2021-01-01
Fix comment for `pizzaToOwner` variable in CryptoPizza example

BriungRi:
Fix the typo in the comment describing the `pizzaToOwner` variable

## Description
The comment describing the mapping was flipped. Rewrote comment to
reflect the actual mapping.

## Related Issue
n/a


samajammin (approved):
Thanks for catching this!

PR #2198, opened 2021-01-04
Fix broken link

staktrace:
Update link to one that works, for what appears to be the new project that replaced the old one.

## Description

wiki.parity.io redirects to openethereum.github.io/wiki but it's a 404. There is a project by that name, which is a Rust ethereum wallet, but the URL just needs fixing.


ryancreatescopy (approved):
Thanks for updating!

ryancreatescopy:
@all-contributors please add @staktrace for bugs and content.

PR #2209, opened 2021-01-05
Removed submission form and added banner that the wave is over

gilbertginsberg:
## Description

Removed submission form and added banner that the wave is over


samajammin (approved):
Thanks!

PR #2249, opened 2021-01-13
Removed "/en" from tutorial links

wackerow:
## Description
Tutorial card links are being generated from the `slug` being pulled from the query data. These include the `/en` prefix, which interferes with routing a user of another language to the appropriate language path. 
This conditionally checks for '/en' at the beginning and removes it. Now automatically generates correct language path in it's place. 

## Related Issue
Going to tutorials page in a language:
[https://ethereum.org/de/developers/tutorials/](https://ethereum.org/de/developers/tutorials/)

All links go to English tutorial page 
[https://ethereum.org/en/developers/tutorials/sending-transactions-using-web3-and-alchemy/](https://ethereum.org/en/developers/tutorials/sending-transactions-using-web3-and-alchemy/)

instead of language page
[https://ethereum.org/de/developers/tutorials/sending-transactions-using-web3-and-alchemy/](https://ethereum.org/de/developers/tutorials/sending-transactions-using-web3-and-alchemy/)


ryancreatescopy (changes requested):
When I navigate to `/de/` and click on `/developer/docs/`, I get the following error:
image

`/de/developers/tutorials/` works as intended, but when I click on a tutorial I get this error:
image


Maybe this is the peril of running it in local, but just in case this is helpful!

wackerow:
I think this is related to the GitHub API rate limits... we seem to be getting a lot of these lately, but will take a closer look

samajammin (approved):

PR #2262, opened 2021-01-15
Legal pages exclusively in English

wackerow:
## Description
- Adjusted `gatsby-node.js` to select out "legal" pages, and prevent building these pages in other languages beyond English.
- Footer links all go to `/en/` path for these pages, regardless of current language

## Related Issue
(Only create legal pages in English)


samajammin (approved):
Nice

PR #2252, opened 2021-01-14
Language rollout v2.0 - index & developers index

wackerow:
## Description
- Completed CrowdIn files `page-developers-index.json` and `page-index.json` migrated into repo utilizing folder structure inline with the english file structure (`src/intl/en/`)
- Existing language files were migrated into each of their respective subdirectories (ie `src/intl/ar.json` -> `src/intl/ar/ar.json`), and conflicting keys-value pairs were removed in favor of newly completed translation files
- `merge-translations.js` modified to parse all languages folders, not just English
- .gitignore updated to ignore all _built_ language JSONs (`src/intl/*.json`)

## Related Issue
(Rolling out the translated Developer Home & Main Page)

PR #2283, opened 2021-01-20
Remove generated intl/{lang}.json files from git

samajammin:
## Description

We updated our `merge-translations` script in #2252 to generate the `${lang}.json` string files for all languages, not just English. This removes these generate files from git's version control.

## Related Issue
#2252


samajammin:
@wackerow FYI [here's an article](https://medium.com/better-programming/how-to-remove-committed-files-from-git-version-control-b6533b8f9044) on how to do this.

The TL;DR is to run `git rm --cached path/to/file`. This keeps the file while removing it from git.

Side learning for me: unfortunately `git rm --cached src/intl/*.json` didn't work, I had to run it for each file individually (e.g. `git rm --cached src/intl/ar.json`).

PR #2289, opened 2021-01-20
Makes dapp categories linkable with URL

wackerow:
## Description
- Updated `dapps` page to track selected category (`finance`, `collectibles`, `gaming`, or `technology`) within query params of the URL
- Will load with selected category already filtered when included in link, ie: `https://ethereum.org/en/dapps?category=technology` pre-selects the technology section
- Edges: Defaults to finance if queryParam is either not included or an invalid category. Uses first if array passed.
- Maintains the scroll-navigation built into the mobile-only component at the bottom of this page (will bring user to top of #explore section)


samajammin (approved):
Nice!

PR #2294, opened 2021-01-21
Removed wallets string fragments

ryancreatescopy:
## Description

Removed fragmented strings on wallets page due to `<b>` tags – flagged by translators.


samajammin (approved):
Nice!

PR #2352, opened 2021-02-01
Add JSON-RPC API docs [Fixes #2322]

### LINKED ISSUE #2322: Migrate/add JSON-RPC docs to ethereum.org
## Description

- [ ]  Make space for JSON RPC docs on ethereum.org
- [ ]  Can the UX/readability of specs be improved?
- [ ]  Publish specs
- [ ]  Set up redirects

## Context

https://github.com/ethereum/eth-wiki/issues/55

wackerow:
## Description
Adds json-rpc end point reference docs to developer docs section
Initial commit places this information within the "Ethereum client APIs" section of developer docs
Endpoints placed on dedicated page, with initial section breakdown using gossip/state/history model.  

Pages:
- https://deploy-preview-2352--ethereumorg.netlify.app/en/developers/docs/apis/json-rpc/
- https://deploy-preview-2352--ethereumorg.netlify.app/en/developers/docs/apis/json-rpc/endpoints

## Related Issue
#2322


ryancreatescopy (changes requested):
I think both pages could use some context-setting. It feels like we jump straight into it without really explaining what it's for. It also feels to me like these docs always relied on developers knowing a lot already? Do we think they're beginner friendly? I realise the scope of this issue was to migrate but I think we should aim to make these more user-friendly if possible. Perhaps that's a separate PR afterwards. cc @samajammin

src/content/developers/docs/apis/json-rpc/index.md
```diff
@@ -5,19 +5,19 @@ lang: en
...
-Geth 1.4 has experimental pub/sub support. [More information](https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB)
+- [Geth v1.4](https://geth.ethereum.org/docs/rpc/pubsub)
 
-Parity 1.6 has experimental pub/sub support. [More information](https://github.com/paritytech/parity/wiki/JSONRPC-Eth-Pub-Sub-Module)
+- [Hyperledger Besu 1.3](https://besu.hyperledger.org/en/stable/HowTo/Interact/APIs/RPC-PubSub/)
 
-Hyperledger Besu 1.3 has pub/sub support. [More information](https://besu.hyperledger.org/en/stable/HowTo/Interact/APIs/RPC-PubSub/)
+_[Jump ahead to endpoints](/developers/docs/apis/json-rpc/reference)_
 
 ## JavaScript API
```
ryancreatescopy, line 18: Don't forget heading IDs e.g. {#javascript-API}

src/content/developers/docs/apis/json-rpc/index.md
```diff
@@ -5,19 +5,19 @@ lang: en
 sidebar: true
 ---
 
-[JSON](http://json.org/) is a lightweight data-interchange format. It can represent numbers, strings, ordered sequences of values, and collections of name/value pairs.
+A remote procedure call, or RPC, is a method of retrieving or being notified of information from a given data source in an asynchronous manner. [JSON-RPC](http://www.jsonrpc.org/specification) is a stateless, light-weight RPC protocol, that defines several data structures and the rules around their processing. It is transport agnostic in that the concepts can be used within the same process, over sockets, over HTTP, or in many various message passing environments. It uses [JSON](http://json.org/) as data format.
```
ryancreatescopy, line 8: I might be wrong here but this feels like a generic explanation of what an RPC is? Is there something we can add to provide context of why it's relevant to Ethereum development? Maybe that's obvious from this overview if you're a developer but it's not clear to me.

src/content/developers/docs/apis/json-rpc/reference/index.md
```diff
@@ -5,7 +5,11 @@ lang: en
 sidebar: true
 ---
 
-## Gossip Methods
+A handful of core JSON-RPC endpoints require data from the Ethereum network, and fall neatly into three main categories: [Gossip](#gossip_methods), [State](#state_methods), and [History](#history_methods)
```
ryancreatescopy, line 8: This opener should reference "Methods" if possible

src/content/developers/docs/apis/json-rpc/index.md
wackerow: Happy to... Was on the fence about keeping it more generic, but can add some tie-in here.

src/content/developers/docs/apis/json-rpc/index.md
wackerow: Good catch... only did this on the `/reference/` page, I'll update this file

src/content/developers/docs/apis/json-rpc/reference/index.md
wackerow: Rethought this.. there was really no need to use that term. Switched to using the term "endpoint" and added this to a pre-req section on the overview page.

src/content/developers/docs/apis/json-rpc/index.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Java {#java}
+
+Run a Besu node on mainnet with the HTTP JSON-RPC service enabled:
+
+```bash
+besu --rpc-http-enabled
+```
+
+More details can be found in the [documentation](https://besu.hyperledger.org/en/stable/Reference/CLI/CLI-Syntax/#rpc-http-enabled).
```
NicolasMassart, line 96: Could we link to this https://besu.hyperledger.org/HowTo/Interact/APIs/Using-JSON-RPC-API/ which explains more how to run Besu with RPC and how to use it. The ref link is correct but we prefer to point to guides instead of refs directly. Also I suggest a clearer link name that will help for search and indexing that a generic one and include some examples of parameters you can find in this doc as they are described for other clients. Thanks!
```suggestion
Read the [Besu JSON-RPC documentation](https://besu.hyperledger.org/en/stable/HowTo/Interact/APIs/Using-JSON-RPC-API/) for details about configuring the service (listening ip, port, cors, ...).
```

wackerow: @NicolasMassart I'm in favor of this, thanks for posting. I agree this link is clearer, and the landing page is more inviting. 👍

src/content/developers/docs/apis/json-rpc/index.md
```diff
@@ -0,0 +1,154 @@
+---
+title: JSON RPC API
+description: A stateless, light-weight remote procedure call (RPC) protocol
+lang: en
+sidebar: true
+---
+
+The JSON-RPC specification allows for an application programming interface (API), similar to REST or GraphQL, with endpoints served from an Ethereum node. Several [JavaScript](/developers/docs/apis/javascript/#available-libraries) and [backend](/developers/docs/apis/backend/#available-libraries) libraries exist that conform to this specification, enabling developers to utilize up-to-date on-chain data while building [web3](/developers/docs/web2-vs-web3/) enabled [Dapps](/glossary/#dapp).
```
ryancreatescopy, line 8: This first sentence really confuses me.

the JSON RPC spec allows for an API with endpoints served from an ethereum node.

Does this mean it allows you to configure an API to call endpoints from an ethereum node?

I think the first sentence has too much going on to be clear. Once we've fixed this I think we can ship!

ryancreatescopy (changes requested):

samajammin:
Great start on this @wackerow! 

I'm going to reach out for feedback on this from a couple devs who often work with the JSON RPC docs. In particular, I'm wondering:
- Is it best to break out the info into 2 separate pages? Or should we just have 1 page - both an intro & the endpoints. I lean towards I single page.
- How should we organize the endpoints? Does the categorization into gossip/state/history make sense? Or should we just list them all alphabetically? Or would some other form of categorization make more sense?

samajammin:
@wackerow got some feedback on the page:

> I'm really glad this is happening 😍
> 
> Some feedback:
> 
> * I'd remove the go/c++/etc sections. Those are dependant on each client, and outdated. I think I'd just mention that every client implements it and you should refer to their docs to know how to enable/disable it. Maybe the entire "JSON-RPC ENDPOINT" section can go
> 
> * Javascript API can be replaced with a section of libraries built to consume the RPC in the different languages. Web3.js and ethers in js, but there are more libs in other ecosystems.
> 
> * "JSON-RPC support section" I think there's a lot of value in documenting this, but I suspect it would be better to do it per endpoint. I have something like this in mind: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#browser_compatibility -- Including Ganache, Hardhat Network and HEVM would be great here.
> 
> * The way curl examples are shown is pretty cool. 
> 
> * I'm not really sure about splitting the endpoints in those section. I'd normally expect them to be organized by module (i.e. eth_ vs web3_)

I think for this PR we focus on:

1. Updating the content so all info is on one page (i.e. /developers/docs/apis/json-rpc/).
2. Remove the "JSON-RPC ENDPOINT" section w/ language-specific instructions (i.e. go/c++/etc). Instead providing an explanation that all clients (of different programming languages) implement the spec & to check individual client documentation for specifics.
3. Remove the "JAVASCRIPT API" section. Replace w/ an explanation that many convencience API libraries exist (both [frontend APIs](https://ethereum.org/en/developers/docs/apis/javascript/) & [backend APIs](https://ethereum.org/en/developers/docs/apis/backend/) that offer a layer of abstraction.

samajammin (changes requested):
See my comment above 😄

samajammin (approved):

PR #2365, opened 2021-02-02
Change Dapps Features to use Shared Styled Card Component

chase-manning:
This PR changes the Dapps Features section to use the shared StyledCard Component.  

Previously the Dapps Features section looked like this:
image
It can be found on [this page](https://ethereum.org/en/dapps/) near the bottom.  

After the change it now looks like this:
image

This change brings more design consistency to the eth website by aligning the card designs across pages.

One additional benefit to this change is a smaller bundle size as we are able to remove the code for the old component as that was the only place it was used.


samajammin:
Thanks for the PR @chase-manning!

@ryancreatescopy what are your thoughts on this? I don't have a strong preference but I like the visual & interactive elements of the BoxGrid component. Gives the page a bit more life IMO.

src/pages-conditional/dapps.js
```diff
@@ -1521,7 +1519,20 @@ const DappsPage = ({ data, location }) => {
             <Translation id="page-dapps-magic-behind-dapps-link" />
           </Link>
         </Box>
-        <BoxGrid items={features} />
+        <FeaturesContainer>
+          <CardContainer>
+            {features.map((card, idx) => {
```
Kristiyan96, line 1524:
```
features.map((card, idx) => (
                <StyledCard
                  key={idx}
                  emoji={card.emoji}
                  title={card.title}
                  description={card.description}
                />
)
```
No need for the return statement

Kristiyan96 (changes requested):
Small change request

src/pages-conditional/dapps.js
chase-manning: Good spotting! Have just updated 😄

ryancreatescopy:
Hi @chase-manning thanks for digging into the codebase. This difference in style is quite deliberate however. I realise it may not be perfect but we're experimenting with more interactive ways to display content that's maybe not quite so relevant for everyone so people can dig into specific things they're interested in. More generally we wanted to explore some ways to diversify some of the layouts we've been using throughout the site. We don't want to use that standard card layout for everything! So I'm afraid we won't be merging this. 

For design changes like this I'd recommend discussing ideas in our discord before going ahead. We'd love to workshop design ideas with the community and don't want anyone to waste their time.

chase-manning:
@ryancreatescopy Thanks Ryan! And no worries I had fun implementing it 😄

chase-manning:
@ryancreatescopy Thanks Ryan! And no worries I had fun implementing it 😄

PR #2389, opened 2021-02-05
Accessibility improvements

wackerow:
## Description
Fixed a few accessibility issues showing up on the browser devtools

Footer.js has `socialLinks` array with four links using icons and lacking alt text. I tried passing in `alt` text to these links, as well as directly to the icon, but wasn't able to get it to work, so that's the only one left.

## Related Issue
ethereum.org:
image

this PR localhost:
image

...btw @samajammin @ryancreatescopy , SEO and Best Practices rank 100%

### DIFF
--- a/src/components/Nav/Mobile.js
+++ b/src/components/Nav/Mobile.js
@@ -92,7 +92,7 @@ const NavListItem = styled.li`
 const StyledNavLink = styled(NavLink)`
   display: flex;
   align-items: center;
-  margin-bottom: 1rem;
+  margin: 0;
 `
 
 const SectionTitle = styled.div`
@@ -105,7 +105,7 @@ const SectionItems = styled.ul`
 `
 
 const SectionItem = styled.li`
-  margin: 0;
+  margin-bottom: 1rem;
   list-style-type: none;
   list-style-image: none;
   opacity: 0.7;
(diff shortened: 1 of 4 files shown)


src/components/Nav/Mobile.js
```diff
@@ -244,15 +244,14 @@ const MobileNavMenu = ({
                   <SectionItems>
                     {section.items.map((item, idx) => {
```
Kristiyan96, line 245: No need for the return statement. 
```
section.items.map((item, idx) => (
  <SectionItem key={idx} onClick={() => toggleMenu()}>
     ...
  </SectionItem>
)
```

src/components/Nav/Mobile.js
```diff
@@ -244,15 +244,14 @@ const MobileNavMenu = ({
...
                       return (
-                        <StyledNavLink
-                          to={item.to}
-                          isPartiallyActive={item.isPartiallyActive}
-                          key={idx}
-                        >
-                          <SectionItem onClick={() => toggleMenu()}>
+                        <SectionItem key={idx} onClick={() => toggleMenu()}>
```
Kristiyan96, line 247: onClick={toggleMenu}

Kristiyan96 (changes requested):
Some small improvements

src/components/Nav/Mobile.js
wackerow: Agreed, these are scattered throughout the codebase and could be cleaned up in batch in a separate issue. I'll push the syntax fix here though while we're looking at it.

samajammin (approved):

PR #2397, opened 2021-02-08
Fixed typo

EthanSK:
trie -> tree


wackerow:
@EthanSK Thanks for catching this discrepancy... Wondering though if this should go the other way, using "trie" instead of the instances of "tree"...

https://eth.wiki/en/fundamentals/patricia-tree
This eth wiki entry uses "trie" throughout when discussing the modified Merkle Patricia trie specs

@samajammin thoughts here?

ryancreatescopy:
In all the technical docs I've seen it uses `trie`, but I defer to the more technical among us. I didn't even realise the two were interchangeable in this context!

wackerow:
hah, yeah a "trie" is a type of "tree", so it's a bit more specific. My vote would be to switch the three instances on the `accounts` page to "trie"

ryancreatescopy:
@EthanSK do you mind changing all instances of `tree` to `trie` and we can approve/merge. Thanks!

EthanSK:
The new PR is here https://github.com/ethereum/ethereum-org-website/pull/2466

I changed tree to trie in a few other places

Issue #2390, opened 2021-02-05
throwing default message error in console


MonarthS:
**To Reproduce**

1. Open homepage of any language other than english
ex: https://ethereum.org/zh-tw/

**Screenshots**

![image](https://user-images.githubusercontent.com/40236229/106996290-b3943a00-67a6-11eb-996d-8729a1ba5a7f.png)

yakkomajuri:
Seems like errors are being thrown deliberately:

https://github.com/ethereum/ethereum-org-website/blob/816b2de5cfd1e0c6e4c3353ab4da2d0f15dd7189/src/utils/translations.js

yakkomajuri:
@samajammin what's the reason for throwing errors in prod?

samajammin:
Thanks for reporting this @MonarthS!

> @samajammin what's the reason for throwing errors in prod?

Great point @yakkomajuri, we should only throw these errors in development. One solution might be to access an ENV variable: https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/#accessing-environment-variables-in-javascript

samajammin:
Solved in #2410 - thanks @yakkomajuri!

PR #2410, opened 2021-02-10
No errors in prod

yakkomajuri:
## Description

Prevents missing translation errors from being thrown in production.

Putting this PR out to get feedback on the changes, then happy to track down all the `console.error` calls this way.

Adds a `consoleError` export in utils that will only throw errors if running in development.

## Related Issue

Solves https://github.com/ethereum/ethereum-org-website/issues/2390.


wackerow (approved):
@yakkomajuri Hey, thanks for giving this a go! Looks like it should do the trick. 
@samajammin Curious your thoughts on expanding something like this... I suppose I'm hesitant to suppress too many warnings/errors. These specifically are known errors at this point and we're taking a look at ways to prevent them.
If we're not going to reuse this, perhaps it's best to just put this function directly within `translations.js`

yakkomajuri:
Will wait on @samajammin here to determine if I should expand to suppress more of the explicit `console.error` calls or move this to `translations.js`

yakkomajuri:
@wackerow @samajammin moved it over to `translations.js`

wackerow (approved):
@yakkomajuri LGTM! Thanks for your help with this 👍

samajammin (approved):
Thanks for this!

samajammin:
@all-contributors please add @yakkomajuri for code.

Issue #2411, opened 2021-02-10
Fix horizontal scroll


ipapandinas:
**Describe the bug**
The NavContainer width set to **100vw** causes a horizontal scroll bar to appear on a 13-inch laptop screen. In fact, the vertical scroll width is included using the *vw* viewport unit.

**Screenshots**
<img width="1440" alt="issueHorizontalScroll" src="https://user-images.githubusercontent.com/26460174/107552253-834c0180-6bd3-11eb-8ad0-56b10b9b0b6d.png">

**To Reproduce**

1. Go to [Ethereum.org]('https://ethereum.org/')
2. See error

**Expected behavior**
No horizontal scroll needed.

**Additional context**
Width: 1440px - 13-inch laptop screen.

PR #2412, opened 2021-02-10
fix(Nav): NavContainer width fixed to avoid horizontal scroll #2411

ipapandinas:
## Description

The NavContainer width set to **100vw** causes a horizontal scroll bar to appear on a 13-inch laptop screen. In fact, the vertical scroll width is included using the *vw* viewport unit.

Setting width at **100%** solve this bug. Moreover I removed the useless max-width rule. It is already set on the `ContentContainer` component who is wrapping `Nav` (in *src/components/Layout.js*).

## Related Issue

#2411

### DIFF
--- a/src/components/Nav/index.js
+++ b/src/components/Nav/index.js
@@ -19,9 +19,7 @@ const NavContainer = styled.div`
   position: sticky;
   top: 0;
   z-index: 1000;
-  width: 100vw;
-  /* xl breakpoint (1440px) + 72px (2rem padding on each side) */
-  max-width: 1504px;
+  width: 100%;
 `
 
 const StyledNav = styled.nav`


wackerow (approved):
@ipapandinas Looks good to me!... I honestly wasn't able to reproduce this, and we have had some recurrent cache-related issues causing little layout glitches to the site. Happy to approve this as I agree the line you added does the trick, but consider clearing your browser cache and trying the live homepage again... curious if it resolves this horizontal scroll.

ipapandinas:
The issue still persists on a 100% zoom, even after a cache clean up.

wackerow:
@ipapandinas good to know.. thanks for checking!

samajammin:
@all-contributors please add @ipapandinas for bug reports & code.

PR #2417, opened 2021-02-11
Update Loopring link

mfinestone:
Link was the DEX v1, which is deprecated.

Updated to Loopring zkRollup DEX v2


samajammin (approved):
Thanks for the update!

Issue #1148, opened 2020-06-25
Skip Netlify builds on bot PRs


samajammin:
**Is your feature request related to a problem? Please describe.**

Netlify builds on every PR, including from bots like @all-contributors
https://github.com/ethereum/ethereum-org-website/pull/1147
This creates unnecessary build queues & bandwidth usage (which we have to pay for).

**Describe the solution you'd like**

Look like there's a config option to ignore builds:
https://docs.netlify.com/configure-builds/file-based-configuration/#ignore-builds

We could do this against some grep on author? Something like:
```
$ git show -s ea6fcfb0 | grep Author
Author: allcontributors[bot] <46447321+allcontributors[bot]@users.noreply.github.com>
```
Or perhaps it will make the most sense to ignore any PRs that don't touch source files. e.g. if anyone updates a README or a GH workflow, no need to re-build.

minimalsm:
Should create a new issue as we've moved to Gatsby cloud.

1) Is this still an issue with Gatsby cloud?

2) Can we solve it by not allowing their GitHub account to trigger PRs (in the same way that only the ethereum.org core team got deploy previews, to begin with)

PR #2431, opened 2021-02-13
Limit Netlify builds to relevant diffs

yakkomajuri:
## Description

> EDIT: Having put this out I now realize Netlify asks for permission to build, so not sure it even is relevant anymore

Adds a shell script to ignore builds based on what files have been changed (rather than the author). 

The idea here is to provide the foundation for not running builds on certain commits that can of course be changed by the maintainers as you need.

Currently, it makes it so that Netlify will build if:

- Something changed in `/src` **OR**
- `.js` or `.json` files were changed elsewhere

Happy to update the pattern if anyone has suggestions. 

## Related Issue

#1148 

## Further explanation

The first part of the script: 

```
git diff --quiet --exit-code HEAD^ HEAD ./src
```

Will exit with code 1 if files changed in `/src`, triggering a build as per [Netlify's docs](https://docs.netlify.com/configure-builds/file-based-configuration/#ignore-builds).

If it exits with code 0 (no files changed in that dir), it will then run the second part:

```
if [[ $(git diff --name-only HEAD^ HEAD .) == *[.](js|json)* ]]; then exit 1; else exit 0; fi
```

Here we are getting only the names of the files changed elsewhere (since there were no changes in `/src`) and seeing if they match a given pattern, which essentially checks if there's an extension `.js` or `.json` somewhere. Here are some examples to try:

**No build**

```
diff="test.foo\nhey.bar"
if [[ $diff == *[.](js|json)* ]]; then echo build; else echo dont build; fi
```

**Build**

```
diff="test.js\nhey.bar"
if [[ $diff == *[.](js|json)* ]]; then echo build; else echo dont build; fi
```

**Build**

```
diff="test.foo\nhey.json"
if [[ $diff == *[.](js|json)* ]]; then echo build; else echo dont build; fi
```

## Limitations

There is a problem here though. While this should catch all PRs from the allcontributors bot for example, the script currently only checks the latest commit against its parent. In the case of most bots, the parent will be a commit in `dev` anyway, which is fine, but if one submits a PR with multiple commits, this check won't actually be referring to `dev`, but rather to the commit's parent in that branch.

One might then think that `git diff dev` might be better, since it will compare branch HEADs. Problem is that then it will trigger on useless commits if even one commit in that branch matches the pattern against `dev`.

Happy to explore other avenues here but thought the best approach was to open a PR to spark discussion :smiley:

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -16,6 +16,9 @@
   # Directory with the serverless Lambda functions to deploy to AWS.
   functions = "public/lambda"
 
+  # Only build on changes to src/ OR changes to .js and .json files outside src/
+  ignore="git diff --quiet --exit-code HEAD^ HEAD ./src || if [[ $(git diff --name-only HEAD^ HEAD .) == *[.](js|json)* ]]; then exit 1; else exit 0; fi"
+
 # Production context: all deploys from the Production branch set in your site's
 # deploy contexts will inherit these settings.
 # [context.production]
--- a/netlify.toml
+++ b/netlify.toml
@@ -17,7 +17,7 @@
   functions = "public/lambda"
 
   # Only build on changes to src/ OR changes to .js and .json files outside src/
-  ignore="git diff --quiet --exit-code HEAD^ HEAD ./src || if [[ $(git diff --name-only HEAD^ HEAD .) == *[.](js|json)* ]]; then exit 1; else exit 0; fi"
+  ignore="git diff --quiet --exit-code dev ./src || if [[ $(git diff --name-only dev .) == *[.](js|json)* ]]; then exit 1; else exit 0; fi"
 
 # Production context: all deploys from the Production branch set in your site's
 # deploy contexts will inherit these settings.


yakkomajuri:
As we can see, no build on this PR! 🎉

samajammin:
Amazing! Thank you for this @yakkomajuri ❤️ 🙏 

Question on the limitations you mention:
> the script currently only checks the latest commit against its parent

So if a contributor submits a PR with many commits, this logic will only check for the files changes in the latest commit? If so, I could see that being problematic... e.g. someone submits a PR with a bunch of changes, then as their final commit, they update the README. That (if I understand correctly) would cause Netlify to ignore the build for the PR.

yakkomajuri:
@samajammin yes that's right - so the alternative here is:

> One might then think that `git diff dev` might be better, since it will compare branch HEADs. Problem is that then it will trigger on useless commits if even one commit in that branch matches the pattern against dev.

Diffing the branches will give more false positives, whereas the current approach gives more false negatives - but indeed it could make more sense in this case.

yakkomajuri:
@samajammin updated to branch-based diffing, which may lead to false positives but won't cause false negatives!

samajammin:
Awesome, thank you @yakkomajuri for the explanation & the update!

samajammin (commented):
This looks great to me! One more edge case I'm thinking through @yakkomajuri...

What if the branch is `master`? e.g. We merge `dev` into `master` in order to deploy the site. In this case, would builds skip when `master` is the same as `dev`? If so, we may want to add logic (if possible) to always build if the branch is `master`. LMK what you think, maybe I'm over-thinking this 🤔

samajammin:
Hey @yakkomajuri - just checking in on my question above. Thanks again for helping out with this!

yakkomajuri:
Hey @samajammin ! Been quite busy with work but will sort this out over the weekend!

yakkomajuri:
Update here - gave this a quick look. Unsure yet how to check the target branch of the PR when checking out a branch.

samajammin (approved):
Approving & merging! If this breaks master builds, we'll quickly be able to tell 😄 

Thanks again!

PR #2456, opened 2021-02-19
Update index.md

MakeSoundNotWar:
Changed "subsequent" to "preceding" in line 14 to reflect the intended meaning of the sentence. In order to change a block's data, the blocks created BEFORE (preceding) it must be changed as opposed to the ones AFTER (subsequent).

### DIFF
--- a/src/content/developers/docs/intro-to-ethereum/index.md
+++ b/src/content/developers/docs/intro-to-ethereum/index.md
@@ -11,7 +11,7 @@ A blockchain is best described as a public database that is updated and shared a
 
 "Block" refers to the fact that data and state is stored in sequential batches or "blocks". If you send ETH to someone else, the transaction data needs to be added to a block for it to be successful.
 
-"Chain" refers to the fact that each block cryptographically references its parent. A block's data cannot be changed without changing all subsequent blocks, which would require the consensus of the entire network.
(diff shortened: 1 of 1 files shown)


wackerow:
@MakeSoundNotWar Hey, thanks for dropping in =) Not sure about this change though, as altering a block can only be done by then adjusting every block that comes _after_ that block, so those blocks reference the correct parent.

For example:
Blocks: `[0 (genesis)] <- [1] <- [2] <- [3]`
This could represent a simple chain, where `[0]` is the genesis block, and every block after references its parent. If you attempt to alter block `[2]`, this will invalidate `[3]`, not `[0]` or `[1]`. In proof-of-work, this would require re-mining every block from that point forward to create a new longest/heaviest chain that includes those earlier block changes. 

This is why the most recent block has a greater chance of being reversed. Also why as more and more "confirmations" stack up, it because that much more difficult the change anything, since you'd have to re-work every block since then too. 

https://andersbrownworth.com/blockchain/blockchain <-- This is a great demo website that demonstrates this concept. 

Hope this helps, going to close for now =)

MakeSoundNotWar:
@wackerow Thank you for clarifying that. I was mistaken in my understanding of the validation process.

wackerow:
@MakeSoundNotWar Not a problem! Appreciate your intent, glad I could help clarify =)

PR #2471, opened 2021-02-23
Allows hash linking alongside queryParams [Fixes #2465]

### LINKED ISSUE #2465: Pre-filter and anchor links
As a content creator I want to link users to a pre-filtered section that exists halfway down a page so that I can link users to the exact content I want them to see.

For example, on a page that discusses NFTs I should be able to link users to the `/#explore` section of the dapps page with the collectibles option pre-filtered. This way users can see the NFT dapps immediately upon navigating. 

Current behaviour pre-filters the page but gives the user no context as to what is pre-filtered.

wackerow:
## Description
- If hash link provided, will navigate to it
- If "category" queryParam provided, this will be set before navigating to any hash link, and will automatically scroll to "explore" section to provide context for queryParam if no overriding hash link is provided

examples:
`/en/dapps` -> No pre-filtering (defaults to Finance), user stays at top of page
`/en/dapps#explore` -> No pre-filtering (defaults to Finance), user is taken to `explore` section
`/en/dapps?category=gaming` ->  Pre-filters to Gaming, user is taken to `explore` section
`/en/dapps?category=gaming#what-are-dapps` -> Pre-filters to Gaming, user is taken to `what-are-dapps` section

## Related Issue #2465 

@ryancreatescopy I defaulted this to an automatic scroll (no animation), but this can be switched to a "smooth" scroll if you wanted to consider adding some more context for the user (ie, they go to `/en/dapps?category=gaming` and it animates the scroll to that section so the user knows where they are, instead of it just jumping there). Personally prefer the `auto` option over `scroll`.

### DIFF
--- a/src/pages-conditional/dapps.js
+++ b/src/pages-conditional/dapps.js
@@ -367,7 +367,7 @@ const DappsPage = ({ data, location }) => {
     } else if (window && queryParamCategories && explore.current) {
       window.scrollTo({
         top: explore.current.offsetTop - 76,
-        behavior: "auto",
+        behavior: "smooth",
       })
     }
   }, [location.search])
--- a/src/pages-conditional/dapps.js
+++ b/src/pages-conditional/dapps.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from "react"
+import React, { useRef, useState, useEffect } from "react"
 import styled from "styled-components"
 import Img from "gatsby-image"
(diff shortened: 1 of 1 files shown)


ryancreatescopy (approved):
Looks awesome. I changed it to a smooth scroll though as I think it's valuable to inform users that there is a more complete dapps page if people want it.

samajammin (approved):
👍

PR #2501, opened 2021-02-27
Remove "inspired by Python" in Solidity's description

fubuloubu:
## Description
Every time I look at this page, I see "Solidity is inspired by C++, *Python* and JavaScript" and think to myself "where's the Python influence?"

Thinking about it further, it also seems kind of weird to have that there, and then mention another language that is also "Pythonic" (aka Vyper) right below it. It really just seems confusing to me. "Which language should I try if I really like Python?", the natural answer should be Vyper I would think.

Feel free to accept or reject this PR, but I just finally decided to do something about my cognitive dissonance.

### DIFF
--- a/src/content/developers/docs/smart-contracts/languages/index.md
+++ b/src/content/developers/docs/smart-contracts/languages/index.md
@@ -20,7 +20,7 @@ Previous knowledge of programming languages, especially of JavaScript or Python,
 
 ## Solidity {#solidity}
 
-- Influenced by C++, Python and JavaScript.
+- Influenced by C++ and JavaScript.
 - Statically typed (the type of a variable is known at compile time).
 - Supports:
   - Inheritance (you can extend other contracts).


ryancreatescopy:
@fubuloubu thanks for making this PR. I believe this was originally in the content because this is how Solidity describes themselves (see second paragraph here: https://docs.soliditylang.org/en/latest/) However I agree that this is confusing as python is a great differentiator from vyper.

ryancreatescopy (approved):

ryancreatescopy:
@all-contributors please add @fubuloubu for content.

leonardoalt:
Considering the discussions in https://github.com/ethereum/solidity/pull/3195#issuecomment-344264124 and https://github.com/ethereum/solidity/pull/11021#issuecomment-787814267, I think this was merged prematurely.

fubuloubu:
@leonardoalt I think for the context of this webpage, it's more confusing to leave it in, even if you see it as more technically correct. We can discuss the change in the Solidity docs more, but I think not having every single language as "derived from Python" in this document will make it clearer to casual users which language has the _feel_ they're looking for, and don't necessarily care about the technical specifics of how the language's inheritance structure works or how the _idea_ of modifiers is similar to a construct in that language (which works in a much different way).

leonardoalt:
Well, it said "influenced" and not "derived", and that's much more subtle than what the basic syntax looks like, which seems to be what you refer by "the feel they're looking for". If removing Python because of `def` and indentation makes sense, then removing JS because of `var`, `let`, scoping rules, etc etc makes even more sense.

fubuloubu:
> @fubuloubu Well, it said "influenced" and not "derived", and that's much more subtle than what the basic syntax looks like, which seems to be what you refer by "the feel they're looking for". If removing Python because of `def` and indentation makes sense, then removing JS because of `var`, `let`, scoping rules, etc etc makes even more sense.

Most people comfortable with Javascript find Solidity more comfortable. The point of this website is to be a resource to newcomers, and pointing them at Solidity if they are looking for a Javascript-like language is the right way to go. Pointing them at Vyper if they really enjoy Python is also the right way to go.

Issue #2515, opened 2021-03-01
Remove the partiallyActive behaviour on Navigation & Footer links


ipapandinas:
**Describe the bug**
Navigation and Footer links are actived if it partially matches the current page route.

**To Reproduce**

1. Go to the [Private Ethereum]('https://ethereum.org/en/enterprise/private-ethereum/') page
2. Open the `Entreprise` dropdown menu or scroll down to footer links
3. Noting the unexpected behaviour

**Expected behavior**
Only the current page link must be active. In the previous exemple, only `Private Ethereum` must be orange.

**Screenshot**

<img width="1210" alt="partiallyActiveLink" src="https://user-images.githubusercontent.com/26460174/109545601-594a7880-7ac9-11eb-82d5-7ae7074945e2.png">

samajammin:
Nice catch @ipapandinas! Thanks for volunteering for this 😄. 

My hunch is that the `isPartiallyActive` props should be set to `false` for those links:
https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Link.js#L65

Let us know if you have questions, otherwise feel free to knock this out!

ipapandinas:
Agreed ! I fixed it on this PR #2523

wackerow:
@ipapandinas thanks again!

PR #2523, opened 2021-03-02
fix: isPartiallyActive set to false for NavDropdown & FooterLink

ipapandinas:
## Description

Removed of the partiallyActive behaviour on Navigation Dropdown & Footer links by setting `isPartiallyActive` props to false for those links.

## Related Issue

#2515


samajammin (approved):
Thanks for this!

Issue #2304, opened 2021-01-23
Automatically add explicit markdown header IDs 


samajammin:
**Is your feature request related to a problem? Please describe.**

Often when contributors add new markdown files, they forget to add explicit header IDs (e.g. see @wackerow's review of #2194).

**Describe the solution you'd like**

We have a script to automatically add these explicit header IDs when they're not present:
https://github.com/ethereum/ethereum-org-website/blob/dev/src/scripts/generate-heading-ids.js

It can be executed from the project root, e.g. 
```
$ yarn generate-heading-ids developers/tutorials/create-and-deploy-a-defi-app
```

Let's explore the potential to add this as a pre-commit hook (can we access the filename?) so that before any markdown files are committed, we run the `generate-heading-ids` script on the file.

**Describe alternatives you've considered**

- Add this as an instruction to the README

**Additional context**

Why add add explicit header IDs: #272
TL;DR it's for anchor link consistency across translations

wackerow:
Encountered some issues trying to implement this in an automatic way.
Using the pre-commit hook causes issues with staging the potential changes that are made when running a script to clean all of these up. It gets messy trying to stage only the new changes from the script with only what the user had previously staged, without potentially overwriting their stash in some cases. 
Perhaps a GH action / bot that could scan for changes to `/contents/` (minus `content/translations`) folder and automatically run a script then and submit a bot PR to fix?

samajammin:
This now has added importance given that we rely on header IDs to determine if a translated file is outdated our not (#4441).

corwintines:
Going to close with above PR

corwintines:
Resurfacing in https://github.com/ethereum/ethereum-org-website/issues/10773 to refresh this issue

PR #2550, opened 2021-03-06
Adds missing custom headers

wackerow:
## Description
- More custom header IDs missing from recent changes


ryancreatescopy (approved):

samajammin (approved):
Nice, thanks for catching @wackerow!

Another good example of why we should do this: https://github.com/ethereum/ethereum-org-website/issues/2304 - feel free to investigate that if you'd like! 😄

PR #2565, opened 2021-03-08
Fix mobile side navigation rendering issue [Fixes #2301]

### LINKED ISSUE #2301: Repeat sidenav on docs page
**Describe the bug**

Kinda hard to replicate this. I think it's a return of the caching issue as it seems to sometimes appear and other times not. Refreshing can sometimes cause it.

First discovered on: https://ethereum.org/en/developers/docs/nodes-and-clients/run-a-node/

**Screenshots**

image

image

random arrow halfway down the page

image

image

okdonga:
## Description

 

* When you access a page under ```/doc``` on **light mode**, mobile navigation is rendered in a broken way (see screenshot below)  _This only happens on light mode, (works fine on dark mode)_  
Screen Shot 2021-03-08 at 11 39 16 AM

 

### Why is this happening  

After some research into this issue, it appears to behave such due to rehydration - where ```shouldShowBanner``` is ```false``` in server side, but ```true``` on client side. 
https://github.com/ethereum/ethereum-org-website/blob/41744cc49ebc9b6be762f85912f9378fc4be1dd6/src/components/Layout.js#L93
I'm suggesting to hold off rendering of ```SideNavMobile``` component until we know the ```path``` of the page being rendered. I wrapped this logic using ```useEffect``` hook to check when ```path``` changes. 

### How to reproduce 
* Turn on light mode
* Open up any page under /Docs 
* Refresh

## Related Issue
#2301

### DIFF
--- a/src/components/Layout.js
+++ b/src/components/Layout.js
@@ -58,7 +58,7 @@ const StyledBannerNotification = styled(BannerNotification)`
 // TODO `Layout` renders twice on page load - why?
 const Layout = (props) => {
   const [isDarkTheme, setIsDarkTheme] = useState(false)
-
+  const [shouldShowSideNav, setShouldShowSideNav] = useState(false)
   // set isDarkTheme based on browser/app user preferences
   useEffect(() => {
     if (localStorage && localStorage.getItem("dark-theme") !== null) {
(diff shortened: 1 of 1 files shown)


samajammin (approved):
Looks great - thanks for this!

PR #2614, opened 2021-03-11
Tutorials filter by language [Fixes #2584]

### LINKED ISSUE #2584: Tutorials index page doesn't filter by language
**Describe the bug**

We recently added our first translations of tutorials (#2531) 🎉  but we forgot we now need to filter out language on the tutorials page.

**To Reproduce**

1. Go to https://ethereum.org/en/developers/tutorials/
2. Search for "The Graph: Remedierea interogării datelor Web3"

**Expected behavior**

- https://ethereum.org/en/developers/tutorials/ should only show English tutorials
- https://ethereum.org/ro/developers/tutorials/ should only show Romanian tutorials
- etc

**Screenshots**

Image 2021-03-09 at 7 24 12 AM

All
All

wackerow:
## Description
- Filters `allTutorials` by lang locale when going to `/{lang}/developers/tutorials` 
- Added work-around to create `allTags` object from the above filtered tutorials list, as pulling via graphql did not allow differentiating tags coming from tutorials of foreign languages.

## Related Issue
#2584

### DIFF
--- a/src/pages/developers/tutorials.js
+++ b/src/pages/developers/tutorials.js
@@ -16,7 +16,7 @@ import TutorialTags from "../../components/TutorialTags"
 import Emoji from "../../components/Emoji"
 import { Page, ButtonSecondary } from "../../components/SharedStyledComponents"
 
-import { getLocaleTimestamp } from "../../utils/time"
+import { getLocaleTimestamp, INVALID_DATETIME } from "../../utils/time"
 
 const SubSlogan = styled.p`
   font-size: 20px;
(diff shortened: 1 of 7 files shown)


ryancreatescopy:
To confirm this ensures that we don't just display all tutorials in all languages all the time?

Although when looking at it I discovered this error:

image

Not sure how this typo slipped through the net. Can we change to "There are no tutorials with all your chosen tags yet"

Thought it might be easy to slip this change into this PR.

src/pages/developers/tutorials.js
```diff
@@ -189,8 +189,10 @@ const ModalTitle = styled.h2`
 const TutorialsPage = ({ data }) => {
   const intl = useIntl()
 
-  const allTutorials = data.allTutorials.nodes.map((tutorial) => {
-    return {
+  // Filter tutorials by language and map to object
+  const allTutorials = data.allTutorials.nodes
+    .filter((tutorial) => tutorial.frontmatter.lang === intl.locale)
```
samajammin, line 194: I think the issue here is that `tutorial.frontmatter.lang` is still `en` for the English pages we copy over.

In gatsby-node.js you can see we don't overwrite that frontmatter prop: https://github.com/ethereum/ethereum-org-website/blob/dev/gatsby-node.js#L145
Instead we add the `language` property in the `intl` object.

Let's update this filter to use that (I believe you can grab it from the `pageContext` prop.

wackerow:
@samajammin Take a look when you can. Added a property to track which languages have tutorials translated. Tutorials page will now display all the English tutorials on the `/en/` path, and also any lang path that does not have any translated (all but Romanian right now). The `/ro/developers/tutorials` path will show only the Romanian tutorials, and filter out others.

also
- Fixed string with poor grammar when no options available for selected tags. 
- Hides the "Published" snippet when the date provided in the markdown frontmatter is invalid.

samajammin:
@wackerow nice! I've removed the `hasTutorials` from the translations metadata & updated the `hasTutorials` util function to infer this from the content version (since v2.2 is the version that includes tutorials).

samajammin (approved):

wackerow:
Nice! I like that much better. I didn't realize 2.2 was the official version for tutorials, thanks for fixing that.

Issue #2583, opened 2021-03-09
Vertical height bug – use-case and eth2 template 


ryancreatescopy:
**Describe the bug**

On certain screen sizes, the left-hand column with the table of contents isn't scrollable. If a list goes beyond the view port it's impossible to navigate to the sections below the viewport until the main content has scrolled to the footer. 

**Expected behavior**

If the table of contents doesn't fit due to vertical height, it should be scrollable so that users can select all items in the table of contents.

**Screenshots**

https://user-images.githubusercontent.com/40891631/110489524-5f52e180-80e7-11eb-91ca-7459453a38e5.mov

NilsKaden:
@wackerow did you start working on this already? I can have a look at this if you like :)

wackerow:
@NilsKaden Thanks for offering to help on this! Please don't hesitate to ask us any questions along the way 🙂 Assigning you (just let us know if you change your mind in the future)

PR #2645, opened 2021-03-16
add css rule to Aside component to address vertical height bug #2583…

mic0des:
Added 'overflow-y: auto' CSS rule to the Styled Aside component in Eth2TableOfContents.js to attempt to address the vertical height bug flagged in #2583 

## Description

Unsure if this is what you guys had in mind. Saw this open issue a few days ago and decided to give it a go. Rather simple fix, but on sufficiently short screens, a scrollbar is now visible on the left-hand Aside component, and disappears on taller viewports. I only saw today prior to this PR that it had been assigned to @NilsKaden ... apologies if we doubled work, or if someone else had a different/better fix.

https://user-images.githubusercontent.com/4992682/111256600-d01a6200-85ef-11eb-9d40-fb1e9df25de0.mp4

## Related Issue

#2583

### DIFF
--- a/src/components/Eth2TableOfContents.js
+++ b/src/components/Eth2TableOfContents.js
@@ -9,7 +9,7 @@ const Aside = styled.aside`
   padding: 0rem;
   text-align: right;
   margin-bottom: 2rem;
-  overflow: scroll;
+  overflow-y: auto;
 `
 
 const OuterList = styled(motion.ul)`
--- a/src/components/Eth2TableOfContents.js
+++ b/src/components/Eth2TableOfContents.js
@@ -9,6 +9,7 @@ const Aside = styled.aside`
   padding: 0rem;
   text-align: right;
   margin-bottom: 2rem;
+  overflow: scroll;
 `
 
 const OuterList = styled(motion.ul)`


wackerow (changes requested):
@mic0des Hey, thanks for helping out with this =) as you alluded to, just try to comment on an issue if you'd like to be assigned to it, just to try and limit stepping on each others toes. 

I think this is a good start, but wondering if we may want to instead apply this styling property to the `InfoColumn` component in the `use-cases.js` template, which would cause the whole panel to slide instead of just the links.
Additionally, personally feel we should remove the scrollbar for this scrollable region, ie
```
  overflow-y: auto;
  scrollbar-width: none;
  ::-webkit-scrollbar {
    display: none;
  }
```
@ryancreatescopy thoughts on this?

(@NilsKaden ok if @mic0des takes this one? we'll try to get you involved on another issue if you're still down.. feel free to take a look and comment 👍)

samajammin (approved):

samajammin:
Thanks for the input @wackerow - as discussed, let's just merge this as-is.

Thanks for this @mic0des!

samajammin:
@all-contributors please add @mic0des for code.

mic0des:
Noted @wackerow - will comment on an issue before jumping in next time now that I know! Thanks guys.

PR #2649, opened 2021-03-16
Added focus state to interactive header buttons

### LINKED ISSUE #2648: improve keyboard accessibility (a11y)
**Describe the bug**

I have noticed a few issues when trying to navigate this website using only a keyboard. 

<s>1. A few interactive elements (such as the theme toggle button) are either not focusable or do not have any focus styles applied</s> Fixed in #2649

2. The mobile navigation doesn't have `display: none` applied and therefor is rendered in the accessibility tree. This means a keyboard user has to tab through all of these hidden nav items before reaching any content.

**To Reproduce**

1. Go to `https://ethereum.org/en/`
2. Try navigating the site with just using a keyboard

**Expected behavior**

- Each interactive element on the page should have a clear focus state.
- The mobile menu should not appear in the accessibility tree when on desktop, and the desktop menu shouldn't be in the accessibility tree on mobile.

jordanoverbye:
## Description

- Created a `NakedButton` component which has the accessibility advantages of a button with none of the styling.
- Used `NakedButton` to get a focus state on the theme toggle button
- Used `NakedButton` instead of a span which means we can remove `tabIndex`, `role` and `onMouseDown` from `NavMobileButton` 

## Related Issue

Fixes some of the issues mentioned in #2648


src/components/Nav/index.js
```diff
@@ -397,19 +396,13 @@ const Nav = ({ handleThemeChange, isDarkTheme, path }) => {
           <MobileIcons>
             <NavMobileButton
               onClick={() => handleMenuToggle("search")}
-              onKeyDown={() => handleMenuToggle("search")}
```
jordanoverbye, line 400: Using a `button` element means we get all of this for free 🎉

wackerow (approved):
@jordanoverbye Welcome! Good catch here, thanks for the PR. This looks good to me from the standpoint of the Dark/Light mode button. Per the issue you posted, I agree fixing having to tab through all the hidden menu items, etc.
I'll pull this PR in now, but will keep the original issue open. Feel free to expand on this idea in a separate PR if you'd like =) Thanks again

PR #2655, opened 2021-03-16
Changed TPS count from 10 to 15. This is the current average TPS

a1irahman:


a1irahman:
I think changing it to "10-15" would also make sense. This would make the documentation as accurate as possible.

a1irahman:
> I think changing it to "10-15" would also make sense. This would make the documentation as accurate as possible.

After further review, I think this makes the most sense. Let me know if I should begin another request for this, rather than append this.

wackerow (changes requested):
@a1irahman Hey, thanks for helping out =)
Taking a look at some [charts](https://blockchair.com/ethereum/charts/transactions-per-second?interval=3m) it does look like saying `10` is a bit out of date... Given that it changes over time, I agree that saying `10-15` keeps it vague while conveying the point that it's not a lot. Hopefully soon we can eliminate the whole section of "network congestion" =)

wackerow:
@a1irahman Easiest to just append this branch with another commit and push it up 👍 thanks!

a1irahman:
@wackerow No problem! I think this project is really helpful, so I am happy to pitch in. I actually based my original request on that exact chart. I agree, I can't wait for network congestion to no longer be a problem. Also, I went ahead and pushed the new version 👍🏼

wackerow (approved):
👍

wackerow:
@a1irahman Thanks again! Feel welcome to submit or comment on other issues if you want to be assigned to anything =)

wackerow:
@all-contributors please add @a1irahman for content

a1irahman:
@wackerow Of Course! I will keep an eye out for opportunities to help in the future 🔥

PR #2675, opened 2021-03-18
improve keyboard accessibility (a11y)

jordanoverbye:
## Description

The mobile navigation isn't hidden correctly and therefor is rendered in the accessibility tree. This means a keyboard user has to tab through all of these hidden nav items before reaching any content.

## Related Issue

Issue #2648 

Related PR: #2649

### DIFF
--- a/src/components/Nav/Mobile.js
+++ b/src/components/Nav/Mobile.js
@@ -3,6 +3,7 @@ import styled from "styled-components"
 import { useIntl } from "gatsby-plugin-intl"
 import { motion } from "framer-motion"
 
+import NakedButton from "../NakedButton"
 import Translation from "../Translation"
 import Icon from "../Icon"
 import Link from "../Link"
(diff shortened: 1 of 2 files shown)


src/components/Nav/index.js
```diff
@@ -393,24 +370,8 @@ const Nav = ({ handleThemeChange, isDarkTheme, path }) => {
             toggleTheme={handleThemeChange}
             linkSections={mobileLinkSections}
           />
-          <MobileIcons>
```
jordanoverbye, line 396: I've moved all of this as it felt weird to have some mobile components here and some in `src/components/Nav/Mobile.js`. I think they should just live in the same place.

src/components/Nav/Mobile.js
```diff
@@ -11,6 +12,25 @@ import Emoji from "../Emoji"
 import { NavLink } from "../../components/SharedStyledComponents"
 import { translateMessageId } from "../../utils/translations"
 
+const Container = styled.div`
+  display: none;
```
jordanoverbye, line 16: Adding `display: none` will hide all mobile elements from the a11y tree on desktop

src/components/Nav/index.js
wackerow: Thanks for doing that =)

wackerow (approved):
@jordanoverbye LGTM! Thanks for putting this together. Trying it out it definitely flows better when using just the keyboard. 👍

PR #2695, opened 2021-03-21
Grants updates

samajammin:
## Description

- Update static templates to show GH edit button
- Link to Grants page from Community page

## Related Issue
None

### DIFF
--- a/src/templates/static.js
+++ b/src/templates/static.js
@@ -131,7 +131,6 @@ const StaticPage = ({ data: { siteData, mdx }, pageContext }) => {
   const intl = useIntl()
   const isRightToLeft = isLangRightToLeft(intl.locale)
 
-  console.log({ mdx })
 
   const lastUpdatedDate = mdx.parent.fields
     ? mdx.parent.fields.gitLogLatestDate
--- a/src/templates/static.js
+++ b/src/templates/static.js
@@ -95,7 +95,7 @@ const HR = styled.hr`
   background: ${(props) => props.theme.colors.border};
 `
 
-// Passing components to MDXProvider allows use across all .md/.mdx files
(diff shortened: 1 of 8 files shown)


samajammin:
@ryancreatescopy @wackerow curious to hear your thoughts - should we make ALL static pages show the GH edit button? Or show that conditionally (e.g. via a frontmatter boolean)? I feel like it's appropriate for the Grants & Community pages. Not sure about Enterprise / all other pages that use the static template.

src/templates/static.js
```diff
@@ -127,16 +127,21 @@ const components = {
...
 }
 
-const StaticPage = ({ data: { mdx } }) => {
+const StaticPage = ({ data: { siteData, mdx }, pageContext }) => {
   const intl = useIntl()
   const isRightToLeft = isLangRightToLeft(intl.locale)
-  const tocItems = mdx.tableOfContents.items
 
-  // TODO some `gitLogLatestDate` are `null` - why?
+  console.log({ mdx })
```
wackerow, line 134: (Intentional?)

wackerow:
@samajammin Agree that ideally it's hidden on enterprise pages. Could make it default to being shown, unless the frontmatter explicitly specifies false.

ryancreatescopy:
I think we can extend to all pages using static template right now. We don't need to approve any PRs for editing enterprise pages if they're controversial changes. But it makes sense that someone should be able to easily fix a typo if they see one.

src/templates/static.js
samajammin: Removing the comment, yes. Adding a console.log, no 😳

samajammin, line 134:
```suggestion
```

PR #2731, opened 2021-03-26
Animates hamburger menu into ETH glyph

wackerow:
## Description
- Replaced hamburger menu icon with SVG that defaults to greek "Xi" shape when closed, and morphs into an ETH glyph when opened. 
- Adjusted z-index so same glyph acts as close button when menu opens, and existing close button was removed.
- Search panel will lay above all of this and function as it did. 

cc: @ryancreatescopy check it out, lemme know what you think

### DIFF
--- a/src/components/Nav/Mobile.js
+++ b/src/components/Nav/Mobile.js
@@ -41,13 +41,13 @@ const MobileModal = styled(motion.div)`
 `
 
 const mobileModalVariants = {
-  open: { display: "block" },
-  closed: { display: "none" },
+  open: { display: "block", opacity: 1 },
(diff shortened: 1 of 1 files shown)


ryancreatescopy (commented):
I think this looks incredible! My only tiny concern is that it's not clear you can press that to close the menu. Would it make sense to have it transition into the eth symbol as the menu slides in and it turns back into a hamburger shortly afterwards? That might look odd but worth trying out IMO

wackerow:
@ryancreatescopy Curious your thoughts on latest

ryancreatescopy (approved):

Issue #2705, opened 2021-03-22
Explicit language path in links


samajammin:
**Describe the bug**

We have cases where links to pages have an explicit language path, e.g. `[ethereum.org/learn](/en/learn/)`.

We should generally avoid this (except if only the English page exists) as our `<Link/>` component automatically prepends the appropriate language path:
https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Link.js#L131

**To Reproduce**

1. Search codebase for common paths, e.g. `/en/learn/`
2. View results

**Expected behavior**

Remove explicit language path from these links

minimalsm:
This issue was resolved by #2735

ryancreatescopy:
Closing for reason above.

PR #2735, opened 2021-03-26
Remove explicit language in path for links

kziechmann:
## Description

- Removing the language ("en" in this case) when explicitly included in the path for the "to" prop of any instance of the Link component
- Also including translations for the 404 error page when available

## Related Issue

Explicit language path in links #2705
https://github.com/ethereum/ethereum-org-website/issues/2705#issue-838081344


src/components/WalletCompare.js
```diff
@@ -501,7 +501,7 @@ const WalletCompare = ({ location }) => {
           <p>
             <em>
               <Translation id="page-find-wallet-not-endorsements" />{" "}
-              <Link to="/en/contributing/adding-products/">
```
wackerow, line 504: This hard-coding of `/en` is intentional here since the contributing page is not translated.

The same is true for any of the pages flagged as `isLegal` in `gatsby-node`:
https://github.com/ethereum/ethereum-org-website/blob/0f71a8d2d6beae0c02224310c29bddf594b9921e/gatsby-node.js#L137
```
const isLegal =
      slug.includes(`/cookie-policy/`) ||
      slug.includes(`/privacy-policy/`) ||
      slug.includes(`/terms-of-use/`) ||
      slug.includes(`/contributing/`)
```

src/pages/languages.js
```diff
@@ -62,7 +62,7 @@ const LanguagesPage = () => {
           </p>
           <p>
             <Translation id="page-languages-interested" />{" "}
-            <Link to="/en/contributing/translation-program/">
```
wackerow, line 65: (same as above)

src/pages/languages.js
```diff
@@ -83,7 +83,7 @@ const LanguagesPage = () => {
           </h2>
           <p>
             <Translation id="page-languages-want-more-paragraph" />{" "}
-            <Link to="/en/contributing/translation-program/">
```
wackerow, line 86: (same as above)

src/pages/404.js
```diff
@@ -1,21 +1,27 @@
...
+        <Translation id="page-404-not-found" />
+      </h1>
       <p>
-        Try using search to find what you're looking for or{" "}
-        <Link to="/en/">return home</Link>.
+        <Translation id="page-404-copy" />
+      </p>
+      <p>
+        <Link to="/">
+          <Translation id="home" />
```
wackerow, line 23: If we're going to use these translation ID's, we need to add corresponding strings to the translation JSON file. In this case, I would recommend adding them to the `src/intl/en/common.json` file. 
ie:
```
"page-404-not-found": "Resource not found.",
"page-404-copy": "Try using search to find what you're looking for.",
"home": "Return home",
```

On this note, @samajammin were we intentionally not translating this page for any reason?

wackerow (changes requested):
@kziechmann hey! Thanks for joining in to help out =) Appreciate this, just had a few notes for ya. @samajammin Let me know if you're okay with teeing the 404 page up for translation. If so, @kziechmann if you could just make the changes suggested and we'll get it merged!

wackerow:
@kziechmann In reference to #2705, if you search for hard-coded English paths within the markdown content, you'll find about 294 results. Keep in mind the /en/contributing/* links should be left as is, since those pages are not built for non-english languages.

Try searching the codebase for the string `](/en/` which will catch all markdown links that begin with `/en/`:
image

Remove the `/en` from each of these (find replace `](/en/` with `](/` within each file, making sure to leave the `/en/contributing/*` links, i.e. these:
image

Hope this helps clarify =) Let us know if you have questions

src/pages/404.js
kziechmann: It looks like there are already translations for "page-404-not-found" and "page-404-copy" in 25 of the files though there are only 4 files with translations for "home" and no 'return home' as a separate translation. Wondering if this provides value or if its better to wait for complete translations.

wackerow: Hm, good call. I spot checked it using Romanian, and that one happened to not have those IDs.

@samajammin Looking close, it appears it's the version 2.0, 2.2 and 1.0 pages that don't have these strings anymore. English (v2.4) has them, and v1.1 have them still. The `en/common.json` file for example has 202 items, yet the `ro/common.json` and `es/common.json` files only have 192 items (of the ones missing are these 404 strings).
Any idea why that is?

samajammin: So it's been a bit since I looked into this but I'm not sure translating the 404 page is actually possible. Since the requested page doesn't exist, the application isn't aware what the language is. I don't believe there's an `intl` object in the page's context. Perhaps we could add some logic into the component to inspect the `pathname` & lookup the appropriate string based on that? Might be best to have this as a standalone issue/PR.

.env.example
```diff
@@ -1,11 +0,0 @@
-# rename this file to .env and supply the values listed below
```
samajammin, line 1: Let's not delete this file.

samajammin (changes requested):
@kziechmann thanks for this 🙌 !

Only a couple small change requests:
- Let's not delete the .env.example file
- For the sake of this PR, let's not attempt to translate the 404 page (see my comment, I'm not sure if it's possible without digging into it, we should address that in a separate PR & test it out)

wackerow (approved):
@kziechmann Looks good! Just reverted the page-404 changes, which we can handle in a separate PR. Thanks for the help on the rest!

wackerow:
@all-contributors please add @kziechmann for bug fixes

PR #2741, opened 2021-03-28
Fixing error in the re-entrancy section

robertdosa:
## Description

The name of the mapping in the Victim contract is balances, but in the cycle description when the Attacker calls Victim.withdraw() the Victim should read balances[msg.sender] instead of balanceOf[msg.sender]


wackerow (approved):
@robertdosa Good catch! Thanks for the PR, looks good to me =)

wackerow:
@all-contributors please add @robertdosa for content

PR #2772, opened 2021-04-01
Add patch request type and ide folders to gitignore

timhc22:
## Description

I have added a patch type issue template, as there seems to be a gap - in the issue types - for minor tweaks which aren't really bugs and aren't really features. (e.g. the other commit here which was adding IDE generated folders to .gitignore).


.gitignore
```diff
@@ -76,3 +76,7 @@ yarn-error.log
 src/data/contributors.json
 # These files are generated by `yarn merge-translations` command
 src/intl/*.json
+
+# IDE
+.idea/
+.viscode/
```
wackerow, line 82: `.vscode/`

wackerow (changes requested):
@timhc22  Thanks for suggesting this... when you say "Suggest a patch for this project" is that in reference to the whole site? I get trimming out some of the technical stuff for simple reports, but just wondering if it's worth separating this from bug reports since the difference seems very nuanced. Perhaps reworking the bug template could be a more optimal approach. @samajammin any thoughts on this?

.gitignore
timhc22: Hey, my logic is that: all bugs are patches, not all patches are bugs. https://semver.org/
Reworking the bug template would be a better solution imo.
I'd say do what works best for the team.
Best regards

wackerow: (@timhc22 note line 82 should be `.vscode/`)
(@samajammin bump 👆 any thoughts on this issue type?)

timhc22: 🧐 ah yes! should be updated now

timhc22:
@samajammin incidentally, it was decision fatigue which led me to opening this issue in the first place (I wasn't sure what to submit as because it definitely wasn't a bug), however, if 'minor patch' issues like mine are few and far between, it is more important to reduce overall decision fatigue, and so let's go with your suggestion!

samajammin:
Thanks for the insight. Remind me @timhc22 - what was the original issue you're referring to?

timhc22:
'add ide folders to gitignore'

I guess it could be considered a bug that they weren't there, but I'd say (like you said) more a 'minor' patch, but you are right that a template for 'major patch' would also be needed if we have that, and that is too much..

PR #2798, opened 2021-04-07
ensure nimbus in list of eth2 bounties

djrtwo:
## Description

Nimbus was added to eth2 bounties  a while back but was not updated in all the copy.

Added to `en` copy. This should be fixed for other languages as well


wackerow (approved):

wackerow:
Thanks @djrtwo!... Added "Nimbus" to the translated strings for languages supporting this page... 
Since I don't actually know these languages, I just adjusted/updated the four client names manually, but the other adjustments you made to the sentence were not translated. I can push this change to Crowdin for official translation and get those in a separate PR when ready.

samajammin:
Groovy - thanks @djrtwo & @wackerow!

Issue #2802, opened 2021-04-07
Builds breaking in Netlify


samajammin:
**Describe the bug**

Catch-all issue to track various ways builds have broken. We're seeing recent build failures:

1. Runtime errors

```
Build exceeded maximum allowed runtime
```

Here's one of the logs:
https://app.netlify.com/sites/ethereumorg/deploys/606d1016cbef7d00076d6ef9

2. GraphQL failures

```
1:57:11 PM: Url path: /developers/local-environment/
1:57:11 PM: Plugin: none
1:57:11 PM: 
1:57:11 PM: 
1:57:11 PM:   FetchError: request to https://api.github.com/graphql failed, reason: Client n  etwork socket disconnected before secure TLS connection was established
1:57:11 PM:   
1:57:11 PM:   - index.js:1461 ClientRequest.<anonymous>
1:57:11 PM:     [repo]/[node-fetch]/lib/index.js:1461:11
1:57:11 PM:   
1:57:11 PM:   - destroy.js:92 emitErrorNT
1:57:11 PM:     internal/streams/destroy.js:92:8
1:57:11 PM:   
1:57:11 PM:   - destroy.js:60 emitErrorAndCloseNT
1:57:11 PM:     internal/streams/destroy.js:60:3
1:57:11 PM:   
1:57:11 PM:   - task_queues.js:84 processTicksAndRejections
1:57:11 PM:     internal/process/task_queues.js:84:21
1:57:11 PM:   
1:57:11 PM: 
1:57:11 PM: not finished run page queries - 2124.630s
1:57:11 PM: error Command failed with exit code 1.
```

**Expected behavior**

Let's explore ways to optimize build times:
https://answers.netlify.com/t/support-guide-how-can-i-optimize-my-netlify-build-time/3907

samajammin:
Note to self: one useful approach here may be to emulate Netlify's buildbot locally:
https://github.com/netlify/build-image

samajammin:
Hmmmmmm
```
Hi @samajammin,

We noticed that, at April 5th, 2021 at 23:05 (UTC) your authentication token (ethereum-org-website, with a user-agent header of axios/0.21.1) accessed the GitHub API, the token you used has an outdated format.

In order to provide additional security benefits to all our customers, we recently updated the format of our API authentication tokens. Please update your authentication tokens to take advantage of these benefits, by visiting https://github.com/settings/tokens.

[...]

PR #2811, opened 2021-04-09
Remove gatsby-source-graphql

samajammin:
## Description

Attempt to solve various Netlify build issues...
- Remove `gatsby-source-graphql`, which queried Github API at build
- Refactor ProductCard to fetch Github API on client

This isn't as ideal from an SEO or performance perspective but hopefully this will solve our woes around failing builds & API limit issues.

## Related Issue
#2802


samajammin:
Hmmm @wackerow one issue here re: local development - now all client API requests will fail (401 status) if running the app with `yarn start` vs. `netlify dev`, as the env won't have access to `GATSBY_GITHUB_TOKEN_READ_ONLY` 🤔

wackerow (approved):
working with both `netlify dev` and `yarn start` locally

PR #2816, opened 2021-04-09
Add styling to links where the language is explicitly stated

### LINKED ISSUE #2815: Non-INTL Footer Links are not styled properly
**Describe the bug**

Footer links which are not react-intl links (i.e they have the prefix /en/.....) not being styled properly. They should be highlighted in a different colour if the url equals the current path.

Links effected:

- Contributing
-  Language Support
-  Privacy policy
-  Terms of Use
-  Cookie Policy

**To Reproduce**

1. Go to https://ethereum.org/en/contributing/
2. Scroll down to footer
4. See unstyled link

**Expected behavior**

The current page should be a different colour to the other links.

**Screenshots**

### See link bottom right
Screenshot 2021-04-09 at 10 21 42

### Expected behaviour example

Screenshot 2021-04-09 at 10 22 38

minimalsm:
## Description

Fixes an issue with lack of styling on footer links that have their language set explicity (#2815).
Wraps the GatsbyLink in a styled component. I couldn't find a similar use cases in the codebase so I hope the naming convention is okay 😃.

One thought -- the styling is the same as the InternalLink, this could maybe be refactored to be more DRY?

## Related Issue 
Closes #2815

### DIFF
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -50,13 +50,6 @@ const ExplicitLangInternalLink = styled(GatsbyLink)`
   &.active {
     color: ${(props) => props.theme.colors.primary};
   }
-  &:hover {
-    svg {
-      fill: ${(props) => props.theme.colors.primary};
-      transition: transform 0.1s;
-      transform: scale(1.2);
-    }
-  }
 `
 
 const GlossaryIcon = styled(Icon)`
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -46,6 +46,22 @@ const InternalLink = styled(IntlLink)`
   }
 `
 
+const ExplicitLangInternalLink = styled(GatsbyLink)`
+  .is-glossary {
+    white-space: nowrap;
+  }
+  &.active {
+    color: ${(props) => props.theme.colors.primary};
+  }
+  &:hover {
+    svg {
+      fill: ${(props) => props.theme.colors.primary};
+      transition: transform 0.1s;
+      transform: scale(1.2);
+    }
+  }
+`
+
 const GlossaryIcon = styled(Icon)`
   margin: 0 0.25rem 0 0.35rem;
   fill: ${(props) => props.theme.colors.primary400};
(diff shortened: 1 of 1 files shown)


ryancreatescopy (approved):
Thanks for this @minimalsm – looks good! Behaviour works as expected, but haven't inspected the code @wackerow

src/components/Link.js
```diff
@@ -46,6 +46,22 @@ const InternalLink = styled(IntlLink)`
   }
 `
 
+const ExplicitLangInternalLink = styled(GatsbyLink)`
+  .is-glossary {
+    white-space: nowrap;
+  }
```
wackerow, line 52: Shouldn't need this, as it only applied to `/glossary/` links which are not in the "english only" category, if you could please remove

src/components/Link.js
```diff
@@ -46,6 +46,22 @@ const InternalLink = styled(IntlLink)`
...
+  &.active {
+    color: ${(props) => props.theme.colors.primary};
+  }
+  &:hover {
+    svg {
+      fill: ${(props) => props.theme.colors.primary};
+      transition: transform 0.1s;
+      transform: scale(1.2);
+    }
+  }
```
wackerow, line 62: Also not sure we need this `&:hover` styling for the issue being reported. Don't believe there are any SVG's that are affected by the explicit english paths. Correct me if you find something, but if not I think this could go.

This also touches on your comment about DRY, and I think if we do reduce the above down to simply the ` &.active {...}` pseudo-selector then this isn't really a problem, and should still fix the reported issue.

wackerow (changes requested):
@minimalsm Good catch spotting this and thanks for the fix! It looks pretty good to me, but think we could trim down a little (see comments). Functioning as expected though, let me know your thoughts below and we'll get this in soon 👍

src/components/Link.js
minimalsm: 👍

src/components/Link.js
minimalsm: Because of how the links were abstracted I kept both for for coverage but to be honest I wasn't sure either way. 

Now that I think about it a missing glossary or hover probably would've been picked up quicker than a missing active state. Amending now 😃

minimalsm:
> @minimalsm Good catch spotting this and thanks for the fix! It looks pretty good to me, but think we could trim down a little (see comments). Functioning as expected though, let me know your thoughts below and we'll get this in soon 👍

@wackerow requested changes have been made, thanks. 😀

wackerow (approved):
@minimalsm lgtm! Thanks again 👍

wackerow:
@all-contributors please add @minimalsm for bug report and code

PR #2821, opened 2021-04-10
Delete ERC-777 and ERC-1555

ensaryusuf:


wackerow (changes requested):
@ensaryusuf Thanks for doing this per our offline discussion... Looking at it again, we may want to just remove the entire section if all that remains is a link between ERC-20 and ERC-721... These two docs are "next to" each other in the documentation, and the doc navigation at the bottom already fulfills the purpose of linking between them.

ensaryusuf:
The changes are complete.

wackerow (approved):
👍

PR #2834, opened 2021-04-12
Added aria-labels to icon links in footer

### LINKED ISSUE #2770: Improve accessibility of icon links

The social icons (Youtube, Twitter, etc ) at the bottom of the home page don't contain any text. This makes it hard for people that use screen readers to figure out what they do.

The simplest solution I can think of would be to add an ariaLabel property to the Link component to allow a link to have an aria-label attribute. Though, maybe a better solution would be to add an accessibleText property to the Icon component that would display the text to screen readers and hide the icon using aria-hidden. Would be good to get a second opinion!

**Additional context**

Robiquet:
## Description

Added an ariaLabel property to the Link component and supplied it with values in the Footer component to improve accessibility.

Couple of things to note:

- I haven't added translations for the aria labels as they are brand names and will be read out by a screen reader.
- To resolve the scenario mentioned in the linked issue I would only need to update the link where isExternal is true and hide arrow is true (line 123 Link.js). In this PR I've updated all links with the aria-label in case people need to use it elsewhere.

Let me know your thoughts

## Related Issue

[Fixes #2770]

### DIFF
--- a/src/components/Footer.js
+++ b/src/components/Footer.js
@@ -101,18 +101,22 @@ const socialLinks = [
   {
     icon: "github",
     to: "https://github.com/ethereum/ethereum-org-website",
+    ariaLabel: "GitHub",
   },
   {
     icon: "twitter",
     to: "https://twitter.com/ethdotorg",
+    ariaLabel: "Twitter",
   },
   {
     icon: "youtube",
     to: "https://youtube.com/channel/UCNOfzGXD_C9YMYmnefmPH0g",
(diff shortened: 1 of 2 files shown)


wackerow (approved):
Solid, LGTM 🎉 ... With this PR:
image
It's a beautiful thing! Thanks @Robiquet!

wackerow:
@all-contributors please add @Robiquet for code

PR #2900, opened 2021-04-21
Eth2 merge banner

ryancreatescopy:
Created a banner that addresses latest talk of the merge happening earlier than expected.

Also updated terminology from "docking" to "merge" on the merge upgrade page. **I haven't** updated links or updated any other content at this point. I want to confirm what our plan is here and wonder if there's an efficient way to update all this at once and inform translators etc.

### DIFF
--- a/src/content/eth2/docking/index.md
+++ b/src/content/eth2/docking/index.md
@@ -1,16 +1,16 @@
 ---
-title: Docking mainnet with Eth2
-description: Learn about the docking - when mainnet Ethereum joins the Beacon Chain coordinated proof-of-stake system.
+title: The merge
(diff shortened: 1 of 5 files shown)


wackerow:
I think we should. In a separate PR we can rename the link and place a redirect in `netlify.toml` if anyone goes to the old link.

wackerow (approved):
love it... fortunately these words are fairly interchangeable grammatically speaking hah

ryancreatescopy:
Not sure this was ready to merge, it only updated terminology in one place and didn't update the URL of the docking page. Was this known when it was merged?

wackerow:
@ryancreatescopy It was... commented above about fixing the links in separate PR which I've started locally, but wanted to get the banner and initial changes going. Can expedite the PR to avoid issues.

PR #2915, opened 2021-04-23
Update blockchain section

samajammin:
...hopefully making it more clear...


wackerow (approved):

wackerow:
@samajammin I like the changes... I honestly feel like as an addition to this, making a larger callout to the https://andersbrownworth.com/blockchain link would be great. The link right now is a small inline link near the bottom of that section, which specifically links to the portion of the demo where you interact with it (which is great, but if you're brand new, you should go to the link I just mentioned, with the videos to be guided through the demo first). 

Maybe just giving that link its own line and title would do it, as the youtube video is already embedded in the "blocks" page.

PR #2928, opened 2021-04-25
Change error message in example Token contract

schnerd:
## Description

I think `exceeded` is the desired word here, not `succeeded`.


wackerow (approved):
@schnerd 👍🏻 Nice, agreed... thanks for the patch

wackerow:
@all-contributors please add @schnerd for content patch

PR #3003, opened 2021-05-09
Fix WalletCard component layout issues and errors in Safari

### LINKED ISSUE #2971: Wallets page bug
image

Robiquet:
## Description

Switch size of Icons from rem to px because Safari was throw errors related to setting width and height on svgs.
Replaced height: 100% with min-height and adjusted margin to fix layout issues in safari

## Related Issue

[Fixes #2971]

### DIFF
--- a/src/components/WalletCard.js
+++ b/src/components/WalletCard.js
@@ -311,7 +311,7 @@ const WalletCard = ({ wallet }) => {
           >
             <FlipIcon>
               <Translation id="page-find-wallet-card-more-info" />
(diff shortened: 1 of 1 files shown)


wackerow (approved):
@Robiquet This looks good to me, thanks for catching this and posting a fix! Will try to get this live asap.

wackerow:
(Tested this out on Desktop Safari / Brave / Firefox, and mobile Safari / Brave... all of which appear to function as intended without the bug noted on current live version of https://ethereum.org/en/wallets/find-wallet/ when viewing with Safari.)

PR #3029, opened 2021-05-11
Adding to Scaling page Further Readings

skylarweaver:
Adding "An Incomplete Guide to Rollups" from Vitalik to recommended readings


ryancreatescopy (changes requested):
Thanks @skylarweaver I wonder whether this would be better placed on the rollups page though

ryancreatescopy:
You're right, I think it makes sense here. We could add it to rollups page too if needed.

PR #3061, opened 2021-05-16
Add URL to wallet issue template

ligi:
The URL to be used for the "go" action - as far as I see this was missing in the template


wackerow (approved):
@ligi Thanks! I agree, I've had to find this myself before, this helps reduce any errors. 👍🏻

PR #3068, opened 2021-05-17
Added example of Gas Fees

theshubhagrwl:
Added an example for more clarity how the transaction fee is calculated.

I added an example which shows how the transaction fee is calculated based on the Gas units and price.

I think adding this example gives more clarity to the reader as the reader will be able to relate to a real life situation as mentioned in the example.


wackerow (approved):
@theshubhagrwl LGTM! 👍🏻  I agree, a simple example like this can be helpful to clarify, thanks!

wackerow:
@all-contributors please add @theshubhagrwl for content

PR #3074, opened 2021-05-19
stablecoins: set the correct URLs for all DEXs

skylenet:
## Description

The DEX links on the stablecoin section contain wrong URLs. This change uses the correct URL for each DEX.


samajammin:
@all-contributors please add @skylenet for bug reports & content.

PR #3099, opened 2021-05-21
Stop refresh on 'Enter' key press while searching

### LINKED ISSUE #3059: Pressing enter in the search bar refreshes the page
**Describe the bug**

Pressing enter in the search bar refreshes the page

**To Reproduce**

1. On the front page of ethereum.org click the search bar text field
2. Enter some text
3. Press enter

The page is refreshed

**Expected behavior**

I would expect either nothing to happen or to be navigated to the first result

- Browser: Chrome, Firefox

Happy to put in a fix if someone on the team agrees it's a bug

minimalsm:
## Description

Hitting <kbd>Enter</kbd> whilst in the search box refreshed the page. Added onSubmit handler to override this behaviour. Now it simply does nothing.. I think this is probably the only thing that can be done for now unless there are plans to implement a search results page?

Open to ideas on this one because I don't think simply breaking the form submission is optimal.

## Related Issue
Closes #3059

### DIFF
--- a/src/components/Search/Input.js
+++ b/src/components/Search/Input.js
@@ -40,12 +40,16 @@ const Input = ({ query, setQuery, refine, ...rest }) => {
     refine(value)
     setQuery(value)
   }
+  
+  const handleSubmit = (event) => {
+    event.preventDefault()
+  }
 
   const intl = useIntl()
   const searchString = translateMessageId("search", intl)
 
   return (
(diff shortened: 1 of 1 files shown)


wackerow (approved):
@minimalsm Good call!... I agree, this is more in line with expected behavior, thanks for patching this 👍🏻

PR #3100, opened 2021-05-21
Fix Spelling Mistake

minimalsm:
## Description

Changed 'Etherium' to correct spelling 😀


wackerow (approved):
@minimalsm Nice 👍🏻 This probably got missed because it was in the Spanish translations, but doesn't look like that spelling is used anywhere else in the Spanish docs. Thanks again!

minimalsm:
Credit goes to Sam for leading me into a rabbit-hole about the different spellings of 'Ethereum' 😅

PR #3109, opened 2021-05-23
Fix possible grammatical mistake

blackblather:
## Description

I believe there was a simple mistake in the text of the _Scaling_ page, in the chapter of _State Channels_.
I think the author meant to use "thus" instead of "this".

### DIFF
--- a/src/content/developers/docs/scaling/index.md
+++ b/src/content/developers/docs/scaling/index.md
@@ -52,7 +52,7 @@ Learn more about [rollups](/developers/docs/scaling/layer-2-rollups/).
 
 #### State Channels {#channels}
 
-State channels utilize multisig contracts to enable participants to transact quickly and freely off-chain, then settle finality with mainnet, thus minimizing network congestion, fees and delays. The two types of channels are currently state channels, and payment channels.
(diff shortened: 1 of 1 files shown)


minimalsm (changes requested):
Nice spot. Personally, I think using simpler language here is preferable.

```diff
+++ This minimizies network congestion, fees and delays. The two types of channels are currently state channels, and payment channels.
```

brettsmentek:
> Nice spot. Personally, I think using simpler language here is preferable.
> 
> ```diff
> +++ This minimizies network congestion, fees and delays. The two types of channels are currently state channels, and payment channels.
> ```

I agree with this phrasing. Typo in `minimizes`, though. I also don't believe there should be a comma in the second sentence, `The two types of channels are currently state channels and payment channels.`

blackblather:
Thank you guys 👍
Yeah, I agree with Brett. The new phrase is simpler, but would also remove the comma at the end.

ryancreatescopy:
@all-contributors please add @blackblather for content and bugs.

ryancreatescopy:
@all-contributors please add @brettsmentek for content.

PR #3131, opened 2021-05-25
Typo corrections

wackerow:
## Description
Two typos noted in Deploy 2.29.1 changes:
- "minimizies"
- Fixing statement about hashing your public key to get your public key.


samajammin (approved):

Aivolt1 (commented):
Each pending TX hash can be accessed as a wallet preventing the funds from reaching end point. Intersecting with the contract and contract deployed address + finding the 64 bit hex in the log gives access to the wallet

Issue #3139, opened 2021-05-26
Add hover effect to the Tooltip


Uttam-Singhh:
Add hover effect to the Tooltip so that the content is visible when hovering over the I button.

Before: 
![IMG_20210526_230715](https://user-images.githubusercontent.com/63050765/119706858-1e61a200-be78-11eb-8d2b-179b9631a64d.jpg)

After:
![IMG_20210526_230736](https://user-images.githubusercontent.com/63050765/119706802-0d189580-be78-11eb-9a91-ff63e2d78213.jpg)

wackerow:
@Uttam-Singhh The above PR is merged, but will keep this open in the meantime per [my comment on that thread](https://github.com/ethereum/ethereum-org-website/pull/3140#issuecomment-856165135). If you feel this issue still needs work, will keep open.

PR #3140, opened 2021-05-26
Added hover effect to Tooltip

Uttam-Singhh:
Solves #3139 
I added the hover effect to the tooltip.
Please let me know if any changes are required.


samajammin:
@Uttam-Singhh thanks for this PR!

@wackerow please take a look. Should the tooltip close when hovering off the tooltip? Currently it stays open unless clicked on.

Uttam-Singhh:
> @Uttam-Singhh thanks for this PR!
> 
> @wackerow please take a look. Should the tooltip close when hovering off the tooltip? Currently it stays open unless clicked on.

Thank you! @samajammin 
I thought so it is good to have it open unless clicked since before also(when using click) it was not going away unless clicked.

ryancreatescopy (changes requested):
I think the behaviour should be as follows:

- if hover, tooltip should disappear as soon as your mouse leaves the trigger component
- if clicked, the trigger must be clicked again to dismiss the tooltip

It's not very intuitive to have a tooltip pop up when hovering and then having to click to remove it (an action you never did in the first place because it was triggered by a hover).

Uttam-Singhh:
> Thanks for working on this! It's almost a great improvement.
> 
> I think the behaviour should be as follows:
> 
> * if hover, tooltip should disappear as soon as your mouse leaves the trigger component
> * if clicked, the trigger must be clicked again to dismiss the tooltip
> 
> It's not very intuitive to have a tooltip pop up when hovering and then having to click to remove it (an action you never did in the first place because it was triggered by a hover).

Oh okay, Thank you for your feedback @ryancreatescopy,
I will work on this!

Uttam-Singhh:
Hello, @samajammin @ryancreatescopy please help me out here!
Today when I tried to run `yarn start` command I am getting an error on loop(attached image), I don't know what had happened.
Yesterday it was working fine but now it is stuck on a loop and not working.
P.S - I already tried `yarn cache clean`, I also tried checkout other branches but still the problem persists.

minimalsm:
> * if hover, tooltip should disappear as soon as your mouse leaves the trigger component
> * if clicked, the trigger must be clicked again to dismiss the tooltip
> 
> It's not very intuitive to have a tooltip pop up when hovering and then having to click to remove it (an action you never did in the first place because it was triggered by a hover).

I agree but I also don't think it is intuitive to click on an element that has a hover effect like this.

ryancreatescopy:
The click is mainly for mobile so that you can view the tooltip without a hover

Uttam-Singhh:
> The click is mainly for mobile so that you can view the tooltip without a hover
> […](#)
> Sent from my iPhone
> On 27 May 2021, at 21:13, Joshua ***@***.***> wrote: ﻿ if hover, tooltip should disappear as soon as your mouse leaves the trigger component if clicked, the trigger must be clicked again to dismiss the tooltip It's not very intuitive to have a tooltip pop up when hovering and then having to click to remove it (an action you never did in the first place because it was triggered by a hover). I agree but I also don't think it is intuitive to click on an element that has a hover effect like this. — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

so should I work more on this or not?
I mean it's working fine for the click(which can be used for mobile) and for desktop we have hover+click both.

minimalsm:
> The click is mainly for mobile so that you can view the tooltip without a hover

Sorry I wasn't very clear. The problem I was suggesting is if someone were to hover the element, then attempt to click the Datasource URL - wouldn't it disappear when leaving a hovered state?

Uttam-Singhh:
Can anyone please guide me more on this!
Do I have to work on this more to get this PR merged??
It means a lot to me please put a light on this! @samajammin @ryancreatescopy @wackerow

wackerow:
@Uttam-Singhh Sorry for delay in getting back to you, appreciate you bearing with us as a small team. Agree this is looking nice but not quite there yet...

When I use it right now, after hovering the tooltip pops up, but it does not go away when I move the mouse away.. It should only stay visible while still hovering over the (i) or within the tooltip (so you can click the link). 

This should only add functionality on top of the previous click/tap toggling, which should continue to be preserved (click/tap on i to open, clicking/tapping on i again or anywhere outside of tooltip causes the tooltip to close)

Uttam-Singhh:
> @Uttam-Singhh Sorry for delay in getting back to you, appreciate you bearing with us as a small team. Agree this is looking nice but not quite there yet...
> 
> When I use it right now, after hovering the tooltip pops up, but it does not go away when I move the mouse away.. It should only stay visible while still hovering over the (i) or within the tooltip (so you can click the link).
> 
> This should only add functionality on top of the previous click/tap toggling, which should continue to be preserved (click/tap on i to open, clicking/tapping on i again or anywhere outside of tooltip causes the tooltip to close)

Actually, there is a concerning case suppose a person is hovering over (i), the tooltip pops up but what if at the same time he clicks on the (i) button then what will happen?? that is why I didn't add up the function of hiding the tooltip when cursor goes away!
and then after clicking if he moves his cursor away then the tooltip will not automatically close itself, one should click somewhere to hide it.

But, yeah I will work on this and will find a good solution!

Uttam-Singhh:
@wackerow I was working on this for the past few days, I tried a lot to implement as you guided but every time I tried something there were bad consequences.
for example, you wanted me to preserve the onclick function but if someone were to hover the element, then attempt to click the Datasource URL then it would disappear  because it is leaving a hovered state.


So, At last, I came to a conclusion to add a **title** to the tooltip.
Whenever someone hovers over the `i` button and waits there for more than a second they will see a title box with a message `more info`.


Please let me know what do you think about this!

jley81 (approved):

ryancreatescopy:
If we can sentence case the title tag e.g. "More info" I'm good with this... Clicking/pressing is the most important action so we should be optimising for that.

wackerow:
@Uttam-Singhh This is functioning, but I just wanted to circle back to the original issue you posted #3139. 

You suggested that the tooltip should be shown when hovering. On the live site right now, hovering will provide a change in color to the "i" icon with the cursor switching to a "pointer" indicating there is a clickable action. This PR now provides a delayed tooltip of "more info", and clicking then provides a different/second tooltip with the link. Happy to pull this in, but I want to make sure you feel this addresses the original issue.

Uttam-Singhh:
@wackerow Yes I am aware of the original issue but as I mentioned above it was very very cumbersome and I wasn't able to find a solution (Please read the above message).

Also @wackerow if you want I can change the name and description of issue #3139 
Should I?

wackerow (approved):

wackerow:
@all-contributors please add @Uttam-Singhh for bug patch

Uttam-Singhh:
> @Uttam-Singhh Yes, I saw the message =), but this issue was originally posted by you, so I wanted to make sure this suits what you were aiming for. Like I mentioned, your changes function just fine and happy to pull in where it's at (* first fixing the capitalization change Ryan suggested).
> 
> If this isn't satisfactory for your original issue, we can leave the issue open and open it up to other members of the community to see if anyone can help out. I can take a closer look myself down the line, but admittedly it's a lower priority for the team at the moment.
> 
> So I will just fix the capitalization now and merge this in. If you're content with this change in regards to the issue you posted, we can close it, otherwise we can keep the issue open and let someone else take a stab at it. Let me know your thoughts on whether or not you're content with the issue being closed.

First of all, Thank you @wackerow, I am aware that my issue was originally posted for a different purpose.
and Yes I totally **agree** with you about **not closing the issue**. I would be very pleased if someone else will work on that feature since it will improve the website.

PR #3141, opened 2021-05-26
Update community lead JD for compensation range, requesting cover letter in application

GeorgeTrotter:


samajammin (approved):
LGTM - I updated the frontmatter as well to match.

PR #3156, opened 2021-05-28
mistake in Ethereum address creation

Rub3cula:
In line 70 where the public address is described, it states that "You get a public address for your account by taking the last 20 bytes of the SHA3 hash of the private key", this should perhaps be "You get a public address for your account by taking the last 20 bytes of the SHA3 hash of the public key"?

## Description
Currently it states the public address is derived from the private key, which is wrong. The Ethereum address is derived from the last 20 bytes of the Keccak-256 hash of the corresponding public key.

### DIFF
--- a/src/content/developers/docs/accounts/index.md
+++ b/src/content/developers/docs/accounts/index.md
@@ -67,7 +67,7 @@ Example:
 
 `fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd036415f`
 
-The public key is generated from the private key using the [Elliptic Curve Digital Signature Algorithm](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm). You get a public address for your account by taking the last 20 bytes of the hash of the public key and adding `0x` to the beginning.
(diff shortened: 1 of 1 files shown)


src/content/developers/docs/accounts/index.md
```diff
@@ -67,7 +67,7 @@ Example:
 
 `fffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd036415f`
 
-The public key is generated from the private key using the [Elliptic Curve Digital Signature Algorithm](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm). You get a public address for your account by taking the last 20 bytes of the SHA3 hash of the private key and adding `0x` to the beginning.
+The public key is generated from the private key using the [Elliptic Curve Digital Signature Algorithm](https://en.wikipedia.org/wiki/Elliptic_Curve_Digital_Signature_Algorithm). You get a public address for your account by taking the last 20 bytes of the hash of the public key and adding `0x` to the beginning.
```
wackerow, line 70: @Rub3cula  🤦🏻  Good call, this was my bad. Read it too quickly while reviewing the last deploy.

Did you remove SHA3 for a reason? I would leave it in personally, either that or Keccak-256

Rub3cula: Very understandable!
I agree, Keccak-256 would be more correct I think

wackerow (approved):
@Rub3cula 👍🏻

wackerow:
@all-contributors please add @Rub3cula for content patch

PR #3159, opened 2021-05-29
Update StatBoxGrid development fallback figures 

minimalsm:
In development the StatBoxGrid component doesn't fetch from coingecko/etherscan/defipulse but the fallback figures were way off. I updated the fallbacks figures to reflect todays figures.

## Development Figures
Screenshot 2021-05-29 at 21 42 53

## Production Figures (5-29-2021) && Proposed New Development Figures
Screenshot 2021-05-29 at 21 42 43


wackerow (approved):
@minimalsm 😂  So two of these were straight up dummy numbers, the 23456789 and 1234567... but the price was legit at time of that update lol, nodes as well 👍🏻

PR #3133, opened 2021-05-26
Transaction fee calculation

miiiguel:
Correcting the formula for transaction fee to gasLimit times gasPrice

PR #3162, opened 2021-05-30
Asterisk in markdown fix

PeersHut:
Changed * into &#42; because * is not visible on webpage

## Description
On line 29 and 30 on the webpage https://ethereum.org/en/developers/docs/gas/
the * are not visible even not in de HTML code. 
I asume that the change from * to &#42; makes the differrence.


wackerow:
@PeersHut Hey thanks for the PR! Definitely appreciated, but this will actually conflict with another PR, #3133 which wraps these statements in back-ticks, which remedies this while maintaining readability. Will leave this open in case that doesn't get merged, but if it does will close this out in favor. Please feel free to check that one out and make sure it works for you to remedy this issue =)

wackerow:
(Closed in lieu of #3133)

PR #3163, opened 2021-05-30
fix a typo

AmirAliM:
I believe the correct function name here is `_mint` and not `_emit`. I can't even find a definition of the function `_emit` anywhere.


wackerow (approved):
@AmirAliM 👍🏻 Good catch, thanks!

wackerow:
@all-contributors please add @AmirAliM for content patch

PR #3170, opened 2021-06-01
Docs: Add Info About Sybil Attacks in 'intro-to-ethereum' Page

### LINKED ISSUE #3169: Including Sybil Attack Info in Intro to Ethereum Page
I suggest adding some info about [Sybil attacks](https://en.wikipedia.org/wiki/Sybil_attack) in the [Intro to Ethereum page](https://ethereum.org/en/developers/docs/intro-to-ethereum/) ([Respective Markdown file](https://github.com/ethereum/ethereum-org-website/blob/dev/src/content/developers/docs/intro-to-ethereum/index.md)). More specifically, I think it should be added under the `what-is-a-blockchain` header, near the info about consensus and proof-of-work.

I think that this is an important piece of information in regards to understanding why proof-of-work has a financial cost behind it. Which is partly to defend against people gaining majority control over the blockchain network (as far as I understand).

I'll make a PR for it right after this issue is published.

Discussion in regards to the PR can be had here, if necessary.

Hakeemmidan:
## Description

Added some details about Sybil attacks on the [Intro to Ethereum page](https://ethereum.org/en/developers/docs/intro-to-ethereum/) ([Respective Markdown file](https://github.com/ethereum/ethereum-org-website/blob/dev/src/content/developers/docs/intro-to-ethereum/index.md)).

## Related Issue
Fixes #3169

### DIFF
--- a/src/content/developers/docs/intro-to-ethereum/index.md
+++ b/src/content/developers/docs/intro-to-ethereum/index.md
@@ -15,7 +15,9 @@ A blockchain is best described as a public database that is updated and shared a
 
 Each new block and the chain as a whole must be agreed upon by every computer in the network. These computers are known as "nodes". This is so everyone has the same data. To accomplish this distributed agreement, blockchains need a consensus mechanism.
 
-Ethereum currently uses a [proof-of-work](/developers/docs/consensus-mechanisms/pow/) consensus mechanism. This means that anyone who wants to add new blocks to the chain must solve a difficult puzzle that requires a lot of computing power. Solving the puzzle "proves" that you have spent the computational resources. Doing this is known as [mining](/developers/docs/consensus-mechanisms/pow/mining/). Mining is typically brute force trial and error but adding a block successfully is rewarded in ETH.
+Ethereum currently uses a [proof-of-work](/developers/docs/consensus-mechanisms/pow/) consensus mechanism. This means that anyone who wants to add new blocks to the chain must solve a difficult puzzle that requires a lot of computing power. Solving the puzzle "proves" that you have spent the computational resources. Doing this is known as [mining](/developers/docs/consensus-mechanisms/pow/mining/).
+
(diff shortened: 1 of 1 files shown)


wackerow (changes requested):
@Hakeemmidan Hey, appreciate you dropping the PR. Not against adding a note about Sybil resistance in the docs, but not sure I agree it belongs in the introduction page in the "What is a blockchain" section. 

Perhaps this would belong better in the `/consensus-mechanisms/pow` markdown. @samajammin Any thoughts here?

(Also, can we revert the yarn.lock changes? Thanks!)

Hakeemmidan:
Hi @wackerow, 

No problem! 

I was thinking about the same thing too. 

Although I also think it may belong somewhere down the line (like `/consensus-mechanisms/pow`), I still think that a brief mention of it in the introduction is somewhat necessary. It gives context to part of the reason why consensus mechanisms are there in the first place. If it was up to me, I'd add it in both places, and just have it be more detailed on the `/consensus-mechanisms/pow` page.

Reverted `yarn.lock`

wackerow (changes requested):
I suppose I'm still not convinced this belongs in one of the first paragraphs introducing Ethereum. I think it's a reasonable thing to comment on in the PoW section but hesitate to introduce this topic immediately in the flow of the docs. Any chance we could find a place for these additions on the PoW page
(ping @samajammin)

samajammin:
Thanks for the PR!

This addition makes more sense to me on the mining page. This intro page is intended to keep things high level- I don't think covering Sybil resistance is necessary here.

Cc @wackerow curious what you think.

samajammin:
Hah whoops just seeing your comment @wackerow - looks like we're on the same page.

Hakeemmidan:
Sounds good. And no problem.
Thank you all for your consideration!
I'll add it to the mining page if I get the time (mostly unlikely). I'm currently working on other things.

Cheers!

wackerow:
@Hakeemmidan Thank you as well for the input! A PR (#3353) is actually in progressing that will address this on the consensus-mechanisms page, hope this helps address your initial concerns.  Cheers!

Hakeemmidan:
@wackerow No problem. It does! Thanks for the update.

Cheers!

PR #3196, opened 2021-06-05
Update oracles/index.md

pawurb:
Current code sample produces the following warning:

```
 Warning: Unused local variable.
    uint80 roundID,
    uint startedAt,
    uint timeStamp,
    uint80 answeredInRound
```
This change fixes it.

### DIFF
--- a/src/content/developers/docs/oracles/index.md
+++ b/src/content/developers/docs/oracles/index.md
@@ -87,13 +87,7 @@ contract PriceConsumerV3 {
      * Returns the latest price
      */
     function getLatestPrice() public view returns (int) {
-        (
-            uint80 roundID,
-            int price,
-            uint startedAt,
-            uint timeStamp,
-            uint80 answeredInRound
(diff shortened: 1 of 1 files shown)


wackerow:
@PatrickAlphaC Any chance you could help double check these changes to the oracle page? 🙏🏻  Looks fine to me, but wanted to make sure there wasn't another reason for these variables

PatrickAlphaC:
He's correct here! Typically, I like to see them in for instructional pieces, so you can know right away everything that a function returns. What do you think is best practice here?

wackerow:
@PatrickAlphaC Right, I can see the value of keeping the variables visible to aid in understanding... @pawurb was this causing a breaking error? It appears to be more of a warning, which in the setting of a dev docs, this may be worth keeping in and just making a note to the reader that they should expect a warning here so they have a reference to what the function returns.

pawurb:
@wackerow it was just a warning. You're right probably better to keep it for clarity. I'll close the issue than.

PR #3210, opened 2021-06-07
Added a functionality to check if PR is merged or not

adi44:

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,6 @@
   <p>👋 Welcome to ethereum.org!</p>
 </h1>
 
-test PR
 
 This is the repo for the [ethereum.org](https://ethereum.org) website, a resource for the Ethereum community. The purpose of the site is to _“Be the best portal to Ethereum for our growing global community"_ - read more about what this means [here](https://ethereum.org/en/about/).
 
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@
   <p>👋 Welcome to ethereum.org!</p>
 </h1>
 
+test PR
+
 This is the repo for the [ethereum.org](https://ethereum.org) website, a resource for the Ethereum community. The purpose of the site is to _“Be the best portal to Ethereum for our growing global community"_ - read more about what this means [here](https://ethereum.org/en/about/).
 
 [ethereum.org](https://ethereum.org) is improved and changed over time through the contributions of community members who submit content, give feedback, or volunteer their time to managing its evolution. If you’re interested in helping to improve [ethereum.org](https://ethereum.org), find out [how to contribute](https://ethereum.org/en/contributing/).
--- a/src/components/Roadmap.js
+++ b/src/components/Roadmap.js
@@ -106,7 +106,7 @@ return ( data.merged===true)
                 "allcontributors[bot]" !== issue.user.login
                  }
               if(!!issue.pull_request.url){
-                return(!!isMerged(issue.pull_request.url))
+                return(isMerged(issue.pull_request.url))
               }
                 
                 
--- a/src/components/Roadmap.js
+++ b/src/components/Roadmap.js
@@ -57,16 +57,9 @@ const Roadmap = () => {
     implemented: blankIssues,
   })
 
-  const checkifMerge= async (url)=>{
-    let resp = await axios.get(url+"/merge")
-    if(resp.status==204){
-      
-      return true
-      
-    }
-    else{
-      return false
-    }
+  const isMerged=async(url)=>{
+    const resp= await axios.get(url+"merge")
+    return resp.status===204
   }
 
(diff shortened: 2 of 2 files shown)


src/components/Roadmap.js
```diff
@@ -57,6 +57,18 @@ const Roadmap = () => {
...
+    if(resp.status===203){
+      
+      return true
+      
+    }
+    else{
+      return false
+    }
+  }
+
```
wackerow, line 71: Cool, will check this out when it builds...
Keep in mind `resp.status === 203` is a boolean statement by itself, and you can just return that, ie:
```
const resp = await axios.get(url + "merge")
return resp.status === 203 // This will return true or false depending on output
```

src/components/Roadmap.js
```diff
@@ -96,9 +108,10 @@ const Roadmap = () => {
           const implemented = issues
             .filter(
               (issue) =>
-                issue.state === "closed" &&
+                (issue.state === "closed" &&
                 "allcontributors[bot]" !== issue.user.login &&
-                !!issue.pull_request
+                !!issue.pull_request && checkifMerge(issue.pull_request.toString())===true)
```
wackerow, line 113: Similarly here... don't need the `===true` since `checkifMerge()` returns a boolean.

Also just noticing, mind adjusting the name to something like `isMerged` since it's a boolean return?

src/components/Roadmap.js
adi44: @wackerow I am not able to check locally. It shows Loading error Please refersh at localhost:8000. I have added github API key and etherscan key still its not working. Can you pls guide me

adi44: Ok Sure I will change it

src/components/Roadmap.js
adi44: ok I will change the naming convention

adi44:
@wackerow I am not able to check locally. It shows Loading error Please refersh at localhost:8000. I have added github API key and etherscan key still its not working. Can you pls guide me?

adi44:
@wackerow It seems working now! pls have a look

wackerow (changes requested):
image
image

Hm, still doesn't seem to be filtering out closed issues. Taking a look, I'm not sure we need to make an additional axios request here since we're already making a request using the lambda function `src/lambda/roadmap.js`. This would be where I would take a look to see if there is anything in the response object that we could use to filter down.

adi44:
Hey, extra axios request is being made to check if PR for specific issue is
merged or not. Currently the roadmap js gives out the list of issues

src/components/Roadmap.js
```diff
@@ -57,6 +57,12 @@ const Roadmap = () => {
     implemented: blankIssues,
   })
 
+  const isMerged=async(url)=>{
+    const resp= await axios.get(url)
+    const data = await resp.data
+return ( data.merged===true)
```
wackerow, line 63: if `data.merged` returns a boolean, we don't need the `===true`... if if doesn't return a boolean, can use `return !!data.merged` to return a boolean from it.

src/components/Roadmap.js
```diff
@@ -96,10 +102,16 @@ const Roadmap = () => {
...
-                "allcontributors[bot]" !== issue.user.login &&
-                !!issue.pull_request
+              {issue.state === "closed" &&
+                "allcontributors[bot]" !== issue.user.login
+                 }
+              if(!!issue.pull_request.url){
+                return(isMerged(issue.pull_request.url))
+              }
+                
+
```
wackerow, line 112: Syntax here is broken... 
```
.filter(
  (issue) => {
    /* code here is run without an automatic return */
  }
  /* below this is invalid syntax since we're outside the curly braces and still within the .filter() method */
  if (!!issue.pull_request.url) {
    return isMerged(issue.pull_request.url)
  }
)
```

README.md
```diff
@@ -8,6 +8,7 @@
   <p>👋 Welcome to ethereum.org!</p>
 </h1>
 
+
```
wackerow, line 11: (please revert changes to this file)

src/components/Roadmap.js
wackerow: And help me out here... do we not already have this data available from the axios call within the `useEffect` hook? Not sure I understand why we're making a separate request. Mind walking me through your thought process here?

wackerow (changes requested):
(see notes)

minimalsm:
Hey @adi44, do you need some more help getting this complete?

adi44:
Hey @joshua I wasn't able to find any other way to solve this issue

minimalsm:
No worries @adi44. Thanks for your efforts on this one. I'm going to close it out in favour of #3822 for now.

PR #3292, opened 2021-06-17
Update transactions/index.md

m4sterbunny:
UX/Accuracy fix
## Description
small rewordings to improve clarity

### DIFF
--- a/src/content/developers/docs/transactions/index.md
+++ b/src/content/developers/docs/transactions/index.md
@@ -1,7 +1,7 @@
 ---
 title: Transactions
 description: An overview of Ethereum transactions – how they work, their data structure, and how to send them via an application.
-lang: en
+lang: en-UK
 sidebar: true
 ---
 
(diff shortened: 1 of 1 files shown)


src/content/developers/docs/transactions/index.md
```diff
@@ -138,9 +138,11 @@ Once the transaction has been submitted the following happens:
...
-   - The larger the block confirmation number the more immutable the transaction is. So for higher value transactions, more block confirmations may be desired.
-
+4. Your transaction is assigned a block confirmation number. This number signifies the first block your transaction was included in. Note:
+  - The block confirmation number provides a baseline to estimate the number of blocks created since your transaction's block. The higher the number, the greater the certainty that the transaction was processed and recognised by the network. 
+  - Recent blocks may be re-organised. This gives the impression that the transaction was reversed. The transaction is still valid but may be included in a different block.
+  - The probability of a re-organisation diminishes with every subsequent block mined, i.e. the larger the block confirmation number, the more immutable the transaction is. So, for higher value transactions, more block confirmations may be desired.
```
wackerow, line 144: I think we need to clear up exactly what is being defined here.

Originally: "block confirmation number"... was being defined as "number of blocks created since the block that your transaction was included in." I would argue this isn't 100% correct, as it would also include the block the txn was included in. 
Your change to "This number signifies the first block your transaction was included in" actually defines _block height_. Also, this number isn't really "assigned" to the transaction. 

---
Just jotted some other thoughts about the wording in this section:
Once your transaction is mined into a block, that counts as one confirmation. At this point, your transaction has already been recognized by the network and _has been processed_, but there is a small chance the block it was included in isn't permanent, and will need to be processed again and included in another block.

src/content/developers/docs/transactions/index.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Transactions
 description: An overview of Ethereum transactions – how they work, their data structure, and how to send them via an application.
-lang: en
+lang: en-UK
```
wackerow, line 4: Our internationalization setup does not yet account for UK English, just 'en'. Mind reverting this?

wackerow (changes requested):
Dropped a few thoughts on this. Think we mainly need to just clean up the wording around exactly what we're defining. There is a difference between the block height your tx was included in, and the number of confirmations your tx is considered to have. 

Reorgs are an additional topic that could be elaborated on separately, but thoughts on how to simplify the concept here welcomed. Left a couple thoughts.

wackerow:
@m4sterbunny I think I noticed this was originally posted in a separate PR? If possible, do me a favor and make changes within this PR instead of opening a new one. Helps keep things organized =)

m4sterbunny:
Closing in favour of PR#3267

PR #3298, opened 2021-06-17
SafeMath removal from Solidity 0.8.0

gsthina:
While implementing the steps on this page, happened to know that Solidity 0.8.0 has an improvement [documented here](https://blog.soliditylang.org/2020/12/16/solidity-v0.8.0-release-announcement/).

## Description

The [documentation](https://ethereum.org/en/developers/tutorials/create-and-deploy-a-defi-app/) to create and deploy Dapps has the source codes with Solidity ^0.6.2. 

The developers who would use a later compiler might face an issue with the usage of SafeMath.

```
using SafeMath for uint256;
```
A comment about the new changes for Solidity 0.8.0 would help the developers to learn the improvement.

Something like "As of Solidity v0.8.0, mathematical operations can be done safely without the need for SafeMath. More info here: https://blog.soliditylang.org/2020/12/16/solidity-v0.8.0-release-announcement", or how it suits better.

### DIFF
--- a/src/content/developers/tutorials/create-and-deploy-a-defi-app/index.md
+++ b/src/content/developers/tutorials/create-and-deploy-a-defi-app/index.md
@@ -192,7 +192,12 @@ import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 
 contract FarmToken is ERC20 {
     using Address for address;
-    using SafeMath for uint256;
+    using SafeMath for uint256; /* This is not required for Solidity ^0.8.0
+                                   
+                                   Everyone who wanted to write a SafeMath function for exponentiation probably noticed that it is rather expensive to do that because the EVM does not provide overflow signalling. Essentially you have to implement your own exp routine without using the exp opcode for the general case.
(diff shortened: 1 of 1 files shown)


wackerow (changes requested):
Okay, so do I understand this correct that Solidity 0.8.0 now does math operations safely without having to import SafeMath?

This particular excerpt doesn't seem to convey that fully. If that's the case, maybe a more concise comment here about "As of Solidity v0.8.0, mathematical operations can be done safely without the need for SafeMath. More info here: ..."
(This takes up a lot more space than it needs to imo 😂 )

gsthina:
@wackerow - Yes, agree with you! I shall make that edit now. Thanks :)

gsthina:
I guess, this PR can be ignored and the #3331 can be considered. Thanks :)

wackerow:
@gsthina Just a note, in the future try to keep updates to a PR within the same branch / PR thread to help with organization, thanks 👍🏻

PR #3332, opened 2021-06-22
homepage charts testing

Andrew-Sofos:

### DIFF
--- a/src/components/StatsBoxGrid.js
+++ b/src/components/StatsBoxGrid.js
@@ -753,8 +753,6 @@ const StatsBoxGrid = () => {
             fillOpacity="0.2"
             connectNulls={true}
           />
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "dotenv": "^8.2.0",
     "ethereum-blockies-base64": "^1.0.2",
     "framer-motion": "^4.1.3",
-    "gatsby": "^3.0.4",
+    "gatsby": "^3.7.2",
```
samajammin, line 24: Is there a reason why we're updating Gatsby here? Seems that could/should be done in a separate PR to isolate any potential issues.

Andrew-Sofos: Hi @samajammin , I didn't do it manually I think the action: yarn add recharts, updated the gatsby in the prosses of installing the library.

PR #3336, opened 2021-06-24
Adds FUD to glossary [Closes #3152]

### LINKED ISSUE #3152: Glossary Suggestion:  F.U.D. Fear, Uncertainty, and Doubt.
F.U.D. Fear, Uncertainty, and Doubt. Often used to refer to negative, misleading, or false information relating to a cryptocurrency, token, or the wider industry.

An example of F.U.D. being used by a prominent member of Society, was Elon Musk. When he announced that Tesla would no longer accept Bitcoin as a form of payment, because it is not friendly to the environment. 

https://abcnews.go.com/Business/bitcoin-plunges-elon-musk-puts-environmental-impact-scrutiny/story?id=77665619

wackerow:
## Related Issue
#3152

### DIFF
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -245,6 +245,10 @@ A service carried out via [smart contract](#smart-contract) that dispenses funds
 
 <DocLink to="/developers/docs/networks/#testnet-faucets" title="Testnet Faucets" />
 
+### fear, uncertainty, doubt (FUD) {#fud}
+
+Used to refer to negative, misleading, or false information relating to a cryptocurrency, token, or the wider industry.
+
 ### finality {#finality}
 
 Finality is the guarantee that a set of transactions before a given time will not change and can't be reverted.


samajammin:
I don't really see this as a crypto-specific term (e.g. top Google result for "fud": https://en.wikipedia.org/wiki/Fear,_uncertainty,_and_doubt) but I'm not strongly opposed to including this.

wackerow:
I placed this PR in response to the posted issue, however, in hindsight I agree with your point @samajammin that this  doesn't really apply to just crypto. This gets into my other comment on the issue about needing to be careful of the precedent set for introducing terms that apply to a much broader set of circumstances than just Ethereum or crypto. In particular, investment terms that predate crypto do not feel like good candidates for the Ethereum glossary. As such, closing this PR. Can leave the existing issue open in the meantime if anyone would like to offer their opinion on the matter.

PR #3346, opened 2021-06-28
Update `/merge` with post-merge cleanup note

wackerow:
## Description
- Adds some information regarding the nuance that there will be a post-merge cleanup fork. 
- Focused on the key point that withdraws will not be capable until then, not immediately after the merge.
- Couple spelling corrections


samajammin (approved):
LGTM, thanks!

PR #3356, opened 2021-06-28
Remove trinity from the suggested ethereum clients

LefterisJP:
Remove trinity from the suggested ethereum clients as it's not actively supported/developed, and thus should not be in something as user-facing as the ethereum website.

Context on what lead to this PR: https://twitter.com/LefterisJP/status/1409617303507963905


wackerow:
Fixed in #3357 👍🏻

PR #3373, opened 2021-07-02
Explicitly types frontmatter schema

wackerow:
## Description
- Adds fields exclusively used in `job.js` template to frontmatter schema. 
- Fixes build error caused by 1d7c5181 after removal of the only markdown file that populated these fields. Solution suggested within logs
```
It is recommended to explicitly type your GraphQL schema if you want to use optional fields. This way you don't have to add the mentioned "dummy content". Visit our docs to learn how you can define the schema for "Frontmatter":
```
- Leaves `job.js` template available for future use
- Adds Netlify redirect from old page to /about/#open-jobs

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -306,6 +306,11 @@ exports.createSchemaCustomization = ({ actions }) => {
       incomplete: Boolean
       template: String
       summaryPoints: [String!]!
+      position: String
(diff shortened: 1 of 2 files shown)


samajammin:
Cool! We should probably add comments for which templates use which frontmatter fields.

Issue #3402, opened 2021-07-08
Translation Program page update #1


lukassim:
**Is your feature request related to a problem? Please describe.**

The update is the first in a series of planned updates, with the aim of adding content to the Translation Program page and expanding the scope of information it offers. 

**Describe the solution you'd like**

I would like to add a video to the top of the page, expand the overview a bit and include some key stats about the Translation program so far.
Additionally, I am planning on adding a section about POAPs for translators and moving the ‘Eth2 Launchpad’ section above the ‘In-progress Translations’ to increase the visibility of the Eth2 Launchpad section. 

**Describe alternatives you've considered**

**Additional context**

The video that I would like to add - https://blog.ethereum.org/img//2020/06/world2ethereum_white_small_%20(1).mp4

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

lukassim:
I would like to be assigned to this issue. :)

PR #3406, opened 2021-07-09
Update Translation Program page

lukassim:
## Description

Adds the following to the Translation Program page:
- World2ethereum video
- Short overview
- Some key stats
- Section about POAPs

Moves Eth2 Launchpad section above In-progress translations

## Related Issue

#3402


wackerow:
@lukassim This looks awesome, thanks! I'm hesitant about the video not because it wouldn't look great, but because it's bulky to load that automatically for the user when they come to that page. 

I'd like to hold on the video for now, so it doesn't block getting the rest of this live. Let's chat options for the video at standup 👍🏻

wackerow (approved):

PR #3417, opened 2021-07-13
MetaMask updates July 2021

mapachurro:
MetaMask has a number of features that are not currently reflected.

## Description

MetaMask features DEX, DeFi, and needed their brand color updated, etc.


src/data/wallets.csv
```diff
@@ -4,7 +4,7 @@ argent,Argent,https://www.argent.xyz/,#FF875B,TRUE,FALSE,FALSE,FALSE,TRUE,TRUE,T
 dharma,Dharma,https://www.dharma.io/,#FFF,TRUE,FALSE,FALSE,FALSE,TRUE,TRUE,TRUE,TRUE,FALSE,FALSE,TRUE,FALSE
 imtoken,imToken,https://token.im/,#FFF,TRUE,FALSE,FALSE,TRUE,TRUE,TRUE,TRUE,FALSE,FALSE,FALSE,FALSE,TRUE
 portis,Portis,https://www.portis.io/,#FFF,TRUE,FALSE,TRUE,FALSE,TRUE,TRUE,FALSE,FALSE,FALSE,FALSE,FALSE,FALSE
-metamask,MetaMask,https://metamask.io/,#FFF,TRUE,FALSE,TRUE,FALSE,FALSE,TRUE,FALSE,FALSE,FALSE,FALSE,FALSE,TRUE
+metamask,MetaMask,https://metamask.io/,#F66A0A,TRUE,FALSE,TRUE,TRUE,TRUE,TRUE,TRUE,FALSE,FALSE,TRUE,FALSE,TRUE
```
corwintines, line 7: Since the background the MetaMask icon is white on the app stores, I don't know if changing this would fit their branding.
When I checked you couldn't purchase more that 2000USD through Wyre when purchasing directly, so this wouldnt be considered high-volume purchasing in my opinion. Be curious of someone elses take on this. Maybe from Canada I can only see Wyre as an option as well so I will do some research on this.
I wouldn't consider MetaMask a hardware wallet, although you can connect it to hardware wallets, it isn't purely a hardware wallet. I think having it appear as one here would give the wrong impression.
You can purchase with your card from MetaMask using Wyre so that is good to add!
I would say you could can access Defi through the wallet if we consider the mobile applications browser as well.

corwintines (changes requested):

src/data/wallets.csv
kumavis: Hi @corwintines 

We had some trouble updating this data as the terms werent defined.
1. We weren't sure how the brand color was being used so we set it to our signature orange. Can revert to white background
2. Our fiat purchase limits
here's the current limits for bank account purchases:
7-day limit — $25,000
30-day limit — $50,000
60-day limit — $75,000
Yearly limit — $250,000
3. Metamask supports hardware wallets, it is not a hardware wallet

hope that clarifies the situation

wackerow: @kumavis Thank you for clarifying!  So for this case I would argue to not turn "has_hardware" true, as there is no hardware sold by MetaMask. Agreed that generally $2000 has been used for cutoff, and sounds like this qualifies. Given MM has integration with DEX's, I think the DeFi integration is appropriate as well. And sounds like cards are a go with Wyre.

Will update with the adjustments and bring this in, thanks @mapachurro and @kumavis!

wackerow (approved):

src/data/wallets.csv
wackerow: And yeah, the "brand color" would more appropriately be named a "background color" for the transparent logo png. White would be more appropriate here.

wackerow:
@all-contributors please add @mapachurro for ideas and bug fix

PR #3482, opened 2021-07-26
Use Ethereum Mainnet as a proper noun

fulldecent:
As per [EIP-2228](https://eips.ethereum.org/EIPS/eip-2228), it is prefered that Ethereum Mainnet be referred in title case because it is a proper noun representing a singular, static thing.

This PR implements the change on the Ethereum website.


wackerow:
Did not know about this EIP, thanks for sharing this @fulldecent! Especially given this well laid out EIP, I'm in favor of refactoring the site to consistently use this casing on formal uses of Mainnet. @samajammin Any thoughts or opposition to this?

That being said, there are a lot more instances on the site than the ones fixed in this PR. Would be happy to work to refactor them when I get some time, or if you or someone else would like to help with the remainder. Would prefer to hold on merging until we make sure we all agree here, and review the rest of the code base.

samajammin (commented):
LGTM, thanks for the PR (& the EIP)!

Agree with @wackerow that we should update references across the site before merging.

fulldecent:
Updated PR to include the full replacements. I reviewed these manually, this is not a batch job.

minimalsm:
Awesome, thanks for this @fulldecent 🍻

PR #3485, opened 2021-07-27
Implementing security.txt

fredriksvantes:
## Description

Adding security.txt to https://ethereum.org/.well-known/security.txt as per https://datatracker.ietf.org/doc/html/draft-foudil-securitytxt-12

## Related Issue
#3486


fredriksvantes:
Public key is in the following PR: #3486

Aldekein:
https://ethereum.org/security_at_ethereum.org.asc gives 404 error and is not part of this PR.

fredriksvantes:
> https://ethereum.org/security_at_ethereum.org.asc gives 404 error and is not part of this PR.

I've now moved the security_at_ethereum.org.asc into this PR (instead of being in #3486)

JustinDrake:
As a side note I don't have access to security@ethereum.org.

(I do have access to eth2security@ethereum.org but that's used for recruiting security researchers, not for disclosing vulnerabilities.)

samajammin (approved):
Thanks for this!

Issue #3489, opened 2021-07-28
ETHWall description length causing scrolling


minimalsm:
**Describe the bug**

The character length of the description on the WallEth wallet is causing the card to require scrolling to read the whole description. No other wallets have this issue.

**To Reproduce**

1. Go to 'https://ethereum.org/en/wallets/find-wallet/'
2. Scroll down to WallETH wallet card
3. See bug

**Expected behavior**
Description should fit within the card without scrolling.

**Screenshots**

![Screenshot 2021-07-28 at 14 41 30](https://user-images.githubusercontent.com/62268199/127340534-96e14fe9-dade-4e10-906e-f9055b3c111f.png)

**Desktop (please complete the following information):**

- OS: MacOS
- Browser Brave
- Version 1.24.82

**Alternatives Considered**
Limit description length so we're not redesigning every time the new max description length is exceeded?

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

mannekin:
Would like to take this on as my first contribution :)

minimalsm:
Closed by #3569

PR #3496, opened 2021-07-29
Hid overflow of wallet descriptions

mannekin:
WallETH wallet card description overflowed and needed a scrollbar to view. It was the only card that needed this. Setting overflow-y to hidden limits description length so we're not redesigning every time the new max description length is exceeded, per issue #3489 .

### DIFF
--- a/src/components/WalletCard.js
+++ b/src/components/WalletCard.js
@@ -139,7 +139,7 @@ const Description = styled.p`
   margin-bottom: 0.5rem;
   line-height: 140%;
   max-height: 100px;
-  overflow-y: auto;
+  overflow-y: hidden;
 `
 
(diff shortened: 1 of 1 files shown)


wackerow:
@mannekin Hey thanks for the PR... 
I don't think we want to just hide the overflow here, which will lead to some content being inaccessible.
There is a large effort to eventually translate everything as well, making string length variable between languages, and leaving the need for flexibility.

We should just try to trim the wording on the copy found here:
https://github.com/ethereum/ethereum-org-website/blob/f44fc4d69328a7e2abfd8354d1b3f30d64c668af/src/intl/en/page-wallets-find-wallet.json#L65

PR #3512, opened 2021-07-30
Add Crowdin badge to display localization progress

samajammin:
## Description
Add Crowdin badge to display localization progress

## Related Issue
None


samajammin:
@wackerow - FYI if we update the Crowdin project URL we'll want to update this as well.

PR #3536, opened 2021-08-02
Link for Release Banner instead of a href

corwintines:
Switch to using Link component instead of using a href.


pettinarip (approved):

samajammin (approved):
🇬🇧 🚀

Issue #3551, opened 2021-08-04
Update London History AFTER launch


minimalsm:
We'll be using rough estimates for the time and ETH price on [the history page](http://localhost:8000/en/history/) to ensure the content is live *before* the London upgrade. We should go back and edit these with the exact accurate information after London is live.

**Want to contribute?**
We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

Issue #3521, opened 2021-07-30
Update content resources for the London network upgrade


samajammin:
**Is your feature request related to a problem? Please describe.**

We should ensure our resources are up to date re: the upcoming network upgrade - London (August 3-5, 2021): https://blog.ethereum.org/2021/07/15/london-mainnet-announcement/

**Describe the solution you'd like**

- [x] update gas page to reflect to fee market structure: https://ethereum.org/en/developers/docs/gas/
    - [x]  Remove the `incomplete` frontmatter field
- [x] update blocks page to explain new variable block size: https://ethereum.org/en/developers/docs/blocks/
- [x] update transactions page to add new `baseFee` field: https://ethereum.org/en/developers/docs/transactions/
- [x] update history page to show London as a historical update: https://ethereum.org/en/history/
- [x] (depending on #3520) update JSON RPC with any endpoint updates: https://ethereum.org/en/developers/docs/apis/json-rpc/
- [x] add CTA on developer docs section (to upgrade libraries - e.g. referencing https://github.com/ethereum/eth1.0-specs/blob/master/network-upgrades/ecosystem-readiness.md)?

Anything I'm missing here? cc @timbeiko @TrentVanEpps (what's Trent's handle)?

**Describe alternatives you've considered**

None

**Additional context**

https://www.notion.so/efdn/Update-ethereum-org-to-reflect-status-of-London-upgrade-760bc603c2544e11bd63f677ee859af3

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

timbeiko:
@tvanepps is his handle :-)

timbeiko:
There was also the wallet filtering page that needs updating, see: https://github.com/ethereum/ethereum-org-website/issues/2715#issuecomment-890942712

samajammin:
Solved in #3537 #3509 #3552 #3553, closing this out.

PR #3552, opened 2021-08-04
Update History page for London

minimalsm:
## Description
Updated the history page for London update

## Related Issue #3521 #3551


wackerow (approved):
@minimalsm Nice! As we chatted, pulling this in, will update after the actual launch with official ETH price / timestamp.

PR #3569, opened 2021-08-05
Shortened WallETH description

mannekin:
The description of WallETH overflowed the wallet card, making a scrollbar appear. It was the only card that did this. To fix it, I just cut the wording to fit the card, per issue #3489 .

PR #3579, opened 2021-08-07
Fixes Issue #3489

siddharths00:
The problem with the WallETH description was that, it had a property of `overflow-y:auto` on line number 142 in the file `ethereum-org-website\src\components\WalletCard.js` which was the reason that it had a scroll bar. All I did was change the value of overflow-y from `auto` to `inherit`.

### DIFF
--- a/src/components/WalletCard.js
+++ b/src/components/WalletCard.js
@@ -139,7 +139,7 @@ const Description = styled.p`
   margin-bottom: 0.5rem;
   line-height: 140%;
   max-height: 100px;
-  overflow-y: auto;
+  overflow-y: inherit;
 `
 
 const StyledButtonLink = styled(ButtonLink)`


siddharths00:
I've raised a [PR](https://github.com/ethereum/ethereum-org-website/pull/3579/commits/10956377ea18a3e9d0832cfe2974fa2e2a4dcb32) for this particular issue. Please take a look and do let me know if there are any issues with it.

minimalsm:
Can you have a look @wackerow at this as an alternative to shortening the WallETH description? Seems fine at all screen sizes in English. 

https://deploy-preview-3579--ethereumorg.netlify.app/en/wallets/find-wallet/

wackerow:
So the decision to have a scroll bar here was deliberate. Changing to this leaves the trade-off of having no padding/margin between the text and the button below it on some screen sized, which is also not ideal.
image
With some more text this will end up overlapping the button.  To prevent clipping off text, and to maintain some padding to the layout, with a fixed height to the cards themselves, this left us with the scroll option. 
For those reasons I'm more in favor of trimming the text like over on #3569, which also benefits as these strings were designed to be very brief to begin with. Also, a longer term solution is in order, which goes back to discussions about revamping the wallet section as a whole in the not-so-distant future. 

@siddharths00 Going to close this out for now, but feel free to continue discussion on the original issue thread 👍🏻

PR #3594, opened 2021-08-10
Add product ordering instructions to adding products page

minimalsm:
## Description
The previous instructions that new product entries on the site should be added to the bottom are no longer visible on the site. This re-adds those instructions to avoid confusion or perceived favouritism between products.


samajammin (changes requested):
Nice! Please add this to the "Adding developer tools" page as well: https://ethereum.org/en/contributing/adding-developer-tools/

minimalsm:
@samajammin done 🙏

samajammin (approved):

PR #3608, opened 2021-08-11
Stop stale issues closing

minimalsm:
## Description
Stops stale GitHub issues from closing automatically

- Removed option to close stale PRs after 7 days
- Updated stale message to reflect changes
- Stale bot will still mark issues as stale which receive no activity.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -15,3 +15,4 @@ jobs:
           stale-issue-label: "Status: Stale"
           stale-issue-message: "This issue is stale because it has been open 45 days with no activity."
           days-before-stale: 45
+          days-before-close: -1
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -13,6 +13,5 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           stale-issue-label: "Status: Stale"
-          stale-issue-message: "This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days"
+          stale-issue-message: "This issue is stale because it has been open 45 days with no activity."
           days-before-stale: 45
-          days-before-close: 7
```
samajammin, line 18: I don't think we want to remove this - looks like the default value is 7: https://github.com/actions/stale#days-before-close

samajammin:
@minimalsm pushed a small change - I believe we want to set the value to a negative number to avoid closing it out: https://github.com/actions/stale#days-before-close

samajammin (approved):

minimalsm:
Ah never noticed that. Thanks.

PR #3630, opened 2021-08-16
adding tutorials from external sources

### LINKED ISSUE #3620: Allow submission of 3rd party tutorials

I think our developer tutorials section should also display popular tutorials on external sites, like the Hardhat tutorial & scaffold-eth tutorials.

I think we could add a static file, e.g. /data/tutorials.json with some basic info, e.g. 
```
{
  url: "",
  title: "",
  description: "",
  author: "",
  authorGithub: "",
  tags: "",
  skillLevel: "",
  publishDate: ""
}
```

Then display theme as the same cards on https://ethereum.org/en/developers/tutorials/

image

**Want to contribute?**

ahmedprusevic:
## Description

Adding tutorials from 3rd party on tutorials page

Added extra sources on turorials page via JSON file.

## Related Issue

Resolves #3620

### DIFF
--- a/src/data/tutorials.json
+++ b/src/data/tutorials.json
@@ -1,0 +1,13 @@
+[
+  {
+    "url": "https://docs.alchemy.com/alchemy/tutorials/how-to-create-an-nft/how-to-mint-a-nft",
+    "title": "How to Mint an NFT",
+    "description": "This tutorial describes how to mint an NFT on the Ethereum blockchain using Web 3",
+    "author": "Alchemy",
+    "authorGithub": "https://github.com/alchemyplatform",
+    "tags": ["solidity"],
+    "skillLevel": "beginner",
+    "timeToRead": "10",
+    "publishDate": "08/08/2021"
+  }
+]
--- a/src/data/tutorials.json
+++ b/src/data/tutorials.json
@@ -8,6 +8,7 @@
     "tags": ["solidity"],
     "skillLevel": "beginner",
     "timeToRead": "10",
+    "lang": "en",
     "publishDate": "08/08/2021"
   }
 ]
--- a/src/data/tutorials.json
+++ b/src/data/tutorials.json
@@ -5,7 +5,7 @@
     "description": "This tutorial describes how to mint an NFT on the Ethereum blockchain using Web 3",
     "author": "Alchemy",
     "authorGithub": "https://github.com/alchemyplatform",
-    "tags": ["solidity"],
+    "tags": ["solidity", "nfts"],
     "skillLevel": "beginner",
     "timeToRead": "10",
     "lang": "en",
--- a/src/pages/developers/tutorials.js
+++ b/src/pages/developers/tutorials.js
@@ -445,7 +445,7 @@ const TutorialsPage = ({ data, pageContext }) => {
                       mr={`0.5em`}
                     />{" "}
                     <Link to={tutorial.to} hideArrow>
-                      External
+                      <Translation id="page-tutorial-external-link" />
                     </Link>
                   </>
                 )}
--- a/src/pages/developers/tutorials.js
+++ b/src/pages/developers/tutorials.js
@@ -202,25 +202,20 @@ const published = (locale, published) => {
 const TutorialsPage = ({ data, pageContext }) => {
   const intl = useIntl()
   // Filter tutorials by language and map to object
-  const allTutorials = data.allTutorials.nodes
-    .filter((tutorial) =>
-      hasTutorials(pageContext.language)
-        ? tutorial.frontmatter.lang === pageContext.language
-        : tutorial.frontmatter.lang === "en"
-    )
-    .map((tutorial) => ({
-      to:
-        tutorial.fields.slug.substr(0, 3) === "/en"
-          ? tutorial.fields.slug.substr(3)
-          : tutorial.fields.slug,
-      title: tutorial.frontmatter.title,
-      description: tutorial.frontmatter.description,
-      author: tutorial.frontmatter.author,
-      tags: tutorial.frontmatter.tags.map((tag) => tag.toLowerCase().trim()),
-      skill: tutorial.frontmatter.skill,
-      timeToRead: tutorial.timeToRead,
-      published: tutorial.frontmatter.published,
-    }))
+  const internalTutorials = data.allTutorials.nodes.map((tutorial) => ({
+    to:
+      tutorial.fields.slug.substr(0, 3) === "/en"
+        ? tutorial.fields.slug.substr(3)
+        : tutorial.fields.slug,
+    title: tutorial.frontmatter.title,
+    description: tutorial.frontmatter.description,
+    author: tutorial.frontmatter.author,
+    tags: tutorial.frontmatter.tags.map((tag) => tag.toLowerCase().trim()),
+    skill: tutorial.frontmatter.skill,
+    timeToRead: tutorial.timeToRead,
+    published: tutorial.frontmatter.published,
+    lang: tutorial.frontmatter.lang || "en",
+  }))
 
   const externalTutorials = foreignTutorials.map((tutorial) => ({
     to: tutorial.url,
@@ -230,12 +225,22 @@ const TutorialsPage = ({ data, pageContext }) => {
     tags: tutorial.tags.map((tag) => tag.toLowerCase().trim()),
     skill: tutorial.skillLevel,
     timeToRead: tutorial.timeToRead,
-    published: tutorial.publishDate,
+    published: new Date(tutorial.publishDate).toISOString(),
+    lang: tutorial.lang || "en",
   }))
+
+  const allTutorials = []
+    .concat(externalTutorials, internalTutorials)
+    .filter((tutorial) =>
+      hasTutorials(pageContext.language)
+        ? tutorial.lang === pageContext.language
+        : tutorial.lang === "en"
+    )
+    .sort((a, b) => new Date(b.published) - new Date(a.published))
 
   // Tally all subject tag counts
   const tagsConcatenated = []
-  for (const tutorial of allTutorials) {
+  for (const tutorial of internalTutorials) {
     tagsConcatenated.push(...tutorial.tags)
   }
   for (const tutorial of externalTutorials) {
@@ -256,13 +261,13 @@ const TutorialsPage = ({ data, pageContext }) => {
 
   const [state, setState] = useState({
     activeTagNames: [],
-    filteredTutorials: [].concat(externalTutorials, allTutorials),
+    filteredTutorials: allTutorials,
   })
 
   const clearActiveTags = () => {
     setState({
       activeTagNames: [],
-      filteredTutorials: [].concat(externalTutorials, allTutorials),
+      filteredTutorials: allTutorials,
     })
   }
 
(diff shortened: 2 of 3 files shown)


src/pages/developers/tutorials.js
```diff
@@ -220,11 +222,25 @@ const TutorialsPage = ({ data, pageContext }) => {
...
 
+  const externalTutorials = foreignTutorials.map((tutorial) => ({
+    to: tutorial.url,
+    title: tutorial.title,
+    description: tutorial.description,
+    author: tutorial.author,
+    tags: tutorial.tags.map((tag) => tag.toLowerCase().trim()),
+    skill: tutorial.skillLevel,
+    timeToRead: tutorial.timeToRead,
+    published: tutorial.publishDate,
```
minimalsm, line 233: Tutorial's date isn't showing which breaks the tutorial's date ordering. The date needs to be in ISO 8601 format to be shown.

Screenshot 2021-08-16 at 23 12 13

One way to get this working would be to wrap ```tutorial.publishDate``` in the ```externalTutorials``` function 

(Not sure what sort of data sanitisation we'd be doing on this before adding it @samajammin)

```suggestion
    published: new Date(tutorial.publishDate).toISOString(),
```

src/data/tutorials.json
```diff
@@ -0,0 +1,13 @@
...
+    "url": "https://docs.alchemy.com/alchemy/tutorials/how-to-create-an-nft/how-to-mint-a-nft",
+    "title": "How to Mint an NFT",
+    "description": "This tutorial describes how to mint an NFT on the Ethereum blockchain using Web 3",
+    "author": "Alchemy",
+    "authorGithub": "https://github.com/alchemyplatform",
+    "tags": ["solidity"],
+    "skillLevel": "beginner",
+    "timeToRead": "10",
+    "publishDate": "08/08/2021"
+  }
```
minimalsm, line 12: In hindsight, we probably want to add a language field to this and filter external tutorials by language similar to how we filter out the other tutorials.

src/pages/developers/tutorials.js
```diff
@@ -259,9 +278,9 @@ const TutorialsPage = ({ data, pageContext }) => {
     }
 
     // If no tags are active, show all tutorials, otherwise filter by active tag
-    let filteredTutorials = allTutorials
+    let filteredTutorials = [].concat(externalTutorials, allTutorials)
     if (activeTagNames.length > 0) {
-      filteredTutorials = allTutorials.filter((tutorial) => {
+      filteredTutorials = state.filteredTutorials.filter((tutorial) => {
```
minimalsm, line 283: Currently, all tutorials are filtered by date in descending order (this is done by GraphQL for free).

With this the external tutorials will always be displayed first and then the ordered internal tutorials. We're probably going to want to combine external and internal tutorials and order them by date descending (are we best doing this all with GraphQL @wackerow ?).

Screenshot 2021-08-16 at 23 35 24

minimalsm (changes requested):
Hey @ahmedprusevic thanks for the PR. Left a few comments but looking good so far 🦾

One thing I'm crazy about is the placement of the arrow (indicating the tutorial is an external link). Not sure the best way to handle this if anyone else wants to weigh in...

Screenshot 2021-08-16 at 23 43 33

src/pages/developers/tutorials.js
ahmedprusevic: They are one array in state, so it would be possible to sort them by date.

src/pages/developers/tutorials.js
ahmedprusevic: I will check , and comeback with commit

src/data/tutorials.json
wackerow: Yes, agreed. We should add the field here, and also build in a check in `tutorials.js` where if a lang is missing, assume 'en'.

src/pages/developers/tutorials.js
wackerow, line 233: One step further, add `lang` here:
```suggestion
    published: new Date(tutorial.publishDate).toISOString(),
    lang: tutorial.lang ? tutorial.lang : "en",
```

src/pages/developers/tutorials.js
```diff
@@ -220,11 +222,25 @@ const TutorialsPage = ({ data, pageContext }) => {
...
+    to: tutorial.url,
+    title: tutorial.title,
+    description: tutorial.description,
+    author: tutorial.author,
+    tags: tutorial.tags.map((tag) => tag.toLowerCase().trim()),
+    skill: tutorial.skillLevel,
+    timeToRead: tutorial.timeToRead,
+    published: tutorial.publishDate,
+  }))
+
```
wackerow, line 235: So I would rearrange a few things here.  We start in this component by declaring `allTutorials` which is pulled from the graph query. 
1. I would _rename_ this to `internalTutorials`, both at the bottom where the query is declared, and also renaming this declaration to `const internalTutorials = data.internalTutorials.nodes`
2. I would cut this `.filter()` (L206) method from this declaration, and use it below in a moment.
3. Leave the `.map()` function (L211), but add `lang: tutorial.frontmatter.lang,` so we store the language for filtering in a moment
4. Then at this location (L235, after declaring both `internalTutorials` and `externalTutorials`) we can bring them together:
```js
const allTutorials = [...internalTutorials, ...externalTutorials]
```
Then apply the `.filter()` method from above, and add a `.sort()` to sort by date (honestly, reverse date would probably be better so new tutorials are highlighted at top... @minimalsm  thoughts?).

You should then be able to just use the new `allTutorials` array the same way as before with the existing logic, reverting below changes.

Let me know if that makes sense and if you're able to try those changes, or you need a hand getting it over the line. 👍🏻

wackerow (changes requested):
@minimalsm So the ↗ arrow is styled in our `Link.js` component using `&:after` css pseudo element... we may have to play around with this to figure out a better way to signal that these tutorials are external, since the arrow may not play nice with the current setup (it can easily be hidden by passing the `hideArrow={true}` prop with `<TutorialCard>` but then we'd be left without any indication)

minimalsm:
> we may have to play around with this to figure out a better way to signal that these tutorials are external

@wackerow Something like this was what I had in mind. Basically, just using the same after pseudo element at the end of the title.

Screenshot 2021-08-17 at 13 53 23

src/pages/developers/tutorials.js
samajammin: Agreed! Yes to reverse sort to show new tutorials at the top.

ahmedprusevic:
> > we may have to play around with this to figure out a better way to signal that these tutorials are external
> 
> @wackerow Something like this was what I had in mind. Basically, just using the same after pseudo element at the end of the title.
> 
> Screenshot 2021-08-17 at 13 53 23

so should this be a part of this task and issue, or should it be a separate issue?

minimalsm:
> so should this be a part of this task and issue, or should it be a separate issue?

Ideally, we'd keep it as part of the same PR. If you're unsure of how to approach it or need any help you can ping here or on Discord.

Scott1UP:
Hey 👋🏼 

The suggested solutions to this problem are a great start. From a UX perspective, it appears your solution needs to:
• Clearly indicate an External vs. On-site tutorial
• Be easily scannable–users should know at a glance where these links go
• Not disrupt the current design language of the component
• Not introduce another element for users to learn

With those principles in mind, I propose the following:
External Tutorials - Design Idea

In addition to appending the ↗ to the title of the tutorial, a small 'External' link tag could be added to the tutorials details section. The 'External' link colour highlights the difference without being too overbearing or disrupting the overall hierarchy of the component.

samajammin:
Hey @Scott1UP - thanks for weighing in! I'm good with that design change 👍

wackerow:
Yeah I like that idea.

@ahmedprusevic 
So first we need to remove the existing external link arrow with the `hideArrow` prop:
```javascript
<TutorialCard key={tutorial.to} to={tutorial.to} hideArrow={true}>
```

Then I would suggest we provide another property in the "tutorial" object to store which links are external.. something like adding `isExternal: false,` to the `internalTutorials` object, and `isExternal: true,` to the `externalTutorials` object. Then we can add conditional elements to the render function that include a new ↗️ arrow where we need it, and the "External" text.

Worth fixing this in the current PR, let me know if you need a hand! 👍🏻

ahmedprusevic:
> Yeah I like that idea.
> 
> @ahmedprusevic
> So first we need to remove the existing external link arrow with the `hideArrow` prop:
> 
> ```js
> <TutorialCard key={tutorial.to} to={tutorial.to} hideArrow={true}>
> ```
> 
> Then I would suggest we provide another property in the "tutorial" object to store which links are external.. something like adding `isExternal: false,` to the `internalTutorials` object, and `isExternal: true,` to the `externalTutorials` object. Then we can add conditional elements to the render function that include a new  arrow where we need it, and the "External" text.
> 
> Worth fixing this in the current PR, let me know if you need a hand! 👍🏻

Solved, idk weather this is the best implementation, if you have any better ideas please leave me a comment here. @wackerow @minimalsm

minimalsm:
Looking good! Thanks for bearing with us on this @ahmedprusevic - I know this has grown in scope a little 😅😬

I removed the `{" "}` after the link Emoji because it was causing some spacing issues (see below).

Screenshot 2021-08-30 at 13 01 24

I think the last thing is that we should be translating the word External instead of hardcoding it. This can be done by adding it to the [page-developers-tutorials.json](https://github.com/ethereum/ethereum-org-website/blob/dev/src/intl/en/page-developers-tutorials.json) file and referencing it as we do the other tags.

**Like this**
```javascript 
<Link to={tutorial.to} hideArrow>
  <Translation id="page-tutorial-external-link" />
</Link>
```

ahmedprusevic:
> Looking good! Thanks for bearing with us on this @ahmedprusevic - I know this has grown in scope a little 
> 
> I removed the `{" "}` after the link Emoji because it was causing some spacing issues (see below).
> 
> Screenshot 2021-08-30 at 13 01 24
> 
> I think the last thing is that we should be translating the word External instead of hardcoding it. This can be done by adding it to the [page-developers-tutorials.json](https://github.com/ethereum/ethereum-org-website/blob/dev/src/intl/en/page-developers-tutorials.json) file and referencing it as we do the other tags.
> 
> **Like this**
> 
> ```js
> <Link to={tutorial.to} hideArrow>
>   <Translation id="page-tutorial-external-link" />
> </Link>
> ```

Solved @minimalsm

minimalsm:
Awesome, thanks for your work on this @ahmedprusevic 🙏😀

cc: @wackerow - I think this is good to go unless you can see anything else? After we merge this I can create a PR to remove the dummy data and populate it with actual external tutorials.

src/data/tutorials.json
```diff
@@ -0,0 +1,14 @@
+[
+  {
+    "url": "https://docs.alchemy.com/alchemy/tutorials/how-to-create-an-nft/how-to-mint-a-nft",
+    "title": "How to Mint an NFT",
+    "description": "This tutorial describes how to mint an NFT on the Ethereum blockchain using Web 3",
+    "author": "Alchemy",
+    "authorGithub": "https://github.com/alchemyplatform",
+    "tags": ["solidity"],
```
samajammin, line 8:
```suggestion
    "tags": ["solidity", "nfts"],
```

samajammin (approved):
Looks amazing! Thanks for this @ahmedprusevic! Appreciate the work & your perseverance on this 💪

minimalsm:
@all-contributors please add @ahmedprusevic for code

Issue #3072, opened 2021-05-18
Include Carl's energy article on NFT page


ryancreatescopy:
https://blog.ethereum.org/2021/05/18/country-power-no-more/

We should align the calculations if possible and/or just link to it as another resource. 

We can also use this to help form the energy page on ethereum.org

samajammin:
Agreed! Relevant to #3020 as well.

minimalsm:
Closed in #3777 😬

Issue #3020, opened 2021-05-11
Learn page: Ethereum energy consumption


samajammin:
**Is your feature request related to a problem? Please describe.**

It's a well known critique on Ethereum/NFTs/blockchains generally.

**Describe the solution you'd like**

An open explanation of energy consumption. Why there's a trade-off for decentralization and why this won't be an issue with Eth2 moving to PoS.

**Describe alternatives you've considered**

Flesh out / optimize content on https://ethereum.org/en/eth2/vision/

**Additional context**

- Bankless folks offered to co-publish if we can find a writer: https://twitter.com/samonchain/status/1383120909624483841?s=20
- Great thread on Ethereum’s current state & roadmap https://twitter.com/smpalladino/status/1391901504752300033?s=21-Could be a good outline for “ethereum energy consumption” page

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

samajammin:
Solid article by @CarlBeek that could serve as the foundation for this!
https://blog.ethereum.org/2021/05/18/country-power-no-more/

heypran:
We can add a new list item in the `Learn` dropdown, linking to a separate page where we can showcase all the relevant energy consumption-related articles/resources. 

<img width="500" alt="Screenshot 2021-06-13 at 1 43 10 AM" src="https://user-images.githubusercontent.com/15847009/121789748-973b5a80-cbe9-11eb-9746-6d719db89793.png">

We can me the page something similar to the history page as below- 
<img width="832" alt="Screenshot 2021-06-13 at 1 47 51 AM" src="https://user-images.githubusercontent.com/15847009/121789812-44ae6e00-cbea-11eb-8aeb-2ddf85e3ca28.png">

samajammin:
https://twitter.com/0xstark/status/1457823831360368640?s=20

PR #3650, opened 2021-08-18
Add Ethereum energy consumption page

jmcook1186:
Hello,

hope it is OK that I had a bash at this. 

## Description 
This PR includes a draft page on ethereum energy consumption for review and refinement as per [Issue 3020](https://github.com/ethereum/ethereum-org-website/issues/3020). I pulled info from the NFT page as suggested in the notes there as well as the blog by Carl Beek as suggested in the issue conversation thread.

Thanks! 

## Related Issue
#3020 #3072


src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,78 @@
+---
+title: Ethereum Energy Consumption
+description: The basic information you need to understand Ethereum's energy consumption.
+lang: en
+sidebar: true
+---
+
+# Introduction to Ethereum's Energy Consumption {#introduction-to-ethereum-energy-consumption}
+
+Since inception, sustainability has been a core value for Ethereum. However, until recently sustainability has sat uncomfortably alongside two other core values: security and scaleability. This page will explain why building out Ethereum to its current state has required a significant upfront carbon cost and why Ethereum is now on the cusp of becoming a truly low-carbon blockchain without undermining its other core principles.
```
minimalsm, line 10: > Since inception, sustainability has been a core value for Ethereum.

I don't think we should be making generalised statements like this. PoS was always in the plan but calling sustainability a core value seems inaccurate.

```suggestion
Since inception, sustainability has been a core value for Ethereum. However, until recently sustainability has sat uncomfortably alongside two other core values: security and scalability. This page will explain why building out Ethereum to its current state has required a significant upfront carbon cost and why Ethereum is now on the cusp of becoming a truly low-carbon blockchain without undermining its other core principles.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,78 @@
...
+
+This only works if there is a cost associated with minign and some unpredictability about which specific node submits the next block. These conditions are met by imposing proof-of-work (PoW). To be eligible to submit a block of transactions a miner must solve an arbitrary computational puzzle faster than any other miner. This creates competition between miners and cost in the form of energy expenditure. In order to successfully defraud the blockchain, a dishonest node operator would have to consistently win the race to solve the PoW puzzle which is both very unlikely and prohibitively expensive. PoW is therefore a way to secure the network. By requiring proof of computational work, PoW underpins the longest-chain consensus mechanism that ensures validity of the blockchain.
+
+Ethereum has implemented the PoW protocol since its inception. While migration away from PoW has always been a fundamental goal of Ethereum, it has also arguably been the most philosophically and technologically challenging because the viable alternatives all required some compromise in one of Ethereum's core principles.
```
minimalsm, line 18: This is good but should be tackled from a slightly higher level. Seems like it intends to be directed to beginners but them makes assumptions about mempools and blocks. For learn pages, where possible, I'd lean towards not getting into the technicals too much.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,78 @@
...
+
+## PoS energy expenditure
+
+As well as building confidence in the PoS mechanism, the Beacon Chain also enables estimates of Ethereum's post-merge energy usage. A recent [blog post on this site](https://blog.ethereum.org/2021/05/18/country-power-no-more/) suggested that the merge to PoS could result in a 99.95% reduction in total energy use, with PoS being ~2000x more efficient than PoS. The energy-expenditure of Ethereum will be roughly equal to the cost of running a home computer for each node on the network, estimated to be about 5,250,000kWh per year for a network of 10,000 nodes.
```
minimalsm, line 34: > Over time, this prices miners out of the market, tending towards centralisation

- Not sure we can/should make the PoS is more/less centralised than PoW argument...
- Don't think the focus on this matters on this page

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,78 @@
...
+
+It's estimated that Eth2 will allow the network to process between 25,000 and [100,000 transactions per second](https://twitter.com/VitalikButerin/status/1312905884549300224?ref_src=twsrc%5Etfw%7Ctwcamp%5Etweetembed%7Ctwterm%5E1312905886327664640%7Ctwgr%5E%7Ctwcon%5Es2_&ref_url=https%3A%2F%2Fwww.coinspeaker.com%2Fvitalik-buterin-ethereum-layer-2%2F), with [100,000 as the theoretical maximum right now](https://ethereum-magicians.org/t/a-rollup-centric-ethereum-roadmap/4698).
+At the bare minimum, Eth2 will allow 64 times the amount of transactions as today which sits at around 15 transactions. That's the amount of shard chains (extra data and capacity) being introduced ([More on shard chains](/eth2/shard-chains/)). That means we can conservatively estimate how long it will take to process 100,000 transactions so we can compare it to the Visa example above.
```
minimalsm, line 41:
```suggestion
At a minimum, Eth2 will allow 64 times the amount of transactions as today which sits at around 15 transactions. That's the number of [shard chains](/eth2/shard-chains/) (extra data and capacity) being introduced. That means we can conservatively estimate how long it will take to process 100,000 transactions so we can compare it to the Visa example above.
```

minimalsm: Also: we should probably be explaining that shared chains allow this growth (not anything else in the eth2 updates) especially because [we're moving away from the ETH2 branding](https://notes.ethereum.org/@timbeiko/great-renaming)

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,78 @@
...
+
+To put it another way, if Visa handled 140,839,000,000 transactions at a cost of 149 kWh per 100,000 transactions that's 209,850,110 kWh energy consumed for the year.
+
+Eth2 in a single year stands to consume 5,256,000 kWh. With a potential of 788,940,000,000 - 3,153,600,000,000 transactions processed in that time.
+
+_We’ve provided the basic comparison to Visa to baseline your understanding of Eth2 energy consumption against a familiar name. However, in practice, it’s not really correct to compare based on number of transactions. Ethereum’s energy output is time-based. If Ethereum did more or less transactions from one minute to the next, the energy output would stay the same._
+
+_It’s also important to remember that Ethereum does more than just financial transactions, it’s a platform for applications, so a fairer comparison might be to many companies/industries including Visa, AWS and more!_
+
+## Criticism
```
minimalsm, line 66: > Criticism

I don't think the centralisation argument is all that relevant to a user who wants to know about how much energy Ethereum uses. I'd probably lean towards dropping this whole section.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,78 @@
...
+
+## The merge
+
+There is now a functional PoS blockchain called the [Beacon Chain]("https://ethereum.org/en/eth2/beacon-chain/") that has been running since December 2020 that is demonstrating the viability of the PoS protocol. Now that the ["London Upgrade"](https://ethereum.org/en/history/#london) has been successfully implemented, the switch from PoW to Pos (known as "the merge") is the next major update coming to Ethereum with a target date of late 2021 - early 2022. The merge to PoS will also bring associated efficiency gains through "sharding" the blockchain into sub-chains and accelerating the adoption of "roll-ups" (collections of transactions settled cheaply off-chain then pushed to the Ethereum blockchain as a single batch transaction).
```
minimalsm, line 30: > as been successfully implemented, the switch from PoW to Pos (known as "the merge") is the next major update coming to Ethereum with a target date of late 2021 - early 2022.

Altair is before the Merge

minimalsm: > The merge to PoS will also bring associated efficiency gains through "sharding" the blockchain into sub-chains

The merge will happen before shard chains now

minimalsm: > accelerating the adoption of "roll-ups"

Not sure what you mean by this. I don't see the merge and roll ups as being linked?

minimalsm (changes requested):
Hey @jmcook1186 - spoke to in DMs but to re-iterate this is a really strong start ✌️  I've left some comments for suggested changes. 

One change that we'll have to make before publishing this is removing references to 'Eth2' as [the term is being deprecated](https://notes.ethereum.org/@timbeiko/great-renaming). (but we can iterate on this)

jmcook1186:
Ok, I amended the page according to the review comments - agree with everything, of course. 

Also noticed that the calculations of energy consumption per 100k transactions that were inherited from the previous version of the page assumed 10k validators which seems to me to be over an order of magnitude too low. I updated them assuming 200k validators, which i think gives a conservative estimate. Also rewrote this section to make it (hopefully) more readable.

Thanks!

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,64 @@
+---
+title: Ethereum Energy Consumption
+description: The basic information you need to understand Ethereum's energy consumption.
+lang: en
+sidebar: true
+---
+
+# Introduction to Ethereum's Energy Consumption {#introduction-to-ethereum-energy-consumption}
+
+Ethereum is now on the cusp of becoming a truly low-carbon blockchain. Achieving this without sacrificing decentralization has been a major technical challenge for Ethereum developers. This page will explain why building out Ethereum to its current state has required a significant upfront carbon cost and how this will be remediated in future ugrades.
```
minimalsm, line 10: As we agreed in DMs - transparency and showing awareness to the problem early is good

```suggestion
Ethereum's current energy expenditure is too high and unsustainable. Resolving energy expenditure concerns without sacrificing security and decentralization has been a significant technical challenge for Ethereum's development. Let's explore why building Ethereum has had a high environmental impact and how upcoming network upgrades will dramatically change this.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Energy secures the network {energy-secures-the-network}
+
+Transactions on the Ethereum blockchain are validated by [miners](/developers/docs/consensus-mechanisms/pow/mining). Miners bundle together transactions into ordered blocks and add them to the Ethereum blockchain. The new blocks are broadcast to all the other node operators who run the transactions independently and verify that they are valid. Any dishonesty shows up as an inconsistency between different nodes. Honest blocks are added to the blockchain and become an immutable part of history.
```
minimalsm, line 14:
```suggestion
Transactions on the Ethereum blockchain are validated by [miners](/developers/docs/consensus-mechanisms/pow/mining). Miners bundle together transactions into ordered blocks and add them to the Ethereum blockchain. The new blocks get broadcast to all the other node operators who run the transactions independently and verify that they are valid. Any dishonesty shows up as an inconsistency between different nodes. Honest blocks are added to the blockchain and become an immutable part of history.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+This only works if there is a cost associated with mining and some unpredictability about which specific node submits the next block. These conditions are met by imposing [proof-of-work](/developers/docs/consensus-mechanisms/pow/) (PoW). To be eligible to submit a block of transactions a miner must solve an arbitrary computational puzzle faster than any other miner. This creates competition between miners and cost in the form of energy expenditure. In order to successfully defraud the blockchain, a dishonest miner would have to consistently win the PoW race which is both very unlikely and prohibitively expensive. PoW is therefore a way to secure the network.
```
minimalsm, line 16:
```suggestion
Reliance on miner honesty only works if there is a cost associated with mining and unpredictability about which specific node submits the next block. These conditions are met by imposing **[proof-of-work](/developers/docs/consensus-mechanisms/pow/) (PoW)**. To be eligible to submit a block of transactions, a miner must solve an arbitrary computational puzzle faster than any other miner. This creates competition between miners and costs in the form of energy expenditure. To successfully defraud the blockchain, a dishonest miner would have to consistently win the proof-of-work race, which is very unlikely and prohibitively expensive. 
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+Ethereum has implemented the PoW protocol since its inception. While migration away from PoW has always been a fundamental goal of Ethereum, it has also arguably been the most philosophically and technologically challenging because the viable alternatives all required some compromise in one of Ethereum's core principles.
```
minimalsm, line 18:
```suggestion
Ethereum has used proof-of-work since genesis. Migrating off of proof-of-work has always been a fundamental goal of Ethereum. Still, it has been philosophically and technologically challenging because the viable alternatives all required to compromise Ethereum's core principles.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+## PoW energy expenditure
+
+PoW is a very robust way to secure the against dishonest changes to the blockchain, but it is problematic for several reasons. Since the right to mine a block requires solving a computational puzzle, miners can increase their odds of success by investing in more powerful hardware, leading to an arms race of increasingly expensive and power-hungry mining equipment. Ethereum's PoW protocol currently consumes as much energy as a medium-sized country.
```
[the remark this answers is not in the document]
minimalsm:
```suggestion
Proof-of-work is a very robust way to secure against dishonest changes to the blockchain, but it is problematic for several reasons. Since the right to mine a block requires solving a computational puzzle, miners can increase their odds of success by investing in more powerful hardware, leading to an arms race of increasingly expensive and power-hungry mining equipment. Ethereum's PoW protocol currently consumes as much energy as a medium-sized country.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+## PoS
+
+A greener future for Ethereum is already being built in the form of a proof-of-stake (PoS) chain. Under PoS, the arbitrary puzzle solving is not necessary, drastically reducing the energy expenditure required to secure the network. Miners are replaced by validators who perform the same function except that instead of expending their assets up-front in the form of computational work, they stake ETH as collateral against dishonest behaviour. If the validator's node is non-responsive, or a fraudulent block is submitted to the chain, the staked assets can be "slashed", strongly incentivizing honesty and securing the network. Similarly to PoW, in order to maintain a fraudulent blockchain, a validator would require 51% of the total ETH staked in the network. Unlike PoW, however, consensus is not based on the longest chain, but on a mechanism known as ["casper"](https://arxiv.org/abs/1710.09437). Migrating from PoW to PoS eliminates the need to expend energy on arbitrary computations.
```
minimalsm, line 26:
```suggestion
A greener future for Ethereum is already being built in the form of a proof-of-stake (PoS) chain. Under proof-of-stake, arbitrary puzzle-solving is unnecessary. This drastically reduces the energy expenditure required to secure the network. Miners get replaced by validators who perform the same function except that instead of expending their assets up-front in the form of computational work, they stake ETH as collateral against dishonest behaviour. If the validator's node is non-responsive or a fraudulent block gets submitted to the chain, the staked assets can be "slashed", strongly incentivizing honesty and securing the network.

Similarly to proof-of-work, to maintain a fraudulent blockchain, a validator would require 51% of the total ETH staked in the network. However, unlike proof-of-work, consensus is not based on the longest chain, but a mechanism known as ["Casper"](https://arxiv.org/abs/1710.09437). Migrating from proof-of-work to proof-of-stake eliminates the need to expend energy on arbitrary computations.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+## The merge
+
+There is now a functional PoS blockchain called the [Beacon Chain]("https://ethereum.org/en/eth2/beacon-chain/") that has been running since December 2020 that is demonstrating the viability of the PoS protocol. The merge refers to the point in time when Ethereum leaves PoW behind and fully adopts PoS. The target date for this is now late 2021 - early 2022.
```
minimalsm, line 30:
```suggestion
There is now a functional proof-of-stake chain called the [Beacon Chain]("https://ethereum.org/en/eth2/beacon-chain/") that has been running since December 2020 that is demonstrating the viability of the proof-of-stake protocol. The merge refers to the point in time when Ethereum leaves proof-of-work behind and fully adopts proof-of-stake. The merge is expected to happen ~Q4 2021/Q1 2022.
```

minimalsm:
Hey @jmcook1186 made some more edits. 

FYI I've suggested the edits above (You can just click 'Commit suggestion' if you agree with the changes) - we recently released a [style guide](https://ethereum.org/en/contributing/style-guide) where we decided to use proof-of-work over PoW consistently to reduce user cognitive load (i.e. they don't have to remember recently learned terms).

jmcook1186:
Hi, agree with all edits - thanks!

jmcook1186:
Realised my mistake in the energy calculations - there may be 180k validators but there can be many validators per node. Eth R&D Discord suggest 3-4k individual nodes currently, so 10k is conservative estimate for post merge. Calculations still need further validation, but at least that piece makes more sense now. Cheers.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+## PoW energy expenditure
+
+Proof-of-work is a very robust way to secure against dishonest changes to the blockchain, but it is problematic for several reasons. Since the right to mine a block requires solving a computational puzzle, miners can increase their odds of success by investing in more powerful hardware, leading to an arms race of increasingly expensive and power-hungry mining equipment. Ethereum's PoW protocol currently consumes as much energy as a medium-sized country.
+
+## PoS
+
+A greener future for Ethereum is already being built in the form of a proof-of-stake (PoS) chain. Under proof-of-stake, arbitrary puzzle-solving is unnecessary. This drastically reduces the energy expenditure required to secure the network. Miners get replaced by validators who perform the same function except that instead of expending their assets up-front in the form of computational work, they stake ETH as collateral against dishonest behaviour. If the validator's node is non-responsive or a fraudulent block gets submitted to the chain, the staked assets can be "slashed", strongly incentivizing honesty and securing the network.
```
minimalsm, line 26:
```suggestion
A greener future for Ethereum is already being built in the form of a **proof-of-stake (PoS)** chain. Under proof-of-stake, arbitrary puzzle-solving is unnecessary. Removing puzzle-solving drastically reduces the energy expenditure required to secure the network. Miners get replaced by validators who perform the same function except that instead of expending their assets up-front in the form of computational work, they stake ETH as collateral against dishonest behavior. If the validator's node is non-responsive or a fraudulent block gets submitted to the chain, the staked assets can be "slashed", strongly incentivizing honesty and securing the network.
```

minimalsm: > If the validator's node is non-responsive [....] the staked assets can be "slashed"

This isn't true. Offline penalties and slashing aren't the same thing (in both mechanism and severity I believe?) 

@wackerow thoughts on whether we should we remove reference to offline penalties altogether or just reword?

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+Similarly to proof-of-work, to maintain a fraudulent blockchain, a validator would require 51% of the total ETH staked in the network. However, unlike proof-of-work, consensus is not based on the longest chain, but a mechanism known as ["Casper"](https://arxiv.org/abs/1710.09437). Migrating from proof-of-work to proof-of-stake eliminates the need to expend energy on arbitrary computations.
+
+## The merge
+
+There is now a functional proof-of-stake chain called the [Beacon Chain]("https://ethereum.org/en/eth2/beacon-chain/") that has been running since December 2020 that is demonstrating the viability of the proof-of-stake protocol. The merge refers to the point in time when Ethereum leaves proof-of-work behind and fully adopts proof-of-stake. The merge is expected to happen ~Q4 2021/Q1 2022.
```
minimalsm, line 32:
```suggestion
There is a functional proof-of-stake chain called the [Beacon Chain](/eth2/beacon-chain/) that has been running since December 2020 that is demonstrating the viability of the proof-of-stake protocol. The merge refers to the point in time when Ethereum leaves proof-of-work behind and fully adopts proof-of-stake. The merge is expected to happen ~Q2 2022. [More on the merge](/eth2/merge/).
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+## PoS energy expenditure
```
minimalsm, line 34:
```suggestion
## Proof-of-stake energy expenditure {#proof-of-stake-energy}
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+As well as building confidence in the PoS mechanism, the Beacon Chain also enables estimates of Ethereum's post-merge energy usage. A recent [blog post on this site](https://blog.ethereum.org/2021/05/18/country-power-no-more/) suggested that the merge to PoS could result in a 99.95% reduction in total energy use, with PoS being ~2000x more efficient than PoW. The energy-expenditure of Ethereum will be roughly equal to the cost of running a home computer for each node on the network.
```
minimalsm, line 36:
```suggestion
As well as building confidence in the proof-of-stake mechanism, the Beacon Chain also enables estimates of Ethereum's post-merge energy usage. A recent [blog post on the Ethereum blog](https://blog.ethereum.org/2021/05/18/country-power-no-more/) suggested that the merge to proof-of-stake could result in a 99.95% reduction in total energy use, with proof-of-stake being ~2000x more efficient than proof-of-work. The energy expenditure of Ethereum will be roughly equal to the cost of running a home computer for each node on the network.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+_We’ve provided the basic comparison to Visa to baseline your understanding of post-merge Ethereum energy consumption against a familiar name. However, in practice, it’s not really correct to compare based on number of transactions. Ethereum’s energy output is time-based. If Ethereum did more or less transactions from one minute to the next, the energy output would stay the same._
+
+_It’s also important to remember that Ethereum does more than just financial transactions, it’s a platform for applications, so a fairer comparison might be to many companies/industries including Visa, AWS and more!_
+
+## Summary
+
+While Ethereum's energy consumption has historically been substantial, there has been major investment of developer time and intellect into transitioning from energy-hungry to energy-efficient block validation. To quote [Bankless](http://podcast.banklesshq.com/) the best way to conserve the energy being burned by PoW is simply to "turn it off", which is the approach Ethereum has committed to take.
```
minimalsm, line 62:
```suggestion
While Ethereum's energy consumption has historically been substantial, there has been a major investment of developer time and intellect into transitioning from energy-hungry to energy-efficient block validation. To quote [Bankless](http://podcast.banklesshq.com/), the best way to conserve the energy consumed by proof-of-work is simply to "turn it off", which is the approach Ethereum has committed to take.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+![image](energy_use_per_transaction.png)
+
+We can use this data to compare Ethereum to a global service like Visa. 100,000 Visa transactions uses 149kWh of energy<sup>[^2]</sup>. Assuming sharding has been implemented, Ethereum's current transaction rate (15 transactions per second) will be increased by at least 64x (the number of shards) not accounting for additional optimisation from rollups. A realistic estimate for post-merge, sharded Ethereum with rollups is [25000 - 100000](https://twitter.com/VitalikButerin/status/1312905884549300224?ref_src=twsrc%5Etfw%7Ctwcamp%5Etweetembed%7Ctwterm%5E1312905886327664640%7Ctwgr%5E%7Ctwcon%5Es2_&ref_url=https%3A%2F%2Fwww.coinspeaker.com%2Fvitalik-buterin-ethereum-layer-2%2F) transactions per second. We can use this information to estimate a maximum and minimum energy expenditure per 100,000 transactions.
```
minimalsm, line 40:
```suggestion
We can use this data to compare Ethereum to a global service like Visa. 100,000 Visa transactions uses 149kWh of energy<sup>[^2]</sup>. Assuming sharding has been implemented, Ethereum's current transaction rate (15 transactions per second) will be increased by at least 64x (the number of shards), not accounting for additional optimization from rollups. A realistic estimate for post-merge, sharded Ethereum with rollups is [25000 - 100000](https://twitter.com/VitalikButerin/status/1312905884549300224?s=20) transactions per second. We can use this information to estimate a maximum and minimum energy expenditure per 100,000 transactions.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+- 25000 transactions per second.
+- `100,000 / 25000 = 4` seconds to process 100,000 transactions.
+
+We can also estimate Ethereum's energy expenditure per second, making a conservative estimate that there are 10,000 active validators securing the network (there are over 180,000 validators on the eacon chain at the moment, but many validators can operate on a single node. Currently, there are 3000-4000 individual nodes, so 10,000 is a conservative estimate for post-merge):
```
minimalsm, line 45:
```suggestion
We can also estimate Ethereum's energy expenditure per second, making a conservative estimate that 10,000 active validators are securing the network (there are over 180,000 validators on the Beacon Chain at the moment, but many validators can operate on a single node. Currently, there are 3000-4000 individual nodes, so 10,000 is a conservative estimate for post-merge):
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,66 @@
...
+
+This is ~0.4% of the energy used by Visa for the same number of transactions, or a reduction in energy expenditure by a factor of ~225.
+
+Repeating the calculation with the maximum transactions-per-second yields 0.1667 kWh per second which is about 0.1% of the energy expenditure of Visa, or a reduction of ~894x.
+
+_We’ve provided the basic comparison to Visa to baseline your understanding of post-merge Ethereum energy consumption against a familiar name. However, in practice, it’s not really correct to compare based on number of transactions. Ethereum’s energy output is time-based. If Ethereum did more or less transactions from one minute to the next, the energy output would stay the same._
+
+_It’s also important to remember that Ethereum does more than just financial transactions, it’s a platform for applications, so a fairer comparison might be to many companies/industries including Visa, AWS and more!_
+
+## Summary
```
minimalsm, line 60:
```suggestion
## A greener Ethereum (#green-ethereum}
```

minimalsm (changes requested):

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+Ethereum has used proof-of-work since genesis. Migrating off of proof-of-work has always been a fundamental goal of Ethereum. Still, it has been philosophically and technologically challenging because the viable alternatives all required to compromise Ethereum's core principles.
+
+## Proof-of-work energy expenditure {#proof-of-work}
+
+Proof-of-work is a robust way to secure against dishonest changes to the blockchain, but it is problematic for several reasons. Since the right to mine a block requires solving a computational puzzle, miners can increase their odds of success by investing in more powerful hardware. These incentives cause an arms race with miners acquiring increasingly power-hungry mining equipment. Ethereum's proof-of-work protocol currently consumes as much energy as a medium-sized country.
```
corwintines, line 20: We should be adding a source when talking about statistics like consuming comparable amounts of energy as some countries. I know that statistic, but we should at least be providing the source on it in this page.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+## The merge {#the-merge}
+
+There is a functional proof-of-stake chain called the [Beacon Chain](/eth2/beacon-chain/) that has been running since December 2020 that is demonstrating the viability of the proof-of-stake protocol. The merge refers to the point in time when Ethereum leaves proof-of-work behind and fully adopts proof-of-stake. The merge is expected to happen ~Q2 2022. [More on the merge](/eth2/merge/).
+
+## Proof-of-stake energy expenditure {#proof-of-stake-energy}
+
+As well as building confidence in the proof-of-stake mechanism, the Beacon Chain also enables estimates of Ethereum's post-merge energy usage. A recent [blog post on the ethereum.org blog](https://blog.ethereum.org/2021/05/18/country-power-no-more/) suggested that the merge to proof-of-stake could result in a 99.95% reduction in total energy use, with proof-of-stake being ~2000x more efficient than proof-of-work. The energy expenditure of Ethereum will be roughly equal to the cost of running a home computer for each node on the network.
+
+![image](energy_use_per_transaction.png)
```
corwintines, line 36: Question about this image. My understanding was that a block was using around this amount of energy, not a transaction. Am I out to lunch on my understanding? If not, then I think we need a different image.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+We can use this data to compare Ethereum to a global service like Visa. 100,000 Visa transactions uses 149kWh of energy<sup>[^2]</sup>. Assuming sharding has been implemented, Ethereum's current transaction rate (15 transactions per second) will be increased by at least 64x (the number of shards), not accounting for additional optimization from rollups. A realistic estimate for post-merge, sharded Ethereum with rollups is [25000 - 100000](https://twitter.com/VitalikButerin/status/1312905884549300224?s=20) transactions per second. We can use this information to estimate a maximum and minimum energy expenditure per 100,000 transactions.
```
corwintines, line 38: Source for visa stat.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+- 25000 transactions per second.
+- `100,000 / 25000 = 4` seconds to process 100,000 transactions.
+
+We can also estimate Ethereum's energy expenditure per second, making a conservative estimate that 10,000 active validators are securing the network (there are over 180,000 validators on the Beacon Chain at the moment, but many validators can operate on a single node. Currently, there are 3000-4000 individual nodes, so 10,000 is a conservative estimate for post-merge):
```
corwintines, line 43: Is there a page we can link out to to see these stats in real time?

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+`1.44kWh daily usage * 10,000 network nodes = 14,400kWh` per day.
```
corwintines, line 45: Where are we getting this 1.44kWh?

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+If we multiply that by the amount of time it takes to process 100,000 transaction: `0.1667 * 4 = 0.667 kWh`.
+
+This is ~0.4% of the energy used by Visa for the same number of transactions, or a reduction in energy expenditure by a factor of ~225.
+
+Repeating the calculation with the maximum transactions-per-second yields 0.1667 kWh per second which is about 0.1% of the energy expenditure of Visa, or a reduction of ~894x.
+
+_We’ve provided the basic comparison to Visa to baseline your understanding of post-merge Ethereum energy consumption against a familiar name. However, in practice, it’s not really correct to compare based on number of transactions. Ethereum’s energy output is time-based. If Ethereum did more or less transactions from one minute to the next, the energy output would stay the same._
+
+_It’s also important to remember that Ethereum does more than just financial transactions, it’s a platform for applications, so a fairer comparison might be to many companies/industries including Visa, AWS and more!_
```
corwintines, line 56: I think we would need to include some more information on infrastructure necessary to develop if we are going to bring in AWS. Infura, graph, ipfs, etc dont come for free in their energy expenditure.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Proof-of-stake {#proof-of-stake}
+
+A greener future for Ethereum is already being built in the form of a **proof-of-stake (PoS)** chain. Under proof-of-stake, arbitrary puzzle-solving is unnecessary. Removing puzzle-solving drastically reduces the energy expenditure required to secure the network. Miners get replaced by validators who perform the same function except that instead of expending their assets up-front in the form of computational work, they stake ETH as collateral against dishonest behavior. If the validator's node is non-responsive or a fraudulent block gets submitted to the chain, the staked assets can be "slashed", strongly incentivizing honesty and securing the network.
+
+Similarly to proof-of-work, to maintain a fraudulent blockchain, a validator would require 51% of the total ETH staked in the network. However, unlike proof-of-work, consensus is not based on the longest chain, but a mechanism known as ["Casper"](https://arxiv.org/abs/1710.09437). Migrating from proof-of-work to proof-of-stake eliminates the need to expend energy on arbitrary computations.
```
wackerow, line 26: This is all true, but at the risk of getting in over my head, not sure the notes about Casper are relevant here. Casper GHOST votes, or greediest heaviest observable subtree, is part of the consensus rules, but I don't see how that saves any energy over longest-chain... the energy benefit comes from a new way to select who gets to propose a block.
In PoS, we have collateral, so users can be randomized to propose new blocks without needing PoW to account for Sybil resistance. I don't believe the chain selection rules have much to do with this.

I think a powerful thing we could note here would be that an attack on the network with PoW only costs the energy to spin up a ton of miners (ASICs/GPUs can be rented and spun up for a quick attack)... in PoS you risk losing your entire stake if the attack is large enough.

src/content/energy_consumption/index.md
wackerow: > [Digiconomist](https://digiconomist.net/ethereum-energy-consumption/) estimates that Ethereum miners currently consume 44.49 TWh per year which works out to 5.13 gigawatt on a continuing basis. This means that PoS is ~2000x more energy efficient based on the conservative estimates above, which reflects a reduction of at least 99.95% in total energy use.
>
> If energy consumption per-transaction is more your speed, that’s ~35Wh/tx (avg ~60K gas/tx) or about 20 minutes of TV. By contrast, Ethereum PoW uses the equivalent energy of a house for 2.8 days per transaction and Bitcoin consumes 38 house-days worth.

This seems to be referring to per-transaction cost, at an average gas of 60k. 

But on that note, I think these stats are a bit out-dated:

image

This shows per-transaction PoW energy usage in Ethereum as 175kWh, vs the 84kWh shown on the image. Not necessarily the end of the world, but we should at least note a timestamp for that image so people understand it's not necessarily fully up-to-date.

src/content/energy_consumption/index.md
wackerow, line 38:
```suggestion
We can use this data to compare Ethereum to a global service like Visa. 100,000 Visa transactions uses 149kWh of energy<sup>[^1]</sup>. Assuming sharding has been implemented, Ethereum's current transaction rate (15 transactions per second) will be increased by at least 64x (the number of shards), not accounting for additional optimization from rollups. A realistic estimate for post-merge, sharded Ethereum with rollups is [25,000 - 100,000](https://twitter.com/VitalikButerin/status/1312905884549300224?s=20) transactions per second. We can use this information to estimate a maximum and minimum energy expenditure per 100,000 transactions.
```


And the source for this is from the `/nft` page... 

```
#### 1. Visa energy consumption {#fn-1}

[Bitcoin network average energy consumption per transaction compared to VISA network as of 2020, Statista](https://www.statista.com/statistics/881541/bitcoin-energy-consumption-transaction-comparison-visa/)

[Visa financials report Q4 2020](https://s1.q4cdn.com/050606653/files/doc_financials/2020/q4/Visa-Inc.-Q4-2020-Operational-Performance-Data.pdf)
```
Will post as code suggestion below.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,76 @@
...
+
+<InfoBanner emoji=":evergreen_tree:">
+  If you think these stats are incorrect or can be made more accurate, please raise an issue or PR. These are estimates by the ethereum.org team made using publicly accessible information and the current Ethereum roadmap. This doesn't represent an official promise from the Ethereum Foundation. 
+</InfoBanner>
+
+## Further reading {#further-reading}
+
+- [A country's worth of power, no more](https://blog.ethereum.org/2021/05/18/country-power-no-more/) – _Carl Beekhuizen, May 18 2021_
+- [Ethereum Energy Consumption Index](https://digiconomist.net/ethereum-energy-consumption/) – _Digiconomist_
+
```
wackerow, line 70:
```suggestion
## Further reading {#further-reading}

- [A country's worth of power, no more](https://blog.ethereum.org/2021/05/18/country-power-no-more/) – _Carl Beekhuizen, May 18 2021_
- [Ethereum Energy Consumption Index](https://digiconomist.net/ethereum-energy-consumption/) – _Digiconomist_

### Footnotes and sources {#footnotes-and-sources}

#### 1. Visa energy consumption {#fn-1}

[Bitcoin network average energy consumption per transaction compared to VISA network as of 2020, Statista](https://www.statista.com/statistics/881541/bitcoin-energy-consumption-transaction-comparison-visa/)

[Visa financials report Q4 2020](https://s1.q4cdn.com/050606653/files/doc_financials/2020/q4/Visa-Inc.-Q4-2020-Operational-Performance-Data.pdf)

```

src/content/energy_consumption/index.md
wackerow: On that note, a lot of this appears to be duplicated from the `/nft` page... not sure we want repeat the exact same information, will hurt SEO scores. Were we planning on removing this from the NFT page? If not, we should rework this so the content is not the same.

src/content/energy_consumption/index.md
wackerow: Similar note as above regarding duplicated material from the `/nft` page. Would suggest editing or removing this so it's not the exact same info on both pages.

wackerow (changes requested):
@jmcook1186 Thanks for the work on this! Left some comments, let me know if you have any questions or if you disagree. Thanks again!

src/content/energy_consumption/index.md
minimalsm: Nice spot. I think **having** the same information on both pages is optimal but agree that they need to rewrite this to be different enough not to be flagged by the crawlers as spammy.

src/content/energy_consumption/index.md
minimalsm: @jmcook1186

src/content/energy_consumption/index.md
minimalsm: @wackerow correct me if I'm wrong but these are rough deductions — I don't think it's possible to accurately measure this?

minimalsm:
I've added the suggestions :) @wackerow @corwintines 

Only outstanding questions for me:

> @jmcook1186 where are we getting this 1.44kWh?

> @wackerow correct me if I'm wrong but these are rough deductions — I don't think it's possible to accurately measure this?

src/content/energy_consumption/index.md
jmcook1186: https://digiconomist.net/ethereum-energy-consumption indicates Ethereum's (PoW) energy consumption to approximate that of Finland. Will add citation to next commit.

src/content/energy_consumption/index.md
jmcook1186: maybe just state a range with citations - e.g. 

"Estimate of PoW energy consumption used in figure based on [May 2021 data](https://blog.ethereum.org/2021/05/18/country-power-no-more/), at time of writing the same source suggested [175.56 Kwh]((https://digiconomist.net/ethereum-energy-consumption))"

will include in next commit - please delete if you feel it's excessive.

src/content/energy_consumption/index.md
jmcook1186: cool, thanks!

src/content/energy_consumption/index.md
jmcook1186: Beaconscan has real time info on validator count on the home page https://beaconscan.com/ - will link here. I don't have a citation for the number of nodes though - the 10,000 estimate is a conservative estimate borrowed from the NFT page.

src/content/energy_consumption/index.md
jmcook1186: from the nft page. Can we ask the author of that page where the value came from? I have not been able to track down a source.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,76 @@
...
+Proof-of-work is a robust way to secure against dishonest changes to the blockchain, but it is problematic for several reasons. Since the right to mine a block requires solving a computational puzzle, miners can increase their odds of success by investing in more powerful hardware. These incentives cause an arms race with miners acquiring increasingly power-hungry mining equipment. Ethereum's proof-of-work protocol currently consumes as much energy as a medium-sized country.
+
+## Proof-of-stake {#proof-of-stake}
+
+A greener future for Ethereum is already being built in the form of a **proof-of-stake (PoS)** chain. Under proof-of-stake, arbitrary puzzle-solving is unnecessary. Removing puzzle-solving drastically reduces the energy expenditure required to secure the network. Miners get replaced by validators who perform the same function except that instead of expending their assets up-front in the form of computational work, they stake ETH as collateral against dishonest behavior. If the validator's node is non-responsive or a fraudulent block gets submitted to the chain, the staked assets can be "slashed", strongly incentivizing honesty and securing the network.
```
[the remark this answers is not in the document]
jmcook1186: yes, fair point - apologies for the sloppy description. Amended in next commit to:

"If the validator is lazy (absent when they are supposed to fulfill some validator duty) their staked ETH can slowly leak away, while provably dishonest behaviour results in the the staked assets being "slashed". This strongly incentivizes active and honest participation in securing the network."

src/content/energy_consumption/index.md
jmcook1186: fair point - the references to casper were a bit of a hangover from a prior, excessively detailed version. Agree with the deletion. 

Also like the idea of emphasising the additional security benefits of PoS. It's quite a big topic though and hard to summarize concisely. Maybe best just to link out to the comprehensive explanations on vitalik.ca?

minimalsm:
Waiting on another team member to review (early next week?)

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
+---
+title: Ethereum Energy Consumption
+description: The basic information you need to understand Ethereum's energy consumption.
+lang: en
+sidebar: true
+---
+
+Ethereum's current energy expenditure is too high and unsustainable. Resolving energy expenditure concerns without sacrificing security and decentralization has been a significant technical challenge for Ethereum's development. Let's explore why building Ethereum has had a high environmental impact and how upcoming network upgrades will dramatically change this.
```
samajammin, line 8: > Ethereum's current energy expenditure is too high and unsustainable.

Agree with the sentiment here but I'm not sure leading this page with a subjective statement is the right approach. It's "too high" - according to who? It's "unsustainable" - is it? Smells like someones opinion.

I'd suggest focusing more on factual statements. e.g. "Ethereum's current energy expenditure is significant - comparable to the power consumption of a small country". We could potentially link out to sources here (e.g. https://digiconomist.net/ethereum-energy-consumption/ though I'm not sure this is an authoritative source).

"Overall, Ethereum represents around 0.1% of global electricity" might be another useful stat for context:
https://kylemcdonald.github.io/ethereum-emissions/

samajammin, line 8:
```suggestion
Ethereum's current energy expenditure is too high and unsustainable. Resolving energy expenditure concerns without sacrificing security and decentralization is a significant technical challenge and has been a focus of research and development for years. Let's explore why building Ethereum has had a high environmental impact and how upcoming network upgrades will dramatically change this.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Energy secures the network {#energy-secures-the-network}
+
+Transactions on the Ethereum blockchain are validated by [miners](/developers/docs/consensus-mechanisms/pow/mining). Miners bundle together transactions into ordered blocks and add them to the Ethereum blockchain. The new blocks get broadcast to all the other node operators who run the transactions independently and verify that they are valid. Any dishonesty shows up as an inconsistency between different nodes. Honest blocks are added to the blockchain and become an immutable part of history.
+
+Reliance on miner honesty only works if there is a cost associated with mining and unpredictability about which specific node submits the next block. These conditions are met by imposing proof-of-work (PoW). To be eligible to submit a block of transactions, a miner must solve an arbitrary computational puzzle faster than any other miner. Solving this puzzle creates competition between miners and costs in the form of energy expenditure. To successfully defraud the blockchain, a dishonest miner would have to consistently win the proof-of-work race, which is very unlikely and prohibitively expensive.
```
samajammin, line 14:
```suggestion
The ability for any miner to add new blocks only works if there is a cost associated with mining and unpredictability about which specific node submits the next block. These conditions are met by imposing proof-of-work (PoW). To be eligible to submit a block of transactions, a miner must solve an arbitrary computational puzzle faster than any other miner. Solving this puzzle creates competition between miners and costs in the form of energy expenditure. To successfully defraud the blockchain, a dishonest miner would have to consistently win the proof-of-work race, which is very unlikely and prohibitively expensive.
```
"Reliance on miner honesty" doesn't feel correct here. We aren't relying on miner honesty - if a miner proposes an invalid block, any node can ignore that block.

samajammin:
```suggestion
Reliance on miner honesty only works if there is a cost associated with mining and unpredictability about which specific node submits the next block. These conditions are met by imposing [proof-of-work (PoW)](/developers/docs/consensus-mechanisms/pow/). To be eligible to submit a block of transactions, a miner must solve an arbitrary computational puzzle faster than any other miner. Solving this puzzle creates competition between miners and costs in the form of energy expenditure. To successfully defraud the blockchain, a dishonest miner would have to consistently win the proof-of-work race, which is very unlikely and prohibitively expensive.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+Ethereum has used proof-of-work since genesis. Migrating off of proof-of-work has always been a fundamental goal of Ethereum. Still, it has been philosophically and technologically challenging because the viable alternatives all required to compromise Ethereum's core principles.
+
+## Proof-of-work energy expenditure {#proof-of-work}
+
+Proof-of-work is a robust way to secure against dishonest changes to the blockchain, but it is problematic for several reasons. Since the right to mine a block requires solving a computational puzzle, miners can increase their odds of success by investing in more powerful hardware. These incentives cause an arms race with miners acquiring increasingly power-hungry mining equipment. Ethereum's proof-of-work protocol currently has a total annualized power consumption approximately equal to that of Finland <sup>[^1]</sup> and carbon footprint similar to Switzerland<sup>[^1]</sup>.
```
samajammin, line 20:
```suggestion
Proof-of-work is a robust way to secure the network and enforce honest changes to the blockchain, but it is problematic for several reasons. Since the right to mine a block requires solving a computational puzzle, miners can increase their odds of success by investing in more powerful hardware. These incentives cause an arms race with miners acquiring increasingly power-hungry mining equipment. Ethereum's proof-of-work protocol currently has a total annualized power consumption approximately equal to that of Finland <sup>[^1]</sup> and carbon footprint similar to Switzerland<sup>[^1]</sup>.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Proof-of-stake {#proof-of-stake}
+
+A greener future for Ethereum is already being built in the form of a **proof-of-stake (PoS)** chain. Under proof-of-stake, arbitrary puzzle-solving is unnecessary. Removing puzzle-solving drastically reduces the energy expenditure required to secure the network. Miners get replaced by validators who perform the same function except that instead of expending their assets up-front in the form of computational work, they stake ETH as collateral against dishonest behavior. If the validator is lazy (absent when they are supposed to fulfill some validator duty) their staked ETH can slowly leak away, while provably dishonest behaviour results in the the staked assets being "slashed". This strongly incentivizes active and honest participation in securing the network.
```
samajammin, line 24:
```suggestion
A greener future for Ethereum is already being built in the form of a [**proof-of-stake (PoS)** chain](/eth2/beacon-chain/). Under [proof-of-stake](/developers/docs/consensus-mechanisms/pos/), arbitrary puzzle-solving is unnecessary. Removing puzzle-solving drastically reduces the energy expenditure required to secure the network. Miners get replaced by validators who perform the same function except that instead of expending their assets up-front in the form of computational work, they stake ETH as collateral against dishonest behavior. If the validator is lazy (offline when they are supposed to fulfill some validator duty) their staked ETH can slowly leak away, while provably dishonest behavior results in the the staked assets being "slashed". This strongly incentivizes active and honest participation in securing the network.
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+Similarly to proof-of-work, to maintain a fraudulent blockchain, a validator would require 51% of the total ETH staked in the network. However, unlike on proof-of-work, where the potential loss of a failed attack is only the cost of generating the hash power needed to mine, on proof-of-stake, the possible loss of an attack is the entire amount of ETH used as collateral. This disincentive structure allows for network security with proof-of-stake while eliminating the need to expend energy on arbitrary computations. Detailed explanations of the network security under PoS can be found [here](https://vitalik.ca/general/2017/12/31/pos_faq.html).
```
samajammin, line 26:
```suggestion
Similarly to proof-of-work, a validator would require 51% of the total ETH staked in the network in order to maintain a fraudulent blockchain. However, unlike on proof-of-work, where the potential loss of a failed attack is only the cost of generating the hash power needed to mine, on proof-of-stake, the possible loss of an attack is the entire amount of ETH used as collateral. This disincentive structure allows for network security with proof-of-stake while eliminating the need to expend energy on arbitrary computations. Detailed explanations of the network security under PoS can be found [here](https://vitalik.ca/general/2017/12/31/pos_faq.html).
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Proof-of-stake energy expenditure {#proof-of-stake-energy}
+
+As well as building confidence in the proof-of-stake mechanism, the Beacon Chain also enables estimates of Ethereum's post-merge energy usage. A recent [blog post on the ethereum.org blog](https://blog.ethereum.org/2021/05/18/country-power-no-more/) suggested that the merge to proof-of-stake could result in a 99.95% reduction in total energy use, with proof-of-stake being ~2000x more efficient than proof-of-work. The energy expenditure of Ethereum will be roughly equal to the cost of running a home computer for each node on the network.
+
+![image](energy_use_per_transaction.png)
+
+<p style="text-align: center;"><small><i>Estimate of PoW energy consumption per tx used in figure based on [May 2021 data](https://blog.ethereum.org/2021/05/18/country-power-no-more/), at time of writing the same source suggested up to [175.56 Kwh]((https://digiconomist.net/ethereum-energy-consumption)) </i></small></p>
```
samajammin, line 38:
```suggestion
<p style="text-align: center;"><small><i>Estimate of PoW energy consumption per tx used in figure based on <a href="https://blog.ethereum.org/2021/05/18/country-power-no-more/" target="_blank" rel="noopener noreferrer">May 2021 data</a>, at time of writing the same source suggested up to <a href="https://digiconomist.net/ethereum-energy-consumption" target="_blank" rel="noopener noreferrer">175.56 Kwh</a></i></small></p>
```

samajammin: These links weren't working - can't use markdown within HTML.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+Let's compare these numbers to a service such as Visa. 100,000 Visa transactions uses 149kWh of energy<sup>[^2]</sup>. Assuming sharding has been implemented, Ethereum's current transaction rate (15 transactions per second) will be increased by at least 64x (the number of shards), not accounting for additional optimization from rollups. A realistic estimate for post-merge, sharded Ethereum with rollups is [25,000 - 100,000](https://twitter.com/VitalikButerin/status/1312905884549300224?s=20) transactions per second. We can use this information to estimate a maximum and minimum energy expenditure per 100,000 transactions.
+
+- 25,000 transactions per second.
+- `100,000 / 25,000 = 4` seconds to process 100,000 transactions.
+
+We can also estimate Ethereum's energy expenditure per second, making a conservative estimate that 10,000 active validators are securing the network (there are over [250,000 validators on the Beacon Chain](https://beaconscan.com/) at the moment, but many validators can operate on a single node. Currently, there are estimated to be 3000-4000 individual nodes, so 10,000 is a conservative estimate for post-merge):
```
samajammin, line 45:
```suggestion
We can also estimate Ethereum's energy expenditure per second, making a conservative estimate that 10,000 active validators are securing the network (there are over [250,000 validators on the Beacon Chain](https://beaconscan.com/) at the moment, but many validators can operate on a single node. Currently, there are estimated to be 3,000-4,000 individual nodes, so 10,000 is a conservative estimate for post-merge):
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+`1.44kWh daily usage * 10,000 network nodes = 14,400kWh` per day.
+There are 86,400 seconds in a day, so `14,400 / 86,400 = 0.1667 kWh` per second.
+
+If we multiply that by the amount of time it takes to process 100,000 transaction: `0.1667 * 4 = 0.667 kWh`.
+
+This is ~0.4% of the energy used by Visa for the same number of transactions, or a reduction in energy expenditure by a factor of ~225.
```
samajammin, line 52:
```suggestion
This is ~0.4% of the energy used by Visa for the same number of transactions, or a reduction in energy expenditure by a factor of ~225 compared to Ethereum's current proof-of-work network.
```
I assume this is comparing to PoW? I think we should make that clear.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+Repeating the calculation with the maximum transactions-per-second yields 0.1667 kWh per second which is about 0.1% of the energy expenditure of Visa, or a reduction of ~894x.
+
+_It's not entirely accurate to compare based on number of transactions as Ethereum's energy usage is time-based. The energy usage of Ethereum is the same in 1 minute regardless if it does 1 or 1,000 transactions._
```
samajammin, line 56:
```suggestion
_Note: it's not entirely accurate to compare based on number of transactions as Ethereum's energy usage is time-based. The energy usage of Ethereum is the same in 1 minute regardless if it does 1 or 1,000 transactions._
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+_We must also consider that Ethereum isn't limited to simple financial transactions but is also a complete platform built for smart contracts and decentralized applications._
```
samajammin, line 58: I feel like we could merge this sentence into the one above, as a set of italicized disclaimers.

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+## A greener Ethereum {#green-ethereum}
+
+While Ethereum's energy consumption has historically been substantial, there has been a major investment of developer time and intellect into transitioning from energy-hungry to energy-efficient block validation. To quote [Bankless](http://podcast.banklesshq.com/), the best way to conserve the energy consumed by proof-of-work is simply to "turn it off", which is the approach Ethereum has committed to take.
+
+<InfoBanner emoji=":evergreen_tree:">
+  If you think these stats are incorrect or can be made more accurate, please raise an issue or PR. These are estimates by the ethereum.org team made using publicly accessible information and the current Ethereum roadmap. This doesn't represent an official promise from the Ethereum Foundation.
```
samajammin, line 65:
```suggestion
  If you think these stats are incorrect or can be made more accurate, please raise an issue or PR. These are estimates by the ethereum.org team made using publicly accessible information and the current Ethereum roadmap. These statements don't represent an official promise from the Ethereum Foundation. 
```

src/content/energy_consumption/index.md
```diff
@@ -0,0 +1,90 @@
...
+While Ethereum's energy consumption has historically been substantial, there has been a major investment of developer time and intellect into transitioning from energy-hungry to energy-efficient block validation. To quote [Bankless](http://podcast.banklesshq.com/), the best way to conserve the energy consumed by proof-of-work is simply to "turn it off", which is the approach Ethereum has committed to take.
+
+<InfoBanner emoji=":evergreen_tree:">
+  If you think these stats are incorrect or can be made more accurate, please raise an issue or PR. These are estimates by the ethereum.org team made using publicly accessible information and the current Ethereum roadmap. This doesn't represent an official promise from the Ethereum Foundation. 
+</InfoBanner>
+
+## Further reading {#further-reading}
+
+- [A country's worth of power, no more](https://blog.ethereum.org/2021/05/18/country-power-no-more/) – _Carl Beekhuizen, May 18 2021_
+- [Ethereum Energy Consumption Index](https://digiconomist.net/ethereum-energy-consumption/) – _Digiconomist_
```
samajammin, line 71: Let's add https://kylemcdonald.github.io/ethereum-emissions/ as well

samajammin (changes requested):
Solid work! Really appreciate this PR 💚 

I've suggested some small tweaks & requested a few changes. A few more things in addition that perhaps @minimalsm we can quickly make after merging.
- We should (probably?) add nav/footer links to this page
- We need to add a breadcrumbs string
- filename should be energy-consumption not energy_consumption to keep URLs consistent

minimalsm (approved):

corwintines (approved):

skylarweaver (commented):
Some thoughts

PR #3653, opened 2021-08-18
Fix inconsistencies on proof-of-work page

minimalsm:
## Description
Fixed inconsistencies with cycling back-and-forth between terms: Proof of work, proof-of-work, PoW, etc...

### DIFF
--- a/src/content/developers/docs/consensus-mechanisms/pow/index.md
+++ b/src/content/developers/docs/consensus-mechanisms/pow/index.md
@@ -8,15 +8,15 @@ incomplete: true
 
 Ethereum, like Bitcoin, currently uses a consensus protocol called [Proof-of-work (PoW)](https://en.wikipedia.org/wiki/Proof_of_work). This allows the nodes of the Ethereum network to agree on the state of all information recorded on the Ethereum blockchain, and prevents certain kinds of economic attacks.
 
-Over the next few years, proof-of-work will be phased out in favour of [proof of stake](/developers/docs/consensus-mechanisms/pos). This will also phase out mining from Ethereum. For more details on timing, check the progress of the [Eth2 merge upgrade](/eth2/merge/).
+Over the next few years, proof-of-work will be phased out in favour of [Proof-of-stake (PoS)](/developers/docs/consensus-mechanisms/pos). This will also phase out mining from Ethereum. For more details on timing, check the progress of the [Eth2 merge upgrade](/eth2/merge/).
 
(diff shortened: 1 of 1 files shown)


wackerow (changes requested):
We've been taking the approach on the page where the hyphens here are dependent on its location in the sentence... Perhaps @ryancreatescopy could help enlighten here?? 😄  But my understanding was:
- Not capitalized unless it starts the sentence
- Hyphenated only if it's in the middle of a sentence
- Not hyphenated when it stands by itself, or starts/ends a sentence

And I would overall encourage the use of the 'PoW' and 'PoS' shorthand as it's quite common in this space, and can help tidy things up, but that's just my opinion. The convention I usually try to follow there is introducing the abbreviation in parentheses the first time the term is used on the page, then just just the abbreviated form. 

Overall I'm fairly indifferent other than the preference for consistency. Would just love to get this ironed out one way or the other so we can document and add to a style guide to help clean it up going forward.

minimalsm:
> And I would overall encourage the use of the 'PoW' and 'PoS' shorthand as it's quite common in this space

Hmm... depends on who our audience is. 

Acronyms reduce sentence size but they're just a layer of abstraction. Readers need to mentally expand recently learned acronyms to the full term. For example, readers convert `PoW` to `proof-of-work` in their heads so the shorter acronym actually takes a little longer to process than the full term.

So I guess the question is: do we expect that many readers are learning these acronyms for the first time? Is there a way to find this out? If not, I'd lean to keeping it as is.

samajammin:
@minimalsm @wackerow so did we decide to always use the hyphenated versions moving forward?

wackerow:
Not sure anything was decided. I believe the point of hyphenating is to bundle the terms together when placed in the middle of sentences to eliminate confusion over surrounding context. All this being said, this is not a hill I'm willing to die on. I just crave some consistency, so whatever way we choose, let's do it. 

As for PoW and PoS abbrevs, again not a hill I'm dying on, just wanted to suggest utilizing them to clean up what are otherwise clearly messy terms to use. @minimalsm Feel free to use your judgment here. 👍🏻

samajammin:
I believe @minimalsm proposed always using hyphenated to help simplify (though it may not be truly grammatically correct) but I'll let him speak to it.

samajammin:
I also agree with you @wackerow - once we define the acronyms (PoW & PoS) I'm fine with using those for simplicity.

samajammin (approved):

PR #3657, opened 2021-08-18
Remove Matic Network link

minimalsm:
## Description
The Matic Network page has been changed (https://matic.network/) to re-direct to Polygon's homepage so we had two adjacent links landing on the same page. This PR removes the second link. 

## Other solutions
Not opposed to removing reference to Matic altogether (Polygon have now removed reference to Matic on their homepage) if anyone has any opinions on that.


jley81 (approved):

wackerow (approved):
Nice, I think you're right that we should tend towards phasing it out. I think this is a nice next step to keep the mention of Matic in there for anyone who may not yet be aware of this, but eventually I'm sure we can just remove mention of Matic all together. 👍🏻

Issue #3670, opened 2021-08-19
Add link to the Web3Bridge website to the site


lukassim:
**Is your feature request related to a problem? Please describe.**

Web3Bridge is an initiative, focused on identifying, training and supporting Ethereum developers from African countries.
https://www.web3bridge.com/

**Describe the solution you'd like**

Add a link to their website and a quick summary to the site to help raise awareness about the initiative.

minimalsm:
Closed by #3671

PR #3671, opened 2021-08-19
Add Web3Bridge to the site

lukassim:
## Description

We should add a section about Web3Bridge to the site to help raise awareness and inform more aspiring African developers to join.

However, there doesn't seem to be an ideal place for this on the community page as it is. I've added it to somewhere passibly relevant for now.

Moving forward, would make a lot of sense to think about adding a local communities section to the community page or something similar, if we are looking to add more local initiatives like this.

## Related Issue

#3670


minimalsm:
Thanks Luka 🎉

Issue #3434, opened 2021-07-16
Text on Eth2 Vision Trilemma Diagram cut off in Brave Browser on Linux


sharon-wang:
**Describe the bug**

Text in the Trilemma Diagram (`DECENTRALIZATION`, `SCALABILITY`) is getting cut off.

**To Reproduce**

Only observed on Linux machine on Brave Browser (looks fine in Firefox). Did not see this problem on Mac in Brave.

1. Go to https://ethereum.org/en/eth2/vision/
2. Scroll down to 'The challenge of decentralized scaling' section
3. See error: 'DECENTRALIZATION' with 'N' cut off

**Expected behavior**

Text in Trilemma diagram should not get cut off. When the `DECENTRALIZATION` or `SCALABILITY` circles in the diagram are selected, their text gets even more cut off (see screenshots below).

**Screenshots**

![Untitled](https://user-images.githubusercontent.com/25834218/126009132-ee524522-b33e-4ecc-80c4-7a2cbceb3a7d.png)
![Untitled2](https://user-images.githubusercontent.com/25834218/126009158-e143d1df-b883-4e85-828b-1d19a900751e.png)

**Desktop (please complete the following information):**

- OS: Linux Ubuntu
- Browser: Brave
- Version 1.25.68 Chromium: 91.0.4472.77

**Want to contribute?**

I'd like to contribute! I'm thinking to just increase the width of the `viewBox` of the SVG.

minimalsm:
Tried on two different macs and I'm unable to replicate this. 

@pettinarip do you have Brave on your Linux machine? If so, could you see if you get the same issue?

enriavil1:
I tried on my Mac and also couldn't replicate this. Maybe we could increase the div size by a little bit, so they fit better, though we might have to see the chain reaction on other devices.

> Tried on two different macs and I'm unable to replicate this.
> 
> @pettinarip do you have Brave on your Linux machine? If so, could you see if you get the same issue?

enriavil1:
I was thinking maybe by adding some padding it would help by when you click on the text

enriavil1:
I did it in this code. could you please try it on ur device? @sharon-wang

samajammin:
Thanks for reporting this @sharon-wang!

> Tried on two different macs and I'm unable to replicate this.

[...]

PR #3744, opened 2021-08-29
tried fixing issue of cutting words through padding #3434

enriavil1:
Adding padding trying to solve the problem of words cutting off

added padding of 0.5rem so we don't change the width of the SVG
## Related Issue
#3434


minimalsm:
Hey @enriavil1 thanks for having a go at this one. @sharon-wang has noted in #3434 that this doesn't fix the problem they were having. The issue is now fixed by #3761.

minimalsm:
@all-contributors please add @enriavil1 for ideas

enriavil1:
> Hey @enriavil1 thanks for having a go at this one. @sharon-wang has noted in #3434 that this doesn't fix the problem they were having. The issue is now fixed by #3761.

I have noticed. Thank you for the help!

PR #3761, opened 2021-08-30
Increase Trilemma viewBox width to fit text on Linux

sharon-wang:
## Description
In Brave Browser on Linux, some text in the Trilemma diagram is cut off (see #3434).

This PR increases the width of the viewBox for the Trilemma diagram SVG, so that the text does not get cut off, as seen in these screenshots:
image
image

## Related Issue
Fixes: #3434. 

## Other
I'll need a hand with testing on Mac and Windows (and perhaps other browsers) to ensure that it still looks reasonable on those platforms/browsers :slightly_smiling_face:

### DIFF
--- a/src/components/Trilemma.js
+++ b/src/components/Trilemma.js
@@ -282,7 +282,7 @@ const Trilemma = () => {
       <Triangle
         width="540"
         height="620"
-        viewBox="-100 100 810 915"
+        viewBox="-100 100 850 915"
         fill="none"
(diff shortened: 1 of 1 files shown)


minimalsm:
Hey @sharon-wang. I have tested this quite extensively on BrowerStack and it is working as expected on all devices tested.

Thanks 🙏! Merging now.

minimalsm:
@all-contributors please add @sharon-wang for bugs

Issue #3790, opened 2021-09-01
Add tutorial modal unwanted scrolling


minimalsm:
**Describe the bug**
The submit tutorial modal allows scrolling beneath the modal.

**To Reproduce**

1. Go to 'https://ethereum.org/en/developers/tutorials/' on mobile
2. Click on 'Submit a tutorial'
3. See error

**Expected behavior**
The modal should be scrollable, the background should not.

**Screenshots**
[Gif of problem](https://media.discordapp.net/attachments/727898649006309377/882653590253080626/tutorials.gif)

**Smartphone (please complete the following information):**

- Device: iPhone X
- Browser: Brave

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

minimalsm:
@ndrozd

SofianeBoumedine:
Hi, I fix we can fix this issues : `
![Uploading brave_irVstb0ui7.png…]()
`

JWLangford:
@minimalsm is this issue still accepting pull requests?

PR #3799, opened 2021-09-02
Add tutorial modal unwanted scrolling #3790

ndrozd:
Additianlly fixes modal positioning on small screens.

## Description

Modal wasn't positioned at the center of screen and was out of screen on mobile, I've fixed that.


minimalsm:
Hey @ndrozd thanks for working on this.

For me, this doesn't look quite right yet. The modal shouldn't meet the edges of the screen like it currently is (see screenshot below from the deploy preview).

Screenshot 2021-09-07 at 13 38 43

The current (screenshot below) tutorial modal should be given some more height without removing the margin at the top of the screen.

Screenshot 2021-09-07 at 13 38 27

My thoughts initially were something like this:

Screenshot 2021-09-07 at 13 38 17

This keeps the top margin and the scrolling behaviour (desirable) but allows the second option to 'peek' out, signalling to users that there is another option to select (which I think the current production version fails at).

minimalsm:
@Scott1UP would appreciate some design eyes on the approach I outlined above

Scott1UP:
Hi @minimalsm and @ndrozd 

Great suggestions and work so far, as always. 

As this is a web-based modal, using a fullscreen format for this (as shown in the initial screenshot would most likely fail most users mental model of how this should behave. 

I've put together a rough design for how this could look. I've also spotted a couple of other tweaks I'd like to suggest; the current method of adding a link after the sentence has finished looks odd. If this is a convention for a reason, then please ignore this suggestion. Second, is the margin around the buttons on the cards is off–I've also adjusted this. 

Current proposed: 
Current-proposed-design-with-button-margin-offset

Newly proposed:
New-design

Let me know your thoughts / questions.

minimalsm:
Sorry, thought I commented on this @Scott1UP @ndrozd 

Yeah I agree with the changes proposed in terms of design. Obviously we don't want to hardcode any specific units to allow this to scale across multiple devices but I think this should work! @ndrozd feel free to ping me here or the team on discord if you need a hand on this one.

minimalsm:
cc: @wackerow

ndrozd:
@minimalsm the thing is that I've removed the double scrolling (nested one) from this modal at all and now it blocks the body from scrolling when popup is open. In result you have a very natural "regular page" feeling when popup is open.

ndrozd:
image

Fixed, current state of modal attached.

wackerow (approved):
@ndrozd Thank you for the help on this! This looks a lot better than it was. 

I see how you removed the nested scrolling in favor of moving the whole modal. Curious what you think @Scott1UP , but I'm not necessarily against this as it does give a strong indicator that there is more content below, it does not truncate any copy, and it doesn't simultaneously scroll the background. 

@minimalsm Okay with these changes if you are 👍🏻

minimalsm:
I'm not keen on having a modal cover this much of the screen. Aesthetically it isn't great (which I can live with) but I think the lack of screen real estate to tap to close the modal might be a slight accessibility problem? 

Screenshot 2021-09-29 at 09 39 48

Scott1UP:
Hey @ndrozd @wackerow 👋🏼 

I'm leaning toward the same view as @minimalsm on this one; it wasn't immediately clear to me that the entire modal would load and therefore become a new 'screen'. If this is the intended behaviour we want, why have it be a modal at all? IMO, the modal window should be just that–a window. 

I don't think this is a good precedent to set going forward for modal behaviour, as we wouldn't want to have this replicated in other areas of the site with lengthier content.

ndrozd:
How do we prevent a nested scroll inside a modal then? 

I feel like we're creating too much of a redundant "tweaks" on mobile by leaving a lot of space on the sides. It's too narrow for users to click there, and also it shrinks the much-needed horizontal space for content. IMO content-rich popups on mobile shouldn't have backdrops, it should be just a page-like experience.

minimalsm:
@ndrozd sorry to be clear I think the spacing on the sides is fine. The lack of space on the bottom here is the UX problem for me. 

Screenshot 2021-09-29 at 11 42 49

ndrozd:
Yeah, I've make it the same as the top one already, you just need to scroll to it. 
I just feel like having to scroll to this gap is better than having to scroll the nested scroll to read the content.

wackerow:
Want to quickly recap the issue at hand to keep organized. @minimalsm posted on #3790 with an issue related to the "Add new tutorial" modal, where the _background_ would still scroll behind the modal. This should be the primary objective here, and is currently solved in this PR, but the changes that make the entire modal move, and extend off the page, are in question now. 

With the suggested design work, I agree it would be nice to implement this polished layout, as the existing modal appears cramped. Looking at this again, it do think it's more ideal to scroll only the content of the modal, this way the `X` close button at the top-right is always accessible on the screen.

So the height of this modal should fit its contents, unless there is enough to fill the screen, and then the _max_ height of the modal should end with the bottom margin equal to the top. In this design, overflowing content within the modal is simply scrollable (again, without scrolling the background, and should only be in the `y` direction).

Open to hear more thoughts, but otherwise @ndrozd, would you mind updating the PR based on this discussion? 🙏🏻

wackerow (changes requested):
(See above)

minimalsm:
Hey @ndrozd, feel free to ping here if you need any help on this 😀👍

wackerow:
Hey @ndrozd, hadn't heard back and this was just patched in #4980, so going to close this out. Appreciate the help getting this going.

PR #3809, opened 2021-09-03
Add date of London upgrade to Gas and Fees page

skylarweaver:
Now that the London upgrade is live, we should make it clear that the `Prior to the London Upgrade` section is now describing the *past*, and the `After the London Upgrade` is now describing the *current* Gas and Fees method. I suggest making this clearer by adding the date that the london upgrade went live. Without the date, if you didn't know what the London upgrade was live or if/when it would go live, you'd be unsure which of those Gas and Fees methods is the *current* method, without researching more about the London upgrade.

Simple minor change, feel free to modify/reject if not ideal.

### DIFF
--- a/src/content/developers/docs/gas/index.md
+++ b/src/content/developers/docs/gas/index.md
@@ -24,7 +24,7 @@ Gas fees are paid in Ethereum's native currency, ether (ETH). Gas prices are den
 
 ## Prior to the London Upgrade {#pre-london}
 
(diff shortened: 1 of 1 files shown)


minimalsm:
Thanks for the PR Skylar, this is definitely a lot better.

minimalsm:
@all-contributors please add @skylarweaver for docs

minimalsm (approved):

PR #3812, opened 2021-09-04
Add API to Glossary

rootqa:
Add API and DAG information

### DIFF
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -28,16 +28,16 @@ An object containing an [address](#address), balance, [nonce](#nonce), and optio
 
 Most generally, this represents an [EOA](#eoa) or [contract](#contract-accouint) that can receive (destination address) or send (source address) [transactions](#transaction) on the blockchain. More specifically, it is the rightmost 160 bits of a [Keccak hash](#keccak-256) of an [ECDSA](#ecdsa) [public key](#public-key).
 
-### application programming interface
-
-An Application Programming Interface (API) is a set of definitions for how to use a piece of software. An API sits between an application and a web server, and facilitates the transfer of data between them.
-
 ### application binary interface (ABI) {#abi}
 
 The standard way to interact with [contracts](#contract-account) in the Ethereum ecosystem,
 both from outside the blockchain and for contract-to-contract interactions.
 
 <DocLink to="/developers/docs/smart-contracts/compiling/#web-applications" title="ABI" />
+
+### application programming interface {#api}
+
+An Application Programming Interface (API) is a set of definitions for how to use a piece of software. An API sits between an application and a web server, and facilitates the transfer of data between them.
 
 ### assert {#assert}
 
@@ -126,10 +126,6 @@ A crosslink provides a summary of a shard's state. It's how [shard](#shard) chai
 <Divider />
 
 ## D {#section-d}
-
-### DAG
-
-DAG stands for Directed Acyclic Graph and without going into too many details, it is a crucial file in ETH and ETC mining. DAG file size grows over time, exactly every 30.000 blocks (Ethereum) or every 100 hours to be precise.
 
 ### Decentralized Autonomous Organization (DAO) {#dao}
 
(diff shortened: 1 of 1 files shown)


minimalsm:
Thanks for the PR @rootqa 🙏  

@samajammin @wackerow any thoughts as to adding these terms to the glossary?

My thoughts are:

#### DAG
- We don't use the term DAG anywhere on the site. 
- Combine this with the merge happening in ~3/4 months and I don't think it makes sense to add

#### API
- Don't mind adding API
> Is the term/definition directly relevant to Ethereum?
- Only negative for me is that it is not specific to Ethereum as per listing criteria above 
- Could be useful if implemented the on hover glossary tooltip

rootqa:
> Thanks for the PR @rootqa 🙏
> 
> @samajammin @wackerow any thoughts as to adding these terms to the glossary?
> 
> My thoughts are:
> 
> #### DAG
> * We don't use the term DAG anywhere on the site.
> * Combine this with the merge happening in ~3/4 months and I don't think it makes sense to add
> 
> #### API
> * Don't mind adding API
> 
> > Is the term/definition directly relevant to Ethereum?
> 
> * Only negative for me is that it is not specific to Ethereum as per listing criteria above
> * Could be useful if implemented the on hover glossary tooltip

hey @minimalsm  thanks for interest so actually geth.ethereum.org and Developer Resources use a DAG and API and i think a lot of new developers are interested in that. thanks again!

src/content/glossary/index.md
```diff
@@ -28,6 +28,10 @@ An object containing an [address](#address), balance, [nonce](#nonce), and optio
 
 Most generally, this represents an [EOA](#eoa) or [contract](#contract-accouint) that can receive (destination address) or send (source address) [transactions](#transaction) on the blockchain. More specifically, it is the rightmost 160 bits of a [Keccak hash](#keccak-256) of an [ECDSA](#ecdsa) [public key](#public-key).
 
+### application programming interface
+
+An API is a set of definitions and protocols for building and integrating application software. API stands for application programming interface.
```
corwintines, line 33:
```suggestion
An Application Programming Interface (API) is a set of definitions for how to use a piece of software. An API sits between an application and a web server, and facilitates the transfer of data between them.
```

src/content/glossary/index.md
```diff
@@ -123,6 +127,10 @@ A crosslink provides a summary of a shard's state. It's how [shard](#shard) chai
 
 ## D {#section-d}
 
+### DAG
+
+DAG stands for Directed Acyclic Graph and without going into too many details, it is a crucial file in ETH and ETC mining. DAG file size grows over time, exactly every 30.000 blocks (Ethereum) or every 100 hours to be precise.
```
corwintines, line 132:
```suggestion
A Directed Acyclic Graph (DAG) is used to make the Ethash Proof-of-Work (PoW) ASIC resistant. The DAG dataset is a resource used when calculating the proof-of-work, and requires choosing subsets of a fixed resource dependent on the nonce and block header. A new DAG dataset is created every 30000 blocks (a 100 hour window, called an epoch), and is a few GB in size.
```
This means that calculating the PoW requires choosing subsets of a fixed resource dependent on the nonce and block header.

corwintines (commented):
I think these are relevant terms, but leaving some suggestions on how to define them.

minimalsm:
Hey @rootqa thanks again for the PR.

I chatted over this with the team and confirmed because of the lack of usage and the fact we're moving to proof-of-stake in the very near future we don't think adding DAG adds value to the glossary.

We agreed that adding API is a nice addition because it clarifies the API vs ABI confusion and makes the term more approachable for people who may not be developers.

Merging the changes now :] 

Be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And remember to [claim your POAP](https://ethereum.org/en/contributing/#poap) also 🥳!

minimalsm:
@all-contributors please add @rootqa for docs

PR #3833, opened 2021-09-06
Fix typo on gas page

skylarweaver:
My prior suggestion included a small typo I just realized 🤦


minimalsm:
Haha not enough coffee in either of us. Thanks for the fix.

PR #3844, opened 2021-09-07
Update "use-case" page redirects

minimalsm:
## Description
Removed the redirect from `ethereum.org/dao` to `ethereum.org/en/developers`. This was added to account for the missing DAO page from the original ethereum.org website (before we added the DAO learn page).

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -79,9 +79,6 @@
   from = "/crowdsale"
   to = "/en/developers/"
 [[redirects]]
-  from = "/dao"
-  to = "/en/developers/"
-[[redirects]]
   from = "/cli"
   to = "/en/developers/"
 [[redirects]]
--- a/netlify.toml
+++ b/netlify.toml
@@ -226,6 +226,23 @@
   to = "/en/developers/docs/programming-languages/dart/"
   force = true
 
+# Use-cases
+[[redirects]]
+  from = "/nft/"
+  to = "/en/nft/"
+[[redirects]]
+  from = "/nfts/"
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -78,9 +78,6 @@
 [[redirects]]
   from = "/crowdsale"
   to = "/en/developers/"
-[[redirects]]
-  from = "/dao"
-  to = "/en/developers/"
```
corwintines, line 83: Would it make sense to redirect to https://ethereum.org/en/dao, or is the point of removing this redirect to account for translated pages in other languages?

jley81 (approved):

netlify.toml
wackerow: Agreed, I'm in favor of having it redirect to the english pages for the use cases, ie:

```
https://ethereum.org/dao => https://ethereum.org/en/dao
https://ethereum.org/nft => https://ethereum.org/en/nft
https://ethereum.org/defi => https://ethereum.org/en/defi
```

Perhaps I'll open an issue to just re-visit this entire `netlify.toml` file and clean up all the redirects, especially as we're planning future content versions. In the meantime will bring this patch in.

wackerow (approved):

netlify.toml
wackerow: Actually, now that I look at this, we're removing a redirect and not replacing it with anything... I'll just add the use-cases redirects I just mentioned to this PR.

minimalsm (commented):
Looks good to me @wackerow

PR #3861, opened 2021-09-08
Fix punctuation in README

kathiresan0907:
## Description

 I really believe in solving and contributing in smallest of the issues so I saw this small error and thought of fixing it. Let this be the beginning to a long journey.


corwintines (approved):
Thanks @kathiresan0907

PR #3874, opened 2021-09-10
Fix Governance page error

minimalsm:
## Description
Fixes a mistake pointed out by David on Discord - ETH is not a token.

https://discord.com/channels/714888181740339261/870313128242602004/885979955324919848

### DIFF
--- a/src/content/governance/index.md
+++ b/src/content/governance/index.md
@@ -47,7 +47,7 @@ _Whilst at the protocol level Ethereum governance is off-chain, many use cases b
 
 There are various stakeholders in the [Ethereum community](/community/), each playing a role in the governance process. Starting from the stakeholders furthest from the protocol and zooming in, we have:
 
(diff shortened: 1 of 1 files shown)


wackerow (approved):
Hah, I still consider ETH a token, just not an ERC-20 token... more of a "native" token. But either way =)

minimalsm:
Yeah, I get your point @wackerow but in the context of actual (ERC-20) tokens this could be confusing 🤔. 

e.g. Ctrl+F 'token' on [the ETH page](https://ethereum.org/en/eth/)

I think it was @austintgriffith who said one of the misconceptions / poor mental models he saw from Enter Ethereum was people equating ETH and ERC-20 tokens.

samajammin:
Ya my 2 gwei on this - I think ETH _should_ continue to be defined as a token. Just because the token is native to the blockchain & isn't a token in the form of a smart contract like an ERC-20, I think it's still a token.

PR #3884, opened 2021-09-12
Fix typo in state channels developer doc

n1te1337:
Fixed a typo: moves -> move


minimalsm:
@n1te1337 nice eyes. Took me a few reads to get it. Thanks for the PR ☺️

PR #3888, opened 2021-09-13
Fix grammatical error in ERC-721 documentation

hugomd:
"Said that" doesn't make grammatical sense in English, it should be "That said", or "In saying that".

## Description
Fixed grammatical error.

## Related Issue
N/A


wackerow (approved):
@hugomd Thanks for the patch!

wackerow:
@all-contributors please add @hugomd for content patch

jley81 (commented):
Approved

Issue #3964, opened 2021-09-21
Docs / Gas / inconsistent rounding of base fee values


metalocal:
Some of the numbers in the table that demonstrates an increasing base fee of 12.5% are not rounded with the usual convention.

It's a minor detail, but rounding to the nearest 0.1 ETH is a subtle bit of polish for the example.

metalocal:
#3965 proposes a small change

minimalsm:
Closed by #3965

PR #3965, opened 2021-09-21
Fix base fee rounding in gas developer doc

metalocal:
Some values in the Gas page of the Docs section of the website are not consistently rounded and formatted.

## Description

The change in numbers uses conventional rounding and rounds to the nearest 0.1 ETH, consistently.

## Related Issue

#3964

### DIFF
--- a/src/content/developers/docs/gas/index.md
+++ b/src/content/developers/docs/gas/index.md
@@ -73,8 +73,8 @@ The base fee is calculated by a formula that compares the size of the previous b
 
 | Block Number | Included Gas | Fee Increase | Current Base Fee |
 | ------------ | -----------: | -----------: | ---------------: |
-| 1            |          15M |           0% |       100.0 gwei |
-| 2            |          30M |           0% |       100.0 gwei |
+| 1            |          15M |           0% |         100 gwei |
+| 2            |          30M |           0% |         100 gwei |
 | 3            |          30M |        12.5% |       112.5 gwei |
 | 4            |          30M |        12.5% |       126.6 gwei |
 | 5            |          30M |        12.5% |       142.4 gwei |
--- a/src/content/developers/docs/gas/index.md
+++ b/src/content/developers/docs/gas/index.md
@@ -77,7 +77,7 @@ The base fee is calculated by a formula that compares the size of the previous b
 | 2            |          30M |           0% |       100.0 gwei |
 | 3            |          30M |        12.5% |       112.5 gwei |
 | 4            |          30M |        12.5% |       126.6 gwei |
-| 5            |          30M |        12.5% |       142.3 gwei |
+| 5            |          30M |        12.5% |       142.4 gwei |
 | 6            |          30M |        12.5% |       160.2 gwei |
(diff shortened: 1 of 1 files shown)


minimalsm:
Hey thanks for the PR @metalocal, one thing I'm not particularly sure of is showing decimals for whole numbers (i.e. `100.0 gwei`). Is there already a widely used convention for this? Not against it necessarily just want to make sure the change is correct as well as asthethic.

cc: @wackerow any thoughts here?

metalocal:
I wouldn't say it's conventional to round whole numbers, but given the rest of the column is formatted with a decimal to the nearest tenth, then I thought it was reasonable consistency.  It's a bit of precision for clarity, eliminating the question of more rounding to nearest ETH.

If that change isn't desired, I will revert the change to the top two rows.

src/content/developers/docs/gas/index.md
```diff
@@ -73,11 +73,11 @@ The base fee is calculated by a formula that compares the size of the previous b
...
 | ------------ | -----------: | -----------: | ---------------: |
-| 1            |          15M |           0% |         100 gwei |
-| 2            |          30M |           0% |         100 gwei |
+| 1            |          15M |           0% |       100.0 gwei |
+| 2            |          30M |           0% |       100.0 gwei |
 | 3            |          30M |        12.5% |       112.5 gwei |
-| 4            |          30M |        12.5% |       126.5 gwei |
-| 5            |          30M |        12.5% |       142.4 gwei |
+| 4            |          30M |        12.5% |       126.6 gwei |
+| 5            |          30M |        12.5% |       142.3 gwei |
```
corwintines, line 80:
```suggestion
| 5            |          30M |        12.5% |       142.4 gwei |
```

corwintines (commented):
I dont mind the 100.0, but it is pretty common to lob off significant digits like that when they are 0 though. But, when denominating currency like that in this chart, might make it more a bit more digestible for users. Open to either of these myself.

src/content/developers/docs/gas/index.md
metalocal: ugh, why is rounding so hard

metalocal:
a8d227dea0067d23a181ddeedd7ca09f3a2a0d48 reverted the change to the whole value since there isn't clear consensus of opinion on the change

minimalsm:
My gripe with it is that most currencies don't have up to 18 significant digits... so saying 100.00 ETH isn't the same as $100.00 (it would be 100.000000000000000000 ETH)

minimalsm:
@all-contributors please add @metalocal for docs

metalocal:
> My gripe with it is that most currencies don't have up to 18 significant digits... so saying 100.00 ETH isn't the same as $100.00 (it would be 100.000000000000000000 ETH)

That's how I felt about all of the ETH values in the table.  I found it unclear upon the first read, and then even more confusing when the values weren't rounded conventionally.  I figured starting off the conversation with a suggestion to not round might lead to greater friction.  I would vote to not round at all - if this team is interested in that change, I can open an issue/PR for that revision as well.

PR #3986, opened 2021-09-22
Improved Prerequisites Content

himanshu3w:
Improved Prequsites Content

### DIFF
--- a/src/content/developers/docs/apis/javascript/index.md
+++ b/src/content/developers/docs/apis/javascript/index.md
@@ -13,7 +13,7 @@ If you want to use JavaScript to connect with an Ethereum node, it's possible to
 
 ## Prerequisites {#prerequisites}
 
-As well as understanding JavaScript, it might be helpful to understand the [Ethereum stack](/developers/docs/ethereum-stack/) and [Ethereum clients](/developers/docs/nodes-and-clients/).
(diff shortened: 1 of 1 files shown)


minimalsm:
Hey @himanshu3w thanks for the PR on this 😬

Open for discussion but I don't see this as an improvement. In its current form, it lists the prerequisites in order of importance and order of difficulty. Flipping it to add JavaScript to the end makes less sense to me.

himanshu3w:
Agreed The Order Of Javascript is at the last what I Mentioned There


Because The starting  1st line was not properly mentioned there.

Maybe you can correct the 1st line properly.
It will be great.
@minimalsm

minimalsm:
Sure, I can have a look. What do you think the problem is with the first line at the moment @himanshu3w?

himanshu3w:
The Beginning Starts with 
=>As well as understanding JavaScript.
and It Seems dramatically Wrong To Me.
If that part is okay, Let me know I will Close This PR.
@minimalsm

himanshu3w:
I am withdrawing this PR

minimalsm:
Yeah, I think grammatically it is correct as is. Thanks for the suggestion 😬

@all-contributors please add @himanshu3w for ideas

himanshu3w:
Thanks for adding me for ideas @minimalsm

Issue #4007, opened 2021-09-24
Add Chainlink Hackathon community event


corwintines:
Describe the solution you'd like

We should add the Chainlink Hackathon to our upcoming events section on the community page

https://chain.link/hackathon

Want to contribute?

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

PR #4008, opened 2021-09-24
Add Chainlink Hackathon community event

corwintines:
fixes: https://github.com/ethereum/ethereum-org-website/issues/4007


minimalsm:
Thanks for adding this @corwintines :)

src/data/community-events.json
```diff
@@ -62,5 +62,13 @@
...
     "startDate": "2021-10-20",
     "endDate": "2021-10-21"
+  },
+  {
+    "title": "Chainlink Hackathon",
+    "to": "https://chain.link/hackathon",
+    "sponsor": "Chainlink",
+    "description": "Join us for our biggest hackathon yet, where you’ll get the chance to build the next generation of hybrid smart contracts alongside developers, creators, artists, and industry-leading mentors from across the blockchain ecosystem.",
+    "startDate": "2021-10-22",
+    "endDate": "2021-11-28"
```
minimalsm, line 72: IMO we don't need to include Chainlink as a sponsor here. And think we should chop their description a little so that it's shorter and doesn't use the first person. Feel free to ignore and merge if you disagree 😬👍

```suggestion
  {
    "title": "Chainlink Hackathon",
    "to": "https://chain.link/hackathon",
    "sponsor": null,
    "description": "Build the next generation of hybrid smart contracts alongside developers, creators, artists, and industry-leading mentors from across the blockchain ecosystem.",
    "startDate": "2021-10-22",
    "endDate": "2021-11-28"
```

corwintines: @minimalsm I think its still worth putting Chainlink as a sponsor for the event, but I can see why it might be redundant as well. Do you know if we can list multiple sponsors? 

As for the description part, im pretty indifferent. I copied their description for this, and I dont know how I feel about changing their description, but im not against it for this either. Would be curious what others thing before pulling this in. 

cc: @samajammin @wackerow @pettinarip 

@PatrickAlphaC maybe you might have an opinion on this as well.

minimalsm: On the sponsor I'm pretty indifferent, just thought it was redundant as you said.

On the description, I think 'Join us' implies it might be an EF event which I think could be misleading.

corwintines: True. I suppose I look at community events as if that post is from the event itself and not the EF, but can understand how it could be taken that way. Open to whatever you think is best here, if you want to just merge it with your request here I am happy with that.

PatrickAlphaC: Wow this is awesome! Glad it's getting added here too!

The description looks pretty good to me!

minimalsm: @PatrickAlphaC thanks for weighing in.

@corwintines I'm indifferent about removing the sponsor info so I've readded that and went with the modified description as Patrick has approved 😬👍

PR #4009, opened 2021-09-24
Add explicit https to twitter link

himanshu3w:
Fixed Link IvanBogatyy Twitter Status for 7$million


minimalsm:
Hey :-) thanks for this. Do you know if this makes much difference? Both seem to direct to the same page and both use HTTPS? (I suppose it's better to always declare HTTPS explicitly)

Merging 👍

himanshu3w:
Actually The Previous Link was a relative link and so was redirecting to 404 error page that why I did That.

Thanks For Merging

Issue #4027, opened 2021-09-25
Keyboard shortcut to jump focus to search input


wackerow:
**Is your feature request related to a problem? Please describe.**
Functionality to use the `/` key to jump focus to the search bar on desktop.

**Describe the solution you'd like**
Add a listener for keystroke `/` (or even additionally `ctrl/cmd+K` as used in Discord or Telegram) to automatically move cursor focus to the search bar. An indicator on the search bar with the `/` character to inform user of option. 
![image](https://user-images.githubusercontent.com/54227730/134756972-73095347-463e-42fd-be57-47f0d169c82b.png)

**Describe alternatives you've considered**
Keep requiring the mouse or tabbing through with keyboard 👎

**Additional context**
A few examples of sites that utilize this behavior are https://github.com, https://coingecko.com or https://coinmarketcap.com. Was thinking of keeping this simple and simply place the cursor in this input, but not necessarily building out a new dropdown like used on these sites, at lease for this issue. 

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

SNikhill:
@wackerow 

Sir, May I please tackle this issue?

SNikhill:
Linked PR: #4031

corwintines:
Closing as we have merged the PR

PR #4031, opened 2021-09-25
Shortcut to Focus Search #4027

SNikhill:
## Description
This merge will create a keyboard shortcut: `/` to focus the search input and change the search icon with a slash icon akin to that on github.com.
image

## Related Issue
- #4027


SNikhill:
So, I was fixing another issue and exploring the code and found the `useKeyPress` hook. I have now modified the code to use the same.

minimalsm:
Hey @SNikhill thanks for the PR and making a start on this 🙌

@wackerow was removing the current Search icon what you had in mind here?

Feedback is below but I wouldn't act on it until we clarify this ☝️

--- 
I think this requires some further design thinking but I think the sizing here looks too large/clunky 

Screenshot 2021-09-27 at 11 04 36

I reduced the size slightly and I definitely think it looks better smaller.

Screenshot 2021-09-27 at 11 04 41

From a UX perspective, I think we will want to reduce the opacity of icon slightly as visually the text 'Search' and the icon '/' look like they hold the same weight to me.

---

I also don't think we should remove the current Search icon on mobile as it is semantically tied to mobile navigation and doesn't make sense to ask users on mobile to press a shortcut to search. 

Screenshot 2021-09-27 at 11 00 32

Screenshot 2021-09-27 at 11 00 07

SNikhill:
@minimalsm 

My Apologies. The part about the Phone Icon skipped my mind. I understand that a `slash` icon will be irrelevant for a mobile screen and I shall make the changes appropriately.

wackerow:
@SNikhill Thanks again for jumping on this! 

My apologies, no I did not intent to suggest removing the search icon 🔎... Although looking at this, I may suggest if we're going to add the `/` indicator that we consider relocating the 🔎 to the left of the Search text, maybe like this:

image

I would consider not using an icon for this and just using `/` string to keep it simpler, and it should look more natural than the 45 degree slash shown above. And yes, to confirm, this should not change the mobile experience.

minimalsm:
I would be ok with adding the functionality and not having the slash icon at all. I still think this is a win.

wackerow:
Yes, agreed... the slash is honestly icing on the cake, but the core functionality is more important.

minimalsm:
@SNikhill just pinging to make sure you've seen this 😬

SNikhill:
@minimalsm 

My Apologies, Sir. I have been engaged in something else. I shall ensure that I create the requested changes as soon as possible.

SNikhill:
> Yes, agreed... the slash is honestly icing on the cake, but the core functionality is more important.

Well, a desktop user will never know if the slash can be used to search (unless we cater to `vim` users) without that slash icon.

SNikhill:
@minimalsm and @wackerow 

As suggested, there now exists a slash icon with a search icon (on the left side) while maintaining the mobile functionality intact.
image

image
image
image

src/components/Search/Input.js
```diff
@@ -45,20 +75,40 @@ const Input = ({ query, setQuery, refine, ...rest }) => {
...
+    if (document.activeElement !== searchInput) {
+      event.preventDefault()
+      searchInput.focus()
+    }
+  }
+  useKeyPress("/", focusSearch)
+
+  useEffect(() => {
+    setShowSearchSlash(window.innerWidth >= 768)
+  }, [])
```
wackerow, line 89: This should break at our themed breakpoint of "1024px" (when mobile view starts)... ideally using the theme context, ie:

First import ThemeContext: `import { ThemeContext } from "styled-components"`
Then something like this:
```
  const themeContext = useContext(ThemeContext)
  const breakpoint = themeContext.breakpoints.l
```

wackerow (changes requested):
Coming along, but couple bugs here..
image
This is currently (and should not be) loading initially, before showing the magnifying glass on the left, and the slash on the right.

Also, again would request we just use a `/` character instead of an icon here. The icon presents an awkward 45-deg line which is not how a normal slash character looks, and adds unnecessary logic to load and style that icon. Would just create a div with a border and slight border-radius, containing a slash. 

See note about using our themed breakpoint instead of hard-coding pixel values. Also, looks like we have two different "SearchIcons" now (`SearchIcon` and `LeftSearchIcon`)... is there a reason for this? I think we should only have one and it should be on the left in this case (would just keep name `SearchIcon`).

PR #4034, opened 2021-09-26
Fix repo URL for Nimbus on ETH2 page

the-emerald:
URL used to point to Eth1 execution client...

## Description

The URL for the Nimbus eth2 client currently points to their eth1 client repo.

## Related Issue

N/A


minimalsm:
Ooops 🤪 nice catch @the-emerald and thanks for the PR 👍

PR #4042, opened 2021-09-27
Fixed Wiki Links Yellow Paper Appendix, Yellow Papaer and YP

himanshu3w:
Fixed Wiki Links Yellow Paper Appendix, Yellow Papaer and YP


wackerow (approved):
👍🏻 Thanks @himanshu3w !

PR #4066, opened 2021-09-28
Updated DAO Direct Redirect Https Link

himanshu3w:
Have Updated The Direct Link For DAO. It will result in faster redirect.
Although this is not necessary.
@minimalsm @wackerow  .
Just Ping Me If this don't seems necessary. I will close The PR


wackerow (approved):
@himanshu3w 👍🏻 Thanks!

PR #4070, opened 2021-09-28
Update POAP section Discord link to go to #poaps channel

himanshu3w:
Updated Discord Link Explicitly For #poap channel.
I think It is good Idea To redirect the user to #poap channel rather than #ethereum-org-development channel of discord Server Because We are Talking About POAP
Note : The Link Attached is of the following type [The Link Will Never Expire, There is No Limit To the Number Of Users]

### DIFF
--- a/src/content/contributing/index.md
+++ b/src/content/contributing/index.md
@@ -76,7 +76,7 @@ If your contribution gets merged into ethereum.org, we'll mint you a unique cont
 
 ### How to claim {#how-to-claim}
 
-1. Join our [Discord server](https://discord.gg/CetY6Y4).
+1. Join our [Discord server](https://discord.gg/Gnz2jWHw6t).
 2. Paste a link to your contribution in the #poaps-🏆 channel.
 3. Wait for a member of our team to send you a link to your minted POAP.
 4. Claim your POAP!


wackerow (commented):
Hey @himanshu3w, thanks for the suggestion. So trying these out, I get funneled into the "Rules" channel for both... I believe this is intentional, correct @minimalsm? I don't think our intentional was to directly lead people into any channel before reading the rules page first.

himanshu3w:
@wackerow Thanks Annotating On This.
I am Closing This PR.
@minimalsm

minimalsm:
@wackerow nah I hit the ethereum-org-development channel also. @himanshu3w I'm good with sending them directly to #poaps for better UX. Thanks for the suggestion

minimalsm:
> Hey @himanshu3w, thanks for the suggestion. So trying these out, I get funneled into the "Rules" channel for both... I believe this is intentional, correct @minimalsm? I don't think our intentional was to directly lead people into any channel before reading the rules page first.

To clarify: any new user will get hit with the rules pop up and can only continue/speak after they accept the rules. the rules channel itself now simply servers as a rules reminder. 

This change will have the user hit the rules pop-up, then be directed to the #poaps channel.

image

himanshu3w:
Nice Catch @minimalsm . You Reopened and Updated 👍 .
Thanks.

PR #4079, opened 2021-09-29
Removed Kauri From The List

himanshu3w:
Removed Kauri From The List of Community Built Resources.
All the Materials Provided By Them is almost already available in other list items other than Kauri.
Reason :  Last Updated Dec 2020, URL : https://github.com/kauri-io/archive
Community Report : Community is No Longer managing The WebSite. They have just hosted it.
@samajammin @minimalsm Take a look


wackerow (approved):
Not necessarily against this given they do appear to be inactive. @minimalsm We should also look into the other links here to confirm if they're still viable resources for the community.

Issue #3908, opened 2021-09-15
Audit Developer Documentation


minimalsm:
This issue is to be used for the management of an audit of the [developer documentation](https://ethereum.org/en/developers/docs/) on ethereum.org. 

If you have any ideas or want to discuss things to add to remove from the existing content listed here please leave a comment below or [join the ethereum.org discord](https://discord.gg/CetY6Y4). 

---

**Foundational Topics**

- [x] README [Merged PR](https://github.com/ethereum/ethereum-org-website/pull/4095)
- [x] Intro to Ethereum - [Merged PR](https://github.com/ethereum/ethereum-org-website/pull/3907)
- [x] Intro to Ether - [Merged PR](https://github.com/ethereum/ethereum-org-website/pull/3929)
- [x] Intro to dapps - [Merged PR](https://github.com/ethereum/ethereum-org-website/pull/4065)
- [ ] Web2 vs Web3
- [ ] Accounts
- [ ] Transactions
- [ ] Blocks
- [ ] Ethereum virtual machine (EVM)
- [ ] Gas
- [ ] Nodes and clients
- [ ] Run a node
- [ ] Nodes as a service - [Merged PR](https://github.com/ethereum/ethereum-org-website/pull/4050) (awaiting more changes)
- [ ] Networks
- [x] Consensus mechanisms - [Merged PR](https://github.com/ethereum/ethereum-org-website/pull/3978)
- [x] Proof of work - [Open PR in progress](https://github.com/ethereum/ethereum-org-website/pull/3999)
- [x] Mining - [Merged PR](https://github.com/ethereum/ethereum-org-website/pull/4112)
- [ ] Proof of stake

**Ethereum Stack**- 

- [x] Intro to the stack - [Merged PR](https://github.com/ethereum/ethereum-org-website/pull/4116)
- [x] Deploying smart contracts - [Merged PR](https://github.com/ethereum/ethereum-org-website/pull/4064)
- [x] Storage - [Merged PR](https://github.com/ethereum/ethereum-org-website/pull/4081)
- [x] Python - [Merged PR #1](https://github.com/ethereum/ethereum-org-website/pull/4000), [Merged PR #2](https://github.com/ethereum/ethereum-org-website/pull/4015)

**Advanced**
- [ ] Scaling
- [x] Layer 2 rollups - [Merged PR](https://github.com/ethereum/ethereum-org-website/pull/3996)
[...]

PR #4081, opened 2021-09-29
Update Storage developer docs

minimalsm:
## Description

This PR is part of a series to [audit the developer docs](https://github.com/ethereum/ethereum-org-website/issues/3908) in preparation for upgrades to our translation program.

## Link
Deploy preview - https://deploy-preview-4081--ethereumorg.netlify.app/en/developers/docs/storage/


src/content/developers/docs/storage/index.md
```diff
@@ -1,18 +1,15 @@
...
 
-As opposed to a centrally located server operated by a single company or organization, decentralized storage systems consist of a peer-to-peer network of user-operators who hold a portion of the overall data, creating a resilient system of file storage and sharing. These can be in a blockchain-based application, or any peer-to-peer based network.
+Ethereum itself can be used as a decentralized storage system, and it is when it comes to code storage in all the smart contracts. However, when it comes to large amounts of data, that isn't what Ethereum was designed for. At the time of writing, the Ethereum chain is around 350GB, and every node on the network needs to be able to store all 350GB of data. If the chain were to expand to large amounts of data (say 5TBs) it wouldn't be feasible for all nodes to continue to run. Also, this can get incredibly expensive to deploy new data due to the size.
```
minimalsm, line 10: > At the time of writing, the Ethereum chain is around 350GB, and every node on the network needs to be able to store all 350GB of data.

Where did we get this figure? I feel like we should probably use Geth full nodes as the benchmark - https://etherscan.io/chartsync/chaindefault

wackerow: Yeah this is an outdated figure. It also keeps expanding, and is dependent on client implementation... maybe we put in a range here of ~500GB to ~1TB at time of writing, perhaps linking to this chart. 

```suggestion
Ethereum itself can be used as a decentralized storage system, and it is when it comes to code storage in all the smart contracts. However, when it comes to large amounts of data, that isn't what Ethereum was designed for. The chain is steadily growing, but at the time of writing, the Ethereum chain is around 500GB - 1TB ([depending on the client](https://etherscan.io/chartsync/chaindefault)), and every node on the network needs to be able to store all of the data. If the chain were to expand to large amounts of data (say 5TBs) it wouldn't be feasible for all nodes to continue to run. Also, the cost of deploying this much data to Mainnet would be prohibitively expensive do to [gas](/developers/docs/gas) fees.
```

src/content/developers/docs/storage/index.md
```diff
@@ -53,7 +50,7 @@ Platforms with contract based persistence:
 
 ### Additional considerations {#additional-considertion}
 
-As a bonus, IPFS doesn't really have an incentive structure for holding data, but it is a great tool in the community. If you'd like data to persist on IPFS, you have to reach out to a pinning service, which will "pin" your data for you.
+As a bonus, IPFS doesn't really have an incentive structure for holding data, but it is an excellent tool in the community. If you'd like data to persist on IPFS, you have to reach out to a pinning service, which will "pin" your data for you.
```
wackerow, line 53:
```suggestion
As a bonus, IPFS doesn't really have an incentive structure for holding data, but it is an excellent tool in the community. If you'd like data to persist on IPFS, you have to reach out to a pinning service, which will "pin" your data for you. Or you can run your own node and contribute to the network!
```

src/content/developers/docs/storage/index.md
```diff
@@ -74,7 +71,7 @@ Types of dStorage with a challenge mechanism:
 
 ### Decentrality {#decentrality}
 
-There aren't great tools to measure the level of decentralization of platforms, but in general you'll want to use tools that don't have some form of KYC to prove they are actually decentralized.
+There aren't great tools to measure the level of decentralization of platforms, but in general, you'll want to use tools that don't have some form of KYC to prove they are genuinely decentralized.
```
wackerow, line 74: Hesitate on this one, since lack of KYC certainly doesn't really prove anything. My thought would be to reword that lack of KYC can provide evidence that a project is not centralized, while staying away from "prove."

```suggestion
There aren't great tools to measure the level of decentralization of platforms, but in general, you'll want to use tools that don't have some form of KYC to provide evidence they are not centralized.
```

wackerow (changes requested):
@minimalsm Looking great! Left a few suggestions, feel free to ignore, though we should at least update the chain size. 👍🏻

src/content/developers/docs/storage/index.md
minimalsm: Yeah on second read you're 100% right this is a bit misleading.

src/content/developers/docs/storage/index.md
minimalsm:
```suggestion
As a bonus, IPFS doesn't really have an incentive structure for holding data, but it is an excellent tool in the community. If you'd like data to persist on IPFS, you have to reach out to a pinning service, which will "pin" your data for you. You can even run your own node and contribute to the network!
```

PR #4087, opened 2021-09-29
Minor grammatical correction.

SlashHash:
"Merkle tree" was misspelled as "Merkle trie".

### DIFF
--- a/src/content/developers/docs/data-and-analytics/block-explorers/index.md
+++ b/src/content/developers/docs/data-and-analytics/block-explorers/index.md
@@ -49,7 +49,7 @@ New blocks are added to Ethereum every ~12 seconds (this can fluctuate) so there
 - Hash – The cryptographic hash that represents the block header (the unique identifier of the block).
 - Parent hash – The hash of the block that came before the current block.
 - Sha3Uncles – The combined hash of all uncles for a given parent.
-- StateRoot – The root hash of Merkle trie which stores the entire state of the system.
+- StateRoot – The root hash of Merkle tree which stores the entire state of the system.
 - Nonce – A value used to demonstrate proof-of-work for a block by the miner.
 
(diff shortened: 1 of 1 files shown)


wackerow:
Hey @SlashHash! Welcome and thanks for the post! The use of "trie" here had been intentional, and was discussed way back on issue #2397. ([More on tree vs trie](https://stackoverflow.com/questions/4737904/difference-between-tries-and-trees)).

We decide to go with the slightly more specific term here, and has been changed in some places, but I am noticing that the Whitepaper does use the term "tree" instead. @minimalsm Curious your thoughts here... personally favor trie, though not a hill I'll die on.

himanshu3w:
Trie is a part of Tree data structure.
In Ethereum We Use trie.

But Since Trie is a part of Tree so often interchangeably used.
When We we are talking in context of ethereum we must use trie instead of tree.
[What Wikipedia Says](https://en.wikipedia.org/wiki/Tree_(data_structure))

minimalsm:
Thanks for the PR @SlashHash 😀

I agree with @wackerow that `trie` is the correct term. Although it's only slightly more accurate to use `trie` > `tree` if we implement this change we'll end up with another PR in a few weeks/months to revert it (and so on...)

Gonna close this out but appreciate you flagging this 👍

minimalsm:
@all-contributors please add @SlashHash for ideas

SlashHash:
Hello everyone, this is my first pull request ever and I don't know how it usually works but I'll just say that I didn't expect such a quick response so please excuse my delayed response and thank you all for the kindness!
   I will admit that it was a mistake on my part and my lack of my knowledge in that specific area. Sorry for wasting the time of everyone involved.

himanshu3w:
We appreciate you shown your interest in the ethereum community. @SlashHash

PR #4108, opened 2021-10-01
Patch to update Lodestar status

minimalsm:
## Description

Changes Lodestar back to `isProductionReady: false`


wackerow (approved):
@minimalsm Thanks buds

PR #4120, opened 2021-10-02
Maximal Extractable Value as title and description

eason-w:
Change the current title and description's "Miner Extractable Value" to "Maximal Extractable Value"

### DIFF
--- a/src/content/developers/docs/mev/index.md
+++ b/src/content/developers/docs/mev/index.md
@@ -1,6 +1,6 @@
 ---
-title: Miner extractable value (MEV)
-description: An introduction to miner extractable value (MEV)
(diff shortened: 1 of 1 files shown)


samajammin:
Thanks for the PR @eason-w!

I'm not convinced we should change the title, at least not yet. Reason being, I think the term is still more widely known as "miner extractable value". e.g. if you look at global Google search volume:
Image 2021-10-04 at 10 35 45 AM
Not a ton of search volume for either but clearly more for "miner extractable value".

Not a super strong opinion here so feel free to re-open for this PR if you'd like to make a case for this.

eason-w:
Ah ok that makes total sense! Thanks for the reply :)

Issue #2737, opened 2021-03-27
Docs Focused Reading (a.k.a. Zen Mode) Toggle


Hakeemmidan:
## **Is your feature request related to a problem? Please describe.**

Documentation part of website currently feels like it has a lot of distracting details in it.

How it looks like now:
> ![eth_docs_now](https://user-images.githubusercontent.com/27740273/112723314-d75e3b80-8f1e-11eb-96e2-4fc38f5c07f9.png)

All these details might distract readers from having a potentially more focused reading time.

## **Describe the solution you'd like**
Adding 'Zen Mode' or 'Focused Reading' _toggle button_ might be helpful in reducing distraction.

It might look something like the following:
> ![eth_docs_zen](https://user-images.githubusercontent.com/27740273/112722758-5f8f1180-8f1c-11eb-92c0-9a9a5d0c0b66.png)

## **Describe alternatives you've considered**

Keeping right-side menu for better usability:
> <img width="1344" alt="eth_docs_side_mnu" src="https://user-images.githubusercontent.com/27740273/112722918-1ab7aa80-8f1d-11eb-8019-89eac8614012.png">

## **Additional context**

One concern that was brought up on Discord is that removing these details might reduce accessibility and SEO ranking. To that, [this](https://css-tricks.com/snippets/css/accessibilityseo-friendly-css-hiding/) was proposed. 

What the proposed solution entails is basically moving content off the page, but not deleting them, which keeps them visible to website crawlers and other similar tools.

minimalsm:
Nice proposal. It doesn't seem intuitive to me where the focused toggle button would be placed on the page (in either state). Any thoughts?

Hakeemmidan:
> Nice proposal. It doesn't seem intuitive to me where the focused toggle button would be placed on the page (in either state). Any thoughts?

I was thinking about adding it to the left of the light/dark mode toggle, but I saw some problems with that, including: 

- It would depend on the docs relative path (since it only appears on the docs page). *Changing the docs path name might easily break things that way.*
- Clicking on it supposedly should make the entire `Nav` element to disappear, except for the 'Zen Mode' toggle. *This might add unnecessary complexity.*

So instead, to make it more independent, I think we should place it on the top-right of the doc layout. It would be in place of the red rectangle:
[...]

Issue #4026, opened 2021-09-25
XL Layout for docs


wackerow:
Having looked at this more closely, I'd recommend a few tweaks to solve the XL-width issue–whilst keeping a good reading (and aesthetic) experience for the user.

First screen: 1920x1080
![1920x1080](https://user-images.githubusercontent.com/83584447/132910444-b669b160-1fa3-42de-b346-554c426dfeff.png)

After 1504px, I recommend moving away from the main content being slightly left-aligned. This allows for a more conventional 'reading focused mode' to begin–whereby the content will have an equal margin on either side of it. 

From my experiments in Figma, the width of the Contents should remain at 640px until it hits 1920 width. At this point, I'd suggest increasing the width of the content to 700px to make the content look more appropriate for larger screen sizes–without damaging the reading experience.

Second screen: 2560x1440
![2560x1440](https://user-images.githubusercontent.com/83584447/132911160-a1755683-567a-4f26-be82-a4d5fd5e93b5.png)

Expanding the entire view to fit XL screens is difficult to get right; the navigation often ends up too far away from the main content and ultimately, one thing will always suffer.

Hope this is of some use. Let me know your thoughts/comments ✌️

_Originally posted by @Scott1UP in https://github.com/ethereum/ethereum-org-website/pull/2671#issuecomment-917183413_

---
Resurfacing this feature request as a new issue and opening up to anyone who would like to contribute, using the above design concept as a guide. (Original issue #1780)

howdyAnkit:
Hi, @minimalsm @wackerow  can I get assigned for this?

minimalsm:
Hey @howdyAnkit thanks for offering to help on this! I've assigned you. Any questions feel free to ping here or on Discord. 👍

Scott1UP:
Hey @wackerow 👋🏼  Thanks for the questions and hopefully my answers can bring some clarity. I'll preface this response with the disclaimer that I am not a web developer by day, so my knowledge of the correct terminology can be a bit flaky at times... 

1. Yes. Based on the tests I ran at 1920x1080 and 2560x1440, it would make sense to increase the width of the docs pages. In my designs, I found a maximum width of 1920px was sufficient even at the 2560x1440 resolution–however, I'll defer to your knowledge on the best way of implementing that same feel from a technical perspective. As long as we are hitting the goals of this issue and improving the experience for the user, I'm happy. 

[...]

PR #4131, opened 2021-10-04
Feature: Zen Mode Toggle for Documentation Pages

Hakeemmidan:
## Description
🧘
Added Zen Mode toggle to documentation (`/docs`) pages:

Screen Shot 2021-10-04 at 5 55 48 PM

### Features
- Persistant
  - Like the theme toggle, the value of `'zen-mode'` gets saved in [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage)
    - That is so when a user closes their ethereum.org tab, or duplicates it, they can keep their settings and get back into flow when they return
- Mobile friendly
  - The toggle doesn't show as an option, and is automatically turned off, when visited on mobile width screens ( < 1024px) since the mobile view of the docs is already pretty minimal
  - If a user shrinks a desktop-sized window to mobile-size with Zen Mode enabled, a more minimal UI may appear but nothing glitchy. Check out the screenshots below:
     - Shrunken window with zen-mode disabled: 
       - Screen Shot 2021-10-04 at 5 42 55 PM
    - Shrunken window with zen-mode enabled:
      - Screen Shot 2021-10-04 at 5 43 10 PM
- Modular
  - The Zen Mode toggle and its effects only appear in `/docs` pages
    - It shouldn't affect any other pages

### Demo (GIF)
eth-org_zen-mode-demo

### Lighthouse
It doesn't affect the SEO or accessibility for the website, but it does slightly lower performance when it (Zen Mode) is enabled (as expected).

#### `dev` branch score
Screen Shot 2021-10-04 at 12 11 07 PM

#### `zen-mode` branch score (with Zen Mode disabled)
Screen Shot 2021-10-04 at 12 02 46 PM

#### `zen-mode` branch score (with Zen Mode enabled)
Screen Shot 2021-10-04 at 12 06 09 PM

## Related Issue
#2737


wackerow (changes requested):
@Hakeemmidan Checking this out and it looks kinda cool!

I do hesitate with the concept of removing the navigation bars / docs navigation with the _only_ way to get them back by toggling that switch back over. 

Also a little torn that we remove the navbar, but not the footer.. leaving the footer seems to leave it only half-zen.

Perhaps we should enable the `esc` key to exit back to non-zen mode as well, and/or provide some other type of indicator at the top/bottom of page to make it more obvious how to get back to the default view.

And on that note, should this mode persist on page refresh? Right now it does, but I feel like if someone hard-refreshes the page it should load to the default view. I do agree with it persisting when navigating between docs using the navigation buttons at the bottom.

Lastely, for intl purposes, anyone have thoughts on just using an Icon to replace the text "Zen Mode", that way it doesn't need to be translated at all, and will eliminate any issues with wider text in other languages?

Hakeemmidan:
> @Hakeemmidan Checking this out and it looks kinda cool!

Thank you! = )

> Also a little torn that we remove the navbar, but not the footer.. leaving the footer seems to leave it only half-zen.

I agree. It now hides the footer as well when it's enabled.

> Perhaps we should enable the `esc` key to exit back to non-zen mode as well, and/or provide some other type of indicator at the top/bottom of page to make it more obvious how to get back to the default view.

I agree about the `esc` key part. It seems like it'd make the user experience more intuitive/friendly. Clicking on `esc` now exits Zen Mode. (I did it using the existing `useKeyPress` hook.)

I'm not sure how an indicator would look like, but I think it's best to have as little distractions on the screen when reading digitally. (Think of reading in an ad-filled news website vs. reading on something like Medium.) 

So personally, I think having the toggle and the `esc` key behavior should be enough. But I could be wrong of course.

> And on that note, should this mode persist on page refresh? Right now it does, but I feel like if someone hard-refreshes the page it should load to the default view. I do agree with it persisting when navigating between docs using the navigation buttons at the bottom.

Reading the docs will most likely be a thing that's done over multiple days for someone that's interested in checking-out/reading many parts of it. So it'd just be more convenient (for those people) to have settings such as the theme and Zen Mode persist on page-refresh.

Nevertheless, I do get the appeal of the other option. It seems (from observation) that page-refresh serves as a reset-function for many people. However, I think that having the button toggle, and now the `esc` key should be easy enough to toggle back and forth between Zen Mode and the default view. So a reset on page-refresh is no longer necessary (in my opinion).

Hakeemmidan:
@wackerow 
> @Hakeemmidan Checking this out and it looks kinda cool!

Thank you! = )

> Also a little torn that we remove the navbar, but not the footer.. leaving the footer seems to leave it only half-zen.

I agree. It now hides the footer as well when it's enabled.

> Perhaps we should enable the `esc` key to exit back to non-zen mode as well, and/or provide some other type of indicator at the top/bottom of page to make it more obvious how to get back to the default view.

I agree about the `esc` key part. It seems like it'd make the user experience more intuitive/friendly. Clicking on `esc` now exits Zen Mode. (I did it using the existing `useKeyPress` hook.)

I'm not sure how an indicator would look like, but I think it's best to have as little distractions on the screen when reading digitally. (Think of reading in an ad-filled news website vs. reading on something like Medium.) 

So personally, I think having the toggle and the `esc` key behavior should be enough. But I could be wrong of course.

> And on that note, should this mode persist on page refresh? Right now it does, but I feel like if someone hard-refreshes the page it should load to the default view. I do agree with it persisting when navigating between docs using the navigation buttons at the bottom.

Reading the docs will most likely be a thing that's done over multiple days for someone that's interested in checking-out/reading many parts of it. So it'd just be more convenient (for those people) to have settings such as the theme and Zen Mode persist on page-refresh.

Nevertheless, I do get the appeal of the other option. It seems (from observation) that page-refresh serves as a reset-function for many people. However, I think that having the button toggle, and now the `esc` key should be easy enough to toggle back and forth between Zen Mode and the default view. So a reset on page-refresh is no longer necessary (in my opinion).

minimalsm:
Hey 👋! Thanks for the work you've done on this 😬.

Agree with Paul that if we implement this it **should not** be consistent across a hard refresh.

**Scenario:**

- A user is reading documentation
- Clicks 'zen mode' toggle
- User closes browser
- Days later: user googles ethereum and lands on a docs page - "hmm this page seems broken"
- User bounces without getting the information they were searching for

Overall, I'm not very keen on this feature. I understand the motivation (improving user focus) but my worry is that we're reducing aesthetic usability for the many to aid few who would actually use this feature. 

According to Jakob's Law: *Users spend most of their time on other sites. This means that users prefer your site to work the same way as all the other sites they already know.*

Most docs don't have a zen/focus mode. According to research by the [Neilsen Norman Group](https://www.nngroup.com/articles/zen-mode/) this might be because the intuition that user productivity will improve by having a zen mode feature for a cleaner UI is incorrect due to the cognitive overhead of learning the feature itself (the stickiness just isn't there).

If we want to implement a feature to make the UI less distracting a togglable sidebar feature (similar to what Notion has) might be a good option here? 

Screenshot 2021-10-06 at 16 47 54

Two relative benefits: 2. a togglable sidebar has lower cognitive overhead because it is more widely used and 2. it doesn't add visual clutter to the page if it is not used (arrows only appear on hover of the sidebar itself)

Another thing that might help (at least on larger screens) is shipping the [XL layout for docs](https://github.com/ethereum/ethereum-org-website/issues/4026)

Check out the research article [Why Zen Mode Isn't the Answer to Everything](https://www.nngroup.com/articles/zen-mode/) for most of the research causing my concerns.

Hakeemmidan:
@corwintines 

> First of all, thank you @Hakeemmidan for the contribution of this feature.

🙏 =)

> I actually dont mind that the refresh persists zen mode. .... I think the functionality for now is good as is, and we can probably iterate on this after getting some feedback. 

Yes, I agree. I think it might be best to get user feedback on things such as the persistence of Zen Mode (if applicable), given that it seems somewhat debatable. [I feel like both ways could work. Not sure which one users would prefer.]

src/components/Layout.js
```diff
@@ -78,9 +78,12 @@ const Layout = (props) => {
       setShouldShowSideNav(true)
 
       if (localStorage.getItem("zen-mode") !== null) {
-        const isMobile = /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
-          navigator.userAgent
-        )
+        let isMobile = false
+        if (typeof window !== undefined) {
```
Hakeemmidan, line 82: > Did a little cleanup, but looks good @Hakeemmidan! Thanks!

@corwintines I think this needs to be 
```JavaScript
if (typeof window !== 'undefined') {
```
Because [the docs say](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof): "The **typeof operator returns a string** indicating the type of the unevaluated operand."

Other than that, the new changes look good! Thank you!

PR #4205, opened 2021-10-12
implemented hovering tooltip, fixed positioning [Fixes #3139]

### LINKED ISSUE #3139: Add hover effect to the Tooltip

erhuve:
## Description
As per this comment: https://github.com/ethereum/ethereum-org-website/pull/3140#issuecomment-856165135 I've attempted my implementation of the original intention of the issue. In the process, I've also repositioned the tooltip for a cleaner layout (which was necessary for my implementation too :p).

For some reason, handling a mouse leave event on Container resulted in turning off visibility immediately. I created a new div similar to ModalReturn to appear on top of the container and handle the mouse off event. I positioned the Content to allow for a "bridge" for the cursor to the content to remain hovering. Click functionality still works.

I'm not too experienced with frontend, so my implementation may be unnecessarily messy and I'm not sure if my CSS is fully responsive. I tested it in my browser, resized window, and the mobile viewport in chrome devtools, which all looked fine at least. Any feedback appreciated. Thanks :)

## Related Issue
#3139 #3140

## Link
https://deploy-preview-4205--ethereumorg.netlify.app/en/


corwintines:
> Thanks @corwintines. I do agree the isMobile approach is a lot more elegant. I do have a couple concerns though:
> 
> 1. Will  checking `clientWidth < 1024` be robust to large touch devices like an iPad?
> 2. With this implementation and removing the style changes, I believe if you attempt to click the link in the content box, the content will instead disappear because your mouse has left the container.
> 3. I forgot to mention this, but I think the reason the cursor is immediately registered as having left the container after entering is because of the rendering of ModalReturn. I could try changing its z-index to see if it fixes this, but otherwise that is why I saw the need for ModalHoverOff; we can see if it is still necessary.

Hmm good points. I think theres a better query I can come up with for your first point using `navigator.userAgent`. Will do some playing around here and report back.

corwintines:
> Thanks @corwintines. I do agree the isMobile approach is a lot more elegant. I do have a couple concerns though:
> 
> 1. Will  checking `clientWidth < 1024` be robust to large touch devices like an iPad?
> 2. With this implementation and removing the style changes, I believe if you attempt to click the link in the content box, the content will instead disappear because your mouse has left the container.
> 3. I forgot to mention this, but I think the reason the cursor is immediately registered as having left the container after entering is because of the rendering of ModalReturn. I could try changing its z-index to see if it fixes this, but otherwise that is why I saw the need for ModalHoverOff; we can see if it is still necessary.

@erhuve alright I think this should be a better implementation for the mobile check. 
```
const Tooltip = ({ content, children }) => {
  const isMobile = /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(navigator.userAgent)
  const [isVisible, setIsVisible] = useState(false)

  return (
    <>
      {(isVisible && isMobile) && <ModalReturn onClick={() => setIsVisible(false)} />}
      <Container
        title="More info"
        onMouseEnter={!isMobile ? () => setIsVisible(true) : null}
        onMouseLeave={!isMobile ? () => setIsVisible(false) : null}
        onClick={isMobile ? () => setIsVisible(!isVisible) : null}
      >
        {children}
        {isVisible && (
          <Content>
            <Arrow />
            {content}
          </Content>
        )}
      </Container>
    </>
  )
}
```

Let me know what you think. 
If you try implementing just this and removing the style changes I think this has the functionality we are looking for here. Ill let you do some testing and see what you think.

corwintines (approved):

corwintines:
@all-contributors can you add @erhuve for code

corwintines:
Thanks @erhuve, make sure you come to collect your POAP in our discord!

PR #4216, opened 2021-10-13
Add in a new 'Block time' portion

Xia-Sam:
## Description

I have added in a new portion called block time in the blocks page. It's my first time contributing. If there's anything i can still improve on, please let me know. Thanks


minimalsm:
*This was marked as resolved in the other thread so posting again.*

> In ethereum, it is between 10 to 19 seconds

We can definitely be more specific than this considering blockchain has been consistently between 12-14 seconds for almost 2 years.

https://ycharts.com/indicators/ethereum_average_block_time

PR #4231, opened 2021-10-14
Improve grammar in introduction to ERC-20

irakojf:
Edit grammar in body to be more readable and easily understood.


irakojf:
> Hey @irakojf thanks for the PR. Agree overall with the sentiment here but I think we should keep the bulleted list to make it simpler for users to wrap their head around.

Thanks @minimalsm and @corwintines. Yes, this page does have a small learning curve, so I think the bulleted list is more readable as well.

PR #4257, opened 2021-10-18
Add IPFS pinning info + comparison vs blockchain embeds

momack2:
Really enjoyed this breakdown! 

I added some more detail about IPFS pinning services and the tradeoffs with other types of directly embedding large-scale blockchain data within chains. 

Hope it's useful!


minimalsm:
Wow @momack2 thanks for the PR! This is great. Be sure to [claim your POAP](https://ethereum.org/en/contributing/#poap) for this 🥳!

PR #4324, opened 2021-10-27
Gatsby v4 upgrade [Fixes #4271]

### LINKED ISSUE #4271: Upgrade Gatsby to v4

Let's upgrade Gatsby to v4! Should help us reduce build times 🤞 

https://www.gatsbyjs.com/docs/reference/release-notes/migrating-from-v3-to-v4/

wackerow:
## Description
- [x] Upgrade codebase to use recently released Gatsby v4
- [x] Upgrade other project dependencies for compatibility (includes `gatsby-plugin-image`)
- [x] Update code base with new syntaxt for images to support `gatsby-plugin-image`
- [x] Test builds on forked Gatsby Cloud account -- Builds passing
- [x] Lambda function builds moved to `public/functions` for compatibility between Gatsby Cloud builds and Netlify serving
- [x] DevOps to set up EF account for Gatsby Cloud
- [x] Set up Gatsby Cloud to build all PR branches for `dev` branch (including `dev`)
- [x] Set up Gatsby Cloud to build our deploy PRs from `dev` into `master`, which will build final production build from `master` branch.
- [ ] Set up Gatby Cloud account to _host_ final production build of `master` branch using existing Netlify account.
- [ ] Update to serve this build as `ethereum.org`

## Related Issue
#4271


pettinarip:
@wackerow I've changed the config for `gatsby-plugin-sitemap` to support the newer version. The problem is the following:

I've used the default configs from the plugin. This means that the `xml`s files are located by default under the `/sitemap` folder.

- Prev version: `https://ethereum.org/sitemap.xml`
- Current version: `https://ethereum.org/sitemap/sitemap-index.xml` and `https://ethereum.org/sitemap/sitemap-0.xml`

Things to consider:
- If we want to keep this new path, we need to make sure that the Google Search Console is reading this path.
- We could do this:
```js
options: {
    output: `/`, // the default value is `/sitemap`
    ...
}
```
- This way, we could have the sitemap like this: `https://ethereum.org/sitemap-index.xml` and `https://ethereum.org/sitemap-0.xml`

netlify.toml
```diff
@@ -14,13 +14,14 @@
   command = "yarn build" # debug with: "yarn build --verbose"
 
   # Directory with the serverless Lambda functions to deploy to AWS.
-  functions = "public/lambda"
+  functions = "public/functions"
```
samajammin, line 17: Do we want to change this? Would we want to rename the directory in the file system also? Or is this a Gatsby v4 thing?

samajammin: All our functions are still in `/src/lambda/` currently.

gatsby-node.js
```diff
@@ -76,7 +80,6 @@ const checkIsMdxOutdated = (path) => {
       intlMatch += match.replace(re, (_, p1, p2) => p1 + p2)
     })
   } catch {
-    console.error("regex error")
```
samajammin, line 79: I believe we still want this error log.

src/content/developers/tutorials/uniswap-v2-annotated-code/index.md
```diff
@@ -382,7 +382,7 @@ This function allows the factory (and only the factory) to specify the two ERC-2
 
 #### Internal Update Functions {#pair-update-internal}
 
-##### \_update {#_update}
+##### \_update {#\_update}
```
samajammin, line 385: Not sure why this change is here - what's the purpose?

src/content/developers/tutorials/uniswap-v2-annotated-code/index.md
```diff
@@ -444,7 +444,7 @@ This price calculation is the reason we need to know the old reserve sizes.
 
 Finally, update the global variables and emit a `Sync` event.
 
-##### \_mintFee {#_mintfee}
+##### \_mintFee {#\_mintfee}
```
samajammin, line 447: Same here

netlify.toml
pettinarip: Yes, this is something we need to change due that we now deploy them from GC. More details here: https://support.gatsbyjs.com/hc/en-us/articles/360054529274-Deploying-Netlify-Functions-from-Gatsby-Cloud

Our source functions still live in `src/lambda`, what has changed is the destination of the compiled functions, `public/functions`. And that destination is the default destination used by Netlify.

Note: I tried keeping this untouched but Netlify wasn't able to find the functions when deployed from GC. It seems that this configuration only works in their CI/CD pipeline (when you link a GH repo and run the builds in their platform). 

Note 2: Besides that, I'm not sure even if the `netlify.toml` is being uploaded from GC. I have the feeling that only the content of the `/public` folder is transferred and Netlify only works as a plain hosting service.

src/content/developers/tutorials/uniswap-v2-annotated-code/index.md
wackerow: Hm, not sure. This appears to have been pulled in when merging `dev` back in, however, noticing that our markdown H5's don't parse these custom headers appropriately as is:

image
image

I feel we should just remove these H5 custom header ids until we can address this issue separately. Can also remove the stray escape `\` that seems to be breaking the others, ie:
image

src/content/developers/tutorials/uniswap-v2-annotated-code/index.md
wackerow: See above

gatsby-node.js
wackerow: Took this out after it was just producing a handful of non-specific errors in the build console. Will replace it with 
```console.error(`regex error in ${englishPath}`)```

wackerow: After chatting with @pettinarip decided to change this to a `console.warn` instead

PR #4335, opened 2021-10-29
Update index.md

Huiuixuan:
A Chinese version.


corwintines:
@Huiuixuan thanks for the contribution. As @lukassim, we are doing our translations in Crowdin, and would be happy to get you setup with that group! Going to close this PR, but please reach out and thank you for the contribution.

PR #4371, opened 2021-11-02
Update consensus mechanisms developer doc

collincusce:
I hope you don't mind, but there was a lot that was factually wrong in this document, so I took a timeout to correct it. 

For example, under "What is consensus?" I had to remove that blockchains required 51% agreement on state. This is false. Anyone can propose a block if they discover a nonce as Proof of Work. Then that block gets added to the chain. If someone finds another block with the same proof of work, they can add to the chain as well. Whoever builds next needs to choose which chain to build on. The 51% portion comes in when someone controls over 51% of the hashpower themselves they can use this to control the chain itself. This is what's known as a safety violation or a 51% attack. It's not the same as having 51% of the network agree.

Rather than write all that out, I decided to remove the incorrect sentence.

There's no such thing as a "consensus mechanism". The correct term is "consensus protocol". As proof, scope out this article... not a primary source, I know, but it's a quick way to show you what I mean.  Ctrl-f "mechanism" then ctrl-f "protocol". See? Not just being anal here, this is the right terminology. The original description used the correct term. https://en.wikipedia.org/wiki/Consensus_(computer_science)

You're going to want to fix the URL path to remove "mechanism" as well.

There's a buncha changes in there. Double check with whoever you need to. 

FYI, casper is susceptible to 34% attacks (actually it's more complicated than that, but it's not worth going over here). Previous version implied 51% and that's not correct. You need 2/3rds positive attestation, which means 1/3rd + 1 of the committee can control the vote. A very easy analysis to follow is found here which back up this claim: https://www.adiasg.me/2020/03/31/casper-ffg-explainer.html


minimalsm:
Hey @collincusce just wanted to say thanks for the PR 😬 Had a quick skim but will try to get this looked at properly later today or early tomorrow.

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -42,57 +40,69 @@ A consensus mechanism in a cryptoeconomic system also helps prevent certain kind
...
 
-Watch more on the different types of consensus mechanisms used on Ethereum:
+Watch more on the different types of consensus protocols used on Ethereum:
 
 <iframe width="100%" height="315" src="https://www.youtube.com/embed/ojxfbN78WFQ" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
 
-### Sybil resistance & chain selection {#sybil-chain}
+### Sybil resistance {#sybil-chain}
+
+Have you ever seen a websit that's being spammed by a ton of fake accounts? These fake accounts are run by bots which try to sell you something. To get around that, a lot of sites employ a system called "captchas". These captchas are difficult for computers to solve but easy for humans, preventing automation from making fake identities. Another way to prevent fake accounts is to force users to pay for a service. This payment associates a cost with automation, preventing armies of bots from taking over a site.
```
nassersaazi, line 87: Have you ever seen a website*

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -42,57 +40,69 @@ A consensus mechanism in a cryptoeconomic system also helps prevent certain kind
...
+Nakamoto consensus is done by [miners](/developers/docs/consensus-mechanisms/pow/mining/) who compete to create new blocks full of processed transactions. They do this through a process known as "proof-of-work" mining (PoW). In PoW, miners are creating random numbers, called a nonce, and plugging that into a hash function to see if they produce a hash which satisfies a condition difficult enough to show they've put work into solving the problem. Once discovered, the miner who found the nonce proposes a new block to the rest of the network. In this new block, the miner pays itself in ETH. If the network decides to continue to build on his block, it becomes part of the "cannonical chain" and the miner has some freshly minted ETH. 
+
+The race is won by whosever computer can solve a math puzzle fastest – this produces the cryptographic link between the current block and the block that went before. Solving this puzzle puts the work in "proof-of-work".
```
nassersaazi, line 57: Isn't it supposed to be "The race is won by whichever* computer..."?

minimalsm: @nassersaazi whosever is technically correct.

> Whosever: belonging to or associated with whichever person; whoever's.

But given that miners are largely server farms now (instead of home miners) I'd be good with changing this.

minimalsm (changes requested):

minimalsm:
Hey @collincusce, sorry for the delay on this. Going to try to get this done this week.

minimalsm (changes requested):
Thanks again for the PR @collincusc. 

For context: to keep this page approachable we chose to start at a high-level, (i.e. we hide some information (nakamoto/proof-of-work) and reveal it slowly to build up a mental model). I'll make some tweaks but I think a lot of these changes are valuable. Thanks again.

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -48,51 +46,63 @@ A consensus mechanism in a cryptoeconomic system also helps prevent certain kind
...
 
-### Proof-of-stake {#proof-of-stake}
+### Casper FFG {#casper-ffg}
 
 Ethereum has plans to upgrade to a **proof-of-stake (PoS)** consensus protocol.
 
 #### Block creation {#pos-block-creation}
 
-Proof-of-stake is done by validators who have staked ETH to participate in the system. A validator is chosen at random to create new blocks, share them with the network and earn rewards. Instead of needing to do intense computational work, you simply need to have staked your ETH in the network. This is what incentivises healthy network behaviour.
+Casper FFG is done by validators who have staked ETH to participate in the system. A validator is chosen at random to create a new block. This randomness constitutes the proposal mechanism that would otherwise be the "work" in proof-of-work for Nakamoto consensus. Once it's a validator's turn to propose a block, they share it with the network and earn rewards. Instead of needing to do intense computational work, they'd simply need to have staked their ETH in the network. This is what incentivises healthy network behaviour as the validator now has skin in the game.
```
wackerow, line 71: Per our style guide, declaring the acronym meaning first, and switched terms to "American English"
```suggestion
**Casper the Friendly Finality Gadget (Casper FFG)** is utilized by validators who have staked ETH to participate in the system. A validator is chosen at random to create a new block. This randomness constitutes the proposal mechanism that would otherwise be the "work" in proof-of-work for Nakamoto consensus. Once it's a validator's turn to propose a block, they share it with the network and earn rewards. Instead of needing to do intense computational work, they'd simply need to have staked their ETH in the network. This is what incentivizes healthy network behavior as the validator now has skin in the game.
```

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -48,51 +46,63 @@ A consensus mechanism in a cryptoeconomic system also helps prevent certain kind
...
 
 #### Security {#pos-security}
 
-A proof-of-stake system is kept secure by the fact that you'd need 51% of the total staked ETH to defraud the chain. And that your stake is slashed for malicious behaviour.
+A proof-of-stake system is kept secure by the fact that you'd need 34% of the total voting power in a committee to defraud the chain. This encouraged further in that, on Eth2, your stake is slashed for malicious behaviour.
```
wackerow, line 75: Wanted to clarify, is this accurate that >1/3 voting power can _defraud_ the chain? Despite being more than half of a supermajority, this still wouldn't give an attacker a supermajority to finalize fraudulent changes, though it could certainly enable a DOS attack by preventing supermajority consensus. 

Also, the "eth2" term is being phased out. Updates to the rest of the site are coming, but I would remove "Eth2" here in favor of "proof-of-stake"

Perhaps:
```suggestion
A proof-of-stake system is kept secure by the fact that you'd need 34% of the total voting power in a committee to defraud the chain. Malicious behavior under proof-of-stake also result in staked ETH being slashed in proportion to the size of the attack.
```

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -48,51 +46,63 @@ A consensus mechanism in a cryptoeconomic system also helps prevent certain kind
...
 <iframe width="100%" height="315" src="https://www.youtube.com/embed/ojxfbN78WFQ" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
 
-### Sybil resistance & chain selection {#sybil-chain}
+### Sybil resistance {#sybil-chain}
+
+Have you ever seen a website that's being spammed by a ton of fake accounts? These fake accounts are run by bots which try to sell you something. To get around that, a lot of sites employ a system called "captchas". These captchas are difficult for computers to solve but easy for humans, preventing automation from making fake identities. Another way to prevent fake accounts is to force users to pay for a service. This payment associates a cost with automation, preventing armies of bots from taking over a site.
+
+Blockchains have a form of captcha of their own. This process of preventing fake identities from taking over a network is called "Sybil resistance". Like captcha or premium accounts on a website, Sybil resistance mechanisms associate some kind of work or value to participating in the blockchain. This causes an innate investment in the success of the network.
```
wackerow, line 89:
```suggestion
Blockchains have a form of captcha of their own. This process of preventing fake identities from taking over a network is called "Sybil resistance." Like captcha or premium accounts on a website, Sybil resistance mechanisms associate some kind of work or value to participating in the blockchain. This causes an innate investment in the success of the network.
```

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -48,51 +46,63 @@ A consensus mechanism in a cryptoeconomic system also helps prevent certain kind
...
+
+**Sybil resistance** describes how a protocol fares against a [Sybil attack](https://en.wikipedia.org/wiki/Sybil_attack). Sybil attacks are when one user or group pretends to be many users. Resistance to this type of attack is essential for a decentralized blockchain and enables miners and validators to be rewarded equally based on resources put in. Proof-of-work and proof-of-stake protect against this by making users expend a lot of energy or put up a lot of collateral. These protections are an economic deterrent to Sybil attacks.
+
+Proof-of-work and proof-of-stake are not consensus protocols by themselves, but since each blockchain needs some form of Sybil resistance, it's easy to conflate the two concepts. They actually are Sybil resistance mechanisms and block author selectors. They are a way to decide who is the author of the latest block. It's this Sybil resistance mechanism combined with proposal mechanism that makes up a truly decentralized consensus protocol.
```
wackerow, line 93: 👍🏻

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -48,51 +46,63 @@ A consensus mechanism in a cryptoeconomic system also helps prevent certain kind
...
+
+### Proposal mechanism {#proposal-mechanism}
 
-Now technically, proof-of-work and proof-of-stake are not consensus protocols by themselves, but they are often referred to as such for simplicity. They are actually Sybil resistance mechanisms and block author selectors; they are a way to decide who is the author of the latest block. It's this Sybil resistance mechanism combined with a chain selection rule that makes up a true consensus mechanism.
+A proposal mechanism is what's used to determine the next block proposed on the network. In Nakamoto consensus, a person who mines a winning nonce gets the right to be the leader for the next block. They propose that block to the network. In Eth2, a random beacon decides the order of proposers from the list of available validators. In both cases, after a block is proposed, it's up to the next proposer to decide if they want to build on that block or on another block proposed in the case of competing proposals. This is what's known as the "chain selection rule".
```
wackerow, line 97: Again, we should eliminate usage of "Eth2" here. Perhaps `Under a proof-of-stake protocol,` instead of `In Eth2,`

wackerow (changes requested):
@collincusce thanks for submitting this! 

Can't disagree with this being a protocol, but would argue that the term "mechanism" encompasses a protocol, and is commonly referenced this way in the community. For SEO purposes, not sure we should remove this term entirely, but not against the transition in general. Curious others thoughts on completely renaming the file paths, or perhaps we could just add a statement that clarifies this so people aren't confused thinking this is something completely different. 

Left a couple other comments, but definitely appreciate your suggested changes.

minimalsm:
@collincusce just checking in you've seen the change requests on this 🙏.

minimalsm:
@collincusce any update?

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -18,65 +18,75 @@ To better understand this page, we recommend you first read our [introduction to
...
+**Casper the Friendly Finality Gadget (Casper FFG)** is utilized by validators who have staked ETH to participate in the system. A validator is chosen at random to create a new block. This randomness constitutes the proposal mechanism that would otherwise be the "work" in proof-of-work for Nakamoto consensus. Once it's a validator's turn to propose a block, they share it with the network and earn rewards. Instead of needing to do intense computational work, they'd simply need to have staked their ETH in the network. This is what incentivizes healthy network behavior as the validator now has skin in the game.
 
 #### Security {#pos-security}
 
-A proof-of-stake system is kept secure by the fact that you'd need 51% of the total staked ETH to defraud the chain. And that your stake is slashed for malicious behaviour.
+A proof-of-stake system is kept secure by the fact that you'd need 34% of the total voting power in a committee to defraud the chain. Malicious behavior under proof-of-stake also result in staked ETH being slashed in proportion to the size of the attack.
```
wackerow, line 59: Could you explain what you mean here by "defraud the chain?"

My understanding is a supermajority (67%) vote is required to justify and finalize epochs, thus if someone controlled 34% they could grind the network to a halt by preventing the chain from finalizing, but they would not be able to finalize their own malicious blocks on their own.

wackerow (changes requested):
Hey @collincusce, got a chance to take a look at this again. Overall I really like it and you do make some good points about the previous wording. I think this is a nice improvement. 

I added a commit with just a couple small tweaks to account for the Eth2 rebranding, and merged it with dev (which seems to have made a bit of a mess of the PR with autoformatting, sorry about that)

Just curious your take on the comment I left about the 34% needed to defraud the chain. Let me know and we'll try to get this in real soon!

src/content/developers/docs/consensus-mechanisms/index.md
collincusce: Wow I totally forgot I did this. 

Honestly, I don't like that term "defraud the chain" ... so much so I question whether I wrote it in the first place LMAO! I probably did, but you're correct, you could halt... 

...or you could do what's known as "equivocation". If you give Coinbase one transaction and FTX another. Then you get to choose which epoch you like and go with it, censoring the partition of the network you don't want to execute. If you're in control over 34% you can do this to the network. I believe this is still possible. Double check with your lads internally. I'm just a guy who learned a thing or two in his brief stint at Ava Labs. I'm hardly a world-renowned expert.

collincusce: Let me rephrase that a bit... I said "Coinbase" and "FTX" but what I really should be saying is "32.5%" in both cases. You own (at least) 34%... You split the difference on the other two transactions by sending half one version of a transaction, another half the other version, and then wait until the epoch to see which version favors you the most.

jmcook1186: I don't think this can happen. 
 
Double-voting - i.e. voting for multiple targets for the same slot - is explicitly a slashable offence and because in this case it is backed by 33% of the total staked ether it would also trigger the inactivity leak. The inactivity leak would cause the clients on the minority fork to gradually bleed their stake, representing a smaller and smaller proportion of the total, eventually allowing the chain to begin finalizing again once a 66% supermajority is re-established. The malicious validators attempting a double-vote would get slashed and ejected from the network.

So a 33% stake is sufficient to halt the Beacon Chain from finalizing, but only temporarily.

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -18,65 +18,75 @@ To better understand this page, we recommend you first read our [introduction to
...
+The race is won by whosever computer can solve a math puzzle fastest – this produces the cryptographic link between the current block and the block that went before. Solving this puzzle puts the work in "proof-of-work."
 
 #### Security {#pow-security}
 
 The network is kept secure by the fact that you'd need 51% of the network's computing power to defraud the chain. This would require such huge investments in equipment and energy; you're likely to spend more than you'd gain.
 
 More on [proof-of-work](/developers/docs/consensus-mechanisms/pow/)
 
-### Proof-of-stake {#proof-of-stake}
+### Casper FFG {#casper-ffg}
```
jmcook1186, line 49: is it right to jump straight into Casper-FFG here? FFG is the finality gadget that compares votes on pairs of checkpoints to give finality to the chain, but it's not the whole story of PoS consensus. There are also LMD-GHOST votes that vote on the head of the chain. Consensus over the head of the chain comes from LMD-GHOST, finality comes from FFG.

So the closest equivalent to ETH1/Bitcoin's Nakamoto consensus mechanism is LMD-GHOST, but instead of the longest chain (which is really the chain with the most accumulated work rather than the longest) it is the chain that has accumulated the greatest attestations backed by the most staked ether that becomes canonical. I think this nuance prevents LMD-GHOST from strictly being Nakamoto consensus. I guess it could be called "Nakamoto inspired"...!

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -18,65 +18,75 @@ To better understand this page, we recommend you first read our [introduction to
...
+
+### Proposal mechanism {#proposal-mechanism}
 
-Now technically, proof-of-work and proof-of-stake are not consensus protocols by themselves, but they are often referred to as such for simplicity. They are actually Sybil resistance mechanisms and block author selectors; they are a way to decide who is the author of the latest block. It's this Sybil resistance mechanism combined with a chain selection rule that makes up a true consensus mechanism.
+A proposal mechanism is what's used to determine the next block proposed on the network. In Nakamoto consensus, a person who mines a winning nonce gets the right to be the leader for the next block. They propose that block to the network. In the new proof-of-stake system, a random beacon decides the order of proposers from the list of available validators. In both cases, after a block is proposed, it's up to the next proposer to decide if they want to build on that block or on another block proposed in the case of competing proposals. This is what's known as the "chain selection rule."
```
jmcook1186, line 82: do you want to mention validator selection via RANDAO here for block proposal on the Beacon Chain?

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -18,65 +18,75 @@ To better understand this page, we recommend you first read our [introduction to
...
 
-**Sybil resistance** measures how a protocol fares against a [Sybil attack](https://wikipedia.org/wiki/Sybil_attack). Sybil attacks are when one user or group pretends to be many users. Resistance to this type of attack is essential for a decentralized blockchain and enables miners and validators to be rewarded equally based on resources put in. Proof-of-work and proof-of-stake protect against this by making users expend a lot of energy or put up a lot of collateral. These protections are an economic deterrent to Sybil attacks.
+### Chain selection rule {#chain-selection-rule}
```
jmcook1186, line 83: Kind of splitting hairs here but it would be more consistent with other Ethereum literature and pages on ethereum.org to refer to "fork choice algorithm" or "fork choice rule" rather than "chain selection rule".

Also see previous comment regardng LMD-GHOST as the fork-choice algorithm on the Beacon Chain.

src/content/developers/docs/consensus-mechanisms/index.md
```diff
@@ -18,65 +18,75 @@ To better understand this page, we recommend you first read our [introduction to
...
+A **chain selection rule** is used to decide which chain is the "correct" chain. Ethereum and Bitcoin use the "longest chain" rule, which means that whichever blockchain is the longest will be the one the rest of the nodes accept as valid and work with. For proof-of-work chains, the longest chain is determined by the chain's total cumulative proof-of-work difficulty.
 
-The combination of proof-of-work and longest chain rule is known as "Nakamoto Consensus."
+The combination of proof-of-work and longest chain rule makes up Nakamoto consensus.
 
-The [Beacon Chain](/upgrades/beacon-chain/) uses a consensus mechanism called [Casper the Friendly Finality Gadget](https://arxiv.org/abs/1710.09437), which is proof-of-stake based.
+The [Beacon Chain](/eth2/beacon-chain/) uses a consensus protocol called [Casper the Friendly Finality Gadget](https://arxiv.org/abs/1710.09437), which is proof-of-stake based.
```
jmcook1186, line 89: see earlier comment that FFG is for finalizing the Beacon Chain, it is not equivalent to the longest chain rule - the closest equivalent is LMD-GHOST.

src/content/developers/docs/consensus-mechanisms/index.md
collincusce: Yeah but the person responsible for slashing is the person with 34%. Slashing doesn't work. At all. We've been saying this.

collincusce: Oh and then the question is "well he wont be staked forever". Perhaps, but this person can censor any new staking transactions which usurps his control over the network as well. This is why at 34% attack is so incredibly damaging. Once that safety threshold is crossed, they have control over the entire network. All. Bets. Are. Off.

jmcook1186: No, sorry but that's not right. 

The person with 33% stake has no control over who is slashed. Slashing happens whenever a validator attests to multiple blocks for the same slot (there are other slashable offences too). This is detected by other validators on the network, and they get a small reward for pointing it out. The magnitude of the slashing would be especially high because a large number of validators are involved (correlation penalty) and the inactivity leak would gradually bleed their stake, re-establishing a supermajority attesting to the majority fork. 

A 33% stake does not allow you to choose who gets slashed.

collincusce: You really are not understanding this. Consensus is the process by which all databases are kept in-sync. The consensus protocol ensures that the slashing occurs. You can present all of the evidence that you want, but the person with 34% has control over making sure that the entire network comes to agreement on what the database (balances) look like. Individual nodes can update their databases to be in disagreement with the 34% node, but that doesn't actually ensure consensus. It's likely to throw those other nodes into an incorrect state causing them to get slashed.

Ok let me give you an example. Coinbase and FTX decide to collude against Binance. Together they have over 34% of the stake. They decide they're going to siphon money from Binance by presenting the network with two options: if number goes up half the network gets option A, if number goes down the other half gets option B. Then they attest positively to the option that benefits them the most and Binance the least. This double vote is only seen by the validators that they shared it to. Binance is completely cut out of the equation entirely. Their votes don't matter. Even if a double vote is detected, the other nodes do not have the power to ensure that a database update is replicated across all nodes. The double vote is meaningless. There is no mechanism design that can rectify this. 

All of that is meaningless though. The question of which vote is the correct vote ... there's only one vote. The 34% chose to cast ONE VOTE on the epoch. They cast a single vote on the correct state and the other 33% saw the wrong one but doesn't have the voting power to make a difference. 

Anyway. I've said all I can about this. It's immaterial, you've designed things as you want. I'll step out and wish y'all good luck.

minimalsm: Thanks for the input @collincusce. I've asked some people working on pos consensus to weigh in here.

vbuterin:
I don't support much of this.

The words "consensus mechanism" are used instead of "consensus protocol" for a very specific reason: the word "mechanism" (cf. "mechanism design") captures the fact that it's not just a set of rules for how validators should behave, but also incentives to encourage validators to behave correctly.

The percent needed to attack consensus is a complicated topic. If the network is asynchronous, then you can cause double-finality with 34% of stake. But if the network is synchronous (same assumption as PoW), then it takes 51% of stake to revert the safe head, though you could _prevent_ the chain from growing with somewhere around 20-30%.

> He'll censor any transaction claiming he's slashing. Slashing doesn't work. At all. It only penalizes well-intentioned actors while completely failing to assist in stopping bad actors. We've been saying this.

Censoring slashed transactions doesn't work because it's an assumption of the system that if an attacker attempts to censor, honest users and validators will coordinate on a minority fork that includes the slashing transactions and leaks the attacker (this is a soft fork, NOT a hard fork; no "irregular state changes" are required).

collincusce:
> The words "consensus mechanism" are used instead of "consensus protocol" for a very specific reason: the word "mechanism" (cf. "mechanism design") captures the fact that it's not just a set of rules for how validators should behave, but also incentives to encourage validators to behave correctly.

I hear you, you want to differentiate the entire system from the protocol itself. Well that line is pretty blurry, but I get why you could want to do that. For instance, Proof of Stake vs Proof of Work... that's Sybil, not consensus. However, there's a lot of ways to do both, so it's not a good enough distinction to wrap it up into just "mechanism design" and call it a consensus mechanism. This was one of my original points here, one I hope you agree with. 

I'm a bit concerned with this:

> The percent needed to attack consensus is a complicated topic. If the network is asynchronous, then you can cause double-finality with 34% of stake. But if the network is synchronous (same assumption as PoW), then it takes 51% of stake to revert the safe head, though you could prevent the chain from growing with somewhere around 20-30%.

Assuming you're talking specifically about Eth2 and epochs. Huge apologies in advance if I misunderstood something. As I've said a lot in here, I'm neither a world-renowned expert in consensus protocols nor Eth2 (especially since it's been quite a moving target over the years), but I was under the impression that transactions are voted on immediately. If so, that's asynchronous, no? Then Eth2 tries to finalize them with the epoch requiring the a 2/3rds majority vote? Which is exactly the same as a 1/3rd safety threshold for the reasons I mentioned (having 34% and splitting the vote on the remainder, no?). I'm a bit confused where 51% comes in...

... What you said is true, but is it true for Eth2? The reason I ask is because I keep hearing different things. Like I said, not an expert, moving target, yadda yadda. I just want to be correct, I don't care about being (personally) right, so help me out.

I want to remind that if 33% is the safety threshold for a protocol, and you add an element that is synchronously safe to 50% safety, that element is the stronger assumption but it's the WEAKER assumption which limits the entire system. By this I mean that the system is only as safe as the weakest part, which is 33%. 

> Censoring slashed transactions doesn't work because it's an assumption of the system that if an attacker attempts to censor, honest users and validators will coordinate on a minority fork that includes the slashing transactions and leaks the attacker (this is a soft fork, NOT a hard fork; no "irregular state changes" are required).

You mean like Ethereum Classic did?

minimalsm:
I think the conversation here is getting circuitous, and we're not going to reach a productive conclusion by carrying on here.

I think the clear next steps are:

1. Integrate the existing improvements that aren't up for debate (mostly the Sybil resistance improvements). I'm happy to pick this up and credit you @collincusce. Should be able to do it by the end of this week.

3. Close this PR

5. Create a ticket for adding clarity on 34%/51% and ask for a researcher to help us write this clarification.

7. Get a review from a proof-of-stake researcher to ensure the page comprehensively covers consensus mechanisms

Appreciate you highlighting these @collincusce

minimalsm:
@all-contributors please add @collincusce for docs, ideas, feedback

newsize686 (commented):
src/components/EthPriceCard.js

PR #4376, opened 2021-11-02
Policy updates, adds Matomo opt-out option

wackerow:
## Description
- Adds component that offers user the ability to opt-out of Matomo tracking. (`MatomoOptOut.js`)
- Uses `localStorage` to store this preference, and `utils/matomo.js` will check this before submitting an event, returning early if set to `true`
- Updated the cookie and privacy policies per legal team, adding the above component


minimalsm (approved):
LGTM. Tested on BrowserStack and for accessibility with a screenreader.

PR #4382, opened 2021-11-03
Fix typos in security learn page

bskrksyp9:
Fix typos in security doc

## Description
Corrected typos on line number 28, 94 and 98., however I am quite skeptical about it, because some of them might be used as 'CSS Selectors' for webpage, so these changes might affect


minimalsm (approved):
Thanks @bskrksyp9 😀

PR #4386, opened 2021-11-04
Small grammar fix

justincjohnson:
## Description

Change `The amount the base fee is adjusted by is proportional to how far from the current block size is from the target.` to `The amount the base fee is adjusted by is proportional to how far the current block size is from the target.`


justincjohnson:
Closing in favor of a new PR with more fixes.  Sorry for the hassle.  First time contributor.

PR #4397, opened 2021-11-05
Fix punctuation in Readme doc

bskrksyp9:
Fix punctuation

## Description
Removed space for words on line number 208, 216


corwintines (approved):
Thanks @bskrksyp9

PR #4437, opened 2021-11-10
Remove metamask weekend event from community page

minimalsm:
## Description
Removes an event added to wrong place on community page 

## Related Issue
#4291


wackerow (approved):
Thanks @minimalsm!

PR #4429, opened 2021-11-09
Add useWeb3 to community page

wslyvh:
Adding 2 links to the community page
- Main website https://www.useweb3.xyz/
- Jobs https://www.useweb3.xyz/jobs

PR #4439, opened 2021-11-10
Broaden article listing policy

samajammin:
## Description

@minimalsm noted that we didn't have a policy around job boards - I suggest we broaden our existing article policy to encompass a wider array of community content resources. I think it better captures the criteria we should assess vs. evaluating something like a jobs board as a product, which we scrutinize much more for security purposes.

## Related Issue
https://github.com/ethereum/ethereum-org-website/pull/4429


wackerow (approved):

wackerow:
image

The diff looks good, not sure why it's causing the build to fail. Netlify logs are a bit of a challenge to load up, but I was able to get to it so I'm pasting it here for reference.

minimalsm (approved):

PR #4441, opened 2021-11-10
Support new content version release in codebase

corwintines:
## Description
Changes to gatsby-node to parse markdown and json files for translation. 

With JSON we now check:
- if translation file exists
- If there are the same amount of keys
- If all the keys are the same

With Markdown we now check:
- The parsed info structure of the mdx


samajammin:
This looks great at a glance @corwintines! I still need to review this in depth. Will do so tomorrow...

One quick thought since we're using the header ids to compare markdown files - we should ensure we have header ids generated for all our markdown files. i.e. please run this script on all English files: https://github.com/ethereum/ethereum-org-website/blob/dev/src/scripts/generate-heading-ids.js

I think ideally we run this as a pre-commit hook on any English file but I'm ok with saving that for a later issue.

gatsby-node.js
```diff
@@ -42,6 +42,107 @@ const getMessages = (path, language) => {
...
+    englishData.match(re).forEach((match) => {
+      englishMatch += match.replace(re, (_, p1, p2) => p1 + p2)
+    })
+    translatedData.match(re).forEach((match) => {
+      intlMatch += match.replace(re, (_, p1, p2) => p1 + p2)
+    })
+
+    return englishMatch !== intlMatch
+  } catch (err) {
+    return true
```
pettinarip, line 75: I would make this catch more granular. As you describe in your doc, "If there is not english file, return true", I would lean toward this:
```ts
const splitPath = path.split("/")
splitPath.splice(7, 2)
const englishPath = splitPath.join("/")

try {
  const englishData = fs.readFileSync(englishPath, "utf-8")
} catch (err) {
  return true
}

// rest of the logic
...
```
Because if any other error happens (like not having a match), the consumer is going to be able to handle that, in the way they want.

In this case, we could `console.error` and error if no header id was found.
```ts
try {
  isOutdated = await checkIsMdxOutdated(node.fileAbsolutePath)
} catch (err) {
  // log
  console.error(`No header id found in file ${node.fileAbsolutePath}`, err)
  // OR throw, in case we want to stop the build
  throw new Error(`No header id found in file ${node.fileAbsolutePath}`)
}
```

gatsby-node.js
```diff
@@ -42,6 +42,107 @@ const getMessages = (path, language) => {
...
 }
 
+/**
+ * Markdown isOutdated check
+ * Parse headers in markdown file (both translated and english) and compare their info structure.
+ * If this structure is not the same, then the file isOutdated.
+ * If there is not english file, return true
+ * @param {string} path filepath for translated mdx file
+ * @returns boolean for if file is outdated or not
+ */
```
pettinarip, line 52: Good docs 👍🏼

gatsby-node.js
```diff
@@ -42,6 +42,107 @@ const getMessages = (path, language) => {
...
+        }
+      }
+
+      return {
+        isOutdated: false,
+        isContentEnglish: false,
+      }
+    }
+  } catch {
+    return true
```
pettinarip, line 142: I don't like this return that is absorbing all the errors and is different from the return type from the function.

Let the function throw its errors and that the consumer handles what to do with it.

pettinarip (changes requested):
Great work @corwintines. I have added a couple of comments to refactor the code a bit.

gatsby-node.js
```diff
@@ -51,7 +152,10 @@ const outdatedMarkdownPages = [
   "/what-is-ethereum/",
 ]
 
-exports.onCreateNode = ({ node, getNode, actions }) => {
+// Loops through all the files dictated by Gatsby (building pages folder), as well as
+// folders flagged through the gatsby-source-filesystem plugin in gatsby-config
+// This seems to be where the source and transform holdup is?
```
pettinarip, line 157: I would suggest removing the comments with questions from the code. I think we can discuss them off record than leaving them there.

And regarding the question, hmm not sure. I think it is more related with the queries execution, when gatsby is preparing to build the pages.

gatsby-node.js
```diff
@@ -42,6 +42,107 @@ const getMessages = (path, language) => {
   }
 }
 
+/**
+ * Markdown isOutdated check
+ * Parse headers in markdown file (both translated and english) and compare their info structure.
```
samajammin, line 47:
```suggestion
 * Parse header ids in markdown file (both translated and english) and compare their info structure.
```
Let's make it clear we're talking about the "header id" (open to another name) - ultimately it's the [URI fragment](https://en.wikipedia.org/wiki/URI_fragment) of the header. The reason we do this is that the header text will differ from language to language & we want to keep links consistent.

gatsby-node.js
samajammin: I like the idea of a console error - no need to disrupt the build IMO.

gatsby-node.js
```diff
@@ -42,6 +42,107 @@ const getMessages = (path, language) => {
...
+  } catch (err) {
+    return true
+  }
+}
+
+/**
+ * JSON isOutdated check
+ * Checks if translated JSON file exists.
+ * If translated file exists, checks that all translations are present (checks keys), and that all the keys are the same.
+ * If translation file exists, isContentEnglish will be false
```
samajammin, line 83: Is there a difference between `translated file` & `translation file`? If not let's use the same term.

gatsby-node.js
```diff
@@ -42,6 +42,107 @@ const getMessages = (path, language) => {
...
+        isOutdated: false,
+        isContentEnglish: false,
+      }
+    }
+  } catch {
+    return true
+  }
+}
+
 const outdatedMarkdownPages = [
```
samajammin, line 146: I don't believe `outdatedMarkdownPages` are used anymore? ✂️

gatsby-node.js
```diff
@@ -63,10 +167,8 @@ exports.onCreateNode = ({ node, getNode, actions }) => {
       slug = slug.replace("/translations", "")
       const split = slug.split("/")
       split.splice(1, 1)
-      const originalPath = split.join("/")
-      if (outdatedMarkdownPages.includes(originalPath)) {
-        isOutdated = true
-      }
+
+      isOutdated = await checkIsMdxOutdated(node.fileAbsolutePath)
```
samajammin, line 171: Would it make sense to just work with relative paths? We already use this logic down below:
```
    const absolutePath = node.fileAbsolutePath
    const relativePathStart = absolutePath.indexOf("src/")
    const relativePath = absolutePath.substring(relativePathStart)
```

samajammin: Ah does `fs` need the absolute path? Probably 😄

gatsby-node.js
```diff
@@ -42,6 +42,107 @@ const getMessages = (path, language) => {
...
+ * If this structure is not the same, then the file isOutdated.
+ * If there is not english file, return true
+ * @param {string} path filepath for translated mdx file
+ * @returns boolean for if file is outdated or not
+ */
+const checkIsMdxOutdated = (path) => {
+  try {
+    const splitPath = path.split("/")
+    splitPath.splice(7, 2)
+    const englishPath = splitPath.join("/")
```
samajammin, line 57: I'm not sure this is right - don't you need to replace the lang (e.g. `ar`) with `en` in the path?

samajammin: I think as-is this only drops the `/content/translations/` from the path.

gatsby-node.js
```diff
@@ -272,29 +375,32 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
     createPage({
       ...page,
       context: {
         ...page.context,
         isOutdated,
         //display TranslationBanner for translation-component pages that are still in English
-        isContentEnglish:
-          langVersion < 2 && !page.component.includes("/index.js"),
+        isContentEnglish,
+        // // We can check if isContentEnglish through our crawling
```
samajammin, line 402: ✂️

samajammin: Also I believe `langVersion` is no longer used, so can delete that variable.

samajammin:
General note on this refactor - this now removes the concept of `version` for each language, so I believe we can delete that key for each language here: https://github.com/ethereum/ethereum-org-website/blob/dev/src/data/translations.json

This also means logic that relies on that should be removed/refactored, e.g. any use of `getLangContentVersion`.

gatsby-node.js
corwintines: YEah makes the fs readfile easier.

gatsby-node.js
corwintines: I need to parse out without __dirname I think to get this right in hindsight. Will add that in.

gatsby-node.js
```diff
@@ -42,16 +42,121 @@ const getMessages = (path, language) => {
...
+ * Checks if translation JSON file exists.
+ * If translation file exists, checks that all translations are present (checks keys), and that all the keys are the same.
+ * If translation file exists, isContentEnglish will be false
+ * @param {*} path url path used to derive file path from
+ * @param {*} lang language abbreviation for language path
+ * @returns {{isOutdated: boolean, isContentEnglish: boolean}}
+ */
+const checkIsPageOutdated = async (path, lang) => {
+  // Files that need index appended on the end. Ex page-index.json, page-developers-index.json, page-eth2-index.json
+  const indexFilePaths = ["", "developers", "eth2"]
```
samajammin, line 97: Should we just change these file names so we don't have to do this extra logic? Perhaps a question for Luka if that would affect anything in Crowdin?

samajammin: Or I suppose we'd need something like this for the homepage, since it's path is `/`?

gatsby-node.js
```diff
@@ -272,29 +376,31 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
+// Only ran when creating component pages
 // Add additional context to translated pages
 // https://www.gatsbyjs.com/docs/creating-and-modifying-pages/#pass-context-to-pages
-exports.onCreatePage = ({ page, actions }) => {
+exports.onCreatePage = async ({ page, actions }) => {
   const { createPage, deletePage } = actions
 
   const isTranslated = page.context.language !== defaultLanguage
   const hasNoContext = page.context.isOutdated === undefined
   const langVersion = getLangContentVersion(page.context.language)
```
samajammin, line 387: We're going to need to re-think how we do this, since version is no longer on a language level.

gatsby-node.js
corwintines: The other two we could probably get away with taking out the -index portion, but the homepage is the tricky one yeah.

src/pages/developers/tutorials.js
```diff
@@ -241,11 +240,7 @@ const TutorialsPage = ({ data, pageContext }) => {
 
   const allTutorials = []
     .concat(externalTutorials, internalTutorials)
-    .filter((tutorial) =>
-      hasTutorials(pageContext.language)
-        ? tutorial.lang === pageContext.language
-        : tutorial.lang === "en"
-    )
+    .filter((tutorial) => tutorial.lang === pageContext.language)
```
samajammin, line 243: I don't think this logic quite gives us what we want. If a language doesn't have any tutorials, we'll now display 0 tutorials for that language, e.g.
Image 2021-11-23 at 9 56 08 PM

What we want is to:
1) if the language has translated tutorials, display those
2) if the language doesn't have translated tutorials, display our English tutorials

I feel like we discussed just refactoring the `hasTutorials` to be a helper function in this file that iterates through the tutorials & to return `true` as soon as it hits one in that native language. If it doesn't find one, return `false`. Basically lodash's `any` or `some`: https://lodash.com/docs/4.17.15#some

samajammin: (we have lodash installed in the project, so you should be able to use it).

gatsby-node.js
```diff
@@ -207,94 +312,66 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
...
+  // This avoids overwriting markdown pages with a component page of the same name.
+  // Note: once all these markdown pages have been replaced with updated JSON translation files,
+  // we can remove this logic and the `/pages-conditional/` directory.
+  const outdatedMarkdown = [`eth`, `dapps`, `wallets`, `what-is-ethereum`]
+  outdatedMarkdown.forEach((page) => {
+    supportedLanguages.forEach((lang) => {
+      const markdownPath = `${__dirname}/src/content/translations/${lang}/${page}/index.md`
+      const langHasOutdatedMarkdown = fs.existsSync(
+        markdownPath,
+        fs.constants.R_OK | fs.constants.W_OK
```
samajammin, line 325: Is this 2nd param needed? What's it do? I notice we don't use it in `fs.existsSync` elsewhere.

corwintines: Probably not. But using existsSync to check if the file exists is all.

src/pages/developers/tutorials.js
corwintines: Ah gotcha, my mistake. Will get that added in now!

src/pages/index.js
```diff
@@ -410,12 +406,6 @@ const HomePage = ({ data }) => {
   const [isModalOpen, setModalOpen] = useState(false)
   const [activeCode, setActiveCode] = useState(0)
 
-  // Language versions 2.4 & above support this homepage content
-  // If current language is below, render LegacyPageHome
-  if (getLangContentVersion(intl.locale) < 2.4) {
-    return <LegacyPageHome />
```
samajammin, line 416: We can delete this LegacyPageHome file since we no longer use it anywhere.

corwintines:
Thanks for the thorough review everyone, I think we are at the finish line now

minimalsm:
I think we might have to add some additional logic for the tutorial page. I believe current logic will always show the outdated banner? (we don't list every single tutorial on CrowdIn as we also have external tutorials)

Screenshot 2021-11-26 at 16 57 07

minimalsm (commented):
I think we might have to add some additional logic for the tutorial page. I believe current logic will always show the outdated banner? (we don't list every single tutorial on CrowdIn as we also have external tutorials)

Screenshot 2021-11-26 at 16 57 07

wackerow:
Getting a buggy layout when the english tutorials are being listed... The `pl`/`ro`/`zh` langs, which have translated tutorials available, are all being displayed correctly.

image
image

wackerow:
Noticing the banner logic on the homepage is now showing the "See English" option, when the template is already showing all the translated strings possible and defaulting to English for the rest of them.
image
By comparison, at https://ethereum.org/cs/developers/tutorials/ you'll notice a similar setup, where we default to the English strings, thus we don't offer the "See English" button
image
Forget if we decided that this was okay for a brief period while we await the finalized translations, but wanted to note it.

src/pages/index.js
```diff
@@ -410,11 +407,10 @@ const HomePage = ({ data }) => {
   const [isModalOpen, setModalOpen] = useState(false)
   const [activeCode, setActiveCode] = useState(0)
 
-  // Language versions 2.4 & above support this homepage content
-  // If current language is below, render LegacyPageHome
-  if (getLangContentVersion(intl.locale) < 2.4) {
+  if (intl.locale !== "en") {
     return <LegacyPageHome />
   }
+  console.log(intl)
```
samajammin, line 413: ✂️

src/content/developers/tutorials/how-to-use-tellor-as-your-oracle/index.md
```diff
@@ -13,7 +13,7 @@ sourceUrl: https://docs.tellor.io/tellor/
 
 Pop Quiz: Your protocol is just about finished, but it needs an oracle to plug in to get an active off chain data feed going...What do you do?
 
-## (Soft) Prerequisites
+## (Soft) Prerequisites {#in-this-tutorial-youll-learn-how-to}
```
samajammin, line 16: Bummer - this script doesn't appear to be working correctly - the header id _should_ match the header text (i.e. this should be:
```
## (Soft) prerequisites {#soft-prerequisites}
```

samajammin:
> I think we might have to add some additional logic for the tutorial page. I believe current logic will always show the outdated banner? (we don't list every single tutorial on CrowdIn as we also have external tutorials)

FYI @minimalsm I don't think this is actually an issue. The tutorials homepage is a React component, so we don't compare header ids.

samajammin (changes requested):
This is looking super close!

Only issues I noticed were some markdown files (seemingly only the developer docs & developer tutorial pages) had wonky header ids (i.e. the header id didn't match the header text). The other markdown pages (e.g. defi & NFTs) seemed to be fine. I suspect because the script to generate the header ids may not work on deep directories?

src/pages/developers/tutorials.js
```diff
@@ -239,18 +239,21 @@ const TutorialsPage = ({ data, pageContext }) => {
     isExternal: true,
   }))
 
-  const allTutorials = []
-    .concat(externalTutorials, internalTutorials)
+  const allTutorials = [].concat(externalTutorials, internalTutorials)
+
+  const hasTutorialsCheck = some(allTutorials, ["lang", pageContext.language])
```
samajammin, line 244: Perhaps more clear as `langHasTutorials`?

samajammin (approved):

pettinarip (approved):
Looks great! @corwintines good job

pettinarip:
@corwintines building it in my local, I see a long list of errors which I don't think adds too much value:
image

should we silence this? or in case we keep it, should we add more details on the error message (e.g. file paths being read)?

samajammin:
Nice catch @pettinarip 

> should we silence this? or in case we keep it, should we add more details on the error message (e.g. file paths being read)?

I think the latter - I'm curious why we're seeing those errors.

corwintines:
I was originally just returning true for these, but it was brought up that we wanted to throw the errors so I console logged them. I believe at the end of the day it just doesnt have access to a file and throws the error from fs.

samajammin:
@corwintines & I dug in - we can remove the fs.readFileSync error.

PR #4452, opened 2021-11-12
Use consistent format for block numbers in history

fvictorio:
Sorry for deleting the template, but this is a pretty self-explanatory change.

### DIFF
--- a/src/content/history/index.md
+++ b/src/content/history/index.md
@@ -119,7 +119,7 @@ The [Beacon Chain](/eth2/beacon-chain/) needed 16384 deposits of 32 staked ETH t
 ### Staking deposit contract deployed {#staking-deposit-contract}
 
 <Emoji text=":calendar:" size={1} mr={"0.5rem"} mb={"0.5rem"} /> <code>Oct-14-2020 09:22:52 AM +UTC</code><br />
-<Emoji text=":bricks:" size={1} mr={"0.5rem"} mb={"0.5rem"} /> Block number: <a href="https://etherscan.io/block/11052984">11052984</a><br />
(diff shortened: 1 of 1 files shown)


fvictorio:
Actually, I _could_ have given a better explanation: the last entries use commas for the block number but the older ones don't. That's it.

corwintines (approved):
Definitely prefer this for readability, thanks @fvictorio!

samajammin:
Thanks @fvictorio! I agree we should be consistent on this.

Looking into this a bit I did notice that etherscan doesn't use any commas or other separators between block numbers (even though they use commas for other numbers), e.g. https://etherscan.io/block/13634817

I also notice in the [Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) they don't use commas:
Image 2021-11-17 at 1 24 19 PM

I do wonder why this is - I agree it hurts legibility a bit. On the other hand, it's easier to copy & paste.

I couldn't find any "official" way to reference block numbers (I suspect there is none) but I might suggest that we don't use commas to keep consistency with these sources. Open to opinions here.

fvictorio:
My two cents: I think that, for this particular context, using commas feel better. The downside is that copy-pasting them also means removing the commas, but there aren't that many entries, so I don't think it's a big deal. In etherscan, OTOH, it makes more sense, because I figure people copy-paste those numbers more often. With respect to the yellow paper... I think they should use commas too in that part.

samajammin (approved):

minimalsm:
Merging now 🎉

Be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And remember to [claim your POAP](https://ethereum.org/en/contributing/#poap) also 🥳!

minimalsm:
@all-contributors please add @fvictorio for docs

PR #4457, opened 2021-11-13
Fix duplication in whitepaper doc

bskrksyp9:
Fix duplication

## Description
A redundant word has been removed


solarpunklabs (commented):
The original is correct. While it can look unusual, especially to non-native English speakers, this is not a duplicate of the word 'is,' but two separate instances of the word. If you read it aloud you will also notice that removing the second 'is' renders the sentence grammatically incorrect. 
Here is some more information about this usage: https://ygdp.yale.edu/phenomena/double-is

bskrksyp9:
Alright thanks for the clarification !

Issue #4547, opened 2021-11-26
Headers covered by nav on ETH2 template when using header-ids in the url


minimalsm:
**Describe the bug**

When using a URL to directly access a header using a header-ID (i.e. `https://ethereum.org/en/eth2/beacon-chain/#beacon-chain-features`) the navigation bar covers the header.

**To Reproduce**

1. Go to 'https://ethereum.org/en/eth2/beacon-chain/'
2. Click on a header-ID chain link or a sidenav link
3. See error

**Expected behavior**

The nav should not cover the heading

**Screenshots**

https://i.gyazo.com/9ff035d5579b47de06a44f701644e8b8.gif

**Desktop (please complete the following information):**

- OS: macOS
- Browser: Brave Browser
- Version: v1.32

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

vluna:
@minimalsm since this is same issue as #4548, I can also update this one.

minimalsm:
Fixed in #4561

PR #4561, opened 2021-11-26
Update header anchors top margin on scroll [Fixes #4548 and #4547]

### LINKED ISSUE #4548: Headers covered by nav on usecases template when using header-ids in the url
**Describe the bug**

When using a URL to directly access a header using a header-ID (i.e. `https://ethereum.org/en/defi/#defi-vs-tradfi`) the navigation bar covers the header.

**To Reproduce**

1. Go to 'https://ethereum.org/en/eth2/beacon-chain/'
2. Click on a header-ID chain link or a sidenav link
3. See error

**Expected behavior**

The nav should not cover the heading

**Screenshots**

https://i.gyazo.com/9ff035d5579b47de06a44f701644e8b8.gif

- OS: macOS
- Browser: Brave Browser
- Version: v1.32

**Want to contribute?**

vluna:
## Description
- Update h2 and h3 style for eth2 and use-cases pages to add a top margin when scroll into them
- Update layout style to allow smooth scroll across the application when using anchors

## Related Issue
- #4548 
- #4547

### DIFF
--- a/src/styles/layout.css
+++ b/src/styles/layout.css
@@ -5,6 +5,7 @@ html {
   -ms-text-size-adjust: 100%;
   -webkit-text-size-adjust: 100%;
   font-size: 16px;
+  scroll-behavior: smooth;
 }
 body {
   margin: 0;
--- a/src/styles/layout.css
+++ b/src/styles/layout.css
@@ -5,7 +5,6 @@ html {
   -ms-text-size-adjust: 100%;
   -webkit-text-size-adjust: 100%;
   font-size: 16px;
-  scroll-behavior: smooth;
 }
 body {
   margin: 0;
--- a/src/templates/eth2.js
+++ b/src/templates/eth2.js
@@ -134,7 +134,15 @@ const H2 = styled.h2`
   font-size: 32px;
   font-weight: 700;
   margin-top: 4rem;
-  scroll-margin-top: 5rem;
+
+  /* Prevent nav overlap */
+  &:before {
+    content: "";
+    display: block;
+    height: 120px;
+    margin-top: -120px;
+    visibility: hidden;
+  }
 
   a {
     display: none;
@@ -170,7 +178,15 @@ const H2 = styled.h2`
 const H3 = styled.h3`
   font-size: 24px;
   font-weight: 700;
-  scroll-margin-top: 5rem;
+
+  /* Prevent nav overlap */
+  &:before {
+    content: "";
+    display: block;
+    height: 120px;
+    margin-top: -120px;
+    visibility: hidden;
+  }
 
   a {
     display: none;
--- a/src/templates/eth2.js
+++ b/src/templates/eth2.js
@@ -134,6 +134,7 @@ const H2 = styled.h2`
   font-size: 32px;
   font-weight: 700;
   margin-top: 4rem;
+  scroll-margin-top: 5rem;
 
   a {
     display: none;
@@ -169,6 +170,7 @@ const H2 = styled.h2`
 const H3 = styled.h3`
   font-size: 24px;
   font-weight: 700;
+  scroll-margin-top: 5rem;
 
   a {
     display: none;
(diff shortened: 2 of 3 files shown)


src/styles/layout.css
```diff
@@ -5,6 +5,7 @@ html {
   -ms-text-size-adjust: 100%;
   -webkit-text-size-adjust: 100%;
   font-size: 16px;
+  scroll-behavior: smooth;
```
lemiesz, line 8: Not supported in latest version of Safari. https://caniuse.com/?search=scroll-behavior 

Might be unimportant have you tested this, with safari though?

vluna: It does not work on Safari. If it is important to add, we can do the scroll behaviour with js.

lemiesz: Does anything in Safari look broken (I tried but was not able to repro issue). 

If not, then this should be fine. Safari support for `scroll-behavior` is currently in beta so should be supported soon enough.

vluna: Nothing was broken,  I just dded the `scroll-behaviour` to make it smoother when using anchors. This was more like an enhancement rather than an actual bug.

lemiesz: Yup got it. Good work! I'm relatively new to open source/github so I don't think I can ship your PR. However I have no further concerns.

minimalsm (changes requested):
Hey @vluna, thanks for taking a crack at this. This seems to be working on desktop resolutions but it does not work on tablet or mobile resolutions.

Tablet resolutions example:
https://i.gyazo.com/1ade55e2c8298dac7ec639578c1045eb.gif

Also, as noted by @lemiesz the support for `scroll-behavior` isn't fantastic yet so we probably want to avoid that. (Note all browsers on iOS run Safari under the hood and Apple requires an OS update to update Safari 😒 so even if they update it tomorrow anyone who doesn't update their OS would be unsupported forever).

We already handle the navigation overlap on the other templates [via the `Header2` and `Header3` components in `SharedStyleComponents`](https://github.com/ethereum/ethereum-org-website/blob/f03dc5d55469d30c86eb1f877226490e647a1c92/src/components/SharedStyledComponents.js#L289). I think to get this to work as expected we should take the same approach of using a `:before` pseudo-element, and simply add this to the H2 and H3 in the `eth2` and `use-cases` templates.

```javascript
  &:before {
    content: "";
    display: block;
    height: 120px;
    margin-top: -120px;
    visibility: hidden;
  }
```

vluna:
> Hey @vluna, thanks for taking a crack at this. This seems to be working on desktop resolutions but it does not work on tablet or mobile resolutions.
> 
> Tablet resolutions example: https://i.gyazo.com/1ade55e2c8298dac7ec639578c1045eb.gif
> 
> Also, as noted by @lemiesz the support for `scroll-behavior` isn't fantastic yet so we probably want to avoid that. (Note all browsers on iOS run Safari under the hood and Apple requires an OS update to update Safari 😒 so even if they update it tomorrow anyone who doesn't update their OS would be unsupported forever).
> 
> We already handle the navigation overlap on the other templates [via the `Header2` and `Header3` components in `SharedStyleComponents`](https://github.com/ethereum/ethereum-org-website/blob/f03dc5d55469d30c86eb1f877226490e647a1c92/src/components/SharedStyledComponents.js#L289). I think to get this to work as expected we should take the same approach of using a `:before` pseudo-element, and simply add this to the H2 and H3 in the `eth2` and `use-cases` templates.
> 
> ```js
>   &:before {
>     content: "";
>     display: block;
>     height: 120px;
>     margin-top: -120px;
>     visibility: hidden;
>   }
> ```

That makes sense! I have updated the PR with the code changes. 
- Change `scroll-margin-top` to `&:before` content as mentioned above
- Remove `scroll-behaviour` for the layout

minimalsm (approved):
@vluna LGTM! 

Thanks vluna! 🚀🚀🚀🚀🚀

minimalsm:
@all-contributors add @vluna for code, bugs

PR #4599, opened 2021-12-01
Adjust codeHash explanation

mariahpickett:
Was a little confused about the purpose and value of codeHash in contract accounts versus externally-owned account after reading this for the first time, so i took a peek at the yellowpaper. the yellowpaper description was almost identical to these docs, but as someone totally unfamiliar with these concepts, the minor grammatical differences in the yellowpaper actually clarified a few things for me.

## Description

Minor grammatical changes

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/3908

### DIFF
--- a/src/content/developers/docs/accounts/index.md
+++ b/src/content/developers/docs/accounts/index.md
@@ -41,9 +41,9 @@ Both account types have the ability to:
 
 Ethereum accounts have four fields:
 
-- `nonce` – a counter that indicates the number of transactions sent from the account. This ensures transactions are only processed once. In a contract account, this number represents the number of contracts created by the account
-- `balance` – the number of wei owned by this address. Wei is a denomination of ETH and there are 1e+18 wei per ETH.
-- `codeHash` – this hash refers to the _code_ of an account on the Ethereum virtual machine (EVM). Contract accounts have code fragments programmed in that can perform different operations. This EVM code gets executed if the account gets a message call. It cannot be changed unlike the other account fields. All such code fragments are contained in the state database under their corresponding hashes for later retrieval. This hash value is known as a codeHash. For externally owned accounts, the codeHash field is the hash of an empty string.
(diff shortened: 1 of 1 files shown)


src/content/developers/docs/accounts/index.md
```diff
@@ -41,9 +41,9 @@ Both account types have the ability to:
...
+- `nonce` – A counter that indicates the number of transactions sent from the account. This ensures transactions are only processed once. In a contract account, this number represents the number of contracts created by the account
+- `balance` – The number of wei owned by this address. Wei is a denomination of ETH and there are 1e+18 wei per ETH.
+- `codeHash` – For externally-owned accounts, this field is the hash of an empty string. For contract accounts, this field is the hash of the Ethereum virtual machine (EVM) _code_ that gets executed if the account receives a message call. These code fragments can perform various operations, and they are contained in the state database under their corresponding hashes for later retrieval. Unlike all other account fields, the codeHash is immutable, meaning it cannot be modified after its creation.
```
minimalsm, line 46:
```suggestion
- `nonce` – A counter that indicates the number of transactions sent from the account. This ensures transactions are only processed once. In a contract account, this number represents the number of contracts created by the account
- `balance` – The number of wei owned by this address. Wei is a denomination of ETH and there are 1e+18 wei per ETH.
- `codeHash` – This hash refers to the _code_ of an account on the Ethereum virtual machine (EVM). Contract accounts have code fragments programmed in that can perform different operations. This EVM code gets executed if the account gets a message call. It cannot be changed, unlike the other account fields. All such code fragments are contained in the state database under their corresponding hashes for later retrieval. This hash value is known as a codeHash. For externally owned accounts, the codeHash field is the hash of an empty string.
```

minimalsm: Thanks for the suggestions on this. I've checked this over a few times and checked with others and we think this re-wording detracts from the readability a bit. I've suggested changes to the casing at the start of the sentence because this was incorrect.

minimalsm (approved):

wackerow (approved):

PR #4600, opened 2021-12-01
Update POS content

Xia-Sam:
## Description
New portion added to pos content: How POS addressses mining power


minimalsm:
Hey @Xia-Sam thanks for the PR on this. The points being made in this new paragraph are already made elsewhere in the doc. I'm not certain we need to add a whole new section on this.

Especially as this doc is focused on what proof-of-stake is, and not what proof-of-stake is in comparison to proof-of-work.

PR #4604, opened 2021-12-01
Refactor isMobile to separate function

### LINKED ISSUE #4297: Refactor isMobile checks into one utility function
There are a few checks for isMobile in the application which implement the same functionality in each of them.

Create a single isMobile utility function to use throughout the application.

**Additional context**
These are getting added in https://github.com/ethereum/ethereum-org-website/pull/4131 and https://github.com/ethereum/ethereum-org-website/pull/4205. After these are merged, the refactor will be best.

**Want to contribute?**

ShubhankarKG:
## Description

This PR extracts various `isMobile` implementations into a consistent function, that is then reused.

## Related Issue

Fixes #4297


minimalsm:
Hey @ShubhankarKG, thanks for the PR on this.

With these changes, the site doesn't appear to be building properly. Were you able to get this to run locally?

ShubhankarKG:
I unfortunately did not check it locally. Let me spin this up tomorrow and I'll get back as soon as I can.

PS: I can not see the logs from the build output. Is it because the build log has a lot of lines ?

ShubhankarKG:
So I ran a local build and found out that this is the issue:

image

So what we can do is, if window is not defined then fallback to width, else opt for the current one. Would like to know your thoughts on this @minimalsm

ShubhankarKG:
I was able to locally build this, I'm unsure what the problem is. Any help here please ?

minimalsm:
Thanks @ShubhankarKG, this appears to be working now 💪💪 Haven't had the chance to extensively test this yet but will get around to it at the beginning of the week.

minimalsm (approved):

src/utils/isMobile.js
```diff
@@ -0,0 +1,9 @@
+export const isMobile = () => {
+  let isMobile = false
+  if (typeof window !== "undefined") {
+    isMobile = /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
+      window.navigator.userAgent
+    )
+  }
```
wackerow, line 7: He @ShubhankarKG, thanks for the submission! Looks good, was just wondering if you should consider adding a check for small screen sizes as well in case this missed anything. 

Maybe something like this:
```suggestion
  if (![typeof window, typeof document].includes("undefined")) {
    isMobile = /Android|webOS|iPhone|iPad|iPod|BlackBerry|IEMobile|Opera Mini/i.test(
      window.navigator.userAgent
    ) | document.documentElement.clientWidth < 1024
  }
```
@pettinarip @corwintines Any thoughts here? Otherwise this looks great =), and would be willing to merge as is.

pettinarip: IMO I'd use a well-known lib like https://github.com/faisalman/ua-parser-js for this (e.g. is what `react-detect-device` is using behind the scenes). Why? mainly because we don't need to worry about the maintainability of this function that can grow in complexity over time.

If we decide to go that way, this can be as simple as:
```ts
function isMoblie() {
	return ua.device.type === 'mobile'
}
```

If we keep our custom function, I would argue against checking the size of the screen here. What we really want to check here is if the device is a mobile device regardless of the size of the screen in order to turn on/off some functionality that those devices do/do not support. On the other hand, media queries should be there to control screen sizes and layout design.

src/components/Tooltip.js
```diff
@@ -54,24 +55,17 @@ const ModalReturn = styled.div`
...
         <ModalReturn onClick={() => setIsVisible(false)} />
       )}
       <Container
         title="More info"
-        onMouseEnter={!isMobile ? () => setIsVisible(true) : null}
-        onMouseLeave={!isMobile ? () => setIsVisible(false) : null}
-        onClick={isMobile ? () => setIsVisible(!isVisible) : null}
+        onMouseEnter={!isMobile() ? () => setIsVisible(true) : null}
+        onMouseLeave={!isMobile() ? () => setIsVisible(false) : null}
+        onClick={isMobile() ? () => setIsVisible(!isVisible) : null}
```
pettinarip, line 68: Let's call `isMobile` once and use the result in the rest.

```ts
import * as utils from "../utils/isMobile"

const Tooltip = ({ content, children }) => {
  const isMobile = utils.isMobile()

  return (
    <>
      {isVisible && isMobile && (
        <ModalReturn onClick={() => setIsVisible(false)} />
      )}
      <Container
        title="More info"
        onMouseEnter={!isMobile ? () => setIsVisible(true) : null}
        onMouseLeave={!isMobile ? () => setIsVisible(false) : null}
        onClick={isMobile ? () => setIsVisible(!isVisible) : null}
```

pettinarip (commented):
Thanks for the PR @ShubhankarKG! looking great. I left a small refactor there. GJ

minimalsm:
@ShubhankarKG thanks for all the work on this ❤️❤️❤️ Bringing this in 🚀🚀🚀

@pettinarip I've put up a PR (#4771) implementing your suggested refactor.

PR #4650, opened 2021-12-05
fix: always convert nodes stats to Number

aghArdeshir:
## Description
When switching between 30 Days and 90 Days with specific data, the chart becomes unreadable.

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/4651

### DIFF
--- a/src/components/StatsBoxGrid.js
+++ b/src/components/StatsBoxGrid.js
@@ -348,7 +348,7 @@ const StatsBoxGrid = () => {
         const data = result
           .map(({ UTCDate, TotalNodeCount }) => ({
             timestamp: new Date(UTCDate).getTime(),
-            value: TotalNodeCount,
+            value: Number(TotalNodeCount),
           }))
           .sort((a, b) => a.timestamp - b.timestamp)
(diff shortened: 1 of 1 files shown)


minimalsm (approved):
Thanks for the PR @aghArdeshir!

I've tested the deploy preview and this definitely looks like it solves the problem 💪

For my own curiosity: I don't really understand why casting `TotalNodeCount` as a number fixes this. Could you explain?

aghArdeshir:
Thanks @minimalsm for your comment. Actually to be honest, I don't know either. First, I thought the problem is with `recharts`. I tried the mock data on a pure `react-recharts` codesandbox: https://codesandbox.io/s/elegant-varahamihira-q1jou?file=/src/App.js:6050-6118

But noted that just giving numbers to recharts seems to make the chart readable. And I couldn't convince myself that this is a bug of recharts. So I just mapped our strings to numbers.

corwintines (approved):
Thanks @aghArdeshir!

corwintines:
@all-contributors can you add @aghArdeshir for code

PR #4652, opened 2021-12-05
Update number of translated words 2021

lukassim:
## Description

Update the number of words the community has translated in 2021 in the stats section.


samajammin (approved):
🚀

PR #4667, opened 2021-12-06
Fix stacking elements in container [Fixes #4598]

### LINKED ISSUE #4598: Stablecoin page layout bug
**Describe the bug**
Horizontal scrolling is occurring at medium screen widths on the stablecoins page. Also noting a bug to the layout of CTA buttons in one of the cards on the same page at similar screen sizes.

**To Reproduce**

1. Go to https://ethereum.org/en/stablecoins/ on desktop
2. Scroll down to "Find a stablecoin" section
3. Reduce window width to around 1025px to 1125px
4. Notice cards start to extend off right side and horizontal scroll bar comes in
5. Also, scroll to "Use your stablecoins" callout, and notice two buttons are stacked and staggered without vertical padding.

**Expected behavior**
Cards should rearrange if too big for the window size. Buttons should rearrange to stacking vertically, with some padding between, and full width once this occurs.

**Screenshots**
image
image

- OS: macOS Big Sur
- Browser: Brave latest

**Want to contribute?**

vluna:
Issue with buttons not stacking properly in `/stablecoins`, `Use you stablecoins` container.

## Description

- Remove `ButtonLinkRight` since it only adds `margin-right` to the button
- Add  `ButtonLinkContainer` to be a container around buttons to display flex in rows
- Update second button to be `ButtonLink`

## Related Issue

#4598

### DIFF
--- a/src/data/eth2-articles.js
+++ b/src/data/eth2-articles.js
@@ -25,25 +25,21 @@ export const benArticles = [
   {
     title: "What’s New in Eth2 – #82",
     description: "19 November 2021",
-    link:
-      "https://hackmd.io/@benjaminion/eth2_news/https%3A%2F%2Fhackmd.io%2F%40benjaminion%2Fwnie2_211119",
+    link: "https://hackmd.io/@benjaminion/eth2_news/https%3A%2F%2Fhackmd.io%2F%40benjaminion%2Fwnie2_211119",
   },
   {
     title: "What’s New in Eth2 – #81",
     description: "5 November 2021",
-    link:
-      "https://hackmd.io/@benjaminion/eth2_news/https%3A%2F%2Fhackmd.io%2F%40benjaminion%2Fwnie2_211105",
+    link: "https://hackmd.io/@benjaminion/eth2_news/https%3A%2F%2Fhackmd.io%2F%40benjaminion%2Fwnie2_211105",
   },
   {
     title: "What’s New in Eth2 – #80",
     description: "22 October 2021",
-    link:
-      "https://hackmd.io/@benjaminion/eth2_news/https%3A%2F%2Fhackmd.io%2F%40benjaminion%2Fwnie2_211022",
+    link: "https://hackmd.io/@benjaminion/eth2_news/https%3A%2F%2Fhackmd.io%2F%40benjaminion%2Fwnie2_211022",
   },
   {
     title: "What’s New in Eth2 – #79",
     description: "8 October 2021",
-    link:
-      "https://hackmd.io/@benjaminion/eth2_news/https%3A%2F%2Fhackmd.io%2F%40benjaminion%2Fwnie2_211008",
+    link: "https://hackmd.io/@benjaminion/eth2_news/https%3A%2F%2Fhackmd.io%2F%40benjaminion%2Fwnie2_211008",
   },
 ]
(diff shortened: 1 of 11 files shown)


corwintines (changes requested):
Hey @vluna, this is looking awesome! I think just need to get the responsiveness down for phone size and we are good to go. Noticing the button bleed over the container.
Screen Shot 2021-12-07 at 4 42 16 PM

vluna:
@corwintines I have fixed the issue by adding `white-space: break-spaces;` to an inherit `ButtonLink`. However, I had to merge the code with current dev I believe and the tests failed. Not really sure about this, so please advise if this is expected or if I should create a new branch and PR for this issue?

src/data/eth2-articles.js
```diff
@@ -25,21 +25,25 @@ export const benArticles = [
   {
     title: "What’s New in Eth2 – #82",
     description: "19 November 2021",
-    link: "https://hackmd.io/@benjaminion/eth2_news/https%3A%2F%2Fhackmd.io%2F%40benjaminion%2Fwnie2_211119",
```
minimalsm, line 28: @vluna Not sure why but your linter has went a bit crazy here. Any chance we could remove the files that aren't concerned with fixing the stacking elements from this PR?

minimalsm (changes requested):

src/pages/stablecoins.js
```diff
@@ -249,8 +249,15 @@ const InfoIcon = styled(Icon)`
   fill: ${(props) => props.theme.colors.text};
 `
 
-const ButtonLinkRight = styled(ButtonLink)`
-  margin-left: 1rem;
+const ButtonLinkWrap = styled(ButtonLink)`
+  white-space: break-spaces;
```
corwintines, line 253:
```suggestion
    width: 100%;
    text-align: left;
    white-space: inherit;
```
This definitely fixes the problem, but I think these suggested styles look a bit better in my opinion. That way we aren't mixing text alignments. Let me know what you think @vluna

corwintines (changes requested):

src/pages/stablecoins.js
vluna: I am not sure this will work for desktop since the button might use 100% width. We could do `white-space: inherit` only, it will produce the same result. However, I was thinking that maybe we should change the `white-space: nowrap` to  `inherit` in the ButtonLink component as this might happen to all the buttons. With `inherit`, it will just break to the next line by default instead of going all the way. 

I believe the `white-space` change should be sufficient when this issue comes up. What do you think @corwintines?

src/data/eth2-articles.js
vluna: @minimalsm I fixed most of the files, just a couple not sure why I could not revert them back. Those couple of files are just white spaces not sure if that would be a problem? If it is, I will just create a new PR with the `/stablecoins` change.

vluna:
Just an update since I am not sure whats going on. I created a new branch and a new PR but all tests failed. Not sure why.

minimalsm:
Hey @vluna thanks for letting us know. We're having some internal issues with our hosting service (the issues aren't on your end causing the tests to fail aren't on your end). Should we close this PR in favour of #4696?

vluna:
As per conversation with @minimalsm, closing this PR in favour of #4696

PR #4671, opened 2021-12-06
Remove en paths from non-English pages

minimalsm:
## Description
- Removes /en/ paths from foreign language pages

## Related Issue
None filed


corwintines (approved):
I think this is good after the commit I pushed, let me know what you think @minimalsm

minimalsm:
Thanks @corwintines :-) nice catch. I had only checked the /translations/ file.

wackerow (approved):

PR #4678, opened 2021-12-07
Fix Translation Acknowledgement bugs

corwintines:
## Description
- Display only one row of contributor names in mobile view for /contributing/translation-program/contributors page. Was displaying two, but that ran too wide for the page and had responsive issues.
- Show all languages for translators in TranslationLeaderboard. May need to talk to Crowdin on how to address these translators having multiple languages like this though.

Preview: https://deploy-preview-4678--ethereumorg.netlify.app/en/contributing/translation-program/contributors/


minimalsm (approved):

samajammin:
Looks good!

Only thought is for situations where someone works on a bunch of languages:
Image 2021-12-08 at 11 42 31 AM

Should we just display the 1st 3 languages, or something along those lines?

wackerow (approved):
Lookin good =)

@samajammin Agree, would probably be reasonable to draw the line somewhere here. To make this more friendly on mobile, think 3 is a reasonable place to start. @lukassim Curious your take on this. 

image

This is limiting to three and I think looks a lot cleaner. Will patch with this and pull in, can adjust easily if needed.

PR #4682, opened 2021-12-07
Implement true randomness on consensus client list

minimalsm:
## Description
- In the spirit of client diversity I refactored the [naive random short](https://medium.com/@oldwestaction/randomness-is-hard-e085decbcbb2) we were using to a lodash shuffle instead.
- Also separated the declaration of the client data from the sorting logic

## Related Issue
None


corwintines:
I like the change when it comes to sorting. I just feel like the whole useState and useEffect is a bit overkill for this if we just sort a list one time. Could we remove these and simplify this component? Curious what you think @minimalism

minimalsm:
@corwintines sounds good but i'm not 100% clear on the tradeoffs here. maybe worth asking someone else to weigh in? 

Only downside I see of doing it this way is that the cards will be [reordered after the page renders](https://i.gyazo.com/f7fecada165bee914ead4dde91f5e730.gif) but the current implementation has layout shift anyway so both solutions are a bit jarring 🤷‍♂️.

corwintines (approved):
Actually after looking further, useEffect would be a proper lifecycle method for this. LGTM!

corwintines:
Just have some merge conflicts, but otherwise good for me.

corwintines:
@minimalsm just have some merge conflicts, but otherwise good for me.

wackerow:
🎉  Nice job!

PR #4695, opened 2021-12-08
Redirect debugging - Gatsby Cloud focus

wackerow:
Patch to recent Gatsby v4 / Gatsby Cloud migration for the redirect bugs encountered. #4693 attempting to solve for the same, but focused on a fix for when the site is deployed to and hosted by Netlify. This branch takes a different approach to focus on using Gatsby recommendations.

- `gatsby-plugin-gatsby-cloud` and `gatsby-plugin-netlify` are both installed in this branch
- Removes `netlify.toml` redirects, leaving only the `createRedirects` method to generate these redirects, pulling from a new `redirects.json` data file in repo root.


redirects.json
```diff
@@ -0,0 +1,293 @@
...
+    "fromPath": "/en/developers/docs/mining/",
+    "toPath": "/en/developers/docs/consensus-mechanisms/pow/mining/"
+  },
+  {
+    "fromPath": "/en/eth2/the-beacon-chain/",
+    "toPath": "/en/eth2/beacon-chain/"
+  },
+  {
+    "fromPath": "/*/beginners",
+    "toPath": "/:splat/what-is-ethereum/"
```
samajammin, line 263: These are incorrect - I think we want `"toPath": "/*/what-is-ethereum/"` - see the "splat redirects" section: https://support.gatsbyjs.com/hc/en-us/articles/1500003051241-Working-with-Redirects

PR #4723, opened 2021-12-10
Arabic: Homepage updated

wackerow:
## Description
- Updates the Arabic (ar) homepage to latest translations from Crowdin.
- Updates logic about when to show the legacy homepage. This is a temporary measure and can be phased out after all languages have been updated to the newer homepage (prioritized in Crowdin)


minimalsm:
LGTM.

Could definitely do a clean-up to support rtl better. But seems like it could get messy real quick... (see nav screenshot)

### Nav

Screenshot 2021-12-13 at 14 37 51

### Mobile nav

This should be flipped

Screenshot 2021-12-13 at 14 37 19

Screenshot 2021-12-13 at 14 28 13

### Footer

Should be rtl

Screenshot 2021-12-13 at 14 40 49

PR #4747, opened 2021-12-11
Malay: New language, homepage

wackerow:
## Description
- Adds Malay (ms) to list of languages
- Import homepage strings from Crowdin
- Utilizes updated legacy homepage logic from #4723

- [x] Check banner logic


minimalsm:
@wackerow where are we at with this 😀?

gatsby-node.js
```diff
@@ -334,6 +334,9 @@ exports.createPages = async ({ graphql, actions, reporter }) => {
       const markdownPath = `${__dirname}/src/content/translations/${lang}/${page}/index.md`
       const langHasOutdatedMarkdown = fs.existsSync(markdownPath)
       if (!langHasOutdatedMarkdown) {
+        // Check if json strings exists for language, if not mark `isContentEnglish` as true
+        const jsonPath = `${__dirname}/src/intl/${lang}/page-${page}.json`
```
samajammin, line 338: Could we just use the existing helper function `checkIsPageOutdated` for this? cc @corwintines

samajammin: I'm also not sure if this check is needed at all, given that we perform `checkIsPageOutdated` within `onCreatePage`, although I'm not _certain_ that is triggered when we call `createPage` below this.

wackerow: I don't believe it does trigger it from that method itself. `exports.onCreatePage` calls `createPage()` itself. Placing this here fixes the "pages-conditional" that were having the issues.

I updated to use `checkIsPageOutdated`, and merged in dev / cleared conflicts.

corwintines (approved):
LGTM! Thanks for catching the page-conditional thing @wackerow!

PR #4749, opened 2021-12-11
Farsi: Homepage updated

wackerow:
## Description
- Updates the Farsi (fa) homepage to latest translations from Crowdin.
- Utilizes updated legacy homepage logic from #4723, and updates `fa` to use newer page

- [x] @lukassim Looks like Crowdin uses the term "Persian" instead of "Farsi"... should we update our site to use the same?


lukassim:
> * [ ]  @lukassim Looks like Crowdin uses the term "Persian" instead of "Farsi"... should we update our site to use the same?

Crowdin uses ''Persian'', while the translation agencies use ''Farsi''. Farsi is the Persian word for the Persian language, but refers specifically to Iran's standard Persian. There are also other variations of Persian, namely Dari (Afghanistan), Tajik (Tajikistan), etc.
Crowdin differentiates between them (we also have Dari in the project), so we should probably do the same.
In case we get the homepage translated into Dari, this would be a different language, despite the fact that they are both variations of Persian.

PR #4758, opened 2021-12-13
Support new report type from Crowdin for TranslationLeaderboard

corwintines:
## Description
Support using `Translation Cost` report from Crowdin instead of `Top Members` so that we can get more granular data around what languages users have translated


wackerow (approved):
Just suggested a couple stylistic adjustments, but either way approving. Having trouble with Gatsby builds at the moment, but locally this looks good. 

Noticed this page isn't loading images properly, will address in separate PR.

minimalsm:
[In the current deploy preview](https://build-30987539-bd63-4c21-9fea-e3c9cb41b18e.gtsb.io/en/contributing/translation-program/acknowledgements/) none of the buttons on the translator leaderboard are working for me. Unsure if this is related to the gatsby issues we are currently experiencing.

PR #4764, opened 2021-12-14
Fix client images to use `gatsby-plugin-image` format

wackerow:
## Description
- Fixed broken images as a result of `gatsby-plugin-image` migration

## Related Issue
None filed, but was causing Ethereum client logos to not display on `get-involved` page


pettinarip (approved):
💪🏼

minimalsm (approved):

PR #4768, opened 2021-12-14
Refactor few components to accept intl strings for attributes

wackerow:
## Description
- Extracts english text from React component attributes, placing them in an intl `json` file instead
- Allow strings to be translated in Crowdin

Files affected:
- `Callout.js`
- `TitleCardList.js`
- `UpgradeStatus.js`


src/components/Callout.js
```diff
@@ -59,8 +59,8 @@ const Callout = ({
     <Content>
       <div>
         {emoji && <Emoji text={emoji} size={3} />}
-        <h3>{title}</h3>
-        <Description>{description}</Description>
+        <h3>{titleKey}</h3>
```
samajammin, line 62: Shouldn't this be wrapped in a `Translation` component?

src/components/Callout.js
```diff
@@ -59,8 +59,8 @@ const Callout = ({
...
+        <Description>{descriptionKey}</Description>
```
samajammin, line 63: Same here?

corwintines (approved):

corwintines:
LGTM, nice @wackerow !

PR #4781, opened 2021-12-14
[testing] Revert and adjust `relativePath` parsing logic

wackerow:
Builds recently stopped working after `relativePath` logic was switched to use `lastIndexOf("src/")`.

Reverts to commit before builds stopped working, and adjusted with new logic for `relativePath`


wackerow:
PR was for debugging. Fixed by #4788

PR #4788, opened 2021-12-16
Remove `gatsby-plugin-lodash`

wackerow:
## Description
Recent builds have been presenting with a large variety of bugs. The errors received followed the theme of `TypeError: ___ is not a function`. A recent bug presented similarly where nav menus were broken, and was fixed by removing the use of `lodash` in that PR. Noted that we have both `lodash` and `gatsby-plugin-lodash` installed. 
Attempting to switch usage of `lodash` library in components, to using the gatby plugin resulted in a broken build.

Removing the gatsby plugin in lieu of the standard `lodash` library seems to fix the array of bugs we've been seeing.

- Removes `gatsby-plugin-lodash` package

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -59,28 +59,6 @@ module.exports = {
         // dev: true,
       },
     },
-    {
-      resolve: `gatsby-plugin-lodash`,
-      options: {
-        disabledFeatures: [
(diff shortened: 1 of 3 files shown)


corwintines (commented):
Looks good

PR #4625, opened 2021-12-02
Fix string fragment issues

corwintines:
Fix string fragments outlined in notion card:

- [https://github.com/ethereum/ethereum-org-website/blob/b4121564662828caa940302bb5591e2968faa0bf/src/intl/en/page-wallets-find-wallet.json#L132](https://github.com/ethereum/ethereum-org-website/blob/b4121564662828caa940302bb5591e2968faa0bf/src/intl/en/page-wallets-find-wallet.json#L132)
- [https://github.com/ethereum/ethereum-org-website/blob/4e8f8826890e949212db94d94fdec1186f36c6c0/src/intl/en/page-eth2-vision.json#L45](https://github.com/ethereum/ethereum-org-website/blob/4e8f8826890e949212db94d94fdec1186f36c6c0/src/intl/en/page-eth2-vision.json#L45)
- [https://github.com/ethereum/ethereum-org-website/blob/c1e1a26f118b8541d674320ac038761a0003bfb8/src/intl/en/page-eth2-get-involved-bug-bounty.json#L42](https://github.com/ethereum/ethereum-org-website/blob/c1e1a26f118b8541d674320ac038761a0003bfb8/src/intl/en/page-eth2-get-involved-bug-bounty.json#L42)

PR #4802, opened 2021-12-16
Remove HTML from json files

minimalsm:
## Description
Removes unnecessary HTML strong tags from JSON files

## Related Issue
#4625


wackerow (approved):

samajammin:
@minimalsm @wackerow if we're not going to use HTML in these translation strings, should we revert the Translation component? https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Translation.js#L2

minimalsm:
I've not removed every instance of this as per your comments yesterday about trying to support it

That PR just removed it from the tutorial/wallet instances. However, we still use it on the eth2 bug-bounty page.

PR #4818, opened 2021-12-17
Add YouTube component

wackerow:
## Description
We use the same structure/formatting for all of our YouTube embeds. Extracted this into a `YouTube.js` component to remove redundancy. Updated codebase to replace all instances of YouTube iframes.

We recently reformatted many single-line React components being used in markdown to be multiline, to force it to display as a `block` element. Given this is a single-line React component, added formatting to use `display: block;` with a slight vertical margin to force it to always take a new line (even if Crowdin exports adjust this within the markdown to be at the end of the previous sentence).


src/content/defi/index.md
```diff
@@ -24,9 +24,7 @@ DeFi is a collective term for financial products and services that are accessibl
 
 There's a booming crypto economy out there, where you can lend, borrow, long/short, earn interest, and more. Crypto-savvy Argentinians have used DeFi to escape crippling inflation. Companies have started streaming their employees their wages in real time. Some folks have even taken out and paid off loans worth millions of dollars without the need for any personal identification.
 
-<figure>
-  <iframe width="100%" height="315px" src="https://www.youtube.com/embed/H-O3r2YMWJ4" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
-</figure>
+<YouTube id="H-O3r2YMWJ4" />
```
minimalsm, line 27: I like the idea here but I think only passing the `id` here adds complexity. 

- Without context, contributors might have to go to the YouTube component or read the style guide to understand how the props here work—a regular URL seems more intuitive.  
- It also seems a little finicky and prone to error to me (i.e. it is a lot easier to copy an entire URL versus only part of it.)

```suggestion
<YouTube src="https://www.youtube.com/embed/H-O3r2YMWJ4" />
```

minimalsm (commented):
@wackerow thanks for knocking this out 😀 Left a comment around the trade-off of passing only an `id` versus the entire src URL, curious to hear your thoughts here.

src/content/defi/index.md
samajammin: Agreed - @wackerow could you please update to pass the entire URL src sring?

src/components/YouTube.js
```diff
@@ -0,0 +1,33 @@
+import React from "react"
+import styled from "styled-components"
+
+const Figure = styled.figure`
+  display: block;
```
minimalsm, line 5: don't need this, it's already a block element
```suggestion
```

src/content/defi/index.md
corwintines: Same. I think once that change is in, then this is good 👍

wackerow: Main issue with this is that we use an "embed" URL:

If user grabs link from the URL bar:
image

If user grabs from "Share" modal:
image

And if user selects "Embed":
image

In either case, none of these are exactly what we want if we're asking users to pass the entire "embed" URL into this component.

If we really don't like the idea of just using the video ID, then perhaps we should expand the logic of this component to accept any link and parse out the ID ourselves within the component. Otherwise I don't know that this will make anything any easier.

corwintines:
I think with @wackerow explanation for why he just used the id makes sense to me for now. Maybe we can refactor this to use src after and use regex to make this easier, but for inclined to merge.

src/content/defi/index.md
samajammin: Ah, that's a fair point!

samajammin (approved):
✨ 📺 ✨

corwintines (approved):

Issue #4821, opened 2021-12-18
Update Ethereum docs to make definition of Wallet and Account clearer


kennethcassel:
The site has a conflicting definition of Wallets on two pages. 

On https://ethereum.org/en/wallets, the definition is: 
> Ethereum wallets are applications that let you interact with your Ethereum account. 

On https://ethereum.org/en/developers/docs/accounts/#a-note-on-wallets the definition is:
> A wallet is the keypair associated with a user-owned account, which allows a user to make transactions from or manage the account.

I brought this up in a [twitter thread](https://twitter.com/KennethCassel/status/1471854438126137352?s=20) and @samajammin agreed with the following suggested terminology from @wslyvh 

Account : key-pair 
Wallet: application that can interact on behalf of it.

We should update the docs at https://ethereum.org/en/developers/docs/accounts/#a-note-on-wallets to reflect this change in terminology.

kennethcassel:
I'd like to be assigned to this issue :)

minimalsm:
Closed in #4822

PR #4822, opened 2021-12-18
Update definition of wallet in developer docs for accounts

kennethcassel:
## Description
This PR updates the definition of a wallet to more closely match terminology commonly used. The conflicting definitions of wallets on the Ethereum website are confusing for new users to Ethereum. This change creates consistency between the pages https://ethereum.org/en/wallets and https://ethereum.org/en/developers/docs/accounts/#a-note-on-wallets for wallet definitions.

Updates the definition at https://ethereum.org/en/developers/docs/accounts/#a-note-on-wallets from

> An account is not a wallet. A wallet is the keypair associated with a user-owned account, which allows a user to make transactions from or manage the account.

to

> An account is not a wallet. An account is the keypair for a user-owned Ethereum account. A wallet is an application that lets you interact with your Ethereum account.

## Related Issue
#4821

### DIFF
--- a/src/content/developers/docs/accounts/index.md
+++ b/src/content/developers/docs/accounts/index.md
@@ -126,7 +126,7 @@ For dapp development, you'll want access to dummy accounts with test ETH so you 
 
 ## A note on wallets {#a-note-on-wallets}
 
-An account is not a wallet. An account is the keypair for a user-owned Ethereum account. A wallet is an application that lets you interact with your Ethereum account.
+An account is not a wallet. An account is the keypair for a user-owned Ethereum account. A wallet is an interface or application that lets you interact with your Ethereum account.
 
 ## A visual demo {#a-visual-demo}
 
--- a/src/content/developers/docs/accounts/index.md
+++ b/src/content/developers/docs/accounts/index.md
@@ -126,7 +126,7 @@ For dapp development, you'll want access to dummy accounts with test ETH so you 
 
 ## A note on wallets {#a-note-on-wallets}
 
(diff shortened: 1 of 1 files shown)


samajammin:
Thanks for the PR! Agree this is important to be consistent about.

One thing worth considering which I just thought about was how libraries & tools define wallets.

e.g. ethers.js https://docs.ethers.io/v4/api-wallet.html
From that perspective a wallet isn't strictly a software application. 

I think we should do a quick bit of research on this. Let us get back to you on this. I'd encourage input from anyone here!

kennethcassel:
Yeah it seems like from the ethers.js library, a wallet is also an API to manage your key-pair. 

So now the definition could be application or API. I'll do some searching as well for some other examples!

src/content/developers/docs/accounts/index.md
```diff
@@ -126,7 +126,7 @@ For dapp development, you'll want access to dummy accounts with test ETH so you
 
 ## A note on wallets {#a-note-on-wallets}
 
-An account is not a wallet. A wallet is the keypair associated with a user-owned account, which allows a user to make transactions from or manage the account.
+An account is not a wallet. An account is the keypair for a user-owned Ethereum account. A wallet is an application that lets you interact with your Ethereum account.
```
minimalsm, line 129: What about

```suggestion
An account is not a wallet. An account is the keypair for a user-owned Ethereum account. A wallet is an interface or application that lets you interact with your Ethereum account.
```

samajammin (approved):

samajammin:
Thanks again @kennethcassel!

samajammin:
@all-contributors please add @kennethcassel for docs

PR #4850, opened 2021-12-20
Add Vitalik's governance articles to Further Reading

skylarweaver:
Add Vitalik's governance articles as recommended reading to the governance page


minimalsm (approved):
Thanks @skylarweaver ❤️

PR #4879, opened 2021-12-22
Update index.md

mohan-chinnappan-n:
blockchain diagram added. source: wikipedia

### DIFF
--- a/src/content/developers/docs/intro-to-ethereum/index.md
+++ b/src/content/developers/docs/intro-to-ethereum/index.md
@@ -11,6 +11,8 @@ A blockchain is a public database that is updated and shared across many compute
 
 "Block" refers to data and state being stored in consecutive groups known as "blocks". If you send ETH to someone else, the transaction data needs to be added to a block to be successful.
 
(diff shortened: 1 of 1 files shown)


corwintines (commented):
I don't think this image fits where it currently is. There is a lot of extra data in that image that hasnt been introduced yet, and seems more confusing for the reader in this current spot.

minimalsm:
I agree with @corwintines on this one. I don't think we want to add a scary-looking blockchain diagram as the first thing users see when they land on the developer docs. The docs have been specifically architected to provide information as it is needed rather than trying to do too much. 

We also translate our website into ~40 languages. Adding images with English text is not something we can translate—although we do this already we are thinking on how we can solve this in the future. For the moment I think unless it provides a disproportionate amount of value (which I feel this does not) we should probably pass on any images that require the accompanying text in order for the image to make sense.

minimalsm:
Thanks for the input, appreciate all ideas.

@all-contributors please add @mohan-chinnappan-n for ideas

PR #4895, opened 2021-12-23
Adding clarification to Externally Owned Accounts (EOAs)

nstrike2:
## Description

This is a test PR.


nstrike2:
Hi!

package.json
```diff
@@ -1,3 +1,4 @@
+// super
```
corwintines, line 1: I dont think this needs to be here

src/content/developers/docs/accounts/index.md
```diff
@@ -25,7 +25,7 @@ Both account types have the ability to:
 
 ### Key differences {#key-differences}
 
-**Externally-owned**
+**Externally-owned (wallet address)**
```
corwintines, line 28: I dont think this is quite right. There are smart contract wallets which would not be externally owned accounts.

PR #4898, opened 2021-12-24
Update rust projects and tools

kelcheone:
Adding rust projects not included in the docs.

## Description

Adding rust toolchain and rust implementation of the EVM, ethers, and updating the archived parity repo with the current running fork openEthereum.


minimalsm (commented):
Hey @kelcheone, thanks for the PR 😬👌

I agree with removing Parity since it has been deprecated but unsure why would we would want to add Open Ethereum in its place if it is also no longer actively maintained?

corwintines:
Agree with @minimalsm comment

kelcheone:
will remove Open Ethereum and make a new PR.

PR #4901, opened 2021-12-25
Choice of words in defi doc

bskrksyp9:
## Description
Replaced words that suits better


minimalsm (commented):
Thanks for the PR as always @bskrksyp9 ❤️

Although both of these changes are better in an academic, formal, sort of way, this page is a non-technical learning page that is written to be approachable by a vast array of different sorts of folks and is therefore written in a very colloquial and easy to read style. I think both changes take us away from this style so these probably aren't changes that we would want.

bskrksyp9:
Alright ! thanks for letting me know about this

PR #4908, opened 2021-12-27
Fix dead link in Vyper tutorial

io12:
## Description

There was a dead link in the [Vyper ERC-721 Contract Walkthrough](https://ethereum.org/en/developers/tutorials/erc-721-vyper-annotated-code/). This PR fixes it.


minimalsm:
Thanks for the fix @io12 😬❤️

PR #4920, opened 2021-12-29
Add PoS reading links

samajammin:
## Description

Add PoS reading links

## Related Issue

None


samajammin:
Didn't take the time to do this but one potential improvement for this page would be to add the points of [this article](https://vitalik.ca/general/2020/11/06/pos2020.html) to the pros & cons table.

minimalsm:
> Didn't take the time to do this but one potential improvement for this page would be to add the points of [this article](https://vitalik.ca/general/2020/11/06/pos2020.html) to the pros & cons table.

Threw up a board card for this

PR #4947, opened 2022-01-03
Update Translator Reports for January

corwintines:
Updating report data for translators for translator acknowledgement page


minimalsm (changes requested):
Thanks @corwintines 

Noticed that the Acolad reviewers are showing up on this. 

Screenshot 2022-01-04 at 15 46 54

Seems they are using the 'Aco_' prefix in their displayName rather than their username as requested (cc: @lukassim)

Screenshot 2022-01-04 at 15 47 31

I was able to filter out 7 of the reviewers by adding an additional two checks [to the TranslationLeaderboard logic ](https://github.com/ethereum/ethereum-org-website/blob/e0450e5fc8a33a26d794f849949f6b2fc13a1b19/src/components/TranslationLeaderboard.js#L187)

```js
!item.user.fullName.includes("Aco_") && !item.user.fullName.includes("Acc_")
```

But there are still a bunch that I don't recognise and suspect are reviewers 😟 

e.g. this Nordic reviewer:

Screenshot 2022-01-04 at 15 58 20

I suspect waiting for all the reviewers to change their names will take a while. IMO the best thing to do would be to filter out what we can by `fullName` for now and message Acolad/the reviewers in the meanwhile.

corwintines:
@lukassim @minimalsm this should be ok now. Added some additional filters to catch fullname instances, as well as the specific users from the list @lukassim compiled.

minimalsm:
Build failing for this now 🤔 doesn't look to be caused by your changes though as it happens before them?

Screenshot 2022-01-05 at 09 42 56

corwintines:
@minimalsm got this building now. I think should be good now

wackerow (approved):

PR #4964, opened 2022-01-04
Indonesian: Docs - Tech Stack

wackerow:
## Description
Updates Indonesian (id) "Docs - Tech Stack" content with latest from Crowdin


corwintines (approved):

minimalsm (changes requested):
Not gone through this entirely yet but there is a bunch of fixes needed:

1. Inline code that has been translated that shouldn't be (i.e. properties on a method)
2. Code blocks (e.g. vending machine smart contract solidity syntax translated into Indonesian)

corwintines (approved):

PR #4973, opened 2022-01-05
change get_balance to getBalance in the python tutorial

gosimitz:
w3.eth does not have a get_balance() function, it is getBalance().

## Description
The Python tutorial references `w3.eth.get_balance()` but it is actually `w3.eth.getBalance()`


gosimitz:
Scratch this, I think I don't have the most up to date version of w3. Just saw that the camel case methods are deprecated.

Issue #4793, opened 2021-12-16
Bug: H3 ::before pseudo-element is preventing click interation with the elements above.


evertonfraga:
Hello! 

H3 element has a pseudo-element (::before) with negative margin top, which prevents clicking on links within that range of 120px.

A quick fix would be to use `pointer-events: none` on that pseudo-element, but I'd chase a cleaner solution.

**To Reproduce**

1. Go to https://ethereum.org/en/dao/
1. Try to click on a link right before a H3 element. 

**Expected behavior**

**Screenshots**

<img width="1017" alt="Screen Shot 2021-12-15 at 11 30 46 PM" src="https://user-images.githubusercontent.com/47108/146303725-582e8641-b866-416e-8686-daacf700758a.png">

**Desktop (please complete the following information):**

- OS: MacOS

minimalsm:
Thanks for raising this @evertonfraga :-) would you like to tackle this issue?

corwintines:
@evertonfraga I believe its for scrolling purposes for links

samajammin:
Thanks for offering to help out @evertonfraga! Good to see you again 😄 

Yup, expand on @corwintines's comments, the reason for the negative margin is so that if you open a URL w/ an anchor tag (e.g. https://ethereum.org/en/developers/docs/nodes-and-clients/#full-node), the browser & nav doesn't overlap the header.

Curious to hear what you come up with , I'm sure there's better approaches than the negative margin.

evertonfraga:
sure, thanks @samajammin! there's `scroll-padding` for that particular problem. >90% support, afaict. 

OK expect a PR from me :)

TonyGaski:
hey anyone here able to help me familiarize myself with this stuff?

corwintines:
@TonyGaski let me know if there is something else you are interested in looking into. If you want to see how this got solved I just opened a PR for this.

evertonfraga:
thanks for fixing that before me, @corwintines :)

PR #5003, opened 2022-01-06
Bug: H3 ::before pseudo-element is preventing click interation with the elements above

corwintines:
## Description
- Added scroll-margin-top to headers
- Added a hook in for when a page is loaded with a anchor tag and scroll to that anchor

## Related Issue
fixes: https://github.com/ethereum/ethereum-org-website/issues/4793

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -436,6 +436,7 @@ export const GlobalStyle = createGlobalStyle`
     margin: 2rem 0;
     font-weight: 700;
     scroll-margin-top: ${theme.variables.navHeight};
+    scroll-snap-margin: ${theme.variables.navHeight};
     @media (max-width: ${theme.breakpoints.m}) {
       font-size: 2.5rem;
     }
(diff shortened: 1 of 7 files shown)


wackerow (changes requested):
Lookin good

https://build-4d984deb-29ab-4739-bbb6-dae49915c8f4.gtsb.io/en/dao/#share-based-membership
Going here seems to properly navigate to the correct section with space for the nav bar, and the link above it remains clickable. 

Just left code suggestions to use our theme variable for the nav height instead of hard-coding.

wackerow (approved):

src/theme.js
```diff
@@ -437,6 +435,7 @@ export const GlobalStyle = createGlobalStyle`
     line-height: 1.4;
     margin: 2rem 0;
     font-weight: 700;
+    scroll-margin-top: ${theme.variables.navHeight};
```
minimalsm, line 438:
```suggestion
    scroll-snap-margin: ${theme.variables.navHeight};
    scroll-margin-top: ${theme.variables.navHeight};
```

minimalsm (commented):
Nice @corwintines. Working well for me. 

Noticed on [caniuse.com](https://caniuse.com/?search=scroll-margin) that if we include polyfill with `scroll-snap-margin` we get an additional ~3% browser coverage for iOS devices. Probably worth it?

src/theme.js
corwintines: I tried using this and it wasn't working. Maybe we can make an issue to change over, but would like to see this get in for now if thats ok with you.

PR #5005, opened 2022-01-06
Fix UI bugs on some versions of new homepage

corwintines:
## Description
- Allow text to wrap in ButtonLink to avoid overflow and horizontal scroll in mobile 

## Related Issue
None


wackerow (approved):
👍🏻

PR #5008, opened 2022-01-07
Run an Ethereum node

wackerow:
Introduces a beginner friendly page on running an Ethereum node:
https://ethereumorgwebsitedev01-runanode.gtsb.io/en/run-a-node/

Intended to be a resource for those who are less technically oriented and those who may have never considered running their own node before. Primary focus of this page is not on staking, though it can be used for users to get started with staking as well.

Reviews and feedback appreciated =)

- [x] Need to create an expandable info component for the "_Who_ should run a node?" section
- [x] Added links from `nodes-and-clients` and `nodes-and-clients/run-a-node` pages 
- [x] Update for "eth2" changes
- [x] Extract intl strings

Can also consider other places to link from


src/components/ExpandableCard.js
```diff
@@ -50,11 +50,21 @@ const Question = styled.div`
...
+    margin-right: 1.5rem;
+  }
+`
+
 const ButtonContainer = styled.div`
   margin-left: 1rem;
 `
 
-const ExpandableCard = ({ children, contentPreview, title }) => {
+const ExpandableCard = ({ children, contentPreview, title, svg }) => {
```
minimalsm, line 67: Not a fan of how we have to click directly on the 'More' text to expand the additional content - [example GIF](https://gyazo.com/574e7a8b9b8dd4b9a0cf7712908574cc). This is quite a small target and [could be difficult for mobile users to interact with](https://www.nngroup.com/articles/touch-target-size/).

minimalsm: I think it'd be an improvement if we make the entire Card clickable - [example GIF](https://gyazo.com/8f4f0e9f028d08572dd5048df9db3bb4). 

I've checked and we also use this on the History page and the [staking community grants page](https://ethereum.org/en/community/grants/staking-community-grants), neither are negatively affected by the change.

src/components/ExpandableCard.js
```diff
@@ -96,7 +106,10 @@ const ExpandableCard = ({ children, contentPreview, title }) => {
     <Card>
```
minimalsm, line 106:
```suggestion
    <Card onClick={() => setIsVisible(!isVisible)}>
```

src/components/ExpandableCard.js
```diff
@@ -96,7 +106,10 @@ const ExpandableCard = ({ children, contentPreview, title }) => {
...
       <Content>
         <Question>
-          <Title>{title}</Title>
+          <Header>
+            {svg && <img src={svg} />}
+            <Title>{title}</Title>
+          </Header>
           <TextPreview>{contentPreview}</TextPreview>
         </Question>
         <ButtonContainer onClick={() => setIsVisible(!isVisible)}>
```
minimalsm, line 115:
```suggestion
        <ButtonContainer>
```

minimalsm: If we agree on this change we also need to add `pointer: cursor` to the `Card` styled-component within ExpandableCard.js

src/pages/run-a-node.js
```diff
@@ -0,0 +1,1011 @@
...
+    image: getImage(data.ethereumInside),
+    alt: <Translation id="page-run-a-node-hero-alt" />,
+    buttons: [
+      {
+        content: <Translation id="page-run-a-node-button-why-run-a-node" />,
+        pathId: "why-run-a-node",
+      },
+      {
+        content: <Translation id="page-run-a-node-button-get-started" />,
+        pathId: "getting-started",
```
minimalsm, line 424: I don't think we should have two primary action buttons here. [More info on best practices.](https://balsamiq.com/learn/articles/button-design-best-practices/#2-visually-differentiate-primary-and-secondary-actions)

```suggestion
        pathId: "getting-started",
        isSecondary: "isSecondary",
```

src/pages/run-a-node.js
```diff
@@ -0,0 +1,1011 @@
...
+  }
+`
+
+const ScrollButtonSecondary = styled.button`
+  text-decoration: none;
+  display: inline-block;
+  padding: 0.5rem 0.75rem;
+  font-size: 1rem;
+  border-radius: 0.25em;
+  text-align: center;
```
minimalsm, line 359:
```suggestion
  text-align: center;
  cursor: pointer;
```

src/pages/run-a-node.js
```diff
@@ -0,0 +1,1011 @@
...
+  display: flex;
+  gap: 1rem;
+  margin-top: 2rem;
+  @media (max-width: ${({ theme }) => theme.breakpoints.s}) {
+    flex-direction: column;
+  }
+`
+
+const DappNodeButtonLink = styled(ResponsiveButtonLink)`
+  background-color: #30bcb2;
```
minimalsm, line 320: This button doesn't pass accessibility contrast tests. Noticed you're using the Dapp node colours for this which I think is a nice touch—maybe we could tweak it slightly to make it compliant? 

Screenshot 2022-01-07 at 23 15 15

src/intl/en/page-run-a-node.json
```diff
@@ -0,0 +1,116 @@
...
+  "page-run-a-node-what-1-text": "Known as a \"client,\" this software downloads a copy of the Ethereum blockchain, keeps it up-to-date with new blocks and transactions, and helps others download and update their own copies.",
+  "page-run-a-node-what-2-subtitle": "With hardware.",
+  "page-run-a-node-what-2-text": "Ethereum is designed to run a client on average consumer-grade computers. You can use any personal computer, but most users opt to run their node on dedicated hardware to eliminate the performance impact on their machine and minimize node downtime.",
+  "page-run-a-node-what-3-subtitle": "While online.",
+  "page-run-a-node-what-3-text": "Running an Ethereum node may sound complicated at first, but it's merely the act of continuously running client software on a computer while connected to the internet. While offline, your node will simply be inactive until it gets back online and catches up with the latest changes.",
+  "page-run-a-node-who-title": "<i>Who</i> should run a node?",
+  "page-run-a-node-who-preview": "Nodes are not just for miners and validators. <i>Anyone</i> can run a node (you don't even need ETH).",
```
minimalsm, line 110: Feel like dropping the brackets here is a bit cleaner.

Screenshot 2022-01-07 at 23 27 38

Screenshot 2022-01-07 at 23 28 00


```suggestion
  "page-run-a-node-who-preview": "Nodes are not just for miners and validators. <i>Anyone</i> can run a node—you don't even need ETH.",
```

src/intl/en/page-run-a-node.json
```diff
@@ -0,0 +1,116 @@
...
+  "page-run-a-node-feedback-prompt": "Did you find this page helpful?",
+  "page-run-a-node-further-reading-title": "Further reading",
+  "page-run-a-node-further-reading-1-link": "Mastering Ethereum - Should I Run a Full Node",
+  "page-run-a-node-further-reading-1-author": "Andreas Antonopoulos",
+  "page-run-a-node-further-reading-2-link": "Ethereum on ARM - Quick Start Guide",
+  "page-run-a-node-further-reading-3-link": "The Limits to Blockchain Scalability",
+  "page-run-a-node-further-reading-3-author": "Vitalik Buterin",
+  "page-run-a-node-getting-started-hardware-title": "Getting started - Part 1: Hardware",
+  "page-run-a-node-getting-started-hardware-1": "You'll need some hardware to get started. Although running node software is possible on a personal computer, having a dedicated machine can greatly enhance the performance of your node, while minimizing its impact on your primary computer.",
```
minimalsm, line 47:
```suggestion
  "page-run-a-node-getting-started-hardware-1": "You'll need some hardware to get started. Although running node software is possible on a personal computer, having a dedicated machine can greatly enhance the performance of your node while minimizing its impact on your primary computer.",
```

src/intl/en/page-run-a-node.json
```diff
@@ -0,0 +1,116 @@
...
+  "page-run-a-node-getting-started-hardware-2": "When selecting hardware, take into consideration that the chain is always slowly growing, and maintenance will inevitably be needed. Increasing specs can help delay the need for node maintenance.",
```
minimalsm, line 48:
```suggestion
  "page-run-a-node-getting-started-hardware-2": "When selecting hardware, consider that the chain is continually growing, and maintenance will inevitably be needed. Increasing specs can help delay the need for node maintenance.",
```

src/pages/run-a-node.js
```diff
@@ -0,0 +1,1011 @@
...
+                    <p>
+                      <ScrollLink onClick={() => scrollToId("rasp-pi")}>
+                        <Translation id="page-run-a-node-build-your-own-ram-note-2" />
+                      </ScrollLink>
+                    </p>
+                  </li>
+                  <li>
+                    <Translation id="page-run-a-node-build-your-own-min-ssd" />
+                    <p>
+                      <small>
```
minimalsm, line 623: This feels too much text to be a 'small',

minimalsm: > Though you can still technically get away with 1 TB, the blockchain continues to grow over time and this will soon no longer be enough.

Also, with the statelessness and state-expiry roadmap, I'm not sure if we should mention this. The idea of having to continually upgrade your node hardware indefinitely might put some potential node runners off.

src/pages/run-a-node.js
```diff
@@ -0,0 +1,1011 @@
...
+          </li>
+        </ul>
+      </Content>
+
+      <Content>
+        <h2>
+          <Translation id="page-run-a-node-getting-started-software-title" />
+        </h2>
+        <GappedContent>
+          <SoftwareHighlight backgroundColor="homeBoxTurquoise">
```
minimalsm, line 756: This section feels a bit cramped to me. One thought I had was maybe making it a bit narrower similar to the '[How to stake](https://ethereum.org/en/eth2/staking)' section on our staking page?

**Proposed**
Screenshot 2022-01-07 at 23 42 07

**Current**
Screenshot 2022-01-07 at 23 42 15

minimalsm: Don't love this solution but 🤷‍♂️

minimalsm (changes requested):
Hey @wackerow this is looking awesome.

Left a few comments largely around small styling tweaks. Made a couple of suggestions on the content but will look more over weekend—takes forever to review the Translation JSONs 😬

src/components/ExpandableCard.js
wackerow: Yeah, personally prefer being able to click the entire card as well. Tried to start by reusing components, but would be down to modify this.

src/intl/en/page-run-a-node.json
wackerow: Agreed, the ol' em dash alternative

src/pages/run-a-node.js
wackerow: Remove it and just leave "2 TB" as the recommended minimum?

I'd not against that, but if you're suggesting changing it to just 1 TB I'd be hesitant until statelessness/expiry are actually live. 

This is a tricky section, since each client is a little different, and future-proofing / trying to minimize future maintenance is somewhat subjective. In the end I'd prefer to keep it simple, so I'm down to just keep it at "2 TB SSD" and leave the note that SSD speeds are essential.

src/pages/run-a-node.js
wackerow: Yeah wouldn't be against adding some horizontal margin on those divs when on suitable screen sizes.

src/components/ExpandableCard.js
wackerow: Only hesitation with doing this... It's nice for opening/closing, but if you want to select any of the expanded text (at least on desktop with a mouse), it will close the card when you un-click to finish your selection.

I tried it by moving `onClick={() => setIsVisible(!isVisible)}` to `<Content>` instead, which works pretty well by only closing when you click the top section of the card, but it does leave the padding un-clickable. Curious your thought on this.

src/intl/en/page-run-a-node.json
```diff
@@ -0,0 +1,115 @@
...
+  "page-run-a-node-title": "Run a node",
+  "page-run-a-node-voice-your-choice-title": "Voice your choice",
+  "page-run-a-node-voice-your-choice-preview": "Don't give up control in the event of a fork.",
+  "page-run-a-node-voice-your-choice-1": "In the event of a chain fork, where two chains emerge with two different sets of rules, running your own node guarantees your ability to choose which set of rules you support. It’s up to you to upgrade to new rules and support proposed changes, or not.",
+  "page-run-a-node-voice-your-choice-2": "If you're staking ETH, running your own node allows you to chose your own client, to minimize your risk of slashing and to react to fluctuating demands of the network over time. Staking with a third party forfeits your vote on which client you think is the best choice.",
+  "page-run-a-node-what-title": "<i>What</i> does it mean to \"run a node\"?",
+  "page-run-a-node-what-1-subtitle": "Run software.",
+  "page-run-a-node-what-1-text": "Known as a \"client,\" this software downloads a copy of the Ethereum blockchain, keeps it up-to-date with new blocks and transactions, and helps others download and update their own copies.",
```
taxmeifyoucan, line 103: > downloads a copy of the Ethereum blockchain, keeps it up-to-date with new blocks and transactions '

One important fact I would add here is that client (cryptographically, mathematically) verifies this data.

src/pages/run-a-node.js
minimalsm: I think we should say 2 TB and hope state bloat is fixed before they'll need to upgrade 😅

taxmeifyoucan: You can also refer table under this https://ethereum.org/en/developers/docs/nodes-and-clients/#recommended-specifications
I would keep just 1TB as minimum, with prunning of old states, dropping ancient data, 1TB will be more than enough.

src/intl/en/page-run-a-node.json
wackerow: Thanks for the feedback @taxmeifyoucan .. Yeah, can't disagree too strongly. Been a little back and forth on this statement, trying to keep it really simple for people in this section. Will keep this in mind for sure as we get other feedback. 👍🏻

src/components/ExpandableCard.js
wackerow: Tried this out... not sure I love it enough to keep it in for this PR, so reverted it back. The "ExpandableCard" is an existing component, which if we'd like to revamp it I have no problem with that, but requesting we address that separately to limit scope creep here.

src/pages/run-a-node.js
wackerow: I think I'm in favor of just recommending 2TB here. At the risk of excluding some people who are willing/able to get a 1 TB but not a 2 TB drive, recommending 2 will minimize future issues and complications in the attempt to keep this geared as much as possible towards non-technical users. This will also keep things really simple as we approach the merge, where the Beacon Chain will also need to be synced. 

In the future as statelessness, and history/state expiry become a reality, we can certainly readdress this minimum, and hopefully drop it entirely someday. In the meantime, of course more complex setups exist to bring cost down, but again this guide is intended to keep things as simple as possible, and I think this is a reasonable way to help accomplish this.

src/pages/run-a-node.js
```diff
@@ -0,0 +1,1015 @@
...
+                <p key={idx}>{<Translation id={item} />}</p>
+              ))}
+            </ExpandableCard>
+          ))}
+        </InfoGrid>
+      </Content>
+
+      <Divider />
+
+      <Content id="getting-started">
```
samajammin, line 590: Rather than:
- "Getting started - Part 1: Hardware"
- "Part 2: Software"

 I think I'd prefer a structure of:
- "Get started"
  - "Part 1: Hardware"
  - "Part 2: Software"

src/intl/en/page-run-a-node.json
samajammin: I agree - came here to say that. Maybe something like:

> Known as a "client," this software downloads a copy of the Ethereum blockchain, keeps it up to date by verifying new activity, and helps others download and update their own secure copies.

samajammin: On the other hand, I'm pretty fine with what you have. I'm guessing these users will understand what transactions & blocks are...

wackerow: Just tossing out another take

> Known as a \"client,\" this software downloads a copy of the Ethereum blockchain and verifies the validity of every block, then keeps it up-to-date with new blocks and transactions, and helps others download and update their own copies.
```suggestion
  "page-run-a-node-what-1-text": "Known as a \"client,\" this software downloads a copy of the Ethereum blockchain and verifies the validity of every block, then keeps it up-to-date with new blocks and transactions, and helps others download and update their own copies.",
```

src/components/ExpandableCard.js
samajammin: I tend to agree but think we can keep this as a separate PR, given that we use this card component all over the place.

src/components/ExpandableCard.js
samajammin: Agreed. As I noted above, this is why I think we should address that separately.

src/pages/run-a-node.js
samajammin: Agree with @wackerow's last comment. I'd rather err on the side of caution.

src/intl/en/page-run-a-node.json
samajammin: LGTM @wackerow 👍

src/pages/run-a-node.js
samajammin: @wackerow noting this still needs to be addressed.

samajammin:
Image 2022-01-19 at 12 00 19 PM

While I like the structure of this section I wonder if these sentence fragments ("With hardware." & "While online.") will create potential translations issues. Curious what @lukassim thinks.

src/intl/en/page-run-a-node.json
```diff
@@ -0,0 +1,115 @@
...
+  "page-run-a-node-what-2-subtitle": "With hardware.",
+  "page-run-a-node-what-2-text": "Ethereum is designed to run a client on average consumer-grade computers. You can use any personal computer, but most users opt to run their node on dedicated hardware to eliminate the performance impact on their machine and minimize node downtime.",
```
samajammin, line 105:
```suggestion
  "page-run-a-node-what-2-text": "Ethereum is designed to be able to run a client on average consumer-grade computers. You can use any personal computer, but most users opt to run their node on dedicated hardware to eliminate the performance impact on their machine and minimize node downtime.",
```

src/intl/en/page-run-a-node.json
```diff
@@ -0,0 +1,115 @@
...
+  "page-run-a-node-what-3-subtitle": "While online.",
+  "page-run-a-node-what-3-text": "Running an Ethereum node may sound complicated at first, but it's merely the act of continuously running client software on a computer while connected to the internet. While offline, your node will simply be inactive until it gets back online and catches up with the latest changes.",
```
samajammin, line 107: Should we mention the why here? e.g. "merely the act of continuously running client software on a computer while connected to the internet, so that your computer can communicate and sync with other peers in the network."

src/intl/en/page-run-a-node.json
```diff
@@ -0,0 +1,115 @@
...
+  "page-run-a-node-who-title": "<i>Who</i> should run a node?",
+  "page-run-a-node-who-preview": "Nodes are not just for miners and validators. <i>Anyone</i> can run a node—you don't even need ETH.",
```
samajammin, line 109:
```suggestion
  "page-run-a-node-who-preview": "Everyone! Nodes are not just for miners and validators. <i>Anyone</i> can run a node—you don't even need ETH.",
```

src/intl/en/page-run-a-node.json
```diff
@@ -0,0 +1,115 @@
...
+  "page-run-a-node-who-copy-1": "You don't need to stake ETH or be a miner to run a home node. In fact, it's every other node on Ethereum that holds miners and validators accountable.",
```
samajammin, line 110:
```suggestion
  "page-run-a-node-who-copy-1": "You don't need to stake ETH or be a miner to run a node. In fact, it's every other node on Ethereum that holds miners and validators accountable.",
```

src/intl/en/page-run-a-node.json
```diff
@@ -0,0 +1,115 @@
...
+  "page-run-a-node-who-copy-2": "You may not get the financial rewards that validators and miners earn, but there are many other benefits of running a home node for any Ethereum user to consider, including privacy, security, reduced reliance on third-party servers, censorship resistance and improved health and decentralization of the network.",
```
samajammin, line 111:
```suggestion
  "page-run-a-node-who-copy-2": "You may not get the financial rewards that validators and miners earn, but there are many other benefits of running a node - both for you and the network - that every Ethereum user should consider, including privacy, security, reduced reliance on third-party servers, censorship resistance and improved health and decentralization of the network.",
```

samajammin:
Dropped in a few comments but didn't finish. Will come back to this tonight! Feel free to hold off on making changes until I've finished.

src/components/ExpandableCard.js
samajammin: @wackerow please create a separate story in this epic to address this.

src/components/FeedbackCard.js
```diff
@@ -38,10 +38,10 @@ const ButtonContainer = styled.div`
   }
 `
 
-const FeedbackCard = () => {
+const FeedbackCard = ({ prompt, className }) => {
   const [feedbackSubmitted, setFeedbackSubmitted] = useState(false)
   const [isHelpful, setIsHelpful] = useState(false)
-
+  const feedbackPrompt = prompt || "Did this page help answer your question?"
```
samajammin, line 44: We should translate this string.

samajammin:
I don't follow this section in red:
Image 2022-01-25 at 7 43 08 PM
I think we could just remove it entirely? We could update the 2nd bullet to "App-like setup with a GUI."

samajammin (changes requested):
So close! Made some minor copy changes. A few comments to address. Once those are made, I think we're ready to ship 🚢

wackerow:
@samajammin So the above was accidentally incomplete copy... Should've said "small in size, but come fully loaded."
Somewhat of an after thought, but was hoping to include this as another subtle reminder that we're not talking large mining rigs... many NUCs only have about an 11 sq cm footprint, standing <5 cm tall.
Thoughts on if we trim out the additional comment about choosing your hardware, move the GUI comment, and leave this small callout?
image

wackerow:
Also, still have two checkboxes before merging... intl strings and eth2 updates. Will get these going in a branch.

samajammin:
@wackerow my take - I we just remove that "these solutions are small in size, but come fully loaded." section & image. We already mention "fully loaded" in the same card above. We also mention nodes run on consumer hardware earlier in the page. I feel like mentioning this in the "fully loaded" solution implies that's not the case for the "build your own".

samajammin:
@wackerow also FWIW I'd say we could ship this page without full intl support. I think it'd be good to collect some community feedback (& potentially make tweaks) before we upload the page to Crowdin for translation.

wackerow:
@samajammin Okay cool, https://github.com/ethereum/ethereum-org-website/pull/5226 has the latest here. After we get that pulled into this branch, should be ready for final review before prod.

samajammin (approved):
🚀

Drblessing:
This resource was a great help setting up a node. Could there be more resources about exposing a client node's RPC/HTTP server for a small remote community? Right now in [Spin up your own Ethereum node](https://ethereumorgwebsitedev01-runanode.gtsb.io/en/developers/docs/nodes-and-clients/run-a-node/#spinning-up-node), it mentions port forwarding the router and setting `--http.addr 0.0.0.0`. It also mentions routing it through tor, or a web server like Nginx, and a separate guide for this would be much appreciated.

jjmstark:
@wackerow I think we should probably change this page a bit to make clear that dappnode is just one company that offers a solution. 

In particular I am thinking of this section:
Screen Shot 2022-04-14 at 12 07 32 PM

This feels like an "endorsement" compared to how we usually show projects/companies on a page. We also use the Dappnode logo in this section in a way that does not make it clear to the reader that we are using someone's logo, and not just an image representing "running a node".  It is only lower on the page that we show another option (Avado). 

Maybe a way to change it would be to instead have this section communicate something like "Now we have simple node-running services like Dappnode or Avado that gives users an app-like experience while managing their node."

Same point about the "find some helpers" section, maybe we ought to be linking to more than just the dappnode community?

PR #5009, opened 2022-01-07
Removed Altair multiplier for "eth2 bug bounty" page

fredriksvantes:
## Description

Removed Altair multiplier for "eth2 bug bounty" page


minimalsm (approved):
Thanks @fredriksvantes

minimalsm:
@lukassim curious you're thoughts about the translated languages that include this. I think we're probably good to let this run its course through the translation pipeline?

lukassim:
@minimalsm In this case, it would be easy to remove everything but the first sentence. However, they're all part of the same string, so we would need to update the source file first.

PR #5011, opened 2022-01-07
Fix overflow of text on /find-wallet/ header

minimalsm:
## Description
- Added a max-width of 45 characters to each Subtitle
- Use the same colour on both subtitles (amused this was a mistake? can revert otherwise)
- Use rem over px (a11y)
- Small refactor to remove redundancy 

## Related Issue
Fixes #5010

Issue #5010, opened 2022-01-07
Text overflowing onto images on some languages (Find-wallet)


minimalsm:
**Describe the bug**
On some Slavic languages, we're seeing text-overflow due to the generally longer length of sentences in these languages (relative to English).

**To Reproduce**

1. Go to 'https://ethereum.org/ru/wallets/find-wallet/'
2. See error

**Expected behavior**
The text should break to a new line if it gets too long

**Screenshots**

Russian:
![Screenshot 2022-01-07 at 13 16 08](https://user-images.githubusercontent.com/62268199/148549343-f05a8986-9817-4e2d-8439-d8e0310a98d7.png)

Slovenian:
![Screenshot 2022-01-07 at 13 17 42](https://user-images.githubusercontent.com/62268199/148549522-77d92e30-fe86-4862-897f-40cec0a6e7d9.png)

Bulgarian
![Screenshot 2022-01-07 at 13 17 42](https://user-images.githubusercontent.com/62268199/148549548-e96ddf46-a447-4cc7-a97b-374ff525809a.png)

PR #5027, opened 2022-01-08
Fix overflow of text on /get-eth/ header

minimalsm:
## Description

Similar to #5011, fixes overflowing text on language with longer sentences.

- Added a max-width of 45 characters to each Subtitle
- Use the same colour on both subtitles (amused this was a mistake? can revert otherwise)
- Use rem over px (a11y)
- Small refactor to remove redundancy

## Related Issue
#5010

### DIFF
--- a/src/pages/get-eth.js
+++ b/src/pages/get-eth.js
@@ -31,10 +31,7 @@ const Subtitle = styled.div`
   text-align: center;
   color: ${(props) => props.theme.colors.text200};
 
-  /* Add margin-bottom to the second subtitle */
-  &:nth-last-of-type(2) {
-    margin-bottom: 2rem;
-  }
+  margin-bottom: ${(props) => props.mb || ""};
 `
 
 const HeroContainer = styled.div`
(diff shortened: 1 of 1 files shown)


corwintines (approved):
LGTM 🚀

src/pages/get-eth.js
```diff
@@ -25,18 +25,16 @@ import {
...
-const SubtitleTwo = styled.div`
-  font-size: 20px;
-  line-height: 140%;
-  margin-bottom: 2rem;
-  text-align: center;
-  color: ${(props) => props.theme.colors.text300};
+  /* Add margin-bottom to the second subtitle */
+  &:nth-last-of-type(2) {
+    margin-bottom: 2rem;
+  }
```
wackerow, line 37: Do we want this on the "second subtitle" (`&:nth-of-type(2)`), or do we want this on the _last_ subtitle (`&:last-of-type`)? Not sure we want to use `&:nth-last-of-type(2)`..

If we just want the margin to be on the last item:
```suggestion
  /* Add margin-bottom to the second subtitle */
  &:last-of-type {
    margin-bottom: 2rem;
  }
```

wackerow (changes requested):
Just made a comment to double check the selector used to add the margin to the second subtitle, otherwise looks good 👍🏻

minimalsm:
The `&:last-of-type` selector selects the last HTML element (div), not the styled-component (Subtitle).

Thinking about that, this solution is tightly coupled to the order of components here. I've pushed a small refactor where we can use `mb` as a prop. If we don't want to do this, we should probably just create `SubtitleTwo` again.

wackerow (commented):
lgtm =)

PR #5080, opened 2022-01-13
Arabic: Use Ethereum

wackerow:
## Description
Updates Arabic (ar) "Use Ethereum" content with latest from Crowdin


minimalsm (changes requested):
We're still using the old pages for

- /what-is-ethereum/
- /eth/
- /dapps/
- /wallets/

Screenshot 2022-01-13 at 12 53 26

PR #5107, opened 2022-01-14
Extract "What happened to Eth2?" intl json strings

wackerow:
`eth2-rebrand <- what-happened`

## Description
Extracts intl strings for "What happened to Eth2?" section into json files


src/intl/en/page-upgrades-index.json
```diff
@@ -169,6 +169,14 @@
   "page-upgrades-vision-btn": "More on the Ethereum vision",
   "page-upgrades-vision-desc": "To bring Ethereum into the mainstream and serve all of humanity, we have to make Ethereum more scalable, secure, and sustainable.",
   "page-upgrades-vision-upper": "Ethereum vision",
+  "page-upgrades-what-happened-to-eth2-title": "What happened to Eth2?",
+  "page-upgrades-what-happened-to-eth2-1": "The term 'Eth2' is being phased out in preparation for the merge.",
+  "page-upgrades-what-happened-to-eth2-1-more": "More on the merge.",
+  "page-upgrades-what-happened-to-eth2-2": "After merging Eth1 and Eth2 into a single chain, there will no longer be two distinct Ethereum networks; there will only be Ethereum.",
```
corwintines, line 175:
```suggestion
  "page-upgrades-what-happened-to-eth2-2": "After merging what is formerly known as 'Eth1' and 'Eth2' into a single chain, there will no longer be two distinct Ethereum networks; there will only be Ethereum.",
```

wackerow: Yeah, in favor of this, with a qualifier and quotes to limit how much it looks like we're currently using these terms

corwintines (approved):

src/intl/en/page-upgrades-index.json
wackerow: Then again this PR is more just to extract the copy that was there... we can always re-address any copy suggestions separately

PR #5117, opened 2022-01-16
Adds STARK as a type of ZK proof on layer-2-rollups doc

mmilenkovic:
- Adding STARK (beside SNARK) as a type of ZK proof.
- Adding external link to explain STARKs vs SNARKs.

### DIFF
--- a/src/content/developers/docs/scaling/layer-2-rollups/index.md
+++ b/src/content/developers/docs/scaling/layer-2-rollups/index.md
@@ -87,7 +87,7 @@ Multiple implementations of Optimistic rollups exist that you can integrate into
 
 ### Zero-knowledge rollups {#zk-rollups}
 
-**Zero-knowledge rollups (ZK-rollups)** bundle (or "roll-up") hundreds of transfers off-chain and generate a cryptographic proof, known as a SNARK (succinct non-interactive argument of knowledge). This is known as a validity proof and is posted on layer 1.
+**Zero-knowledge rollups (ZK-rollups)** bundle (or "roll-up") hundreds of transfers off-chain and generate a cryptographic proof. These proofs can come in a form of a SNARK (Succinct Non-interactive ARgument of Knowledge) or a STARK (Scalable Transparent ARgument of Knowledge). This is known as a validity proof and is posted on layer 1.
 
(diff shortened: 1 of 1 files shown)


src/content/developers/docs/scaling/layer-2-rollups/index.md
```diff
@@ -87,7 +87,7 @@ Multiple implementations of Optimistic rollups exist that you can integrate into
 
 ### Zero-knowledge rollups {#zk-rollups}
 
-**Zero-knowledge rollups (ZK-rollups)** bundle (or "roll-up") hundreds of transfers off-chain and generate a cryptographic proof, known as a SNARK (succinct non-interactive argument of knowledge). This is known as a validity proof and is posted on layer 1.
+**Zero-knowledge rollups (ZK-rollups)** bundle (or "roll-up") hundreds of transfers off-chain and generate a cryptographic proof. These proofs can come in a form of a SNARK (Succinct Non-interactive ARgument of Knowledge) or a STARK (Scalable Transparent ARgument of Knowledge). This is known as a validity proof and is posted on layer 1.
```
minimalsm, line 90: - I don't think we want to be so blunt about where the acronyms come from. It seems messy and doesn't add much in terms of understanding.
- Small grammar fix

```suggestion
**Zero-knowledge rollups (ZK-rollups)** bundle (or "roll-up") hundreds of transfers off-chain and generate a cryptographic proof. These proofs can come in the form of SNARKs (succinct non-interactive argument of knowledge) or STARKs (scalable transparent argument of knowledge). SNARKs and STARKs are known as validity proofs and get posted to layer 1.
```

minimalsm (changes requested):
Hey @mmilenkovic Thanks for suggesting this 😀!

Definitely agree we should be explaining both. Made a comment about a small stylistic change I think we should make to this.

mmilenkovic:
@minimalsm Agree with proposed changes, already made a commit.

BTW Not sure if you have control over Glossary but I proposed adding "[Zero-knowledge proof](https://github.com/ethereum/ethereum-org-website/issues/5121)", "[SNARK](https://github.com/ethereum/ethereum-org-website/issues/5119)" and "[STARK](https://github.com/ethereum/ethereum-org-website/issues/5120)"
When/if those are included we can link those as well.

minimalsm:
Yeah, good thinking—if we reach consensus on adding these to the glossary it's a good idea to loop back here and add them.

minimalsm:
@all-contributors please add @mmilenkovic for docs

PR #5134, opened 2022-01-17
Updated ECH invite link 

SunitRoy2703:
## Description

Updated ECH invite link from `discord.io` page to original discord invite link

## Related Issue

Users don't have to get redirected from `discord.io` the slash page

### DIFF
--- a/src/content/community/get-involved/index.md
+++ b/src/content/community/get-involved/index.md
@@ -42,7 +42,7 @@ If you’re not a developer, it can be hard to know where to start in Ethereum. 
 
 ### Offer to take notes for community calls {#take-notes}
 
-- There are many open-source community calls, and having notetakers is a huge help. If you’re interested, join the [Ethereum Cat Herders discord](https://discord.io/EthCatHerders), and introduce yourself!
+- There are many open-source community calls, and having notetakers is a huge help. If you’re interested, join the [Ethereum Cat Herders discord](https://discord.com/invite/Nz6rtfJ8Cu), and introduce yourself!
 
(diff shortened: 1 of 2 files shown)


wackerow (approved):
Yeah definitely faster this way, too bad we can't have speed and the nice URL hah. Thanks @SunitRoy2703!

SunitRoy2703:
I have just added the invite link, which their `discord.io` page was redirecting to

Please check if there is any problem with this

@minimalsm @wackerow

minimalsm:
> @minimalsm I'm cool with this as long as there isn't any particular spam preventing reason.

Ya agree if it doesn't serve a particular purpose we should just use the native invite. I'm just cognizant that the ECH discord was recently hacked so it might be a security concern. I've pinged Pooja on Discord to weigh in.

poojaranjan:
We decided to go by the customized link for convenience and safety. But, if there is a preference to have native link for Discord, it shouldn't be a problem for us. 
To be sure that the invite link added to the Ethereum.org website should be for unlimited period, I'd recommend adding https://discord.gg/Nz6rtfJ8Cu

wackerow (approved):

wackerow:
Thanks for weighing in @poojaranjan! And thank you @SunitRoy2703 for the fix!

wackerow:
@all-contributors please add @SunitRoy2703 for bug patch

SunitRoy2703:
> Thanks for weighing in @poojaranjan! And thank you @SunitRoy2703 for the fix!

Thank you so much!

Issue #4714, opened 2021-12-10
Style events to look like a blockchain


corwintines:
**Is your feature request related to a problem? Please describe.**
Would like to have the events listed on `/community/events` styled to have a connection between them to look like a blockchain. Some inspiration to draw from: https://bitcoin.org/en/events

**Additional context**
- Feel free to reach out to @corwintines about this

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

tburnam:
I'd love to tackle this!

corwintines:
@tburnam assigning to you :) Let me know if you need anything.

minimalsm:
@tburnam just checking you've seen your assignment to this :-)

vluna:
@minimalsm I would like to help if possible 🙂 Let me know if there is anything I can do.

minimalsm:
@tburnam no just submitting it as a PR is fine. 

Any update on the status of this ❤️?

vluna:
@minimalsm I would like to give it a go if there is no update on this.

minimalsm:
Thanks offering again @vluna 🙏 Hmm yeah feel free to have a swing at this.

minimalsm:
Closed in #5150

PR #5150, opened 2022-01-19
Style events like a blockchain

vluna:
## Description
- Update overall style for the events
- Update `EventCard` component with new style for the events
- Update `UpcomingEventsList` component with new changes

## Related Issue
#4714


vluna:
@minimalsm I am not super familiar with gatsby/react, so please let me know if there is a better way 🙂

src/components/EventCard.js
```diff
@@ -40,20 +80,51 @@ const Title = styled.h3`
...
+          <Location>
+            <LocationText>{location}</LocationText>
+            <Emoji text=":round_pushpin:" size={1} ml={`0.5em`} />
+          </Location>
+          <Title>{title}</Title>
+          <Description>{description}</Description>
+          <ButtonLink to={to}>View Event</ButtonLink>
+        </StyledCardContentRight>
+      )}
+    </StyledCard>
```
wackerow, line 126: Couple thoughts here on ways to reduce redundant code...
1- Pass `isEven` as a prop to `StyledCardContent` (`<StyledCardContent isEven={isEven}>`) and then pick this up in the CSS to conditionally render on the left or right.
2- Could try using `&:nth-child(even)` from the parent to style all even children differently. 
3- Could otherwise declare a constant before the return in this component that holds all the children that are currently written twice, then reuse that constant (as a mini component) in the final return. 

The CSS approach I think would be the cleanest, but otherwise the first suggestion would suffice and would remove the need to have this duplicated.

src/components/UpcomingEventsList.js
```diff
@@ -82,15 +106,15 @@ const UpcomingEventsList = () => {
...
-              />
-            </EventListItem>
+            <EventCard
+              key={idx}
+              title={title}
+              to={to}
+              date={date}
+              description={formattedDetails}
+              location={location}
+              idx={idx}
```
wackerow, line 116: If we go with the approach on `EventCard` of passing `isEven` as a prop, you could calculate that here, naming it the same to make it easy to track, and then you could remove needing to declare this as a constant in the EventCard component. 

 ```suggestion
              isEven={idx % 2 === 0}
```

wackerow (changes requested):
@vluna Very cool! Thanks for your work on this so far. 

Left a couple comments around trying to minimize code redundancy. Also, getting some rough results on mobile sizes. We'll need to fix this up to be mobile responsive. 

image

I would suggest this switches to just a vertical string of boxes with vertical lines in the middle connecting them. Open to thoughts from others though.

vluna:
> @vluna Very cool! Thanks for your work on this so far.
> 
> Left a couple comments around trying to minimize code redundancy. Also, getting some rough results on mobile sizes. We'll need to fix this up to be mobile responsive.
> 
> image
> 
> I would suggest this switches to just a vertical string of boxes with vertical lines in the middle connecting them. Open to thoughts from others though.

Thanks for pointing that out. I guess the responsiveness slipped my mind.

src/components/EventCard.js
vluna: Nice! Thanks for the suggestions, will work on them 🙂

vluna:
@wackerow Thanks for the feedback and suggestions. I have updated the events page to be mobile friendly.

@minimalsm Let me know what you think 🙂

minimalsm:
Hey @vluna,

Thanks a ton for this. Apologies for the lack of image reference on this. I think this could use some design tweaking but definitely beyond our capacity until we've got a designer on staff—you've done an awesome job here. Merging now ❤️❤️❤️❤️❤️❤️❤️❤️

minimalsm:
Just waiting on build after resolving merge conflicts here :-)

minimalsm (approved):

PR #5158, opened 2022-01-19
Fix: svg hover bug on WalletCard

wackerow:
## Description
Fixes glitch on WalletCard component when hovering over a flipped card. 

## Related Issue
None filed. 
Can reproduce by going here on desktop: https://ethereum.org/en/wallets/find-wallet/
Then flipping any card over and hovering over that card. 
image


minimalsm (approved):
Nice catch @wackerow

PR #5161, opened 2022-01-21
Fix Uniswap grants link

samajammin:
## Description

Fix dead link

## Related Issue
None


wackerow (approved):
Just added the same fix for the intl versions. Kept the label as just "Uniswap" for these to avoid translation issues before the markdown is officially translated.

PR #5168, opened 2022-01-22
Run-a-node layout updates after review

wackerow:
`run-a-node <- run-a-node-rc`

Figma prototype: https://www.figma.com/proto/Fsuy1ZcRm262sBUzdMSkn7/Run-A-Node?page-id=0%3A1&node-id=2096%3A635&viewport=241%2C48%2C0.38&scaling=min-zoom&starting-point-node-id=217%3A183

- Updates hero CTA labels
  - first linking to "What does it mean to run a node?" section immediately beneath (no longer skipping content)
  - second button for those ready to take action, scrolling user to "Getting started"
- Forces the "Who should run a node?" card to be open
  - Updates the new "ExpandableInfo" component to accept a `forceOpen` prop which hides the clickable chevron, and loads in expanded state.
- Moves highlights that introduce DAppNode to beginning of "Getting started" section, which immediately follows the "What, Who, Why" education sections
  - Introduces DAppNode directly in second box, removing comment about "Now, more user-friendly graphical user interfaces (GUIs) exist" which implied we were about to list multiple options.
  - Updated "phone tapping" glyph that goes with the now third card.
- Next is "Choose your adventure," where copy about needing some basic hardware starts it off, then giving user the option of buying fully loaded, or building own.
  - Fully loaded has two links to external shops
  - Build your own has secondary button for "Start building" that scrolls user down to next section
- Next is "Build your own" section, broken into two boxes:
  - Hardware: Same recs as before with new icon (open to a better one)
  - Software: Linking out to DAppNode Setup instructions
- "Find some helpers" section follows this up, along with "Further reading"
- Staking section unchanged, but the "Plan on staking?" and "A note on Raspberry Pis" sections are now located below the staking callout. Links _to_ these are still mentioned and located above, but this consolidates all staking content in one place to minimize confusion for those not staking.

- [x] Mobile responsive
- [ ] Extract strings for intl (can wait til merged into `run-a-node`)

@samajammin Have a look whenever you get a chance, but this should address everything we chatted about in some form. Made it as a separate branch to leave the other as-is in the meantime.


samajammin:
I like the structure change of "Choose your adventure"

Image 2022-01-24 at 3 35 19 PM

BUT I think we should keep the same 50/50 layout, like you had previously:
Image 2022-01-24 at 3 36 38 PM

I think having both with the same card layout makes it clear these are the 2 options (adventures) we're suggesting.

samajammin:
Potential content for the "build your own" card:
Image 2022-01-24 at 3 38 11 PM

- I don't think "good" is a strong adjective - perhaps just "A cheaper option for slightly more technical users"
- Similar to the bullet points under the "buy fully loaded" I think we could come up with a few, e.g.
  - More customizable setup
  - Greater flexibility with node software (DAppNode doesn't support all client installations, correct?)

samajammin:
Image 2022-01-24 at 3 44 43 PM

- "part 3" doesn't really make sense here anymore
- Again feels weird reference "Discord", seems like we're saying the company vs. any particular Ethereum community servers
- Are there any more servers we can add here, e.g. EthStaker? Or is that too staking-focused?

samajammin (changes requested):
A few suggestions - otherwise looks awesome!

samajammin (approved):
🔥

Issue #4447, opened 2021-11-11
Wallet Listing: Numio


mitdralla:
**Is your wallet security tested? Please explain security measures i.e. security audit, internal security team or some other method.**

The app is constantly subject to multiple layers of QA testing before (and after) undergoing internal audits and penetration tests. Major releases undergo 3rd party security auditing.

Numio is a non-custodial mobile application focused on user security and UI/UX. A user's key pair is generated locally on their phone and their private keys never leave their device. Users sign their own transactions manually and only the encrypted data is transmitted.

**When did your wallet go live to users?**

Android went live on Google Play on the 31st December 2020 and iOS went live in the App Store on the 6th July 2021.

**Does your wallet have an active development team?**

Yes. The wallet is actively developed and new releases go out on average every 10 days.

**Is your wallet open-source?**

No. Numio is not open source, however elements of the application are.

Open source elements include our CDN & SDK, zkSync API, NPM relayer, and various meta API’s.

Our full open-source GitHub repo can be found here - https://github.com/TeamNumio/TeamNumio

**Is your wallet globally accessible?**

Numio is globally accessible via the Google Play and Apple App Stores. The app requires no KYC to use.

**Is your wallet custodial, non-custodial, or a hardware wallet?**

Numio is non-custodial - only the user has access to their Private Keys, which are easily accessible via the settings menu.

**Please describe the measures taken to ensure the wallet's security and provide documentation wherever possible**

As stated in the first response, the app is constantly subject to multiple layers of QA testing before (and after) undergoing internal audits and penetration tests. Major releases undergo 3rd party security auditing. While we do not currently have any of this audit data currently publicly available, we are looking to make something available in the near future.

We work closely with our partners, including zkSync and Storj, to ensure that their technology is integrated in accordance with their best practices. This can be verified with those teams.

Device permissions are only requested based on a user's choice of function and if essential to the operation of the application. Any permissions are at the minimum viable amount to ensure that a function works. Numio keeps up to date with all Google Play and App Store privacy requirements and describes, via custom modal and plain English, exactly what each permission is for. 

By design, users are unable to capture the screen during the seed phrase screen and (optional) biometrics registrations screens to enhance users security. Seed phrases are not stored on the device.

Outside of secure blockchain technology, the Numio application uses SHA256/RSA256 for encryption both within the app, and as a byproduct of integrated technology, including FaceTech - https://dev.facetec.com/security-best-practices.

[...]

PR #5219, opened 2022-01-25
Update Numio Wallet brand colour and swaps status

milvinae:
## Description

Change brand_color from #FFFFFF to # 3F3F3F so the background is on brand.

Change has_dex_integrations to TRUE as Token Swaps are live on Numio.

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/4447


minimalsm (approved):
Thanks @milvinae,

Updated on our end also.

PR #5226, opened 2022-01-26
Run-a-node updates

wackerow:
`run-a-node <- run-a-node-rc2`

## Description
- Adds missing card hover to two cards with buttons
- Updates "what is a node" copy to includes note about verifying the validity of blocks
- Adjusts copy for fully-loaded/plug-n-play section
- Extracts English strings to intl jsons
- Merges with `dev` and clears conflicts
- Updates to match paths and naming after eth2 rename
- Extracts English from FeedbackCard.js

cc: @samajammin


samajammin (approved):
👍

PR #5235, opened 2022-01-26
Identify additional areas to reference blog post

corwintines:
## Description
- Link out to renaming blogpost in side banner on upgrades template pages
- Add Eth1 and Eth1 terms to glossary and link out to renaming blogpost
- Add a banner to /upgrades/ and /staking/ pages that links out to renaming blogpost


src/components/Layout.js
```diff
@@ -93,6 +97,11 @@ const Layout = (props) => {
         document.getElementById(idTag[1]).scrollIntoView(false)
       }
     }
+
+    // TODO: Remove Feb 9 2022
+    props.path.includes("/upgrades/") || props.path.includes("/staking/")
```
samajammin, line 102: Why not just set a variable here vs. component state?

wackerow: Yeah, I'd say if it's not dismissible, we don't really need to track any state. I can patch this.

corwintines: Makes sense. Thanks for addressing @wackerow.

samajammin (approved):

PR #5243, opened 2022-01-27
Changed Tenderly description in smart-contracts deploying docs

Ignjatovic:
## Changed the description for Tenderly to better reflect the current state of the product

Changed desc and added the link to Tenderly docs page.

### DIFF
--- a/src/content/developers/docs/smart-contracts/deploying/index.md
+++ b/src/content/developers/docs/smart-contracts/deploying/index.md
@@ -38,9 +38,10 @@ Once deployed, your contract will have an Ethereum address like other [accounts]
 
 - [Remix](https://remix.ethereum.org)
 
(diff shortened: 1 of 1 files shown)


minimalsm:
@all-contributors please add @Ignjatovic for docs

Ignjatovic:
@minimalsm glad to contribute, it ain't much but it's honest work :D But for real, I would like to post some more sensible materials as we produce them (at Tenderly), since they will be oriented towards solving actual issues in the space and proposing actual solutions, stuff like this - https://blog.tenderly.co/mempool-simulated/. This article is shill(y) of course since it's on our blog and it's not really a proper use case example, but you get the idea (this would be great for the contract testing segment imho). So I wanted to consult a bit on what you guys think.

PR #5097, opened 2022-01-14
Cleanup console warnings during build

wackerow:
## Description
Fixes the following console warnings received when building site:

```log
/src/components/Layout.js 96:6
warning  React Hook useEffect has a missing dependency: 'props.location.hash'.
```
- Adds location to dependency array

```log
/src/components/Morpher.js 152:6
warning  React Hook useEffect has missing dependencies: 'morphTimeout', 'morpher', 'state.text', and 'state.words'.
```
- Investigated in the past with no threat here. Warning suppressed

```log
/src/components/YouTube.js 24:7
warning  <iframe> elements must have a unique title property
```
- Added `title` property to our one use of iframe with default text of "YouTube"
- Added optional `title` prop to YouTube component

```log
/src/pages/developers/learning-tools.js 87:6
warning  React Hook useEffect has a missing dependency: 'sandboxes'.
```
- Fixed useEffect for handling sandboxes array

```log
warn There are routes that match both page and redirect. Pages take precendence over redirects so the redirect will not work:
  - page: "/" and redirect: "/" -> "/en/"
  - page: "/eth2/" and redirect: "/eth2/" -> "/en/eth2"
  - page: "/community/" and redirect: "/community/" -> "/en/community/"
  - page: "/developers/" and redirect: "/developers/" -> "/en/developers/"
  - page: "/languages/" and redirect: "/languages/" -> "/en/languages/"
 - page: "/upgrades/" and redirect: "/upgrades/" -> "/en/upgrades/"
 - page: "/staking/" and redirect: "/staking/" -> "/en/staking/"
```
- Removed these redirects from `redirects.json`

_This does not address warnings related to gatsby plugins._

PR #5251, opened 2022-01-27
Fix redirect bug on page loads without lang

corwintines:
## Description
When going to a page without a lang the site gets in a redirect loop and never loads. Added back in redirects removed in https://github.com/ethereum/ethereum-org-website/pull/5097/files to resolve

### DIFF
--- a/src/components/Layout.js
+++ b/src/components/Layout.js
@@ -76,26 +76,26 @@ const Layout = (props) => {
     }
   }, [])
 
-  // useEffect(() => {
-  //   if (props.path.includes("/docs/")) {
-  //     setShouldShowSideNav(true)
+  useEffect(() => {
+    if (props.path.includes("/docs/")) {
+      setShouldShowSideNav(true)
 
-  //     if (localStorage.getItem("zen-mode") !== null) {
-  //       setIsZenMode(localStorage.getItem("zen-mode") === "true" && !isMobile())
-  //     }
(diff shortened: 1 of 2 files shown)


wackerow:
Getting the same homepage bug on Brave mobile (iPhone) with the preview link for this PR
https://ethereumorgwebsitedev01-redirectissue.gtsb.io/

Suggests to me there may be a deeper problem than just these redirect

src/components/Layout.js
```diff
@@ -76,26 +76,26 @@ const Layout = (props) => {
...
-    }
-
-    if (props.location.hash && !props.location.hash.includes("gatsby")) {
-      const idTag = props.location.hash.split("#")
-      if (document.getElementById(idTag[1]) !== null) {
-        document.getElementById(idTag[1]).scrollIntoView(false)
-      }
-    }
-  }, [props.path, props.location])
+  // useEffect(() => {
```
samajammin, line 79: Why comment all this out?

samajammin (changes requested):
Redirects look good but let's undo that commented code.

wackerow:
So I'm still getting the redirect when I try this on mobile, both Brave and Safari with private tabs. This is a Gatsby link though, and our main concern is fixing prod, which is Netlify, so I'll bring this in to see if it fixes it on Netlify side.

PR #5262, opened 2022-01-28
Update proof of stake in style-guide

anderselowsson:

### DIFF
--- a/src/content/contributing/style-guide/index.md
+++ b/src/content/contributing/style-guide/index.md
@@ -130,22 +130,25 @@ When referring to the Ethereum Mainnet (i.e. not referring to a testnet) use the
 - Main net
 - Ethereum mainnet
 
-### Proof-of-work / Proof-of-stake {#proof-of}
+### Proof of stake {#proof-of}
 
-Proof-of-work should be capitalized at the beginning of a sentence. In any other instance, all letters should be lower case. In either case, proof-of-work should be hyphenated between each word.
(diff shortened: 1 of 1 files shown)


minimalsm:
_Posting my comment from Discord here for transparency_.

Hey @anderselowsson thanks for the PR ☺️ curious for other opinions but we discussed this a while ago and decided to treat proof-of-stake as a trigram.

Technically we could leave proof-of-stake unhyphenated when it stands alone, but the grammar decision we landed on is still sound, so we decided to hyphenate in 100% of cases.

The team agreed that this would improve the clarity of our docs for readers (many of whom are not native English speakers).
Another thing to consider here is that we have hundreds of open-source contributors who submit changes to our site. We don't want to 1) spend a lot of time thinking about grammar and 2) have contributors playing proof-of-stake > proof of stake ping pong 🏓(e.g. we've had four or five PRs to change 'Merkle trie' to 'Merkle tree' 😅)

But to be clear: the ethereum.org style-guide obviously doesn't speak for the Ethereum ecosystem :-).

But if our specific style decisions can push consistency, it's a win I'll happily take 😀 (a point I raised with EIP editors a few months back... IMO the way we talk about Ethereum specific terms should be standardized)

minimalsm:
Thanks again for the PR @anderselowsson,

We've discussed this again and agreed that we'd like to keep this as is so as not to introduce additional overhead in maintenance. Closing this out for now.

anderselowsson:
No problem!

PR #5306, opened 2022-02-02
Remove public from solidity constructor

jixbo:
Visibility for constructors is not needed anymore, and hardhat throws a warning when public is used.
https://docs.soliditylang.org/en/v0.7.0/070-breaking-changes.html#functions-and-events


minimalsm (approved):
Thanks @jixbo this looks good to me.

Waiting on someone better versed with Solidity to weigh in before merging :-)

samajammin (approved):
Thanks!

Issue #5273, opened 2022-01-31
Skip link scrolls too far down the page


minimalsm:
**Describe the bug**

Navigating to the main content ID (#main-content) using the skip-link takes me to the 'Stake your ETH' section (well lower in the page than I should be taken to). This bug is persistent across all pages I tested.

**Other pages tested:**
- DeFi 
- governance
- Developer docs, 
- Developer tutorials

**To Reproduce**

1. Go to 'https://ethereum.org/en/run-a-node/'
2. Hit tab to see skip link
3. Hit enter to visit main content

**or**

Visit https://ethereum.org/en/run-a-node/#main-content

**Expected behavior**

A skip link should take the user to the first meaningful content after the navigation bar.

**Desktop (please complete the following information):**

- OS: macOS Monteray 
- Browser: Brave Browser
- Version: 1.34.80

**Additional context**

The main-content ID is on the correct element.

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

vluna:
@minimalsm I am not 100% sure about the solution for this, but after looking for a couple of days online I think I found a solution.

vluna:
@minimalsm I have created the PR, I am sure maybe this is not what is expected but thought I will give it a try.

PR #5308, opened 2022-02-02
Update skip to main content component

vluna:
## Description
- Update `SkipLink` component to support `SkipLinkAnchor` to set anchor for the link
- Update to scroll to on top of main content

## Related Issue
#5273

### DIFF
--- a/src/components/Layout.js
+++ b/src/components/Layout.js
@@ -23,7 +23,7 @@ import { useKeyPress } from "../hooks/useKeyPress"
 
 import { isLangRightToLeft } from "../utils/translations"
 import { isMobile } from "../utils/isMobile"
-import SkipLink from "./SkipLink"
+import { SkipLink, SkipLinkAnchor } from "./SkipLink"
 
 const ContentContainer = styled.div`
   position: relative;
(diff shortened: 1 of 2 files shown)


minimalsm:
Thanks @vluna,

I'm getting the same issue with this as I was in #5273. (Brave Version 1.34.80)

Can I check what browser you're on?

vluna:
@minimalsm I tested it in:
- Chrome: 97.0.4692.99
- Brave: 1.35.100 Chromium: 98.0.4758.87

minimalsm:
Could you try on a mobile view? This seems to work at certain resolutions for me @vluna

vluna:
> Could you try on a mobile view? This seems to work at certain resolutions for me @vluna

@minimalsm could you provide a specific resolution? I tried different ones and it seems to work just fine. Could you maybe take a video or a way to show where it goes wrong? Now, I am just wondering maybe I did not do it correctly or I got confused in to what it is required. The current behaviour in the PR is the following:

Steps:
- When user goes to `/run-a-node` page, and presses `tab` on the keyboard
- The focus should go to the content (currently there is no visual of focus as the id is above the actual content as a workaround)
- Then, the user presses `tab` again and the focus should be in `Learn More` button after the short description

Is this correct?

minimalsm (approved):
Weird. I've tested this again and it's now working for me... 

Think this is good to go but would appreciate another set of eyes to check this before we merge @corwintines @wackerow

pettinarip (approved):
LGTM! thanks @vluna

PR #5333, opened 2022-02-03
Added Mint glossary term

RahulBansal123:
## Description

Added a new glossary term: Mint

### DIFF
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -532,7 +532,7 @@ A network [node](#node) that finds valid [proof-of-work](#pow) for new blocks, b
 
 ### Mint {#mint}
 
-Minting is the process of creating new coins and bringing them into circulation so that they can be traded. It's a decentralised mechanism to create a new token without the involvement of the central authority.
+Minting is the process of creating new tokens and bringing them into circulation so that they can be traded. It's a decentralized mechanism to create a new token without the involvement of the central authority.
 
 <Divider />
 
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -532,7 +532,7 @@ A network [node](#node) that finds valid [proof-of-work](#pow) for new blocks, b
 
 ### Mint {#mint}
 
-It refers to the process of issuing new coins/tokens.
+Minting is the process of creating new coins and bringing them into circulation so that they can be traded. It's a decentralised mechanism to create a new token without the involvement of the central authority.
 
 <Divider />
 
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -530,6 +530,10 @@ A network [node](#node) that finds valid [proof-of-work](#pow) for new blocks, b
   Mining
 </DocLink>
 
+### Mint {#mint}
+
+It refers to the process of issuing new coins/tokens.
+
 <Divider />
 
 ## N {#section-n}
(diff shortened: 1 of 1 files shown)


src/content/glossary/index.md
```diff
@@ -530,6 +530,10 @@ A network [node](#node) that finds valid [proof-of-work](#pow) for new blocks, b
   Mining
 </DocLink>
 
+### Mint {#mint}
+
+It refers to the process of issuing new coins/tokens.
```
minimalsm, line 535: I think we need a better, more detailed description here. 

Issuing isn't the right word here IMO. Minting is explicitly **creating** new currency that didn't previously exist.

minimalsm (changes requested):
Hey @RahulBansal123, thanks for the suggestion and PR ❤️❤️❤️

We'd like to add this to the glossary. As per my comments below, I think we need to improve this definition first. 

I think the next step here would be looking at other, existing definitions for inspiration. [Here is one example from Coinmarketcap.](https://coinmarketcap.com/alexandria/glossary)

src/content/glossary/index.md
```diff
@@ -530,6 +530,10 @@ A network [node](#node) that finds valid [proof-of-work](#pow) for new blocks, b
   Mining
 </DocLink>
 
+### Mint {#mint}
+
+Minting is the process of creating new coins and bringing them into circulation so that they can be traded. It's a decentralised mechanism to create a new token without the involvement of the central authority.
```
minimalsm, line 535:
```suggestion
Minting is the process of creating new tokens and bringing them into circulation so that they can be traded. It's a decentralised mechanism to create a new token without the involvement of a central authority.
```

minimalsm: @samajammin any thoughts on saying coins/tokens here (or both). I think token catches everything and coins creates a weird mental model.

minimalsm: Also, is it important that we mention trading tokens here? Is that their purpose? Not sure myself... many tokens aren't explicitly designed to have trading value (e.g. governance tokens)

RahulBansal123: @minimalsm I think we should use both coins and tokens.

samajammin: Not a string opinion but I think "token" makes sense, as minting is most relates to token contracts (i.e. ERC-20 or ERC-721 contracts).

wackerow: I don't have a very strong opinion here either, but I honestly feel "coin" is the more awkward option given how we often talk about minting tokens of NFTs.

I think given the context it's simplest and makes most sense to just go with "token." It fits the idea of the "token contracts" we're familiar with (not "coin contracts")

```suggestion
Minting is the process of creating new tokens and bringing them into circulation so that they can be traded. It's a decentralized mechanism to create a new token without the involvement of the central authority.
```
(Also, small American English tweak per style guide)

src/content/glossary/index.md
```diff
@@ -530,6 +530,10 @@ A network [node](#node) that finds valid [proof-of-work](#pow) for new blocks, b
   Mining
 </DocLink>
 
+### Mint {#mint}
+
+Minting is the process of creating new tokens and bringing them into circulation so that they can be traded. It's a decentralized mechanism to create a new token without the involvement of the central authority.
```
minimalsm, line 535:
```suggestion
Minting is the process of creating new tokens and bringing them into circulation so that they can be used. It's a decentralized mechanism to create a new token without the involvement of the central authority.
```

minimalsm (approved):
Thanks again @RahulBansal123 

Be sure to claim your [2022 POAPs](https://ethereum.org/en/contributing/#poap) also 🥳!

minimalsm:
@all-contributors please add @RahulBansal123 for docs, content

PR #5348, opened 2022-02-06
fix: Use correct terminology to refer to the EVM state

Sabnock01:
## Description
Changes terminology from "EVM storage" and "EVM memory" to "EVM state" to reflect a more accurate description of smart contracts' place in the EVM.

### DIFF
--- a/src/content/developers/docs/intro-to-ethereum/index.md
+++ b/src/content/developers/docs/intro-to-ethereum/index.md
@@ -49,7 +49,7 @@ The amount of ether paid corresponds to the time required to do the computation.
 
 ## What are smart contracts? {#what-are-smart-contracts}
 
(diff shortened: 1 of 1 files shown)


Sabnock01:
@minimalsm I would think so. The particular article also uses the term "state" with respect to the EVM several times prior so there is some logic to it. The original motivation for the change was my question in the Discord on whether smart contracts were stored in "memory" or "storage" since both were mentioned in the article. Micah Zoltu suggested the correct term probably ought to be "state."

gichiba:
Makes sense to me! I think the term 'state' is general enough to be understood by any reader who is digesting everything else in this article

minimalsm (approved):
Sorry for the delay here (I was on vacation)

I've gathered a few opinions to double-check and it seems everyone agrees! Thanks again for the contribution @Sabnock01  ✌️

Be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And remember to [claim your POAP](https://ethereum.org/en/contributing/#poap) also 🥳! We're currently working on our 2022 POAPs and they'll be released soon.

Issue #2585, opened 2021-03-09
Absolute reference for both Intl and English MDX images


samajammin:
**Describe the bug**

We recently added our first translations of tutorials (#2531) 🎉  but we are not properly referencing assets.

**To Reproduce**

1. Go to https://ethereum.org/ro/developers/tutorials/getting-started-with-ethereum-development-using-alchemy/
2. See 404s in console from failed image fetches

**Expected behavior**

Images should work!

**Desktop (please complete the following information):**
All

**Smartphone (please complete the following information):**
All

**Additional context**

A couple approaches I see here, neither which I like
- A) Copy over the images from English tutorial directories into translations
- B) Update the relative path of image files in translation tutorials to point to English images
- C) Refactor our file structure to have all markdown in the same directory

I think a better solution might be to 
- D) Find a plugin / configure a way to have all relative image references in markdown files start from the same absolute reference, e.g. so anywhere in the project we could use
```
![Some image alt text](./some-image.png)
``` 
Then our project would know that actually references `/src/assets/some-image.png`. I'm not sure how to accomplish that yet but that would save us time from manually copying files / updating relative image references.

More context on C) approach here:
https://github.com/gatsbyjs/gatsby/discussions/30244#discussioncomment-482912

samajammin:
Offline discussion: approach C) seems to be our best bet here.

minimalsm:
@samajammin 

Until this is solved I believe we're blocked on the CrowdIn to Github automation?

samajammin:
Depends on the implementation but I don't think this should impact the GH <> Crowdin integration at all. Particularly if we can implement option D.

samajammin:
For the record we are currently go with approach B (manually updated image paths of translated files) to solve this (which is NOT ideal).

pettinarip:
Might be worth to look at https://www.gatsbyjs.com/plugins/gatsby-plugin-module-resolver/ where we could start by the basic resolution of `src`:

```ts
module.exports = {
  plugins: [
    {
      resolve: 'gatsby-plugin-module-resolver',
      options: {
        root: './src',
      }
    }
  ]
}
```

Note: I don't know if this is going to work from inside the md files, this is something to test.

corwintines:
While we dont do absolute paths, we have a script made to resolve these issues

PR #5355, opened 2022-02-07
Local plugin to fix translated mdx images urls

pettinarip:
## Description
The following proposal tries to solve having different images urls between source and translated md files. With this, we won't need to adjust the images urls in translated files (we wil keep the same as in the source file).

## Solution
By using a local gatsby plugin we can intercept the nodes created by `gatsby-plugin-mdx` and adjust the images urls. We can think about it as a "translator" for urls.
- If we are in a source file => the plugin will ignore it
- If we are in a translated file => the plugin will modify the url by pointing them to the source file folder (what we were doing manually)

Check working example https://ethereumorgwebsitedev01-mdximagespaths.gtsb.io/ro/developers/tutorials/getting-started-with-ethereum-development-using-alchemy/

## Related Issue
#2585 

## TODO
- [x] Change variable names that contains `path` with `url`
- [x] Add more docs

### DIFF
--- a/src/content/translations/ro/developers/tutorials/getting-started-with-ethereum-development-using-alchemy/index.md
+++ b/src/content/translations/ro/developers/tutorials/getting-started-with-ethereum-development-using-alchemy/index.md
@@ -19,7 +19,7 @@ source: Mediu
 sourceUrl: https://medium.com/alchemy-api/getting-started-with-ethereum-development-using-alchemy-c3d6a45c567f
 ---
 
-![Logouri Ethereum și Alchemy](../../../../../developers/tutorials/getting-started-with-ethereum-development-using-alchemy/ethereum-alchemy.png)
+![Logouri Ethereum și Alchemy](./ethereum-alchemy.png)
 
 Acesta este un ghid pentru începători pentru a demara cu dezvoltarea Ethereum folosind [Alchemy](https://alchemyapi.io/), cea mai importantă platformă de programator blockchain alimentând milioane de utilizatori din 70% din aplicațiile blockchain de top, inclusiv Maker, 0x, MyEtherWallet, Dharma și Kyber.
 
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -127,6 +127,10 @@ module.exports = {
         // See: https://www.gatsbyjs.org/docs/mdx/plugins/
         gatsbyRemarkPlugins: [
           {
+            // Local plugin to adjust the images urls of the translated md files
+            resolve: require.resolve(`./plugins/gatsby-remark-image-urls`),
+          },
+          {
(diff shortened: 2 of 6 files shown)


src/content/translations/ro/developers/tutorials/getting-started-with-ethereum-development-using-alchemy/index.md
```diff
@@ -19,7 +19,7 @@ source: Mediu
 sourceUrl: https://medium.com/alchemy-api/getting-started-with-ethereum-development-using-alchemy-c3d6a45c567f
 ---
 
-![Logouri Ethereum și Alchemy](../../../../../developers/tutorials/getting-started-with-ethereum-development-using-alchemy/ethereum-alchemy.png)
+![Logouri Ethereum și Alchemy](./ethereum-alchemy.png)
```
pettinarip, line 22: The plugin will basically translate `./ethereum-alchemy.png` into `../../../../../developers/tutorials/getting-started-with-ethereum-development-using-alchemy/ethereum-alchemy.png`

wackerow (approved):
@pettinarip This is great! Nice and clean.

This appears to be backward compatible, so the existing images are not breaking, which is great. I think we bring this in and separately we can make an issue to clean up all the existing images.

Last think would just be to double check w/ @lukassim if we need to re-upload any of these to Crowdin. The files should have the short relative paths by default, but just want to double check we didn't manually overwrite any of those paths. In either case, we should be good here! 👍

pettinarip:
> I think we bring this in and separately we can make an issue to clean up all the existing images.

@wackerow this approach allows us to avoid touching the current files. We can keep everything as it is.

My idea was that starting from now, the files will progressively update themselves to the short relative path. We shouldn't necessarily do any new PR to manually update the files.

Maybe is something we can discuss as a team:
- Do we prefer to update all the translated files at once to start using the same images urls that the source files have?
- Or do we prefer to keep everything as it is and while we upload newly translated files the urls are going to be updating themselves?

corwintines (approved):
Looks great, thanks for linking out to gatsby stuff for context :)

minimalsm:
> The files should have the short relative paths by default, but just want to double check we didn't manually overwrite any of those paths. In either case, we should be good here! 👍

@wackerow I've manually overwritten all of these paths (so when you download from Crowdin you get the previously correct path (e.g. `../../../../../developers/tutorials/getting-started-with-ethereum-development-using-alchemy/ethereum-alchemy.png`)

But, this is a relatively easy fix that requires 0 review from Acolad :-) the main concern would be if there are other changes to the source content that might set back progress in the markdown files with images.

corwintines:
Merging based on conversation in standup today

PR #5370, opened 2022-02-08
Update Alchemy link in NFT tutorial

btandowsky:
Cleaning up the link to Alchemy's NFT API since it previously went to a specific section that wasn't the top


minimalsm (approved):
Thanks as always @btandowsky ❤️❤️❤️❤️❤️❤️

PR #5397, opened 2022-02-11
add "checkpoint" to glossary

jmcook1186:
## Description
Last pr from me in this bundle...!

adding "checkpoint" to glossary with definition:

"The [Beacon Chain](#beacon-chain) has a tempo divided into slots (12 seconds) and epochs (32 slots). The first slot in each epoch is a checkpoint. When a [supermajority](#supermajority) of validators attests to the link between two checkpoints, they can be [justified](#justification) and then when another checkpoint is justified on top, they can be [finalized](#finality)."

## Related Issue
n/a


minimalsm (approved):
Thanks Joseph!

PR #5413, opened 2022-02-13
Simplify cpubenchmark link

pettinarip:
## Description
This is a proposal to make the section a bit simpler. Just felt redundant having the 2 same urls in the same space.

## Problem
image

@wackerow let me know what do you think? perhaps you have some reason why we have done it that way and I'm not considering it


wackerow (approved):
@pettinarip  Only reason I broke it out was to keep it cleaner in the translated strings, and partially by habit since we used to be limited here. Not at all against doing it this way.

corwintines (approved):

PR #5424, opened 2022-02-14
Extracted missed intl strings

wackerow:
## Description
Extracts two additional strings from run-a-node page that were missed for intl

## Related Issue
None filed


corwintines (approved):
LGTM

PR #5428, opened 2022-02-15
Update index.md

licitrasimone:
I have inserted an article in my opinion very important on the future of ethereum with the addition of eth2.


KeKo6988 (approved):

corwintines (changes requested):
@licitrasimone I took at the article you linked out to here. I think we have better resources on the website already, but they could certainly be linked out to from here. What do you think about linking to https://ethereum.org/en/history/#the-history-of-ethereum which addresses the arrow glacier part of this article, and https://ethereum.org/en/upgrades/ when talking about the roadmap. It will be more up to date as time progresses, and will be more up to date as the roadmap changes this way as well.

corwintines:
I realize on second look that this is for the community page. My mistake. I think this page is too specific to be listed here. We link out to this website just above this, so going to close out this PR.

PR #5430, opened 2022-02-16
update index.md

seekasra:
The definition of blockchain is modified in the document, to remove confusion. Since the blockchain is an immutable database its best to use the word accumulate instead of update.


corwintines:
I think updated is still the most appropriate term as when a new block is added the state is being updated. If you disagree feel free to open this back up and we can debate this a bit more.

PR #5470, opened 2022-02-22
Solidity is not "heavily inspired by JavaScript"

Arachnid:

### DIFF
--- a/src/content/developers/docs/programming-languages/javascript/index.md
+++ b/src/content/developers/docs/programming-languages/javascript/index.md
@@ -24,7 +24,7 @@ You can use these libraries to interact with smart contracts on Ethereum so it's
 
 ### Smart contracts {#smart-contracts}
 
-If you're a JavaScript developer and wanting to write your own smart contract, you may want to get familiar with [Solidity](https://solidity.readthedocs.io). This is the most popular smart contract language and it's heavily inspired by JavaScript.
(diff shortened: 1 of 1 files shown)


franzihei:
Yes, absolutely agree with this change. I didn't know there were other places that talk about Solidity, otherwise I would have suggested this change myself! :) See https://ethereum.org/en/developers/docs/smart-contracts/languages/ where we said "most profoundly influenced by C++" - but of course Solidity takes concepts from other languages as well. Generally one can say that Solidity will feel familiar for anybody who has experience with curly-bracket languages (https://en.wikipedia.org/wiki/List_of_programming_languages_by_type#Curly-bracket_languages).

franzihei:
This really makes me wonder why people still believe Solidity is inspired by JavaScript... Btw we wrote a section on language influences [here in the docs](https://docs.soliditylang.org/en/latest/language-influences.html).

minimalsm (approved):
Thanks for the PR @Arachnid and thanks for confirming @franzihei. I feel like I see this everywhere (possibly the fault of the ethereum.org docs). I'll keep a lookout for anywhere else on the site we discuss this and fix those cases also.

minimalsm:
@Arachnid you're also able to [claim a 2022 contributor POAP](https://ethereum.org/en/contributing/#poap) for this 🥳!

minimalsm:
@all-contributors please add @Arachnid for docs

PR #5478, opened 2022-02-22
Add a side-note as to why ether is burnt

m-zaink:
While reading the documentation, I realized that ether burn was explained but it wasn't clarified as to why it is done.
I understood why minting is required but why to burn, I was clueless.

I have added a side-note under {#burning-ether} section regarding the same.

I'm hoping this will add more clarity to the documentation.

### DIFF
--- a/src/content/developers/docs/intro-to-ether/index.md
+++ b/src/content/developers/docs/intro-to-ether/index.md
@@ -41,6 +41,10 @@ As well as creating ether through block rewards, ether can get destroyed by a pr
 
 Ether burn occurs in every transaction on Ethereum. When users pay for their transactions, their base gas fee gets destroyed by the protocol. Depending on network demand, [some blocks](https://etherscan.io/block/12965263) burn more ether than they mint.
 
+> Ether burn is essentially a mechanism to control the supply & demand dynamics. 
+> 
+> By burning a certain amount of ether in every transaction, total amount of ether in circulation is controlled, thereby hedging it against such devils as inflation.
+
 ## Denominations of ether {#denominations}
 
 Since many transactions on Ethereum are small, ether has several denominations which may be referenced for smaller amounts. Of these denominations, Wei and gwei are particularly important.
--- a/src/content/developers/docs/intro-to-ether/index.md
+++ b/src/content/developers/docs/intro-to-ether/index.md
@@ -39,7 +39,9 @@ Ether is minted when a miner creates a block on the Ethereum blockchain. As an i
 
 As well as creating ether through block rewards, ether can get destroyed by a process called 'burning'. When ether gets burned, it gets removed from circulation permanently.
 
-Ether burn occurs in every transaction on Ethereum. When users pay for their transactions, their base gas fee gets destroyed by the protocol. Depending on network demand, [some blocks](https://etherscan.io/block/12965263) burn more ether than they mint.
+Ether burn occurs in every transaction on Ethereum. When users pay for their transactions, a base gas fee, set by the network according to transactional demand, gets destroyed. This, coupled with variable block sizes and a maximum gas fee, simplifies transaction fee estimation on Ethereum. When network demand is high, [blocks](https://etherscan.io/block/12965263) can burn more ether than they mint, effectively offsetting ether issuance.
+
+Burning the base fee prevents various ways in which miners could affect its level. For example, if miners got the base fee, they could include their own transactions for free and raise the base fee for everyone else. Alternatively, they could refund the base fee to some users off-chain, leading to a more opaque and complex transaction fee market.
 
 
 ## Denominations of ether {#denominations}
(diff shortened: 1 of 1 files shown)


m-zaink:
Totally agree. Your one-liner is a good suggestion @minimalsm. 

I'm rephrasing it as follows -

> Ether is burned to prevent inflation by limiting its overall supply.

Does this sound fine? Just trying to keep the tone of the entire article similar.

minimalsm:
> Ether is burned to prevent inflation by limiting its overall supply.

@m-zaink I think I'd use the word 'reducing' instead of 'limiting' here. Ether burn does not **limit** supply, per se.

m-zaink:
But think about it, the total amount of ether doesn't reduce per se.
It just gets thrown onto a bin account from which this ether can never be recovered.

So with that in mind, I think _limit_ is more appropriate.

minimalsm:
Bitcoin is limited to a supply of 21 million bitcoin. Ether has no such limit. The word limit here creates a confusing mental model. 

> But think about it, the total amount of ether doesn't reduce per see.

But the total supply of ether does reduce. If it is no longer accessible then the supply has been reduced.

m-zaink:
Fair enough @minimalsm.

I'll update this PR accordingly.

wackerow:
Hey @m-zaink, thanks for the PR here.

My initial thoughts on this are that the _reason_ ETH is burned is _not_ just because it must have a limited supply. Would be curious what @timbeiko has to say here, but EIP-1559 started as a UX upgrade. Gas fees block-to-block were unpredictable and people were overpaying for transactions. The baseFee was introduced as a protocol-calculated minimum price for gas to be included into a block, but to prevent it from being gamed by miners/validators the decision was made to _burn_ this baseFee, otherwise the block producers would directly benefit from driving the gas prices higher.

Burning ETH in this way has the side-effect of limiting supply and producing a deflationary measure, but this is not the direct reason ETH is burned.

src/content/developers/docs/intro-to-ether/index.md
```diff
@@ -41,6 +41,8 @@ As well as creating ether through block rewards, ether can get destroyed by a pr
 
 Ether burn occurs in every transaction on Ethereum. When users pay for their transactions, their base gas fee gets destroyed by the protocol. Depending on network demand, [some blocks](https://etherscan.io/block/12965263) burn more ether than they mint.
```
timbeiko, line 42: I understand this might not be the best page on ethereum.org to get into the weeds, but I tried to flesh out the rationale for burning the base fee, and to explain its impact on issuance. FWIW, I generally think "inflation" should be avoided because it has a "purchasing power" connotation which doesn't really apply when discussing ether issuance. 

```suggestion
Ether burn occurs in every transaction on Ethereum. When users pay for their transactions, a base gas fee, set by the network according to transactional demand, gets destroyed. This, coupled with variable block sizes and a maximum gas fee, simplifies transaction fee estimation on Ethereum. When network demand is high, [blocks](https://etherscan.io/block/12965263) can burn more ether than they mint, effectively offsetting ether issuance.

Burning the base fee prevents various ways in which miners could affect its level. For example, if miners got the base fee, they could include their own transactions for free and raise the base fee for everyone else. Alternatively, they could refund the base fee to some users off-chain, leading to a more opaque and complex transaction fee market.
```

src/content/developers/docs/intro-to-ether/index.md
```diff
@@ -39,7 +39,10 @@ Ether is minted when a miner creates a block on the Ethereum blockchain. As an i
...
 
-Ether burn occurs in every transaction on Ethereum. When users pay for their transactions, their base gas fee gets destroyed by the protocol. Depending on network demand, [some blocks](https://etherscan.io/block/12965263) burn more ether than they mint.
+Ether burn occurs in every transaction on Ethereum. When users pay for their transactions, a base gas fee, set by the network according to transactional demand, gets destroyed. This, coupled with variable block sizes and a maximum gas fee, simplifies transaction fee estimation on Ethereum. When network demand is high, [blocks](https://etherscan.io/block/12965263) can burn more ether than they mint, effectively offsetting ether issuance.
+
+Burning the base fee prevents various ways in which miners could affect its level. For example, if miners got the base fee, they could include their own transactions for free and raise the base fee for everyone else. Alternatively, they could refund the base fee to some users off-chain, leading to a more opaque and complex transaction fee market.
```
minimalsm, line 44:
```suggestion
Burning the base fee prevents various ways the miners could manipulate it otherwise. For example, if miners got the base fee, they could include their own transactions for free and raise the base fee for everyone else. Alternatively, they could refund the base fee to some users off-chain, leading to a more opaque and complex transaction fee market.
```

minimalsm (approved):
Pulling this in, thanks @m-zaink 

Be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And we've just released our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) also 🥳!

minimalsm:
@timbeiko thanks for weighing in here (you can also claim a POAP, if you'd like one)

minimalsm:
@all-contributors please add @m-zaink for docs

PR #5490, opened 2022-02-24
Farsi: Homepage

wackerow:
## Description
Updates Farsi (fa) homepage with latest from Crowdin


wackerow:
Agree we should look for feedback on this and try to address. We should address this as a separate issue though since this is an existing issue currently already live on the website. This PR just updates to the latest translation strings from Crowdin.

corwintines (approved):
Seems like the number stuff can be handled in a separate issue.

PR #5503, opened 2022-02-24
Rephrase requirement for translator certificate

lukassim:
## Description

Rephrase the requirement for the translator certificate to emphasize the fact that the certificate is issued for 5,000 approved words, not translated words.


corwintines (approved):
LGTM

PR #5511, opened 2022-02-25
CORS esp domain

pettinarip:
## Description
From the ESP site, we need to access the Matomo script located in the eth.org domain. This will enable us to access it.


samajammin (approved):

minimalsm (approved):
lgtm

PR #5517, opened 2022-02-27
Add translation banner for legal pages

corwintines:
## Description
- Add translation banner for legal pages that lets users know in their language that these pages wont be translated.


src/components/TranslationBannerLegal.js
```diff
@@ -0,0 +1,145 @@
...
+                localStorage.setItem("dont-show-translation-legal-banner", true)
+                setIsOpen(false)
+              }}
+            >
+              <Translation id="translation-banner-no-bugs-dont-show-again" />
+            </ButtonPrimary>
+          </ButtonRow>
+          <BannerClose
+            onClick={() => setIsOpen(false)}
+            isPageRightToLeft={isPageRightToLeft}
```
minimalsm, line 135: The cross to close the banner isn't displaying on the correct side on [/ar/ legal pages](https://ethereumorgwebsitedev01-translationbannerlegalpages.gtsb.io/ar/contributing/)

Screenshot 2022-02-28 at 12 46 45

Strangely, it is working on [non-legal pages.](https://ethereumorgwebsitedev01-translationbannerlegalpages.gtsb.io/ar/developers/tutorials/)

Screenshot 2022-02-28 at 12 49 15

src/components/TranslationBannerLegal.js
```diff
@@ -0,0 +1,145 @@
...
+
+const TranslationBannerLegal = ({
+  shouldShow,
+  originalPagePath,
+  isPageRightToLeft,
+}) => {
+  const [isOpen, setIsOpen] = useState(shouldShow)
+
+  useEffect(() => {
+    if (localStorage.getItem("dont-show-translation-legal-banner") === "true") {
```
minimalsm, line 103: After clicking 'don't show' I'm getting a flash of the banner when I refresh ([see video of bug.](https://i.gyazo.com/79d02ed7ddad1f648d260f83df4188e9.mp4))

minimalsm (changes requested):
Thanks Corwin.

Looking good! Noted a small bug below.

src/components/TranslationBannerLegal.js
corwintines: Woops, small mistake in render function. Fixed.

src/components/TranslationBannerLegal.js
corwintines: Changed the default state here, seems to fix this issue now.

corwintines:
@minimalsm I think this is ready for rereview. Thanks for feedback.

src/components/TranslationBannerLegal.js
```diff
@@ -0,0 +1,146 @@
...
+  // Default to isOpen being false, and let the useEffect set this.
+  const [isOpen, setIsOpen] = useState(false)
+
+  useEffect(() => {
+    if (localStorage.getItem("dont-show-translation-legal-banner") === "true") {
+      setIsOpen(false)
+    } else {
+      setIsOpen(shouldShow)
+    }
+  }, [originalPagePath, shouldShow])
```
pettinarip, line 109: `originalPagePath` should not be necessary anymore

pettinarip (approved):
lgtm

src/components/TranslationBannerLegal.js
wackerow:
```suggestion
  }, [shouldShow])
```

src/components/TranslationBannerLegal.js
```diff
@@ -0,0 +1,146 @@
...
+              <StyledEmoji ml={"0.5rem"} size={1.5} text=":bug:" />
+            </H3>
+          </Row>
+          <p>
+            <Translation id="translation-banner-no-bugs-content" />
+          </p>
+          <ButtonRow>
+            <ButtonPrimary
+              onClick={() => {
+                localStorage.setItem("dont-show-translation-legal-banner", true)
```
wackerow, line 127: Should we consider altering this key to be unique to the page?

Noticed if I were to go to the cookies or privacy policy, and clear that alert, I would not get it again on the style guide where it may not be clear the association is the same.

Just a thought, though I don't want to make the UI overbearing either. 

If we do decide to make it unique for each page:
```suggestion
                localStorage.setItem(`dont-show-translation-legal-banner-${originalPagePath}`, true)
```

wackerow (changes requested):
Looks good, just left a comment about having the "hide" preference unique for each page that this applies to. Not a hard stance, just thought I'd bring it up.

Otherwise we should just be able to ditch the extra useEffect dependency and should be good 👍

src/components/TranslationBannerLegal.js
corwintines: I can get behind this idea.

corwintines:
@wackerow @pettinarip I am adding the originalPagePath back into the useEffect as its being used to check the localStorage path, and if its not in there then the component would not get rendered if you change between legal pages. To test this out to see the bug here are the steps to reproduction:
1 - remove originalPagePath from useEffect
2 - go to /fr/terms-of-use/
3 - Click dont show again button
4 - click the privacy policy page in the header or footer
5 - notice its not rendering on this page, even though it should be given the key its checking localStorage for

wackerow (approved):

PR #5554, opened 2022-03-01
Italian: Learn pages

wackerow:
## Description
Updates Italian (`it`) **learn pages** content with latest from Crowdin


minimalsm (approved):
Noticed a few problems but they stem from the source content and nothing we can do until next update. I've compiled a list and will change in Crowdin.

PR #5583, opened 2022-03-04
Open PDFs in browser

pettinarip:
## Description

Currently, the pdfs are downloaded by default when you click on them. With this change they will open in a new tab.

This will improve the UX and will be useful to link or bookmark the files inside eth.org.

### DIFF
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -157,7 +157,7 @@ const Link = ({
   // Download link for internally hosted PDF's (ex: whitepaper)
   if (isPdf && !isExternal) {
     return (
-      <a href={to} download>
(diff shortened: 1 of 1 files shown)


polarpunklabs:
Perfect.

minimalsm (approved):
Nice

We should change the copy of the link too as 'Download ... as PDF' is no longer accurate.

pettinarip:
> We should change the copy of the link too as 'Download ... as PDF' is no longer accurate.

Uhmm yes nice catch, fxing...

minimalsm:
no need, @pettinarip I've just done this in #5584 :-)

PR #5630, opened 2022-03-09
add alchemy nft api to nft standards

corwintines:
## Description
- add alchemy nft api docs to further reading for erc-721 and erc-1155 pages. This api supports these standards.


minimalsm (commented):
Do we think developer tooling is appropriate on these pages?

corwintines:
I would argue yes when theres specific tooling.

samajammin (approved):

PR #5635, opened 2022-03-09
Remove rocketpool from staking page [Fixes #5633]

### LINKED ISSUE #5633: Remove RocketPool as the recommended staking solution
Placing RocketPool as recommended staking solution for users with less than 32 ETH breaks the idea that Ethereum Foundation should stay neutral to the protocols that build on top. There should be a link to all the solutions out there, and the user should be able to evaluate all of them and choose the one he likes the most.

tsudmi:

### DIFF
--- a/src/intl/en/page-staking.json
+++ b/src/intl/en/page-staking.json
@@ -32,7 +32,6 @@
   "page-staking-more-sharding": "More on sharding",
   "page-staking-pool": "Stake in a pool",
   "page-staking-pool-desc": "If you have less than 32 ETH, you will be able to add a smaller stake to staking pools. Some companies can do it all on your behalf so you won't have to worry about staying online. Here are some companies to check out.",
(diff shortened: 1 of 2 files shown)


djrtwo:
I agree that rocketpool should not be *independently* highlighted like this on the ethereum.org website.
I think the page should either give a better range of possibilities or not address the issue at all

jyturley:
Why not add other options instead of removing options here?  Plus, I don't think we should shy away from any data that would promote decentralization.

NessDan:
Rocket pool is currently the best option for a truly decentralized pool. I think adding others that fit the criteria is fair, but the rest are either permissioned or DAO controlled (unless there's an alternative I'm forgetting?)

minimalsm:
Thanks, @tsudmi @djrtwo @jyturley @NessDan 

To keep things focussed here, let's keep the discussion on the issue itself: #5633 

Going to lock conversation here temporarily to keep all thoughts and opinions in one place.

minimalsm:
As discussed in #5633, we have decided not to move forward with this PR. For more information on our plans to address these concerns and the staking page, please check out #5647.

We appreciate all the feedback given to make ethereum.org better 🙏

PR #5645, opened 2022-03-10
Move layer 2 translations

corwintines:
## Description
- create and move layer 2 translations to `layer-2.json` in prep for new page


samajammin:
Just noting order of operations here - I think we should ship new pages in English (without translation support) first before configuring pages for translations. This makes it easier to quickly iterate based on community feedback before uploading for translations.

More on that process here:
https://www.notion.so/efdn/New-page-process-fbf2d59a4b284025992fc23d990bd689

minimalsm (approved):

PR #5653, opened 2022-03-11
fix: add note for Arbitrum

smartcontracts:
## Description

Adds a disclaimer for Arbitrum. According to Arbitrum's documentation, Arbitrum's fraud proofs can currently only be used by [whitelisted addresses](https://developer.offchainlabs.com/docs/mainnet). This whitelist is currently not open to the general public, which has important effects on the security model of the system (only the Arbitrum team can currently use the fraud proof system). This is important to note (alongside the notes described for other L2s).


corwintines (approved):
I agree wtih adding this note given the current whitelist on Arbitrum. Thanks @smartcontracts!

minimalsm:
@all-contributors please add @smartcontracts for docs

PR #5669, opened 2022-03-14
supporting a big picture perspective

thewild-being:
I see very often how other people tend to get lost in the technical details of a said technology and they lose the big picture that allows them to see things from first principle. I wanted to add ## Dictation as a consensus metaphor as a way to help those who are learning to not get lost in the technical minutiae.


minimalsm:
Hey @thewild-being thanks for the PR. Agree with the idea of simplifying our documentation as much as possible. However, I don't think that adding this makes sense in its current form. I understand consensus mechanisms relatively well and I found this confusing. I asked someone who didn't understand consensus mechanisms well to look it over and they too were a bit confused. I just don't think it's a good metaphor... 

Open to other ideas about how we can improve this but would appreciate if you open an issue first so that people can weigh in and discuss ideas before opening a PR. Thanks again ❤️

minimalsm:
@all-contributors please add @thewild-being for ideas

PR #5686, opened 2022-03-16
Update README

pettinarip:
## Description

Updates the README file with:
- the new env variables that control the languages and content that is created when you run `yarn start`.
- GC as our hosting service for preview deploys.

Bonus:
- Updated code sections with the correct syntax highlighting.


minimalsm (approved):
Nice work :-)

PR #5691, opened 2022-03-17
Fix header ids scrolling

pettinarip:
## Description
Scroll to the anchor's start position by default, not the end.

Bonus:
- Created a utils function to reuse the code.

Example:
https://ethereumorgwebsitedev01-fixscrollintoview.gtsb.io/en/whitepaper/

### DIFF
--- a/src/components/Layout.js
+++ b/src/components/Layout.js
@@ -21,6 +21,7 @@ import { ZenModeContext } from "../contexts/ZenModeContext"
 import { useKeyPress } from "../hooks/useKeyPress"
 
 import { isLangRightToLeft } from "../utils/translations"
+import { scrollIntoView } from "../utils/scrollIntoView"
 import { isMobile } from "../utils/isMobile"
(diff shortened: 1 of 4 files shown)


src/components/Layout.js
```diff
@@ -90,9 +91,7 @@ const Layout = (props) => {
 
     if (props.location.hash && !props.location.hash.includes("gatsby")) {
       const idTag = props.location.hash.split("#")
-      if (document.getElementById(idTag[1]) !== null) {
-        document.getElementById(idTag[1]).scrollIntoView(false)
```
pettinarip, line 94: The problem we had. `scrollIntoView(false)` will put the anchor at the bottom of the viewport.

pettinarip:
> Merging as this solves the immediate bug and I will create a separate issue for each of these

@minimalsm cool yes, that is more a style decision we need to make, whether we want to have a margin or not. Implementing that or not shouldn't affect anything.

PR #3704, opened 2021-08-24
Add "MEV" page to developer docs [Resolves #3610]

nathanhleung:
## Description

This PR adds a page about miner extractable value (MEV) to the developer docs.

Let me know what you think — I am also sending the deploy preview out to other folks I know in the MEV space for more feedback.

## Related Issue

#3610

Link to deploy preview of the page: https://deploy-preview-3704--ethereumorg.netlify.app/en/developers/docs/miner-extractable-value/

PR #5698, opened 2022-03-18
Update explainer for "miner extractable value" deprecation

Buttaa:


minimalsm:
Hey @Buttaa, thanks for the PR and for re-highlighting this.

Using `miner extractable value` over `maximal` was actually a deliberate decision we made [when creating this page in September 2021](https://github.com/ethereum/ethereum-org-website/pull/3704).

Some thoughts on our logic at the time:

### Usage

The term `maximal extractable value` was fairly new and not being used widely. For that reason, [we believed at the time](https://github.com/ethereum/ethereum-org-website/pull/3704#discussion_r702328263) that most people who knew the term `maximal extractable value` knew what `miner extractable value` was, but the opposite was definitely not true. 

**How does this stack up today? Are MEV circles commonly using `maximal` now?**

### Discoverability

At the time, the resources explaining MEV were poor and we wanted to provide a much needed, easy to find, source of truth. Search volume for miner extractable value was exponentially higher and in order to educate more users, it didn't make sense to strip out miner extractable value altogether.

**We should investigate the current search volumes and see if it makes sense to make modifications elsewhere on the page to serve users who are searching `miner extractable value`.**

--- 

Overall, I'm I agree that it's probably time but we should perform a little bit of due diligence first.

cc: @samajammin 

_Also noting: this page needs a massive overhaul for the switch to proof-of-stake as part of our Merge prep_

wackerow:
Miners days are numbered... time to switch

samajammin:
Overall I support these changes 👍 but for discoverability I think it's important to maintain mentions of "miner extractable value" on the page. Search volume for "miner extractable value" is still ~100x greater than "maximal extractable value":
Screen Shot 2022-03-25 at 11 19 19 AM

wackerow:
Hey @Buttaa, thanks for nudging us on this change. To the above points, added/edited some copy to be more explicit about the name change, giving it its own section for SEO purposes. 
cc: @samajammin @minimalsm

corwintines (approved):

PR #5713, opened 2022-03-20
docs: grammar/punctuation changes in Spanish page-index 

eduardopelitti:
## Description

Replaced "un guía", which in Spanish would refer to a person who acts as a guide, with "una guía" which refers to a guide as an object/concept. Also removed the ";" as I think it's being misused here, I believe the phrase works better when expressed in Spanish rather than directly translated.

Also added a missing "1" at the end of the min ETH purchase value.


minimalsm:
Thanks for the PR @eduardopelitti ❤️

@pettinarip could you have a look?

pettinarip (approved):
Thanks @eduardopelitti! looks good 💪🏼

PR #5724, opened 2022-03-22
Moved up the gas expense paragraph; removed kLast reference

tnkrxyz:
## Description

The paragraph in the Uniswap-V2 Contract Walk Through document about gas saving by packing variables refers to three variables, `reserve0`, `reserve1`, and `blockTimestampLast`, and is moved up to be closer to the relevant code chunks. `kLast` is a regular uint256 and not packed in a single slot with other variables and its reference removed.


minimalsm:
Thanks @tnkrxyz 😀

@qbzzt could you check this one out also?

qbzzt (approved):
You're right. Thank you for the correction.

minimalsm (approved):

PR #5727, opened 2022-03-22
Refactor to avoid using ids in the search inputs

pettinarip:
## Description

We were having duplicate ids with the search inputs since we have 2, one for mobile and another one for desktop. This is a refactor to avoid using ids to focus the inputs.

### DIFF
--- a/src/components/Search/index.js
+++ b/src/components/Search/index.js
@@ -1,4 +1,4 @@
-import React, { useState, createRef } from "react"
+import React, { useState, useRef } from "react"
 import { Link as GatsbyLink } from "gatsby"
 import { useIntl } from "gatsby-plugin-intl"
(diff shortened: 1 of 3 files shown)


src/components/Search/index.js
```diff
@@ -112,43 +113,45 @@ const StyledHighlight = styled(Highlight)`
...
-          />
-        )}
-        {hit.content && (
-          <StyledSnippet attribute="content" hit={hit} tagName="mark" />
-        )}
-      </GatsbyLink>
-    </div>
-  )
-}
+const PageHit =
```
pettinarip, line 116: (this diff block is just prettier fixing formatting things)

minimalsm (approved):
Nice 💪

PR #5656, opened 2022-03-11
Layer 2 page

corwintines:
## Description
[WIP] Layer 2 page

PR #5756, opened 2022-03-25
L2 page tweaks

samajammin:
## Description

A few proposed changes, based largely off feedback here: https://twitter.com/ryancreatescopy/status/1506893580065026051?s=20&t=st3EwiUqCDhF8yYzVCh9qg

- Define L2 right off the bat in a new section at the top
- Re-word existing sections to "How layer 2 works" & "Use layer 2"
- Misc copy updates

## Related Issue
#5656

### DIFF
--- a/src/pages/layer-2.js
+++ b/src/pages/layer-2.js
@@ -412,8 +412,8 @@ const Layer2Page = ({ data }) => {
           Layer 2 (L2) is a collective term to describe a specific set of
           Ethereum scaling solutions.{" "}
           <b>
-            A layer 2 is separate blockchain that's connected to Ethereum and
-            inherits the security guarantees of Ethereum
+            A layer 2 is separate blockchain that extends Ethereum and inherits
+            the security guarantees of Ethereum
           </b>
           .
         </p>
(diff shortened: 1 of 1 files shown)


src/pages/layer-2.js
```diff
@@ -406,19 +406,34 @@ const Layer2Page = ({ data }) => {
         </StatsContainer>
       </PaddedContent>
 
+      <PaddedContent id="what-is-layer-2">
```
samajammin, line 409: Noting this layout currently looks poor - the focus was on content not design 😅 

This top section probably shouldn't be full width:
Image 2022-03-24 at 10 42 22 PM

nloureiro: Today we will look into it.

src/pages/layer-2.js
```diff
@@ -406,19 +406,37 @@ const Layer2Page = ({ data }) => {
...
+        <h2>What is layer 2?</h2>
+        <p>
+          Layer 2 (L2) is a collective term to describe a specific set of
```
samajammin, line 412: Regarding copy for defining L2, I very much like Barnabé's explanation:
https://barnabe.substack.com/p/understanding-rollup-economics-from?s=r

He's speaking specifically to rollups but I like how he frames them as an extension the protocol they build on. I think saying L2s extend Ethereum is better than merely saying they're connected to Ethereum.

corwintines (approved):

PR #5804, opened 2022-03-29
Update client diversity chart

minimalsm:
## Description
- Replaces our existing client diversity chart with a new shiny one 😀
- Updates data attributions

## Related Issue
None

Preview deploy: https://ethereumorgwebsitedev01-addnewclientdiversityimage.gtsb.io/en/developers/docs/nodes-and-clients/client-diversity/


wackerow (approved):
@minimalsm Very quickly thinking we need a better solution for this to be honest. 
https://pools.invis.cloud Showing different (and improved) stats for this. When were these snapshots taken? Looks like the data updated in the image, but the copy still mentioned "at time of writing in January 2022".

Not against pulling this in in the meantime, but would love to think about more sustainable solutions here. Could look into an API key for the above link, not sure if anyone has any suggestions here to pull more up-to-date data.

Issue #5790, opened 2022-03-28
Repeated ids on the stats charts


pettinarip:
**To Reproduce**
Steps to see the bug:

1. Go to the home page https://ethereum.org/
2. Scroll down to the stats charts
3. Each chart has the following repeated ids in its implementation: `colorUv` and `colorPv`
![image](https://user-images.githubusercontent.com/468158/160425215-8158ec2a-d40a-4dd7-8b32-cbc36e92b586.png)

**Expected behavior**

Modify the way we set these ids (`colorUv` and `colorPv`) to have a unique id per chart created.

We could define them as `colorUv-${chartType}`.

**Additional context**

The ids are defined in the `GridItem` component that lives inside the `StatsBoxGrid` file.

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

HursitTarcan:
Can you assign me for this?

Steph-crown:
Hi, can I hop on this?

PR #5825, opened 2022-03-29
Update StatsBoxGrid.js

kum9748ar:
redefined linear Gradient id's

## Description
redefined linear Gradient id's with `-${chartType}`

## Related Issue
https://github.com/ethereum/ethereum-org-website/issues/5790

### DIFF
--- a/src/components/StatsBoxGrid.js
+++ b/src/components/StatsBoxGrid.js
@@ -3,14 +3,14 @@ import styled from "styled-components"
 import { useIntl } from "gatsby-plugin-intl"
 import axios from "axios"
 import { kebabCase } from "lodash"
-
 import { AreaChart, ResponsiveContainer, Area, XAxis } from "recharts"
+
 import Translation from "./Translation"
 import Tooltip from "./Tooltip"
 import Link from "./Link"
 import Icon from "./Icon"
 
-import { isLangRightToLeft } from "../utils/translations"
+import { isLangRightToLeft, translateMessageId } from "../utils/translations"
 import { getData } from "../utils/cache"
 
 const Value = styled.span`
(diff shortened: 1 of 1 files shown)


src/components/StatsBoxGrid.js
```diff
@@ -182,11 +182,11 @@ const GridItem = ({ metric, dir }) => {
         margin={{ left: -5, right: -5 }}
       >
         <defs>
-          <linearGradient id="colorUv" x1="0" y1="0" x2="0" y2="1">
+          <linearGradient id="colorUv-${chartType}" x1="0" y1="0" x2="0" y2="1">
             <stop offset="5%" stopColor="#8884d8" stopOpacity={1} />
             <stop offset="95%" stopColor="#8884d8" stopOpacity={0} />
           </linearGradient>
-          <linearGradient id="colorPv" x1="0" y1="0" x2="0" y2="1">
+          <linearGradient id="colorPv-${chartType}" x1="0" y1="0" x2="0" y2="1">
```
pettinarip, line 189: `chartType` var doesn't exist. I used it just as an example in the issue description.

Perhaps we could use the `title` here that is being destructured in line 150. For safety reasons, you could do: `_.kebabCase(title)` to avoid having spaces.

What do you think?

pettinarip (changes requested):
Hey @kum9748ar thanks for the PR!

I've added a comment there. Please take a look and LMK in case you need help.

src/components/StatsBoxGrid.js
pettinarip: We already have `lodash` installed. You can just do `import { kebabCase } from "lodash"` 👍🏼

pettinarip (approved):

kum9748ar:
@pettinarip yeah I have  checked now it looks nice thanks for refactoring

pettinarip:
@all-contributors please add @kum9748ar for a11y

PR #5827, opened 2022-03-29
Revert tvl lambda name

pettinarip:
Revert the new lambda called `tvl` to the old name `defipulse` to fix possible issue with GC -> Netlify deploy.


wackerow (approved):

wackerow (changes requested):
@pettinarip We should merge this into `dev` first

PR #5831, opened 2022-03-30
Add aria-label to exchange search input #5793

### LINKED ISSUE #5793: Add aria-label to exchange search input
**Describe the bug**
The select/search input doesn't have an aria-label description.

**To Reproduce**

1. Go to https://ethereum.org/en/get-eth/
2. Scroll down to the select input
image

**Expected behavior**

Add aria-label to the underlying input element.

**Want to contribute?**

KaranKaira:
## Description
Added aria-label to exchange search input

## Related Issue
Fixes #5793


src/components/EthExchanges.js
```diff
@@ -549,6 +549,7 @@ const EthExchanges = () => {
         <Translation id="page-get-eth-exchanges-intro" />
       </Intro>
       <StyledSelect
+        aria-label={getDefaultMessage("page-get-eth-exchanges-search")}
```
minimalsm, line 552: Using `getDefaultMessage` means that this will always resolve to English aria-labels. We should use `translateMessageId` here instead to make this accessible across languages. 

[See here for an example](https://github.com/ethereum/ethereum-org-website/blob/1829598d2a2e3819a7429bd9414db64b6b6724cc/src/components/Nav/Mobile.js#L269).

src/pages-conditional/dapps.js
```diff
@@ -1134,7 +1134,7 @@ const DappsPage = ({ data, location }) => {
           <Translation id="get-started" />
         </H2>
         <p>
-          <Translation id="page-dapps-get-started-subtitle" />{" "}
+          <Translation id="page-dapps-get-started-subtitle" />
```
minimalsm, line 1137: I think this and the other changes got added accidentally from a different PR? 

Could you remove them from this PR please 🙏

minimalsm (changes requested):

minimalsm:
@pettinarip do you think we should create a new translation string for this? 

The current string 'Type where you live' makes sense in the visual context provided but less so when only using a screenreader. i.e. is where I live my town? city? state? 

Also, it's bad practice to include unnecessary verbs ('type') as screen readers already announce an element's role (input).

src/components/EthExchanges.js
KaranKaira: Sure!

src/pages-conditional/dapps.js
KaranKaira: Hey @minimalsm , I have changed`getDefaultMessage` to `translateMessageId`.
>  Could you remove them from this PR please 🙏

Can you tell how to delete these changes from a pull request.
Thanks

pettinarip:
> do you think we should create a new translation string for this?

Now that you say this I totally agree. Hmm maybe we could reuse the header string in this section `What country do you live in?`
image

What do you think? I think that one will work better for screen readers

pettinarip:
@KaranKaira we can go with that text. Lets use that one for now 👍🏼

KaranKaira:
@pettinarip , made changes .. have a look

minimalsm (approved):

minimalsm:
@all-contributors please add @KaranKaira for code

Issue #5788, opened 2022-03-28
No labels on the wallets checkboxes


pettinarip:
**Describe the bug**

Missing labels on the wallets checkboxes.

**To Reproduce**
Steps to reproduce:

1. Go to https://ethereum.org/en/wallets/find-wallet/
2. Scroll down to the "Choose the wallet" section
3. The cards checkboxes don't have labels
![image](https://user-images.githubusercontent.com/468158/160419996-98d8e2fb-9dcf-4dd6-b4ce-308044098cf3.png)

**Expected**

- Each underlying checkbox should have a wrapping `label` element or an `aria-label` attribute.

**Bonus**

- Add styles to highlight the checkbox when they are on focus (probably with `tabindex="0"`).

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

HursitTarcan:
I can fix this

PR #5844, opened 2022-03-30
Aria-labels added the wallets checkboxes [Solves #5788]

HursitTarcan:
## Description
- 'Checkbox' in Checkbox.js has a new ariaContent property (line 56, Checkbox.js).
- 'HiddenCheckbox' has a new attribute 'aria-label' with the ariaContent value in it (line 67, Checkbox.js).
- 'StyledCheckbox' in SelectableCard.js has the title.id value (translated with the translateMessageId function) passed for 
   'ariaContent' (line 58, SelectableCard.js).

## Related Issue
- Related to issue: [#5788]

## Steps to reproduce
1. Add a new property to Checkbox. E.g: 'ariaContent' (line 56, Checkbox.js).
2. Add an aria-label attribute to HiddenCheckbox (line 67, Checkbox.js).
3. Put the ariaContent (created in step 1) in the aria-label (line 67, Checkbox.js).
4. In StyledCheckbox add the ariaContent property (line 58, SelectableCard.js).
5. Now take 'title.props.id' and use the translateMessageId-method as value for ariaContent (line 58, SelectableCard.js).

### DIFF
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -49,7 +49,14 @@ const Label = styled.span`
   margin-left: 0.5rem;
 `
 
-const Checkbox = ({ callback, checked, children, className, size = 2 }) => {
+const Checkbox = ({
+  callback,
+  checked,
+  children,
+  ariaContent,
+  className,
+  size = 2,
+}) => {
   const handleClick = () => {
     if (callback) {
       callback()
@@ -57,7 +64,7 @@ const Checkbox = ({ callback, checked, children, className, size = 2 }) => {
   }
   return (
     <CheckboxContainer className={className} onClick={handleClick}>
-      <HiddenCheckbox checked={checked} readOnly />
+      <HiddenCheckbox checked={checked} readOnly aria-label={ariaContent} />
       <StyledCheckbox checked={checked} className="styled-checkbox" size={size}>
         <Icon checked={checked} viewBox="0 0 24 24">
           <polyline points="20 6 9 17 4 12" />
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -56,6 +56,7 @@ const Checkbox = ({
   ariaContent,
   className,
   size = 2,
+  ...rest
 }) => {
   const handleClick = () => {
     if (callback) {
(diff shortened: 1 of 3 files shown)


src/components/WalletCompare.js
```diff
@@ -114,50 +114,50 @@ const walletFeatures = [
...
+    title: "page-find-wallet-swaps",
+    description: "page-find-wallet-swaps-desc",
   },
   {
     id: "has_multisig",
     emoji: ":busts_in_silhouette:",
-    title: <Translation id="page-find-wallet-multisig" />,
-    description: <Translation id="page-find-wallet-multisig-desc" />,
+    title: "page-find-wallet-multisig",
+    description: "page-find-wallet-multisig-desc",
```
pettinarip, line 160: @HursitTarcan I left only the keys here so that we translate this strings in the `WalletCompare`. This way you receive a string prop and not a component/element.

src/components/Checkbox.js
```diff
@@ -49,15 +49,22 @@ const Label = styled.span`
...
+}) => {
   const handleClick = () => {
     if (callback) {
       callback()
     }
   }
   return (
     <CheckboxContainer className={className} onClick={handleClick}>
-      <HiddenCheckbox checked={checked} readOnly />
+      <HiddenCheckbox checked={checked} readOnly aria-label={ariaContent} />
```
pettinarip, line 67: Lets not define a new prop name (`ariaContent`) for this. We can do the following:
```jsx
const Checkbox = ({ callback, ...rest }) => {
  ....
  return (
    ...
    <HiddenCheckbox checked={checked} readOnly {...rest} />
  )
```

and from the consumer view, you can just use the native aria props
```jsx
<StyledCheckbox checked={isSelected} aria-label={title} />
```

pettinarip (changes requested):
Looking good @HursitTarcan I left a small request, let me know if you can implement that or LMK.

src/components/Checkbox.js
HursitTarcan: I made the changes, thanks for helping!

pettinarip (approved):
Awesome @HursitTarcan 🎉  thank you!

pettinarip:
@all-contributors please add @HursitTarcan for a11y

PR #5748, opened 2022-03-24
Staking epic revamp

wackerow:
## Description
Introduces redesign of the ethereum.org/staking page. This includes three new subpages focusing on different paths users can take to stake:
- solo staking (`/staking/solo/`)
- staking as a service (`/staking/saas/`)
- and pooled staking (`/staking/pools/`)

This also introduces new product listings for the following categories
- Node tools (for solo staking)
- SaaS providers
- Pooling services
- Key generators (for solo staking and SaaS staking)

Provides more in depth explanation of the differences between these options, and trade-offs between each.
Guides, frequently asked questions, and further readings to help users fill in any gaps.

Preview deploy:
https://ethereumorgwebsitedev01-stakingepic.gtsb.io/en/staking

Feedback always welcome!

## Related issue
[Closes #5646, Closes #5647]

PR #5864, opened 2022-03-31
SVG upgrade using gatsby-plugin-react-svg

wackerow:
## Description
- [`yarn add gatsby-plugin-react-svg`](https://www.gatsbyjs.com/plugins/gatsby-plugin-react-svg/)

- Adds ability to use imported `.svg` files as React components, compatible with styled-components.
- Enables rendering imported SVGs as true `<svg>` elements with access to sub-elements, instead of it being rendered as an `img` which removes access to svg properties.

Usage:

```js
import SvgImage from "/path/to/svg-image.svg"
```
```js
<SvgImage />
```
or
```js
const StyledSvgImage = styled(SvgImage)``
```

This affects existing SVGs that are now being imported as components and passed to ``. These have been refactored to simply import the variable using PascalCase, and then using this directly as a component. CSS selectors have been updated to select for `svg` instead of `img`.

## Related Issue
None filed, but related to features being used in `staking-epic` branch on #5748


src/components/ExpandableCard.js
```diff
@@ -81,7 +81,7 @@ const StyledFakeLink = styled(FakeLink)`
   white-space: nowrap;
 `
 
-const ExpandableCard = ({ children, contentPreview, title, svg, alt }) => {
+const ExpandableCard = ({ children, contentPreview, title, Svg, alt }) => {
```
pettinarip, line 84: I would suggest using lowercase for props to follow conventions. Then you can redefine it
```jsx
const Svg = props.svg
```
https://airbnb.io/javascript/react/#props

pettinarip (approved):
LGTM @wackerow 💪🏼

src/pages/run-a-node.js
```diff
@@ -595,19 +595,22 @@ const RunANodePage = ({ data }) => {
...
-              alt={<Translation id={alt} />}
-              key={idx}
-            >
-              {body.map((item, idx) => (
-                <p key={idx}>{<Translation id={item} />}</p>
-              ))}
-            </ExpandableCard>
-          ))}
+          {whyRunANodeCards.map(({ Image, title, preview, body, alt }, idx) => {
+            console.log({ Image })
```
pettinarip, line 599: remove

src/pages/run-a-node.js
```diff
@@ -595,19 +595,22 @@ const RunANodePage = ({ data }) => {
...
+            return (
+              <ExpandableCard
+                contentPreview={<Translation id={preview} />}
+                title={<Translation id={title} />}
+                Svg={Image}
+                alt={<Translation id={alt} />}
```
pettinarip, line 605: This need to be switched to `translateMessageId(alt)` since we are not rendering this as a component.

pettinarip (changes requested):

pettinarip (approved):

PR #5880, opened 2022-04-01
Add API tooltip to L2 page

corwintines:
## Description
- Add API tooltip to l2 page stats
- remove fee data for L2 projects not listed


src/pages/layer-2.js
```diff
@@ -199,21 +221,27 @@ const Layer2Page = ({ data }) => {
...
-          ) / feeData.data.length
+          ) / feeData.length
 
         const intlFeeAverage = new Intl.NumberFormat(intl.locale, {
           style: "currency",
           currency: "USD",
           notation: "compact",
           minimumSignificantDigits: 2,
-          maximumSignificantDigits: 3,
+          maximumSignificantDigits: 2,
```
minimalsm, line 244: Currently, the 'Average layer 2 ETH transfer fee' is showing up as 1 decimal point. I was verrrry confused by this because in school, we were taught decimal places ==== significant digits. [Apparently not!](https://simple.wikipedia.org/wiki/Arithmetic_precision)

Screenshot 2022-04-01 at 21 20 32

Not 100% sure if this is what you're after but it looks like there is a [`minimumFractionDigits` and `maximumFractionDigits`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) that we could use here to get it to display like regular currency ($1.10, in this case).

Tested a little and it seems to be working as we would expect.

Screenshot 2022-04-01 at 21 49 19
Screenshot 2022-04-01 at 21 49 39

minimalsm (commented):
The tooltip looks good and works on all screensizes for me. Tested on my mobile too 🎉😀

wackerow (approved):

Issue #5731, opened 2022-03-23
Migrate materials from ethdocs.org


jmcook1186:
## Background

The homestead guide at [ethdocs.org](https://ethdocs.org/en/latest/index.html) is now a historical archive. While it is full of interesting information it is now very out of date and users can easily be confused by contradictory information provided there compared to newer resources such as ethereum.org. Therefore, this issue is to suggest selective migration of material from ethdocs.org to ethereum.org such that the latter is becomes the go-to resource.

This has been discussed previously, e.g. on the ethdocs github [here](https://github.com/ethereum/homestead-guide/issues/473), but no action has yet been taken. Here, I have done a page-by-page audit of ethdocs.org and identified pages that can simply be deprecated using 301 redirects to the equivalent information on etheruem.org and which pages contain relevant material not currently included on ethereum.org that should be migrated across.

Note that there is crossover with another issue #5690 that suggests similar actions for the ethwiki. There, I suggested several new pages for ethereum.org that can also include material from ethdocs.org. 

Below is a page-by-page audit for ethdocs with recommendations. Below are specific PRs that would address the suggested actions.

```
|   Suggestions     |                  Notes                         |  File
|-------------------|------------------------------------------------|-----------------------------------------------|
|                   |                                                ├── source                                      |
| Sunset            | Info specific to homestead guide               │   ├── about.rst                               |
|                   |                                                |   |                                           |
| Sunset            | OOD and geth specific. Redirect to geth docs   │   ├── account-management.rst                  |
|                   | and/or eth.org                                 |   |                                           |
|                   | /developers/docs/accounts/                     |   |                                           |
|                   |                                                |   |                                           |
| Sunset            | OOD - redirect to eth.org/developers/          │   ├── connecting-to-clients                   |
|                   | nodes-and-clients                              |   |   |                                       |
|                   |                                                |   |   |                                       |
|                   | docs/programming-languages/ruby                │   │   ├── ethereum-ruby                       |
|                   |                                                │   │   │   └── index.rst                       |
|                   |                                                │   │   ├── index.rst                           |
| Sunset            | /docs/programming-languages/dot-net/           │   │   ├── nethereum                           |
|                   |                                                │   │   │   └── index.rst                       |
| Sunset            | /docs/programming-languages/java/              │   │   ├── web3j                               |
|                   |                                                │   │   │   └── index.rst                       |
| Sunset            | /docs/apis/javascript/                         │   │   ├── web3.js                             |
[...]

Issue #5690, opened 2022-03-17
Migrate materials from eth.wiki


jmcook1186:
## Background

The material hosted at eth.wiki is now very stale and contains information that is years out of date. This can be confusing for users because the old information can contradict newer information available elsewhere and it is not clear what the canonical source of information should be. While the wiki is rich with Ethereum history, it seems like a good time to migrate some information from those sites onto ethereum.org and sunset those older sites.

There have been several suggestions along similar lines recently but so far only relatively minor actions have been taken, e.g.

https://github.com/ethereum/eth-wiki/issues/55

Therefore, I'm raising this issue to jump-start progress on this and get the good content from the wiki migrated over to ethereum.org.

## Proposal

I have been through the wiki page by page and made suggestions for which materials I think should be migrated to ethereum.org and which materials should be archived/taken offline. Below is a table with those file-by-file suggestions. Underneath the table is a high level summary of the actions that could be taken int he form of 12 PRs to ethereum.org to preserve the good information currently only available in the wiki.
              
                    

```                    
|   Suggestions     |                  Notes                         |  File                                          |
|-------------------|------------------------------------------------|------------------------------------------------|
|                   |                                                ├── archive                                      |
|Sunset             | redirect to eth.foundation                     │   ├── licensing.md                             |
|                   |                                                |   |                                            |
|                   |                                                |   |                                            |
|Sunset             | historically interesting but uneccessary       │   ├── poll-for-token-proposal-EIP-20.md        |
|                   | redirect to /developers/docs/standards/        |   |                                            |
|                   | tokens/erc-20/                                 |   |                                            |
|                   |                                                |   |                                            |
|Sunset             | redirect to eth.org                            │   ├── proposal-reversion-notification.md       |
|                   | /developers/docs/apis/json-rpc                 |   |                                            |
|                   |                                                |   |                                            |
|Sunset             | redirect to eth.org                            |   ├── proposal-transaction-proxy-hooks.md      |
|                   | /developers/docs/apis/json-rpc                 |   |                                            |
|                   |                                                |   |                                            |
|Sunset             | OOD redirect to eth.org                        │   └── serpent.md                               |
|                   |  (/docs/smart-contracts/languages/             |                                                |
|                   |                                                |                                                |
|                   |                                                ├── clients                                      |
|                   |                                                |   |                                            |
|Sunset             | OOD redirect to eth.org                        │   ├── clients-tools-dapp-browsers-wallets...   |
[...]

PR #5897, opened 2022-04-04
migrate info on pow-mining from ethdocs and eth.wiki

jmcook1186:
## Description

This PR aims to migrate information from the ethdocs.org and eth.wiki sites to appropriate pages on ethereum.org so that the information is preserved when those websites are deprecated.

To this end, I have done the following:

1) Added some introductory material to the existing pow mining page
2) Added a brief high-level paragraph on mining algorithms to the existing pow-mining page
3) Added a new directory, `/mining-algorithms`, beneath `/consensus-mechanisms/pow`
4) Added a landing page and two subdirectories to `/mining-algorithms`: `/dagger-hashamoto` and `ethash`
5) Migrate info from ethdocs and eth.wiki pages on mining algorithms into `dagger-hashamoto/index.md` and `ethhash/index.md`
6) Update menus and links to make new pages visible.

## Related Issue

This resolves suggested PR 12 from #5690 and PR 3 from #5731


src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+    return a ^ b
+
+def isprime(x):
+    for i in range(2, int(x**0.5)):
+         if x % i == 0:
+             return False
+    return True
+```
+
+### Data Sizes
```
minimalsm, line 258:
```suggestion
### Data Sizes {#data-sizes}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+
+The following lookup tables provide approximately 2048 tabulated epochs of data sizes and cache sizes. They were generated with the _Mathematica_ function provided here:
```
minimalsm, line 260: > the _Mathematica_ function

What does this refer to?

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+Dagger Hashimoto was a precursor research implementation and specification for the mining algorithm for Ethereum, while it has been superceded by [Ethash](#ethash). The information on this page is for historical interest - it is not up to date with current Ethereum mining.
+
+Dagger Hashimoto aims to simultaneously satisfy two goals:
+
+1.  **ASIC-resistance**: the benefit from creating specialized hardware for the algorithm should be as small as possible, ideally to the point that even in an economy where ASICs have been developed the speedup is sufficiently small that it is still marginally profitable for users on ordinary computers to mine with spare CPU power.
+2.  **Light client verifiability**: a block should be relatively efficiently verifiable by a light client.
+
+With an additional modification, we also specify how to fulfill a third goal if desired, but at the cost of additional complexity:
+
+**Full chain storage**: mining should require storage of the complete blockchain state (due to the irregular structure of the Ethereum state trie, we anticipate that some pruning will be possible, particularly of some often-used contracts, but we want to minimize this).
```
minimalsm, line 20: This intro seems way too long to me. I think we can split it into:

- Intro
- Prerequisites
- Overview

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
+---
+title: Dagger-Hashamoto
+description: A detailed look at the Dagger-Hashamoto algorithm.
+lang: en
+sidebar: true
+incomplete: false
+---
+
+This page includes detailed information about the Dagger Hashamoto mining algorithm. Through a series of substanbtial upgrades, this algorithm was superceded by [Ethash](/ethash). However, this original version has historical significance and is included here because it was an important innovation in Ethereum's development. Proof-of-work mining itself will be deprecated in favour of proof-of-stake during the merge which is expected to happen in summer 2022. To understand the information in this page it is recommended to first read our introductory pages on [proof-of-work consensus](/developers/docs/consensus-mechanisms/pow) and [mining](/developers/docs/consensus-mechanisms/pow/mining).
```
minimalsm, line 9: > Proof-of-work mining itself will be deprecated in favour of proof-of-stake during the merge which is expected to happen in summer 2022.

I think this is unnecessary here. At very least, it is explained in the prerequisites.

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+Dagger Hashimoto was a precursor research implementation and specification for the mining algorithm for Ethereum, while it has been superceded by [Ethash](#ethash). The information on this page is for historical interest - it is not up to date with current Ethereum mining.
```
minimalsm, line 11: > The information on this page is for historical interest - it is not up to date with current Ethereum mining.

Given the closeness of the merge IMO we should just remove this

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
minimalsm: I actually think if we move the prerequisites below we can drop this paragraph entirely.

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
minimalsm, line 11:
```suggestion
## Prerequisites {#prerequisites}

To better understand this page, we recommend you first read up on [proof-of-work consensus](/developers/docs/consensus-mechanisms/pow), [mining](/developers/docs/consensus-mechanisms/pow/mining), and [mining algorithms](/developers/docs/consensus-mechanisms/pow/mining/mining-algorithms).
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
minimalsm, line 9:
```suggestion
Dagger Hashimoto was the original research implementation and specification for Ethereum's mining algorithm. Dagger Hashimoto was superseded by [Ethash](#ethash).
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+Dagger Hashimoto aims to simultaneously satisfy two goals:
+
+1.  **ASIC-resistance**: the benefit from creating specialized hardware for the algorithm should be as small as possible, ideally to the point that even in an economy where ASICs have been developed the speedup is sufficiently small that it is still marginally profitable for users on ordinary computers to mine with spare CPU power.
```
minimalsm, line 15: This gets kind of hard to read, do you think cutting this at the first sentence retains the sentiment here?

```suggestion
1.  **ASIC-resistance**: the benefit from creating specialized hardware for the algorithm should be as small as possible
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+lang: en
+sidebar: true
+incomplete: false
+---
+
+This page includes detailed information about the Dagger Hashamoto mining algorithm. Through a series of substanbtial upgrades, this algorithm was superceded by [Ethash](/ethash). However, this original version has historical significance and is included here because it was an important innovation in Ethereum's development. Proof-of-work mining itself will be deprecated in favour of proof-of-stake during the merge which is expected to happen in summer 2022. To understand the information in this page it is recommended to first read our introductory pages on [proof-of-work consensus](/developers/docs/consensus-mechanisms/pow) and [mining](/developers/docs/consensus-mechanisms/pow/mining).
+
+Dagger Hashimoto was a precursor research implementation and specification for the mining algorithm for Ethereum, while it has been superceded by [Ethash](#ethash). The information on this page is for historical interest - it is not up to date with current Ethereum mining.
+
+Dagger Hashimoto aims to simultaneously satisfy two goals:
```
minimalsm, line 13:
```suggestion
## Dagger Hashimoto {#dagger-hashimoto}

Dagger Hashimoto aims to satisfy two goals:
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+1.  **ASIC-resistance**: the benefit from creating specialized hardware for the algorithm should be as small as possible, ideally to the point that even in an economy where ASICs have been developed the speedup is sufficiently small that it is still marginally profitable for users on ordinary computers to mine with spare CPU power.
+2.  **Light client verifiability**: a block should be relatively efficiently verifiable by a light client.
```
minimalsm, line 16: Relative to what?

```suggestion
2.  **Light client verifiability**: a block should be efficiently verifiable by a light client.
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+With an additional modification, we also specify how to fulfill a third goal if desired, but at the cost of additional complexity:
+
+**Full chain storage**: mining should require storage of the complete blockchain state (due to the irregular structure of the Ethereum state trie, we anticipate that some pruning will be possible, particularly of some often-used contracts, but we want to minimize this).
+
+## DAG Generation
```
minimalsm, line 22:
```suggestion
## DAG Generation {#dag-generation}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+        x = encode_int(x)
+    return decode_int(utils.sha3(x))
+
+def dbl_sha3(x):
+    if isinstance(x, (int, long)):
+        x = encode_int(x)
+    return decode_int(utils.sha3(utils.sha3(x)))
+```
+
+### Parameters
```
minimalsm, line 61:
```suggestion
### Parameters {#parameters}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+      "diff": 2**14,                    # Difficulty (adjusted during block evaluation)
+      "epochtime": 100000,              # Length of an epoch in blocks (how often the dataset is updated)
+      "k": 1,                           # Number of parents of a node
+      "w": w,                          # Used for modular exponentiation hashing
+      "accesses": 200,                  # Number of dataset accesses during hashimoto
+      "P": SAFE_PRIME_512               # Safe Prime for hashing and random number generation
+}
+```
+
+`P` in this case is a prime chosen such that `log₂(P)` is just slightly less than 512, which corresponds to the 512 bits we have been using to represent our numbers. Note that only the latter half of the DAG actually needs to be stored, so the de-facto RAM requirement starts off at 1 GB and grows by 441 MB per year.
```
minimalsm, line 83:
```suggestion
`P` in this case is a prime chosen such that `log₂(P)` is just slightly less than 512, which corresponds to the 512 bits we have been using to represent our numbers. Note that only the latter half of the DAG actually needs to be stored, so the de-facto RAM requirement starts at 1 GB and grows by 441 MB per year.
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Dagger graph building
```
minimalsm, line 85:
```suggestion
### Dagger graph building {#dagger-graph-building}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+            x ^= o[x % i]
+        o.append(pow(x, params["w"], P))
+    return o
+```
+
+Essentially, it starts off a graph as a single node, `sha3(seed)`, and from there starts sequentially adding on other nodes based on random previous nodes. When a new node is created, a modular power of the seed is computed to randomly select some indices less than `i` (using `x % i` above), and the values of the nodes at those indices are used in a calculation to generate a new a value for `x`, which is then fed into a small proof of work function (based on XOR) to ultimately generate the value of the graph at index `i`. The rationale behind this particular design is to force sequential access of the DAG; the next value of the DAG that will be accessed cannot be determined until the current value is known. Finally, modular exponentiation is used to further hash the result.
+
+This algorithm relies on several results from number theory. See the appendix below for a discussion.
+
+## Light Client Evaluation
```
minimalsm, line 106:
```suggestion
## Light client evaluation {#light-client-evaluation}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+                x ^= quick_calc_cached(x % p)
+            cache[p] = pow(x, w, P)
+        return cache[p]
+
+    return quick_calc_cached(p)
+```
+
+Essentially, it is simply a rewrite of the above algorithm that removes the loop of computing the values for the entire DAG and replaces the earlier node lookup with a recursive call or a cache lookup. Note that for `k=1` the cache is unnecessary, although a further optimization actually precomputes the first few thousand values of the DAG and keeps that as a static cache for computations; see the appendix for a code implementation of this.
+
+## Double Buffer of DAGs
```
minimalsm, line 134:
```suggestion
## Double buffer of DAGs {#double-buffer}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+    o = [init]
+    for i in range(1, length):
+        x = picker = (picker * init) % P
+        for _ in range(params["k"]):
+            x ^= o[x % i]
+        o.append(pow(x, params["w"], P))
+    return o
+```
+
+Essentially, it starts off a graph as a single node, `sha3(seed)`, and from there starts sequentially adding on other nodes based on random previous nodes. When a new node is created, a modular power of the seed is computed to randomly select some indices less than `i` (using `x % i` above), and the values of the nodes at those indices are used in a calculation to generate a new a value for `x`, which is then fed into a small proof of work function (based on XOR) to ultimately generate the value of the graph at index `i`. The rationale behind this particular design is to force sequential access of the DAG; the next value of the DAG that will be accessed cannot be determined until the current value is known. Finally, modular exponentiation is used to further hash the result.
```
minimalsm, line 102:
```suggestion
Essentially, it starts off a graph as a single node, `sha3(seed)`, and from there starts sequentially adding on other nodes based on random previous nodes. When a new node is created, a modular power of the seed is computed to randomly select some indices less than `i` (using `x % i` above), and the values of the nodes at those indices are used in a calculation to generate a new a value for `x`, which is then fed into a small proof of work function (based on XOR) to ultimately generate the value of the graph at index `i`. The rationale behind this particular design is to force sequential access of the DAG; the next value of the DAG that will be accessed cannot be determined until the current value is known. Finally, modular exponentiation hashes the result further.
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+This algorithm relies on several results from number theory. See the appendix below for a discussion.
+
+## Light Client Evaluation
+
+The intent of the above graph construction is to allow each individual node in the graph can be reconstructed by computing a subtree of only a small number of nodes, and requiring only a small amount of auxiliary memory. Note that with k=1, the subtree is only a chain of values going up to the first element in the DAG.
```
minimalsm, line 108:
```suggestion
The above graph construction intends to allow each node in the graph to be reconstructed by computing a subtree of only a small number of nodes and requiring only a small amount of auxiliary memory. Note that with k=1, the subtree is only a chain of values going up to the first element in the DAG.
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+                             params, decode_int(block.prevhash), nonce)
+    return result * params["diff"] < 2**256
+```
+
+Also, note that Dagger Hashimoto imposes additional requirements on the block header:
+
+- For two-layer verification to work, a block header must have both the nonce and the middle value pre-sha3
+- Somewhere, a block header must store the sha3 of the current seedset
+
+## Appendix
```
minimalsm, line 255: Not a blocker but we could use `<fn>` tags to implement the appendix [similar to our NFT page](https://ethereum.org/en/nft/#fn-1)

minimalsm, line 255:
```suggestion
## Appendix {#appendix}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+```
+
+Essentially, it is simply a rewrite of the above algorithm that removes the loop of computing the values for the entire DAG and replaces the earlier node lookup with a recursive call or a cache lookup. Note that for `k=1` the cache is unnecessary, although a further optimization actually precomputes the first few thousand values of the DAG and keeps that as a static cache for computations; see the appendix for a code implementation of this.
+
+## Double Buffer of DAGs
+
+In a full client, a [_double buffer_](https://en.wikipedia.org/wiki/Multiple_buffering) of 2 DAGs produced by the above formula is used. The idea is that DAGs are produced every `epochtime` number of blocks according to the params above. The client does not use the latest DAG produced, but the previous one. The benefit of this is that it allows the DAGs to be replaced over time without needing to incorporate a step where miners must suddenly recompute all of the data. Otherwise, there is the potential for an abrupt temporary slowdown in chain processing at regular intervals and dramatically increasing centralization and thus 51% attack risks within those few minutes before all data is recomputed.
```
minimalsm, line 136:
```suggestion
In a full client, a [_double buffer_](https://wikipedia.org/wiki/Multiple_buffering) of 2 DAGs produced by the above formula is used. The idea is that DAGs are produced every `epochtime` number of blocks according to the params above. Instead of the client using the latest DAG produced, it uses the one previous. The benefit of this is that it allows the DAGs to be replaced over time without needing to incorporate a step where miners must suddenly recompute all of the data. Otherwise, there is the potential for an abrupt temporary slowdown in chain processing at regular intervals and dramatically increasing centralization. Thus 51% attack risks within those few minutes before all data are recomputed.
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+The algorithm used to generate the actual set of DAGs used to compute the work for a block is as follows:
```
minimalsm, line 138:
```suggestion
The algorithm used to generate the set of DAGs used to compute the work for a block is as follows:
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+        return {"front": {"dag": produce_dag(params, seedset["front_hash"], dagsz),
+                          "block_number": 0}}
+    else:
+        return {"front": {"dag": produce_dag(params, seedset["front_hash"], dagsz),
+                          "block_number": seedset["front_number"]},
+                "back": {"dag": produce_dag(params, seedset["back_hash"], dagsz),
+                         "block_number": seedset["back_number"]}}
+```
+
+## Hashimoto
```
minimalsm, line 175:
```suggestion
## Hashimoto {#hashimoto}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+The idea behind the original Hashimoto is to use the blockchain as a dataset, performing a computation which selects N indices from the blockchain, gathers the transactions at those indices, performs an XOR of this data, and returns the hash of the result. Thaddeus Dryja's original algorithm, translated to Python for consistency, is as follows:
```
minimalsm, line 177:
```suggestion
The idea behind the original Hashimoto is to use the blockchain as a dataset, performing a computation that selects N indices from the blockchain, gathers the transactions at those indices, performs an XOR of this data, and returns the hash of the result. Thaddeus Dryja's original algorithm, translated to Python for consistency, is as follows:
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+    hash_output_A = sha256(prev_hash + merkle_root + nonce)
+    txid_mix = 0
+    for i in range(64):
+        shifted_A = hash_output_A >> i
+        transaction = shifted_A % len(list_of_transactions)
+        txid_mix ^= list_of_transactions[transaction] << i
+    return txid_max ^ (nonce << 192)
+```
+
+Unfortunately, while Hashimoto is considered RAM hard, it relies on 256-bit arithmetic, which has considerable computational overhead. To address this issue, dagger hashimoto only uses the least significant 64 bits when indexing its dataset.
```
minimalsm, line 190:
```suggestion
Unfortunately, while Hashimoto is considered RAM hard, it relies on 256-bit arithmetic, which has considerable computational overhead. However, Dagger Hashimoto only uses the least significant 64 bits when indexing its dataset to address this issue.
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+```python
+def hashimoto(dag, dagsize, params, header, nonce):
+    m = dagsize / 2
+    mix = sha3(encode_int(nonce) + header)
+    for _ in range(params["accesses"]):
+        mix ^= dag[m + (mix % 2**64) % m]
+    return dbl_sha3(mix)
+```
+
+The use of double sha3 allows for a form of zero-data, near-instant pre-verification, verifying only that a correct intermediate value was provided. This outer layer of PoW is highly ASIC-friendly and fairly weak, but exists to make DDoS even more difficult since that small amount of work must be done in order to produce a block that will not be rejected immediately. Here is the light-client version:
```
minimalsm, line 201: Should this be SHA-3?

minimalsm: > PoW

*proof-of-work

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+    m = dagsize // 2
+    mix = sha3(nonce + header)
+    for _ in range(params["accesses"]):
+        mix ^= quick_calc(params, seed, m + (mix % 2**64) % m)
+    return dbl_sha3(mix)
+```
+
+## Mining and Verifying
+
+Now, let us put it all together into the mining algo:
```
minimalsm, line 214:
```suggestion
Now, let us put it all together into the mining algorithm:
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+```python
+def quick_hashimoto(seed, dagsize, params, header, nonce):
+    m = dagsize // 2
+    mix = sha3(nonce + header)
+    for _ in range(params["accesses"]):
+        mix ^= quick_calc(params, seed, m + (mix % 2**64) % m)
+    return dbl_sha3(mix)
+```
+
+## Mining and Verifying
```
minimalsm, line 212:
```suggestion
## Mining and verifying {#mining-and-verifying}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+Given that `P` is prime, then an appropriate `w` for a modular exponentation hashing function can be chosen using the following result:
+
+> Observation 3. Let `P` be a prime; `w` and `P-1` are relatively prime if and only if for all `a` and `b` in `ℤ/Pℤ`:<center>`aʷ mod P ≡ bʷ mod P` if and only if `a mod P ≡ b mod P`</center>
+
+Thus, given that `P` is prime and `w` is relatively prime to `P-1`, we have that `|{pow(x, w, P) : x ∈ ℤ}| = P`, implying that the hashing function has the minimal collision rate possible.
+
+In the special case that `P` is a safe prime as we have selected, then `P-1` only has factors 1, 2, `(P-1)/2` and `P-1`. Since `P` > 7, we know that 3 is relatively prime to `P-1`, hence `w=3` satisfies the above proposition.
+
+## More Efficient Cache-based Evaluation Algorithm
```
minimalsm, line 299:
```suggestion
## More efficient cache-based evaluation algorithm {#cache-based-evaluation}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+From the above proposition, we can recognize that iterating `(picker * init) % P` will have a cycle length of at least `(P-1)/2`. This is because we selected `P` to be a safe prime approximately equal to be a higher power of two, and `init` is in the interval `[2,2**256+1]`. Given the magnitude fo `P`, we should never expect a cycle from modular exponentiation.
+
+When we are assigning the first cell in the DAG (the variable labeled `init`), we compute `pow(sha3(seed) + 2, 3, P)`. At first glance, this does not guarantee that the result is neither `1` nor `P-1`. However, since `P-1` is a safe prime, we have the following additional assuance, which is a corollary of Observation 1:
+
+> Observation 2. Let `x` be a member of the multiplicative group `ℤ/Pℤ` for a safe prime `P`, and let `w` be a natural number. If `x mod P ≠ 1 mod P` and `x mod P ≠ P-1 mod P`, as well as `w mod P ≠ P-1 mod P` and `w mod P ≠ 0 mod P`, then `xʷ mod P ≠ 1 mod P` and `xʷ mod P ≠ P-1 mod P`
+
+[lagrange]: https://en.wikipedia.org/wiki/Lagrange%27s_theorem_(group_theory)
+
+### Modular Exponentiation as a Hash Function
```
minimalsm, line 287:
```suggestion
### Modular exponentiation as a hash function {#modular-exponentiation}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+Also, note that Dagger Hashimoto imposes additional requirements on the block header:
+
+- For two-layer verification to work, a block header must have both the nonce and the middle value pre-sha3
+- Somewhere, a block header must store the sha3 of the current seedset
+
+## Appendix
+
+As noted above, the RNG used for DAG generation relies on some results from number theory. First, we provide assurance that the Lehmer RNG that is the basis for the `picker` variable has a wide period. Second, we show that `pow(x,3,P)` will not map `x` to `1` or `P-1` provided `x ∈ [2,P-2]` to start. Finally, we show that `pow(x,3,P)` has a low collision rate when treated as a hashing function.
+
+### Lehmer Random Number Generator
```
minimalsm, line 259:
```suggestion
### Lehmer random number generator {#lehmer-random-number}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+The general route that the algorithm takes is as follows:
+
+1. There exists a **seed** which can be computed for each block by scanning through the block headers up until that point.
+2. From the seed, one can compute a **16 MB pseudorandom cache**. Light clients store the cache.
+3. From the cache, we can generate a **1 GB dataset**, with the property that each item in the dataset depends on only a small number of items from the cache. Full clients and miners store the dataset. The dataset grows linearly with time.
+4. Mining involves grabbing random slices of the dataset and hashing them together. Verification can be done with low memory by using the cache to regenerate the specific pieces of the dataset that you need, so you only need to store the cache.
+
+The large dataset is updated once every 30000 blocks, so the vast majority of a miner's effort will be reading the dataset, not making changes to it.
+
+## Definitions
```
minimalsm, line 20:
```suggestion
## Definitions {#definitions}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+CACHE_MULTIPLIER=1024             # Size of the DAG relative to the cache
+EPOCH_LENGTH = 30000              # blocks per epoch
+MIX_BYTES = 128                   # width of mix
+HASH_BYTES = 64                   # hash length in bytes
+DATASET_PARENTS = 256             # number of parents of each dataset element
+CACHE_ROUNDS = 3                  # number of rounds in cache production
+ACCESSES = 64                     # number of accesses in hashimoto loop
+```
+
+### A note regarding "SHA3" hashes described in this specification
```
minimalsm, line 39:
```suggestion
### The use of 'SHA3' {#sha3}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+
+Ethereum's development coincided with the development of the SHA3 standard, and the
+standards process made a late change in the padding of the finalized hash algorithm, so that Ethereum's
+"sha3_256" and "sha3_512" hashes are not standard sha3 hashes, but a variant often referred
+to as "Keccak-256" and "Keccak-512" in other contexts. See discussion, e.g. [here](https://github.com/ethereum/EIPs/issues/59), [here](http://ethereum.stackexchange.com/questions/550/which-cryptographic-hash-function-does-ethereum-use), or [here](http://bitcoin.stackexchange.com/questions/42055/what-is-the-approach-to-calculate-an-ethereum-address-from-a-256-bit-private-key/42057#42057).
+
+Please keep that in mind as "sha3" hashes are referred to in the description of the algorithm below.
+
+## Parameters
```
minimalsm, line 48:
```suggestion
## Parameters {#parameters}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+    sz = DATASET_BYTES_INIT + DATASET_BYTES_GROWTH * (block_number // EPOCH_LENGTH)
+    sz -= MIX_BYTES
+    while not isprime(sz / MIX_BYTES):
+        sz -= 2 * MIX_BYTES
+    return sz
+```
+
+Tables of dataset and cache size values are provided in the appendix.
+
+## Cache Generation
```
minimalsm, line 70:
```suggestion
## Cache generation {#cache-generation}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+            o[i] = sha3_512(map(xor, o[(i-1+n) % n], o[v]))
+
+    return o
+```
+
+[sergio2014]: http://www.hashcash.org/papers/memohash.pdf
+
+The cache production process involves first sequentially filling up 32 MB of memory, then performing two passes of Sergio Demian Lerner's _RandMemoHash_ algorithm from [_Strict Memory Hard Hashing Functions_ (2014)](http://www.hashcash.org/papers/memohash.pdf). The output is a set of 524288 64-byte values.
+
+## Data aggregation function
```
minimalsm, line 96:
```suggestion
## Data aggregation function {#date-aggregation-function}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+```python
+FNV_PRIME = 0x01000193
+
+def fnv(v1, v2):
+    return ((v1 * FNV_PRIME) ^ v2) % 2**32
+```
+
+Please note, even the yellow paper specifies fnv as v1\*(FNV_PRIME ^ v2), all current implementations consistently use the above definition.
+
+## Full dataset calculation
```
minimalsm, line 109:
```suggestion
## Full dataset calculation {#full-dataset-calculation}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+```
+
+Essentially, we combine data from 256 pseudorandomly selected cache nodes, and hash that to compute the dataset node. The entire dataset is then generated by:
+
+```python
+def calc_dataset(full_size, cache):
+    return [calc_dataset_item(cache, i) for i in range(full_size // HASH_BYTES)]
+```
+
+## Main Loop
```
minimalsm, line 135:
```suggestion
## Main loop {#main-loop}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+ def get_seedhash(block):
+     s = '\x00' * 32
+     for i in range(block.number // EPOCH_LENGTH):
+         s = serialize_hash(sha3_256(s))
+     return s
+```
+
+Note that for smooth mining and verifying, we recommend pre-computing future seedhashes and datasets in a separate thread.
+
+## Appendix
```
minimalsm, line 206:
```suggestion
## Appendix {#appendix}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+    # zero-pad target to compare with hash on the same digit
+    target = zpad(encode_int(2**256 // difficulty), 64)[::-1]
+    from random import randint
+    nonce = randint(0, 2**64)
+    while hashimoto_full(full_size, dataset, header, nonce) > target:
+        nonce = (nonce + 1) % 2**64
+    return nonce
+```
+
+## Defining the Seed Hash
```
minimalsm, line 192:
```suggestion
## Defining the seed hash {#seed-hash}
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1009 @@
...
+
+def hashimoto_full(full_size, dataset, header, nonce):
+    return hashimoto(header, nonce, full_size, lambda x: dataset[x])
+```
+
+Essentially, we maintain a "mix" 128 bytes wide, and repeatedly sequentially fetch 128 bytes from the full dataset and use the `fnv` function to combine it with the mix. 128 bytes of sequential access are used so that each round of the algorithm always fetches a full page from RAM, minimizing translation lookaside buffer misses which ASICs would theoretically be able to avoid.
+
+If the output of this algorithm is below the desired target, then the nonce is valid. Note that the extra application of `sha3_256` at the end ensures that there exists an intermediate nonce which can be provided to prove that at least a small amount of work was done; this quick outer PoW verification can be used for anti-DDoS purposes. It also serves to provide statistical assurance that the result is an unbiased, 256 bit number.
+
+## Mining
```
minimalsm, line 177:
```suggestion
## Mining {#mining}
```

minimalsm (changes requested):
Looking great! Found it very well put together and educational for me.

Comments are mostly around site conventions or some simplifications I think we could make, but I'll leave those up to you.

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
wackerow: Looks good, but think we want `/ethash/` as the link, not `#ethash`

```suggestion
Dagger Hashimoto was the original research implementation and specification for Ethereum's mining algorithm. Dagger Hashimoto was superseded by [Ethash](/ethash/).
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+The order of `x` cannot be `1`, since by Fermat's Little Theorem we have:
+
+<center><pre>x<sup>P-1</sup> mod P ≡ 1</pre></center>
+Hence `x` must be a multiplicative identity of `ℤ/nℤ`, which is unique.  Since we assumed that `x ≠ 1` by assumption, this is not possible.
+
+The order of `x` cannot be `2` unless `x = P-1`, since this would violate that `P` is prime.
+
+<div align="right">◻</div>
+
```
wackerow, line 278: ?
```suggestion
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+## Appendix
+
+As noted above, the RNG used for DAG generation relies on some results from number theory. First, we provide assurance that the Lehmer RNG that is the basis for the `picker` variable has a wide period. Second, we show that `pow(x,3,P)` will not map `x` to `1` or `P-1` provided `x ∈ [2,P-2]` to start. Finally, we show that `pow(x,3,P)` has a low collision rate when treated as a hashing function.
+
+### Lehmer Random Number Generator
+
+While the `produce_dag` function does not need to produce unbiased random numbers, a potential threat is that `seed**i % P` only takes on a handful of values. This could provide an advantage to miners recognizing the pattern over those that do not.
+
+To avoid this, a result from number theory is appealed to. A [_Safe Prime_](https://en.wikipedia.org/wiki/Safe_prime) is defined to be a prime `P` such that `(P-1)/2` is also prime. The _order_ of a member `x` of the [multiplicative group](https://en.wikipedia.org/wiki/Multiplicative_group_of_integers_modulo_n) `ℤ/nℤ` is defined to be the minimal `m` such that <center><pre>xᵐ mod P ≡ 1</pre></center>
```
wackerow, line 263: Is there a need for centering here?
```suggestion
To avoid this, a result from number theory is appealed to. A [_Safe Prime_](https://en.wikipedia.org/wiki/Safe_prime) is defined to be a prime `P` such that `(P-1)/2` is also prime. The _order_ of a member `x` of the [multiplicative group](https://en.wikipedia.org/wiki/Multiplicative_group_of_integers_modulo_n) `ℤ/nℤ` is defined to be the minimal `m` such that <pre>xᵐ mod P ≡ 1</pre>
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+Given these definitions, we have:
+
+> Observation 1. Let `x` be a member of the multiplicative group `ℤ/Pℤ` for a safe prime `P`. If `x mod P ≠ 1 mod P` and `x mod P ≠ P-1 mod P`, then the order of `x` is either `P-1` or `(P-1)/2`.
+
+_Proof_. Since `P` is a safe prime, then by [Lagrange's Theorem][lagrange] we have that the order of `x` is either `1`, `2`, `(P-1)/2`, or `P-1`.
+
+The order of `x` cannot be `1`, since by Fermat's Little Theorem we have:
+
+<center><pre>x<sup>P-1</sup> mod P ≡ 1</pre></center>
```
wackerow, line 272: Same here?
```suggestion
<pre>x<sup>P-1</sup> mod P ≡ 1</pre>
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,327 @@
...
+
+<div align="right">◻</div>
+
+From the above proposition, we can recognize that iterating `(picker * init) % P` will have a cycle length of at least `(P-1)/2`. This is because we selected `P` to be a safe prime approximately equal to be a higher power of two, and `init` is in the interval `[2,2**256+1]`. Given the magnitude fo `P`, we should never expect a cycle from modular exponentiation.
+
+When we are assigning the first cell in the DAG (the variable labeled `init`), we compute `pow(sha3(seed) + 2, 3, P)`. At first glance, this does not guarantee that the result is neither `1` nor `P-1`. However, since `P-1` is a safe prime, we have the following additional assuance, which is a corollary of Observation 1:
+
+> Observation 2. Let `x` be a member of the multiplicative group `ℤ/Pℤ` for a safe prime `P`, and let `w` be a natural number. If `x mod P ≠ 1 mod P` and `x mod P ≠ P-1 mod P`, as well as `w mod P ≠ P-1 mod P` and `w mod P ≠ 0 mod P`, then `xʷ mod P ≠ 1 mod P` and `xʷ mod P ≠ P-1 mod P`
+
+[lagrange]: https://en.wikipedia.org/wiki/Lagrange%27s_theorem_(group_theory)
```
wackerow, line 285: This syntax does not render
```suggestion
```

wackerow (changes requested):
@jmcook1186 🙏🏻 Thank you again for these. Great addition to fill out the background of PoW, appreciate your work on this. This seems super close.. bunch of code suggestions left, mostly non-controversial I feel.

Let me know if you want a hand tuning this up, we'll try to get this in soon =)

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
jmcook1186: I think this comment is resolved by latest commits

src/content/developers/docs/consensus-mechanisms/pow/mining/index.md
```diff
@@ -14,7 +14,11 @@ To better understand this page, we recommend you first read up on [transactions]
...
+
+Mining ether = Securing the Network = Verifying Computation
+
+Ethereum, like Bitcoin, currently uses a [proof-of-work (PoW)](/developers/docs/consensus-mechanisms/pow/) consensus mechanism. Mining is the lifeblood of proof-of-work. Ethereum miners - computers running software - using their time and computation power to process transactions and produce blocks. The Ethereum blockchain is in many ways similar to the Bitcoin blockchain, although it does have some differences. The main difference between Ethereum and Bitcoin with regard to the blockchain architecture is that, unlike Bitcoin, Ethereum blocks contain a copy of both the transaction list and the most recent state (the root hash of the merkle patricia trie encoding the state to be more precise). Aside from that, two other values, the block number and the difficulty, are also stored in the block.
```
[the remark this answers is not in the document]
jmcook1186: ok - removed

src/content/developers/docs/consensus-mechanisms/pow/mining/index.md
```diff
@@ -60,6 +64,12 @@ Watch Austin walk you through mining and the proof-of-work blockchain.
 
 <YouTube id="zcX7OJ-L8XQ" />
 
+## The mining algorithm
+
+The Ethereum mining algorithm has undergone several upgrades since its inception. The original algorithm, "Dagger Hashamoto" was based around the provision of a large, transient, randomly generated dataset which forms a [Directed Acyclic Graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph) (the Dagger-part), with miners attempting to solve a particular constraint on it, partly determined through a block’s header-hash. This algorithm was novel because it had high memory-access bandwidth requirements but could be run using a modest processor, making it GPU-friendly but resistant to the type of ASIC-driven hardware arms race that could pose a centralization risk. After substantial upgrades to the algorithm, it was renamed to "Ethash". Ethash is the mining algorithm Ethereum miners use today.
```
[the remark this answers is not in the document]
jmcook1186: tbh I'm struggling to insert something here that doesn't feel like a tangent to the main point of the paragraph. Instead I linked out to an external article - ok?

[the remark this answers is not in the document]
wackerow: Personally cool with this, and it doesn't derail the topic too much this way.

wackerow (approved):
💪🏻 Thanks again @jmcook1186!

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1007 @@
+---
+title: Ethash
+description: A detailed look at the Ethash algorithm.
+lang: en
+sidebar: true
+incomplete: false
+---
+
+[Ethash](https://github.com/ethereum/wiki/wiki/Ethash) is a modified version of the [Dagger-Hashimoto](/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto) algorithm. Ethash PoW is memory hard, making it ASIC resistant. Memory hardness is achieved with a proof of work algorithm that requires choosing subsets of a fixed resource dependent on the nonce and block header. This resource (a few gigabytes in size) is called a DAG. The DAG is totally different every 30000 blocks, a 125-hour window called an epoch (roughly 5.2 days) and takes a while to generate. Since the DAG only depends on block height, it can be pre-generated but if it's not, the client needs to wait until the end of this process to produce a block. If clients do not pre-generate and cache DAGs ahead of time the network may experience massive block delay on each epoch transition. Note that the DAG does not need to be generated for verifying the PoW essentially allowing for verification with both low CPU and small memory.
```
MariusVanDerWijden, line 9: Ethash is not really asic resistant anymore
"totally different"

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/index.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Mining algorithms
+description: A detailed look at the algorithms used for Ethereum mining.
+lang: en
+sidebar: true
+incomplete: false
+---
+
+Ethereum mining has used two mining algorithms, Dagger Hashamoto and Ethash. Dagger Hashamoto is no longer used to mine Ethereum, being superseded by Ethash. However, it has historical significance as an important innovation in Ethereum's development. Proof-of-work mining itself will be deprecated in favor of proof-of-stake during [The Merge](/merge/), which is forecast to happen in Q3-Q4 2022.
```
MariusVanDerWijden, line 9: Dagger-Hash*i*moto 
Dagger-Hashimoto was never used on mainnet, it was only ever the basis for ethash (and a research direction pre-mainnet afaik): https://github.com/ethereum/wiki/wiki/Dagger-Hashimoto

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/index.md
```diff
@@ -0,0 +1,33 @@
...
+title: Mining algorithms
+description: A detailed look at the algorithms used for Ethereum mining.
+lang: en
+sidebar: true
+incomplete: false
+---
+
+Ethereum mining has used two mining algorithms, Dagger Hashamoto and Ethash. Dagger Hashamoto is no longer used to mine Ethereum, being superseded by Ethash. However, it has historical significance as an important innovation in Ethereum's development. Proof-of-work mining itself will be deprecated in favor of proof-of-stake during [The Merge](/merge/), which is forecast to happen in Q3-Q4 2022. 
+
+The fundamental idea of both algorithms is that a miner tries to find a nonce input using brute force computation so that the result is below a certain difficulty threshold. This difficulty threshold can be dynamically adjusted, allowing block production to happen at a regular tempo.
```
MariusVanDerWijden, line 11:
```suggestion
The fundamental idea of both algorithms is that a miner tries to find a nonce input using brute force computation so that the result is below a certain difficulty threshold. This difficulty threshold can be dynamically adjusted, allowing block production to happen at a regular interval.
```

src/content/developers/docs/consensus-mechanisms/pow/mining/index.md
```diff
@@ -60,6 +64,12 @@ Watch Austin walk you through mining and the proof-of-work blockchain.
 
 <YouTube id="zcX7OJ-L8XQ" />
 
+## The mining algorithm {#mining-algorithm}
+
+The Ethereum mining algorithm has undergone several upgrades since its inception. The original algorithm, "Dagger Hashamoto" was based around the provision of a large, transient, randomly generated dataset which forms a [Directed Acyclic Graph](https://en.wikipedia.org/wiki/Directed_acyclic_graph) (the Dagger-part), with miners attempting to solve a particular constraint on it, partly determined through a block’s header-hash. This algorithm was novel because it had high memory-access bandwidth requirements but could be run using a modest processor, making it GPU-friendly but resistant to the type of ASIC-driven hardware arms race that could pose a centralization risk (more on [problems with ASICS](https://www.investopedia.com/investing/why-centralized-crypto-mining-growing-problem/)). After substantial upgrades to the algorithm, it was renamed to "Ethash". Ethash is the mining algorithm Ethereum miners use today.
```
MariusVanDerWijden, line 69: Again the mining algorithm has never been upgraded on mainnet, dagger-hashimoto was only the research base on which ethash was constructed
Also dagger-hash**i**moto

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
jmcook1186: thanks, updated

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash/index.md
```diff
@@ -0,0 +1,1007 @@
...
+lang: en
+sidebar: true
+incomplete: false
+---
+
+[Ethash](https://github.com/ethereum/wiki/wiki/Ethash) is a modified version of the [Dagger-Hashimoto](/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto) algorithm. Ethash PoW is memory hard, which was thought to make the algorithm ASIC resistant, but ASIC Ethash-mining has since been shown to be possible. Memory hardness is achieved with a proof of work algorithm that requires choosing subsets of a fixed resource dependent on the nonce and block header. This resource (a few gigabytes in size) is called a DAG. The DAG is changed every 30000 blocks, a 125-hour window called an epoch (roughly 5.2 days) and takes a while to generate. Since the DAG only depends on block height, it can be pre-generated but if it's not, the client needs to wait until the end of this process to produce a block. If clients do not pre-generate and cache DAGs ahead of time the network may experience massive block delay on each epoch transition. Note that the DAG does not need to be generated for verifying the PoW essentially allowing for verification with both low CPU and small memory.
```
minimalsm, line 9:
```suggestion
[Ethash](https://github.com/ethereum/wiki/wiki/Ethash) is a modified version of the [Dagger-Hashimoto](/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto) algorithm. Ethash proof-of-work is [memory hard](https://wikipedia.org/wiki/Memory-hard_function), which was thought to make the algorithm ASIC resistant, but ASIC Ethash-mining has since been shown to be possible. Memory hardness is achieved with a proof of work algorithm that requires choosing subsets of a fixed resource dependent on the nonce and block header. This resource (a few gigabytes in size) is called a DAG. The DAG is changed every 30000 blocks, a 125-hour window called an epoch (roughly 5.2 days) and takes a while to generate. Since the DAG only depends on block height, it can be pre-generated but if it's not, the client needs to wait until the end of this process to produce a block. If clients do not pre-generate and cache DAGs ahead of time the network may experience massive block delay on each epoch transition. Note that the DAG does not need to be generated for verifying the proof-of-work essentially allowing for verification with both low CPU and small memory.
```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/index.md
```diff
@@ -0,0 +1,33 @@
...
+
+More on [Dagger-Hashimoto](/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashimoto).
+
+## Ethash {#ethash}
+
+Ethash is Ethereum's current mining algorithm. Ethash was effectively a new name given to a specific version of Dagger-Hashimoto after the algorithm got significantly updated, whilst still inheriting the fundamental principles of its predecessor. Ethereum mainnet has only ever used Ethash - Dagger Hashimoto was an R&D version of the mining algorithm that was superseded before mining started on Ethereum mainnet.
+
+[More on Ethash](/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash).
```
minimalsm, line 33:
```suggestion
## Ethash {#ethash}

Ethash is Ethereum's current mining algorithm. Ethash was effectively a new name given to a specific version of Dagger-Hashimoto after the algorithm got significantly updated, whilst still inheriting the fundamental principles of its predecessor. Ethereum mainnet has only ever used Ethash - Dagger Hashimoto was an R&D version of the mining algorithm that was superseded before mining started on Ethereum mainnet.

[More on Ethash](/developers/docs/consensus-mechanisms/pow/mining-algorithms/ethash).

## Further reading {#further-reading}

_Know of a community resource that helped you? Edit this page and add it!_

```

src/content/developers/docs/consensus-mechanisms/pow/mining-algorithms/dagger-hashamoto/index.md
```diff
@@ -0,0 +1,328 @@
...
+                             params, decode_int(block.prevhash), nonce)
+    return result * params["diff"] < 2**256
+```
+
+Also, note that Dagger-Hashimoto imposes additional requirements on the block header:
+
+- For two-layer verification to work, a block header must have both the nonce and the middle value pre-sha3
+- Somewhere, a block header must store the sha3 of the current seedset
+
+## Appendix {#appendix}
```
minimalsm, line 259:
```suggestion
- Somewhere, a block header must store the sha3 of the current seedset

## Further reading {#further-reading}

_Know of a community resource that helped you? Edit this page and add it!_

## Appendix {#appendix}
```

minimalsm (approved):
LGTM :) Pushed some commits to add further reading section to each page. Merging now 🚀

Some low priority things we should tidy up at some point:

- Memory hard - should we explain what this means? (I’ve added a link out to wikipedia for now)
- Appendix - We should link to appendix using footnotes (see NFT page for example of this on site)
- EthHash page needs prerequisites
- UI bug introduced in our developer docs sidenav by the depth (I will create an issue)

**Dagger Hashamoto page**

“## More efficient cache-based evaluation algorithm {#cache-based-evaluation}” section

- Providing a sentence for context here would be good

PR #5911, opened 2022-04-05
Overwrite Persian numbering for StatsBoxes

minimalsm:
## Description
Context: As noted in #5490, we were mixing Persian and Hindu-Arabic numerals. After consulting Farsi speakers, we realised this was undesirable.

**This PR overwrites the number formatting of the Farsi language on our StatsBox components to use Hindu-Arabic formatting.**

- Updates homepage StatsBox
- Updates layer-2 page StatsBox

## Additional context
- We should do the same for our staking pages

## Preview
- https://ethereumorgwebsitedev01-overwritepersiannumbers.gtsb.io/fa/
- https://ethereumorgwebsitedev01-overwritepersiannumbers.gtsb.io/fa/layer-2

## Related Issue
#5490 #5748

### DIFF
--- a/src/utils/formatLocaleNumbers.js
+++ b/src/utils/formatLocaleNumbers.js
@@ -1,0 +1,9 @@
+// Overwrites the default Persian numbering of the Farsi language to use Hindu-Arabic numerals (0-9)
+// Context: https://github.com/ethereum/ethereum-org-website/pull/5490#pullrequestreview-892596553
+export const formatLocaleNumbers = (locale) => {
+  if (locale === "fa") {
+    return "en"
+  }
+
+  return intl.locale
+}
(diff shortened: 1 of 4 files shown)


src/utils/formatLocaleNumbers.js
```diff
@@ -0,0 +1,9 @@
+// Overwrites the default Persian numbering of the Farsi language to use Hindu-Arabic numerals (0-9)
+// Context: https://github.com/ethereum/ethereum-org-website/pull/5490#pullrequestreview-892596553
+export const formatLocaleNumbers = (locale) => {
```
samajammin, line 3: Not opposed to a new file but we do have an existing `translations.js` utils file that I feel like this function could fit within, given it deals with language formatting.

samajammin: Also given I'm already nitpicking 😅 I think the name could be a bit clearer - it's slightly misleading since this function doesn't actually format anything. Maybe something like `getLocaleNumberFormat`?

minimalsm: > Not opposed to a new file but we do have an existing `translations.js` utils file that I feel like this function could fit within, given it deals with language formatting.

Good point. Moved it in there.

> Also given I'm already nitpicking 😅 I think the name could be a bit clearer - it's slightly misleading since this function doesn't actually format anything. Maybe something like getLocaleNumberFormat?

Not nitpicking at all it was a bad name 😬 updated it to `getLocaleForNumberFormat`.

Kinou90100 (commented):
#5911
#5910
@Christoc90
@Kinou90100

PR #5915, opened 2022-04-05
Option components a11y improvements

pettinarip:
## Description

Turn the Option component into a button.


minimalsm:
@all-contributors please add @pettinarip for a11y

PR #5928, opened 2022-04-06
Fix hidden anchor links on focus

pettinarip:
## Description

Problem: when a heading is focused, the anchor links remain hidden.

This PR addresses fixing how we hide/show them + a small refactor to avoid repeating the styles.


minimalsm (approved):
LGTM :-)

PR #5826, opened 2022-03-29
Add Ethereum Vancouver to list of meetups

adrianmcli:
## Description

This PR adds the [Ethereum Vancouver](https://www.meetup.com/ethvancouver-meetup-group/) meetup group to the list of meetups. Here are some social links:

- http://ethvancouver.org/ 
- http://discord.ethvancouver.org/
- http://meetup.ethvancouver.org/

Note that this is a different group than the previous "Vancouver Ethereum Meetup". That group has since rebranded to the Vancouver Blockchain Meetup. In contrast, this new meetup aims to be Ethereum and EVM-focused and will focus on more technical matters as well.

I would recommend removing the other Vancouver meetup from the list since they are no longer Ethereum-focused and it is a meetup group run by a staunch BSV supporter. I have nothing against that meetup, but it may cause confusion for those looking for a local meetup. They also haven't had an event for almost a year. This removal can probably be a separate PR if the reviewers agree it should be removed.

PR #5931, opened 2022-04-07
Remove legacy meetup (they are now BSV maxis)

adrianmcli:
## Description

As I mentioned briefly in #5826, the old Vancouver Ethereum meetup group is now a BSV maxi group.

If you click the link: https://www.meetup.com/Vancouver-Ethereum-Meetup

You will be greeted by a "Bitcoin is BSV" meetup:

image

My understanding is that the Ethereum Foundation would not want to confuse users and demonstrate supporting a BSV maxi group that has nothing to do with Ethereum on their website. Therefore, I propose removing this meetup from the list of Ethereum meetups.

### DIFF
--- a/src/data/community-meetups.json
+++ b/src/data/community-meetups.json
@@ -66,12 +66,6 @@
     "link": "https://www.meetup.com/ethvancouver-meetup-group/"
   },
   {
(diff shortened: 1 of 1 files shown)


minimalsm (approved):
Oh, dear. Nice catch, thank you @adrianmcli

PR #5935, opened 2022-04-07
migrate Ethereum history page from ethdocs.org

jmcook1186:
## Description

Ethereum history from inception to first production network (homestead) is added to the history page on ethereum.org. The text is migrated from ethdocs.org with only minor typographic changes. Removed/updated dead links.

Also embedded image of Vitalik's announcement of Ethereum genesis on bitcointalk.org.

UPDATE: Added "design rationale" page to this PR. This is a direct, unedited transfer of information from wiki -> ethereum.org. I feel the page needs preserving but does not fit well on ethereum.org right now. I've migrated it into `/history/design-rationale`.

## Related Issue

#5731


minimalsm:
Hey @jmcook1186,

I support adding these details, but I think we have a separation of concerns problem here. I believe the best approach is to separate these into two pages:

1. the history of Ethereum (i.e. a story-driven narrative)
2. the network upgrades history (i.e. the current history page)

There will obviously be some crossover with this approach, but I think this is best when we consider the discussion we had around fleshing out this story-driven narrative and translating it to provide a free, approachable resource for all to learn Ethereum's story so far.

What do you think?

src/content/history/design-rationale/index.md
```diff
@@ -0,0 +1,227 @@
...
+4.  **Distinction between terminating and non-terminating nodes**: technically, the "is this node terminating" flag is unnecessary, as all tries in Ethereum are used to store static key lengths, but we added it anyway to increase generality, hoping that the Ethereum MPT implementations will be used as-is by other cryptographic protocols.
+5.  **Using sha3(k) as the key in the "secure tree"** (used in the state and account storage tries): this makes it much more difficult to DoS the trie by setting up maximally unfavorable chains of diverge nodes 64 levels deep and repeatedly calling SLOAD and SSTORE on them. Note that this makes it more difficult to enumerate the tree; if you want to have enumeration capability in your client, the simplest approach is to maintain a database mapping `sha3(k) -> k`.
+
+## RLP
+
+RLP ("recursive length prefix") encoding is the main serialization format used in Ethereum, and is used everywhere - for blocks, transactions, account state data and wire protocol messages. RLP is formally described [here](./rlp.md).
```
wackerow, line 79: Similar to above, we may want to pull this until we have the `data-structures` PR merged (or at least make sure we're set on the path name first, then aim to merge these together)
```suggestion
RLP ("recursive length prefix") encoding is the main serialization format used in Ethereum, and is used everywhere - for blocks, transactions, account state data and wire protocol messages.
```

src/content/history/design-rationale/index.md
```diff
@@ -0,0 +1,227 @@
...
+    [ medstate, gas_used, logbloom, logs ]
+
+Where:
+
+- `medstate` is the state trie root after processing the transaction
+- `gas_used` is the amount of gas used after processing the transaction
+- `logs` is a list of items of the form `[address, [topic1, topic2...], data]` that are produced by the `LOG0` ... `LOG4` opcodes during the execution of the transaction (including by the main call and sub-calls). `address` is the address of the contract that produced the log, the topics are up to 4 32-byte values, and the data is an arbitrarily sized byte array.
+- `logbloom` is a bloom filter made up of the addresses and topics of all logs in the transaction.
+
+There is also a bloom in the block header, which is the OR of all of the blooms for the transactions in the block. The purpose of this construction is to make the Ethereum protocol light-client friendly in as many ways as possible. For more details on Ethereum light clients and their use cases, see the [light client page (principles section)](../concepts/light-client-protocol.md#principles).
```
wackerow, line 115: Broken link from migration... link back to ethdocs.org? Remove link? Not sure we have a great alternative to link to within the repo on light clients, so may err on the side of just removing this. 
```suggestion
There is also a bloom in the block header, which is the OR of all of the blooms for the transactions in the block. The purpose of this construction is to make the Ethereum protocol light-client friendly in as many ways as possible.
```

src/content/history/design-rationale/index.md
```diff
@@ -0,0 +1,227 @@
...
+
+- A simulator that measures centralization risks is available at <https://github.com/ethereum/economic-modeling/blob/master/ghost.py>
+- A high-level discussion can be found at <https://blog.ethereum.org/2014/07/11/toward-a-12-second-block-time/>
+
+Design decisions in our block time algorithm include:
+
+- **12 second block time**: 12 seconds was chosen as a time that is as fast as possible, but is at the same time substantially longer than network latency. A [2013 paper](http://www.tik.ee.ethz.ch/file/49318d3f56c1d525aabf7fda78b23fc0/P2P2013_041.pdf) by Decker and Wattenhofer in Zurich measures Bitcoin network latency, and determines that 12.6 seconds is the time it takes for a new block to propagate to 95% of nodes; however, the paper also points out that the bulk of the propagation time is proportional to block size, and thus in a faster currency we can expect the propagation time to be drastically reduced. The constant portion of the propagation interval is about 2 seconds; however, for safety we assume that blocks take 12 seconds to propagate in our analysis.
```
wackerow, line 132: Dead link
```suggestion
- **12 second block time**: 12 seconds was chosen as a time that is as fast as possible, but is at the same time substantially longer than network latency. A 2013 paper by Decker and Wattenhofer in Zurich measures Bitcoin network latency, and determines that 12.6 seconds is the time it takes for a new block to propagate to 95% of nodes; however, the paper also points out that the bulk of the propagation time is proportional to block size, and thus in a faster currency we can expect the propagation time to be drastically reduced. The constant portion of the propagation interval is about 2 seconds; however, for safety we assume that blocks take 12 seconds to propagate in our analysis.
```

wackerow (changes requested):
@jmcook1186 Loving these PR's of yours, great work! I like the new breakdown where the story is separated a bit from the strict upgrade timeline. 

Need to get an easier link to jump to the `/network-upgrades/` subpage. Otherwise left a handful of small semantic adjustments, and suggestions on removing some dead links from the design rationale page.
After that I think this is good to go =)

src/content/history/index.md
```diff
@@ -1,14 +1,14 @@
 ---
-title: History and Forks of Ethereum
-description: A history of the Ethereum blockchain including major milestones, releases, and forks.
+title: Upgrade Timeline
```
minimalsm, line 2:
```suggestion
title: Ethereum's upgrade timeline
```

src/content/history/index.md
```diff
@@ -1,14 +1,14 @@
...
+description: An Ethereum network upgrade timeline
 lang: en
 sidebar: true
 sidebarDepth: 1
 ---
 
-# The history of Ethereum {#the-history-of-ethereum}
+# Upgrade Timeline
```
minimalsm, line 9:
```suggestion
# Ethereum's upgrade timeline {#upgrade-timeline}
```

minimalsm:
@jmcook1186 I've moved the network upgrade's timeline back to (/history/) and moved the new page to (/history/story-of-ethereum/) for now. We link out to /history/ a lot across the site which would break those links and we'd also have to change translated versions of ethereum.org to match, as well as update our translation management software mid update, which might confuse our reviewers.

Support the change of structure overall, but let's get the content in first and change the architecture when the timing is more optimal.

minimalsm:
Closing for now until we get a better sense of what we want to do here and if we want to do it.

PR #5954, opened 2022-04-08
Catalan (ca): update "Use Cases" content bucket translation files 

corwintines:
## Description
- Update `Use cases` content bucket translation files for Catalan


minimalsm (approved):
LGTM other than horizontal scroll (same as Dutch in #5955). I suspect we introduced this bug in a different PR.

PR #5959, opened 2022-04-08
Japanese (jp): update "Use Cases" content bucket translation files

corwintines:
## Description
- Update `Use cases` content bucket translation files for Japanese


minimalsm (approved):
LGTM

PR #5961, opened 2022-04-08
Turkish (tr): update "Use Cases" content bucket translation files

corwintines:
## Description
- Update `Use cases` content bucket translation files for Turkish


minimalsm (approved):
Horizontal scroll but LGTM

Issue #5904, opened 2022-04-05
List of exchanges too long for some countries


minimalsm:
**Describe the bug**

We added support for more exchanges (#5839). The increased number of exchanges is causing our list of exchanges to be very long for some countries, which seems visually awkward.

**To Reproduce**
Steps to reproduce the behaviour:

1. Go to 'https://ethereumorgwebsitedev01-addexchanges.gtsb.io/en/get-eth/'
3. Scroll down to '**What country do you live in?**'
4. Select '**United Kingdom**'
5. See list of exchanges

**Expected behavior**

- Unsure exact behaviour we want, but the component should more elegantly scale with a high number of exchanges.
- I don't think we want to hide any exchanges

**Screenshots**

<img width="1139" alt="Screenshot 2022-04-05 at 09 31 29" src="https://user-images.githubusercontent.com/62268199/161713241-fc40de58-6c29-499f-b907-4442b5d93baa.png">

**Want to contribute?**

We love contributions from the Ethereum community! This change requires a design proposal before implementation. Please comment here if you're interested in helping out with the design of this issue. If you'd like input from our design community, [join the ethereum.org Discord](https://discord.gg/CetY6Y4) and use our design channel for discussion.

HursitTarcan:
@minimalsm can I try to fix this?

minimalsm:
Hey @HursitTarcan 👋  This change requires a design proposal before implementation. If you'd like to discuss the design changes please use our #design channel [on Discord](https://discord.gg/CetY6Y4).

HursitTarcan:
## Description

The idea to fix this is to take a fixed height and add a scrollbar to the right. We can take a `fixed height of e.g. 5&1/2 items`, this makes the user realise it is scrollable. The `scrollbar on the right` will also let the user know that the list is scrollable (soo it is also an idea to keep both lists the same height to make it more visually appealing). I included a screenshot below (in both themes): 

![image](https://user-images.githubusercontent.com/75273616/162438995-be27ffe1-cfbb-4604-a429-8ea802c16336.png)
![image](https://user-images.githubusercontent.com/75273616/162439396-4b6a469e-c88c-4768-9430-e078705825f8.png)

These changes are made with google dev tools by adding the following css code: 
- `overflow-y: scroll`
- `height: 320px`

![image](https://user-images.githubusercontent.com/75273616/162439996-8e8b2d22-47d6-47cc-a38e-1583c316d689.png)

[...]

PR #5977, opened 2022-04-08
[DO NOT MERGE] Design Proposal [Issue #5904]

HursitTarcan:
## Description
Design proposal for issue [#5904]. 

The idea to fix this is to take a fixed height and add a scrollbar to the right. We can take a `fixed height of e.g. 5&1/2 items`, this makes the user realise it is scrollable. The `scrollbar on the right` will also let the user know that the list is scrollable (soo it is also an idea to keep both lists the same height to make it more visually appealing). I included a screenshot below (in both themes): 

image
image

These changes are made with google dev tools by adding the following css code: 
- `overflow-y: scroll`
- `height: 320px`

image

Note: css `overflow-y: scroll` does not show on safari browser. We will have to use `webkit-scrollbar` for this, this way we will also be able to change the style of the scrollbar. E.g. `give it a smaller width`, `give it more visually appealing colors` and `changing the style of the thumb (rounding corners etc.)`.  See screenshot below for an example scrollbar design: 

image

[NOTE] This is only a desktop issue, no mobile changes have to be done. 

## Related Issue
PR related to [Issue #5904]


minimalsm:
Hey @HursitTarcan, thanks for the detailed proposal here. I'm not a designer 😬 but overall I can see value in your approach.

As there are no changes here I think we should post this directly on the issue itself to keep the conversation focused over there. Could you do that and close this out? Thanks ❤️

PR #5978, opened 2022-04-08
sync glossaries between eth.org, ethdocs.org & eth.wiki

jmcook1186:
## Description

This PR collects together entries in the ethdocs.org and eth.wiki glossaries and merges them into the glossary on ethereum.org. There are quite a few new additions. There were also many entries in the glossaries of both ethdocs.org and eth.wiki that I decided NOT to merge into ethereum.org. These are listed below so the PR reviewers can challenge any that they think are actually useful to include (especially those from ethdocs relating to trust and identity - I excluded them as superfluous but think they might actually be quite nice to include).

### Entries from ethdocs NOT merged into ethereum.org in this PR:

```
**deposit**
Digital property placed into a contract involving another party such that if certain conditions are not satisfied that property is automatically forfeited and either credited to a counterparty as insurance against the conditions, or destroyed (= burnt = equally distributed) or donated to some charitable funds. 

**web of trust** 
The idea that if A highly rates B, and B highly rates C, then A is likely to trust C. Complicated and powerful mechanisms for determining the reliability of specific individuals in specific concepts can theoretically be gathered from this principle.

**incentive compatibility**
A protocol is incentive-compatible if everyone is better off “following the rules” than attempting to cheat, at least unless a very large number of people agree to cheat together at the same time (collusion).

**identity** 
A set of cryptographically verifiable interactions that have the property that they were all created by the same person. 

**digital identity** 
The set of cryptographically verifiable transactions signed by the same public key define the digital identity’s behavior. In many real world scenarios (voting) it is desireable that digital identities coincide with real world identities. Ensuring this without violence is an unsolved problem. 

**unique identity** 
A set of cryptographically verifiable interactions that have the property that they were all created by the same person, with the added constraint that one person cannot have multiple unique identities.

**gas price oracle**
A helper function of the Geth client that tries to find an appropriate default gas price when sending transactions.

**serpent** 
Serpent is a high-level language whose syntax is similar to that of Python and it is designed to compile to code for the Ethereum Virtual Machine.

**virtual machine **
In computing, it refers to an emulation of a particular computer system

**NAT** 
Network address translation (NAT) is a methodology of remapping one IP address space into another by modifying network address information in Internet Protocol (IP) datagram packet headers while they are in transit across a traffic routing device.

**homestead** 
Homestead is the second major version release of the Ethereum platform. Homestead includes several protocol changes and a networking change that makes possible further network upgrades: EIP-2 Main homestead hardfork changes; EIP-7 Hardfork EVM update (DELEGATECALL); EIP-8 devp2p forward compatibility. Homestead will launch when block 1,150,000 is reached on the Mainnet. On the Testnet, Homestead will launch at block 494,000. 

**metropolis** 
The third stage of Ethereum’s release. This is the stage when the user interfaces come out (e.g. Mist), including a dapp store, and non-technical users should feel comfortable joining at this point.

**serenity** 
The fourth stage of Ethereum’s release. This is when things are going to get fancy: the network is going to change its mining process from Proof-of-Work to Proof-of-Stake.

**frontier** 
Ethereum was planned to be released in four major steps with Frontier being the name for the first phase. The Frontier release went live on July 30th, 2015. The command line Frontier phase was mainly meant to get mining operations going with the full reward of 5 ether per block and also to promote the emergence of ether exchanges. Frontier surpassed earlier modest expectations and has nurtured tremendous growth of the ecosystem. 

**olympic** 
The Frontier pre-release, which launched on May 9th 2015. It was meant for developers to help test the limits of the Ethereum blockchain. 

**morden** 
Morden is the first Ethereum alternative testnet. It is expected to continue throughout the Frontier and Homestead era.

**consortium chain**
A blockchain where the consensus process is controlled by a pre-selected set of nodes.

**micropayment** 
A micropayment is a financial transaction involving a very small sum of money (<1 USD) and usually one that occurs online.

**crypto-fuel **
Similar to ‘gas’, referring to the amount of cryptocurrency required to power a transaction.

**SPV client** 
A client that downloads only a small part of the blockchain, allowing users of low-power or low-storage hardware like smartphones and laptops to maintain almost the same guarantee of security by sometimes selectively downloading small parts of the state without needing to spend megabytes of bandwidth and gigabytes of storage on full blockchain validation and maintenance. See light client.

**uncle rate** 
The number of uncles produced per block. 

**static node** 
A feature supported by Geth, the Golang Ethereum client, which makes it possible to always connect to specific peers. Static nodes are re-connected on disconnects. For details, see the section on static nodes.

**compiler** 
A program that translates pieces of code written in high level languages into low level executable code.

**network id** 
A number which identifies a particular version of the Ethereum network.

**pending transaction** 
A transaction that is not yet confirmed by the Ethereum network.

**pegging** 
Locking down the exchange rate of the coins/tokens in two chains (usually a main and a side chain) in a certain direction. 

**2-way pegging** 
Locking down the exchange rate of the coins/tokens in two chains (usually a main and a side chain) in both directions.

**checksum** 
A count of the number of bits in a transmission that is included with the unit so that the receiving end can verify that the entirety of the message has been transmitted.

**port** 
A network port is a communication endpoint used by a one of the existing standards of establishing a network conversation (e.g. TCP, UDP).

**IPC** 
Interprocess communication (IPC) is a set of programming interfaces that allow a programmer to coordinate activities among different program processes that can run concurrently in an operating system.

**attach** 
The command used to initiate the Ethereum Javascript console. 

**daemon** 
A computer program that runs as a background process instead of in direct control by an interactive user.
 
**system service** 
See base layer service 

**base layer service** 
Services such as SWARM and Whisper which are built into the Ethereum platform. 

**js** 
Javascript.

**fast sync** 
Instead of processing the entire block-chain one link at a time, and replay all transactions that ever happened in history, fast syncing downloads the transaction receipts along the blocks, and pulls an entire recent state database.

**ICAP format** 
The format of the IBANs defined using the Inter-exchange Client Address Protocol.

**geth** 
Ethereum client implemented in the Golang programming language, based on the protocol as defined in the Ethereum Yellow Paper. 

**eth** 
Ethereum client implemented in the C++ programming language, based on the protocol as defined in the Ethereum Yellow Paper. 

**ethereumjs** 
Ethereum client implemented in the Javascript/Node programming language, based on the protocol as defined in the Ethereum Yellow Paper. 

**pyethereum** 
Ethereum client implemented in the Python programming language, based on the protocol as defined in the Ethereum Yellow Paper. 

**ethereumj **
Ethereum client implemented in the Java programming language, based on the protocol as defined in the Ethereum Yellow Paper. 

**ethereumh **
Ethereum client implemented in the Haskell programming language, based on the protocol as defined in the Ethereum Yellow Paper. 

**parity** 
Ethereum client implemented in the Rust programming language, based on the protocol as defined in the Ethereum Yellow Paper.

**HLL (obsolete) **
Acronym for Higher Level Language, which is what Serpent and Solidity are. HLL is what early Ðapp developers called Ethereum programming languages that did not touch the low level elements. This phrase has been phased out. 

**CLL (obsolete) **
Acronym for C Like Language, which Mutan was. This acronym has been phased out. 

**ES1, ES2, and ES3 (obsolete) **
“Ethereum Script” versions 1,2 and 3. There were early versions of what would become the Ethereum Virtual Machine (EVM). 

**log event **
Contracts are triggered by transactions executed as part of the block verification. If conceived of as a function call, contract execution is asynchronous, and therefore they have no return value. Instead contracts communicate to the outside world with log events. The log events are part of the transaction receipt which is produced when the transaction is executed. The receipts are stored in the receipt trie, the integrity of which is guaranteed by the fact that the current root of the receipt trie is part of the block header alongside the roots of state and state-trie. In a broad sense from the external perspective receipts are part of the Ethereum system state except that they are not readable contracts internally.
```

### Entries from eth.wiki NOT merged into ethereum.org in this PR:
```
**Stale**: a stale is a block that is created when there is already another block with the same parent out there; stales typically get discarded and are wasted effort.

**Uncle inclusion mechanism**: Ethereum has a mechanism where a block may include its uncles; this ensures that miners that create blocks that do not quite get included into the main chain can still get rewarded.

**Slashing condition**: a condition which, if triggered by a validator, causes the validator’s deposit to be destroyed. 

**Prepare and commit**: two types of messages that validators can send in many types of consensus protocols; see https://medium.com/@VitalikButerin/minimal-slashing-conditions-20f0b500fc6c 

**Fault**: an action taken by a validator (or more generally, a participant in a mechanism) that they would not have taken had they correctly followed the protocol 

**Liveness fault**: a validator failing to submit a message that according to the protocol they should have submitted (or submitting a message later than they should have) 

**Censorship fault**: a validator failing to accept valid messages from other validators
Invalidity fault: a validator sending a message that a computer running the correct algorithm could not possibly send, unless its internal state is manipulated with in some way other than rewinding. 
Uniquely attributable fault: a fault such that there exists clear evidence which can be used to determine exactly which validator committed the fault. For example, liveness faults are not uniquely attributable because if a message from A fails to reach B, it could be because A failed to send that message, or because B failed to listen to it, whereas equivocation faults are uniquely attributable.

**Data availability problem and Fisherman’s dilemma**: see https://github.com/ethereum/research/wiki/A-note-on-data-availability-and-erasure-coding 

**Validity**: the property of a state that it is indeed the result of executing a valid history of transactions

**Tight coupling**: chains A and B are tightly coupled if (i) any state of A points to some state of B (and vice versa), and (ii) a state of A should not be considered admissible unless both that state itself and the state of B that it points to are valid and data-available. 

**Loose coupling**: chains A and B are loosely coupled if (i) any state of A points to some state of B (and vice versa), and (ii) they are not tightly coupled.

**Whisper**: an upcoming P2P messaging protocol 

**Swarm**: an upcoming P2P data storage protocol optimized for static web hosting

**Self-executing**: Functioning by itself, not controlled by any other party other than itself. Self-executing smart contracts would cut costs/overhead by removing the need for an arbitrator and trust towards a third party.
```

## Related Issue

#5731
#5690

### DIFF
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -103,6 +103,10 @@ The data in a block which is unique to its content and the circumstances in whic
 
 The process of transmitting a confirmed block to all other nodes in the network.
 
+### block reward {#block-reward}
+
+The amount of ether rewarded to the producer of a new valid block.
+
 ### block time {#block-time}
 
 The average time interval between blocks being added to the blockchain.
@@ -551,7 +555,7 @@ A [transaction](#transaction) sent from a [contract account](#contract-account) 
 
 ### issuance
 
-The minting and granting of new cryptocurrency to a miner who has found a new block.
+The minting of new ether to reward block proposal, attestation and whistle-blowing.
 
 ## K {#section-k}
 
@@ -563,7 +567,7 @@ Also known as a "password stretching algorithm," it is used by [keystore](#keyst
   Smart contract security
 </DocLink>
 
-### keyfile {#keyfile}
+### keystore {#keyfile}
 
 Every account’s private key/address pair exists as a single keyfile in an Ethereum client. These are JSON text files which contains the encrypted private key of the account, which can only be decrypted with the password entered during account creation.
 
@@ -571,9 +575,6 @@ Every account’s private key/address pair exists as a single keyfile in an Ethe
 
 Cryptographic [hash](#hash) function used in Ethereum. Keccak-256 was standardized as [SHA](#sha)-3.
 
-### keystore file {#keystore-file}
-
-A JSON-encoded file that contains a single (randomly generated) [private key](#private-key), encrypted by a passphrase for extra security.
 
 <Divider />
 
@@ -642,10 +643,6 @@ The process of verifying transactions and contract execution on the Ethereum blo
 ### mining pool {#mining-pool}
 
 The pooling of resources by miners who share their processing power and split [block rewards](#block-reward).
-
-### block reward {#block-reward}
-
-The amount of ether rewarded to the producer of a new valid block.
 
 ### miner {#miner}
 
@@ -852,7 +849,7 @@ A scaling solution that uses a separate chain with different, often faster, [con
 
 ### signing {#signing}
 
-Demonstrating that a transaction originated from a specific person by encoding it using a private key.
+Demonstrating cryptographically that a transaction was approved by the holder of a specific private key.
 
 ### singleton {#singleton}
 
@@ -968,7 +965,7 @@ Short for "test network," a network used to simulate the behavior of the main Et
 
 ### token {#token}
 
-A fungible virtual good that can be traded.
+A tradable virtual good defined in smart contracts on the Ethereum blockchain.
 
 ### token standard {#token-standard}
 
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -135,7 +135,7 @@ The first of two [hard forks](#hard-fork) for the [Metropolis](#metropolis) deve
 
 ## C {#section-c}
 
-### Casper-FFG
+### Casper-FFG {casper-ffg}
 
 Casper-FFG is a proof-of-stake consensus protocol used in conjunction with the [LMD-GHOST](#lmd-ghost) fork choice algorithm to allow [consensus clients](#consensus-client) to agree on the head of the Beacon Chain.
 
@@ -287,6 +287,14 @@ A deliberate blockchain fork, where a user with a sufficiently large amount of m
 
 A cryptographic algorithm used by Ethereum to ensure that funds can only be spent by their owners. It's the preferred method for creating public and private keys. Relevant for account [address](#address) generation and [transaction](#transaction) verification.
 
+### encryption {#encryption}
+
+Encryption is the conversion of electronic data into a form unreadable by anyone except the owner of the correct decryption key.
+
+### entropy {#entropy}
+
+In the context of cryptography, lack of predictability or level of randomness. When generating secret information, such as [private keys](#private-key), algorithms usually rely on a source of high entropy to ensure the output is unpredictable.
+
 ### epoch {#epoch}
 
 A period of 32 [slots](#slot) (6.4 minutes) in the [Beacon Chain](#beacon-chain)-coordinated system. [Validator](#validator) [committees](#committee) are shuffled every epoch for security reasons. There's an opportunity at each epoch for the chain to be [finalized](#finality). The term is also used on the [execution layer](#execution-layer) to mean the interval between each regeneration of the database used as a seed by the PoW algorithm [Ethash](#Ethash). The epoch in specified as 30000 blocks.
@@ -336,14 +344,6 @@ A design document providing information to the Ethereum community, describing a 
 The ENS registry is a single central [contract](#smart-contract) that provides a mapping from domain names to owners and resolvers, as described in [EIP](#eip) 137.
 
 [Read more at ens.domains](https://ens.domains)
-
-### encryption
-
-Encryption is the conversion of electronic data into a form unreadable by anyone except the owner of the correct decryption key.
-
-### entropy {#entropy}
-
-In the context of cryptography, lack of predictability or level of randomness. When generating secret information, such as [private keys](#private-key), algorithms usually rely on a source of high entropy to ensure the output is unpredictable.
 
 ### execution client {#execution-client}
 
@@ -622,7 +622,7 @@ The [fork-choice algorithm](#fork-choice-algorithm) used by Ethereum's consensus
 
 Short for "main network," this is the main public Ethereum [blockchain](#blockchain). Real ETH, real value, and real consequences. Also known as layer 1 when discussing [layer 2](#layer-2) scaling solutions. (Also, see [testnet](#testnet))
 
-### memory-hard
+### memory-hard {#memory-hard}
 
 Memory hard functions are processes that experience a drastic decrease in speed or feasibility when the amount of available memory even slightly decreases. An example id the Ethereum mining algorithm [Ethash](#ethash).
 
@@ -761,7 +761,7 @@ Sale of cryptocurrency before the actual launch of the network.
 
 A secret number that allows Ethereum users to prove ownership of an account or contracts, by producing a digital signature (see [public key](#public-key), [address](#address), [ECDSA](#ecdsa)).
 
-### private chain
+### private chain {#private-chain}
 
 A fully private blockchain is one with write permissions limited to one organization.
 
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -43,6 +43,10 @@ both from outside the blockchain and for contract-to-contract interactions.
 
 An Application Programming Interface (API) is a set of definitions for how to use a piece of software. An API sits between an application and a web server, and facilitates the transfer of data between them.
 
+### ASIC {#asic}
+
+Application-specific integrated circuit, in this case referring to an integrated circuit custom built for cryptocurrency mining.
+
 ### assert {#assert}
 
 In [Solidity](#solidity), `assert(false)` compiles to `0xfe`, an invalid opcode, which uses up all remaining [gas](#gas) and reverts all changes. When an `assert()` statement fails, something very wrong and unexpected is happening, and you will need to fix your code. You should use `assert()` to avoid conditions that should never, ever occur.
@@ -87,6 +91,26 @@ A collection of required information (a block header) about the comprised [trans
   Blocks
 </DocLink>
 
+### block(chain) explorer
+
+A website that allows easy searching and extraction of data from the blockchain.
+
+### block header {#block-header}
+
+The data in a block which is unique to its content and the circumstances in which it was created. It includes the hash of the previous block’s header, the version of the software the block is mined with, the timestamp and the merkle root hash of the contents of the block.
+
+### block propagation {#block-propagation}
+
+The process of transmitting a confirmed block to all other nodes in the network.
+
+### block time
+
+The average time interval between the mining of two blocks. After the merge the block time will be the time between two Beacon blocks (under optimal conditions there will be one block per slot, which occurs every 6 seconds).
+
+### block validation {#block-validation}
+
+Checking of the coherence of the cryptographic signature of the block with the history stored in the entire blockchain.
+
 ### blockchain {#blockchain}
 
 In Ethereum, a sequence of [blocks](#block) validated by the [proof-of-work](#pow) system, each linking to its predecessor all the way to the [genesis block](#genesis-block). There is no block size limit; it instead uses varying [gas limits](#gas-limit).
@@ -94,6 +118,10 @@ In Ethereum, a sequence of [blocks](#block) validated by the [proof-of-work](#po
 <DocLink to="/developers/docs/intro-to-ethereum#what-is-a-blockchain">
   What is a Blockchain?
 </DocLink>
+
+### bootnode
+
+The nodes which can be used to initiate the discovery process when running a node. The endpoints of these nodes are recorded in the Ethereum source code.
 
 ### bytecode {#bytecode}
 
@@ -106,6 +134,10 @@ The first of two [hard forks](#hard-fork) for the [Metropolis](#metropolis) deve
 <Divider />
 
 ## C {#section-c}
+
+### Casper-FFG
+
+Casper-FFG is a proof-of-stake consensus protocol used in conjunction with the [LMD-GHOST](#lmd-ghost) fork choice algorithm to allow [consensus clients](#consensus-client) to agree on the head of the Beacon Chain.
 
 ### checkpoint {#checkpoint}
 
@@ -122,6 +154,10 @@ Converting code written in a high-level programming language (e.g., [Solidity](#
 ### committee {#committee}
 
 A group of at least 128 [validators](#validator) assigned to beacon and shard blocks at random by [the Beacon Chain](#beacon-chain).
+
+### computational infeasibility {#computational-infeasibility}
+
+a process is computationally infeasible if it would take an impracticably long time (eg. billions of years) to do it for anyone who might conceivably have an interest in carrying it out.
 
 ### consensus {#consensus}
 
@@ -161,15 +197,19 @@ A crosslink provides a summary of a shard's state. It's how [shard](#shard) chai
 
 <Divider />
 
+### cryptoeconomics {#cryptoeconomics}
+
+The economics of cryptocurrencies.
+
 ## D {#section-d}
 
-### Decentralized Autonomous Organization (DAO) {#dao}
+### Đ {#D-with-stroke}
 
-A company or other organization that operates without hierarchical management. DAO may also refer to a contract named "The DAO" launched on April 30, 2016, which was then hacked in June 2016; this ultimately motivated a [hard fork](#hard-fork) (codenamed DAO) at block 1,192,000, which reversed the hacked DAO contract and caused Ethereum and Ethereum Classic to split into two competing systems.
+Đ (D with stroke) is used in Old English, Middle English, Icelandic, and Faroese to stand for an uppercase letter “Eth”. It is used in words like ĐEV or Đapp (decentralized application), where the Đ is the Norse letter “eth”. The uppercase eth (Ð) is also used to symbolize the cryptocurrency Dogecoin. This is commonly seen in older Ethereum literature but is used less often today.
 
-<DocLink to="/dao/">
-  Decentralized Autonomous Organizations (DAOs)
-</DocLink>
+### DAG {#DAG}
+
+DAG stands for Directed Acyclic Graph. It is a data structure composed of nodes and links between them. Ethereum uses a DAG in its [proof of work](#proof-of-work) algorithm, [Ethash](#ethash).
 
 ### Dapp {#dapp}
 
@@ -177,6 +217,22 @@ Decentralized application. At a minimum, it is a [smart contract](#smart-contrac
 
 <DocLink to="/developers/docs/dapps/">
   Introduction to Dapps
+</DocLink>
+
+### Data availability {#data-availability}
+
+The property of a state that any node connected to the network could download any specific part of the state that they wish to
+
+### decentralization {#decentralization}
+
+The concept of moving the control and execution of processes away from a central entity.
+
+### Decentralized Autonomous Organization (DAO) {#dao}
+
+A company or other organization that operates without hierarchical management. DAO may also refer to a contract named "The DAO" launched on April 30, 2016, which was then hacked in June 2016; this ultimately motivated a [hard fork](#hard-fork) (codenamed DAO) at block 1,192,000, which reversed the hacked DAO contract and caused Ethereum and Ethereum Classic to split into two competing systems.
+
+<DocLink to="/dao/">
+  Decentralized Autonomous Organizations (DAOs)
 </DocLink>
 
 ### decentralized exchange (DEX) {#dex}
@@ -213,6 +269,18 @@ A short string of data a user produces for a document using a [private key](#pri
 
 <Divider />
 
+### discovery {#discovery}
+
+The process by which an Ethereum node finds other nodes (computers running client software) to connect to.
+
+### distributed hash table {#distributed-hash-table}
+
+A distributed hash table (DHT) is a data structure containing `(key, value)` pairs used by Ethereum nodes to identify peers to connect to and determine which protocols to use to communicate.
+
+### double spend {#double-spend}
+
+A deliberate blockchain fork, where a user with a sufficiently large amount of mining power/stake sends a transaction moving some currency off-chain (e.g. exiting into fiat money or making an off-chain purchase) then reorganising the blockchain to remove that transaction. A successful double spend leaves the attacker with both their on and off-chain assets.
+
 ## E {#section-e}
 
 ### elliptic curve digital signature algorithm (ECDSA) {#ecdsa}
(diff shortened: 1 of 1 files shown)


src/content/glossary/index.md
```diff
@@ -43,6 +43,10 @@ both from outside the blockchain and for contract-to-contract interactions.
 
 An Application Programming Interface (API) is a set of definitions for how to use a piece of software. An API sits between an application and a web server, and facilitates the transfer of data between them.
 
+### ASIC {#asic}
+
+Application-specific integrated circuit, in this case referring to an integrated circuit custom built for cryptocurrency mining.
```
minimalsm, line 48:
```suggestion
Application-specific integrated circuit. This usually refers to an integrated circuit, custom-built for cryptocurrency mining.
```

src/content/glossary/index.md
```diff
@@ -87,6 +91,26 @@ A collection of required information (a block header) about the comprised [trans
   Blocks
 </DocLink>
 
+### block(chain) explorer
```
minimalsm, line 94:
```suggestion
### block explorer {#block-explorer}
```

src/content/glossary/index.md
```diff
@@ -87,6 +91,26 @@ A collection of required information (a block header) about the comprised [trans
...
+
+A website that allows easy searching and extraction of data from the blockchain.
```
minimalsm, line 96: This seems vague.

src/content/glossary/index.md
```diff
@@ -87,6 +91,26 @@ A collection of required information (a block header) about the comprised [trans
...
+
+### block header {#block-header}
+
+The data in a block which is unique to its content and the circumstances in which it was created. It includes the hash of the previous block’s header, the version of the software the block is mined with, the timestamp and the merkle root hash of the contents of the block.
+
+### block propagation {#block-propagation}
+
+The process of transmitting a confirmed block to all other nodes in the network.
+
+### block time
```
minimalsm, line 106:
```suggestion
### block time {#block-time}
```

src/content/glossary/index.md
```diff
@@ -87,6 +91,26 @@ A collection of required information (a block header) about the comprised [trans
...
+
+The average time interval between the mining of two blocks. After the merge the block time will be the time between two Beacon blocks (under optimal conditions there will be one block per slot, which occurs every 6 seconds).
```
minimalsm, line 108: > After the merge the block time will be the time between two Beacon blocks (under optimal conditions there will be one block per slot, which occurs every 6 seconds).

I don't think we need to explain this

src/content/glossary/index.md
```diff
@@ -95,6 +119,10 @@ In Ethereum, a sequence of [blocks](#block) validated by the [proof-of-work](#po
   What is a Blockchain?
 </DocLink>
 
+### bootnode
```
minimalsm, line 122:
```suggestion
### bootnode {#bootnode}
```

src/content/glossary/index.md
```diff
@@ -213,20 +269,48 @@ A short string of data a user produces for a document using a [private key](#pri
 
 <Divider />
 
+### discovery {#discovery}
+
+The process by which an Ethereum node finds other nodes (computers running client software) to connect to.
```
minimalsm, line 274: Don't need to explain what a node is here

```suggestion
The process by which an Ethereum node finds other nodes to connect to.
```

src/content/glossary/index.md
```diff
@@ -856,15 +1028,15 @@ A [node](#node) in a [proof-of-stake](#pos) system responsible for storing data,
   Staking in Ethereum
 </DocLink>
 
-### Validity proof {#validity-proof}
+### validity proof {#validity-proof}
```
minimalsm, line 1031: ty

src/content/glossary/index.md
```diff
@@ -837,6 +1005,10 @@ At a technical level, your transaction fee relates to how much [gas](#gas) your
 
 Reducing transaction fees is a subject of intense interest right now. See [Layer 2](#layer-2)
 
+### trustlessness {#trustlessness}
+
+Refers to the ability of a network to trustworthily mediate transactions without any of the involved parties needing to trust anyone else.
```
minimalsm, line 1010:
```suggestion
The ability of a network to mediate transactions without any of the involved parties needing to trust a third party
```

src/content/glossary/index.md
```diff
@@ -813,6 +977,10 @@ Short for "test network," a network used to simulate the behavior of the main Et
   Testnets
 </DocLink>
 
+### token {#token}
+
+A fungible virtual good that can be traded. More formally, a token is a database mapping addresses to numbers with the property that the primary allowed operation is a transfer of N tokens from A to B, with the conditions that N is non-negative, N is not greater than A’s current balance, and a document authorizing the transfer is digitally signed by A.
```
minimalsm, line 982: > More formally, a token is a database mapping addresses to numbers with the property that the primary allowed operation is a transfer of N tokens from A to B, with the conditions that N is non-negative, N is not greater than A’s current balance, and a document authorizing the transfer is digitally signed by A.

This is too in the weeds for the glossary

src/content/glossary/index.md
```diff
@@ -709,6 +861,10 @@ A scaling solution that uses a separate chain with different, often faster, [con
   Sidechains
 </DocLink>
 
+### signing {#signing}
+
+Demonstrating that some data originated from a certain person by encoding it using a private key.
```
minimalsm, line 866: demonstrating that a transaction originated from... ?

minimalsm, line 866:
```suggestion
Demonstrating that a transaction originated from a specific person by encoding it using a private key.
```

src/content/glossary/index.md
```diff
@@ -679,8 +823,16 @@ A type of [layer 2](#layer-2) scaling solution that batches multiple transaction
 
 <Divider />
 
+### RPC {#rpc}
+
+Remote Procedure Call, a protocol that a program uses to request a service from a program located in another computer in a network without having to understand the network details.
```
minimalsm, line 828:
```suggestion
**Remote procedure call (RPC)** is a protocol that a program uses to request a service from a program located on another computer in a network without having to understand the network details.
```

src/content/glossary/index.md
```diff
@@ -621,10 +753,18 @@ An off-chain scaling solution that uses [fraud proofs](#fraud-proof), like [Opti
...
+
+Sale of cryptocurrency before the actual launch of the network.
+
 ### private key (secret key) {#private-key}
 
 A secret number that allows Ethereum users to prove ownership of an account or contracts, by producing a digital signature (see [public key](#public-key), [address](#address), [ECDSA](#ecdsa)).
 
+### private chain {#private-chain}
+
+A fully private blockchain is one with write permissions limited to one organization.
```
minimalsm, line 766: Is this accurate? I find it useful to think of private chains as what they are not (open, decentralized) rather than what they are (closed, centralized). How closed they are seems arbitrary to me.

src/content/glossary/index.md
```diff
@@ -382,6 +466,10 @@ A virtual fuel used in Ethereum to execute smart contracts. The [EVM](#evm) uses
 
 The maximum amount of [gas](#gas) a [transaction](#transaction) or [block](#block) may consume.
 
+### gas price {#gas-price}
+
+Price in ether of one unit of gas specified in a transaction. With the launch of Homestead, the default gas price reduces from 50 shannon to 20 shannon (~60% reduction).
```
minimalsm, line 471: Let's not talk about Homestead here :P

```suggestion
Price in ether of one unit of gas specified in a transaction.
```

src/content/glossary/index.md
```diff
@@ -418,6 +510,10 @@ A [wallet](#wallet) using the hierarchical deterministic (HD) key creation and t
 
 A value used to generate the master [private key](#private-key) and master chain code for an HD [wallet](#wallet). The wallet seed can be represented by mnemonic words, making it easier for humans to copy, back up, and restore private keys.
 
+### hexadecimal {#hexadecimal}
+
+Common representation format for byte sequencing. Its advantage is that values are represented in a compact format using two characters per byte (the characters [0-9][a-f]).
```
minimalsm, line 515: Not convinced we should add this.

Our listing criteria:
- [ ] This is a term not already found in the glossary (for similar terms, please consider the benefits of a new term vs updating an existing term)
- [ ] This term/definition is not a product advertisement or contain other promotional content
- [ ] **This term/definition is directly relevant to Ethereum**

We should be careful not to try to cram every computer science term into the glossary.

src/content/glossary/index.md
```diff
@@ -789,6 +949,10 @@ A [layer 2](#layer-2) solution where a channel is set up between participants, w
 
 Supermajority is the term given for an amount exceeding 2/3 (66%) of the total staked ether on the [Beacon Chain](#beacon-chain). A supermajority vote is required for blocks to be [finalized](#finality) on the Beacon Chain.
 
+### syncing {#syncing}
+
+The process of downloading the entire blockchain.
```
minimalsm, line 954: Latest version?

```suggestion
The process of downloading the entire latest version of a blockchain to a node.
```

minimalsm (commented):
Overall I think this looks good.

I'd like some other input on these to make sure we think they fit into our current listing criteria:

- [ ] This is a term not already found in the glossary (for similar terms, please consider the benefits of a new term vs updating an existing term)
- [ ] This term/definition is not a product advertisement or contain other promotional content
- [ ] This term/definition is directly relevant to Ethereum

One thing we should be careful of is not try to cram every computer science related term into the glossary because Ethereum makes use of it in some way.

src/content/glossary/index.md
jmcook1186: Changed to:
"A website that allows a user to search for information from, and about, a blockchain. This includes retrieving individual transactions, activity associated with specific addresses and information about the network."

src/content/glossary/index.md
jmcook1186: ok, changed to:

"The average time interval between blocks being added to the blockchain."

src/content/glossary/index.md
jmcook1186: ok! removed.

src/content/glossary/index.md
jmcook1186: idk, I think this definition makes sense. One "organization" can be one person or one hundred people with multiple affiliations - the point is that not everyone can append data to that particular blockchain.

Suggest changing to:

"A permissioned blockchain with restrictions on who can add data." ?

src/content/glossary/index.md
jmcook1186: ok!

src/content/glossary/index.md
wackerow: It's not just writing data necessarily either. If I spin up a chain on my own network with restricted access, the public cannot read it either. I could potentially share it with whoever I want though, so it's also not strictly limited to any one entity or organization. 

Could simplify to:
```suggestion
A fully private blockchain is one with permissioned access, not publicly available for use. 
```

src/content/glossary/index.md
```diff
@@ -460,6 +553,10 @@ A [transaction](#transaction) sent from a [contract account](#contract-account)
 
 <Divider />
 
+### issuance
+
+The minting and granting of new cryptocurrency to a miner who has found a new block.
+
```
wackerow, line 559: This is not fully accurate since the launch of the Beacon Chain... Issuance on the Beacon Chain is not just from block proposals, but also for attestations and whistleblowing. 
https://twitter.com/trent_vanepps/status/1509538632016801792

src/content/glossary/index.md
```diff
@@ -470,6 +567,10 @@ Also known as a "password stretching algorithm," it is used by [keystore](#keyst
   Smart contract security
 </DocLink>
 
+### keyfile {#keyfile}
+
+Every account’s private key/address pair exists as a single keyfile in an Ethereum client. These are JSON text files which contains the encrypted private key of the account, which can only be decrypted with the password entered during account creation.
+
```
wackerow, line 573: Is the term "keyfile" used very often? Feel like "keystore" is much more common. 

Also, this is confusing and makes it sound like all the private/public keys to all accounts exist inside the client software. 

Feel like rewording to not mention Ethereum "clients" here, but perhaps "wallet" or "wallet client" instead?

src/content/glossary/index.md
```diff
@@ -534,6 +639,18 @@ The act of passing a [message](#message) from one account to another. If the des
 
 The third development stage of Ethereum, launched in October 2017.
 
+### mining {#mining}
+
+The process of verifying transactions and contract execution on the Ethereum blockchain in exchange for a reward in ether with the mining of every block.
+
+### mining pool{#mining-pool}
```
wackerow, line 646:
```suggestion
### mining pool {#mining-pool}
```

src/content/glossary/index.md
```diff
@@ -534,6 +639,18 @@ The act of passing a [message](#message) from one account to another. If the des
...
+
+The pooling of resources by miners who share their processing power and split [mining rewards](#mining-reward).
+
+### mining reward {#mining-reward}
+
+The amount of ether given to a miner that mined a new block.
+
```
wackerow, line 653: Think we should think post-merge here... more accurately this could be listed as "block reward"

```suggestion
### block reward {#block-reward}

The amount of ether rewarded to the producer of a new valid block.

```

src/content/glossary/index.md
```diff
@@ -534,6 +639,18 @@ The act of passing a [message](#message) from one account to another. If the des
 
 The third development stage of Ethereum, launched in October 2017.
 
+### mining {#mining}
+
+The process of verifying transactions and contract execution on the Ethereum blockchain in exchange for a reward in ether with the mining of every block.
+
+### mining pool{#mining-pool}
+
+The pooling of resources by miners who share their processing power and split [mining rewards](#mining-reward).
```
wackerow, line 648: (If we switch "mining rewards" to "block rewards" below)
```suggestion
The pooling of resources by miners who share their processing power and split [block rewards](#block-reward).
```

src/content/glossary/index.md
wackerow: **Note:** This code suggestion actually would need to be relocated up to where it belongs in the list in ABC order

src/content/glossary/index.md
```diff
@@ -558,6 +675,10 @@ Referring to the Ethereum network, a peer-to-peer network that propagates transa
   Networks
 </DocLink>
 
+### network hashrate {#network-hashrate}
+
+The number of hash calculations miners on the Ethereum network can make per second collectively.
+
```
wackerow, line 681: If we're going to have both "hashrate" and "network hashrate", perhaps link to the other here?
```suggestion
### network hashrate {#network-hashrate}

The collective [hashrate](#hashrate) produced by the entire Ethereum mining network.

```

src/content/glossary/index.md
```diff
@@ -613,6 +734,14 @@ An oracle is a bridge between the [blockchain](#blockchain) and the real world.
 
 One of the most prominent interoperable implementations of the Ethereum client software.
 
+### peer {#peer}
+
+Connected computers running Ethereum client software that have identical copies of the [blockchain](#blockchain).
+
+### peer to peer network {#peer-to-peer-network}
+
+A network of computers ([peers](#peer)) that are collectively able to perform functionalities normally only possible with centralized, server-based services.
```
wackerow, line 743:
```suggestion
A network of computers ([peers](#peer)) that are collectively able to perform functionalities without the need for centralized, server-based services.
```

src/content/glossary/index.md
```diff
@@ -621,19 +750,27 @@ An off-chain scaling solution that uses [fraud proofs](#fraud-proof), like [Opti
   Plasma
 </DocLink>
 
+### presale {#presale}
+
+Sale of cryptocurrency before the actual launch of the network.
+
```
wackerow, line 756: Should we note that this was a form of initial ETH issuance that did not come from actual proof-of-work mining or proof-of-stake validating? Okay with just keeping it simple too, just a thought

src/content/glossary/index.md
```diff
@@ -661,6 +798,10 @@ An attack that consists of an attacker contract calling a victim contract functi
   Re-entrancy
 </DocLink>
 
+### reputation {#reputation}
+
+The property of an identity that other entities believe that identity to be either (1) competent at some specific task, or (2) trustworthy in some context, i.e., not likely to betray others even if short-term profitable.
+
```
wackerow, line 804: May push back on this one... as stated this is not really an Ethereum-specific term. Would remove unless we want to rework to be an Ethereum related definition for on-chain reputation, but things are still early on that front so may be best to wait on it for now. 
```suggestion
```

src/content/glossary/index.md
```diff
@@ -709,6 +858,10 @@ A scaling solution that uses a separate chain with different, often faster, [con
   Sidechains
 </DocLink>
 
+### signing {#signing}
+
+Demonstrating that a transaction originated from a specific person by encoding it using a private key.
+
```
wackerow, line 864: Hrm, don't love this definition.
"Signing" more broadly is using the private keys along with an arbitrary message to generate a cryptographic signature that can easily be verified to have come from the private keys behind the public address. You're not really encoding anything, since a signature cannot be decoded to get the original message. Given a public address/key and a message, these can be signed with the private keys, producing a signature, which can be easily verified to confirm that the private keys must have been available to produce the valid signature.
Also, don't love the use of "specific person" here... keys/addresses are not people and people are not keys.

src/content/glossary/index.md
```diff
@@ -813,6 +974,10 @@ Short for "test network," a network used to simulate the behavior of the main Et
   Testnets
 </DocLink>
 
+### token {#token}
+
+A fungible virtual good that can be traded.
+
```
wackerow, line 980: Token's can be both fungible and non-fungible. Not sure I have a great definition here though, but wouldn't declare fungibility here by default

wackerow (changes requested):
@jmcook1186 Juicy =) Went through all the ones you included in the PR and made comments. I'm cool with starting here, and the other options you left in the PR description can be always be raised as issues and added independently. 

Thanks again for your work on this!

src/content/glossary/index.md
```diff
@@ -179,6 +219,22 @@ Decentralized application. At a minimum, it is a [smart contract](#smart-contrac
   Introduction to Dapps
 </DocLink>
 
+### Data availability {#data-availability}
+
+The property of a state that any node connected to the network could download any specific part of the state that they wish to
+
+### decentralization {#decentralization}
+
+The concept of moving the control and execution of processes away from a central entity.
```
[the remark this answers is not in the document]
jmcook1186: hard not to go down rabbit holes here - I opted to be as high level as possible to keep it uncontroversial

src/content/glossary/index.md
jmcook1186: yea good point! 
I guess "virtual goods that can be traded on a blockchain" is better?
@minimalsm opinion?

src/content/glossary/index.md
jmcook1186: keyfiles are kept in a  keystore...

but yea maybe keystore is the better term to include instead of keyfile. Will update.

src/content/glossary/index.md
jmcook1186: yep, fair. Updated.

src/content/glossary/index.md
jmcook1186: or "A tradable virtual good defined in smart contracts on the Ethereum blockchain"

src/content/glossary/index.md
jmcook1186: fair. updated to: "Demonstrating cryptographically that a transaction was approved by the holder of a specific private key."

src/content/glossary/index.md
minimalsm: Personally, I prefer 'digital' over 'virtual' and 'asset' over 'good' for maximum comprehenability.

"A tradable digital asset defined in smart contracts on the Ethereum Blockchain." 

(although, if you want to argue ether is a token, this 👇 might be better). I think either is fine. And anyone who starts talking about soulbound non-fungible tokens can eff off 😆

"A tradable digital asset on the Ethereum Blockchain."

src/content/glossary/index.md
```diff
@@ -621,19 +743,27 @@ An off-chain scaling solution that uses [fraud proofs](#fraud-proof), like [Opti
   Plasma
 </DocLink>
 
+### presale {#presale}
+
+Sale of cryptocurrency before the actual launch of the network.
+
```
minimalsm, line 749:
```suggestion
```

PR #5987, opened 2022-04-09
Add new information to Plasma page

emmanuel-awosika:
Suggest updating the pros vs cons section for Plasma chains and adding a comparison of Plasma against similar scaling solutions, like sidechains and sharding.


minimalsm:
Hey @emmanuel-awosika, thanks for the PR. We will get to this as soon as we can ❤️

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,36 @@ You should have a good understanding of all the foundational topics and a high-l
...
+|                                                                                                                        | Withdrawals are delayed by several days to allow for challenges. For fungible assets this can be mitigated by liquidity providers, but there is an associated capital cost. |
+|                                                                                                                        | If too many users try to exit their Plasma chain simultaneously (due to malicious activity or network failure), they could flood the root chain and congest the network. | 
+
+## How does Plasma differ from sidechains and sharding? 
+
+Plasma, sidechains, and sharding all follow a similar pattern where a collection of "child chains" rely on a "main chain" for consensus. At intervals, cryptographic proofs of transaction validity (hashes) are submitted to the main chain to prevent fraudulent activity on child chains. 
+
+However, the three scaling solutions have slight differences. Here is a quick comparison of Plasma chains, sidechains, and shard chains:
```
wackerow, line 30:
```suggestion
However, the three scaling solutions have slight differences. Here is a quick comparison of plasma chains, sidechains, and shard chains:
```

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,36 @@ You should have a good understanding of all the foundational topics and a high-l
...
+
+### Plasma vs sidechains  
+
+A [sidechain](https://ethereum.org/en/developers/docs/scaling/sidechains/) is a blockchain operating independently but connects to the main blockchain (Ethereum Mainnet) via a two-way bridge. [Bridges](https://ethereum.org/en/bridges/) allow users to exchange tokens between both blockchains, so they can transact on the sidechain—reducing congestion on the base layer and improving scalability.
```
wackerow, line 34: (Internal links should not have the `(https://ethereum.org/en` in front)

```suggestion
A [sidechain](/developers/docs/scaling/sidechains/) is an independently operated blockchain connected to the Ethereum Mainnet via a two-way bridge. [Bridges](/bridges/) allow users to exchange tokens between the two blockchains, so they can transact on the sidechain, reducing congestion on the base layer and improving scalability. 
```

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,36 @@ You should have a good understanding of all the foundational topics and a high-l
...
+## How does Plasma differ from sidechains and sharding? 
+
+Plasma, sidechains, and sharding all follow a similar pattern where a collection of "child chains" rely on a "main chain" for consensus. At intervals, cryptographic proofs of transaction validity (hashes) are submitted to the main chain to prevent fraudulent activity on child chains. 
+
+However, the three scaling solutions have slight differences. Here is a quick comparison of Plasma chains, sidechains, and shard chains:
+
+### Plasma vs sidechains  
+
+A [sidechain](https://ethereum.org/en/developers/docs/scaling/sidechains/) is a blockchain operating independently but connects to the main blockchain (Ethereum Mainnet) via a two-way bridge. [Bridges](https://ethereum.org/en/bridges/) allow users to exchange tokens between both blockchains, so they can transact on the sidechain—reducing congestion on the base layer and improving scalability. 
+With sidechains, there's the risk of losing your funds. Sidechains use a separate consensus mechanism and, ergo, cannot rely on the security of Ethereum Mainnet (Layer 1).
```
wackerow, line 35:
```suggestion
Since sidechains use a separate consensus mechanism, and because these chains are typically much smaller than Ethereum Mainnet, bridging assets to these chains involves increased risk to your funds. Given the lack of security guarantees inherited from Mainnet in the sidechain model, users risk total loss of funds in the event of an attack on the sidechain.```

File: src/content/developers/docs/scaling/plasma/index.md [hunk 4]
```diff
@@ -18,9 +18,36 @@ You should have a good understanding of all the foundational topics and a high-l
...
+
+Conversely, Plasma chains are somewhat safer than sidechains because they derive their security from Ethereum's base layer. Like sidechains, Plasma chains can have a  different consensus protocol. 
+
+The difference is that Plasma publishes Merkle roots for each block on the Ethereum main chain. Block roots are small pieces of information we can use to verify information about transactions that happen on a Plasma chain. If an attack happens on a Plasma chain, users can safely exit to the main chain and withdraw their funds using the appropriate proofs.
```
wackerow, line 39:
```suggestion
Conversely, plasma chains derive their security from Mainnet are somewhat safer than sidechains.  Both sidechains and plasma chains can have different consensus protocols, but the difference is that plasma chains publishes Merkle roots for each block on Ethereum Mainnet. Block roots are small pieces of information we can use to verify information about transactions that happen on a plasma chain. If an attack happens on a plasma chain, users can safely exit to Mainnet and withdraw their funds using the appropriate proofs. 
```

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,36 @@ You should have a good understanding of all the foundational topics and a high-l
...
+With sidechains, there's the risk of losing your funds. Sidechains use a separate consensus mechanism and, ergo, cannot rely on the security of Ethereum Mainnet (Layer 1).
+
+Conversely, Plasma chains are somewhat safer than sidechains because they derive their security from Ethereum's base layer. Like sidechains, Plasma chains can have a  different consensus protocol. 
+
+The difference is that Plasma publishes Merkle roots for each block on the Ethereum main chain. Block roots are small pieces of information we can use to verify information about transactions that happen on a Plasma chain. If an attack happens on a Plasma chain, users can safely exit to the main chain and withdraw their funds using the appropriate proofs. 
+
+A simple way to differentiate between Plasma and sidechains is that Plasma chains are non-custodial, while sidechains are custodial.
```
wackerow, line 41:
```suggestion
```
Don't think I'd introduce these terms like this (with existing meaning in crypto) without explaining it further. Don't think it's really necessary though and would consider just removing.

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,36 @@ You should have a good understanding of all the foundational topics and a high-l
...
+A simple way to differentiate between Plasma and sidechains is that Plasma chains are non-custodial, while sidechains are custodial. 
+
+### Plasma vs sharding
+
+Both Plasma chains and [shard chains](https://ethereum.org/en/upgrades/shard-chains/) periodically publish cryptographic proofs on the main chain. However, both have different security properties. 
+Shard chains are tightly coupled to the main chain, but plasma chains are not. Even if one shard block is invalid, the entire chain will reorganize and discard the invalid block. Thus, shard chains can be as safe as the main chain. 
+
+While Plasma chains can benefit from Layer 1 security, they are not as tightly linked to the main chain as shard chains. It's possible to commit an invalid Plasma block header to the main chain since the larger network has little information about the state of the Plasma chain. 
+
+With sharding, the main chain knows how sharded sub-chains should operate and can reject an invalid shard transition. The upside to Plasma is the effects of a malicious activity can be restricted to the child chain, while the effect on malicious activity on shard chains would affect the entire network.
```
wackerow, line 50: Overall a little hesitant to publish anything more on shard chains for the moment, given danksharding updates, and  especially hesitant in the active voice too, where it gives the impression this is live on Ethereum today.

I think we can save this still though, to fill out the comparison. I would consider rephrasing this a bit to be slightly more generalized, and discuss the general concept of a shard of data, and how its validity can be enforced by Mainnet, compared to plasma where it is not.

Also, afaik no need to capitalize "plasma." (and link needs domain/language trimmed out)

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -35,5 +62,6 @@ Multiple projects provide implementations of Plasma that you can integrate into
 
 - [EthHub on Plasma](https://docs.ethhub.io/ethereum-roadmap/layer-2-scaling/plasma/)
 - [Learn Plasma](https://www.learnplasma.org/en/)
+- [Sidechains vs Plasma vs Sharding](https://vitalik.ca/general/2019/06/12/plasma_vs_sharding.html)
```
wackerow, line 65: Another good one by V that touches on this
```suggestion
- [A quick reminder of what "shared security" means and why it's so important](https://old.reddit.com/r/ethereum/comments/sgd3zt/a_quick_reminder_of_what_shared_security_means/)
- [Sidechains vs Plasma vs Sharding](https://vitalik.ca/general/2019/06/12/plasma_vs_sharding.html)
```

wackerow (changes requested):
@emmanuel-awosika Thanks again for the submission! Left some comments to help polish this... 
In short, as-written my concern is it makes sidechains sound more closely tied to Mainnet than they are. Let me know what you think, but I feel like we could adjust a few terms to help reduce confusion.

Also left a note on comparing to shard chains. Afaik, I don't believe actual shard _chains_ are a part of danksharding. I think we could generalize this to be more accurate when just looking at data sharding, or potentially just hold on it. Curious if any others have thoughts.

src/content/developers/docs/scaling/plasma/index.md
emmanuel-awosika: Apologies for that. @minimalsm only explained how to use internal link paths in content for the website recently.

src/content/developers/docs/scaling/plasma/index.md
emmanuel-awosika: Sure! Just realized we're trying to reach crypto-curious people, too, not just experts (who might understand the terms).

src/content/developers/docs/scaling/plasma/index.md
emmanuel-awosika: So, I reworked the content here and took out the last paragraph. Let me know if the changes are good enough.

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,36 @@ You should have a good understanding of all the foundational topics and a high-l
...
+|                                                                                                                        | Withdrawals are delayed by several days to allow for challenges. For fungible assets this can be mitigated by liquidity providers, but there is an associated capital cost. |
+|                                                                                                                        | If too many users try to exit their Plasma chain simultaneously (due to malicious activity or network failure), they could flood the root chain and congest the network. | 
+
+## How does Plasma differ from sidechains and sharding? 
+
+Plasma, sidechains, and sharding all follow a similar pattern where a collection of "child chains" rely on a "main chain" for consensus. At intervals, cryptographic proofs of transaction validity (hashes) are submitted to the main chain to prevent fraudulent activity on child chains.
```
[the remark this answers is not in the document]
wackerow: > I was trying to show the baseline similarities between the thre

> Vitalik uses a "hub-and-spoke" model to describe all three

For sure... honestly a bit torn on this approach. I just hesitate because the spokes connecting a side chain to Mainnet are _significantly_ weaker than the spokes that will connect data shards to the Beacon Chain, or that connect plasma chains to Mainnet...

> only problem is that the consensus backbone part may not really apply to sidechains.

...to that point exactly. Just want to try and make that clear for readers, since misunderstanding these spokes can potentially lead to loss of funds. I think I would focus more on the differentiating factors here, while keeping any similarities high level.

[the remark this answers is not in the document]
emmanuel-awosika: Yeah, so I changed the language to suggest that the only similarity between the three is that they connect to Ethereum Mainnet "in some way." I also stated that the level and strength of those connections vary, which can have implications for security. I think that keeps the similarities high-level enough, don't you think?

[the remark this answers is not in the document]
minimalsm: > Vitalik uses a "hub-and-spoke" model to describe all three--that is, the main chain serves as a "consensus backbone" while the connected chains (plasma child chains, sidechains, and shard chain) handle actual transactions. We can use this mental model, but the only problem is that the consensus backbone part may not really apply to sidechains.

One thing to be cognizant of here is how fast Ethereum is moving. Terminology (and Vitalik's views, generally) from 2019, are probably dramatically different than they are now.

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,31 @@ You should have a good understanding of all the foundational topics and a high-l
...
+Since sidechains use a separate consensus mechanism, and because these chains are typically much smaller than Ethereum Mainnet, bridging assets to these chains involves increased risk to your funds. Given the lack of security guarantees inherited from Mainnet in the sidechain model, users risk total loss of funds in the event of an attack on the sidechain.```
+
+Conversely, plasma chains derive their security from Mainnet are somewhat safer than sidechains.  Both sidechains and plasma chains can have different consensus protocols, but the difference is that plasma chains publishes Merkle roots for each block on Ethereum Mainnet. Block roots are small pieces of information we can use to verify information about transactions that happen on a plasma chain. If an attack happens on a plasma chain, users can safely exit to Mainnet and withdraw their funds using the appropriate proofs.
```
minimalsm, line 37:
```suggestion
Conversely, plasma chains derive their security from Mainnet. This makes them measurably more secure than sidechains.  Both sidechains and plasma chains can have different consensus protocols, but the difference is that plasma chains publish Merkle roots for each block on Ethereum Mainnet. Block roots are small pieces of information we can use to verify information about transactions that happen on a plasma chain. If an attack happens on a plasma chain, users can safely withdraw their funds back to Mainnet using the appropriate proofs. 
```

minimalsm: Do we explain how they derive security from Mainnet?

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,31 @@ You should have a good understanding of all the foundational topics and a high-l
...
+|  Plasma chains can be adapted to specific use-cases that are unrelated to the main chain. Anyone, including businesses, can customize Plasma smart contracts to provide scalable infrastructure that works in different contexts.                                                       | Relies on one or more operators to store data and serve it upon request.                                                                                              |
+|                                                                                                                        | Withdrawals are delayed by several days to allow for challenges. For fungible assets this can be mitigated by liquidity providers, but there is an associated capital cost. |
+|                                                                                                                        | If too many users try to exit their Plasma chain simultaneously (due to malicious activity or network failure), they could flood the root chain and congest the network. |
```
minimalsm, line 24: We should phrase everything in relation to Mainnet. In this context, is the root chain ever not Mainnet?

```suggestion
|                                                                                                                        | If too many users try to exit simultaneously, Ethereum Mainnet could get congested. | 
```

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,31 @@ You should have a good understanding of all the foundational topics and a high-l
...
+|                                                                                                                        | Withdrawals are delayed by several days to allow for challenges. For fungible assets this can be mitigated by liquidity providers, but there is an associated capital cost. |
+|                                                                                                                        | If too many users try to exit their Plasma chain simultaneously (due to malicious activity or network failure), they could flood the root chain and congest the network. | 
+
+## How does Plasma differ from sidechains and sharding?
+
+Plasma, sidechains, and sharding are fairly similar because they all connect to Ethereum Mainnet in some way. However, the level and strength of these connections vary, which affects the security properties of each scaling solution. 
+
+Here's a quick comparison of plasma chains, sidechains, and shard chains: 
+
```
minimalsm, line 31: Not crazy about the tone here (docs tone generally is quite authoritative). I think we can remove this and let headers guide users without losing any clarity. What do you think?

```suggestion
```

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,31 @@ You should have a good understanding of all the foundational topics and a high-l
...
+### Plasma vs sidechains
```
minimalsm, line 32:
```suggestion
### Plasma vs sidechains {#plasma-vs-sidechains}
```

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,31 @@ You should have a good understanding of all the foundational topics and a high-l
...
+
+## How does Plasma differ from sidechains and sharding?
+
+Plasma, sidechains, and sharding are fairly similar because they all connect to Ethereum Mainnet in some way. However, the level and strength of these connections vary, which affects the security properties of each scaling solution. 
+
+Here's a quick comparison of plasma chains, sidechains, and shard chains: 
+
+### Plasma vs sidechains  
+
+A [sidechain](/developers/docs/scaling/sidechains/) is an independently operated blockchain connected to the Ethereum Mainnet via a two-way bridge. [Bridges](/bridges/) allow users to exchange tokens between the two blockchains, so they can transact on the sidechain, reducing congestion on the base layer and improving scalability.
```
minimalsm, line 34:
```suggestion
A [sidechain](/developers/docs/scaling/sidechains/) is an independently operated blockchain connected to Ethereum Mainnet via a two-way bridge. [Bridges](/bridges/) allow users to exchange tokens between the two blockchains to transact on the sidechain, reducing congestion on Ethereum Mainnet and improving scalability.
```

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,31 @@ You should have a good understanding of all the foundational topics and a high-l
...
+## How does Plasma differ from sidechains and sharding?
+
+Plasma, sidechains, and sharding are fairly similar because they all connect to Ethereum Mainnet in some way. However, the level and strength of these connections vary, which affects the security properties of each scaling solution. 
+
+Here's a quick comparison of plasma chains, sidechains, and shard chains: 
+
+### Plasma vs sidechains  
+
+A [sidechain](/developers/docs/scaling/sidechains/) is an independently operated blockchain connected to the Ethereum Mainnet via a two-way bridge. [Bridges](/bridges/) allow users to exchange tokens between the two blockchains, so they can transact on the sidechain, reducing congestion on the base layer and improving scalability. 
+Since sidechains use a separate consensus mechanism, and because these chains are typically much smaller than Ethereum Mainnet, bridging assets to these chains involves increased risk to your funds. Given the lack of security guarantees inherited from Mainnet in the sidechain model, users risk total loss of funds in the event of an attack on the sidechain.```
```
minimalsm, line 35: Slight tweak for clarity 
```suggestion
Sidechains use a separate consensus mechanism and are typically much smaller than Ethereum Mainnet. As a result, bridging assets to these chains involves increased risk; given the lack of security guarantees inherited from Ethereum Mainnet in the sidechain model, users risk the loss of funds in an attack on the sidechain.
```

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,31 @@ You should have a good understanding of all the foundational topics and a high-l
...
+
+Conversely, plasma chains derive their security from Mainnet are somewhat safer than sidechains.  Both sidechains and plasma chains can have different consensus protocols, but the difference is that plasma chains publishes Merkle roots for each block on Ethereum Mainnet. Block roots are small pieces of information we can use to verify information about transactions that happen on a plasma chain. If an attack happens on a plasma chain, users can safely exit to Mainnet and withdraw their funds using the appropriate proofs. 
+
+### Plasma vs sharding
```
minimalsm, line 39:
```suggestion
### Plasma vs sharding {#plasma-vs-sharding}
```

src/content/developers/docs/scaling/plasma/index.md
```diff
@@ -18,9 +18,31 @@ You should have a good understanding of all the foundational topics and a high-l
...
+
+Both plasma chains and [shard chains](/upgrades/shard-chains/) periodically publish cryptographic proofs on the main chain. However, both have different security properties. 
+
+Shard chains commit "collation headers" to Mainnet, which contain detailed information about each data shard. As such, nodes on the main chain verify and enforce the validity of data shards. This reduces the possibility of invalid shard transitions and protects the network against malicious activity.  
+
+Plasma is different because Mainnet only receives minimal information about the state of child chains. This means Mainnet cannot effectively verify transactions conducted on child chains, making the latter less secure.
```
minimalsm, line 45:
```suggestion
Both plasma chains and [shard chains](/upgrades/shard-chains/) periodically publish cryptographic proofs to Ethereum Mainnet. However, both have different security properties. 

Shard chains commit "collation headers" to Mainnet containing detailed information about each data shard. Nodes on Mainnet verify and enforce the validity of data shards, reducing the possibility of invalid shard transitions and protecting the network against malicious activity.  

Plasma is different because Mainnet only receives minimal information about the state of child chains. This means Mainnet cannot effectively verify transactions conducted on child chains, making them less secure. 
```

minimalsm (changes requested):
Looking great @emmanuel-awosika, a few small tweaks suggested. LMK if you have any questions 😀

src/content/developers/docs/scaling/plasma/index.md
emmanuel-awosika: Sure, we could remove it. I'm used to writing website content, so some of the language carried over. I may have used that line in other places, too, but I'll keep it out of new content.

emmanuel-awosika:
> LGTM 🎉 let's ship it 😎 thanks again @emmanuel-awosika

Thanks for helping with creating the page, @minimalsm. :)

PR #5934, opened 2022-04-07
Update Proof-of-stake (PoS)  by adding a point for cons.

chrboesch:
Updated site Proof-of-stake (PoS)  by adding a point for cons.

## Decription
I added another point for the table pro/cons.

## Related Issue
If you're an expert on the topic and want to contribute, edit this page and sprinkle it with your wisdom. 
Yes, I am.

PR #6001, opened 2022-04-11
Updated site Proof-of-stake (PoS) by adding a point for cons.

chrboesch:
## Description

I added another point for the table pro/cons: No interest for Proof-of-Stake below the minimum amount.
I think that's an important point that can be worked around with a pool, but still remains.

## Related Issue

If you're an expert on the topic and want to contribute, edit this page and sprinkle it with your wisdom.
Yes, I am.


minimalsm:
Hey @chrboesc, I've already addressed your first point of this PR here: https://github.com/ethereum/ethereum-org-website/pull/5934

As for the second point, I'm not sure I understand it? Are you saying the minimum threshold for becoming a solo staker is a limitation? If so, this is an intentional limitation of how Ethereum has implemented its consensus mechanism (read: feature, not a bug), rather than a problem with proof-of-stake itself.

chrboesch:
@minimalsm Shure, it is not a bug. But it is a restriction (in this case intentional) that could prevent a miner from participating in mining. My point is to list all the points, even the (supposedly) negative ones. This is the only way to argue and convince. I think proof-of-stake is the right way to go. However, this also means dealing with the limits (intentional or unintentional). And thus find the best possible solution.

PR #6015, opened 2022-04-12
Add redirect for non-hyphenated layer2 path

wackerow:
Found myself forgetting if the hyphen was required or not. Added simply redirect to allow users to just type ethereum.org/layer2 and not have to worry about the hyphen.


minimalsm (approved):

minimalsm:
Nice idea 😎

PR #6024, opened 2022-04-13
Update Twitter reference

samajammin:
## Description

Update reference to earlier tweet.

## Related Issue

None but a few folks pointed out this tweet that's earlier than the current one we reference:
https://twitter.com/himgajria/status/1266415636789334016


minimalsm (approved):
Ty for doing this @samajammin

PR #6029, opened 2022-04-14
New color for the theme (mostly seen on safari)

nloureiro:
The theme color is set to #1c1ce1 and on safari mobile fills the top of the screen. this color does not work on das theme

Screen Shot 2022-04-14 12 16 33 AM

changed from #1c1ce1 to #222222 to work good on light and dark theme

Open the ethereum.org on safari mobile or even in the desktop when scrolling

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -49,7 +49,7 @@ module.exports = {
         short_name: `ethereum.org`,
         start_url: `/en/`,
         background_color: `#ffffff`,
-        theme_color: `#1c1ce1`,
+        theme_color: `#222222`,
         display: `standalone`,
         icon: `src/assets/favicon.png`,
       },


minimalsm (commented):
Nice catch, @nloureiro! I wasn't aware of this behaviour but can see how it would be annoying (it's also very similar to the personal hotspot banner indicator, which might be confusing). 

One thing I'm not crazy about is how over scrolling on the light theme shows the darker colour:

8c7ff8c069b89e1dc172260f5cb24303

As we already handle theming elsewhere, I think we can fix this by explicitly not passing `theme_color` to our meta tag by adding `theme_color_in_head: false` as an option here. [See gatsby docs](https://www.gatsbyjs.com/plugins/gatsby-plugin-manifest/#remove-theme-color-meta-tag).

Example:

0ad7d370fc4452da3fd41148870bb4b1

What do you think?

_I don't **think** this causes any issues anywhere else but would love a second opinion._

samajammin (approved):
Amazing! Thanks for catching this, I've noticed that in the past & was always curious why that was the case 😄

minimalsm:
@all-contributors please add @nloureiro for code

PR #5933, opened 2022-04-07
Optimize load of translation reports

pettinarip:
Currently, the whole translation JSON reports used in `/contributing/translation-program/acknowledgements/` are being bundled in the final js files.

## Description

This PR changes that, loading the files in the Gatbsy data layer in order to use graphql to pull only the necessary data.

PR #6046, opened 2022-04-14
Revert JSON optimization

pettinarip:
Revert #5933 to test if this is the issue with the stuck builds we currently have in GC.


samajammin:
See the email from GC support - I suspect the issue is on their end.

pettinarip:
Not needed

PR #6088, opened 2022-04-21
Add TrueUp Web3 job board

amtaylor:
Added link to a new Web3 job board with 12,800 open jobs


minimalsm:
Hey thanks for the PR :-)

I'm not comfortable adding this for a few reasons:

1. Advertises thousands of jobs but only lets you see ~15, but requires sign up and personal information to see jobs (data farming/selling to recruiters/etc)
2. Requires LinkedIn for sign up (kinda tone-deaf of crypto culture)
3. The crypto jobs it is listing are the same jobs found on all of the other job boards (doesn't add anything novel, new) — I suspect it is mostly scraping Lever/Bamboo, etc...

amtaylor:
Thanks for the response and detailed reasoning. Those are all fair. This does have way more job openings than the other resources and people in web3 are actively using it, but I don't have a problem with your logic/decision.

PR #6138, opened 2022-04-24
Add "how to get onto layer 2" button in layer 2 page hero

corwintines:
## Description
- Add a "how to get onto layer 2" button in the layer 2 page hero that takes the user to the onboard section

### DIFF
--- a/src/pages/layer-2.js
+++ b/src/pages/layer-2.js
@@ -272,7 +272,7 @@ const Layer2Page = ({ data }) => {
         isSecondary: "isSecondary",
       },
       {
-        content: "How to get onto layer 2",
+        content: "Move to layer 2",
         pathId: "how-to-get-onto-layer-2",
         isSecondary: "isSecondary",
       },
--- a/src/pages/layer-2.js
+++ b/src/pages/layer-2.js
@@ -271,6 +271,11 @@ const Layer2Page = ({ data }) => {
         pathId: "use-layer-2",
         isSecondary: "isSecondary",
       },
+      {
+        content: "How to get onto layer 2",
+        pathId: "how-to-get-onto-layer-2",
+        isSecondary: "isSecondary",
+      },
     ],
   }
 
@@ -577,8 +582,8 @@ const Layer2Page = ({ data }) => {
             <p>
               The Ethereum community has taken a strong stance that it would not
               throw out decentralization or security in order to scale. Until{" "}
-              <Link to="/upgrades/shard-chains/">sharding</Link>, Ethereum Mainnet
-              (layer 1) is only able to process{" "}
+              <Link to="/upgrades/shard-chains/">sharding</Link>, Ethereum
+              Mainnet (layer 1) is only able to process{" "}
               <Link to="https://ethtps.info/Network/Ethereum">
                 roughly 15 transactions per second
               </Link>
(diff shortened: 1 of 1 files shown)


minimalsm:
@nloureiro @konopkja, any thoughts on this?

Screenshot 2022-04-28 at 11 36 26

I'm personally not a fan of having multiple secondary buttons, but if we did, is this how we would style them?

minimalsm:
Given the short travel distance, the Primary button feels redundant to me. E.g:

6a22b5d47e90a97d251a508023661d09

What if we did something like this instead?

Screenshot 2022-04-28 at 11 42 22

konopkja:
We know that people are currently clicking on both buttons a lot. (matomo)

We dont know if its just because people like clicking on buttons :) or whether its because of the topic.

If we remove "what is L2" - there is a potential that some users might misinterpret the content of this page.

but maybe that assumption wont be a problem, I am open to removing that anchor and running another heatmap to see its performance in comparison to the current design.

If there is a problem after that, we will know what to do :)

@nloureiro any thoughts?

PS: i think if there are maximum 3 button, 4+ would be problematic from my point of view.  Alternatively, for the new open design system we could think of how to improve onpage navigation (especially thinking of consistency with the use case pages that have different template)

nloureiro:
I'm ok with this; more on the side of being a long page, these anchors can help target different profiles.
What is layer 2 > beginner that wants to learn
Use layer 2 > intermediate that wants to start using L2s
How to get onto layer 2 > more advanced that want a shortcut.

This being said, the decision fatigue might kick in if we have too many options. For me, 3 might be the limit.

ps: for long pages, we might want to create a floating bar with the section title or some floating index of the flow of the page (just a random idea)

minimalsm:
> If there is a problem after that, we will know what to do better. So i am down for your suggestion @minimalsm :)

Weakly held preference, I'll ultimately defer to you guys @konopkja @nloureiro 

> Also mobile version needs to be taken into consideration.

This is current page on mobile with 3 buttons

Screenshot 2022-05-04 at 18 24 46

Worth noting this is made a lot worse on more verbose languages (Ukrarian/Russian)

Screenshot 2022-05-04 at 18 25 23

nloureiro:
> > If there is a problem after that, we will know what to do better. So i am down for your suggestion @minimalsm :)
> 
> Weakly held preference, I'll ultimately defer to you guys @konopkja @nloureiro
> 
> > Also mobile version needs to be taken into consideration.
> 
> This is current page on mobile with 3 buttons
> 
> Screenshot 2022-05-04 at 18 24 46
> 
> Worth noting this is made a lot worse on more verbose languages (Ukrarian/Russian)
> 
> Screenshot 2022-05-04 at 18 25 23


In some more verbose languages, it's not the most elegant solution, but it's not that bad. the benefit of the 3 buttons it's bigger.

src/pages/layer-2.js
```diff
@@ -271,6 +271,11 @@ const Layer2Page = ({ data }) => {
         pathId: "use-layer-2",
         isSecondary: "isSecondary",
       },
+      {
+        content: "How to get onto layer 2",
```
minimalsm, line 275: We should use a dynamic verb in the CTA here IMO

```suggestion
        content: "Move to layer 2",
```

minimalsm (approved):

PR #3330, opened 2021-06-22
Add Chainstack in the range of Enterprise infrastructure providers

neewy:
## Description

Added chainstack.com as an infrastructure provider of Enterprise Ethereum nodes.

PR #6154, opened 2022-04-26
add chainstack provider to "Nodes and clients" page

14MR:
Adds Chainstack as an alternative node provider to "Nodes and clients" page.

## Description

Adds Chainstack as an alternative node provider to "Nodes and clients" page.

## Related Issue

Related PR: https://github.com/ethereum/ethereum-org-website/pull/3330

### DIFF
--- a/src/content/developers/docs/nodes-and-clients/index.md
+++ b/src/content/developers/docs/nodes-and-clients/index.md
@@ -103,7 +103,7 @@ If you're more of a technical user, learn how to [spin up your own node](/develo
 
 ## Alternatives {#alternatives}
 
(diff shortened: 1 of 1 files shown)


minimalsm:
Hey @14MR, thanks for the PR ❤️ we'll review this as soon as we can.

minimalsm:
Hey @14MR, thanks for raising this.

You've highlighted a clear issue here. In the name of credible neutrality, we probably shouldn't be linking out to anyone (our nodes as a service page does this).  It's a bit of a silly situation because, as you can imagine, we merge this, then we get another PR to add another provider, then another, etc... before long the paragraph is 50% service providers and is detrimental to users 😅 

I'll create a card for myself to go through and remove the links already in this section. We should endeavour to only link to services in this way on clearly defined pages or sections (instead of mid-sentence) as we can define clear product listing criteria for pages like this, instead of just how a PR reviewer feels about a specific product.

minimalsm:
@all-contributors please add @14MR for ideas

PR #6183, opened 2022-05-02
Fix grammar so-named -> so named

epicfaace:
There should be no hyphen in "so named".


minimalsm (approved):
Thanks @epicfaace 

Be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And we've just released our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) also 🥳!

minimalsm:
Worth noting here that I still think this sentence structure is a mess and we should revisit it.

PR #6201, opened 2022-05-03
fix template string bug for l2 stats

corwintines:
## Description
- Fix template string for negative percent change for L2 stats


samajammin (approved):
👍

PR #6192, opened 2022-05-02
docs: Fixes the anchor tag for ABI link

tarunbatra:
## Description

The URL pointing to the glossary entry of ABI, has a trailing slash which makes the anchor tag not work. The fix is too remove the trailing slash.

PR #6207, opened 2022-05-04
Remove trailing slashes from glossary links

minimalsm:
## Description

In #6192, a contributor noted that a trailing slash breaks the URL id selector in glossary links. This builds on #6192 to remove all instances of this.

For example:
https://ethereum.org/en/glossary/#mainnet/ (with trailing slash) vs visiting https://ethereum.org/en/glossary/#mainnet (no slash)

## Related Issue
#6192


samajammin (approved):
Great job catching this! ❤️

PR #6208, opened 2022-05-04
Remove trailing slashes from urls with header-ids

minimalsm:
## Description
Follow on from: #6207.

I realised any trailing slash will break a header id. This removes broken occurences on the website.

## Related Issue
#6207


samajammin (approved):
👍

Issue #6202, opened 2022-05-04
Codeblocks in RTL languages


minimalsm:
**Describe the bug**

In RTL languages (Farsi, Arabic), code blocks inherit RTL styling causing them to be reversed and be illegible.

**To Reproduce**

1. Go to https://ethereumorgwebsitedev01-farsifoundational.gtsb.io/fa/developers/docs/accounts/
1. See error

**Expected behavior**

Codeblocks should remain LTR in RTL languages

**Screenshots**

<img width="810" alt="Screenshot 2022-05-04 at 13 19 06" src="https://user-images.githubusercontent.com/62268199/166679769-5ea5486d-731e-485a-b69f-19b31b5604ce.png">

samajammin:
Hmmm I'm not seeing this issue:
<img width="1010" alt="Screen Shot 2022-05-04 at 4 09 44 PM" src="https://user-images.githubusercontent.com/8097623/166699326-5a1068a3-b9fe-4f36-a4a0-2202255d914f.png">

PR #6212, opened 2022-05-04
Codeblocks in RTL languages bug fixed [Solves #6202]

HursitTarcan:
## Description
- Added a `direction` property with `ltr` value to the `CodeBlock` component. This way the content inside of a code block stays in LTR, no matter the language. 

## Describe your changes in detail
- Added: `direction: ltr;` (src/components/Codeblock.js, line 24)

## Related Issue
- Related to [Codeblocks in RTL languages #6202]


minimalsm:
Hey @HursitTarcan, thanks for the PR but this is already solved by #6202 😅 

Please be sure to check if there is an open PR before beginning work ❤️

PR #6216, opened 2022-05-04
Updated index.md [Dev/Docs/Standards]

HursitTarcan:
## Description
- Removed punctuation marks from list. 

## Related Issue
- Not related to an issue.

### DIFF
--- a/src/content/developers/docs/standards/index.md
+++ b/src/content/developers/docs/standards/index.md
@@ -29,10 +29,10 @@ Certain EIPs relate to application-level standards (e.g. a standard smart-contra
 
 ### Token standards {#token-standards}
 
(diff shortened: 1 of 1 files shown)


minimalsm:
Thanks @HursitTarcan, I'm not sure why we'd want to remove these?

HursitTarcan:
@minimalsm I thought about removing those since we have two lists on that page, and the list above doesn't have any punctuation marks. I included a screenshot. 

image

minimalsm:
Hmmm, the list above is only links with attributions (no sentences). I think we should keep them as they are full sentences here but i can see the confusion. Perhaps we should add something to the style-guide on this... 🤔

PR #6225, opened 2022-05-05
add info on multisigs to /smart-contracts/index.md

jmcook1186:
## Description

Adds a simple paragraph on multisig contracts to `/developers/docs/smart-contracts/index.md`.
Satisfies part of wiki migration.

## Related Issue
#5690

### DIFF
--- a/src/content/developers/docs/smart-contracts/index.md
+++ b/src/content/developers/docs/smart-contracts/index.md
@@ -93,6 +93,10 @@ There are ways to get around this using [oracles](/developers/docs/oracles/).
 
 Another limitation of smart contracts is the maximum contract size. A smart contract can be a maximum of 24KB or it will run out of gas. This can be circumnavigated by using [The Diamond Pattern](https://eips.ethereum.org/EIPS/eip-2535).
 
+## Multisig Contracts
+
+"Mulisigs" are contracts that require multiple valid signatures to execute. This is very useful for avoiding single points of failure for contracts holding susbtantial amounts of ether or other tokens. Multisigs also divide responsibility for contract execution and key management between multiple parties and prevent the loss of a single private key leading to irreversible loss of funds. For these reasons, multisig contracts are often used in DAO governance. Multisigs require a majority N of M possible signatures in order to execute. `N = 3, M = 5` and `N = 4, M = 7` are commonly used. A 4/7 multisig requires 4 out of 7 possible valid signatures. This means the funds are still retrievable even if 3 signatures are lost. It also means that a majority of key-holders must agree and sign in order for the contract to execute. Multisig contracts can be used as extra-secure wallets.
+
 ## Smart contract resources {#smart-contract-resources}
 
 **OpenZeppelin Contracts -** **_Library for secure smart contract development._**


src/content/developers/docs/smart-contracts/index.md
```diff
@@ -93,6 +93,10 @@ There are ways to get around this using [oracles](/developers/docs/oracles/).
 
 Another limitation of smart contracts is the maximum contract size. A smart contract can be a maximum of 24KB or it will run out of gas. This can be circumnavigated by using [The Diamond Pattern](https://eips.ethereum.org/EIPS/eip-2535).
 
+## Multisig Contracts
+
+"Mulisigs" are contracts that require multiple valid signatures to execute. This is very useful for avoiding single points of failure for contracts holding susbtantial amounts of ether or other tokens. Multisigs also divide responsibility for contract execution and key management between multiple parties and prevent the loss of a single private key leading to irreversible loss of funds. For these reasons, multisig contracts are often used in DAO governance. Multisigs require a majority N of M possible signatures in order to execute. `N = 3, M = 5` and `N = 4, M = 7` are commonly used. A 4/7 multisig requires 4 out of 7 possible valid signatures. This means the funds are still retrievable even if 3 signatures are lost. It also means that a majority of key-holders must agree and sign in order for the contract to execute. Multisig contracts can be used as extra-secure wallets.
```
minimalsm, line 98:
```suggestion
Mulisigs (multiple signature wallets) are contracts that require multiple valid signatures to execute a transaction. This is very useful for avoiding single points of failure for contracts holding substantial amounts of ether or other tokens. Multisigs also divide responsibility for contract execution and key management between multiple parties and prevent the loss of a single private key leading to irreversible loss of funds. For these reasons, multisig contracts are often used in DAO governance. Multisigs require a majority N of M possible signatures to execute. `N = 3, M = 5` and `N = 4, M = 7` are commonly used. A 4/7 multisig requires 4 out of 7 possible valid signatures. This means the funds are still retrievable even if three signatures are lost. It also means that a majority of key holders must agree and sign to execute the contract. Multisig contracts can be used as extra-secure wallets.
```

src/content/developers/docs/smart-contracts/index.md
```diff
@@ -93,6 +93,10 @@ There are ways to get around this using [oracles](/developers/docs/oracles/).
 
 Another limitation of smart contracts is the maximum contract size. A smart contract can be a maximum of 24KB or it will run out of gas. This can be circumnavigated by using [The Diamond Pattern](https://eips.ethereum.org/EIPS/eip-2535).
 
+## Multisig Contracts
```
minimalsm, line 96:
```suggestion
## Multisig contracts {#multisig}
```

src/content/developers/docs/smart-contracts/index.md
minimalsm: '_Multisig contracts can be used as extra-secure wallets._'

This sentence feels a little redundant. Do people commonly use a multisig by themselves (they are the sole holder of multiple keys)? If not, it feel like this is implied in the paragraph already.

minimalsm: Also, I've suggested adding '(multiple signature wallets)' above but I'm not sure this is correct... feel free to remove

minimalsm (changes requested):
Thanks @jmcook1186 ❤️

src/content/developers/docs/smart-contracts/index.md
wackerow: > Mulisigs (multiple signature wallets) are contracts that

Would suggest avoiding the word "wallet" in this specific context in favor of "account", as it is not something that hold keys, but is in fact a smart contract account. Suggestion: 

```
Multisig (multiple-signature) contracts are smart contract accounts that...
```

---

> multisig contracts are often used in DAO governance.

Would reword to not suggest/encourage this as the norm (there are much better DAO models out there than a simple multisig)

```
multisig contracts can be used for simple DAO governance. 
```

---

> Multisigs require a majority N of M possible signatures to execute. 

It is not required for N to be greater than 50% of M... you could technically have a 1 of 10 multisig or 2 of 10, etc. 

```
Multisigs require N signatures out of M possible acceptable signatures (M > 1, N ≤ M) in order to execute
```

Or ditch the parenthesis addition

---

> A 4/7 multisig requires 4 out of 7 possible valid signatures. This means the funds are still retrievable even if three signatures are lost.

If we're spelling out numbers under 10 (which I agree with)..

```
A 4/7 multisig requires four out of seven possible valid signatures. This means the funds are still retrievable even if three signatures are lost.
```

---

> It also means that a majority of key holders must agree and sign to execute the contract.

Similar to above, doesn't have to be a majority... depends how it's set up. 

```
In this case it also means that a majority of key holders must agree and sign to execute the contract.
```

---

> Multisig contracts can be used as extra-secure wallets.

@minimalsm I don't think this is referring so much to an individual using it this way, but in either case I'm not sure this adds too much, and it's certainly not always the case that they're more secure. Would suggest just removing this sentence. 

*Bringing it all together...*
```suggestion
Multisig (multiple-signature) contracts are smart contract accounts that require multiple valid signatures to execute a transaction. This is very useful for avoiding single points of failure for contracts holding substantial amounts of ether or other tokens. Multisigs also divide responsibility for contract execution and key management between multiple parties and prevent the loss of a single private key leading to irreversible loss of funds. For these reasons, multisig contracts can be used for simple DAO governance. Multisigs require N signatures out of M possible acceptable signatures (where N ≤ M, and M > 1) in order to execute. `N = 3, M = 5` and `N = 4, M = 7` are commonly used. A 4/7 multisig requires four out of seven possible valid signatures. This means the funds are still retrievable even if three signatures are lost. In this case it also means that a majority of key-holders must agree and sign in order for the contract to execute.
```

minimalsm (approved):
LGTM, thanks

PR #6286, opened 2022-05-10
dev/networks: add sepolia faucet

q9f:
## Description

added the recently launched sepolia faucet.


minimalsm (approved):
Thanks as always @q9f

PR #6293, opened 2022-05-10
Feedback widget [Fixes #6252]

### LINKED ISSUE #6252: User feedback widget
Ethereum.org plays a vital role in onboarding new Ethereum users and builders into the ecosystem.  Therefore, it is important for us to make sure that we validate our designs and content with our community. This quarter we’ll enable an easy feedback process for all site visitors directly inside the website. 

We already have a passive feedback widget in the dev section at the end of the articles. Ideally, this feature should be on every page. Also, the current position is problematic because only 15 - 35% of visitors scroll the whole page and thus most visitors are excluded from the option of giving us feedback on the quality of the website. 

### **Feature: Include a floating feedback widget inside each subpage.**

**the dilemma to consider:**
We want to collect enough feedback for it to be relevant, but we do not want to lower the user experience of using the website with aggressive call to actions.

The solution must not be too intrusive, yet it should be constantly available to the user should they decide to give feedback.

We ideated two potential solutions to vote on, but you can propose an alternative as well.

**Widget Behavior (open for discussion):**
- Include a simple feedback widget inside each subpage.
- Always visible if needed
- Option to provide more detailed comments via paperform.co survey
- No personal data collected

**figma link to preliminary designs and feedback questions:**
https://www.figma.com/proto/JCHPZkyTRbFt5yHbKv7g9A/Feedback-widget?page-id=1190%3A1209&node-id=1198%3A1100&viewport=244%2C48%2C0.21&scaling=min-zoom&starting-point-node-id=1198%3A1100

**Additional context**

This relates to the "🗣️ User insights" epic outlined in our Q2 roadmap (#6161)

wackerow:
## Description
Adds FeedbackWidget component per specs in design Figma. 

- Circle in corner always visible, with bottom positioning conditional on the page path (some pages have a bottom nav bar that this accounts for)
- Clicking will open a feedback card, with a simple Yes/No prompt if the page was helpful
- Clicking outside will close, and also has an X icon available
- If user selects Yes or No, a thank you is presented with a link for optional further feedback
- State of card stored in component and currently resets on page refresh
- Matomo events triggers for various actions along the way for further insight

TODO:
- [x] Confirm `url` param is being passed to PaperForm (may need addition of `url` parameter in form)
- [x] Fix re-opening bug when tapping the dot to close
- [x] Fix z-index bug with mobile menu
- [x] Fix color of icon and button text
- [x] Remove trailing slashes from path checks (for determining bottom padding offset)

## Related Issue
[Fixes #6252]


konopkja:
@wackerow the color on white template should be changed to white in the buttons and the circle could be inversed colors (white circle and blue icon)
Screenshot 2022-05-11 at 11 59 38

konopkja:
Screenshot 2022-05-11 at 12 35 27
can menu have higher z-index than the widget? dont think we need to show it on top of the menu

minimalsm:
Positioning feels too tucked away on the corner, especially on large screen sizes, but even on mobile. What do you think @konopkja?

Screenshot 2022-05-11 at 09 16 27

minimalsm:
As a user, when I click the button a second time I expect the pop up to close.

buggif

minimalsm:
Do we feel the feedback widget is necessary on the homepage?

konopkja:
> As a user, when I click the button a second time I expect the pop up to close.
> 
> buggif

interesting, this seems to be mobile issue only. nice catch

konopkja:
> Positioning feels too tucked away on the corner, especially on large screen sizes, but even on mobile. What do you think @konopkja?
> 
> Screenshot 2022-05-11 at 09 16 27

i think for the first iteration of this i wanted to make sure that its not too aggressive as its always visible. I am curious about suggestions how can we make it slightly more prominent, especially on larger screens? One solution could be simply increasing the size.

konopkja:
> 



> Do we feel the feedback widget is necessary on the homepage?

whats your rationale behind this question? I am thinking about it this way: any page can be helpful or not, including the homepage.

minimalsm:
> whats your rationale behind this question? I am thinking about it this way: any page can be helpful or not, including the homepage.

The feedback surveys seem skewed towards content-driven pages (which is fine). Pages like the homepage or community hub are more traffic routing pages. I don't expect users to 'learn' very much from the homepage.

nloureiro:
Screen Shot 2022-05-11 04 34 15 PM
Screen Shot 2022-05-11 04 32 33 PM

I prefer the icon in white on the dark theme. Does anyone have an opinion on this?

wackerow:
image
image

In the Figma the icon color matched the background when not opened, and matched the text color when open. Sounds like we really just want it to be white all the time, regardless of light/dark mode.

wackerow:
And @nloureiro, I agree. I think it looks best with it always being a white foreground, on both light and dark mode. 

Just pushed patches for pretty much everything noted, except what to do on XL screens. Open to suggestions, but don't think that needs to be a blocker on this. 

Will double check when preview build finishes, and if it looks good from my end will check off the above TODOs and flip this back to "needs review."

src/components/FeedbackWidget.js
```diff
@@ -0,0 +1,254 @@
...
+
+  const getSurveyUrl = () => {
+    if (!feedbackSubmitted) return null
+    if (isStaking) return surveyUrls.staking[isHelpful ? "yes" : "no"]
+    return surveyUrls.default[isHelpful ? "yes" : "no"]
+  }
+
+  const pathsWithBottomNav = ["/staking", "/dao", "/defi", "/nft"]
+
+  const bottomOffset = () => {
```
pettinarip, line 189: Consider using https://reactjs.org/docs/hooks-reference.html#usememo in this case. You will receive the offset value instead of a function and you will only be calling it when `location` changes.

wackerow: Yeah good call.. I tried this originally and it caused an error but I think I figured out what I did wrong. Will patch this and push 👍🏻

wackerow: @pettinarip Hm actually, getting some issues depending on node version. With Node 18 I don't have issues, but with anything lower I get an error:

```log
  Error: Unable to deserialize cloned data due to invalid or unsupported v
  ersion.
  
  - v8.js:343 deserialize
    v8.js:343:7
  
  - serializer.js:249 deserialize
    [ethereumorg-website]/[@parcel]/core/lib/serializer.js:249:48
  
  - RequestTracker.js:854 loadRequestGraph
    [ethereumorg-website]/[@parcel]/core/lib/RequestTracker.js:854:52
  
  - RequestTracker.js:821 Function.init
    [ethereumorg-website]/[@parcel]/core/lib/RequestTracker.js:821:17
  
  - Parcel.js:256 Parcel._init
    [ethereumorg-website]/[@parcel]/core/lib/Parcel.js:256:28
  
  - Parcel.js:273 Parcel.run
    [ethereumorg-website]/[@parcel]/core/lib/Parcel.js:273:7
  
  - compile-gatsby-files.ts:48 compileGatsbyFiles
    [ethereumorg-website]/[gatsby]/src/utils/parcel/compile-gatsby-files.t
    s:48:29
  
  - initialize.ts:171 initialize
    [ethereumorg-website]/[gatsby]/src/services/initialize.ts:171:3
  

not finished compile gatsby files - 0.395s

error Command failed with exit code 1.
```

https://support.gatsbyjs.com/hc/en-us/articles/360052322954-Environment-Variables-Specific-to-Gatsby-Cloud

Looks like Gatsby defaults to 14, and I don't see an override anywhere. Going to hold on patching this for now given the node version issue. Should still work just fine without the `useMemo` just might make it a bit less efficient. Curious your take on this Pablo.

pettinarip: Hum, weird error. Can't see why this is related to the node version 🤔 Will take a closer look later on this but feel free to skip this, its totally a nice to have.
Just to confirm, this is an issue you see in your local or in GC?

wackerow: Local... just committed it to give it a try. If the build fails I'll just revert

wackerow: I have a feeling the node version errors I'm getting are related to Apple silicon

wackerow:
Noticed as well the dot only needs to be elevated on screen sized below `breakpoint.l` (1024px), which is when the bottom menu is displayed on certain pages. Added a commit to use this offset within a media query.

src/components/FeedbackWidget.js
wackerow: Yep, pretty sure that's it. All good here

samajammin:
When clicking on the widget on the [defi page](https://ethereumorgwebsitedev01-feedbackwidget.gtsb.io/en/defi/), the popup appears awkwardly high up for me (Brave on Mac): 
Screen Shot 2022-05-13 at 10 35 46 AM

Interestingly this only appears to be an issue on dark mode. Light mode looks fine:
Screen Shot 2022-05-13 at 10 36 44 AM

samajammin:
Should the feedback widget "reset" after every page transition? I think it should.

Here's the flow I experienced:
1. Navigate to page (e.g. homepage)
2. Open widget & give feedback (clicked "yes" this is helpful)
3. Close widget (without clicking "Open short survey")
4. Navigate to different page (e.g. languages page)
5. Open widget - it says “thanks for feedback” & prompts me to "Open short survey"
    - Shouldn’t it start fresh on new page? i.e. "Is this page helpful?"

samajammin:
I like how clicking on the widget dulls out the rest of the page (with the overlay). Should the overlay also overlay the site nav? I think it should. Currently it doesn't:
Screen Shot 2022-05-13 at 10 41 18 AM

konopkja:
> I like how clicking on the widget dulls out the rest of the page (with the overlay). Should the overlay also overlay the site nav? I think it should. Currently it doesn't: Screen Shot 2022-05-13 at 10 41 18 AM

agreed, it should

konopkja:
> Should the feedback widget "reset" after every page transition? I think it should.
> 
> Here's the flow I experienced:
> 
> 1. Navigate to page (e.g. homepage)
> 2. Open widget & give feedback (clicked "yes" this is helpful)
> 3. Close widget (without clicking "Open short survey")
> 4. Navigate to different page (e.g. languages page)
> 5. Open widget - it says “thanks for feedback” & prompts me to "Open short survey"
>    
>    * Shouldn’t it start fresh on new page? i.e. "Is this page helpful?"

yes it should reset

samajammin:
Noticed on mobile (iPhone 13, Safari) the widget occasionally jumps in position when navigating between pages - not sure why. Only happened on some pages:

e.g. NFT page:
IMG_9710

e.g. Staking page:
IMG_9711

Refreshing the page appears to fix.

samajammin:
Should we have a state for after a user clicks on "Open short survey"?

Currently clicking it does nothing:
Screen Shot 2022-05-13 at 10 50 13 AM

Perhaps we:
A) thank them, with prompt to join our Discord for support if needed
B) close the widget
C) keep as is

konopkja:
> 

I vote for closing the widget and reseting it once user clicks on the survey button.

There is a link to discord presented at the end of the survey. 

I also like the idea of navigating users to discord earlier, but would wait with its implementation for now.

src/components/FeedbackWidget.js
pettinarip: Cool! good to know.

wackerow:
> There is a link to discord presented at the end of the survey.

If we're presenting the Discord link at the end of the Paperform, I agree it's easiest and least obtrusive to just close the widget when the survey link is opened. 

> I vote for closing the widget and reseting it once user clicks on the survey button.

Agree with closing, but disagree we should reset the component state. At this point the user has responded with 👍🏻 👎🏻  and is still on the same page... I think leaving the "Thank you" content with the link to the survey (in case they closed it, or need it again) is appropriate, while this will all still reset if the page is refreshed.

TODO's gathered from feedback:

- [x] Fix dimming region to include nav bar
- [x] Fix layout bug for callout container on "bottom menu" pages
- [x] Reset component on navigation
- [x] Close prompt on survey open
- [ ] Fix dot position on mobile when navigating between pages

This last one is because it currently uses fixed positioning and is being rendered from the `Layout.js` component for simplicity. This avoids having the import this into every template and page, but leaves us limited in the component ability to respond to the layout of each page.
This approach adds some logic to check the path for the `use-cases` and `staking` pages which all have a bottom menu on mobile that needs to be accounted for in the layout, and adjusts the dot/widget position up another 6.75rem in those cases. Works find lower in the page, but at the top (like on page load) the bottom menu is not yet visible, and only appears after scrolling down a little. 

So this causes the dot to look awkwardly elevated at the top ofon a

wackerow:
ages

konopkja:
Agree, when user is still on the same page it could keep the thank you page. Its more about when a new page loads user should be able to provide feedback again on that new page.

corwintines:
Screen Shot 2022-05-15 at 11 07 24 PM

I'm wondering if there's a way to have this button scroll up after a user has scrolled past a certain point on the page, vs leaving it here on the page load. @wackerow if you want to brainstorm/pair on this let me know, I think we could figure something out around this.

src/components/FeedbackWidget.js
```diff
@@ -0,0 +1,278 @@
...
+  top: 0.5rem;
+  cursor: pointer;
+  &:hover {
+    transform: scale(1.1);
+    transition: transform 0.2s ease-in-out;
+  }
+  transition: transform 0.2s ease-in-out;
+`
+
+const FeedbackWidget = ({ prompt, className }) => {
```
corwintines, line 139: I don't see `prompt` used anywhere. Is this for future-proofing in case we want to have a custom prompt? If not, I think we can remove it for now.

samajammin: Agreed if it's not used we should remove.

samajammin:
> Two options:
> We hold and investigate a better way to make this responsive to content within each page so it dynamically adjusts to bottom menus
> We accept the occasional awkward dot height for the benefit of simplicity.
> I'd prefer 2.. changes have been pushed, lemme know what you all think.

@wackerow what if we applied the same behavior for this feedback widget on mobile as we do for those bottom menus? Makes sense to me that they would appear at the same time - neither are needed right on initial page load.

corwintines:
@samajammin @wackerow pushed an experiment up using margin-top to offset the button until a certain height (150vh in this case). The feedback widget will be off-screen until the user scrolls down a certain amount and then it will scroll in. Let me know what you think.

wackerow:
LGTM... thoughts from others? Anything that should block this at this point?

wackerow:
Another Gatsby Cloud time-out... re-triggered deploy for cfdc642 and it build successfully, just not updating here:
https://ethereumorgwebsitedev01-feedbackwidget.gtsb.io/

corwintines (approved):

konopkja:
Untitled

i noticed this little weird behavior loop, not very important, but recording it here.

PR #6300, opened 2022-05-11
Filter out stablecoins not on Ethereum

corwintines:
## Description
- update stablecoins filter to take out stablecoins not on Ethereum.

### DIFF
--- a/src/pages/stablecoins.js
+++ b/src/pages/stablecoins.js
@@ -309,7 +309,6 @@ const StablecoinsPage = ({ data }) => {
       DUSD: { type: CRYPTO, url: "https://dusd.finance/" },
       PAXG: { type: ASSET, url: "https://www.paxos.com/paxgold/" },
       AMPL: { type: ALGORITHMIC, url: "https://www.ampleforth.org/" },
-      UST: { type: ALGORITHMIC, url: "https://www.terra.money/" },
       FRAX: { type: ALGORITHMIC, url: "https://frax.finance/" },
       MIM: { type: ALGORITHMIC, url: "https://abracadabra.money/" },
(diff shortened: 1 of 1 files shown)


wackerow (approved):

src/pages/stablecoins.js
```diff
@@ -322,14 +321,25 @@ const StablecoinsPage = ({ data }) => {
...
-          "https://api.coingecko.com/api/v3/coins/markets?vs_currency=usd&order=market_cap_desc&per_page=250&page=1&sparkline=false"
+
+        const ethereumData = await getData(
+          "https://api.coingecko.com/api/v3/coins/markets?vs_currency=usd&category=ethereum-ecosystem&order=market_cap_desc&per_page=100&page=1&sparkline=false"
+        )
+        const stablecoinData = await getData(
+          "https://api.coingecko.com/api/v3/coins/markets?vs_currency=usd&category=stablecoins&order=market_cap_desc&per_page=100&page=1&sparkline=false"
+        )
+
+        const filteredData = stablecoinData.filter(
```
samajammin, line 332: Perhaps we should add a comment to make clear what's happening here? Or perhaps a different variable name instead of `filteredData` to clarify, e.g. `ethereumStablecoinData`?

samajammin (commented):
One small comment - agree with this change overall, I think we should only display stablecoins on Ethereum 👍

src/pages/stablecoins.js
corwintines: Will give another pass on variable names, and some documentation. Guilty of not writing enough of that.

corwintines:
@samajammin @minimalsm addressed the comments :)

PR #6310, opened 2022-05-12
Layout patch [Fixes #6228]

### LINKED ISSUE #6228: All CardList components now have bottom margin
**Describe the bug**

All CardLists now have a bottom margin, which I don't believe is intended. The background color is the same despite the borders:

e.g. on https://ethereum.org/en/get-eth/#country-picker
Screen Shot 2022-05-05 at 3 55 15 PM

**To Reproduce**

See example link above.

**Expected behavior**

I don't think we should have a margin here...

This is what those CardLists use to look like:
Screen Shot 2022-05-05 at 3 59 26 PM

**Screenshots**

See above.

All

All

**Additional context**

Looks like it came from this commit?
https://github.com/ethereum/ethereum-org-website/pull/5748/commits/746f3e10a67804e9aa000b9a9e0d537d3c76b467

I suspect @corwintines was only trying to apply this to the staking resource links:
Screen Shot 2022-05-05 at 3 56 45 PM
Which I do think look better separated by margins but we should apply to only that instance vs. across the app.

wackerow:
## Description
Removes `margin-bottom` from `CardList.js`, and custom adds back a 1rem gap for the staking guides list specifically.

## Related Issue
[Fixes #6228]

### DIFF
--- a/src/components/CardList.js
+++ b/src/components/CardList.js
@@ -36,7 +36,6 @@ const ItemLink = styled(Link)`
   padding: 1rem;
   width: 100%;
   color: #000000;
-  margin-bottom: 1rem;
   &:hover {
     border-radius: 4px;
     box-shadow: 0 0 1px ${(props) => props.theme.colors.primary};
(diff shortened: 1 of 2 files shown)


corwintines (approved):
Looks good, thanks for fixing my bad styling 🤦

Issue #5998, opened 2022-04-11
Filter out preview deploy traffic from Matomo production data


samajammin:
**Describe the bug**

Not 100% sure this is a bug but noticed cases in Matomo of data reported from our staging environments (i.e. preview deploys), e.g. https://matomo.ethereum.org/index.php?module=CoreHome&action=index&idSite=4&period=day&date=yesterday#?idSite=4&period=range&date=2022-03-21,2022-04-05&segment=&category=CustomReports_CustomReports&subcategory=13

**To Reproduce**

1. Go to https://matomo.ethereum.org/index.php?module=CoreHome&action=index&idSite=4&period=day&date=yesterday#?idSite=4&period=range&date=2022-03-21,2022-04-05&segment=&category=CustomReports_CustomReports&subcategory=13
2. See `ethereumorgwebsitedev...` preview deploy entry

**Expected behavior**

Matomo should only collect data from production traffic (i.e. https://ethereum.org/)

We should
1. confirm if we are indeed collecting preview deploy traffic
2. if so, sort out how to filter that traffic out

**Screenshots**

![Image 2022-04-11 at 11 17 22 AM](https://user-images.githubusercontent.com/8097623/162804486-5e1e04c0-42e6-4ea1-a2df-1a7e256194f3.jpg)

**Additional context**

This is the plugin we use:
https://github.com/kremalicious/gatsby-plugin-matomo

We may want to configure options to solve for this:
https://github.com/kremalicious/gatsby-plugin-matomo#options

pettinarip:
In GC `NODE_ENV` is `production` for Production Builds and Pull Request Builds. https://support.gatsbyjs.com/hc/en-us/articles/360052322954-Environment-Variables-Specific-to-Gatsby-Cloud

What I'm planning to do is to basically avoid loading the plugin if `GATSBY_IS_PREVIEW` is true. Basically because we done have control over that on the plugin.

wackerow:
Fixed with #6366

PR #6366, opened 2022-05-17
Avoid loading matomo in preview deploys

pettinarip:
We are sending testing data from our preview deploys to Matomo because the plugin sends data when `NODE_ENV === 'production'`.
Ref. https://support.gatsbyjs.com/hc/en-us/articles/360052322954-Environment-Variables-Specific-to-Gatsby-Cloud.

## Description
Since we don't have control over these conditions in the plugin, we are not going to load the plugin when `GATSBY_CLOUD === 'true'`.

So, the end state of this is the following:
- In prod (Netlify), `NODE_ENV='production' & GATSBY_CLOUD=undefined` => will load Matomo normally
- In preview deploys, `NODE_ENV='production' & GATSBY_CLOUD='true'` => will NOT load Matomo
- In dev, `NODE_ENV='dev'` => will load Matomo but the plugin handles this case. It will not send any data in dev mode

## Related Issue

#5998

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -9,7 +9,7 @@ const ignoreContent = (process.env.IGNORE_CONTENT || "")
   .split(",")
   .filter(Boolean)
 
-const isPreviewDelpoy = process.env.GATSBY_IS_PREVIEW === "true"
+const isPreviewDeploy = process.env.GATSBY_CLOUD === "true"
 
 const ignoreTranslations = Object.keys(allLanguages)
   .filter((lang) => !supportedLanguages.includes(lang))
@@ -232,7 +232,7 @@ const config = {
 
 // Avoid loading Matomo in preview deploys since NODE_ENV is `production` in
 // there and it will send testing data as production otherwise
-if (!isPreviewDelpoy) {
+if (!isPreviewDeploy) {
   config.plugins = [
     ...config.plugins,
     // Matomo analtyics
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -9,11 +9,13 @@ const ignoreContent = (process.env.IGNORE_CONTENT || "")
   .split(",")
   .filter(Boolean)
 
+const isPreviewDelpoy = process.env.GATSBY_IS_PREVIEW === "true"
+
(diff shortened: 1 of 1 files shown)


wackerow (approved):
LGTM! Was going to suggest we add `GATSBY_IS_PREVIEW` to our `.env.example` but it looks like locally these will still be filtered out by the existing `matomo.js` logic, so I guess it's not really necessary. Looks good!

pettinarip:
From the docs:
```
GATSBY_IS_PREVIEW: true only in the CMS Preview environment, for both [legacy preview builder and incremental preview builder.](https://support.gatsbyjs.com/hc/en-us/articles/360055676874)
```
GATSBY_IS_PREVIEW only works in the CMS previews and not in the pull requests.

I think the only option left here is to use GATSBY_CLOUD variable. Where:
```
GATSBY_CLOUD: Always true. Useful for checking if your build is running on Gatsby Cloud.
```
For this use case it will work as expected.

pettinarip:
The last build was succesfull. https://ethereumorgwebsitedev01-matomopreviewdeploys.gtsb.io/

wackerow (approved):

gatsby-config.js
```diff
@@ -240,3 +229,25 @@ module.exports = {
...
+
+// Avoid loading Matomo in Gatsby Cloud envs since NODE_ENV is `production` in
+// there and it will send testing data as production otherwise
+if (!isGatsbyCloud) {
+  config.plugins = [
+    ...config.plugins,
+    // Matomo analtyics
+    {
+      resolve: "gatsby-plugin-matomo",
+      options: {
```
samajammin, line 241: The Matomo plugin also has a `dev` option:
https://github.com/kremalicious/gatsby-plugin-matomo#options

Should we set that to `true` when in dev env @pettinarip?

samajammin, line 241: The Matomo plugin also has a `dev` option:
https://github.com/kremalicious/gatsby-plugin-matomo#options

Should we set that to `true` when in dev env @pettinarip?

Issue #6386, opened 2022-05-18
Question mark wrapped alone in French version


SuperDelphi:
**Describe the bug**

On the [Ethereum upgrades](https://ethereum.org/fr/upgrades/) page (French version), in the first dark section (see screenshots), the question mark is left alone in a new line, in some screen resolutions.

**To Reproduce**

1. Go to [here](https://ethereum.org/fr/upgrades/)
2. Scroll down a bit to the dark section (see screenshots)

**Expected behavior**

The question mark should not be wrapped alone.

**Screenshots**

Currently this (please note where the question mark stands)
![Capture d’écran 2022-05-18 à 10 32 27](https://user-images.githubusercontent.com/44942598/168994862-cffc34ee-522b-40ab-a463-963c757a222f.png)

Desired result
![Capture d’écran 2022-05-18 à 10 34 31](https://user-images.githubusercontent.com/44942598/168995308-630d3d61-9791-4f7f-a03e-6cabd89eafb5.png)

**Desktop (please complete the following information):**

All

**Smartphone (please complete the following information):**

All

**Additional context**

We need to replace the space before the question mark with an **unbreakable thin space**.

**Want to contribute?**

Yes, sure!

SuperDelphi:
Hey, I would have wanted to fix it actually aha! (waiting for approval)
But if it's your first PR on ethereum.org, I leave it to you!

minimalsm:
Hey @SuperDelphi, thanks for raising this. All yours!

minimalsm:
Thanks for offering @strifedefi, I'll be putting up some other good first issues soon :)

strifedefi:
@SuperDelphi Apologies did not mean to step on your toes! 😃 @minimalsm Thank you I'll be on the lookout!

SuperDelphi:
> @SuperDelphi Apologies did not mean to step on your toes! 😃 @minimalsm Thank you I'll be on the lookout!

No problem at all! Thanks for offering, I appreciate that

HursitTarcan:
> > @SuperDelphi Apologies did not mean to step on your toes! 😃 @minimalsm Thank you I'll be on the lookout!
> 
> No problem at all! Thanks for offering, I appreciate that

Love this civilized chat

SuperDelphi:
@minimalsm By the way, I have yet this issue's PR to be reviewed ˆˆ
https://github.com/ethereum/ethereum-org-website/pull/6395
So we can just close this issue aha! Thanks

PR #6395, opened 2022-05-19
Fix question marks on upgrades page (French)

SuperDelphi:
Replaced normal spaces with unbreakable thin spaces ( ) before question marks (?), according to the rules of the French language. Prevents question marks to be wrapped alone on a new line.

## Description

In French, you should put unbreakable thin spaces before question marks, notably because without them, the question marks could be wrapped alone without the previous word.

## Related Issue

Originally, the related issue pointed out a certain question mark, but as I browsed the whole page, I encountered other occurrences of the same issue. Then I realized that every question mark on the page didn't follow the rule. I fixed them.

https://github.com/ethereum/ethereum-org-website/issues/6386


corwintines (approved):
Interesting. TIL. Thanks @SuperDelphi!

minimalsm (approved):
Sorry for the delay @SuperDelphi, I wanted to make sure I had the time to check these were updated in Crowdin before I merged these changes. Just updated them all and added a note to do a more broad sweep across the codebase at a later date. And one again, thanks and nice catch!

minimalsm:
@all-contributors please add @SuperDelphi for translation, bugs, docs

PR #6403, opened 2022-05-20
update LUNA -> WBTC in bridges example

jmcook1186:
## Description

The current "Bridges" page uses ETH -> Luna as its asset bridging example. this no longer seems appropriate, so I have updated it to BTC/WBTC.


minimalsm (approved):
Nice catch ❤️

Issue #5087, opened 2022-01-13
Better support for RTL languages


minimalsm:
**Is your feature request related to a problem? Please describe.**

Until now we haven't had significant amounts of RTL ([right-to-left-script](https://en.wikipedia.org/wiki/Right-to-left_script)) content on ethereum.org. With #5080 we translate our [Use Ethereum bucket](https://ethereum.org/en/contributing/translation-program/content-versions/#use-ethereum-pages) for Arabic and we'll soon have this bucket translated for Farsi (Persian). 

**Describe the solution you'd like**

For a better UX for those reading a RTL language we should make refactors to better support RTL.

**Additional context**

[Right-to-left script](https://en.wikipedia.org/wiki/Right-to-left_script)

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

SNikhill:
@minimalsm 
Sir, is there any way in which I could be of assistance with this?

samajammin:
Thanks for volunteering @SNikhill!

I suspect the best next step is to outline specific changes we could make & align on approaches to implement those changes. @SNikhill feel free to make suggestions!

minimalsm:
Hey @SNikhill, thanks for offering to help 😀

This is a pretty significant refactor. I think the highest value would be to follow the order of our [content-buckets](https://ethereum.org/en/contributing/translation-program/content-versions/). A very impactful first step would be having proper RTL support for our homepage (bucket one) As far as implementation goes we already [have some RTL support](https://github.com/ethereum/ethereum-org-website/search?q=rtl) but unsure if the approach we have taken is optimal, feel free to suggest alternative approaches if you come up with any—we're definitely open to ideas!

SNikhill:
Okay, sure, sir. I shall see what a content bucket is and start with that.
Yes, I remember seeing that `isLangRightToLeft()` function. I shall let you know as I explore further and have any questions. :)

SNikhill:
Things I noticed:
(NB: I shall be updating this list)

Language: Arabic (I don't understand it but, do know that it is RTL)
- [ ] If I am on the home page and select a code example, the same tends to be rendered in an incorrect way
![image](https://user-images.githubusercontent.com/51415616/150630738-e7674054-b8ab-40c6-b6c5-9e8bf867d173.png)

SNikhill:
@minimalsm 

Sir, above are some things from the Home Page. Not all of them are a priority.

[...]

PR #6415, opened 2022-05-21
Support RTL (Initial)

SNikhill:
## Description

This merge will fix the home page to better support RTL languages

The support for the RTL graph is not a part of this PR as configuring Recharts is more cumbersome than thought.

- I have used the Arabic Language to test out the same.

## Related Issue

- #5087


SNikhill:
I am using "auto" as the direction when it is RTL language as the same would help in case the content is a mixture of both LTR and RTL language.

For Example, in the footer, some links are in English and some are in Arabic.

SNikhill:
(The issue with code-examples is already taken care of by some other change)

SNikhill:
A conflict. `Link.js` was shifted to `Link.tsx`. Looks like the project is being shifted to TypeScript.

minimalsm (approved):
Sorry for the delay and thanks @SNikhill

Issue #6452, opened 2022-05-24
"Trustless" definition for pooled staking options


wackerow:
**Describe the issue**
Recently our staking content was updated to include more details about different ways to stake, and the products/services that can be used to help in each of those categories. In attempt to prevent/limit subjective bias, some objective attributes were assessed for these products/services.

One of these attributes was a "**Trustless**" indicator for pooled staking options, and some feedback has suggested it its worth discussing publicly how this label is defined and how it applies to pooled staking services. This issue can be used as a place for discussion around how this term should be defined to best help inform users, and also for presenting arguments for/against the currently listed services being labeled as "Trustless" or not.

Currently this label is defined on the site as:

> Service does not require trusting any humans to custody your keys or distribute rewards

The goal of this label was to indicate to users which staking pools rely on trusting someone to either hold onto your keys, or to manually distribute rewards. 

This can be confounded by the fact that _currently_ any staked ETH is locked until Shanghai, and although a validator can exit, these funds can still not be retrieved.

- Does this mean we should consider all of these services trusted until withdrawals are enabled by the protocol?
- What mechanisms exist today for non-node operating pooled stakers to force exit a validator balance?
- Which services use these mechanisms?
- How should this change over time?
- Does holding keys to a liquidity token affect the application of this label in any way? How do we differentiate the power a user has by holding keys to an ERC-20 liquidity token vs signing/withdrawal keys for a validator?

---

I would love to hear input from others regarding these questions, thoughts on this definition in general, and also any input (especially from team members of these projects) regarding how specific products would fit in here. If supporting evidence can be provided that would also help in the discussion and help guide where we go from here.

Jstar101:
I agree that it is important to define each indicator as specifically as possible in order to remove any uncertainty around definitions which could leave room for subjective bias. I also agree with the high-level definition of trustless, meaning key steps are taken to remove human/centralised control over pool funds. In my mind, there are 4 key parts to consider:
1) The custody of the keys
2) The distribution of rewards
3) The ability for non-node operating pooled stakers to ensure funds are accessible when requested
4) The ability for a protocol to be changed/smart contracts upgraded by a centralised/controlled

The first two parts are already considered within the current definition and this should continue to remain the case.

1) Custody - tokenisation of deposited ETH alongside the ability for withdrawal credentials to be set by an upgradeable smart contract. 
2) Distribution of rewards - a network of decentralised oracles which determine the number of rewards to be minted and automatically distributed at each time period.

[...]

PR #6461, opened 2022-05-25
Update StakeWise info

### LINKED ISSUE #6460: Update StakeWise information
StakeWise is listed as one of the pool staking options at https://ethereum.org/en/staking/pools/

We would appreciate ticking the Trustless and Diverse Clients boxes for StakeWise. Here is why:

# Trustless

Non-custodial - StakeWise migrated to `0x01` withdrawal prefix in April 2021. All our validators are created with `0x0100000000000000000000002296e122c1a20fca3cac3371357bdad3be0df079` withdrawal credentials since then. The withdrawals will be received by the [Pool Escrow](https://etherscan.io/address/0x2296e122c1a20fca3cac3371357bdad3be0df079) contract and users will burn their sETH2, rETH2 tokens and withdraw ETH from it.

Rewards automatically updated - A set of decentralized oracles update the rewards on a daily basis. The oracle code is available at https://github.com/stakewise/oracle/blob/master/README.md

DAO committee controlling exit signatures - The StakeWise DAO committee controls the exit signatures for our node operators meaning the DAO can force exit validators if required (i.e. pool funds cannot be held hostage by node operators) https://docs.stakewise.io/node-operator/generate-deposit-data#4.-send-encrypted-key-shares

Smart contracts upgradeability - Smart contracts can only be upgraded following a successful DAO vote, underpinning the security/trustlessness of all the above points.

# Diverse Clients
https://www.rated.network/ gives an overview of the current client diversity and new node operators joining the network continue to use a diversified set of clients as detailed in the node operator onboarding posts:
- https://forum.stakewise.io/t/swip-10-onboard-cryptomanufaktur-as-node-operator-in-stakewise-with-250-keys-on-the-ethereum-network/684
- https://forum.stakewise.io/t/verihash-allocate-250-keys-on-the-ethereum-network/713
- https://forum.stakewise.io/t/stakewise-labs-allocate-250-keys-on-the-ethereum-network/719
- https://forum.stakewise.io/t/finoa-consensus-services-allocate-250-keys-on-the-ethereum-network/763

It is also worth flagging that as part of the deployment package for StakeWise node operators, operators are able to transition between clients in real-time should there ever be a need to change clients in an emergency. This is something that node operators are required to prove they can do in the final round of stress testing on our testnet before they are allowed onto the mainnet. The list of EL and CL clients they can migrate between can be found at https://github.com/stakewise/helm-charts/tree/main/charts

tsudmi:
Fixes #6460

### DIFF
--- a/src/data/staking-products.json
+++ b/src/data/staking-products.json
@@ -546,9 +546,9 @@
       ],
       "isFoss": true,
       "hasBugBounty": false,
-      "isTrustless": false,
+      "isTrustless": true,
       "hasPermissionlessNodes": false,
-      "pctMajorityClient": null,
+      "pctMajorityClient": 43.1,
       "platforms": ["Browser"],
       "ui": ["GUI"],
       "socials": {
--- a/src/data/staking-products.json
+++ b/src/data/staking-products.json
@@ -582,7 +582,7 @@
       "isTrustless": true,
       "hasPermissionlessNodes": false,
       "pctMajorityClient": 0.0,
-      "platforms": ["Browser"],
+      "platforms": ["Browser", "Wallet"],
       "ui": ["GUI"],
       "twitter": "https://twitter.com/stakefish",
       "telegram": "https://t.me/stakefish",


src/data/staking-products.json
```diff
@@ -546,9 +546,9 @@
       ],
       "isFoss": true,
       "hasBugBounty": false,
-      "isTrustless": false,
+      "isTrustless": true,
       "hasPermissionlessNodes": false,
-      "pctMajorityClient": null,
+      "pctMajorityClient": 43.1,
```
minimalsm, line 551: Meta comment (unrelated to this PR) @wackerow can we keep this with even though we don't have a super majority? 

In the future would it need to evolve to cover > 1/3 clients?

minimalsm (approved):
Hey @tsudmi, thanks for the PR :-) LGTM.

@wackerow you mind having a look before I merge?

src/data/staking-products.json
wackerow: This is already a flag only for "majority" not "supermajority" implying the client would only need to have >50% of the network (still a risk).

We can think about a better way to rank these now that client diversity is improving. We may want to consider simply hiding this indicator if the network stats reach a certain quality threshold, where emphasis on client diversity isn't as critical. The goal here was to try and flag projects that were more strongly contributing to the supermajority problem, but this is already less critical.

wackerow: Perhaps it's worth opening a discussion on this as well, similar to #6452 for the "Trustless" tag

wackerow (approved):

PR #6479, opened 2022-05-26
TS - templates migration

pettinarip:
## Description

Migration to TS of the templates.

## Related Issue

#6392


src/templates/docs.tsx
```diff
@@ -165,15 +165,22 @@ const Contributors = styled(FileContributors)`
...
+  data: { siteData, pageData: mdx },
+  pageContext,
+}: PageProps<Queries.DocsPageQuery, Context>) => {
   const { isZenMode } = useContext(ZenModeContext)
-  const mdx = data.pageData
-  const isRightToLeft = isLangRightToLeft(mdx.frontmatter.lang)
 
-  const tocItems = mdx.tableOfContents.items
+  if (!siteData || !mdx?.frontmatter) {
+    throw new Error("Docs page template query does not return expected values")
```
samajammin, line 175: 👍

src/templates/staking.tsx
```diff
@@ -338,15 +340,19 @@ const components = {
   StakingGuides,
 }
 
-const StakingPage = ({ data, pageContext, location }) => {
-  const mdx = data.pageData
-  const isRightToLeft = isLangRightToLeft(mdx.frontmatter.lang)
-  const tocItems = mdx.tableOfContents.items
-  const { summaryPoints } = mdx.frontmatter
+const StakingPage = ({
+  data: { pageData: mdx },
```
samajammin, line 344: Same here - should we just rename `pageData` to mdx in the query? 😄

src/templates/static.tsx
```diff
@@ -143,16 +147,28 @@ const components = {
...
-    : mdx.parent.mtime
+  if (!siteData || !mdx?.frontmatter || !mdx.parent) {
+    throw new Error(
+      "Static page template query does not return expected values"
+    )
+  }
+
+  const isRightToLeft = isLangRightToLeft(mdx.frontmatter.lang as Lang)
+
+  // FIXME: remove this any, currently not sure how to fix the ts error
```
samajammin, line 164: Hmmm what was the TS error? Is the type inconsistent?

Not sure if the comment below this is related:
```
// TODO some `gitLogLatestDate` are `null` - why?
```

src/templates/tutorial.tsx
```diff
@@ -142,41 +145,49 @@ const Contributors = styled(FileContributors)`
   border-radius: 4px;
 `
 
-const TutorialPage = ({ data, pageContext }) => {
-  const pageData = data.pageData
-  const isRightToLeft = isLangRightToLeft(pageData.frontmatter.lang)
+const TutorialPage = ({
+  data: { siteData, pageData: mdx },
```
samajammin, line 149: Same here 😄

samajammin (approved):
Nice! LGTM. Made a few comments but those are unrelated to your PR.

src/templates/docs.tsx
```diff
@@ -165,15 +165,22 @@ const Contributors = styled(FileContributors)`
   }
 `
 
-const DocsPage = ({ data, pageContext }) => {
+const DocsPage = ({
+  data: { siteData, pageData: mdx },
```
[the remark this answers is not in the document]
wackerow: Yeah, would make sense. Looks like we immediately rename is in all cases except for the Tutorials template, which is now converted to `mdx` in this PR.

wackerow (approved):

src/templates/docs.tsx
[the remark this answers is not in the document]
pettinarip: Agree. This was a try to make them consistent but doing them at the query level is better.

src/templates/static.tsx
pettinarip: So, the auto-generated types for `mdx` are kind of weird. I have an error with in the next line where we call `parent.fields`.

The generated type for `mdx` is saying that it could possibly be `{}`, not sure if that is correct.

The error:
```
Property 'fields' does not exist on type '{} | { readonly mtime: string; readonly fields: { readonly gitLogLatestDate: string | null; } | null; }'.
  Property 'fields' does not exist on type '{}'.
```

PR #6486, opened 2022-05-27
TS migration - /pages-conditional

pettinarip:
## Description

- Migrated `/pages-conditional` files.
- Refactored how we pull our wallets images from graphql

## Related Issue

#6392


samajammin:
@pettinarip could you explain a bit about the new /schema/ directory? 

Is this where we should now define the schema of any /src/data/ file? What are `sdls` vs. `builders`? Should we have documentation around this moving forward?

src/components/ButtonLink.tsx
```diff
@@ -107,34 +107,17 @@ const ButtonLink: React.FC<IProps> = ({
...
-      <SecondaryScrollLink
-        onClick={handleOnClick}
-        className={className}
-        {...props}
-      >
-        {children}
-      </SecondaryScrollLink>
-    )
-  }
+  const Link = isSecondary ? SecondaryLink : PrimaryLink
```
samajammin, line 110: 👍

src/components/Link.tsx
```diff
@@ -66,6 +66,7 @@ const GlossaryIcon = styled(Icon)`
 export interface IProps {
   to?: string
   href?: string
+  title?: string
```
samajammin, line 69: Why add `title` attribute out of curiosity?

src/components/WalletCompare.js
```diff
@@ -187,6 +187,9 @@ const WalletCompare = ({ location }) => {
           has_high_volume_purchases
           has_dex_integrations
           has_multisig
+          image {
+            ...walletCardImage
```
samajammin, line 191: Wait what!? We can now use dynamic image references?? 🤩

src/schema/builders/WalletsCsv.ts
```diff
@@ -0,0 +1,35 @@
...
+export const WalletsCsv = {
+  name: "WalletsCsv",
+  interfaces: ["Node"],
+  extensions: {
+    infer: true,
+  },
+  fields: {
+    image: {
+      type: "File",
+      resolve: async (source, _args, context) => {
```
samajammin, line 14: Rad 🤩

samajammin: I think we'll just need to make clear that we need to save wallet images as `${source.id}.png`.

wackerow: 🤯

wackerow (approved):
Was this originally targeting the branch from #6479? I see the base automatically switched over to dev from somewhere else, so no preview deploy.

#6479 is now merged into dev, and I updated this branch with latest from dev. Locally this is building great for me. All the wallet images are working correctly, bounty board looks good, no errors when spot checking other pages.

samajammin:
> Was this originally targeting the branch from https://github.com/ethereum/ethereum-org-website/pull/6479? I see the base automatically switched over to dev from somewhere else

Yes, nice GH feature 😄

src/schema/builders/WalletsCsv.ts
pettinarip: Yea this is pretty cool from Gatsby where you can customize the schema. More on this here: https://www.gatsbyjs.com/docs/reference/graphql-data-layer/schema-customization/#foreign-key-fields

> I think we'll just need to make clear that we need to save wallet images as `${source.id}.png`.

Agree. Where do you think we should add more doc? in this file or in a new doc file under `/docs`?

pettinarip:
> @pettinarip could you explain a bit about the new /schema/ directory?

There was a lack of explanation in my PR. My bad.

I tried to organize a bit how we are defining our schema in the `createSchemaCustomization` hook. I moved all of what was there to the new `/schema` folder.

There are two ways to define a schema in Gatsby (I believe):
- GraphQL SDL, the traditional way using template literals: https://graphql.org/learn/schema/
- And a more "programmatic" approach, which has more flexibility, Gatsby Type Builders https://www.gatsbyjs.com/docs/reference/graphql-data-layer/schema-customization/#gatsby-type-builders with `schema.buildObjectType`

That is why you see two folders in `/schema`
- `sdls`
- `builders`

pettinarip:
> Was this originally targeting the branch from #6479? I see the base automatically switched over to dev from somewhere else, so no preview deploy*.

Thanks @wackerow. Yes, it was targeting that branch.

src/components/WalletCompare.js
pettinarip: Yes! pretty cool. Would like to slowly transition to this on the rest of the queries.

src/components/Link.tsx
pettinarip: Probably this was not the correct solution.

The problem was that for example, in the `eth.tsx` page, we have this:
```tsx
<ButtonLink to="/get-eth/" title="where to buy eth">
```

That ButtonLink is going to be a Link underneath.

Help me with the thinking here, in this case, ButtonLink shouldn't send a `title` prop, right? because it is going to be a Link. The `title` only make sense in a `button` tag.

samajammin: Ah, I see. Ya given that `ButtonLink` isn't actually a button, no tooltip is generated, so I don't think it needs a `title` prop. I'd say let's remove that from those instances.

src/schema/builders/WalletsCsv.ts
samajammin: Perhaps a new file under `/docs` makes sense around schema? We could add this note as well as the layout structure you described here: https://github.com/ethereum/ethereum-org-website/pull/6486#issuecomment-1146411041

samajammin: Otherwise a md file in the directory https://github.com/ethereum/ethereum-org-website/tree/ts-pages-conditional/src/schema may make sense so if people navigate to it e.g. via GH.

samajammin: I suppose I'd favor in `/docs` to keep those doc resources in one place.

src/pages-conditional/dapps.tsx
```diff
@@ -314,28 +317,56 @@ const MoreButtonContainer = styled.div`
...
-const GAMING = "gaming"
+enum CatergoryType {
+  FINANCE = "finance",
+  TECHNOLOGY = "technology",
+  COLLECTIBLES = "collectibles",
+  GAMING = "gaming",
+}
 
-const DappsPage = ({ data, location }) => {
+interface Catergory {
```
samajammin, line 327:
```suggestion
interface Category {
```
Small typo 😉

src/pages-conditional/dapps.tsx
```diff
@@ -314,28 +317,56 @@ const MoreButtonContainer = styled.div`
...
+  benefitsTitle?: string
+  benefitsDescription?: string
+  benefits?: Array<{
+    emoji: string
+    title: string
+    description: string
+  }>
+}
+
+interface Catergories {
```
samajammin, line 339:
```suggestion
interface Categories {
```

src/pages-conditional/dapps.tsx
```diff
@@ -314,28 +317,56 @@ const MoreButtonContainer = styled.div`
   margin-bottom: 1rem;
 `
 
-const FINANCE = "finance"
-const TECHNOLOGY = "technology"
-const COLLECTIBLES = "collectibles"
-const GAMING = "gaming"
+enum CatergoryType {
```
samajammin, line 320:
```suggestion
enum CategoryType {
```

gatsby-node.ts
```diff
@@ -338,11 +340,7 @@ export const createPages: GatsbyNode<any, Context>["createPages"] = async ({
         )
         createPage({
           path: `/${lang}/${page}/`,
-          component: path.resolve(
-            page === "wallets"
-              ? `src/pages-conditional/${page}/index.js`
-              : `src/pages-conditional/${page}.js`
-          ),
+          component: path.resolve(`src/pages-conditional/${page}.tsx`),
```
samajammin, line 343: Oh nice, good call renaming the file 😄

samajammin: Can't remember why I put it in a directory 😅

gatsby-node.ts
```diff
@@ -394,54 +392,14 @@ export const onCreatePage: GatsbyNode<any, Context>["onCreatePage"] = async ({
...
-    createTypes(typeDefs)
+    const { sdls, builders } = Schema
 
-    // Optimization. Ref: https://www.gatsbyjs.com/docs/scaling-issues/#switch-off-type-inference-for-sitepagecontext
-    createTypes(`
-    type SitePage implements Node @dontInfer {
-      path: String!
-    }
-  `)
+    createTypes([
```
samajammin, line 399: Love this reorganization overall 👍

samajammin (changes requested):
Awesome work! Looks great overall, just a few comments, mainly to:
- fix typos
- add a `/docs` page to explain the new schema dirctory

src/schema/builders/WalletsCsv.ts
pettinarip: Yes, to keep things consistent and easy to find I will go with the `/docs`.

gatsby-node.ts
pettinarip: :) well...I like the idea of having folders and having the following structure
```
/page
  /index.tsx // only jsx
  /queries.ts
  /styles.ts
```
to have cleaner and smaller files. Would like to propose this approach in the new Learn page.

PR #6487, opened 2022-05-27
Refactor Card component to TS

minimalsm:
## Description
- Migrates Card component to Typescript
- I was going to migrate all Cards that extend it but it looks like all of our cards are standalone components 🤔

## Related Issue
#6392

### DIFF
--- a/src/components/Card.tsx
+++ b/src/components/Card.tsx
@@ -17,7 +17,20 @@ const Description = styled.p`
 
 const TopContent = styled.div``
 
-const Card = ({ emoji, title, description, children, className }) => (
+interface IProps {
+  emoji?: string
+  title: string
+  description: string
+  className?: string
+}
+
+const Card: React.FC<IProps> = ({
+  emoji,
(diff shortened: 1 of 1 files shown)


src/components/Card.tsx
```diff
@@ -17,7 +17,20 @@ const Description = styled.p`
 
 const TopContent = styled.div``
 
-const Card = ({ emoji, title, description, children, className }) => (
+interface IProps {
```
pettinarip, line 20:
```suggestion
export interface IProps {
```
Just as a convention, to keep all components props interfaces exposed.

pettinarip (approved):
Awesome! 💪🏼 added a small comment there

pettinarip:
> * I was going to migrate all Cards that extend it but it looks like all of our cards are standalone components thinking

Uhmm yes, something to improve perhaps 👍🏼 to compose them better and reuse code.

GustavoRSSilva (approved):

Issue #4916, opened 2021-12-29
Added term to glossary: Soft Fork


nschaferx:
**Checklist**

- [x] This is a term not already found in the glossary (for similar terms, please consider the benefits of a new term vs updating an existing term)
- [x] This term/definition is not a product advertisement or contain other promotional content
- [x] This term/definition is directly relevant to Ethereum

**Soft Fork**

**A temporary divergence in a [blockchain](https://ethereum.org/en/glossary/#blockchain) that occurs when the [consensus rules](https://ethereum.org/en/glossary/#consensus-rules) become stricter. Contrary to a [hard fork](https://ethereum.org/en/glossary/#hard-fork), a soft fork is backward compatible; upgraded nodes can validate blocks created by nonupgraded nodes as long as they follow the new consensus rules. However, nonupgraded nodes are ultimately encouraged to download the latest version of the blockchain's software to avoid complications related to transactions.**

**N/A**

**N/A**

**N/A**

minimalsm:
Hey @nschaferx, I agree we should add Soft Fork to the glossary. 

I think we can come up with a cleaner definition. 

Would you like to be assigned to this?

nschaferx:
Sure thing!

minimalsm:
@nschaferx assigned! Be sure to ping here if you need any help getting this over the line.

nschaferx:
I believe omitting the last sentence and changing "become stricter" to "are updated" would simplify the definition.

minimalsm:
Agree! But I wouldn't worry too much about the definition being perfect right now. We can resolve that in the PR itself :-)

jumptiger8:
> 同意！但我不会太担心现在的定义是否完美。我们可以在 PR 本身解决这个问题:-)

Can I participate in the contribution of this word?

jumptiger8:
I see that this word has not been updated for a long time and has not appeared in Ethereum vocabulary.

nschaferx:
Yes you may, I was told by the lead to let contributors edit.

corwintines:
Closing as the above PR addresses this

PR #6492, opened 2022-05-28
Added term to glossary: Soft Fork

pushkar2112:
Changes corresponding to issue #4916

## Description
Add the term Soft Fork to the Ethereum Glossary, which wasn't present originally.

## Related Issue

#4916

### DIFF
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -873,6 +873,10 @@ Short for "succinct non-interactive argument of knowledge", a SNARK is a type of
   Zero-knowledge rollups
 </DocLink>
 
+### Soft Fork {#soft-fork}
+
+A temporary divergence in a [blockchain](#blockchain) that occurs when the [consensus rules](#consensus-rules) become stricter. Contrary to a [hard fork](#hard-fork), a soft fork is backward compatible; upgraded nodes can validate blocks created by nonupgraded nodes as long as they follow the new consensus rules. However, nonupgraded nodes are ultimately encouraged to download the latest version of the blockchain's software to avoid complications related to transactions.
+
 ### Solidity {#solidity}
 
 A procedural (imperative) programming language with syntax that is similar to JavaScript, C++, or Java. The most popular and most frequently used language for Ethereum [smart contracts](#smart-contract). Created by Dr. Gavin Wood.
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -875,7 +875,7 @@ Short for "succinct non-interactive argument of knowledge", a SNARK is a type of
 
 ### soft fork {#soft-fork}
 
-A temporary divergence in a [blockchain](#blockchain) that occurs when the [consensus rules](#consensus-rules) become stricter. Contrary to a [hard fork](#hard-fork), a soft fork is backwards compatible; upgraded nodes can validate blocks created by non-upgraded nodes as long as they follow the new consensus rules.
+A divergence in a [blockchain](#blockchain) that occurs when the [consensus rules](#consensus-rules) become change. Contrary to a [hard fork](#hard-fork), a soft fork is backwards compatible; upgraded nodes can validate blocks created by non-upgraded nodes as long as they follow the new consensus rules.
 
 ### Solidity {#solidity}
 
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -875,7 +875,7 @@ Short for "succinct non-interactive argument of knowledge", a SNARK is a type of
 
 ### soft fork {#soft-fork}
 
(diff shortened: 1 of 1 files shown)


src/content/glossary/index.md
```diff
@@ -873,6 +873,10 @@ Short for "succinct non-interactive argument of knowledge", a SNARK is a type of
   Zero-knowledge rollups
 </DocLink>
 
+### Soft Fork {#soft-fork}
+
+A temporary divergence in a [blockchain](#blockchain) that occurs when the [consensus rules](#consensus-rules) become stricter. Contrary to a [hard fork](#hard-fork), a soft fork is backward compatible; upgraded nodes can validate blocks created by nonupgraded nodes as long as they follow the new consensus rules. However, nonupgraded nodes are ultimately encouraged to download the latest version of the blockchain's software to avoid complications related to transactions.
```
minimalsm, line 878:
```suggestion
A temporary divergence in a [blockchain](#blockchain) that occurs when the [consensus rules](#consensus-rules) become stricter. Contrary to a [hard fork](#hard-fork), a soft fork is backwards compatible; upgraded nodes can validate blocks created by nonupgraded nodes as long as they follow the new consensus rules. However, nonupgraded nodes are ultimately encouraged to download the latest version of the blockchain's software to avoid complications related to transactions.
```

src/content/glossary/index.md
```diff
@@ -873,6 +873,10 @@ Short for "succinct non-interactive argument of knowledge", a SNARK is a type of
   Zero-knowledge rollups
 </DocLink>
 
+### Soft Fork {#soft-fork}
```
minimalsm, line 876: Glossary convention is that everything should be lowercase unless it is an acronym or a proper noun.

```suggestion
### soft fork {#soft-fork}
```

src/content/glossary/index.md
```diff
@@ -873,6 +873,10 @@ Short for "succinct non-interactive argument of knowledge", a SNARK is a type of
   Zero-knowledge rollups
 </DocLink>
 
+### soft fork {#soft-fork}
+
+A temporary divergence in a [blockchain](#blockchain) that occurs when the [consensus rules](#consensus-rules) become stricter. Contrary to a [hard fork](#hard-fork), a soft fork is backwards compatible; upgraded nodes can validate blocks created by nonupgraded nodes as long as they follow the new consensus rules. However, nonupgraded nodes are ultimately encouraged to download the latest version of the blockchain's software to avoid complications related to transactions.
```
minimalsm, line 878: > However, nonupgraded nodes are ultimately encouraged to download the latest version of the blockchain's software to avoid complications related to transactions.

I'd say this part strays from the job of a glossary (starting to describe the Ethereum specific details instead of simply defining a soft fork at a high level). I think we should delete the last sentence. The rest LGTM 😎👍

minimalsm, line 878:
```suggestion
A temporary divergence in a [blockchain](#blockchain) that occurs when the [consensus rules](#consensus-rules) become stricter. Contrary to a [hard fork](#hard-fork), a soft fork is backwards compatible; upgraded nodes can validate blocks created by non-upgraded nodes as long as they follow the new consensus rules. However, nonupgraded nodes are ultimately encouraged to download the latest version of the blockchain's software to avoid complications related to transactions.
```

jmcook1186: IMO a soft fork is one that makes non-breaking changes (i.e. nodes that do not implement the fork can still participate in a network of upgraded nodes).

Not keen on "temporary divergence" - temporary suggests it will be reverted eventually, and divergence implies incorrectness or irrelevance (IMO). Also not sure why a soft fork necessarily makes a set of rules "stricter" nor are forks specific to the "consensus rules".

pushkar2112: @jmcook1186 @minimalsm Sure, I agree with u guys! We sure could make changes to this.

This is a definition that I found online,

A soft fork is a change to the software protocol where only previously valid transaction blocks are made invalid. Because old nodes will recognise the new blocks as valid, a soft fork is backwards-compatible. This kind of fork requires only a majority of the miners upgrading to enforce the new rules, as opposed to a hard fork that requires all nodes to upgrade and agree on the new version.

We can modify and build the definition upon this. I think this covers the concerns raised by @jmcook1186. Let me know ur views as u review this.

Cheers!!

minimalsm:
@wackerow could you have a look :-)

src/content/glossary/index.md
jmcook1186: Fair enough - I'd equated "consensus rules" with "consensus mechanism" but a soft fork could touch any part of the client software - the specificity seems misleading to me.

Point taken re "stricter", but what about a soft fork that introduces new features? They can be run by clients that update and not by clients that don't without invalidating either - I wouldn't consider any rules to have become stricter in that case, if anything the feature-set, has expanded, updated clients can "do more" and therefore the rule-set has become less strict.

PR #6494, opened 2022-05-28
Fix a date and add ETHVenice

jhonnyvianello:
## Fix a date and add a new event
Fixed date of HackFS event and added ETHVenice event 🎆
I attach the links here of the official events sites.
HackFS : https://hackfs.ethglobal.com/
ETHVenice : https://ethvenice.com/


minimalsm (approved):
🎉🎉🎉🎉

PR #6498, opened 2022-05-29
Fix punctuation on community get involved page

hursittarcan:
## Description
- Removed punctuation mark from list. 
- Changed `... help people create value through DAO.` to `... help people create value through DAO`.


minimalsm (approved):
Thanks @hursittarcan

PR #6500, opened 2022-05-30
Misc TypeScript Migrations

WilliamWelsh:
I wanted to help with the TypeScript rollout, so I've migrated the following so far:

#### src/data
- src/data/addresses.js
- src/data/eth-upgrade-articles.js

#### src/api
- src/api/etherscanBlock.js
- src/api/etherscan.js
- src/api/l2beat.js
- src/api/roadmap.js
- src/api/translations.js
- src/api/txs.js

### src/lambda
- src/lambda/etherscanBlock.js
- src/lambda/etherscan.js
- src/lambda/l2beat.js
- src/lambda/roadmap.js
- src/lambda/translation-build.js
- src/lambda/translations.js
- src/lambda/txs.js

### src/utils
- src/utils/calculateStakingRewards

## Related Issue

#6392


src/utils/calculateStakingRewards.ts
```diff
@@ -1,4 +1,4 @@
-const calculateStakingRewards = (totalAtStake) => {
+const calculateStakingRewards = (totalAtStake: number) => {
```
minimalsm, line 1: Shouldn't we explicitly declare a return type here?

src/lambda/txs.ts
```diff
@@ -21,12 +23,15 @@ const lambda = async (apiKey) => {
...
+    }
   }
 }
 
 const handler = () => {
   return lambda(process.env.ETHERSCAN_API_KEY)
 }
 
-module.exports = { handler, lambda }
+export { handler, lambda }
```
minimalsm, line 37: Would it be better to use named exports directly on the functions here now that we've dropped CommonJS?

minimalsm (changes requested):
Hey @WilliamWelsh, thanks for the PR 😎

I'm a TypeScript noob so I'll leave it to @pettinarip to double-check this. I've left a few open questions that I'm unsure of (though I might be entirely wrong 🙂)

I can't comment directly on the file but for the `eth-upgrade-articles` I think we should be explicitly declaring types e.g.:

```ts
interface Article {
  title: string,
  description: string,
  link: string,
}
```

src/utils/calculateStakingRewards.ts
WilliamWelsh: Yes, fixed. Thx :)

src/lambda/txs.ts
WilliamWelsh: I agree, I was following this example: https://github.com/ethereum/ethereum-org-website/blob/dev/src/lambda/coinmetrics.ts

src/data/eth-upgrade-articles.ts
```diff
@@ -1,4 +1,11 @@
-export const dannyArticles = [
+interface Article {
+  title: string
+  description: string
+  link: string
+}
+
+/** Articles by Danny Ryan (Ethereum Foundation) */
+export const dannyArticles: Article[] = [
```
pettinarip, line 8: We are using the `Array<Article>` form. 
```suggestion
export const dannyArticles: Array<Article> = [
```

src/data/eth-upgrade-articles.ts
```diff
@@ -26,24 +33,25 @@ export const dannyArticles = [
   },
 ]
 
-export const benArticles = [
+/** Articles by Ben Edgington (PegaSys, ConsenSys) */
+export const benArticles: Article[] = [
```
pettinarip, line 37:
```suggestion
export const benArticles: Array<Article> = [
```

src/lambda/etherscan.ts
```diff
@@ -1,6 +1,8 @@
-const axios = require("axios")
+import axios from "axios"
 
-const lambda = async (apiKey) => {
+import type { HandlerResponse } from "@netlify/functions"
+
+const lambda = async (apiKey: string | undefined): Promise<HandlerResponse> => {
```
pettinarip, line 5: `apiKey` should not be undefined. 
```suggestion
const lambda = async (apiKey: string): Promise<HandlerResponse> => {
```

And then in the handler function please do something like:
```ts
const handler = () => {
  let apiKey = process.env.ETHERSCAN_API_KEY

  if(!apiKey) {
    throw new Error('required env ETHERSCAN_API_KEY not set')
  }

  return lambda(apiKey)
}
```

src/lambda/txs.ts
pettinarip: Uhmm I missed that... @WilliamWelsh if you are down, I encourage you to switch them to named exports. Would be nice to have 👍🏼

pettinarip (changes requested):
This is looking amazing @WilliamWelsh thanks for tackling so many files. Left a couple of comments there. LMK if you have any question.

src/data/eth-upgrade-articles.ts
```diff
@@ -26,24 +33,25 @@ export const dannyArticles = [
...
+    title: "What’s New in Eth2 – #90",
     description: "25 March 2022",
     link: "https://hackmd.io/@benjaminion/eth2_news/https%3A%2F%2Fhackmd.io%2F%40benjaminion%2Fwnie2_220325",
   },
   {
-    title: "What’s New in Eth2 – #88",
+    title: "What’s New in Eth2 – #89",
     description: "11 March 2022",
     link: "https://hackmd.io/@benjaminion/eth2_news/https%3A%2F%2Fhackmd.io%2F%40benjaminion%2Fwnie2_220311",
   },
```
pettinarip, line 57: Also, pls, rename this file to `ethUpgradeArticles` as we are trying to have consistent file names.

WilliamWelsh: 👍

src/lambda/etherscan.ts
WilliamWelsh: I also went ahead and added `environment.d.ts` (https://github.com/ethereum/ethereum-org-website/pull/6500/commits/571c2e5e3fab687c1b803b33f9cf0fad6cca5272)

src/data/eth-upgrade-articles.ts
WilliamWelsh: 👍

src/data/eth-upgrade-articles.ts
WilliamWelsh: 👍

environment.d.ts
```diff
@@ -0,0 +1,16 @@
...
+      GATSBY_ALGOLIA_APP_ID: string
+      GATSBY_ALGOLIA_SEARCH_KEY: string
+      GATSBY_GITHUB_TOKEN_READ_ONLY: string
+      GATSBY_FUNCTIONS_PATH: string
+      CROWDIN_API_KEY: string
+    }
+  }
+}
+
+export {}
```
pettinarip, line 16: Thanks for doing that, good idea.

We already have a `src/global.d.ts`. Would you mind in putting this in there?

src/lambda/translations.ts
```diff
@@ -0,0 +1,37 @@
+import axios from "axios"
+
+import type { HandlerResponse } from "@netlify/functions"
+
+const lambda = async (apiKey: string): Promise<HandlerResponse> => {
```
pettinarip, line 5:
```suggestion
export const lambda = async (apiKey: string): Promise<HandlerResponse> => {
```

src/lambda/txs.ts
```diff
@@ -1,6 +1,8 @@
-const axios = require("axios")
+import axios from "axios"
 
-const lambda = async (apiKey) => {
+import type { HandlerResponse } from "@netlify/functions"
+
+const lambda = async (apiKey: string): Promise<HandlerResponse> => {
```
pettinarip, line 5:
```suggestion
export const lambda = async (apiKey: string): Promise<HandlerResponse> => {
```

pettinarip (changes requested):
@WilliamWelsh a few small comments, the rest is looking great.

src/lambda/translations.ts
WilliamWelsh: Oops 😅

pettinarip:
@all-contributors please add @WilliamWelsh for code

PR #6512, opened 2022-05-31
Migrated ActionCard.js to TS

hursittarcan:
## Description
- Migrated `ActionCard` component to TypeScript. 

## Related Issue
- Related to [#6392]

### DIFF
--- a/src/components/ActionCard.tsx
+++ b/src/components/ActionCard.tsx
@@ -18,8 +18,8 @@ const ChildrenContainer = styled.div`
 `
 
 const ImageWrapper = styled.div<{
-    isRight: boolean | undefined
-    isBottom: boolean | undefined
+  isRight: boolean | undefined
+  isBottom: boolean | undefined
 }>`
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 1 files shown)


src/components/ActionCard.tsx
```diff
@@ -62,16 +62,27 @@ const Card = styled(Link)`
...
-  to,
-  alt,
-  image,
-  title,
-  description,
-  children,
-  className,
-  isRight,
-  isBottom = true,
+export interface IProps {
```
pettinarip, line 65: You made everything optional but there are some props that we do want to make them as required.
How about this?
```ts
export interface IProps {
  to: string
  alt?: string
  image: string
  title: string
  description: string
  className?: string
  isRight?: boolean
  isBottom?: boolean
}
```

pettinarip (changes requested):
Awesome @hursittarcan thanks for the PR! 

Please, check the comment I made.

Also, when I ran `yarn type-check` I found an error with the `<ImageWrapper>` component used here. To fix it, you will need to declare the extra props that we are using in this case.

```ts
// on line #20
const ImageWrapper = styled.div<{
  isRight: boolean | undefined
  isBottom: boolean | undefined
}>`
```

After doing that, run again `yarn type-check` to make sure that everything is good.

src/components/ActionCard.tsx
```diff
@@ -0,0 +1,112 @@
...
+                                          to,
+                                          alt,
+                                          image,
+                                          title,
+                                          description,
+                                          children,
+                                          className,
+                                          isRight,
+                                          isBottom = true
+                                      }) => {
```
pettinarip, line 89: Hmm weird how this format is in here.. I inspected the last commit and a lot of spaces were added to most of these lines. Do you know why?

pettinarip (changes requested):
Looking great! I noticed a misalignment in the format. Can you fix that?

The format should be covered by `prettier`. Try to do the next:
1. Make a small format change manually in the file
2. Make a new commit
3. `prettier` should format it automatically when you make a new commit

hursittarcan:
Hey @pettinarip, 
I believe the formatting is fixed now. I think the extra spaces were because of my IDE.

PR #6508, opened 2022-05-30
Add new information to sidechains page

emmanuel-awosika:
Reworked the content to add more structure to the page. Also updated pros/cons and described the architecture of sidechains in detail.

## Description

I felt the page needed some structure, so I reworked it and added a few things to provide readers with more information. Here's a brief overview of the changes:

- Rewrote the introduction (the original was too long imo and could be more concise)
- Included new details about sidechains' relationship with Mainnet and contrasted them with Layer 2 scaling solutions 
- Created a section to describe how sidechains work, with new details on how assets are transferred
- Updated the "Use Sidechains" section; replaced the Gnosis chain link (the other was a dead link) and added Loom Network
- Updated pros and cons; added some points and fleshed out others

PR #6502, opened 2022-05-30
Add new information to Validium page

emmanuel-awosika:
Added a detailed description of the Validium scaling solution and briefly explained the differences between validiums and ZK-rollups. The update also contains an explanation of volitions (scaling solutions that allow users to choose between a validium and a ZK-rollup).

## Description

I edited the page to include a detailed overview of validiums and how they work (especially, how they approach data availability). Then I also explained volitions, which allow users to switch between ZK-rollups and validiums. Other additions include:
- An update to the pros and cons,
- Additions to the "Further Reading" section (included some useful articles)
- Updating the "Use Validium" section: Added links to the actual projects' website and documentation pages. 

PR Preview: https://ethereumorgwebsitedev01-emmanuelawosikaethereumor15817.gtsb.io/en/developers/docs/scaling/validium/

PR #6591, opened 2022-06-07
Remove incomplete frontmatter

samajammin:
## Description

Remove UI banner from sidechains & validium pages:

> This page is incomplete and we'd love your help. Edit this page and add anything that you think might be useful to others.

Thanks to recent PRs from @emmanuel-awosika, we've recently added a good amount of info to both pages:
- https://github.com/ethereum/ethereum-org-website/pull/6508
- https://github.com/ethereum/ethereum-org-website/pull/6502

Doesn't mean we can't expand/edit further in the future, but I think they're to a point where they don't really need that flag anymore.

## Related Issue
- https://github.com/ethereum/ethereum-org-website/pull/6508
- https://github.com/ethereum/ethereum-org-website/pull/6502

### DIFF
--- a/src/content/developers/docs/scaling/sidechains/index.md
+++ b/src/content/developers/docs/scaling/sidechains/index.md
@@ -3,7 +3,6 @@ title: Sidechains
 description: An introduction to sidechains as a scaling solution currently utilized by the Ethereum community.
 lang: en
 sidebar: true
-incomplete: true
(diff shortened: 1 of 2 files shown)


corwintines (approved):
I agree, I think these are in a state that we can remove this. It would be good to expand into different parts of rollups (sequencers, provers, validators, etc), but this is a good start.

PR #6621, opened 2022-06-09
Add page on data availability

emmanuel-awosika:
This page discusses the concept of data availability, why it matters, and solutions to the data availability problem.

## Description

@samajammin suggested a page on data availability, so I decided to start things off with this contribution. The page breaks down the concept of data availability and how the data availability relates to security and scalability. It also has a brief description of different solutions to the data availability problem.

### DIFF
--- a/src/content/developers/docs/data-availability.md
+++ b/src/content/developers/docs/data-availability.md
@@ -5,7 +5,15 @@ lang: en
 sidebar: true
 ---
 
-In Ethereum, a node proposing a new block is required to make the underlying data available to the remaining peer-to-peer nodes on the network. This concept, known as data availability, is fundamental to the security and functionality of every blockchain, including Ethereum. Data availability is so important that an entire field of research studying the so-called “data availability problem” exists. 
+Trustlessness is a major premise of public blockchains ("don't trust, verify"). One of the ways Ethereum reduces trust assumptions is by enforcing rules on data availability: a block producer proposing a new block is required publish the data that went into building the block on the network.  
+
+All nodes on the Ethereum network execute the transactions in blocks they receive from peers to ensure that the changes proposed by a block producer precisely match those computed independently by the node. This is how nodes verify that new information is valid, rather than having to trust that block producers are honest. This is not possible if any data is missing. 
+
+Data availability is important because if we can’t reproduce something with the data we have available, from the blockchain’s perspective, it does not exist. Access to a block's data enables validating nodes to trustlessly replay transactions using their version of Ethereum's world state and independently verify the correctness of each block. 
+
+## Prerequisites {#prerequisites} 
+
+You should have a good understanding of [blockchain fundamentals](/developers/docs/intro-to-ethereum/), especially [consensus mechanisms](/developers/docs/consensus-mechanisms/). This page also assumes the reader is familiar with [blocks](/developers/docs/blocks/), [transactions](/developers/docs/transactions/), [nodes](/developers/docs/nodes-and-clients/), [scaling solutions](/developers/docs/scaling/), and other relevant topics. 
 
 ## What is data availability? {#what-is-data-availability}
 
@@ -16,23 +24,33 @@ Each block has two major parts:
 - The **block header**: This contains general information (metadata) about the block, such as the timestamp, block hash, block number, etc.
 - The  **block body**: This contains the actual transactions processed as part of the block. 
 
-To propose a block, [miners](/developers/docs/consensus-mechanisms/pow/mining/) (or validators in [proof-of-stake](/developers/docs/consensus-mechanisms/pos/)) must publish the entire block (block header and block body). Publishing the entire block lets nodes participating in consensus download the block and confirm that all transactions in the block are valid.
-
-[Full nodes](/developers/docs/nodes-and-clients/#full-node) are different from [light nodes](/developers/docs/nodes-and-clients/#light-node) that only download the block header. Without full nodes verifying transactions, block proposers could get away with inserting malicious transactions in blocks. 
+When proposing new blocks, block producers must publish the entire block, including the transaction data (contained in the block body). Nodes participating in consensus can then download the block's data and re-execute the transactions to confirm their validity. Without nodes verifying transactions, block proposers could get away with inserting malicious transactions in blocks. 
 
 ### The data availability problem {#the-data-availability-problem}
 
 We can encapsulate the data availability problem into the question: "how do we verify that the data for a newly produced block is available?". This data being available is crucial as the security of Ethereum assumes that full nodes have access to block data. 
 
-If a miner proposes a block without all the data being available, it could reach finality whilst containing invalid transactions. Even if the block is valid, the block's data not being fully available to validate has negative implications for users and the functionality of the network.
+If a block producer proposes a block without all the data being available, it could reach finality whilst containing invalid transactions. Even if the block is valid, the block's data not being fully available to validate has negative implications for users and the functionality of the network.
 
-Most early discussions around data availability concerned the visibility of transaction data to full nodes (validators) participating in Ethereum’s consensus. However, recent data availability research is also focused on verifying data availability with light clients. For light clients, the problem of data availability concerns validating a block’s data availability without having to download every entire block. 
+The data availability problem is also relevant when discussing [scaling solutions](/developers/docs/scaling/), such as rollups. These protocols increase throughput by executing transactions off Ethereum Mainnet. However, for them to derive security from Ethereum, they must post transaction data on Mainnet, allowing anyone to verify the correctness of computations performed off the main chain. 
+
+#### Data availability and light clients 
+
+Although the classic notion of data availability concerned the visibility of transaction data to validating nodes, newer research has focused on verifying data availability with light clients. For light clients, the data availability problem concerns validating a block’s availability without having to download the entire block. 
+
+A light client is an Ethereum node that only syncs to the latest block header and requests other information from full nodes. As they don't download blocks, light clients cannot validate transactions or help secure Ethereum. 
+
+However, work is underway to ensure light clients can prove data availability without needing to download blocks. If light clients can verify the availability of a block, they can contribute to Ethereum's security by alerting other nodes to a block's unavailability. 
+
+A related area of research is focused on mechanisms for making data provably available in a stateless Ethereum. The [stateless client concept](https://ethresear.ch/t/the-stateless-client-concept/172) is a proposed version of Ethereum, where validating nodes don't have to store state data before verifying blocks. 
+
+Statelessness is expected to improve the security, scalability, and long-term sustainability of Ethereum. With lower hardware requirements for validating nodes, more validators can join the network and secure it against malicious actors. 
 
 ### Data availability vs. data retrievability {#data-availability-vs-data-retrievability}
 
 Data availability is different from data retrievability. Data availability is the ability of nodes to download transaction data for a block while it is being proposed for addition to the chain. In other words, data availability is relevant when a block is yet to pass consensus.
 
-Data retrievability is the ability of nodes to download transactions from a block *after* it has passed consensus. Essentially, data retrievability is concerned with retrieving data from historical blocks. A block becomes “historical” when other blocks have been built on it. 
+Data retrievability is the ability of nodes to retrieve *historical information* from the blockchain. A blockchain's history is made up of ancient blocks and receipts that store information about past events. While historical blockchain data may be necessary for archiving purposes, nodes can validate the chain and process transactions without it.
 
 The core Ethereum protocol is primarily concerned with data availability, not data retrievability. Ethereum will not store data for every transaction it has processed forever, as doing so increases storage requirements for full nodes, negatively impacting Ethereum's decentralization.
 
@@ -44,22 +62,19 @@ Fortunately, data retrievability is a much easier problem to solve than data ava
 
 ### Blockchain security {#blockchain-security}
 
-Data availability is crucial for blockchain security, or else “data withholding attacks” would become commonplace. A data withholding attack occurs when a miner or validator publishes a block but doesn’t share the transaction data used to build the block. 
+Data availability is crucial for blockchain security, or else “data withholding attacks” would become commonplace. A data withholding attack occurs when a block producer publishes a block but doesn’t share the transaction data used to build the block. 
 
-If a data withholding attack happens, full nodes cannot verify the correctness of updates to Ethereum’s world state. This gives malicious block producers leeway to subvert protocol rules and advance invalid state transitions on the Ethereum network. 
+If a data withholding attack happens, full nodes cannot verify the correctness of updates to Ethereum’s world state. This gives malicious block proposers leeway to subvert protocol rules and advance invalid state transitions on the Ethereum network. 
 
-This is important because other network participants, such as light clients, rely on full nodes to verify the network’s state. Unlike full nodes, light clients only check block headers and don’t download block data. Therefore, the rules around data availability ensure full nodes can validate blocks and prevent the chain from getting corrupted. 
+Visibility of block data to full nodes is critical because other network participants, such as light clients, rely on full nodes to verify the network’s state. Unlike full nodes, light clients only check block headers and don’t download block body. Therefore, the rules around data availability ensure full nodes can validate blocks and prevent the chain from getting corrupted. 
 
 ### Decentralized scalability {#decentralized-scalability}
 
-[Ethereum’s goal is to scale computation without trading off decentralization and security[(/upgrades/vision/). Due to the constraints of the monolithic blockchain architecture, data availability is critical to achieving decentralized scalability.
-
+[Ethereum’s goal is to scale computation without trading off decentralization and security](/upgrades/vision/). Due to the constraints of the monolithic blockchain architecture, data availability is critical to achieving decentralized scalability.
 
 #### Data availability and layer 2 scaling {#data-availability-and-layer-2-scaling}
 
-[Layer 2 scaling solutions](/layer-2/), such as [rollups](/glossary/#rollups), scale network throughput and latency by processing transactions off Ethereum's main execution layer. These off-chain transactions are usually compressed and posted on Ethereum in batches. 
-
-Thousands of transactions could happen off-chain, but Ethereum nodes need to process one on-chain transaction. This reduces congestion on the base layer and reduces fees for users, while ensuring faster transactions. 
+[Layer 2 scaling solutions](/layer-2/), such as [rollups](/glossary/#rollups), scale network throughput and latency by processing transactions off Ethereum's main execution layer. Off-chain transactions are compressed and posted on Ethereum in batches—thousands of transactions could happen off-chain, but Ethereum needs to process *one* on-chain transaction associated with each batch submission. This reduces congestion on the base layer and reduces fees for users, while ensuring faster transactions. 
 
 However, for Ethereum to guarantee the security of rollups, it needs a mechanism for verifying the validity of off-chain transactions. This is where data availability comes into the picture. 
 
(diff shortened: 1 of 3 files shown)


src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
+---
+title: Data availability 
+description: An overview of problems and solutions relating to data availability in Ethereum 
+lang: en
+sidebar: true
+---
+
+In Ethereum, a node proposing a new block is required to make the underlying data available to the remaining peer-to-peer nodes on the network. This concept, known as data availability, is fundamental to the security and functionality of every blockchain, including Ethereum. Data availability is so important that an entire field of research studying the so-called “data availability problem” exists.
```
minimalsm, line 8: I think we can explain this in simpler terms. The high-level ELI5 I'd expect to see here is that if we can’t reproduce something with the data we have available, from the blockchain’s perspective, it does not exist.

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Blockchain security {#blockchain-security}
+
+Data availability is crucial for blockchain security, or else “data withholding attacks” would become commonplace. A data withholding attack occurs when a miner or validator publishes a block but doesn’t share the transaction data used to build the block. 
+
+If a data withholding attack happens, full nodes cannot verify the correctness of updates to Ethereum’s world state. This gives malicious block producers leeway to subvert protocol rules and advance invalid state transitions on the Ethereum network. 
+
+This is important because other network participants, such as light clients, rely on full nodes to verify the network’s state. Unlike full nodes, light clients only check block headers and don’t download block data. Therefore, the rules around data availability ensure full nodes can validate blocks and prevent the chain from getting corrupted. 
+
+### Decentralized scalability {#decentralized-scalability}
```
minimalsm, line 53: I'd love to see an explicit section on monolithic vs modular blockchains. I think we imply it a lot but explaining how something like explaining monolithic blockchains do not have a data availability problem because every node replicates all data of every block, but in order to scale without increasing node requirements (decentralisation) we need to go modular. 

Not 100% sure where in this article that would best live, I'd leave that to your interpretation/vision.

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
+---
+title: Data availability 
+description: An overview of problems and solutions relating to data availability in Ethereum 
+lang: en
+sidebar: true
+---
+
+In Ethereum, a node proposing a new block is required to make the underlying data available to the remaining peer-to-peer nodes on the network. This concept, known as data availability, is fundamental to the security and functionality of every blockchain, including Ethereum. Data availability is so important that an entire field of research studying the so-called “data availability problem” exists. 
+
+## What is data availability? {#what-is-data-availability}
```
minimalsm, line 10: All developer docs should have a prerequisites section. Ideally we have this in mind when writing/reviewing as it informs the level of granularity we should go into about specific topics—each developer doc isn’t a stand-alone article, ideally, it should live relative to every other developer doc with clearly defined expected knowledge to comprehend.

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
+---
+title: Data availability
```
minimalsm, line 2: Meta: where do you see this living in the docs? Probably the advanced section? Inside or after directly after scaling?

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+To solve this problem, we combine data availability sampling with [erasure coding](https://en.wikipedia.org/wiki/Erasure_code) to create “data availability proofs”. Erasure coding is a technique that allows us to double a dataset by adding redundant pieces (called erasure codes). If the original data is lost, the erasure codes can be used to reconstruct the original piece of data. 
+
+When implemented in blockchains, erasure codes improve data availability because a small fraction of the data is enough to reconstruct the whole transaction set in a block. In this system, a malicious block producer would need to withhold more than 50% of the block to perform a data withholding attack. Previously, a block producer only needed to seize 1% of block data to act maliciously. 
+
+With erasure-coded blocks, light clients have statistical certainty that the entire block data was published on the network. It also means that light clients don't have to rely on full nodes to alert them of the unavailability of a block. 
+
+[More on data availability proofs.](https://github.com/ethereum/research/wiki/A-note-on-data-availability-and-erasure-coding)
+
```
minimalsm, line 145: This stops quite abruptly based on previous content. I'd expect some sort of wrap-up section here. Perhaps pointing to the solution, Ethereum is leaning toward or the further work being done? Not sure but it just feels a little blunt of a stop to me.

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+description: An overview of problems and solutions relating to data availability in Ethereum 
+lang: en
+sidebar: true
+---
+
+In Ethereum, a node proposing a new block is required to make the underlying data available to the remaining peer-to-peer nodes on the network. This concept, known as data availability, is fundamental to the security and functionality of every blockchain, including Ethereum. Data availability is so important that an entire field of research studying the so-called “data availability problem” exists. 
+
+## What is data availability? {#what-is-data-availability}
+
+“Data availability” is a guarantee that all transaction data for a block was published by the block proposer and is available to other network participants. As explained in “[Intro to Ethereum](/developers/docs/intro-to-ethereum/)”, Ethereum transactions are processed in [blocks](/developers/docs/blocks/) and chained together to form the “blockchain”.
```
minimalsm, line 12: This feels like a very roundabout way of saying that all data should be available for transactions to reproduce the current world state

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Off-chain data availability {#off-chain-data-availability} 
+
+Off-chain data availability solutions move data storage off Ethereum's base layer. The block producer doesn't publish the data on-chain, but provides a cryptographic commitment to prove the availability of the data. 
+
+This is a method used by some scaling solutions, such as [validiums](/developers/docs/scaling/validium/), to reduce the amount of data posted on Ethereum. The drawback is that this system requires trusting block producers to release the data if necessary.  
+
+## What are some solutions to the data availability problem? {#solutions-to-data-availability-problem}
+
+As mentioned, the data availability problem concerns the ability to verify availability of the transaction data for a newly proposed block. Solutions to this problem employ some mechanism for guaranteeing data availability.
```
minimalsm, line 104: Can we order the subheadings here from most to least impactful from the Ethereum ecosystem's perspective? In my mind data availability sampling is the favourite atm?

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+[More on proof-of-stake data availability committees.](https://blog.matter-labs.io/zkporter-a-breakthrough-in-l2-scaling-ed5e48842fbf) 
+
+### Data availability sampling {#data-availability-sampling}
+
+Data availability sampling is a cryptographic mechanism for guaranteeing data availability. Data availability sampling allows blockchain nodes to verify that data for a proposed block is available without having to download the entire block. 
+
+In a DAS system, a node samples small, random chunks of a block over multiple rounds to verify data availability. With many nodes sampling different parts of a block simultaneously, its  availability can be verified with high statistical certainty. 
+
+When applied to blockchains, like Ethereum, data availability sampling ensures light clients also participate in guaranteeing the chain's security and functionality. Light clients can run without expensive hardware, making it easier for anyone to validate on the Ethereum network.
```
minimalsm, line 130: Should we indicate that this is the solution we're leaning toward?

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+The problem is that on-chain data availability places bottlenecks on scalability. Full nodes have to download the same data and check it, which reduces the number of transactions that can be processed per second.
+
+More importantly, on-chain DA has negative implications for decentralization since it increases the data that full nodes must hold. If Ethereum’s state spirals, validators must invest in more expensive machines—which would probably reduce the number of people willing to run a validating node.  
+
+### Off-chain data availability {#off-chain-data-availability} 
+
+Off-chain data availability solutions move data storage off Ethereum's base layer. The block producer doesn't publish the data on-chain, but provides a cryptographic commitment to prove the availability of the data. 
+
+This is a method used by some scaling solutions, such as [validiums](/developers/docs/scaling/validium/), to reduce the amount of data posted on Ethereum. The drawback is that this system requires trusting block producers to release the data if necessary.
```
minimalsm, line 100: We should be stronger in explaining the negative implications of off-chain DA here. 

If I read this as someone who didn't understand DA at all, these would feel more equally weighted in terms of potential solutions to the data availability problem than they actually are.

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+To solve this problem, we combine data availability sampling with [erasure coding](https://en.wikipedia.org/wiki/Erasure_code) to create “data availability proofs”. Erasure coding is a technique that allows us to double a dataset by adding redundant pieces (called erasure codes). If the original data is lost, the erasure codes can be used to reconstruct the original piece of data. 
+
+When implemented in blockchains, erasure codes improve data availability because a small fraction of the data is enough to reconstruct the whole transaction set in a block. In this system, a malicious block producer would need to withhold more than 50% of the block to perform a data withholding attack. Previously, a block producer only needed to seize 1% of block data to act maliciously. 
+
+With erasure-coded blocks, light clients have statistical certainty that the entire block data was published on the network. It also means that light clients don't have to rely on full nodes to alert them of the unavailability of a block. 
+
+[More on data availability proofs.](https://github.com/ethereum/research/wiki/A-note-on-data-availability-and-erasure-coding)
```
minimalsm, line 144: Forgive my ignorance but isn't this a subsection of DAS? i.e. the current research suggests that DAS will include erasure coding and KZG commitment already?

src/content/developers/docs/data-availability.md
minimalsm: Shouldn't we talk here about Danksharding?

minimalsm (changes requested):
Hey @emmanuel-awosika, this is brilliant so far.

Done a big picture/high-level structural review, overall this is pretty solid, great work 😎. Some changes suggested here: 

- I'd like to see prerequisites added; 
- a little more depth on monolithic vs modular; 
- also had some questions about the solutions section, 
- and I think we should add some information on danksharding. 

Working on doing a copyediting review now.

src/content/developers/docs/data-availability.md
jmcook1186: +1. I think it is also a nice place to include a "don't trust, verify" reminder, as ultimately this is why we care about data availability.

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+Each block has two major parts: 
+
+- The **block header**: This contains general information (metadata) about the block, such as the timestamp, block hash, block number, etc.
+- The  **block body**: This contains the actual transactions processed as part of the block. 
+
+[Miner nodes](/developers/docs/consensus-mechanisms/pow/mining/) (or validators in [proof-of-stake](/developers/docs/consensus-mechanisms/pos/)) proposing new blocks must publish both the block header and block body. This allows nodes participating in [consensus](/developers/docs/consensus-mechanisms/) (full nodes) to download the block and confirm that transactions follow Ethereum’s consensus protocol. 
+
+[Full nodes](/developers/docs/nodes-and-clients/#full-node) are different from [light nodes](/developers/docs/nodes-and-clients/#light-node) that only download the block header. Without full nodes verifying transactions, block proposers could get away with inserting malicious transactions in blocks. 
+
```
jmcook1186, line 22: This section kind of makes it seem like miners are not also full nodes. Besides, referring to miners puts a short expiry date ont he page since block production will be done by validators after the merge. Suggest we refer to "block producers" instead of miners/validators? This section could be simplified by replacing with something like:

"All nodes execute the transactions in blocks they receive from peers to ensure that the changes proposed by the block producer precisely match those computed independently by the node. This is how nodes verify that new information is valid, rather than having to trust that block producers are honest. This is not possible if any data is missing".

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+### The data availability problem {#the-data-availability-problem}
+
+The data availability problem is encapsulated in one question: “How do we verify that the data for a newly produced block is actually available?” As mentioned, the security Ethereum rests on the assumption that full nodes have access to block data. 
+
+If a block producer publishes a new block, but hides some (or all) of the underlying information, they might be able to finalize invalid transactions. Even if a block is valid, withholding data still has negative implications for users and the functionality of the network.   
+
+Most early discussions around data availability concerned the visibility of transaction data to full nodes (validators) participating in Ethereum’s consensus. However, newer DA research is also focused on verifying data availability with light clients. For light clients, the problem of data availability concerns validating a block’s data availability without having to download the entire block.
```
jmcook1186, line 29: I think an equally critical aspect is DA for zk and optimistic rollups

jmcook1186: and also how data will be provably available on a stateless Ethereum (e.g. Portal network development)

src/content/developers/docs/data-availability.md
jmcook1186: +1 It kind of undermines the "don't trust, verify" maxim that I suggest we repeat at the start of the intro

src/content/developers/docs/data-availability.md
jmcook1186: Could be nice - just to say something like "state-of-the-art scaling proposals such as Danksharding rely upon DAS to verify that the entire contents of a blob have been seen by the network, relieving individual nodes from the burden of downloading and validating it all directly". I think there is some nuance to whether this translates into data retrievability - afaik it doesn't.

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+## What are some solutions to the data availability problem? {#solutions-to-data-availability-problem}
+
+As mentioned, the data availability problem concerns the ability to verify availability of the transaction data for a newly proposed block. Solutions to this problem employ some mechanism for guaranteeing data availability. 
+
+### Data availability committees {#data-availability-committees}
+
+Pure validiums store transaction data off-chain with a block producer, making them centralized to an extent. This reduces decentralization and security, since the block producer can publish invalid transactions and conceal the rollup's true state by hiding transaction data. 
+
+Some validiums mitigate this problem by forcing block producers to store transaction data with trusted parties that form the Data Availability Committee (DAC). Members of the DAC are expected to securely store the data off-chain and publish it on-chain in event of a dispute. They also publish attestations on-chain to prove that the said data is indeed available.
```
minimalsm, line 110: > Some validiums mitigate this problem by forcing block producers

I don't think mitigate is the correct phrasing here. (I'd personally argue that this strategy is wishful thinking in many cases)

minimalsm (changes requested):
Copy editing run-through: never quite finished but thought I'd leave this here for now.

src/content/developers/docs/data-availability.md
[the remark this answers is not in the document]
jmcook1186: AFAIK light clients will only download block headers unless they specifically request other data. Precisely how the data is then served to them isn't quite certain but likely to initially be via req/resp to a full node, then later by requesting from the p2p layer.

src/content/developers/docs/data-availability.md
emmanuel-awosika: The advance section would be ideal. I also suggest place it after the page on scaling.

src/content/developers/docs/data-availability.md
emmanuel-awosika: Based on @jmcook1186's suggestion, I added a new section to explain how Ethereum is approaching data availability (i.e., by using DAS).

src/content/developers/docs/data-availability.md
emmanuel-awosika: Yes, that's correct. I just made the data availability proofs an H4 to nest it under the DAS section.

src/content/developers/docs/data-availability.md
emmanuel-awosika: Added new text touching on both points. LMK what you think.

src/content/developers/docs/data-availability.md
emmanuel-awosika: I created a new intro with ideas from both suggestions. Happy to hear your thoughts on it.

src/content/developers/docs/data-availability.md
emmanuel-awosika: Added this now.

src/content/developers/docs/data-availability.md
emmanuel-awosika: Fixed this.

src/content/developers/docs/data-availability.md
emmanuel-awosika: I used your suggestion as the basis for the new intro, since it's the ELI5 explanation @minimalsm wanted to see in the intro. I did adjust the original text in this section to a simpler variant based on your suggestion.

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+This is important because other network participants, such as light clients, rely on full nodes to verify the network’s state. Unlike full nodes, light clients only check block headers and don’t download block data. Therefore, the rules around data availability ensure full nodes can validate blocks and prevent the chain from getting corrupted. 
+
+### Decentralized scalability {#decentralized-scalability}
+
+Due to Ethereum’s constraints, scalability has become a widely discussed topic. However, Ethereum’s vision is to [scale computation without trading off decentralization and security[(/upgrades/vision/). Data availability is critical to achieving decentralized scalability as explained below:
+
+
+#### Data availability and layer 2 scaling {#data-availability-and-layer-2-scaling}
+
+[Layer 2 scaling solutions](/layer-2/), such as [rollups](/glossary/#rollups), scale network throughput and latency by processing transactions off Ethereum's main execution layer. These off-chain transactions are usually compressed and posted on Ethereum in batches.
```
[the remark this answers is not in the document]
emmanuel-awosika: > * Layer 2s execute off of Mainnet too
> That's what I said in the paragraph, or am I missing something there?

> > Usually compressed and posted on Ethereum in batches"
> 
> Usually? Are there cases where this isn't true for an L2?

No, it was a redundant word. Took it out now.

src/content/developers/docs/data-availability.md
emmanuel-awosika: Changed the sentence to read "...validiums attempt to solve this problem..". And yes, DACs have risks, which I explained briefly in the intro to the following section (PoS-based DACs).

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Why is data availability important? {#why-is-data-availability-important}
+
+### Blockchain security {#blockchain-security}
+
+Data availability is crucial for blockchain security, or else “data withholding attacks” would become commonplace. A data withholding attack occurs when a miner or validator publishes a block but doesn’t share the transaction data used to build the block. 
+
+If a data withholding attack happens, full nodes cannot verify the correctness of updates to Ethereum’s world state. This gives malicious block producers leeway to subvert protocol rules and advance invalid state transitions on the Ethereum network. 
+
+This is important because other network participants, such as light clients, rely on full nodes to verify the network’s state. Unlike full nodes, light clients only check block headers and don’t download block data. Therefore, the rules around data availability ensure full nodes can validate blocks and prevent the chain from getting corrupted.
```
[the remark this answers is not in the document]
emmanuel-awosika: Fixed this.

src/content/developers/docs/data-availability.md
emmanuel-awosika: So, I weaved the distinction between modular and monolithic blockchains into the off-chain DA vs on-chain DA section. That sounded better than creating a new section for the topic.

src/content/developers/docs/data-availability.md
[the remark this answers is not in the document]
emmanuel-awosika: Added a new section explaining the relevance of the data availability to light clients.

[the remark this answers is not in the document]
emmanuel-awosika: +1 on this. Most of the available material on DAS suggests that light clients will only have to download random chunks of a block's data to verify availability. The idea is that, with enough light clients sampling the same block, it's harder to conceal the unavailability of a block.

For context (from [Vitalik](https://hackmd.io/@vbuterin/sharding_proposal#ELI5-data-availability-sampling)): 
> Data availability sampling is in some ways a mirror image of randomly sampled committees. There is still sampling going on, in that each node only ends up downloading a small portion of the total data, but the sampling is done client-side, and within each blob rather than between blobs. Each node (including client nodes that are not participating in staking) checks every blob, but instead of downloading the whole blob, they privately select N random indices in the blob (eg. N = 20), and attempt to download the data at just those positions.

> The goal of this task is to verify that at least half of the data in each blob is available. If less than half of the data is available, then it is almost certain that at least one of the indices that any given client samples will not be available, and so the client will reject the blob.

src/content/developers/docs/data-availability.md
emmanuel-awosika: I already included some info about sharding and data availability in the "decentralized scalability" section, where I also talk about layer 2s. But, since you requested, I moved the sharding + DAS part to the concluding section and added @jmcook1186's addition about Danksharding.

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+Each block has two major parts: 
+
+- The **block header**: This contains general information (metadata) about the block, such as the timestamp, block hash, block number, etc.
+- The  **block body**: This contains the actual transactions processed as part of the block. 
+
+[Miner nodes](/developers/docs/consensus-mechanisms/pow/mining/) (or validators in [proof-of-stake](/developers/docs/consensus-mechanisms/pos/)) proposing new blocks must publish both the block header and block body. This allows nodes participating in [consensus](/developers/docs/consensus-mechanisms/) (full nodes) to download the block and confirm that transactions follow Ethereum’s consensus protocol. 
+
+[Full nodes](/developers/docs/nodes-and-clients/#full-node) are different from [light nodes](/developers/docs/nodes-and-clients/#light-node) that only download the block header. Without full nodes verifying transactions, block proposers could get away with inserting malicious transactions in blocks.
```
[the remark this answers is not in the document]
emmanuel-awosika: Edited this part.

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Data availability vs. data retrievability {#data-availability-vs-data-retrievability}
+
+Data availability is a different idea from data retrievability. Data availability is the ability of peer-to-peer nodes to download transaction data for a block while it is being proposed for addition to the chain. In other words, data availability is relevant when a block is yet to pass consensus. 
+
+Data retrievability is the ability of nodes to download transactions from a block *after* it has passed consensus. Essentially, data retrievability is concerned with retrieving data from historical blocks. A block becomes “historical” when other blocks have been built on it. 
+
+The core Ethereum protocol is primarily concerned with data availability, not data retrievability. It isn’t the blockckhain’s duty to store data for every transaction it reaches consensus on. Besides, storing state data in perpetuity increases hardware requirements for full nodes and may harm decentralization.
```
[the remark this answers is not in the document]
emmanuel-awosika: > This seems a little simplistic. Obviously, Ethereum needs to allow data to be retrieved at some point past the block being confirmed.

Added a more specific description of data retrievability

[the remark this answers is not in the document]
emmanuel-awosika: > > storing state data in perpetuity increases hardware requirements for full node
> 
> This technically isn't true. Full nodes are typically pruned, and requests are made to archive nodes to reconstruct the entirety of the chain.

Here are some Vitalik quotes for context on why I included that part:

> One of the longest and still unresolved challenges in the Ethereum protocol is that of how to manage the problem of growing state. Many actions in Ethereum (creating an account, filling a storage slot, sending ETH to a new account…) add objects to the state that all full nodes must store in order to be able to verify or produce new blocks. These actions require a one-time gas cost to be paid by the sender of the transaction, but impose permanent ongoing costs on the network, and the nodes that need to store this new data (and future nodes that need to download it when syncing).

> This is a major imbalance in the system’s design, and risks the Ethereum system becoming more and more difficult to use over time as the state fills with no-longer-needed “junk data”. The purpose of this document is to describe in more detail what the problem is, and what are some of the paths toward solving it. If we successfully implement a solution, it could pave the way to safely making significant increases to the block gas limit.

> The Ethereum state size is growing quickly. It is currently around 35 GB for just the state and over 100 GB including all Merkle proofs, and is increasing by roughly half that amount per year. State storage is additionally a weak point in Ethereum’s economics: it is the only mechanism by which a participant can pay a cost once to burden consensus nodes forever. In order to maintain the scalability and sustainability of Ethereum, we need some solution. Two paths to a solution exist, and have existed for a long time: weak statelessness and state expiry:

src/content/developers/docs/data-availability.md
```diff
@@ -0,0 +1,153 @@
...
+
+Data retrievability is the ability of nodes to download transactions from a block *after* it has passed consensus. Essentially, data retrievability is concerned with retrieving data from historical blocks. A block becomes “historical” when other blocks have been built on it. 
+
+The core Ethereum protocol is primarily concerned with data availability, not data retrievability. It isn’t the blockckhain’s duty to store data for every transaction it reaches consensus on. Besides, storing state data in perpetuity increases hardware requirements for full nodes and may harm decentralization. 
+
+Nevertheless, data retrievability is a much easier problem to solve than data availability. The ability to retrieve historical blockchain data only needs a weak `1-of-N` assumption, meaning only one honest node has to store the data to make it retrievable. It is also assumed that many entities, such as blockchain explorers, have incentives to store archival data and can make it available to others on request. 
+
+[More on solutions to the data retrievability problem](https://notes.ethereum.org/@vbuterin/data_sharding_roadmap#Who-would-store-historical-data-under-sharding).
```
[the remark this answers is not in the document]
emmanuel-awosika: > We only use the 'More on ....' style to link internally. Weakly in favour of moving these sections to further reading.

I'd have done that, but for these reasons:
- The further reading section already has many articles
- Some of the links point to relevant sections of each article, not the entire article (like the one on data retrievability).
- I felt linking to resources on each concept was more useful compared to throwing everything in the further reading section. The articles in the FR section tackle the overall concept of DA, not the individual solutions to the DA problem.

src/content/developers/docs/data-availability.md
emmanuel-awosika: Advanced section would make sense. Let's add it after scaling--I think DA is more than just scaling and needs its own page.

src/content/developers/docs/data-availability.md
emmanuel-awosika: Maybe you can explain this with more detail? Not totally sure what you mean here.

src/content/developers/docs/data-availability.md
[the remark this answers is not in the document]
emmanuel-awosika: Fixed this.

minimalsm (approved):
Added to developer docs and merging 😀

Thanks @emmanuel-awosika and to everyone who helped review!

PR #6623, opened 2022-06-09
Update Card.tsx

wackerow:
Was looking to use this in another PR, but given it is a more global change I have broken it out here.

- Tweaked `Card` component to conditionally render elements, only if they are available

### DIFF
--- a/src/components/Card.tsx
+++ b/src/components/Card.tsx
@@ -19,8 +19,8 @@ const TopContent = styled.div``
 
 export interface IProps {
   emoji?: string
-  title: string
-  description: string
+  title?: string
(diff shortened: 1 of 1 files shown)


pettinarip (commented):
@wackerow could you mark those fields as optional?
```ts
title?: string
description?: string
``

wackerow:
@pettinarip As it stands, all of these are optional. Did you want me to hard type this while we're here? Noticing we don't seem to be doing this for the other components that have been migrated to .tsx

wackerow:
nm, brain toot

pettinarip (approved):

PR #6627, opened 2022-06-09
add note about wallet addresses to layer 2 page

corwintines:
## Description
- Add a note about EOA and smart contract wallets on layer 2

### DIFF
--- a/src/pages/layer-2.js
+++ b/src/pages/layer-2.js
@@ -681,14 +681,17 @@ const Layer2Page = ({ data }) => {
         </p>
         <p>
           NOTE: When bridging over and using layer 2, it is important to note
-          that you will control the address for you EOA account just like on
-          Ethereum Mainnet. However, if you are using a smart contract wallet,
-          such as <Link to="https://gnosis-safe.io/">Gnosis Safe</Link> or{" "}
+          that you will control the address for your{" "}
+          <Link to="/glossary/#eoa">EOA account</Link> (an account where only a
+          single private key controls the account) just like on Ethereum
+          Mainnet. However, if you are using a{" "}
+          <Link to="/glossary/#contract-account">contract account</Link>, such
+          as <Link to="https://gnosis-safe.io/">Gnosis Safe</Link> or{" "}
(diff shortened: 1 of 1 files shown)


src/pages/layer-2.js
```diff
@@ -679,6 +679,15 @@ const Layer2Page = ({ data }) => {
         <p>
           <Translation id="layer-2-use-layer-2-1" />
         </p>
+        <p>
+          NOTE: When bridging over and using layer 2, it is important to note
+          that you will control the address for you EOA account just like on
+          Ethereum Mainnet. However, if you are using a smart contract wallet,
```
timbeiko, line 685: I would link a couple examples here, such as Gnosis Safe, Argent, etc.

src/pages/layer-2.js
```diff
@@ -679,6 +679,15 @@ const Layer2Page = ({ data }) => {
         <p>
           <Translation id="layer-2-use-layer-2-1" />
         </p>
+        <p>
+          NOTE: When bridging over and using layer 2, it is important to note
+          that you will control the address for you EOA account just like on
```
timbeiko, line 684: I would expand on "EOA" account, e.g. an account where only a single private key controls the funds / users a recovery phrase / etc. given that this warning is probably more for novice than advanced users.

src/pages/layer-2.js
```diff
@@ -679,6 +679,15 @@ const Layer2Page = ({ data }) => {
...
+          Ethereum Mainnet. However, if you are using a smart contract wallet,
+          you will not have control over this address on a layer 2 by default
+          unless deployed using `CREATE2`. If you are bridging or sending funds
```
timbeiko, line 687: > unless deployed using `CREATE2`

wouldn't you need to also redeploy on the L2 before? If there is an extra step to take we should maybe list it or simplify the sentence.

timbeiko (commented):
Thanks for getting this up so quickly! Added a couple minor comments 😄

minimalsm (approved):

PR #154, opened 2019-08-06
Korean site translation

sbrichards:
- Add Korean translation pages
- Add sitewide "Languages" dropdown menu

Preview:
https://deploy-preview-154--ethereumorg.netlify.com

PR #6628, opened 2022-06-09
Feat/convert e files

vdusart:
Moving files starting with letter "E" to TS

# Related Issue


src/components/EthExchanges.tsx
```diff
@@ -543,9 +558,15 @@ const EthExchanges = () => {
   const exchangesArray = Object.keys(exchanges)
   const walletProvidersArray = Object.keys(walletProviders)
   // Construct arrays for CardList
-  let filteredExchanges = []
-  let filteredWalletProviders = []
-  let filteredWallets = []
+  let filteredExchanges: Array<{
```
corwintines, line 561: I think you can get away with `Array<Wallet>` here

src/components/EthPriceCard.tsx
```diff
@@ -123,7 +137,7 @@ const EthPriceCard = ({ className, isLeftAlign }) => {
     price = <Translation id="loading-error-refresh" />
   }
 
-  const isNegativeChange = state.percentChangeUSD && state.percentChangeUSD < 0
+  const isNegativeChange = state.percentChangeUSD < 0
```
corwintines, line 140:
```suggestion
  const isNegativeChange = state?.percentChangeUSD < 0
```
Just looking at the changes from before makes me think percentChange can be falsy, so just through the ?. operand here I think would make this equivalent to what was there before.

rsalvo77 (approved):

src/components/EthExchanges.tsx
vdusart: Yeah that's what I thought but I figured it could be confusing.
I'm gonna fix it!

src/components/EthExchanges.tsx
```diff
@@ -133,8 +134,18 @@ export const cardListImage = graphql`
   }
```
pettinarip, line 134: Lets add some types on the upper components, `NoResults` and `NoResultsSingle`

src/components/EthExchanges.tsx
```diff
@@ -516,11 +527,15 @@ const EthExchanges = () => {
...
   )
 
-  const [state, setState] = useState({ selectedCountry: {} })
+  const [state, setState] = useState<{
+    selectedCountry: {
+      country?: string
+    }
+  }>({ selectedCountry: {} })
 
   const handleSelectChange = (selectedOption) => {
```
pettinarip, line 540: Define types for this function

src/components/EthExchanges.tsx
```diff
@@ -133,8 +134,18 @@ export const cardListImage = graphql`
...
+}
+
+export interface IProps {}
+
 // TODO move component into get-eth.js page?
-const EthExchanges = () => {
+const EthExchanges: React.FC<IProps> = () => {
   const intl = useIntl()
   const placeholderString = translateMessageId(
     "page-get-eth-exchanges-search",
```
pettinarip, line 151: We need types for the query in line `#154 `
```ts
// Add a name to the query & run `yarn start` to generate the query type def
const data = useStaticQuery<Queries.EthExchangesQuery>(graphql`
    query EthExchanges {
```

src/components/EthExchanges.tsx
```diff
@@ -543,9 +558,15 @@ const EthExchanges = () => {
...
-  let filteredExchanges = []
-  let filteredWalletProviders = []
-  let filteredWallets = []
+  let filteredExchanges: Array<{
+    title: string
+    description: string
+    link: string
+    image: string
+    alt: string
+  }> = []
```
pettinarip: Looks like this type is the same we are using for the `Wallet` interface. Maybe we can rename it and use it in both places.

pettinarip (changes requested):
Thanks for the contribution @vdusart. This is looking really good. I've left a couple of change requests.

The `EthExchanges` component is pretty messy. If you prefer to skip it let me know and I can take care of it, in case it demands too much time to migrate it.

src/components/EthPriceCard.tsx
pettinarip: Agree!

vdusart:
@pettinarip I tried do everything as you wanted.
Just 2 things:
1- I don't really understand what you meant with that : https://github.com/ethereum/ethereum-org-website/pull/6628#discussion_r894934254

2- I don't know enough graphql  to fix the error due to that: https://github.com/ethereum/ethereum-org-website/pull/6628#discussion_r894935804

So yes if you have time I can give it to you and convert other files ! :smile:

pettinarip:
@vdusart,

1. Sorry, I wasn't clear. What I meant was that there are 2 more components defined in this file (called `NoResults` and `NoResultsSingle`) that also need to be migrated to TS. They are in line `#103`.
2. Cool, I'll assign that file to myself so that I don't forget to migrate it. Please, revert the changes you did on `EthExchanges` then, and really glad you keep migrating other files 🙏🏼

pettinarip (approved):
🚀

src/components/DeveloperDocsLinks.tsx
```diff
@@ -3,6 +3,7 @@ import { ListItem } from "./SharedStyledComponents"
 import Translation from "./Translation"
 import Link from "./Link"
 import docLinks from "../data/developer-docs-links.yaml"
+import { TranslationKey } from "../utils/translations"
```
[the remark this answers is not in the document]
pettinarip: Done

pettinarip:
@all-contributors please add @vdusart for code

Issue #6626, opened 2022-06-09
[TS Code Style]Emoji component has a required size props


Mousticke:
**Describe the bug**

This is purely coding style issue.
The `Emoji` component has a `size` props with a default value
```ts
const Emoji: React.FC<IProps> = ({ size = 1.5, text, ...props }) 
```
with a defined interface `IProps`

```ts
export interface IProps extends MarginProps {
  size: number
  text: string
}
```

Here `size` is required but set with a default value in the functional components. Other components require the `Emoji` component and have to set the `size` because of the interface.
We should replace the required props as an optional props.

**Expected behavior**

We expect this interface in `src/components/Emoji.tsx`
```ts
export interface IProps extends MarginProps {
  size?: number
  text: string
}
```

**Additional context**

Part of the Typescript rollout: #6392 
Emoji PR for the migration: #6540  

**Want to contribute?**

You can assign me this issue if you want.

pettinarip:
Good catch @Mousticke. I've assigned it to you

PR #6631, opened 2022-06-10
TS migrate asset download

Mousticke:
## Description

Migrate `AssetDownload` to typescript. The component is massively used in the page `assets`.

### Props
Either you have a `svg` or `image`. But we can never have both in the same time from the logic inside the component and the way of this component being used in `pages/assets.js`. That's the purpose of `IPropsWithSVG` and `IPropsWithImage`.

### Declare svg module
importing a svg in React cause the typescript compiler to raise an error : "Could not find module...". svg are not React component so we have to define a module that export it as a React component. That's the purpose of `assets.d.ts`. For the next coming assets file, we can import them as a module in this file.

## Extra

Added a clean command in `package.json` if there are issues related to schema namespace or seg fault compiler error. `.env` parameters `GATSBY_BUILD_LANGS` and `IGNORE_CONTENT` are not enough sometimes.

## Related Issue

Fix: #6626 
Refs: #6392

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,7 @@
   "scripts": {
     "build": "gatsby build",
     "build:lambda": "netlify-lambda build src/lambda",
+    "clean": "gatsby clean",
     "copy-contributors": "node src/scripts/copy-contributors.js",
     "format": "prettier --write \"**/*.{js,jsx,json,md}\"",
     "generate-heading-ids": "node src/scripts/generate-heading-ids.js",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -87,6 +87,7 @@
   "scripts": {
     "build": "gatsby build",
     "build:lambda": "netlify-lambda build src/lambda",
+    "clean": "gatsby clean",
```
pettinarip, line 90: 👍🏼

pettinarip (approved):
Cool! Thanks @Mousticke this looks great to me.

PR #6645, opened 2022-06-11
TS migrate BoxGrid

Mousticke:
## Description

Migrate `Boxgrid` component to typescript.
Add interfaces : 
- `IBoxProps` : Array of items passed to the `BoxGrid` component coming from `dapps` page
- `IBaseItem` : Base data single item coming from `dapps` page
- `IGridItemProps` : Items wrapping a single item `IBaseItem`, styling `IBoxStyle` and state parameters of the Box Component
- `IBoxStyle` : Wrap all styling properties
- `IOpener` : `isOpen` boolean used in `GridItem` component and for `Box` styling
- `IGridStyle` : `columnNumber` number used in `GridItem` component and for `Box` styling
- `IColor` : `color` string used in `GridItem` component and for `Box` styling

## Related Issue

Refs: #6392

### DIFF
--- a/src/components/BoxGrid.tsx
+++ b/src/components/BoxGrid.tsx
@@ -3,6 +3,35 @@ import styled from "styled-components"
 
 import Emoji from "./Emoji"
 
+export interface IOpener {
+  isOpen: boolean
+}
+
+export interface IGridStyle {
+  columnNumber: number
+}
+
+export interface IColor {
+  color: string
+}
+
+export interface IBaseItem {
+  emoji: string
+  title: string
+  description: string
+}
+
+export interface IBoxStyle extends IOpener, IGridStyle, IColor {}
+
+export interface IGridItemProps extends IBaseItem, IBoxStyle {
+  index: number
+  callback: (idx: number) => void
+}
+
+export interface IBoxProps {
+  items: Array<IBaseItem>
+}
+
 const Title = styled.h3`
   font-size: 2.5rem;
   font-weight: 400;
@@ -35,7 +64,7 @@ const StyledEmoji = styled(Emoji)`
   }
 `
 
-const Box = styled.div`
+const Box = styled.div<IBoxStyle>`
   grid-row-start: ${(props) => (props.isOpen ? `1` : `auto`)};
   grid-row-end: ${(props) => (props.isOpen ? `span 2` : `auto`)};
   grid-column-start: ${(props) => (props.isOpen ? props.columnNumber : `auto`)};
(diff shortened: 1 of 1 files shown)


src/components/BoxGrid.tsx
```diff
@@ -3,6 +3,35 @@ import styled from "styled-components"
...
+  color: string
+}
+
+export interface IBaseItem {
+  emoji: string
+  title: string
+  description: string
+}
+
+export interface IBoxStyle extends IOpener, IGridStyle, IColor {}
```
pettinarip, line 24: Lets try to keep it simple and avoid creating a lot of interfaces unless we clearly see a purpose for doing that (like reusing those interfaces in other places/components).

For this case, lets just do:
```ts
export interface IBoxStyle {
  isOpen: boolean
  columnNumber: number
  color: string
}
```

src/components/BoxGrid.tsx
```diff
@@ -3,6 +3,35 @@ import styled from "styled-components"
...
+
+export interface IGridItemProps extends IBaseItem, IBoxStyle {
+  index: number
+  callback: (idx: number) => void
+}
+
+export interface IBoxProps {
```
pettinarip, line 31: As we are only exporting 1 component in this file, rename this interface to `IProps` to keep it consistent with the rest of the components.

I would suggest only exporting this interface and `IBaseItem` as they are the only interfaces the consumers are going to interact with. I would keep the rest of the interfaces private (no `export`).

src/components/BoxGrid.tsx
```diff
@@ -70,9 +99,9 @@ const Box = styled.div`
 `
 
 // Represent string as 32-bit integer
-const hashCode = (string) => {
+const hashCode = (stringPhrase: string) => {
```
pettinarip, line 102:
```suggestion
const hashCode = (stringPhrase: string): number => {
```

pettinarip (changes requested):
Looking great @Mousticke a few comments to improve readability and to follow conventions.

src/components/BoxGrid.tsx
Mousticke: Thank you for noticing this issue

src/components/BoxGrid.tsx
Mousticke: I have rename `IBaseItem` into `IBoxItem` and followed your recommendations

pettinarip (approved):
Nice @Mousticke, thanks for that 🚀

PR #6661, opened 2022-06-13
add execution-layer info to client diversity page

jmcook1186:
## Description

adds short paragraph about execution layer diversity

### DIFF
--- a/src/content/developers/docs/nodes-and-clients/client-diversity/index.md
+++ b/src/content/developers/docs/nodes-and-clients/client-diversity/index.md
@@ -51,7 +51,7 @@ The execution layer data were obtained from [Ethernodes](https://ethernodes.org)
 
 Up to date client diversity data for the consensus layer is now available at https://clientdiversity.org/.
 
-## Execution Layer
+## Execution layer {#execution-layer}
 
 Up until now, the conversation around client diversity has focused mainly on the consensus layer. However, the execution client [Geth](https://geth.ethereum.org) currently accounts for around 85% of all nodes. This is problematic for the same reasons as for consensus clients. For example, a bug in Geth affecting transaction-handling or constructing execution payloads could lead to bugged information being finalized by the consensus clients. Therefore, Ethereum would be healthier with a more even distribution of execution clients, ideally with no client representing more than 33% of the network.
 
--- a/src/content/developers/docs/nodes-and-clients/client-diversity/index.md
+++ b/src/content/developers/docs/nodes-and-clients/client-diversity/index.md
@@ -51,6 +51,10 @@ The execution layer data were obtained from [Ethernodes](https://ethernodes.org)
 
 Up to date client diversity data for the consensus layer is now available at https://clientdiversity.org/.
 
+## Execution Layer
+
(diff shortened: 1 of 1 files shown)


src/content/developers/docs/nodes-and-clients/client-diversity/index.md
```diff
@@ -51,6 +51,10 @@ The execution layer data were obtained from [Ethernodes](https://ethernodes.org)
 
 Up to date client diversity data for the consensus layer is now available at https://clientdiversity.org/.
 
+## Execution Layer
+
+Up until now, the conversation around client diversity has focused mainly on the consensus layer. However, the execution client [Geth](https://geth.ethereum.org) currently accounts for around 85% of all nodes. This is problematic for the same reasons as for consensus clients. For example, a bug in Geth affecting transaction-handling or constructing execution payloads could lead to bugged information being finalized by the consensus clients. Therefore, Ethereum would be healthier with a more even distribution of execution clients, ideally with no client representing more than 33% of the network.
```
minimalsm, line 56:
```suggestion
Until now, the conversation around client diversity has focused mainly on the consensus layer. However, the execution client [Geth](https://geth.ethereum.org) currently accounts for around 85% of all nodes. This percentage is problematic for the same reasons as for consensus clients. For example, a bug in Geth affecting transaction handling or constructing execution payloads could lead to consensus clients finalizing invalid transactions. Therefore, Ethereum would be healthier with a more even distribution of execution clients, ideally with no client representing more than 33% of the network.
```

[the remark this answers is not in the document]
minimalsm: As the miners switch off and the validators start to run execution clients I'd say absolutely yes. However, Coinbase or Lido could just as easily spin up 100% Geth clients and cause the same problem. 

I'd like to think the pressure put upon them already would prevent that from happening but who knows 😅

minimalsm: Wasn't sure if 'bugged information being finalized' meant transactions, I tried to make it clearer... Please LMK if I misinterpreted

jmcook1186: I'm really wary of "invalid transactions" because it suggests transactions that violate the EVM rules set-out in the whitepaper (e.g. minting new ether, emptying arbitrary accounts, etc). "Bugged" could mean missing, maliciously or accidentally re-ordered or censored, etc.

But it's not a great word, maybe "problematic" or similar?

src/content/developers/docs/nodes-and-clients/client-diversity/index.md
```diff
@@ -51,6 +51,10 @@ The execution layer data were obtained from [Ethernodes](https://ethernodes.org)
 
 Up to date client diversity data for the consensus layer is now available at https://clientdiversity.org/.
 
+## Execution Layer
```
minimalsm, line 54:
```suggestion
## Execution layer {#execution-layer}
```

src/content/developers/docs/nodes-and-clients/client-diversity/index.md
minimalsm: Went with 'problematic or bugged transactions' for the sake of pulling this in @jmcook1186. LMK if that isn't correct 😀

PR #6662, opened 2022-06-13
TS migrate BeaconChainActions

Mousticke:
## Description

Migrate component to typescript

## Related Issue

Refs: #6392


src/components/BeaconChainActions.tsx
```diff
@@ -69,36 +71,51 @@ export const DataLogo = graphql`
...
-        ...DataLogo
-      }
+type ChildImage = {
+  childImageSharp: {
+    gatsbyImageData: Record<string, unknown>
+  }
+}
+
+const BeaconStaticQuery = graphql`
+  query {
```
pettinarip, line 81: In these cases, we need to give the query a name, like `query BeaconStatic {` in order to let gatsby generate the correct types. You build those query types by running `yarn develop`. You will find the new query type under this generated file: `src/gatsby-types.d.ts`.

Finally, you will be able to use them like this:
```ts
const data = useStaticQuery<Queries.BeaconStaticQuery>(BeaconStaticQuery)
```
Note that `Queries` is a global namespace.

pettinarip (changes requested):
Thanks for the contribution @Mousticke. Left a comment about the generation of query types.

src/components/BeaconChainActions.tsx
pettinarip: Ignore this since Gatsby is only generating the query types for the pages.

pettinarip (approved):
Perfect @Mousticke gj!

pettinarip:
@all-contributors please add @Mousticke for code

PR #6671, opened 2022-06-13
Feat/convert f files ts

vdusart:
Moving files starting with letter "F" to TS

# Related Issue


src/components/FeedbackWidget.tsx
```diff
@@ -197,22 +203,22 @@ const FeedbackWidget = ({ className }) => {
...
   const handleOpen = () => {
     setIsOpen(true)
     trackCustomEvent({
       eventCategory: `FeedbackWidget toggled`,
       eventAction: `Clicked`,
-      label: `Opened feedback widget`,
+      eventName: `Opened feedback widget`,
     })
   }
   const handleSubmit = (choice) => {
```
pettinarip, line 217: Lets define types for the function args + the return value
```suggestion
  const handleSubmit = (choice: boolean): void => {
```

src/components/FeedbackWidget.tsx
```diff
@@ -197,22 +203,22 @@ const FeedbackWidget = ({ className }) => {
     trackCustomEvent({
       eventCategory: `FeedbackWidget toggled`,
       eventAction: `Clicked`,
-      label: `Closed feedback widget`,
+      eventName: `Closed feedback widget`,
     })
   }
   const handleOpen = () => {
```
pettinarip, line 209:
```suggestion
  const handleOpen = (): void => {
```

pettinarip: We should do this for every inner function we have here.

src/components/FeedbackWidget.tsx
```diff
@@ -197,22 +203,22 @@ const FeedbackWidget = ({ className }) => {
...
-      label: `Opened feedback widget`,
+      eventName: `Opened feedback widget`,
     })
   }
   const handleSubmit = (choice) => {
     trackCustomEvent({
       eventCategory: `Page is helpful feedback`,
       eventAction: `Clicked`,
-      label: choice,
+      eventName: choice,
```
pettinarip, line 221: @wackerow do you know what should be the expected behavior here?

2 main concerns:
1. We were using `label` here, but it doesn't exist in the trackCustomEvent function.
2. If we want to use `eventName`, we should send a string but `choice` is a boolean. Should we send the stringified version of it, `"true" or "false"`?

src/components/FileContributors.tsx
```diff
@@ -47,9 +48,11 @@ const Container = styled.div`
   }
 `
 
-const SkeletonContainer = styled(Container)`
+const SkeletonContainer = styled(Container)<{
+  loading: boolean
+}>`
   justify-content: flex-start;
-  position: absolute;
+  position: absolute;é
```
pettinarip, line 55:
```suggestion
  position: absolute;
```

src/components/FileContributors.tsx
```diff
@@ -224,7 +227,17 @@ const COMMIT_HISTORY = gql`
...
+  editPath?: string
+}
+
+const FileContributors: React.FC<IProps> = ({
+  relativePath,
+  className,
+  editPath,
+}) => {
   const [isModalOpen, setModalOpen] = useState(false)
   const intl = useIntl()
```
pettinarip, line 243: Lets define an interface for `Commit`. Something like
```ts
interface Commit {
  author: {
    name: string
    email: string
    avatarUrl: string
    user: {
      login: string
      url: string
    }
  }
  committedDate: string
}

and use it down below as `Array<Commit>`.
```

src/components/Footer.tsx
```diff
@@ -384,7 +396,7 @@ const Footer = () => {
                   {section.links.map((link, linkIdx) => (
                     <ListItem key={linkIdx}>
                       <FooterLink
-                        dir={isPageRightToLeft ? "auto" : "ltr"}
+                        dir={isPageRightToLeft ? "rtl" : "ltr"}
```
pettinarip, line 399: `dir` can be `auto` https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir

Lets update the `Link` props to accept it `dir?: "ltr" | "rtl" | "auto"`

src/components/Footer.tsx
```diff
@@ -121,12 +122,23 @@ const socialLinks = [
   },
 ]
 
-const Footer = () => {
+export interface LinkSection {
```
pettinarip, line 125: Good job 👍🏼

pettinarip (changes requested):
Good job on this @vdusart. I have some change requests but is looking good. Thanks for the contribution.

pettinarip (approved):
Nice, good job @vdusart 🚀

PR #6691, opened 2022-06-14
Converted Staking components and pages to TS

heeckhau:
## Description

Converted the staking components and pages to TypeScript

## Related Issue

TypeScript rollout #6392


pettinarip (changes requested):
Hey @heeckhau thanks for the contribution!

I some files like `Pill` and `StakingCommunityCallout` where you only changed the files extension. Please, take a look at [this doc](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/typescript.md) to follow some of our suggestions & conventions on migrating our files. For example, we always want to have an interface called `IProps` in each component.

heeckhau:
@pettinarip updated

heeckhau:
@pettinarip I figured out how PageProps from graphql work and I have updated the page files of this PR

src/components/Pill.tsx
```diff
@@ -27,8 +27,19 @@ const Secondary = styled.div`
   font-size: 0.75rem;
   border-radius: 0.25rem;
 `
+export interface IProps {
+  children: JSX.Element
```
pettinarip, line 31: No need to specify `children` here since it is defined by `React.FC`

src/components/Pill.tsx
```diff
@@ -27,8 +27,19 @@ const Secondary = styled.div`
...
+  className: string
```
pettinarip, line 32:
```suggestion
  className?: string
```

src/components/Pill.tsx
```diff
@@ -27,8 +27,19 @@ const Secondary = styled.div`
...
+  isSecondary: boolean
```
pettinarip, line 33:
```suggestion
  isSecondary?: boolean
```

src/components/Staking/StakingCommunityCallout.tsx
```diff
@@ -53,6 +57,7 @@ const StakingCommunityCallout = (props) => {
       alt={translateMessageId("page-staking-image-alt", intl)}
       titleKey={"page-staking-join-community"}
       descriptionKey={"page-staking-join-community-desc"}
+      id={id}
```
pettinarip, line 60: There is no need to pass down this props since we are already doing `{...props}` above.

src/components/Staking/StakingCommunityCallout.tsx
```diff
@@ -29,7 +29,11 @@ const StyledButtonLink = styled(ButtonLink)`
   }
 `
 
-const StakingCommunityCallout = (props) => {
+export interface IProps {
+  id?: string
+}
+
+const StakingCommunityCallout: React.FC<IProps> = (props, id) => {
```
pettinarip, line 36: `id` is coming inside the `props` arg.
```suggestion
const StakingCommunityCallout: React.FC<IProps> = (props) => {
```

src/components/Staking/StakingComparison.tsx
```diff
@@ -63,12 +64,24 @@ const StyledPoolGlyph = styled(PoolGlyph)`
     fill: ${({ theme }) => theme.colors.stakingBlue};
   }
 `
-
-const StakingComparison = ({ page, className }) => {
+export interface IProps {
```
pettinarip, line 67: Lets reorganize it a bit to improve the readability and define a little better some types. We can put this on top of the function definition:
 
```ts
interface DataType {
  title: TranslationKey
  linkText: TranslationKey
  to: string
  matomo: EventOptions
  color: string
  glyph: ReactNode
}

type StakingPageType = "solo" | "saas" | "pools"

interface StakingPageData extends DataType {
  content: TranslationKey
}

export interface IProps {
  page: StakingPageType
  className?: string
}
```

and then in `data` you can just do:
```ts
const data: {
    [key in StakingPageType]: Array<StakingPageData>
  } = {
   ...
```

src/components/Staking/StakingConsiderations.tsx
```diff
@@ -47,7 +48,7 @@ const ListItem = styled.li`
   box-sizing: border-box;
   position: relative;
   height: 2rem;
-  ${({ theme, active }) =>
+  ${({ theme, active }: { theme: DefaultTheme; active: boolean }) =>
```
pettinarip, line 51: In this cases where we have custom props, lets pass this custom prop types like this:
```ts
const ListItem = styled.li<{ active: boolean }>`
```
that way you don't need to define the types as you did in this line.

src/components/Staking/StakingGuides.tsx
```diff
@@ -11,7 +11,7 @@ const StyledCardList = styled(CardList)`
   gap: 1rem;
 `
 
-const StakingGuides = () => {
+const StakingGuides: React.FC = () => {
```
pettinarip, line 14: By convention, lets always define the `IProps` interface.

src/components/Staking/StakingHierarchy.tsx
```diff
@@ -50,7 +50,9 @@ const Container = styled.div`
   }
 `
 
-const Section = styled.div`
+const Section = styled.div<{
+  number: string
```
pettinarip, line 54: Maybe we can just accept it as a `number` and change how we are using them on the consumers.
```tsx
<Section number={1}>
```

src/components/Staking/StakingHierarchy.tsx
```diff
@@ -248,7 +252,7 @@ const Line = styled.aside`
   }
 `
 
-const StakingHierarchy = () => {
+const StakingHierarchy: React.FC = () => {
```
pettinarip, line 255: Lets define the `IProps`

pettinarip (changes requested):
Thanks @heeckhau , I've made a partial review since it is a lot. I have left some comments. There are still some files that were only renamed to `tsx` without any type definitions.

I'll try to jump into this next week & help you out.

src/components/Pill.tsx
heeckhau: Learned something new. Thanks

src/components/Staking/StakingCommunityCallout.tsx
heeckhau: 👍

src/components/Staking/StakingComparison.tsx
heeckhau: I was doubting here, since the pages are hardcoded in multiple places anyway. But I agree. Thanks

src/components/Staking/StakingGuides.tsx
heeckhau: OK

src/components/Staking/StakingHierarchy.tsx
heeckhau: I went for the smalles change originally. But yes, this is better.

heeckhau:
@pettinarip Thanks a lot for this great review. I made the requested changes.

pettinarip (changes requested):
Looking great. I'll push some changes with some minor fixes but overall I think you did an excellent job @heeckhau. Thanks for the hard work you put in here.

pettinarip (approved):

pettinarip:
@heeckhau be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And we've just released our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) also 🥳!

PR #6702, opened 2022-06-15
Adding Explore by The Dapp List

nirbhikjangid:
Explore is a community-curated & governed registry of projects in Web3. Adding the DAOs category here so that community can explore DAOs that are supported by the community.


minimalsm:
Hey :-) thanks for the PR.

Adding a DAO aggregator to a list of DAOs doesn't seem right to me 🤔 feel free to suggest elsewhere adding elsewhere if it makes sense/fits our producing listing criteria but this isn't the right spot IMO.

PR #6722, opened 2022-06-17
Update channel emoji

vdusart:
## Description

Update the emoji of the poaps channel


minimalsm (approved):
Nice catch @vdusart 😀

PR #6732, opened 2022-06-17
Expand Optimistic Rollups docs

emmanuel-awosika:
@corwintines suggested expanding the rollup docs to include details about other players in the system (validators, sequencers, etc.), which this edit does. I also added details I felt may be relevant to developers and users who want to interact with optimistic rollups.

## Description

Here is a high-level overview of some of the changes I suggested:

- A section explaining the relationship between optimistic rollups and Ethereum 
- A section describing types of fraud proving schemes 
- A section explaining how optimistic rollups work (transaction execution, state commitment, etc.)
- A section on rollup fees (felt this would be quite useful for would-be users)
- A section explaining how rollups scale Ethereum via `calldata` compression

### DIFF
--- a/src/content/developers/docs/scaling/optimistic-rollups/index.md
+++ b/src/content/developers/docs/scaling/optimistic-rollups/index.md
@@ -224,9 +224,9 @@ Optimistic rollups use several techniques to achieve transaction data compressio
 
 Doing some rough calculations on these figures can help show the scalability improvements afforded by an optimistic rollup:
 
-1. The target size for every block is 15 million gas and it costs 16 gas to verify one byte of data; dividing the average block size by 16 gas (15,000,000/6) shows the average block can hold **937,500 bytes of data**.
-2. If a basic rollup transaction uses 12 bytes, then the average Ethereum block can process **78,125 rollup transactions** (937,5000/12) or **39 rollup batches** (if each batch holds an average of 2000 transactions).
-3. If a new block is produced on Ethereum every 15 seconds, then the rollup's processing speeds would amount to roughly **5,208 transactions per second**. This is done by dividing the number of basic rollup transactions an Ethereum block can hold (78,125) by the average block time (15 seconds).
+1. The target size for every block is 15 million gas and it costs 16 gas to verify one byte of data. Dividing the average block size by 16 gas (15,000,000/16) shows the average block can hold **937,500 bytes of data**.
+2. If a basic rollup transaction uses 12 bytes, then the average Ethereum block can process **78,125 rollup transactions** (937,5000/12) or **39 rollup batches** (if each batch holds an average of 2,000 transactions).
+3. If a new block is produced on Ethereum every 15 seconds, then the rollup's processing speeds would amount to roughly **5,208 transactions per second**. This is done by dividing the number of basic rollup transactions an Ethereum block can hold (**78,125**) by the average block time (**15 seconds**).
 
 This is a fairly optimistic estimate, given that optimistic rollup transactions cannot possibly comprise an entire block on Ethereum. However, it can give give a rough idea of how much scalability gains that optimistic rollups can afford Ethereum users (current implementations offer up to 2,000 TPS).
 
@@ -256,7 +256,7 @@ Multiple implementations of Optimistic rollups exist that you can integrate into
 
 <RollupProductDevDoc rollupType="optimistic" />
 
-**Optimistic rollups reading**
+## Further reading on optimistic rollups 
 
 - [Everything you need to know about Optimistic Rollup](https://research.paradigm.xyz/rollups)
 - [EthHub on optimistic rollups](https://docs.ethhub.io/ethereum-roadmap/layer-2-scaling/optimistic_rollups/)
@@ -264,4 +264,4 @@ Multiple implementations of Optimistic rollups exist that you can integrate into
 - [How does Optimism's Rollup really work?](https://research.paradigm.xyz/optimism)
 - [OVM Deep Dive](https://medium.com/ethereum-optimism/ovm-deep-dive-a300d1085f52)
 - [What is the Optimistic Virtual Machine?](https://www.alchemy.com/overviews/optimistic-virtual-machine)
-- [More on layer 2](/layer-2/)
+- [How Do Optimistic Rollups Work? (The Complete Guide)](https://www.alchemy.com/overviews/optimistic-rollups)
--- a/src/content/developers/docs/scaling/optimistic-rollups/index.md
+++ b/src/content/developers/docs/scaling/optimistic-rollups/index.md
@@ -1,6 +1,6 @@
 ---
 title: Optimistic Rollups
-description: An introduction to optimistic rollups, a scaling solution used by the Ethereum community. 
+description: An introduction to optimistic rollups as a scaling solution used by the Ethereum community. 
 lang: en
 sidebar: true
 ---
@@ -11,9 +11,7 @@ As computation is the slow, expensive part of using Ethereum, optimistic rollups
 
 ## Prerequisites {#prerequisites}
 
-You should have a good understanding of all the foundational topics and a high-level understanding of [Ethereum scaling](/developers/docs/scaling/). Implementing scaling solutions such as rollups is an advanced topic as the technology is less battle-tested and continues to be researched and developed.
-
-Looking for a more beginner-friendly resource? See our [introduction to layer 2](/layer-2/).
+You should have read and understood our pages on [Ethereum scaling](/developers/docs/scaling/) and [layer 2](/layer-2/). 
 
 ## What is an optimistic rollup? {#what-is-an-optimistic-rollup} 
 
@@ -25,13 +23,25 @@ Optimistic rollups are considered “optimistic” because they assume off-chain
 
 Optimistic rollups instead rely on a fraud-proving scheme scheme to detect cases where transactions are not calculated correctly. After a rollup batch is submitted on Ethereum, there's a time window (called a challenge period) during which anyone can challenge the results of a rollup transaction by computing a [fraud proof](/glossary/#fraud-proof). 
 
-If the fraud proof succeeds, the rollup protocol rejects the invalid block and punishes the malicious operator. If the rollup batch remains unchallenged after the challenge period elapses, it is deemed valid and accepted on Ethereum. Others can continue to build on an unconfirmed rollup block, but with a caveat: blocks descending from a parent block will be rejected if the parent is later declared invalid. 
+If the fraud proof succeeds, the rollup protocol re-executes the transaction(s) and updates the rollup's state accordingly. The other effect of a successful fraud proof is that the sequencer responsible for including the incorrectly executed transaction in a block receives a penalty. 
+
+If the rollup batch remains unchallenged (i.e., all transactions are correctly executed) after the challenge period elapses, it is deemed valid and accepted on Ethereum. Others can continue to build on an unconfirmed rollup block, but with a caveat: transaction results will be reversed if based on an incorrectly executed transaction published previously. 
 
 ## How do optimistic rollups interact with Ethereum? {#optimistic-rollups-and-Ethereum} 
 
-An optimistic rollup is an off-chain protocol managed by a set of smart contracts deployed on the Ethereum network. Transactions executed off-chain are added to a record stored in the main rollup contract. Like the Ethereum blockchain, this transaction record is immutable and forms the "optimistic rollup chain."
+Optimistic rollups are [off-chain scaling solutions](/developers/docs/scaling/#off-chain-scaling) built to operate on top of Ethereum. Each optimistic rollup is managed by a set of smart contracts deployed on the Ethereum network. Optimistic rollups process transactions off the main Ethereum chain, but post off-chain transaction (in batches) to an on-chain rollup contract. Like the Ethereum blockchain, this transaction record is immutable and forms the "optimistic rollup chain."
 
-Although optimistic rollups exist as separate protocols, they are secured by the Ethereum network. Among other things, Ethereum guarantees the correctness of a rollup’s off-chain computation and the availability of data behind the computation. 
+The architecure of an optimistic rollup comprises the following parts: 
+
+**On-chain contracts**: The optimistic rollups's operation is controlled by smart contracts running on Ethereum. This includes contracts that store rollup blocks, monitor state updates on the rollup, and track user deposits. In this sense Ethereum serves as the base layer or "layer 1" for optimistic rollups.
+
+**Off-chain virtual machine (VM)**: Although contracts managing the optimistic rollup protocol run on Ethereum, the rollup protocol performs computation and state storage on another virtual machine separate from the [Ethereum Virtual Machine](/developers/docs/evm/). The off-chain VM is where applications live and state changes executed; it serves as the upper layer or "layer 2" for an optimistic rollup. 
+
+As optimistic rollups are designed to run programs either written or compiled for the EVM, the off-chain VM incorporates many EVM design specs. Additionally, fraud proofs computed on-chain allows the Ethereum network to enforce the validity of state changes computed in the off-chain VM. 
+
+Optimistic rollups are described as 'hybrid scaling solutions' because, while they exist as separate protocols, their security properties derive from Ethereum. Among other things, Ethereum guarantees the correctness of a rollup’s off-chain computation and the availability of data behind the computation. This makes optimistic rollups more secure than pure off-chain scaling protocols (e.g., [sidechains](/developers/docs/scaling/sidechains/)) that do not rely on Ethereum for security. 
+
+Optimistic rollups rely on the main Ethereum protocol for the following: 
 
 ### Data availability {data-availability}
 
@@ -61,7 +71,7 @@ Another role Ethereum plays in the context of optimistic rollups is that of a se
 
 Ethereum Mainnet provides a hub for optimistic rollups to verify fraud proofs and resolve disputes. Moreover, transactions conducted on the rollup are only final *after* the rollup block is accepted on Ethereum. Once a rollup transaction is committed to Ethereum’s base layer, it cannot be rolled back (except in the highly unlikely case of a chain reorganization).
 
-## How optimistic rollups work {#how-optimistic-rollups-work} 
+## How do optimistic rollups work? {#how-optimistic-rollups-work} 
 
 ### Transaction execution and aggregation {#transaction-execution-and-aggregation}
 
@@ -89,13 +99,13 @@ Here is [an example](https://etherscan.io/tx/0x9102bfce17c58b5fc1c974c24b6bb7a92
 
 ### State commitments {#state-commitments} 
 
-At any point in time, the optimistic rollup’s state—accounts, balances, contract code, etc—is organized as a [Merkle tree](/whitepaper/#merkle-trees) called a “state tree”. The root of this Merkle tree, referencing the rollup’s latest state, is hashed and stored in the rollup contract. Every state transition on the chain produces a new rollup state, which an operator commits to by computing a new state root.
+At any point in time, the optimistic rollup’s state (accounts, balances, contract code, etc.) is organized as a [Merkle tree](/whitepaper/#merkle-trees) called a “state tree”. The root of this Merkle tree (state root), which references the rollup’s latest state, is hashed and stored in the rollup contract. Every state transition on the chain produces a new rollup state, which an operator commits to by computing a new state root.
 
 The operator is required to submit both old state roots and new state roots when posting batches. If the old state root matches the existing state root in the on-chain contract, the latter is discarded and replaced with the new state root. 
 
 The rollup operator is also required to commit a Merkle root for the transaction batch itself. This allows anyone to prove the inclusion of a transaction in the batch (on L1) by presenting a [Merkle proof](/developers/tutorials/merkle-proofs-for-offline-data-integrity/). 
 
-State commitments, especially state roots, are neccessary for detecting fraud in optimistic rollups. They also make it easier to reverse an invalid block; the rollup contract only has to revert to the last valid block's state root to declare it the canonical block. 
+State commitments, especially state roots, are necessary for proving the correctness of state changes in an optimistic rollup. The rollup contract accepts new state roots from operators immediately they are posted, but can later delete invalid state roots to restore the rollup to its correct state. 
 
 ### Fraud proving {#fraud-proving}
 
@@ -143,7 +153,9 @@ Optimistic rollups are designed for interoperability with Ethereum Mainnet and a
 
 To use an optimistic rollup, users deposit ETH, ERC-20 tokens, and other accepted assets in the rollup’s [bridge](/developers/docs/bridges/) contract on L1. The bridge contract will relay the transaction to L2, where an equivalent amount of assets is minted and sendt to the user’s chosen address on the optimistic rollup. 
 
-Typically, user-generated transactions (like an L1 > L2 deposit) are queued until the sequencer re-submits them to the rollup contract. However, to preserve censorship resistance, users are allowed to submit a transaction to the rollup contract if it has been delayed past the maximum time allowed. 
+User-generated transactions (like an L1 > L2 deposit) are usually queued until the sequencer re-submits them to the rollup contract. However, to preserve censorship resistance, optimistic rollups allow users are allowed to submit a transaction directly to the on-chain rollup contract if it has been delayed past the maximum time allowed. 
+
+Some optimistic rollups adopt a more straightforward approach to prevent sequencers from censoring users. Here, a block is defined by all transactions submitted to the L1 contract since the previous block (e.g., deposits) in addition to the transactions processed on the rollup chain. If a sequencer ignores an L1 transaction, it will publish the (provably) wrong state root; therefore, sequencers cannot delay user-generated messages once posted on L1. 
 
 ##### Exiting the rollup {#exiting-the-rollup} 
 
@@ -251,4 +263,5 @@ Multiple implementations of Optimistic rollups exist that you can integrate into
 - [The Essential Guide to Arbitrum](https://newsletter.banklesshq.com/p/the-essential-guide-to-arbitrum)
 - [How does Optimism's Rollup really work?](https://research.paradigm.xyz/optimism)
 - [OVM Deep Dive](https://medium.com/ethereum-optimism/ovm-deep-dive-a300d1085f52)
+- [What is the Optimistic Virtual Machine?](https://www.alchemy.com/overviews/optimistic-virtual-machine)
 - [More on layer 2](/layer-2/)
--- a/src/content/developers/docs/scaling/optimistic-rollups/index.md
+++ b/src/content/developers/docs/scaling/optimistic-rollups/index.md
@@ -27,8 +27,6 @@ Optimistic rollups instead rely on a fraud-proving scheme scheme to detect cases
 
 If the fraud proof succeeds, the rollup protocol rejects the invalid block and punishes the malicious operator. If the rollup batch remains unchallenged after the challenge period elapses, it is deemed valid and accepted on Ethereum. Others can continue to build on an unconfirmed rollup block, but with a caveat: blocks descending from a parent block will be rejected if the parent is later declared invalid. 
 
-rollup will advance to a new state and the operator can start adding new blocks. 
-
 ## How do optimistic rollups interact with Ethereum? {#optimistic-rollups-and-Ethereum} 
 
 An optimistic rollup is an off-chain protocol managed by a set of smart contracts deployed on the Ethereum network. Transactions executed off-chain are added to a record stored in the main rollup contract. Like the Ethereum blockchain, this transaction record is immutable and forms the "optimistic rollup chain."
@@ -183,15 +181,15 @@ Finally, we should note that L2 > L1 message calls between contracts need to acc
 
 ## How do optimistic rollup fees work? {#how-do-optimistic-rollup-fees-work}
 
-Many optimistic rollups use a “gas price” scheme, much like Ethereum, to denote how much users pay per transaction. How much users pay for transactions on optimistic rollups depends on the following components:
+Optimistic rollups use a gas fee scheme, much like Ethereum, to denote how much users pay per transaction. Fees charged on optimistic rollups depends on the following components:
 
-1. **State write**: There is a fixed cost for writing to Ethereum’s state (i.e., submitting a transaction on the Ethereum blockchain). Optimistic rollups reduce this cost by batching transactions and spreading fixed costs across multiple users.  
+1. **State write**: Optimistic rollups publish transaction data and block headers (consisting of the previous block header hash, state root, batch root) to Ethereum as `calldata`. The minimum cost of an Ethereum transaction is 21,000 gas. Optimistic rollups can reduce the cost of writing the transaction to L1 by batching multiple transactions in a single block (which amortizes the 21k gas over multiple user transactions). 
 
-2. **Data publication**: Optimistic rollups publish transaction data and block headers (consisting of the previous block header hash, state root, batch root) to Ethereum as `calldata`. Calldata costs are currently governed by [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559), which stipulates a cost of 16 gas for non-zero bytes and 4 gas for zero bytes of calldata, respectively. The cost paid on each transaction is influenced by how much calldata needs to be posted on-chain for it. 
+Beyond the base transaction fee, the cost of each state write depends on the size of `calldata` posted to L1. Calldata costs are currently governed by [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559), which stipulates a cost of 16 gas for non-zero bytes and 4 gas for zero bytes of calldata, respectively. To reduce user fees, rollup operators compresss transactions to reduce the number of `calldata` bytes published on Ethereum. 
 
-3. **L2 operator fees**: This is the amount paid to the rollup operator as compensation for computational costs incurred in processing transactions, much like miner fees on Ethereum. 
+3. **L2 operator fees**: This is the amount paid to the rollup nodes as compensation for computational costs incurred in processing transactions, much like miner fees on Ethereum. Rollup nodes charge lower transaction fees since L2s have higher processing capacities and aren't faced with the network congestions that force miners on Ethereum to prioritize transactions with higher fees. 
 
-Optimistic rollups apply several mechanisms to reducing fees for users, including reducing `calldata` published per transaction. You can see [here](https://l2fees.info/) for a real-time overview of how it costs to use Ethereum optimistic rollups. 
+Optimistic rollups apply several mechanisms to reducing fees for users, including batching transactions and compressing `calldata` to reduce data publication costs. You can check the [L2 fee tracker](https://l2fees.info/) for a real-time overview of how much it costs to use Ethereum-based optimistic rollups. 
 
 ## How do optimistic rollups scale Ethereum? {#scaling-ethereum-with-optimistic-rollups}
 
@@ -228,7 +226,7 @@ The introduction of [data sharding](/upgrades/shard-chains/) on Ethereum is expe
 | ---------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------- |
 |Offers massive improvements in scalability without sacrificing security or trustlessness.    | Delays in transaction finality due to potential fraud challenges.                                                                                                     |
 | Transaction data is stored on the layer 1 chain, improving transparency, security, censorship-resistance, and decentralization. | Centralized rollup operators (sequencers) can influence transaction ordering.                                                                   | 
-| Fraud proving guarantees trustless finality and allows honest minorities to secure the chain.                     | A malicious operator can steal funds by posting invalid blocks and state commitments.                                                                              | 
+| Fraud proving guarantees trustless finality and allows honest minorities to secure the chain.                     | If there are no honest nodes a malicious operator can steal funds by posting invalid blocks and state commitments.                                                    | 
 | Computing fraud proofs is open to regular L2 node, unlike validity proofs (used in ZK-rollups) that require special hardware.  | Security model relies on at least one honest node executing rollup transactions and submitting fraud proofs to challenge invalid state transitions.                  | 
 | Rollups benefit from "trustless liveness" (anyone can force the chain to advance by executing transactions and posting assertions)      | Users must wait for the one-week challenge period to expire before withdrawing funds back to Ethereum.                                                              |
 | Optimistic rollups rely on well-designed cryptoeconomic incentives to increase security on the chain.                            | Rollups must post all transaction data on-chain, which can increase costs. 
--- a/src/content/developers/docs/scaling/optimistic-rollups/index.md
+++ b/src/content/developers/docs/scaling/optimistic-rollups/index.md
@@ -5,7 +5,7 @@ lang: en
 sidebar: true
 ---
 
-Optimistic rollups sit in parallel to the main Ethereum chain on layer 2. They can offer improvements in scalability because they reduce computation on Ethereum’s base layer. After processing transactions off-chain, optimistic rollups publish the results on Ethereum, but don't re-execute the transactions on Mainnet. 
+Optimistic rollups are off-chain protocols designed to extend throughput of Ethereum's base layer. They reduce computation on the main Ethereum chain by processing transactions off-chain, offering significant improvements in processing speeds. Unlike other scaling solutions, such as [sidechains](/developers/docs/scaling/sidechains/), optimistic rollups derive security from Mainnet by publishing transaction results on-chain. 
 
 As computation is the slow, expensive part of using Ethereum, optimistic rollups can offer up to 10-100x improvements in scalability. Optimistic rollups also write transactions to Ethereum as `calldata`, reducing gas costs for users. 
 
@@ -23,7 +23,7 @@ Optimistic rollup operators bundle multiple off-chain transactions together in l
 
 Optimistic rollups are considered “optimistic” because they assume off-chain transactions are valid and don't publish proofs of validity for transaction batches posted on-chain. This separates optimistic rollups from [zero-knowledge rollups](/developers/docs/scaling/zk-rollups) that publish cryptographic [proofs of validity](/glossary/#validity-proof) for off-chain transactions. 
 
-Optimistic rollups instead rely on a fraud-proving scheme to detect malicious transactions. After a batch is submitted on Ethereum, there's a time window (called a challenge period) during which anyone can challenge the validity of the rollup block by computing a [fraud proof](/glossary/#fraud-proof). 
+Optimistic rollups instead rely on a fraud-proving scheme to detect malicious transactions. After a rollup batch is submitted on Ethereum, there's a time window (called a challenge period) during which anyone can challenge the validity of a rollup transaction by computing a [fraud proof](/glossary/#fraud-proof). 
 
 If the fraud proof succeeds, the rollup protocol reverses the invalid block and punishes the malicious operator. If the rollup batch remains unchallenged after the challenge period elapses, it is deemed valid and accepted on Ethereum. The rollup will advance to a new state and the operator can start adding new blocks. 
 
@@ -31,13 +31,13 @@ If the fraud proof succeeds, the rollup protocol reverses the invalid block and 
 
 An optimistic rollup is an off-chain protocol managed by a set of smart contracts deployed on the Ethereum network. Transactions executed off-chain are added to a record stored in the main rollup contract. Like the Ethereum blockchain, this transaction record is immutable and forms the "optimistic rollup chain."
 
-Although optimistic rollups exist as separate protocols, they derive security directly from Ethereum Mainnet. Among other things, Ethereum guarantees the correctness of a rollup’s off-chain computation and the availability of data behind the computation. 
+Although optimistic rollups exist as separate protocols, they are secured by the Ethereum network. Among other things, Ethereum guarantees the correctness of a rollup’s off-chain computation and the availability of data behind the computation. 
 
 ### Data availability {data-availability}
 
 As mentioned, optimistic rollups post transaction data to Ethereum as `calldata`. Since the rollup chain's execution is based on submitted transactions, anyone can use this information—anchored on Ethereum’s base layer—to execute the rollup’s state and verify the correctness of state transitions. . 
 
-Data availability is critical because, without access to state data, challengers cannot construct fraud proofs to dispute invalid rollup operations. With Ethereum providing data availability, the risk of rollup operators getting away with malicious acts (e.g., submitting invalid blocks) reduces. 
+Data availability is critical because without access to state data, challengers cannot construct fraud proofs to dispute invalid rollup operations. With Ethereum providing data availability, the risk of rollup operators getting away with malicious acts (e.g., submitting invalid blocks) reduces. 
 
 Ethereum’s role as a data availability layer for optimistic rollups is expected to improve with the introduction of [shard chains](/upgrades/shard-chains/). This is important because the processing capacity of rollups is limited by the data throughput of the L1 chain. By splitting data storage across multiple shards, Ethereum can increase the space available for optimistic rollups to dump data. 
 
@@ -67,7 +67,7 @@ Ethereum Mainnet provides a hub for optimistic rollups to verify fraud proofs an
 
 Users submit transactions to “operators”, nodes responsible for processing transactions on the optimistic rollup. Also known as a “validator” or “aggregator”, the operator aggregates transactions, compresses the underlying data, and publishes the block on Ethereum. 
 
-Although anyone can become a validator, nodes must provide a bond before producing blocks, much like a [proof-of-stake system](/developers/docs/consensus-mechanisms/pos/). This bond can be slashed if the validator posts an invalid block, which incentivizes validators to act honestly. 
+Although anyone can become a validator, L2 nodes must provide a bond before producing blocks, much like a [proof-of-stake system](/developers/docs/consensus-mechanisms/pos/). This bond can be slashed if the validator posts an invalid block or builds on an old-but-invalid block (even if their block is valid). This way optimistic rollups utilize cryptoeconomic incentives to ensure validators act honestly. 
 
 Other validators on the L2 chain are expected to execute the submitted transactions using their copy of the rollup’s state. If a validator’s final state is different from the operator’s proposed state, they can start a challenge and compute a fraud proof. 
 
@@ -81,7 +81,7 @@ As mentioned, the operator of an optimistic rollup bundles off-chain transaction
 
 Calldata is a non-modifiable, non-persistent area in a smart contract that behaves like [memory](/developers/docs/smart-contracts/anatomy/#memory). While calldata persists on-chain as part of the blockchain's [history logs](https://docs.soliditylang.org/en/latest/introduction-to-smart-contracts.html?highlight=memory#logs), it is not stored as a part of Ethereum's state. Because calldata does not touch any part of Ethereum's state, it is cheaper to use compared to `memory` or `storage`.
 
-The `calldata` keyword is also used in Solidity to pass data to a smart contract function at execution time. Calldata identifies the function being called during a transaction and holds inputs to the function in the form of an arbitrary sequence of bytes.
+The `calldata` keyword is also used in Solidity to pass arguments to a smart contract function at execution time. Calldata identifies the function being called during a transaction and holds inputs to the function in the form of an arbitrary sequence of bytes.
 
 In the context of optimistic rollups, `calldata` is used to send compressed transaction data to the on-chain contract. The rollup operator adds a new batch by calling the required function in the rollup contract and passing the compressed data as function arguments. 
 
@@ -91,11 +91,13 @@ Using `calldata` reduces user fees since most costs that rollups incur come from
 
 ### State commitments {#state-commitments} 
 
-At any point in time, the optimistic rollup’s state is organized as a [Merkle tree](/whitepaper/#merkle-trees) called a “state tree”. The root of this Merkle tree, referencing the rollup’s latest state, is hashed and stored in the rollup contract. Each new block added to the rollup chain produces a new rollup state, which an operator commits to by computing a new state root.
+At any point in time, the optimistic rollup’s state—accounts, balances, contract code, etc—is organized as a [Merkle tree](/whitepaper/#merkle-trees) called a “state tree”. The root of this Merkle tree, referencing the rollup’s latest state, is hashed and stored in the rollup contract. Every state transition on the chain produces a new rollup state, which an operator commits to by computing a new state root.
 
 The operator is required to submit both old state roots and new state roots when posting batches. If the old state root matches the existing state root in the on-chain contract, the latter is discarded and replaced with the new state root. 
 
 The rollup operator is also required to commit a Merkle root for the transaction batch itself. This allows anyone to prove the inclusion of a transaction in the batch (on L1) by presenting a [Merkle proof](/developers/tutorials/merkle-proofs-for-offline-data-integrity/). 
+
+State commitments, especially state roots, are neccessary for detecting fraud in optimistic rollups. They also make it easier to reverse an invalid block; the rollup contract only has to revert to the last valid block's state root to declare it the canonical block. 
 
 ### Fraud proving {#fraud-proving}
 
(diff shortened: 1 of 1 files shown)


src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,36 +1,213 @@
 ---
 title: Optimistic Rollups
-description: Introduction to optimistic rollups
+description: An introduction to optimistic rollups, a scaling solution used by the Ethereum community. 
 lang: en
 sidebar: true
 ---
 
+Optimistic rollups sit in parallel to the main Ethereum chain on layer 2. They can offer improvements in scalability because they reduce computation on Ethereum’s base layer. After processing transactions off-chain, optimistic rollups publish the results on Ethereum, but don't re-execute the transactions on Mainnet.
```
samajammin, line 8: Just a general thought/note - I like framing L2s as "extending" the base layer, e.g. https://twitter.com/toghrulmaharram/status/1534359799311253505?s=20&t=VxDjdiP67hBtXKgvB-oOkQ

emmanuel-awosika: > Just a general thought/note - I like framing L2s as "extending" the base layer, e.g. https://twitter.com/toghrulmaharram/status/1534359799311253505?s=20&t=VxDjdiP67hBtXKgvB-oOkQ

That makes sense. I'll edit it to reflect the suggestion. The text was in the original page, so I wanted to leave it (and just flesh it out a bit), but I guess better clarifying rollups wouldn't hurt.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+As mentioned, optimistic rollups post transaction data to Ethereum as `calldata`. Since the rollup chain's execution is based on submitted transactions, anyone can use this information—anchored on Ethereum’s base layer—to execute the rollup’s state and verify the correctness of state transitions. . 
+
+Data availability is critical because without access to state data, challengers cannot construct fraud proofs to dispute invalid rollup operations. With Ethereum providing data availability, the risk of rollup operators getting away with malicious acts (e.g., submitting invalid blocks) reduces. 
+
+### Censorship resistance {#censorship-resistance}
+
+Optimistic rollups also rely on Ethereum for censorship resistance. In an optimistic rollup a centralized entity (the operator) is responsible for processing transactions and submitting rollup blocks to Ethereum. This has some implications: 
+
+- Rollup operators can censor users by going offline, refusing to produce blocks, or submit user transactions to L1.
```
qbzzt, line 46: How about "Rollup operators can censor users by going offline completely, or by refusing to either produce blocks that include certain transactions in them"?

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+- Rollup operators can censor users by going offline, refusing to produce blocks, or submit user transactions to L1. 
+
+- Rollup operators can prevent users from withdrawing funds deposited in the rollup contract by withholding state data necessary to Merkle proofs of ownership. Withholding state data can also conceal the rollup’s state from users and prevent them from interacting with the rollup.  
+
+Optimistic rollups solve this problem by forcing operators to publish data associated with state updates on Ethereum. Publishing rollup data on-chain has the following benefits:
+
+- If an optimistic rollup operator goes offline or stops producing transaction batches, another node can use available data to reproduce the rollup’s last state and continue block production. 
+
+- Users can use transaction data to create Merkle proofs proving ownership of funds and withdraw their assets from the rollup.
```
qbzzt, line 54: After this I should you should add this point:

> Users can also submit their transactions on L1 instead of to the sequencer, in which case the sequencer has to include the transaction within a certain time limit to continue to produce valid blocks.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+Although optimistic rollups exist as separate protocols, they are secured by the Ethereum network. Among other things, Ethereum guarantees the correctness of a rollup’s off-chain computation and the availability of data behind the computation. 
+
+### Data availability {data-availability}
+
+As mentioned, optimistic rollups post transaction data to Ethereum as `calldata`. Since the rollup chain's execution is based on submitted transactions, anyone can use this information—anchored on Ethereum’s base layer—to execute the rollup’s state and verify the correctness of state transitions. . 
+
+Data availability is critical because without access to state data, challengers cannot construct fraud proofs to dispute invalid rollup operations. With Ethereum providing data availability, the risk of rollup operators getting away with malicious acts (e.g., submitting invalid blocks) reduces.
```
qbzzt, line 40: Maybe instead of "reduces" use "is reduced"?

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+Optimistic rollups are considered “optimistic” because they assume off-chain transactions are valid and don't publish proofs of validity for transaction batches posted on-chain. This separates optimistic rollups from [zero-knowledge rollups](/developers/docs/scaling/zk-rollups) that publish cryptographic [proofs of validity](/glossary/#validity-proof) for off-chain transactions. 
+
+Optimistic rollups instead rely on a fraud-proving scheme to detect malicious transactions. After a rollup batch is submitted on Ethereum, there's a time window (called a challenge period) during which anyone can challenge the validity of a rollup transaction by computing a [fraud proof](/glossary/#fraud-proof). 
+
+If the fraud proof succeeds, the rollup protocol reverses the invalid block and punishes the malicious operator. If the rollup batch remains unchallenged after the challenge period elapses, it is deemed valid and accepted on Ethereum. The rollup will advance to a new state and the operator can start adding new blocks.
```
qbzzt, line 28: "If the fraud proof succeeds, the rollup protocol reverses the invalid block and punishes the malicious operator."

The invalid block doesn't get reversed. Instead, the transactions are executed correctly and the state is updated to the correct value.

Also, the rollup doesn't wait the challenge period (typically seven days) until it adds new blocks. It adds them immediately, and if there is a successful challenge then the results of all those new blocks have to be recalculated.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+Optimistic rollup operators bundle multiple off-chain transactions together in large batches before submitting to Ethereum. This approach enables spreading fixed costs across multiple transactions in each batch, reducing fees for end-users. Optimistic rollups also use compression techniques to reduce the amount of data posted on Ethereum. 
+
+Optimistic rollups are considered “optimistic” because they assume off-chain transactions are valid and don't publish proofs of validity for transaction batches posted on-chain. This separates optimistic rollups from [zero-knowledge rollups](/developers/docs/scaling/zk-rollups) that publish cryptographic [proofs of validity](/glossary/#validity-proof) for off-chain transactions. 
+
+Optimistic rollups instead rely on a fraud-proving scheme to detect malicious transactions. After a rollup batch is submitted on Ethereum, there's a time window (called a challenge period) during which anyone can challenge the validity of a rollup transaction by computing a [fraud proof](/glossary/#fraud-proof).
```
qbzzt, line 26: It's not the transactions that are malicious, but the results. Maybe "...scheme to detect cases where transactions are not calculated correctly"? This is explained in greater detail here: https://help.optimism.io/hc/en-us/articles/4416604320027-What-happens-when-a-transaction-is-challenged-successfully-

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+Ethereum Mainnet provides a hub for optimistic rollups to verify fraud proofs and resolve disputes. Moreover, transactions conducted on the rollup are only final *after* the rollup block is accepted on Ethereum. Once a rollup transaction is committed to Ethereum’s base layer, it cannot be rolled back (except in the highly unlikely case of a chain reorganization).
+
+## How optimistic rollups work {#how-optimistic-rollups-work} 
 
-Optimistic rollups sit in parallel to the main Ethereum chain on layer 2. They can offer improvements in scalability because they don't do any computation by default. Instead, after a transaction, they propose the new state to Mainnet or "notarise" the transaction.
+### Transaction execution and aggregation {#transaction-execution-and-aggregation}
 
-With Optimistic rollups, transactions are written to the main Ethereum chain as `calldata`, optimising them further by reducing the gas cost.
+Users submit transactions to “operators”, nodes responsible for processing transactions on the optimistic rollup. Also known as a “validator” or “aggregator”, the operator aggregates transactions, compresses the underlying data, and publishes the block on Ethereum.
```
qbzzt, line 66: This makes it sound there are multiple transaction sources at the same time, which is incorrect (at least for Optimism, I don't know Arbitrum well enough).

How about

> Users submit their transactions to nodes that forward them to the sequencer (there is only one sequencer at a time). The sequencer then aggregates the transactions, compresses them, and publishes the block on Ethereum.

A node does not need to be a validator, and a validator does not need to provide node services.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+#### Submitting rollup blocks to Ethereum {#submitting-blocks-to-ethereum} 
+
+As mentioned, the operator of an optimistic rollup bundles off-chain transactions into a batch and sends it to Ethereum for notarization. This process involves compressing transaction-related data and publishing it on Ethereum as `calldata`. 
+
+Calldata is a non-modifiable, non-persistent area in a smart contract that behaves like [memory](/developers/docs/smart-contracts/anatomy/#memory). While calldata persists on-chain as part of the blockchain's [history logs](https://docs.soliditylang.org/en/latest/introduction-to-smart-contracts.html?highlight=memory#logs), it is not stored as a part of Ethereum's state. Because calldata does not touch any part of Ethereum's state, it is cheaper to use compared to `memory` or `storage`.
```
qbzzt, line 80: It is cheaper compared to `storage`. But I wouldn't compare it to `memory`, because `memory` also isn't part of the state.

The issue is that the only way to provide any information to the blockchain is through `calldata`, so that cost has to be paid anyway. The other costs are the result of processing which we don't need.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+The `calldata` keyword is also used in Solidity to pass arguments to a smart contract function at execution time. Calldata identifies the function being called during a transaction and holds inputs to the function in the form of an arbitrary sequence of bytes.
+
+In the context of optimistic rollups, `calldata` is used to send compressed transaction data to the on-chain contract. The rollup operator adds a new batch by calling the required function in the rollup contract and passing the compressed data as function arguments. Using `calldata` reduces user fees since most costs that rollups incur come from storing data on-chain.
```
qbzzt, line 84: It doesn't have to be a rollup contract, BTW. In our (Optimism's) next release it will be just an address, so there won't be any processing done.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+Finally, we should note that L2 > L1 message calls between contracts need to account for delays (L1 > L2 calls are typically executed after some minutes). This is because messages sent to Mainnet from the optimistic rollup cannot be executed until the challenge window expires.
+
+## How do optimistic rollup fees work? {#how-do-optimistic-rollup-fees-work}
+
+Many optimistic rollups use a “gas price” scheme, much like Ethereum, to denote how much users pay per transaction. How much users pay for transactions on optimistic rollups depends on the following components:
+
+1. **State write**: There is a fixed cost for writing to Ethereum’s state (i.e., submitting a transaction on the Ethereum blockchain). Optimistic rollups reduce this cost by batching transactions and spreading fixed costs across multiple users.
```
qbzzt, line 200: It's a cost, but not a fixed one. It depends on the L1 gas price.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
-| All transaction data is stored on the layer 1 chain, meaning it's secure and decentralized.                      | An operator can influence transaction ordering.                             |
+|Offers massive improvements in scalability without sacrificing security or trustlessness.    | Delays in transaction finality due to potential fraud challenges.                                                                                                     |
+| Transaction data is stored on the layer 1 chain, improving transparency, security, censorship-resistance, and decentralization. | Centralized rollup operators (sequencers) can influence transaction ordering.                                                                   | 
+| Fraud proving guarantees trustless finality and allows honest minorities to secure the chain.                     | A malicious operator can steal funds by posting invalid blocks and state commitments.                                                                              |
```
qbzzt, line 243: If there are no honest nodes a malicious operator can...

It could happen, but it's extremely unlikely.

qbzzt (commented):
Here are more comments, from the relevant to the pedantic.

src/content/developers/docs/scaling/optimistic-rollups/index.md
emmanuel-awosika: Thanks for pointing out that out. I changed the sentence to read: " Because calldata does not touch any part of Ethereum's state, it is cheaper for storing data on-chain." Does that work?

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+The asserter is required to provide a “one-step proof” verifying the validity of the disputed single-step computation. If the asserter fails to provide the one-step proof, or the L1 verifier deems the proof invalid, they lose the challenge. 
+
+Some notes about this type of fraud proof: 
+
+1. Multi-round interactive fraud proving is considered efficient because it minimizes the work the L1 chain must do in dispute arbitration. Instead of replaying the entire transaction, Ethereum only needs to re-execute one step in the rollup's execution. 
+
+2. Bisection protocols reduce the amount of data posted on-chain (no need to publish state commits for every transaction). Also, optimistic rollup transactions don't have constrained by Ethereum's gas limit. Conversely, optimistic rollups re-executing transactions must make sure an L2 transaction has a lower gas limit to emulate its execution within a single Ethereum transaction.  
+
+##### Why fraud proofs matter for optimistic rollups {#fraud-proof-benefits}
```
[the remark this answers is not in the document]
emmanuel-awosika: Fixed this now.

src/content/developers/docs/scaling/optimistic-rollups/index.md
emmanuel-awosika: By "fixed costs", I mean the operator has to pay a fee for sending an on-chain transaction, regardless of how big or small the batch is. How much they pay for calldata is highly variable from what I learned. For context, here's an [article](https://forum.celestia.org/t/ethereum-rollup-call-data-pricing-analysis/141) I read on "variable" vs "fixed " costs for rollups.

src/content/developers/docs/scaling/optimistic-rollups/index.md
emmanuel-awosika: Sure, but I think it was still important to point out. If you notice, I mentioned in the next line: "Security model relies on at least one honest node executing rollup transactions and submitting fraud proofs to challenge invalid state transitions."

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+To avoid waiting a week before withdrawing funds to Ethereum, optimistic rollup users can employ a **liquidity provider** (LP). A liquidity provider assumes ownership of a pending L2 withdrawal and pays the user on L1 (in exchange for a fee). 
+
+Liquidity providers can check the validity of the user’s withdrawal request (by executing the chain themselves) before releasing funds. This way they have assurances that the transaction will be confirmed eventually (trustless finality). 
+
+A user can have their exit from the rollup expedited by submitting the withdrawal transaction to an aggregator/sequencer. They can also send it directly to the L1 contract, which will delay the transaction, but protects against censorship. 
+
+#### 2. EVM compatibility {#evm-compatibility} 
+
+For developers, the advantage of optimistic rollups is their compatibility—or, better still, equivalence—with the [Ethereum Virtual Machine (EVM)](/developers/docs/evm/). EVM-compatible rollups rollups comply with specifications outlined in the [Ethereum Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) and support the EVM at the bytecode level.
```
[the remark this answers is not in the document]
emmanuel-awosika: Fixed this.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+After the withdrawal request is initiated on the L2 rollup, the transaction is included in the next batch, while the user’s assets on the rollup are burned. Once the batch is published on Ethereum, the user can compute a Merkle proof verifying the inclusion of their exit transaction in the block. Then it is a matter of waiting through the delay period to finalize the transaction on L1 and withdraw funds to Mainnet. 
+
+To avoid waiting a week before withdrawing funds to Ethereum, optimistic rollup users can employ a **liquidity provider** (LP). A liquidity provider assumes ownership of a pending L2 withdrawal and pays the user on L1 (in exchange for a fee). 
+
+Liquidity providers can check the validity of the user’s withdrawal request (by executing the chain themselves) before releasing funds. This way they have assurances that the transaction will be confirmed eventually (trustless finality). 
+
+A user can have their exit from the rollup expedited by submitting the withdrawal transaction to an aggregator/sequencer. They can also send it directly to the L1 contract, which will delay the transaction, but protects against censorship.
```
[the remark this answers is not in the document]
emmanuel-awosika: Fixed this now.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+The operator is required to submit both old state roots and new state roots when posting batches. If the old state root matches the existing state root in the on-chain contract, the latter is discarded and replaced with the new state root. 
+
+The rollup operator is also required to commit a Merkle root for the transaction batch itself. This allows anyone to prove the inclusion of a transaction in the batch (on L1) by presenting a [Merkle proof](/developers/tutorials/merkle-proofs-for-offline-data-integrity/). 
+
+State commitments, especially state roots, are neccessary for detecting fraud in optimistic rollups. They also make it easier to reverse an invalid block; the rollup contract only has to revert to the last valid block's state root to declare it the canonical block.
```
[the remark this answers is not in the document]
emmanuel-awosika: But the docs say the state root for the invalid block (and any others that built on it) will get deleted from the State Commitment Chain. Isn't that the same as reversing the block (even if the transactions are re-executed)?

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+Optimistic rollups are designed for interoperability with Ethereum Mainnet and allow users to pass messages and arbitrary data between L1 and L2. They are also compatible with the EVM, so you can port existing [dapps](/developers/docs/dapps/) to rollups or create new dapps using Ethereum development tools. 
+
+#### 1. Asset movement {#asset-movement} 
+
+##### Entering the rollup {#entering-the-rollup} 
+
+To use an optimistic rollup, users deposit ETH, ERC-20 tokens, and other accepted assets in the rollup’s [bridge](/developers/docs/bridges/) contract on L1. The bridge contract will relay the transaction to L2, where an equivalent amount of assets is minted and sendt to the user’s chosen address on the optimistic rollup. 
+
+Typically, user-generated transactions (like an L1 > L2 deposit) are queued until the sequencer re-submits them to the rollup contract. However, to preserve censorship resistance, users are allowed to submit a transaction to the rollup contract if it has been delayed past the maximum time allowed.
```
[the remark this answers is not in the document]
emmanuel-awosika: Not sure I totally understand this part. I got this from Paradigm's [article](https://research.paradigm.xyz/optimism#data-availability-batches) on Optimism's rollup:

> Transactions are submitted to the CTC either by the sequencer, a prioritized party who can insert transactions into the chain, or via a first-in-first-out queue which feeds into the CTC. To preserve L1’s censorship resistance guarantees, anybody can submit transactions to this queue, forcing them to be included into the CTC after a delay.

And this is from the [Arbitrum docs](https://developer.offchainlabs.com/docs/inside_arbitrum#inboxes-fast-and-slow):

>Anyone else can submit a message, but messages submitted by non-sequencer nodes will be put into the "delayed inbox" queue, which is managed by an L1 Ethereum contract. Messages in the delayed inbox queue will wait there until the sequencer chooses to "release" them into the main inbox, where they will be added to the end of the inbox. A well-behaved sequencer will typically release delayed messages after about ten minutes, for reasons explained below. Alternatively, if a message has been in the delayed inbox queue for longer than a maximum delay interval (currently 24 hours) then anyone can force it to be promoted into the main inbox. (This ensures that the sequencer can only delay messages but can't censor them.)

Perhaps you can clarify what part of the text needs to change?

src/content/developers/docs/scaling/optimistic-rollups/index.md
emmanuel-awosika: So, I got this from the Paradigm research piece I quoted earlier: 

> We now can introduce the [State Commitment Chain](https://github.com/ethereum-optimism/contracts-v2/blob/0ad4dcfdef11ef87e278a8159de8414c8e329ba1/contracts/optimistic-ethereum/OVM/chain/OVM_StateCommitmentChain.sol) (SCC). The SCC contains a list of state roots, which, in the optimistic case, correspond to the result of applying each transaction in the CTC against the previous state. If this is not the case, then the fraud verification process allows the invalid state root, and all following it, to be deleted, so that the correct state root for those transactions may be proposed.

That said, I want to ask: if the invalid block is not rejected or reversed, what happens to its position in the chain? 

As to the concluding part, I changed it to read: 

> Others can continue to build on an unconfirmed rollup block, but with a caveat: blocks descending from a parent block will be rejected if the parent is later declared invalid.

src/content/developers/docs/scaling/optimistic-rollups/index.md
qbzzt: I agree it needs to be pointed out. I'm just suggesting that 

> If there are no honest nodes a malicious operator can steal funds by posting invalid blocks and state commitments

is more accurate than 

> A malicious operator can steal funds by posting invalid blocks and state commitments

The text here makes it seem like it's unavoidable, at least until they read the next line.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+## How do optimistic rollup fees work? {#how-do-optimistic-rollup-fees-work}
+
+Many optimistic rollups use a “gas price” scheme, much like Ethereum, to denote how much users pay per transaction. How much users pay for transactions on optimistic rollups depends on the following components:
+
+1. **State write**: There is a fixed cost for writing to Ethereum’s state (i.e., submitting a transaction on the Ethereum blockchain). Optimistic rollups reduce this cost by batching transactions and spreading fixed costs across multiple users.  
+
+2. **Data publication**: As mentioned, `calldata` must be published to Ethereum for every transaction. Calldata costs are currently governed by [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559), which stipulates a cost of 16 gas for non-zero bytes and 4 gas for zero bytes of calldata, respectively. The cost paid on each transaction is influenced by how much calldata needs to be posted on-chain for it.
```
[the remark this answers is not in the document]
qbzzt: That's great, thank you.

src/content/developers/docs/scaling/optimistic-rollups/index.md
qbzzt: Great I can't resolve comments because I don't have permission, but I'll hide the ones that are no longer relevant.

src/content/developers/docs/scaling/optimistic-rollups/index.md
qbzzt: I'm tired, so I'll do the simple comments now and deal with the complex ones later this weekend, OK?

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,254 @@
...
+
+### Fraud proving {#fraud-proving}
+
+As explained, optimistic rollups allow anyone to publish blocks without providing proofs of validity. However, to ensure the chain remains safe, optimistic rollups specify a time window during which anyone can dispute a state transition. Hence, rollup blocks are called “assertions” since anyone can dispute their validity. 
+
+If someone disputes an assertion, then the rollup protocol will initiate the fraud proof computation. Every type of fraud proof is interactive—someone must post an assertion before another person can challenge it. The difference lies in how many rounds of interaction are required to compute the fraud proof. 
+
+#### Single-round interactive proving {#single-round-interactive-proving}
+
+Single-round interactive proving schemes replay the disputed transaction on Ethereum to detect fraud. The process starts with the challenger posting a bond and initiating the fraud proof computation, after which they provide relevant information, such as the old state root, their version of the new state root, the disputed transaction, and state data.
```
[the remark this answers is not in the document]
qbzzt: That's good, thank you.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,242 @@
...
+
+Optimistic rollups instead rely on a fraud-proving scheme scheme to detect cases where transactions are not calculated correctly. After a rollup batch is submitted on Ethereum, there's a time window (called a challenge period) during which anyone can challenge the results of a rollup transaction by computing a [fraud proof](/glossary/#fraud-proof). 
+
+If the fraud proof succeeds, the rollup protocol rejects the invalid block and punishes the malicious operator. If the rollup batch remains unchallenged after the challenge period elapses, it is deemed valid and accepted on Ethereum. Others can continue to build on an unconfirmed rollup block, but with a caveat: blocks descending from a parent block will be rejected if the parent is later declared invalid. 
+
+rollup will advance to a new state and the operator can start adding new blocks.
```
qbzzt, line 30: Delete this line? It is the last survivor of a dead paragraph, IIRC.

src/content/developers/docs/scaling/optimistic-rollups/index.md
Nana12345678910: Yes I bought some eth last year to invest on MetaMask

Nana12345678910: How do I recover my eth

src/content/developers/docs/scaling/optimistic-rollups/index.md
qbzzt: > if the invalid block is not rejected or reversed, what happens to its position in the chain?

I'm not sure about Arbitrum, but in Optimism the block just contains transactions, and it cannot be invalid. What can be invalid is the post-block state, which is written elsewhere (in the state commitment chain, SCC). The SCC is reverted, but then it needs to be recalculated with the correct result for that block and then subsequent blocks. The transactions themselves are still part of the chain.

src/content/developers/docs/scaling/optimistic-rollups/index.md
Nana12345678910: I think they rewrote the code to 256

src/content/developers/docs/scaling/optimistic-rollups/index.md
qbzzt: Yes.

src/content/developers/docs/scaling/optimistic-rollups/index.md
qbzzt: I think it's OK to leave it as is for now.

src/content/developers/docs/scaling/optimistic-rollups/index.md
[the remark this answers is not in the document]
qbzzt: In Optimism the block is just a place where you put some background information that could be required to execute the transactions, followed by the transactions themselves.

src/content/developers/docs/scaling/optimistic-rollups/index.md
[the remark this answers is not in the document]
qbzzt: I submitted a question about this to our protocol developers. I'll answer you when they answer me.

src/content/developers/docs/scaling/optimistic-rollups/index.md
qbzzt: How about this?

1. **State write**: The minimum cost of an Ethereum transaction is 21k gas plus either 16 or 4 gas per calldata byte (bytes with zeros cost 4 gas, all other values cost 16). Optimistic rollups can reduce the cost of writing the transaction to L1 by batching multiple transactions in a single block (which amortizes the 21k gas over multiple user transactions). They can also compress the transactions to reduce the number of calldata bytes.

emmanuel-awosika: Thanks for the suggestion. I'd mentioned calldata costs separately from the state write, but combined them based on your suggestion. Here's the new text:

> Optimistic rollups publish transaction data and block headers (consisting of the previous block header hash, state root, batch root) to Ethereum as `calldata`. The minimum cost of an Ethereum transaction is 21,000 gas. Optimistic rollups can reduce the cost of writing the transaction to L1 by batching multiple transactions in a single block (which amortizes the 21k gas over multiple user transactions). 

> Beyond the base transaction fee, the cost of each state write depends on the size of `calldata` posted to L1. Calldata costs are currently governed by [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559), which stipulates a cost of 16 gas for non-zero bytes and 4 gas for zero bytes of calldata, respectively. To reduce user fees, rollup operators compresss transactions to reduce the number of `calldata` bytes published on Ethereum.

src/content/developers/docs/scaling/optimistic-rollups/index.md
emmanuel-awosika: That makes sense. I just changed the text to what you suggested.

src/content/developers/docs/scaling/optimistic-rollups/index.md
emmanuel-awosika: Fixed this.

src/content/developers/docs/scaling/optimistic-rollups/index.md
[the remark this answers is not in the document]
emmanuel-awosika: Alright, then. LMK when you get a reply.

src/content/developers/docs/scaling/optimistic-rollups/index.md
emmanuel-awosika: Okay, then.

src/content/developers/docs/scaling/optimistic-rollups/index.md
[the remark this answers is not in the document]
emmanuel-awosika: Okay, I had to go through both the Paradigm article and Arbitrum docs again, but I think I can clarify my point now:

In both Optimism and Arbitrum, the invalid batch is rejected, but the transactions themselves are still available to be re-executed and applied to the rollup's state (by the node posting the valid assertion).  I don't think invalid transactions are "erased" in Arbitrum; for a batch to be invalid, it means someone else has posted another batch that correctly executes those transactions (they have to post correct assertions before challenging). If the challenger's block/state root is valid, it becomes the new state root in the on-chain contract--but the initial state root (proposed by the malicious node) has to be deleted first. 

Here's a quote from the Paradigm article for context: 

> Finally, [finalizeFraudVerification](https://github.com/ethereum-optimism/contracts-v2/blob/0ad4dcfdef11ef87e278a8159de8414c8e329ba1/contracts/optimistic-ethereum/OVM/verification/OVM_FraudVerifier.sol#L147) is called on the Fraud Verifier contract which checks if the state transitioner is complete and if yes, it calls [deleteStateBatch](https://github.com/ethereum-optimism/contracts-v2/blob/f6069a881fbf6c35687a1676a73c67596b3ef4f9/contracts/optimistic-ethereum/OVM/chain/OVM_StateCommitmentChain.sol#L86) which proceeds to delete all state root batches after (including) the disputed transaction from the SCC. **The CTC remains unchanged, so that the original transactions are re-executed in the same order**.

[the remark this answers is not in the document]
qbzzt: Sorry, I'm on a trip - I'll check on Wednesday, OK?

src/content/developers/docs/scaling/optimistic-rollups/index.md
qbzzt: I verified this with the developers. Yes, transactions are never invalid, you could say they get removed and immediately re-executed, but I think it's clearer to say their execution result is corrected.

src/content/developers/docs/scaling/optimistic-rollups/index.md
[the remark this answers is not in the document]
qbzzt: OK, I checked it with our developers. In Optimism nobody else posts a new batch, the batch itself is still valid. The only thing that changes is the state root.

src/content/developers/docs/scaling/optimistic-rollups/index.md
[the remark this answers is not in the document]
qbzzt: OK, I got the reply.

What happens is bedrock (our new version, to be released in the second half of 2022) is that a block is defined as all the transactions submitted on L1 since the previous block (a.k.a. deposits) followed by all the transactions published by the sequencer. So if the sequencer ignores an L1 transaction it publishes the wrong (provably wrong) state root.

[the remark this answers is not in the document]
qbzzt: So the sequencer can't even delay messages once they are posted on L1.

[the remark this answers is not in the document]
emmanuel-awosika: > OK, I got the reply.
> 
> What happens is bedrock (our new version, to be released in the second half of 2022) is that a block is defined as all the transactions submitted on L1 since the previous block (a.k.a. deposits) followed by all the transactions published by the sequencer. So if the sequencer ignores an L1 transaction it publishes the wrong (provably wrong) state root.

So I've had some time to think about the page a bit, and I think it'd be ineffective to use Optimism's design as the sole basis of the content--especially as users will likely interact with other ORs. For instance, the censorship-resistance mechanism described here is the standard for Arbitrum, Boba, and Metis (the latter two are running forks of Optimism's current version).

Nevertheless, I also used your comment to add a new description of the censorship-resistance mechanism: 
> Some optimistic rollups adopt a more straightforward approach to prevent sequencers from censoring users. Here, a block is defined by all transactions submitted to the L1 contract since the previous block (e.g., deposits) in addition to the transactions processed on the rollup chain. If a sequencer ignores an L1 transaction, it will publish the (provably) wrong state root; therefore, sequencers cannot delay user-generated messages once posted on L1

src/content/developers/docs/scaling/optimistic-rollups/index.md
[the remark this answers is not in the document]
emmanuel-awosika: Changed the text to read: 

> State commitments, especially state roots, are necessary for proving the correctness of state changes in an optimistic rollup. The rollup contract accepts new state roots from operators immediately they are posted, but can later delete invalid state roots to restore the rollup to its correct state. 

I think it explains how rollups operate "optimistically" and shows the core objective of posting and storing state commitments: verifying the rollup's canonical state. Let me know your thoughts on it.

src/content/developers/docs/scaling/optimistic-rollups/index.md
emmanuel-awosika: So, I changed the text to read: 

> If the fraud proof succeeds, the rollup protocol re-executes the transaction(s) and updates the rollup's state accordingly. The other effect of a successful fraud proof is that the sequencer responsible for including the incorrectly executed transaction in a block receives a penalty. 

> If the rollup batch remains unchallenged (i.e., all transactions are correctly executed) after the challenge period elapses, it is deemed valid and accepted on Ethereum. Others can continue to build on an unconfirmed rollup block, but with a caveat: transaction results will be reversed if based on an incorrectly executed transaction published previously. 

LMK what you think.

qbzzt: Approve.

src/content/developers/docs/scaling/optimistic-rollups/index.md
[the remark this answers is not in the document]
qbzzt: Approved

src/content/developers/docs/scaling/optimistic-rollups/index.md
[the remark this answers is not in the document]
emmanuel-awosika: Great.

src/content/developers/docs/scaling/optimistic-rollups/index.md
emmanuel-awosika: Good. Thanks for reviewing.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
 ---
 title: Optimistic Rollups
-description: Introduction to optimistic rollups
+description: An introduction to optimistic rollups as a scaling solution used by the Ethereum community. 
 lang: en
 sidebar: true
 ---
 
+Optimistic rollups are layer 2 (L2) protocols designed to extend throughput of Ethereum's base layer. They reduce computation on the main Ethereum chain by processing transactions off-chain, offering significant improvements in processing speeds. Unlike other scaling solutions, such as [sidechains](/developers/docs/scaling/sidechains/), optimistic rollups derive security from Mainnet by publishing transaction results on-chain.
```
corwintines:
```suggestion
Optimistic rollups are layer 2 (L2) protocols designed to extend the throughput of Ethereum's base layer. They reduce computation on the main Ethereum chain by processing transactions off-chain, offering significant improvements in processing speeds. Unlike other scaling solutions, such as [sidechains](/developers/docs/scaling/sidechains/), optimistic rollups derive security from Mainnet by publishing transaction results on-chain, or [plasma chains](https://github.com/developers/docs/scaling/plasma/), which also verify transactions on Ethereum with fraud proofs, but store transaction data elsewhere.
```

Figured since we referenced validiums on zk-rollup page, worth a similar reference for plasma here :)

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+Optimistic rollup operators bundle multiple off-chain transactions together in large batches before submitting to Ethereum. This approach enables spreading fixed costs across multiple transactions in each batch, reducing fees for end-users. Optimistic rollups also use compression techniques to reduce the amount of data posted on Ethereum.
+
+Optimistic rollups are considered “optimistic” because they assume off-chain transactions are valid and don't publish proofs of validity for transaction batches posted on-chain. This separates optimistic rollups from [zero-knowledge rollups](/developers/docs/scaling/zk-rollups) that publish cryptographic [proofs of validity](/glossary/#validity-proof) for off-chain transactions.
+
+Optimistic rollups instead rely on a fraud-proving scheme scheme to detect cases where transactions are not calculated correctly. After a rollup batch is submitted on Ethereum, there's a time window (called a challenge period) during which anyone can challenge the results of a rollup transaction by computing a [fraud proof](/glossary/#fraud-proof).
```
corwintines, line 24:
```suggestion
Optimistic rollups instead rely on a fraud-proving scheme to detect cases where transactions are not calculated correctly. After a rollup batch is submitted on Ethereum, there's a time window (called a challenge period) during which anyone can challenge the results of a rollup transaction by computing a [fraud proof](/glossary/#fraud-proof).
```

Extra scheme

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+If the rollup batch remains unchallenged (i.e., all transactions are correctly executed) after the challenge period elapses, it is deemed valid and accepted on Ethereum. Others can continue to build on an unconfirmed rollup block, but with a caveat: transaction results will be reversed if based on an incorrectly executed transaction published previously.
+
+## How do optimistic rollups interact with Ethereum? {#optimistic-rollups-and-Ethereum}
+
+Optimistic rollups are [off-chain scaling solutions](/developers/docs/scaling/#off-chain-scaling) built to operate on top of Ethereum. Each optimistic rollup is managed by a set of smart contracts deployed on the Ethereum network. Optimistic rollups process transactions off the main Ethereum chain, but post off-chain transaction (in batches) to an on-chain rollup contract. Like the Ethereum blockchain, this transaction record is immutable and forms the "optimistic rollup chain."
```
corwintines, line 32:
```suggestion
Optimistic rollups are [off-chain scaling solutions](/developers/docs/scaling/#off-chain-scaling) built to operate on top of Ethereum. Each optimistic rollup is managed by a set of smart contracts deployed on the Ethereum network. Optimistic rollups process transactions off the main Ethereum chain, but post off-chain transactions (in batches) to an on-chain rollup contract. Like the Ethereum blockchain, this transaction record is immutable and forms the "optimistic rollup chain."
```

Think this makes more sense as transactions

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+**Off-chain virtual machine (VM)**: Although contracts managing the optimistic rollup protocol run on Ethereum, the rollup protocol performs computation and state storage on another virtual machine separate from the [Ethereum Virtual Machine](/developers/docs/evm/). The off-chain VM is where applications live and state changes executed; it serves as the upper layer or "layer 2" for an optimistic rollup.
+
+As optimistic rollups are designed to run programs either written or compiled for the EVM, the off-chain VM incorporates many EVM design specs. Additionally, fraud proofs computed on-chain allows the Ethereum network to enforce the validity of state changes computed in the off-chain VM.
+
+Optimistic rollups are described as 'hybrid scaling solutions' because, while they exist as separate protocols, their security properties derive from Ethereum. Among other things, Ethereum guarantees the correctness of a rollup’s off-chain computation and the availability of data behind the computation. This makes optimistic rollups more secure than pure off-chain scaling protocols (e.g., [sidechains](/developers/docs/scaling/sidechains/)) that do not rely on Ethereum for security.
```
corwintines, line 42:
```suggestion
Optimistic rollups are described as 'hybrid scaling solutions' because, while they exist as separate protocols, their security properties are derived from Ethereum. Among other things, Ethereum guarantees the correctness of a rollup’s off-chain computation and the availability of data behind the computation. This makes optimistic rollups more secure than pure off-chain scaling protocols (e.g., [sidechains](/developers/docs/scaling/sidechains/)) that do not rely on Ethereum for security.
```

I think this might be a bit more clear/clean, but feel free to ignore.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+Optimistic rollups rely on the main Ethereum protocol for the following:
+
+### Data availability {#data-availability}
+
+As mentioned, optimistic rollups post transaction data to Ethereum as `calldata`. Since the rollup chain's execution is based on submitted transactions, anyone can use this information—anchored on Ethereum’s base layer—to execute the rollup’s state and verify the correctness of state transitions. .
```
corwintines, line 48:
```suggestion
As mentioned, optimistic rollups post transaction data to Ethereum as `calldata`. Since the rollup chain's execution is based on submitted transactions, anyone can use this information—anchored on Ethereum’s base layer—to execute the rollup’s state and verify the correctness of state transitions.
```
Extra period

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+Data availability is critical because without access to state data, challengers cannot construct fraud proofs to dispute invalid rollup operations. With Ethereum providing data availability, the risk of rollup operators getting away with malicious acts (e.g., submitting invalid blocks) is reduced.
+
+### Censorship resistance {#censorship-resistance}
+
+Optimistic rollups also rely on Ethereum for censorship resistance. In an optimistic rollup a centralized entity (the operator) is responsible for processing transactions and submitting rollup blocks to Ethereum. This has some implications:
```
corwintines, line 54: I think it's important to emphasize that this is the current state of optimistic rollups and their centralizing points, but I'm also thinking it might be worth noting that there are plans in roadmaps to get this decentralized. Doesn't need to be a blocker right now as this is reality, just might be a nice to have.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+Another role Ethereum plays in the context of optimistic rollups is that of a settlement layer. A settlement layer anchors the entire blockchain ecosystem, establishes security, and provides objective finality if a dispute occurs on another chain (optimistic rollups in this case) that requires arbitration.
+
+Ethereum Mainnet provides a hub for optimistic rollups to verify fraud proofs and resolve disputes. Moreover, transactions conducted on the rollup are only final _after_ the rollup block is accepted on Ethereum. Once a rollup transaction is committed to Ethereum’s base layer, it cannot be rolled back (except in the highly unlikely case of a chain reorganization).
+
+## How do optimistic rollups work? {#how-optimistic-rollups-work}
+
+### Transaction execution and aggregation {#transaction-execution-and-aggregation}
+
+Users submit transactions to “operators”, nodes responsible for processing transactions on the optimistic rollup. Also known as a “validator” or “aggregator”, the operator aggregates transactions, compresses the underlying data, and publishes the block on Ethereum.
```
corwintines, line 78:
```suggestion
Users submit transactions to “operators”, which are nodes responsible for processing transactions on the optimistic rollup. Also known as a “validator” or “aggregator”, the operator aggregates transactions, compresses the underlying data, and publishes the block on Ethereum.
```

I think this helps the sentence comprehension

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+Although anyone can become a validator, L2 nodes must provide a bond before producing blocks, much like a [proof-of-stake system](/developers/docs/consensus-mechanisms/pos/). This bond can be slashed if the validator posts an invalid block or builds on an old-but-invalid block (even if their block is valid). This way optimistic rollups utilize cryptoeconomic incentives to ensure validators act honestly.
```
corwintines, line 80:
```suggestion
Although anyone can become a validator, optimistic rollup validators must provide a bond before producing blocks, much like a [proof-of-stake system](/developers/docs/consensus-mechanisms/pos/). This bond can be slashed if the validator posts an invalid block or builds on an old-but-invalid block (even if their block is valid). This way optimistic rollups utilize cryptoeconomic incentives to ensure validators act honestly.
```

I think this makes a little more sense in the context of this page.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+Other validators on the L2 chain are expected to execute the submitted transactions using their copy of the rollup’s state. If a validator’s final state is different from the operator’s proposed state, they can start a challenge and compute a fraud proof.
```
corwintines, line 82:
```suggestion
Other validators on the optimistic rollup chain are expected to execute the submitted transactions using their copy of the rollup’s state. If a validator’s final state is different from the operator’s proposed state, they can start a challenge and compute a fraud proof.
```

Swapping L2 for optimistic rollup given context of this page

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+#### Submitting rollup blocks to Ethereum {#submitting-blocks-to-ethereum}
+
+As mentioned, the operator of an optimistic rollup bundles off-chain transactions into a batch and sends it to Ethereum for notarization. This process involves compressing transaction-related data and publishing it on Ethereum as `calldata`.
+
+Calldata is a non-modifiable, non-persistent area in a smart contract that behaves mostly like [memory](/developers/docs/smart-contracts/anatomy/#memory). While calldata persists on-chain as part of the blockchain's [history logs](https://docs.soliditylang.org/en/latest/introduction-to-smart-contracts.html?highlight=memory#logs), it is not stored as a part of Ethereum's state. Because calldata does not touch any part of Ethereum's state, it is cheaper for storing data on-chain.
+
+The `calldata` keyword is also used in Solidity to pass arguments to a smart contract function at execution time. Calldata identifies the function being called during a transaction and holds inputs to the function in the form of an arbitrary sequence of bytes.
```
corwintines, line 94:
```suggestion
The `calldata` keyword is also used in Solidity to pass arguments to a smart contract function at execution time. `calldata` identifies the function being called during a transaction and holds inputs to the function in the form of an arbitrary sequence of bytes.
```

Same here, standardizing use of calldata like we did on zk rollup page

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+The operator is required to submit both old state roots and new state roots when posting batches. If the old state root matches the existing state root in the on-chain contract, the latter is discarded and replaced with the new state root.
+
+The rollup operator is also required to commit a Merkle root for the transaction batch itself. This allows anyone to prove the inclusion of a transaction in the batch (on L1) by presenting a [Merkle proof](/developers/tutorials/merkle-proofs-for-offline-data-integrity/).
+
+State commitments, especially state roots, are necessary for proving the correctness of state changes in an optimistic rollup. The rollup contract accepts new state roots from operators immediately they are posted, but can later delete invalid state roots to restore the rollup to its correct state.
```
corwintines, line 108:
```suggestion
State commitments, especially state roots, are necessary for proving the correctness of state changes in an optimistic rollup. The rollup contract accepts new state roots from operators immediately after they are posted, but can later delete invalid state roots to restore the rollup to its correct state.
```

Sentence cleanup

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+The asserter is required to provide a “one-step proof” verifying the validity of the disputed single-step computation. If the asserter fails to provide the one-step proof, or the L1 verifier deems the proof invalid, they lose the challenge.
+
+Some notes about this type of fraud proof:
+
+1. Multi-round interactive fraud proving is considered efficient because it minimizes the work the L1 chain must do in dispute arbitration. Instead of replaying the entire transaction, the L1 chain only needs to re-execute one step in the rollup's execution.
+
+2. Bisection protocols reduce the amount of data posted on-chain (no need to publish state commits for every transaction). Also, optimistic rollup transactions don't have constrained by Ethereum's gas limit. Conversely, optimistic rollups re-executing transactions must make sure an L2 transaction has a lower gas limit to emulate its execution within a single Ethereum transaction.
```
corwintines, line 132:
```suggestion
2. Bisection protocols reduce the amount of data posted on-chain (no need to publish state commits for every transaction). Also, optimistic rollup transactions are not constrained by Ethereum's gas limit. Conversely, optimistic rollups re-executing transactions must make sure an L2 transaction has a lower gas limit to emulate its execution within a single Ethereum transaction.
```

Think this makes a bit more sense.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+Malicious nodes can try to delay the confirmation of a valid rollup block by starting false challenges. However, fraud proofs will eventually prove the rollup block’s validity and cause it to be confirmed.
+
+This also relates to another security property of optimistic rollups: the validity of the chain relies on the existence of _one_ honest node. The honest node can advance the chain correctly by either posting valid assertions or disputing invalid assertions. Whatever the case, malicious nodes who enter into disputes with the honest node will lose their stakes during the fraud proving process.
+
+### L1/L2 interoperability {#l1-l2-interoperability}
+
+Optimistic rollups are designed for interoperability with Ethereum Mainnet and allow users to pass messages and arbitrary data between L1 and L2. They are also compatible with the EVM, so you can port existing [dapps](/developers/docs/dapps/) to rollups or create new dapps using Ethereum development tools.
```
corwintines, line 148:
```suggestion
Optimistic rollups are designed for interoperability with Ethereum Mainnet and allow users to pass messages and arbitrary data between L1 and L2. They are also compatible with the EVM, so you can port existing [dapps](/developers/docs/dapps/) to optimistic rollups or create new dapps using Ethereum development tools.
```

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+#### 1. Asset movement {#asset-movement}
+
+##### Entering the rollup
+
+To use an optimistic rollup, users deposit ETH, ERC-20 tokens, and other accepted assets in the rollup’s [bridge](/developers/docs/bridges/) contract on L1. The bridge contract will relay the transaction to L2, where an equivalent amount of assets is minted and sendt to the user’s chosen address on the optimistic rollup.
+
+User-generated transactions (like an L1 > L2 deposit) are usually queued until the sequencer re-submits them to the rollup contract. However, to preserve censorship resistance, optimistic rollups allow users are allowed to submit a transaction directly to the on-chain rollup contract if it has been delayed past the maximum time allowed.
```
corwintines, line 156:
```suggestion
User-generated transactions (like an L1 > L2 deposit) are usually queued until the sequencer re-submits them to the rollup contract. However, to preserve censorship resistance, optimistic rollups users are allowed to submit a transaction directly to the on-chain rollup contract if it has been delayed past the maximum time allowed.
```
Either this

```suggestion
User-generated transactions (like an L1 > L2 deposit) are usually queued until the sequencer re-submits them to the rollup contract. However, to preserve censorship resistance, optimistic rollups allow users to submit a transaction directly to the on-chain rollup contract if it has been delayed past the maximum time allowed.
```

or this. No strong preference, but should use one of them to clean this up.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+To avoid waiting a week before withdrawing funds to Ethereum, optimistic rollup users can employ a **liquidity provider** (LP). A liquidity provider assumes ownership of a pending L2 withdrawal and pays the user on L1 (in exchange for a fee).
+
+Liquidity providers can check the validity of the user’s withdrawal request (by executing the chain themselves) before releasing funds. This way they have assurances that the transaction will be confirmed eventually (i.e., trustless finality).
+
+#### 2. EVM compatibility {#evm-compatibility}
+
+For developers, the advantage of optimistic rollups is their compatibility—or, better still, equivalence—with the [Ethereum Virtual Machine (EVM)](/developers/docs/evm/). EVM-compatible rollups rollups comply with specifications in the [Ethereum Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) and support the EVM at the bytecode level.
```
corwintines, line 172:
```suggestion
For developers, the advantage of optimistic rollups is their compatibility—or, better still, equivalence—with the [Ethereum Virtual Machine (EVM)](/developers/docs/evm/). EVM-compatible rollups comply with specifications in the [Ethereum Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) and support the EVM at the bytecode level.
```

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+ii. Developers and project teams using optimistic rollups can take advantage of Ethereum's infrastructure. This includes programming languages, code libraries, testing tools, client software, deployment infrastructure, and so on.
+
+Using existing tooling is important because these tools have been extensively audited, debugged, and improved over the years. It also removes the need for Ethereum developers to learn how to build with an entirely new development stack.
+
+#### 3. Cross-chain contract calls {#cross-chain-contract-calls}
+
+Users (externally owned accounts) interact with L2 contracts by submitting a transaction to the rollup contract or having a sequencer or validator do it for them. Optimistic rollups also allow contract accounts on Ethereum to interact with L2 contracts using bridging contracts to relay messages and pass data between L1 and L2. This means you can program an L1 contract on Ethereum Mainnet can to invoke functions belonging to contracts on an L2 ptimistic rollup.
```
corwintines, line 184:
```suggestion
Users (externally owned accounts) interact with L2 contracts by submitting a transaction to the rollup contract or having a sequencer or validator do it for them. Optimistic rollups also allow contract accounts on Ethereum to interact with L2 contracts using bridging contracts to relay messages and pass data between L1 and L2. This means you can program an L1 contract on Ethereum Mainnet to invoke functions belonging to contracts on an L2 optimistic rollup.
```

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+## How do optimistic rollup fees work? {#how-do-optimistic-rollup-fees-work}
+
+Optimistic rollups use a gas fee scheme, much like Ethereum, to denote how much users pay per transaction. Fees charged on optimistic rollups depends on the following components:
+
+1. **State write**: Optimistic rollups publish transaction data and block headers (consisting of the previous block header hash, state root, batch root) to Ethereum as `calldata`. The minimum cost of an Ethereum transaction is 21,000 gas. Optimistic rollups can reduce the cost of writing the transaction to L1 by batching multiple transactions in a single block (which amortizes the 21k gas over multiple user transactions).
+
+Beyond the base transaction fee, the cost of each state write depends on the size of `calldata` posted to L1. Calldata costs are currently governed by [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559), which stipulates a cost of 16 gas for non-zero bytes and 4 gas for zero bytes of calldata, respectively. To reduce user fees, rollup operators compresss transactions to reduce the number of `calldata` bytes published on Ethereum.
```
corwintines, line 200:
```suggestion
2. **`calldata`**: Beyond the base transaction fee, the cost of each state write depends on the size of `calldata` posted to L1. `calldata` costs are currently governed by [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559), which stipulates a cost of 16 gas for non-zero bytes and 4 gas for zero bytes of `calldata`, respectively. To reduce user fees, rollup operators compresss transactions to reduce the number of `calldata` bytes published on Ethereum.
```

I believe you meant to add something like this since we skip from 1 to 3 :)

corwintines: Maybe data availability is a better term to be bolded for 2 instead of calldata. Open to what you think is best here @emmanuel-awosika

corwintines (changes requested):
Sorry @emmanuel-awosika, should have made requests in one review. Above I left some change requests if you don't mind taking a look :)

src/content/developers/docs/scaling/optimistic-rollups/index.md
emmanuel-awosika: Second one sounds better.

src/content/developers/docs/scaling/optimistic-rollups/index.md
```diff
@@ -1,46 +1,252 @@
...
+
+To avoid waiting a week before withdrawing funds to Ethereum, optimistic rollup users can employ a **liquidity provider** (LP). A liquidity provider assumes ownership of a pending L2 withdrawal and pays the user on L1 (in exchange for a fee).
+
+Liquidity providers can check the validity of the user’s withdrawal request (by executing the chain themselves) before releasing funds. This way they have assurances that the transaction will be confirmed eventually (i.e., trustless finality).
+
+#### 2. EVM compatibility {#evm-compatibility}
+
+For developers, the advantage of optimistic rollups is their compatibility—or, better still, equivalence—with the [Ethereum Virtual Machine (EVM)](/developers/docs/evm/). EVM-compatible rollups rollups comply with specifications in the [Ethereum Yellow Paper](https://ethereum.github.io/yellowpaper/paper.pdf) and support the EVM at the bytecode level.
+
+EVM-compatibility in optimistic rollups has the following benefits:
```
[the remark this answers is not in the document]
emmanuel-awosika: It should be "has" since "EVM-compatibility", which describes a singular property, is the subject of the sentence. "Have" would be used if the statement read: "EVM-compatible rollups *have* the following benefits..."

src/content/developers/docs/scaling/optimistic-rollups/index.md
emmanuel-awosika: I separated the "state write" part from "data publication" (which covers `calldata`) in an earlier version of the page. But, after feedback from @qbzzt, I opted to bundle both concepts under the state write section. You'll notice I said, "Beyond the base transaction fee, the cost of each state write depends on the size of `calldata` posted to L1...:" to bridge both parts. 

I feel it makes sense to combine them since they are part of the same activity (writing to Ethereum's state). This is not a strong opinion, though--I can add "data publication" to represent the cost of `calldata`, like I did before.

src/content/developers/docs/scaling/optimistic-rollups/index.md
corwintines: Will leave this for now after talking with @emmanuel-awosika. This can be updated as these rollup roadmaps are worked on.

src/content/developers/docs/scaling/optimistic-rollups/index.md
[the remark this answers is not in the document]
corwintines: Makes sense, thanks for clarifying :)

corwintines (approved):
I think this is good to go, big improvement of the content here, thanks so much @emmanuel-awosika! Great work as always!

emmanuel-awosika:
> I think this is good to go, big improvement of the content here, thanks so much @emmanuel-awosika! Great work as always!

Glad I could help improve the page. :)

PR #6733, opened 2022-06-17
Whitepaper PDF context added.

polarpunklabs:
Added a small contextual note that the PDF is the best version for academics or researchers seeking a historical or canonical version of the paper. The version that is on the website is actually a newer version and does not match the PDF . Would be good if someone could pinpoint the source of the website version. I would personally prefer if they were the same, but this should be a good accommodation for now.


minimalsm (approved):
Thanks @polarpunklabs

PR #6765, opened 2022-06-22
Upgrade home icon to svg

wackerow:
## Description
- Upgrades home icon to SVG
- As a result the `svg:hover` selector in the `Link.tsx` component causes this to scale by 2% on hover (since it is now an SVG)... which I thought was a nice bonus, so I kept it in there
- Note, there is an existing SVG with a heavier stroke width that didn't quite match our existing home logo, hence the new file. If we feel it's too redundant, would suggest we remove the old one and replace it with this new one on the assets page. (This PR doesn't touch the assets page)

<-- old / new -->
image


Nana12345678910 (approved):
Pull request

PR #6766, opened 2022-06-22
refactor: migrate `OrderedList`, `DocLink`, and `Morpher` to ts

byhow:
## Description

Migrate `OrderedList`, `DocLink`, and `Morpher` to typescript
## Related Issue
https://github.com/ethereum/ethereum-org-website/issues/6392

### DIFF
--- a/src/components/DocLink.tsx
+++ b/src/components/DocLink.tsx
@@ -1,8 +1,14 @@
 import React from "react"
-import styled from "styled-components"
+import styled, { DefaultTheme } from "styled-components"
 import Icon from "./Icon"
 import Link from "./Link"
 import Emoji from "./Emoji"
+
+export interface IProps {
+  to: string
+  theme?: DefaultTheme
+  className?: string
+}
 
 const Container = styled(Link)`
   position: relative;
@@ -43,6 +49,7 @@ const Arrow = styled(Icon)`
   margin: 0rem 1.5rem;
   align-self: center;
   min-width: 2rem;
+  color: ${({ theme }) => theme.colors.text};
 `
 
 const EmojiCell = styled.div`
@@ -50,15 +57,15 @@ const EmojiCell = styled.div`
   align-items: center;
 `
 
-const DocLink = ({ to, children, className }) => (
-  <Container to={to} className={className}>
+const DocLink: React.FC<IProps> = ({ to, children, className, theme }) => (
+  <Container to={to} className={className || ""}>
     <EmojiCell>
       <Emoji size={1} text=":page_with_curl:" mr={`1rem`} />
     </EmojiCell>
     <TextCell>
       <Title>{children}</Title>
     </TextCell>
-    <Arrow name="arrowRight" color={({ theme }) => theme.colors.text} />
+    <Arrow name="arrowRight" />
   </Container>
 )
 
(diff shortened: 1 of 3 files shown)


byhow:
During migration, another issue was caugth when running `yarn type-check`. More context here: #6767

byhow:
Unable to view the build as I don't have access. Locally `yarn type-check` looks fine with only errors on `error TS2503: Cannot find namespace 'Queries'`, which happened in previous PRs. 

I also tried to do a local build but the build was stuck on `⠴ source and transform nodes` with:
```sh
> yarn clean
> yarn start / yarn start:static / yarn build
```

src/components/OrderedList.tsx
```diff
@@ -2,8 +2,13 @@
...
 
-//Styles
-const Content = styled.div`
+export interface IProps {
+  listData: (string | HTMLParagraphElement)[]
+  className?: string
+}
+
+// Styles
+const Content = styled.div<{ size?: `${string}px` }>`
```
pettinarip, line 11: Feel free to remove the `size` logic from here since we are not using it.

But if you want to leave it, lets define it as a `number` and handle the `px` suffix where we use it.
```suggestion
const Content = styled.div<{ size?: number }>`
```

pettinarip (changes requested):
A few comments @byhow. Thanks for the contribution.

src/components/DocLink.tsx
```diff
@@ -43,22 +48,23 @@ const Arrow = styled(Icon)`
...
 
 const EmojiCell = styled.div`
   display: flex;
   align-items: center;
 `
 
-const DocLink = ({ to, children, className }) => (
-  <Container to={to} className={className}>
+const DocLink: React.FC<IProps> = ({ to, children, className }) => (
+  <Container to={to} className={className || ""}>
```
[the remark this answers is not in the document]
byhow: got it, updated

src/components/OrderedList.tsx
```diff
@@ -2,8 +2,13 @@
 import React from "react"
 import styled from "styled-components"
 
-//Styles
-const Content = styled.div`
+export interface IProps {
+  listData: (string | HTMLParagraphElement)[]
```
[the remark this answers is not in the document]
byhow: got it, type widened

src/components/OrderedList.tsx
byhow: since we are not using the `size` logic, I will remove it to prevent confusion in legacy code. LMK if this is ok! Or I can just update according to ^ suggestion

vdusart:
> A few comments @vdusart. Thanks for the contribution.

I think you have the wrong person, let's thank @byhow instead haha

byhow:
@pettinarip no worries! totally understandable lol. there's a lot of people to keep track of

src/components/OrderedList.tsx
```diff
@@ -22,7 +27,7 @@ const Content = styled.div`
     position: absolute;
     top: -2px; /* adjusts circle + number up and down */
     left: -3rem;
-    width: ${({ size }) => (size ? size : "35px")};
+    width: "35px";
```
pettinarip, line 30:
```suggestion
    width: 35px;
```

pettinarip (approved):
LGTM thanks for your work @byhow gj!

Issue #6281, opened 2022-05-10
Epic: pre-merge prep


wackerow:
Context: Epic from Q2 Roadmap https://github.com/ethereum/ethereum-org-website/issues/6161

## Epic Description
After a long wait, the Merge is almost upon us. The switch to proof-of-stake means that large sections of ethereum.org will need rewriting to remain up to date. 

## Involves 
- Scoping out what content changes, additions, or subtractions are required related to the Merge upgrade
- Preparing appropriate content changes, 
- Coordinating with external contributors to ensure accuracy
- Timing releases  appropriately to provide the community with accurate information about the Merge during the transition

## Goals
**Pre-Merge Prep**
- Help clear up misconceptions of **every day users** regarding how the Merge will be executed, and the impact it will/will not have. Also reassuring these users that there is nothing they need to do with their funds.
- Inform **dapp developers** of important changes that may result from the Merge
- Inform **node operators** of important changes, including the need to run both an execution layer (EL) client as well as a consensus layer (CL) client to sync the head of the chain
- Inform **stakers** of important changes, including that their nodes will require the above CL+EL setup, and they will need to provide a traditional Ethereum address to receive unburnt fees in blocks they propose

**Post-merge, longer-term changes\***
- Ensure references to proof-of-work are updated to either be historical references, or removed
- Update developer documentation to ensure all fundamentals are accurately represented under the proof-of-stake model
- Update `/upgrade/merge` and to be reflective of past tense and latest information
- Update `/energy-consumption` page with updated latest information

## Checkpoints
[Page Audit Spreadsheet](https://docs.google.com/spreadsheets/d/1Nisi3_FY8nwOlLwPsTPm0jlgtAhypH0-pjatGDMSz7M/edit#gid=0)

**Pre-merge update PR:**

- [x] Banner notification: some page content may be out of date related to the merge #6629
  - /developers/docs/consensus-mechanisms/pow/mining-algorithms/
  - /developers/docs/development-networks/
  - /developers/docs/intro-to-ether/
  - /developers/docs/mev/
  - /developers/docs/mining/
  - /developers/docs/networking-layer/
  - /developers/docs/networks/
  - /developers/docs/nodes-and-clients/
  - /developers/docs/nodes-and-clients/client-diversity/
  - /developers/docs/nodes-and-clients/nodes-as-a-service/
  - /developers/tutorials/run-light-node-geth/
  - /developers/tutorials/run-node-raspberry-pi/
  - /nft/
[...]

PR #6780, opened 2022-06-22
update mev page for pre-merge

jmcook1186:
## Description

minor changes to content to make it appropriate for pre-merge, for example changign references to miners to "miners/validators", plus some typo fixes.

## Related Issue

#6281


src/content/developers/docs/mev/index.md
```diff
@@ -9,19 +9,19 @@ Maximal extractable value (MEV) refers to the maximum value that can be extracte
...
+This concept was first applied in the context of [proof-of-work](/developers/docs/consensus-mechanisms/pow/), and was initially referred to as "miner extractable value". This is because in proof-of-work, miners control transaction inclusion, exclusion, and ordering. However, after the transition to proof-of-stake via [The Merge](/upgrades/merge) validators will be responsible for these roles, and mining will no longer be applicable. The value extraction methods here will still persist after this transition, and thus the term "miner extractable value" is invalidated. "Maximal extractable value" is now used as a more inclusive replacement.
```
minimalsm, line 12:
```suggestion
This concept was first applied in the context of [proof-of-work](/developers/docs/consensus-mechanisms/pow/), and was initially referred to as "miner extractable value". This is because in proof-of-work, miners control transaction inclusion, exclusion, and ordering. However, after the transition to proof-of-stake via [The Merge](/upgrades/merge) validators will be responsible for these roles, and mining will no longer be applicable. The value extraction methods here will still persist after this transition, and thus the term "miner extractable value" is not longer valid. "Maximal extractable value" is now used as a more inclusive replacement.
```

minimalsm (approved):

src/content/developers/docs/mev/index.md
wackerow:
```suggestion
This concept was first applied in the context of [proof-of-work](/developers/docs/consensus-mechanisms/pow/), and was initially referred to as "miner extractable value". This is because in proof-of-work, miners control transaction inclusion, exclusion, and ordering. However, after the transition to proof-of-stake via [The Merge](/upgrades/merge) validators will be responsible for these roles, and mining will no longer be applicable. The value extraction methods here will still persist after this transition, and thus the term "miner extractable value" is no longer valid. "Maximal extractable value" is now used as a more inclusive replacement.
```

wackerow (approved):

PR #6788, opened 2022-06-22
Include Arbitrum key in layer-2.json

smartcontracts:
Key was missing, meaning the note wasn't being shown on the website.


wackerow (approved):
Thanks @smartcontracts! Like the handle =)

PR #6829, opened 2022-06-26
Ts migration(src/pages): Migrated assigned files

M-Ivan:
## Description

Migrated: 
*  src/pages/404.js
*  src/pages/assets.js
*  src/pages/bug-bounty.js
*  src/pages/community.js 

## Related Issue

PR Is related with [6392](https://github.com/ethereum/ethereum-org-website/issues/6392#)


src/pages/404.tsx
```diff
@@ -10,14 +10,14 @@ const StyledPage = styled(Page)`
...
 
-const NotFoundPage = () => (
+const NotFoundPage: React.FC = () => (
   <StyledPage>
     <Content>
       <h1>
         <Translation id="we-couldnt-find-that-page" />
       </h1>
       <p>
-        <Translation id="try-using-search" />{" "}
```
pettinarip, line 20: We need that extra space

src/pages/404.tsx
```diff
@@ -10,14 +10,14 @@ const StyledPage = styled(Page)`
   margin-top: 4rem;
 `
 
-const NotFoundPage = () => (
+const NotFoundPage: React.FC = () => (
```
pettinarip, line 13: Since this is a Page component, we are not defining it as a `React.FC`. Lets keep the consistency across the other pages by only defining the props with `PageProps`.
```suggestion
const NotFoundPage = (props: PageProps) => {
```

src/pages/assets.tsx
```diff
@@ -74,7 +74,34 @@ const Header = styled.header`
...
+      fallback: string
+    }
+  }
+}
+
+export interface IProps {
+  data: Record<string, IGatsbyChildImageSharp>
+}
+
+const AssetsPage: React.FC<IProps> = ({ data }) => {
```
pettinarip, line 104: As mentioned in the other PR, lets use `PageProps` from gatsby and use the auto-generated types using `Queries.{NameOfTheQuery}`.

src/pages/assets.tsx
```diff
@@ -116,7 +143,7 @@ const AssetsPage = ({ data }) => {
           <AssetDownload
             title={translateMessageId("page-assets-hero", intl)}
             alt={translateMessageId("page-assets-hero", intl)}
-            image={data.hero}
+            image={getImage(data.hero)}
```
pettinarip, line 146: I think we don't need to do `getImage` here since `AssetDownload` is already doing that for us.

src/pages/bug-bounty.tsx
```diff
@@ -229,7 +230,39 @@ const Faq = styled.div`
...
+  link: string
+  image: any
+}
+
+export interface ISpec {
+  title: JSX.Element
+  link: string
+}
+
+const BugBountiesPage: React.FC<IProps> = ({ data, location }) => {
```
pettinarip, line 265: Same as the other pages, lest use `PageProps`.

src/pages/bug-bounty.tsx
```diff
@@ -326,7 +361,7 @@ const BugBountiesPage = ({ data, location }) => {
     ? getImage(data.lighthouseDark)
     : getImage(data.lighthouseLight)
 
-  const specs = [
+  const specs: ISpec[] = [
```
pettinarip, line 364: For consistency,
```suggestion
  const specs: Array<ISpec> = [
```

src/pages/community.tsx
```diff
@@ -305,7 +306,25 @@ const StyledCallout = styled(Callout)`
...
+  to: string
+}
+
+export interface IGetInvolvedCard {
+  emoji: string
+  title: string
+  description: string
+}
+
+const CommunityPage: React.FC<IProps> = ({ data }) => {
```
pettinarip, line 327: Same here, `PageProps`

src/pages/community.tsx
```diff
@@ -316,7 +335,7 @@ const CommunityPage = ({ data }) => {
     alt: translateMessageId("page-community-hero-alt", intl),
   }
 
-  const cards = [
+  const cards: ICard[] = [
```
pettinarip, line 338:
```suggestion
  const cards: Array<ICard> = [
```

pettinarip (changes requested):
@M-Ivan thanks for the contribution!

Similar feedback as in your other PR. Lets use `PageProps` + the auto-generated query type.
- You can take a look at the `src/pages/index.tsx` file to take it as an example.
- Also, take a look at this doc for more context https://github.com/ethereum/ethereum-org-website/blob/dev/docs/typescript.md

src/components/AssetDownload.tsx
```diff
@@ -26,7 +26,7 @@ interface IPropsWithSVG extends IPropsBase {
 }
 interface IPropsWithImage extends IPropsBase {
   svg?: never
-  image: string
+  image: Record<string, unknown>
```
M-Ivan, line 29: @pettinarip since this prop is supposed to receive the GatsbyImageData object i changed this type to match any object. It was set to string before and it was a pain to make it work when passing a Gatsby auto-generated type (from the Queries namespace).

src/pages/assets.tsx
```diff
@@ -117,7 +118,7 @@ const AssetsPage = ({ data }) => {
           <AssetDownload
             title={translateMessageId("page-assets-hero", intl)}
             alt={translateMessageId("page-assets-hero", intl)}
-            image={data.hero}
+            image={data.hero as Record<string, unknown>}
```
M-Ivan, line 121: @pettinarip Type assetion here is for the AssetDownload component prop to don't cry. It actually complains a lot because the props of the component expect a non-nullable object, even tho in the component this specific prop value is used for a gatsby call is using getSrc, a null-safe method ([docs here](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-plugin-image/#getimage))

pettinarip: This is fine but there are some issues with the `getImage` types. In the `overrides.d.ts` we are overriding the types for this function to basically accepts `any` as its argument. So, in order to keep it simple, we can just accept `any` in `AssetDownload` while we wait for a proper fix.

pettinarip (approved):
@M-Ivan thanks for the PR! this looks great! I'm pushing some minor changes and reverting the md changes.

src/components/AssetDownload.tsx
pettinarip: This is nice! as commented in the other comment, we will leave it as `any` for now. Once we have better types from gatsby we will refactor and fix all the images types 👍🏼

PR #6830, opened 2022-06-26
refactor: migrate `SharedStyledComponents` and `studio` page to ts

byhow:
## Description

Migrating the component `SharedStyledComponents` and page `studio` to ts.

On top of that, some other pages and components also needed some minor updates.

## Related Issue
#6392


src/components/ButtonLink.tsx
```diff
@@ -85,9 +85,8 @@ const SecondaryScrollLink = styled(StyledScrollButton)`
   }
 `
 
-interface IProps extends ILinkProps {
+interface IProps extends ILinkProps, MarginProps {
   toId?: string
-  isSecondary?: boolean
```
byhow, line 90: Moving `isSecondary` to the base `Link` as the line where it says:
```tsx
 <Link isSecondary to="/developers/">
```
also needed this prop

src/components/FileContributors.tsx
```diff
@@ -136,7 +136,7 @@ const SkeletonButtonContainer = styled(ButtonContainer)`
   }
 `
 
-const ContributorsButton = styled(ButtonSecondary)`
+const ContributorsButton = styled(ButtonSecondary)<{ loading: boolean }>`
```
byhow, line 139: Need this prop as this particular component used `loading`

src/components/SharedStyledComponents.tsx
```diff
@@ -384,7 +384,7 @@ export const OptionContainer = styled.div`
   }
 `
 
-export const Option = styled.button`
+export const Option = styled.button<{ isActive: boolean }>`
```
byhow, line 387: Missing type from `studio`

src/components/SharedStyledComponents.tsx
```diff
@@ -233,7 +233,7 @@ export const ButtonPrimary = styled(Button)`
   }
 `
 
-export const ButtonSecondary = styled(Button)`
+export const ButtonSecondary = styled(Button)<MarginProps>`
```
byhow, line 236: Missing marginal types such as `mr`

src/components/SharedStyledComponents.tsx
```diff
@@ -412,7 +412,7 @@ export const Option = styled.button`
   }
 `
 
-export const OptionText = styled.span`
+export const OptionText = styled.span<{ fontSize?: string }>`
```
byhow, line 415: Missing font size and also has a base value

src/pages-conditional/dapps.tsx
```diff
@@ -1242,7 +1242,7 @@ const DappsPage = ({
           ))}
         </StyledCardGrid>
       </Content>
-      <FullWidthContainer ref={explore}>
```
byhow, line 1245: Doesn't seems like `FullWidthContainer` has used this prop at all as well as the parent components of it, and the `explore` here was only used to grab the DOM. LMK if this is not expected!

src/components/Link.tsx
```diff
@@ -71,6 +71,7 @@ export interface IProps {
   className?: string
   isPartiallyActive?: boolean
   ariaLabel?: string
+  isSecondary?: boolean
```
byhow, line 74: ditto

src/components/SharedStyledComponents.tsx
pettinarip: Cool, lets add that type to the `Button` component instead of the `ButtonSecondary`.

pettinarip (changes requested):
A couple of change requests @byhow but it is looking great!

Merge `dev` into this branch because there were a couple of new updates and this `SharedStyledComponents` is used a lot across the app.

src/components/ButtonLink.tsx
byhow: Thank you! Changes reverted based on ^.

Do you want me to include the update of all of this case to use `ButtonLink` instead of `Link` in this PR as well?

src/components/SharedStyledComponents.tsx
byhow: updated!

src/pages-conditional/dapps.tsx
byhow: Got it. So there's an issue explained in 8ac930142fd1d137657bd697d54df42f3447657c but the gist is that when I type it with:
```ts
const FullWidthContainer = styled(Page)<{
  ref: React.RefObject<HTMLElement>
}>`
```
It will complain that it wont be able to match the type of `RefObject`, so I did some digging and found out that this particular property of this type is ignored [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e358cde5edeaa8156bc5b9086dd9f5443c2cadc7/types/react/index.d.ts#L818)

I am not too sure what the resolution to this could be. For now, I changed the name from `ref` to `scrollRef`, and the typing will work

pettinarip: There is no need to define the `ref` prop here. That prop is already defined by React. It is not a custom prop.

So,
1. We should keep `const FullWidthContainer = styled(Page)` as before
2. Keep using `ref` in `<FullWidthContainer ref={explore}>`
3. One thing I noticed is that we will need to change the ref type, from `HTMLElement` to `HTMLDivElement`, in `useRef`

src/pages/studio.tsx
```diff
@@ -71,9 +71,9 @@ const StudioRedirectPage = () => (
       <ButtonLink mr={"1rem"} isSecondary to="https://remix.ethereum.org">
         Use Remix
       </ButtonLink>
-      <Link isSecondary to="/developers/">
+      <ButtonLink isSecondary to="/developers/">
```
pettinarip, line 74: 💯 thanks!

src/components/ButtonLink.tsx
pettinarip: I saw that you already did that, awesome!

pettinarip (changes requested):
Perfect @byhow. Thanks for the adjustments so far.

src/components/ButtonLink.tsx
byhow: Yes, within this PR, the one in `studio.tsx` will be the only case. I was more referring to all the other use cases that should use `ButtonLink` instead of `<Link isSecondary ...`, if we needed to change the rest of them outside of this migration PR

src/components/Card.tsx
```diff
@@ -19,8 +19,8 @@ const TopContent = styled.div``
 
 export interface IProps {
   emoji?: string
-  title?: ReactNode
-  description?: ReactNode
+  title?: string | ReactNode
+  description?: string | ReactNode
```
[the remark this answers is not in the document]
byhow: got it, updated

src/pages-conditional/dapps.tsx
byhow: Ohhh I think that's why I was unable to type it since `ref` got a type mismatch from `HTMLElement` instead of the correct type `HTMLDivElement`, got a bit confused. Thanks for clearing it up! Changes updated

pettinarip (approved):
Awesome! thanks for the adjustments @byhow GJ

pettinarip:
@all-contributors please add @byhow for code

PR #6797, opened 2022-06-22
Learn revamp - what is ethereum page

pettinarip:
## Description

Revamp "What is Ethereum?" page.

https://ethereumorgwebsitedev01-whatisethrevamp.gtsb.io/en/what-is-ethereum/

## TODOs
- [x] Fetch infromation for the stats grid.
- [x] Display tooltip in stats grid.
- [x] Polish mobile version.
- [x] Add custom events tracking

PR #6832, opened 2022-06-27
adds energy expenditure graph to energy page

jmcook1186:
## Description
Adds simple graph comparing Ethereum energy expenditure to other industries


wackerow:
Looks great!

Only small concern is the contrast in light mode... @nloureiro curious your take on if this is adequate enough contrast:
image

nloureiro:
> 

I guess I was struggling a bit. I bring to have the same image for both themes and maintain contrast. Let me try to twiq a bit more.

That being said, this is not the final image. I think we can do a better visual metaphor with the same data, but for now, it helps to passe the message.

minimalsm (approved):
Merging this for now but we should replace it with the recharts component that @pettinarip created for the What is Ethereum page (#6797)

PR #6837, opened 2022-06-27
Update myetherwallet data

gamalielhere:
## Description

Change myetherwallet data for withdrawing to bank account to true


minimalsm (approved):
Thanks @gamalielhere

PR #6457, opened 2022-05-24
Find wallet revamp

corwintines:
## Description
New find wallets page for ethereum.org.

**Preview deploy:**
https://ethereumorgwebsitedev01-findwalletrevamp.gtsb.io/en/wallets/find-wallet/

## Related Issue
- https://github.com/ethereum/ethereum-org-website/issues/6274

## Work left checklist
- [x] Cant have same feature filter in multiple comparison dropdowns (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1158973395)
- [x] Focus state on profile filters (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1158976961)
- [x] Active profile filter background color hard to read with (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1158982619_
- [x] Reset filters (https://github.com/ethereum/ethereum-org-website/pull/6457#discussion_r900271258)
- [x] Filter column widths desktop view (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1159006056)
  - Note: This seemed to be fixed with removing the icons
- [x] Showing wallets text color and description edit (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1159007492)
- [x] When deselecting all device options, should untoggle that device option (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1159008725)
- [x] Remove icons in filter dropdown options (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1159011161)
- [x] General type/copy review (https://github.com/ethereum/ethereum-org-website/pull/6457#discussion_r900280402)
- [x] Toggle bug when switching between filter panels (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1159018464)
- [x] Padding in filter compare dropdowns (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1159030178) (might be solved when removing icons)
- [x] Expand clickable area for toggles and checkboxes (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1159031371)
- [x] Fix sticky styling for headers (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1159052436)
- [x] Update compare icons to green check and red x (https://github.com/ethereum/ethereum-org-website/pull/6457#issuecomment-1160224470)
[...]

PR #6856, opened 2022-06-29
Remove find wallet hero image alts

minimalsm:
## Description
- Remove unnecessary translations for decorative hero image on find-wallets page
- Add alt="" instead
- Separate PR to keep the file changes in #6457 readable

## Related Issue
#6457


wackerow (commented):
I'm not sure I understand why we're doing this here and not on other similar hero images. We've typically attempted to place alt text on all images, is there a reason for sending an empty string just for the find wallets hero?

minimalsm:
@wackerow we should have blank alt tags on every decorative image. I removed these ones in a separate PR because I noticed this problem there and I didn't want to add noise to #6457.

PR #6889, opened 2022-07-01
Turkish import 'community' from Crowdin

wackerow:
## Description
- Updates Turkish `tr` content with latest from Crowdin for Community content bucket

Noting the bullets links here are still not translated. @lukassim Should we revert that file and wait til it's fully translated?

## Related Issue
Ongoing translation efforts


wackerow:
Path `/community/online/index.md` has not fully been translated yet. This file is also having some parsing issues where the italicized text nested within HTML is being exported from Crowdin using markdown syntax, which cannot be used when nested inside HTML. The English version uses `<i></i>` tags. I used a regex pattern to fix this for this file, but then realized the text wasn't even translated yet.

Reverting this file until translations completed.

minimalsm (approved):

PR #6891, opened 2022-07-01
refactor: migrate Tag to typescript

didoshotev:
## Description
Migrate [Tag](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Tag.js) to typescript

## Related Issue

Refs: #6392

### DIFF
--- a/src/components/Tag.tsx
+++ b/src/components/Tag.tsx
@@ -51,7 +51,6 @@ const StyledIcon = styled(Icon)`
 
 export interface IProps {
   name: string
-  key: string
   onSelect: (tagName: string) => void
   value: string
   isActive: boolean
@@ -60,7 +59,6 @@ export interface IProps {
 
 const Tag: React.FC<IProps> = ({
   name,
-  key,
   onSelect,
   value,
   isActive = true,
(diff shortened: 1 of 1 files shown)


src/components/Tag.tsx
```diff
@@ -46,21 +49,27 @@ const StyledIcon = styled(Icon)`
   margin-left: 1em;
 `
 
-const Tag = ({
+export interface IProps {
+  name: string
+  key: string
```
pettinarip, line 54: There is no need to specify `key` in this component. `key` is used on the consumer, when you need to do a `.map` or a loop but inside the component, you don't need to pass that as a prop.

pettinarip (changes requested):
Looking great @didoshotev 💪🏼 thanks for the PR.

I've left a comment there to do some cleanup.

src/components/Tag.tsx
didoshotev: Make sense, thank you.

pettinarip (approved):
Awesome @didoshotev 🚀

pettinarip:
@didoshotev be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And we've just released our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) also 🥳!

pettinarip:
@all-contributors please add @didoshotev for code

Issue #4253, opened 2021-10-18
Catch missing lang property error before build


minimalsm:
**Is your feature request related to a problem? Please describe.**

In [this PR](https://github.com/ethereum/ethereum-org-website/pull/4252) we created a check that will throw an error during the build if a markdown page does not have a lang attribute. The problem with this is that the build could run for 30+ minutes before throwing an error.

**Describe the solution you'd like**

We should implement a check to catch this **before** we reach the point of doing a build. This could be something like a pre-commit hook or a test.

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

narayand16:
My approach is to use husky pre-commit hook which will check this lang in the MDX file. Can you assign this to me ?

minimalsm:
Hey, @noopta sure. I have assigned you. I think the suggested approach by @narayand16 is probably sufficient to solve this for now.

Thanks for offering @narayand16; if @noopta ends up not working on this, I'll ping you to see if you're still interested...

minimalsm:
Hey @narayand16 

Sorry for the late reply on this. Do you want to work on this still?

narayand16:
Hi Joshua,

I would like to work on this issue. Can you assign it to me ?

It would be great if you could give a little brief about the issue and what
we are trying to accomplish.

Thanks,
Narayan Deshmukh

akht:
Hi @minimalsm , can I take up this issue?

SNikhill:
@minimalsm 

May I tackle this one?

minimalsm:
Closing this out as this was solved as part of #7088

PR #6899, opened 2022-07-02
Hook to check Language Attribute in a Markdown file

SNikhill:
## Description

Create a script to verify whether a markdown file
part of the `src/content` folder contains a non-empty `lang` attribute
as a part of its front matter.
 

I have used `git diff` with some regex to see that the Added
markdown file **changes** contains `lang`.

## Caveats

- This hook will work only for newly created markdown files
- This hook will perform a **rudimentary** `lang` attribute check based on a regex match
image

image

## Testing

You will need to create some markdown files at random places
to test out the same.

1. Create a markdown file with an invalid or without lang attribute at a folder other than the `src/content`
2. **Stage the file**
3. Run `npm run-script check-markdown-language-tag`

It will not be detected by the script and hence, the hook will proceed.

- Repeat the above steps but, this time move the markdown file to a place under `src/content`

The script should throw an error

## Related Issue

- related to #4253

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "format": "prettier --write \"**/*.{js,jsx,json,md}\"",
     "generate-heading-ids": "node src/scripts/generate-heading-ids.js",
     "merge-translations": "node src/scripts/merge-translations.js",
+	"check-markdown-language-tag": "node src/scripts/checkMarkdownLanguage.mjs",
     "start": "gatsby develop",
     "start:lambda": "netlify-lambda serve src/lambda",
     "start:static": "gatsby build && gatsby serve",
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "prettier": "^2.2.1",
     "pretty-quick": "^3.1.0",
     "react-test-renderer": "^17.0.1",
+    "ts-node": "^10.8.2",
     "typescript": "^4.6.3"
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "format": "prettier --write \"**/*.{js,jsx,json,md}\"",
     "generate-heading-ids": "node src/scripts/generate-heading-ids.js",
     "merge-translations": "node src/scripts/merge-translations.js",
-	"check-markdown-language-tag": "node src/scripts/checkMarkdownLanguage.mjs",
+    "check-markdown-language-tag": "ts-node --esm src/scripts/checkMarkdownLanguage.mts",
     "start": "gatsby develop",
     "start:lambda": "netlify-lambda serve src/lambda",
     "start:static": "gatsby build && gatsby serve",
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged"
+      "pre-commit": "npm run-script check-markdown-language-tag && pretty-quick --staged"
     }
   },
   "volta": {
--- a/src/scripts/checkMarkdownLanguage.mjs
+++ b/src/scripts/checkMarkdownLanguage.mjs
@@ -3,28 +3,30 @@ import { spawn } from "child_process"
 let allMarkdownFilesBuffer = ""
 
 const getAllMarkdownFiles = () => {
-  const allMarkdownFilesCommand = spawn("git", [
-    "diff",
-    "--cached",
-    "--name-only",
-    "--diff-filter",
-    "AM",
-    "--",
-    "*.md",
-  ])
+  return new Promise((resolve, reject) => {
+    const allMarkdownFilesCommand = spawn("git", [
+      "diff",
+      "--cached",
+      "--name-only",
+      "--diff-filter",
+      "AM",
+      "--",
+      "*.md",
+    ])
 
-  allMarkdownFilesCommand.stdout.on("data", (data) => {
-    console.log(data.toString("utf8"))
-    allMarkdownFilesBuffer = data
-  })
+    allMarkdownFilesCommand.stdout.on("data", (data) => {
+      console.log(data.toString("utf8"))
+      allMarkdownFilesBuffer = data
+    })
 
-  allMarkdownFilesCommand.stderr.on("data", (data) => {
-    console.log(data.toString("utf8"))
-  })
+    allMarkdownFilesCommand.stderr.on("data", (data) => {
+      reject(data)
+    })
 
-  allMarkdownFilesCommand.on("error", (error) => {
-    console.error("SHITT", error)
+    allMarkdownFilesCommand.on("error", (error) => {
+      reject(error)
+    })
   })
 }
 
-getAllMarkdownFiles()
+await getAllMarkdownFiles()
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -93,6 +93,7 @@
     "format": "prettier --write \"**/*.{js,jsx,json,md}\"",
     "generate-heading-ids": "node src/scripts/generate-heading-ids.js",
     "merge-translations": "node src/scripts/merge-translations.js",
+	"check-markdown-language-tag": "node src/scripts/checkMarkdownLanguage.mjs",
```
SNikhill, line 96: This is a JS file and not a TS file. I can write (and would prefer) typescript but, I didn't want to add the compilation step.

src/scripts/checkMarkdownLanguage.mjs
```diff
@@ -0,0 +1,91 @@
...
+      reject(error)
+    })
+  })
+}
+
+const doAllMarkdownsHaveALanguage = async () => {
+  try {
+    const allMarkdownFilesBuffer = await getAllMarkdownFiles()
+    const markdownFilesWithLanguageBuffer = await getMarkdownFilesWithLanguage()
+    const comparisonResult = Buffer.compare(
```
SNikhill, line 63: The file names are printed to the STDOUT and the same is reported by NodeJS as a Buffer.
If the file names are the same then the Buffer should be the same too.

SNikhill: (Assuming that while this process is running, no other process printed something onto the STDOUT [highly unlikely])

SNikhill:
Please, keep in mind that I am not a NODEJS developer. This is one of the very few scripts I have written and hence I apologise if I am using an unconventional pattern.

src/scripts/checkMarkdownLanguage.mjs
```diff
@@ -13,9 +13,9 @@ const getMarkdownFilesWithLanguage = () => {
     const markdownFilesWithLanguageCommand = spawn("git", [
       ...commonDiffCommand,
       "-G",
-      "lang:",
+      "lang:[ ]*\\w+",
```
SNikhill, line 16: Every markdown file under `src/content` needs to have a non-empty lang attribute

src/scripts/checkMarkdownLanguage.mjs
```diff
@@ -38,7 +38,7 @@ const getAllMarkdownFiles = () => {
     const allMarkdownFilesCommand = spawn("git", [
       ...commonDiffCommand,
       "--",
-      "*.md",
+      "src/content/*.md",
```
SNikhill, line 40: Limiting the check to files under `src/content` path

src/scripts/checkMarkdownLanguage.mjs
```diff
@@ -0,0 +1,90 @@
...
+const doAllMarkdownsHaveALanguage = async () => {
+  try {
+    console.log("Checking Markdown Content Files")
+    const allMarkdownFilesBuffer = await getAllMarkdownFiles()
+    const markdownFilesWithLanguageBuffer = await getMarkdownFilesWithLanguage()
+    const comparisonResult = Buffer.compare(
+      allMarkdownFilesBuffer,
+      markdownFilesWithLanguageBuffer
+    )
+    return comparisonResult === 0 ? true : false
```
SNikhill, line 66: Should I also print out the names of the file that didn't match up?

src/scripts/checkMarkdownLanguage.mjs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+const throwError = (message) => {
+  console.log("\x1b[91m%s\x1b[0m", "ERROR", message)
+}
+
+doAllMarkdownsHaveALanguage()
+  .then((result) => {
+    if (result) {
+      process.exitCode = 0
```
SNikhill, line 79: Let me know if a proper success message needs to be displayed.

SNikhill:
@minimalsm 

Thank you for assigning this issue. It was fun to learn about Node and create a script. :)

package.json
```diff
@@ -94,7 +94,7 @@
     "format": "prettier --write \"**/*.{js,jsx,json,md}\"",
     "generate-heading-ids": "node src/scripts/generate-heading-ids.js",
     "merge-translations": "node src/scripts/merge-translations.js",
-	"check-markdown-language-tag": "node src/scripts/checkMarkdownLanguage.mjs",
+    "check-markdown-language-tag": "ts-node --esm src/scripts/checkMarkdownLanguage.mts",
```
SNikhill, line 97: @pettinarip 

I am using the `ts-node` package with the `esm` flag to run the script with ES Module support

SNikhill: (Reference: #6943)

minimalsm:
Hey @SNikhill, thanks again for working on this.

On testing, I am unable to get the hook to throw an error. I tried adding random characters (see screenshot below) and also completely deleting the lang prop from the markdown (in src/content/bridges/index.md) but the checks run green. Am I using it as intended or is there a problem?

Screenshot 2022-07-11 at 14 47 10

samajammin:
Awesome, thanks for this PR!

This makes me think I wonder what we can generalize here in terms of validating markdown files. For instance, every markdown page should (at least) have:
- `title`
- `description`
- `lang`

I also wonder if we could use our schema to help validate this. Curious what you think @pettinarip?

SNikhill:
@minimalsm Thank you for catching that bug with the numbers. The Regex has been fixed to capture just letters.

Make sure that you **stage** the markdown file after you make changes to it.

SNikhill:
@samajammin 

That is a good suggestion and I believe it can be done but, I am not sure if `git diff` would be the right approach for it.

SNikhill:
@minimalsm 

You are right that there is a caveat that if `lang` for an existing file is removed, it is not detected.

`diff` with `-G` seems to match even if a line with a matching regex pattern was removed (which is basically what is happening in the test case).

I shall create a fix and let you know.

SNikhill:
> Hey @SNikhill thanks for the PR!
> 
> I had a similar experience where the script is not throwing any error when I remove the `lang` prop and/or set something like `lang: asdasdasd`.
> 
> For now, I think this is a practical solution to the problem.
> 
> However, I think a more robust one would be to dig into using a linter, like ESlint. Would be ideal to have a custom rule which checks for us all these properties.
> 
>     * We could enable Gatbsy + ESlint: https://www.gatsbyjs.com/docs/how-to/custom-configuration/eslint/
> 
>     * Use something like this -> eslint plugin for md: https://github.com/mdx-js/eslint-mdx/tree/master/packages/eslint-plugin-mdx
> 
>     * And figure out how to set a custom rule to validate frontmatter properties.

I agree, that a custom lint rule would definitely be a more robust solution.

I shall update you as I improve the hook and possibly create the rule.

SNikhill:
@pettinarip and @minimalsm 

I wanted to try out something and have modified the hook.

- It will now only consider newly created markdown files.

My previous test scenario was flawed. I failed to consider what would happen if the `git diff` command produced no output.

Now the hooks will check for the `empty output` case and will also list out the progress and the files found.

minimalsm:
As discussed @SNikhill, marking this as blocked until #7088 is implemented to see if we can use one of those functions as an export to run a commit hook here.

SNikhill:
Understood @minimalsm

I shall keep an eye on the other PR and I shall make the changes as it is reviewed and merged. :)

minimalsm:
As discussed, we no longer really have this issue due to the markdown checker being run automatically on crowdin-import. Closing this out for now but I appreciate your work here as it massively informed the changes we made to improve the productivity of this workflow.

PR #6909, opened 2022-07-04
refactor: migrate listed components to typescript 

didoshotev:
## Description

migrate listed components:
src/components/PageHero.js
src/components/SkipLink.js

In PageHero content prop, I have set it to any, since I don't see any TS interface for this type of data. This can be changed once we have defined interface.

Please let me know, if you want to separate `SkipLinkAnchor` into different file, as there are 2 components in the same file (SkipLink) currently.

## Related Issue

Refs: #6392


src/components/PageHero.tsx
```diff
@@ -89,7 +93,20 @@ const StyledButtonLink = styled(ButtonLink)`
...
+  className?: string
+}
+
+const PageHero: React.FC<IProps> = ({
+  content,
+  isReverse = false,
+  children,
+  className,
+}) => {
   const { buttons, title, header, subtitle, image, alt } = content
```
pettinarip, line 110: For now, we could add the content interface in this file with this structure. Could be something like:
```ts
interface IButton {
  path: string
  pathId: string
}

interface IContent {
  buttons?: Array<IButton>
  title: string
  header: string
  subtitle: string
  image: string
  alt: string
}
```

pettinarip (changes requested):
Thanks for the PR @didoshotev! left one comment.

didoshotev:
@pettinarip I have made the requested changes and pulled the new changes from dev. Let me know if something else needs to be added for this PR.

pettinarip (approved):
Thanks @didoshotev. I've pushed a few minor things but overall it looks great! thanks for applying the suggestions.

PR #6937, opened 2022-07-05
Refactor tutorial skill level to use Translation component

minimalsm:
## Description
Currently, most non-English languages inconsistently translate our tutorial skill levels (Beginner, Intermediate, Advanced). This is because, for every tutorial, the translator must decide how to translate the skill tag.

For example:
- [French tutorial](https://ethereum.org/fr/developers/tutorials/) use _moyen_ (medium) and _intermédiaire_ for intermediate
- [Chinese simplified ](https://ethereum.org/zh/developers/tutorials/)uses _入门_ (getting started) and _初学者_ (beginner) for beginner 
- [Chinese simplified](https://ethereum.org/zh/developers/tutorials/) uses _中级_ (intermediate) and _中等难度_ (medium difficulty) for intermediate
- [Italian tutorials](https://ethereum.org/it/developers/tutorials/) use _principante_ (beginner) and _base_ (basic) for beginner

This PR puts the skill levels into the page-developers-tutorial JSON to allow them to be translated once (consistently). 

## Related Issue
None


src/pages/developers/tutorials.js
```diff
@@ -213,6 +213,30 @@ const published = (locale, published) => {
...
+  const lowercasedSkill = skill.toLowerCase()
+
+  if (Skill.BEGINNER === lowercasedSkill) return SkillTranslationId.BEGINNER
+  if (Skill.INTERMEDIATE === lowercasedSkill)
+    return SkillTranslationId.INTERMEDIATE
+  if (Skill.ADVANCED === lowercasedSkill) return SkillTranslationId.ADVANCED
+
+  // Returning the original string to preserve translations until we are ready to remove them.
+  return skill
+}
```
wackerow, line 238: @minimalsm If I am correct, this results in the same logic, simplified a bit:
```suggestion
const getSkillTranslationId = (skill) => {
  if (Object.values(Skill).includes(skill.toLowerCase()))
    return `page-tutorial-${lowercasedSkill}`
  // Returning the original string to preserve translations until we are ready to remove them.
  return skill
}
```

wackerow (commented):
Goal here seems to be to restrict the "skill" frontmatter property to English-only on all intl tutorials, using one of these three options, correct? Just noting the way it's set up will only match an intl string if it's the _English_ string 'beginner,' 'intermediate,' or 'advanced."

Two questions
1. When do we convert this to TypeScript? Should we just do this now?
2. When do we go through and tidy up the markdown files here to only use these three options? Should we just do this now?

src/pages/developers/tutorials.js
wackerow: This could shorten it even further :-)
```suggestion
const getSkillTranslationId = (skill) =>
  Object.values(Skill).includes(skill.toLowerCase())
    ? `page-tutorial-${lowercasedSkill}`
    : skill
```

minimalsm:
> Goal here seems to be to restrict the "skill" frontmatter property to English-only on all intl tutorials, using one of these three options, correct? Just noting the way it's set up will only match an intl string if it's the _English_ string 'beginner,' 'intermediate,' or 'advanced."

Yes, it should only match English terms. We should then hide the string in Crowdin so it isn't translated.

> 1. When do we convert this to TypeScript? Should we just do this now?

[Someone is already assigned to refactoring it](https://github.com/ethereum/ethereum-org-website/issues/6392).

> 2. When do we go through and tidy up the markdown files here to only use these three options? Should we just do this now?

Could do, it wasn't a priority with this but we could quite easily pull all of the 'correct' skill tag translations and add them manually to the corresponding languages JSON file. If we do I think we should throw a console.warn in to let us know of any we might miss. 

The potentially slow part is: should we double-check with native speakers that we're picking their preferred terminology for this?

src/pages/developers/tutorials.js
minimalsm: > This could shorten it even further :-)

I think shortening it this much makes it a little obscure/hard to read 🤔 What do you think?

wackerow: Depends if you're used to ternary expressions, implicit arrow function returns and boolean logic I suppose =)
Not a hill I'm dying on, but would just argue the original setup has a fair amount of redundancy

wackerow:
Personally don't think it's critical to "double check" first... we have translations to work with and could go with the majority to start. We place those into `.json` files as intl strings, and simultaneously we'll upload the latest English `.json` file with these new strings, "Beginner" "Intermediate" "Advanced", and eventually that will be translated and will override if we were "wrong."

I'm fine either way, including that cleanup in this PR or separately. Don't think it'll be too much work.

wackerow:
Noticing this doesn't touch the actual tutorial page, leaving them in English on those pages... will patch this before we merge

src/components/TutorialMetadata.tsx
```diff
@@ -38,6 +38,7 @@ const PillContainer = styled.div`
 const SkillPill = styled(Pill)`
   align-self: flex-start;
   margin-bottom: 0.5rem;
+  white-space: nowrap;
```
wackerow, line 41: Added this because Chinese characters were wrapping every character. Open to suggestions, since this may cause issues in other languages if these terms are longer.

PR #6943, opened 2022-07-06
Add crowdin-import.js node script

wackerow:
### Description
- Adds `crowdin-import.ts` ts-node script to `/src/scripts/`
- Expects unzipped contents of Crowdin export download to be copied into `./.crowdin` folder
- Adds package script `yarn crowdin-clean`: Deletes/regenerates the root `.crowdin` folder
- Adds package script `yarn crowdin-import`: Runs `crowdin-import.ts` script
- Package `minimist` added, implementing optional `-v` or `-f` flag parameters (see script comments)
- `crowdin-import.ts` contains an object `USER_SELECTION` where bucket number can be listed for import. Running the above yarn command will copy all selected translation bucket markdown/json files to their corresponding locations within the repo.
- `USER_SELECTION` object designed as a template to make it easy to just open, add the bucket numbers, save, and execute. Empty languages/buckets are simply ignored.


src/scripts/crowdin-import.js
```diff
@@ -0,0 +1,369 @@
+const { copyFileSync, existsSync, mkdirSync, readdirSync } = require("fs")
+const { resolve, join } = require("path")
+
+/**
+ * 0. Console summary flags
+ */
+const VERBOSE = false
+const FULL_BUCKET_NAME_SUMMARY = false
```
pettinarip, line 8: Would be nice if we could pass these params as args to the script. You could install something like https://www.npmjs.com/package/minimist to help you with that

src/scripts/crowdin-import.js
```diff
@@ -0,0 +1,369 @@
...
+ * ie: /Home/alice/Downloads/Ethereum.org Translations (translations)/
+ */
+const CROWDIN_ROOT_OVERRIDE = ""
+
+/**
+ * 2. Select languages and buckets to import by adding the number of the
+ * corresponding content bucket to the `buckets` array
+ * ie. `buckets: [1, 10],` would import the "Homepage" and "Learn" buckets
+ */
+const SELECTION = [
```
pettinarip, line 22: Nice to have. Migrate this to TS and run it using `ts-node`.

With TS we could could read all the langs from our `src/utils/languages.ts` and keep this in sync. Just an idea:

```ts
const baseBucketsPerLang = supportedLanguages.reduce((res, lang) => {
  return {
    [lang]: []
  }
}, {})

const userSelection = {
  // overwrite w user selection
  en: [1, 10]
}

// or transform this to an array as you have
const bucketsPerLang = {
  ...baseBucketsPerLang,
  ...userSelection
}
```

src/scripts/crowdin-import.js
```diff
@@ -0,0 +1,369 @@
...
+    lang: "zh-tw",
+    buckets: [],
+  },
+]
+/**
+ * 3. Execute script by running `yarn crowdin-import`
+ */
+
+/******************************
+ * Primary script logic below *
```
pettinarip, line 217: Would be nice to use this block to explain how to use the script (in case we add arguments) and what the script does from a black box perspective (inputs & outputs).

src/scripts/crowdin-import.js
```diff
@@ -0,0 +1,369 @@
...
+    const source = resolve(`${_path}/${item}`)
+    if (item.endsWith(".json")) {
+      const jsonDestinationPath = join(
+        repoRoot,
+        "src",
+        "intl",
+        repoLangCode,
+        item
+      )
+      if (VERBOSE)
```
pettinarip, line 285: Nice to have.
```js
function log(message) {
  if(VERBOSE) {
    console.log(message)
  }
}
```

pettinarip (approved):
This looks great @wackerow I haven't fully test it yet but lgtm so far. I've only added a couple of nice to have.

minimalsm:
@wackerow is this good to go?

src/scripts/crowdin-import.js
wackerow: Just to note here, the `SELECTION` object does not need to contain all the languages; the array is hard-coded this way as a template, to make it easier to just enter the buckets, instead of having to type out the object structure at all.  Building the object using code doesn't really add anything (empty langs will just be ignored).

With our import flow, I don't see us using any base buckets since rarely if ever are we importing an updated bucket for every language at the same time.

That being said, I do like the simplified `userSelection` object. Had gone with an array iteration approach, but think this looks cleaner.

pettinarip: Perfect. Make sense.

src/scripts/crowdin-import.ts
```diff
@@ -0,0 +1,348 @@
...
+ * 1. Copy languages folder from Crowdin export to ./.crowdin
+ *    ie. ./.crowdin/{lang-codes}
+ *    Tip: Run `yarn crowdin-clean` to initialize the `.crowdin` folder. Can
+ *    also be used to erase contents when finished.
+ *
+ * 2. Select buckets to import by adding the number of the corresponding
+ *    content bucket to the chosen language array below
+ *    ie. `es: [1, 10],` would import the "Homepage" and "Learn" buckets for Spanish
+ *
+ * 3. Save file without committing \*
```
pettinarip, line 23:
```suggestion
 * 3. Save file without committing
```

src/scripts/crowdin-import.ts
```diff
@@ -0,0 +1,348 @@
...
+ *
+ * 4. Execute script by running `yarn crowdin-import`
+ *
+ * 5. If successful, copy `GATSBY_BUILD_LANGS={langs}` output and paste in
+ *    your `.env`, then build site to test results.
+ *
+ * Remember: Revert working changes to this file before committing Crowdin import
```
pettinarip, line 30: (Not strong) idea:

Perhaps we could remove this temporary `USER_SELECTION` state (which conflicts with git) in favor of the `.crowdin` folder structure.

Instead of importing everything in the `.crowdin` folder, the user only needs to import the buckets they need, and the script will import everything it reads from that folder. So you would be moving this `USER_SELECTION` state to the folder itself.

--------
However, I prefer the current approach. I think its probably faster from an operational perspective.

src/scripts/crowdin-import.ts
```diff
@@ -0,0 +1,348 @@
...
+ */
+
+type UserSelectionObject = { [key: string]: number[] }
```
pettinarip, line 33: We are using the `Array` syntax in the codebase.
```suggestion
type UserSelectionObject = { [key: string]: Array<number> }
```

src/scripts/crowdin-import.ts
```diff
@@ -0,0 +1,348 @@
...
+  "pt-br": "pt-BR",
+  pt: "pt-PT",
+  ml: "ml-IN",
+  sr: "sr-CS",
+}
+/**
+ * Names for each bucket in order, zero indexed.
+ * Used for lookup in summary if FULL_BUCKET_NAME_SUMMARY (-f,--full) flag enabled.
+ */
+const BUCKET_NAMES: string[] = [
```
pettinarip, line 144:
```suggestion
const BUCKET_NAMES: Array<string> = [
```

pettinarip (approved):
Just tested it, it worked perfectly 💯

package.json
```diff
@@ -51,6 +51,7 @@
     "is-relative-url": "^3.0.0",
     "lodash": "^4.17.21",
     "luxon": "^1.24.1",
+    "minimist": "^1.2.6",
```
samajammin, line 54: This can probably go under `devDependencies` if it's just for scripts/CLI stuff.

src/scripts/crowdin-import.ts
```diff
@@ -0,0 +1,348 @@
...
+ * 4. Execute script by running `yarn crowdin-import`
+ *
+ * 5. If successful, copy `GATSBY_BUILD_LANGS={langs}` output and paste in
+ *    your `.env`, then build site to test results.
+ *
+ * Remember: Revert working changes to this file before committing Crowdin import
+ */
+
+type UserSelectionObject = { [key: string]: number[] }
+const USER_SELECTION: UserSelectionObject = {
```
samajammin, line 34: More of a nice-to-have but perhaps we could generate this object from `src/utils/languages.ts` to ensure this stays in sync with the languages we support?

wackerow: @samajammin Yes... originally implemented this verbose hard-coded form intentionally to make it easy for users to just add numbers to the array without having to setup the object themself in any way. 

But, just left a comment above responding to Pablo where I have some ideas for v2, where we generate a git-ignored file using a third script (something like `crowdin-setup`) which could generate an object based on the TS langs available, making the full object still available for the user, just not in the `crowdin-imports.ts` file anymore.

This would solve this issue, and also the issue with asking users to alter a git-tracked file but to essentially try not to commit these changes.

package.json
wackerow: Good call, thanks

src/scripts/crowdin-import.ts
```diff
@@ -0,0 +1,348 @@
...
+ * @param repoLangCode Language code used within the repo
+ * @returns void
+ */
+const scrapeDirectory = (
+  _path: string,
+  contentSubpath: string,
+  repoLangCode: string
+): void => {
+  if (!existsSync(_path)) return
+  const ls: string[] = readdirSync(_path).filter(
```
wackerow, line 207:
```suggestion
  const ls: Array<string> = readdirSync(_path).filter(
```

src/scripts/crowdin-import.ts
```diff
@@ -0,0 +1,348 @@
...
+      )
+    }
+  })
+}
+
+// Filter out empty requests and map selection to usable array
+type SelectionItem = {
+  repoLangCode: string
+  crowdinLangCode: string
+  buckets: number[]
```
wackerow, line 258:
```suggestion
  buckets: Array<number>
```

src/scripts/crowdin-import.ts
```diff
@@ -0,0 +1,348 @@
...
+}
+const importSelection: SelectionItem[] = Object.keys(USER_SELECTION)
```
wackerow, line 260:
```suggestion
const importSelection: Array<SelectionItem> = Object.keys(USER_SELECTION)
```

src/scripts/crowdin-import.ts
```diff
@@ -0,0 +1,348 @@
...
+    }
+    // Initialize working directory and check for existence
+    const _path: string = join(crowdinRoot, crowdinLangCode)
+    if (!existsSync(_path)) {
+      trackers.langs[
+        repoLangCode
+      ].error = `Path doesn't exist for lang ${crowdinLangCode}`
+      return
+    }
+    const langLs: string[] = readdirSync(_path)
```
wackerow, line 291:
```suggestion
    const langLs: Array<string> = readdirSync(_path)
```

src/scripts/crowdin-import.ts
```diff
@@ -0,0 +1,348 @@
...
+      // Update tracker
+      trackers.langs[repoLangCode].buckets.push(BUCKET_NAMES[bucket - 1])
+    })
+  }
+)
+
+// Construct console summary
+type SummaryItem = {
+  repoLangCode: string
+  buckets: string[] | number[]
```
wackerow, line 311:
```suggestion
  buckets: Array<string> | Array<number>
```

src/scripts/crowdin-import.ts
```diff
@@ -0,0 +1,348 @@
...
+  jsonCopyCount: number
+  mdCopyCount: number
+  error?: string
+}
+const summary: SummaryItem[] = importSelection.map(
```
wackerow, line 316:
```suggestion
const summary: Array<SummaryItem> = importSelection.map(
```

src/scripts/crowdin-import.ts
pettinarip: I like that idea. Would make the process faster.

I also think that the 3rd script (`cleanup`) is not necessary. Every time you call the `setup`, should regenerate the `.crowdin` folder for you keeping the hole process simpler and minimal.

PR #6952, opened 2022-07-06
Convert p & r files

vdusart:
Moving files starting with letter "P" and "R" to TS

# Related Issue


pettinarip (approved):
Looks great @vdusart. Thanks for the PR. I've only added a couple of minor changes.

PR #6953, opened 2022-07-06
Fix types

pettinarip:
Fix some type bugs introduced in the latest TS migrations.

### DIFF
--- a/src/templates/staking.tsx
+++ b/src/templates/staking.tsx
@@ -422,8 +422,8 @@ const StakingPage = ({
       </HeroContainer>
       <Page dir={isRightToLeft ? "rtl" : "ltr"}>
         <PageMetadata
-          title={mdx.frontmatter.title}
-          description={mdx.frontmatter.description}
+          title={mdx.frontmatter.title || ""}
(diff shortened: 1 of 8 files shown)


minimalsm (approved):

src/templates/staking.tsx
```diff
@@ -344,11 +344,12 @@ const StakingPage = ({
...
   location,
 }: PageProps<Queries.StakingPageQuery, Context>) => {
-  if (!mdx?.frontmatter) {
+  if (!mdx?.frontmatter || !mdx?.frontmatter?.title)
     throw new Error(
       "Staking page template query does not return expected values"
     )
-  }
+  if (!mdx?.frontmatter?.title)
+    throw new Error("Required `title` property missing for staking template")
```
wackerow, line 352: @pettinarip Do you want to add this `OR` statement here? Just noting as-is, this second `if` statement will never run

ie: If (!mdx?.frontmatter?.title) => Throws the first error, and will never get to the second

pettinarip: Ouch 🤦🏼 fixing!

PR #6956, opened 2022-07-06
Remove Gatsby warning on jobs.tsx

minimalsm:
## Description

Hiding this warning by commenting out `jobs.tsx` when not in use.

Screenshot 2022-07-06 at 17 27 25

## Related Issue
None


wackerow (approved):
Hm, feel like there may be a more elegant way to fix this where it simply conditional on if we have any markdown files using the `job` template, but I suppose this works in the meantime.

@pettinarip Any thoughts on this?

I don't think the `StaticQuery` suggested in the error would be appropriate, since the query depends on the markdown file path being passed.

pettinarip:
I would just remove the template if we are not using it. In case we want to use it again, we can `git revert {commit}` it.

PR #6958, opened 2022-07-06
Add missing sharding breadcrumbs

minimalsm:
## Description
"Shard-chain" breadcrumbs were removed when we updated ethereum.org but "sharding" wasn't added to `common.json`

Sharding

## Related Issue
None


wackerow:
Added a commit to fix this for all the intl versions as well. This string will still need to be translated to "sharding", but we can at least fallback to the translation for "shard chains" in the meantime

wackerow (approved):

PR #6992, opened 2022-07-10
Updated MergeArticle List

shelleyolivia:
## Description

Updated the list with relevant files about the merge, particularly around faucet changes and technical details that include definitions, timing, and merge steps. I organized the list to be in alphabetical order based on contributor.

## Preview

https://ethereumorgwebsitedev01-shelleyoliviaethereumorgw95126.gtsb.io/en/upgrades/merge/#further-reading

### DIFF
--- a/src/components/MergeArticleList.tsx
+++ b/src/components/MergeArticleList.tsx
@@ -7,10 +7,20 @@ const Container = styled.div`
 `
 
 const reads: Array<CardListItem> = [
+   {
+    title: "Technical Aspects of Ethereum Merge",
+    description: "Alchemy",
+    link: "https://www.alchemy.com/overviews/the-ethereum-merge",
+  },
   {
-    title: "Ethmerge",
-    description: "Ethmerge",
-    link: "https://ethmerge.com/",
+    title: "How to Migrate from Ropsten to Goerli",
+    description: "Alchemy",
+    link: "https://www.alchemy.com/overviews/migrate-from-ropsten-to-goerli",
+  },
+  {
+    title: "The Merge is Coming",
+    description: "Alchemy",
+    link: "https://www.alchemy.com/the-merge",
   },
   {
     title:
@@ -37,6 +47,11 @@ const reads: Array<CardListItem> = [
     title: "Engine API specs",
     description: "Ethereum Foundation",
     link: "https://github.com/ethereum/execution-apis/tree/main/src/engine",
+  },
+  {
+    title: "Ethmerge",
+    description: "Ethmerge",
+    link: "https://ethmerge.com/",
   },
 ]
 
--- a/src/components/MergeArticleList.tsx
+++ b/src/components/MergeArticleList.tsx
@@ -7,15 +7,10 @@ const Container = styled.div`
 `
 
 const reads: Array<CardListItem> = [
-   {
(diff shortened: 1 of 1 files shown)


src/components/MergeArticleList.tsx
```diff
@@ -7,10 +7,20 @@ const Container = styled.div`
...
+   {
+    title: "Technical Aspects of Ethereum Merge",
+    description: "Alchemy",
+    link: "https://www.alchemy.com/overviews/the-ethereum-merge",
+  },
   {
-    title: "Ethmerge",
-    description: "Ethmerge",
-    link: "https://ethmerge.com/",
+    title: "How to Migrate from Ropsten to Goerli",
```
minimalsm, line 16: I'm not sure this content is right for the further reading. It's very specific to dapp developers and we probably want to keep the further reading a little more generalised.

src/components/MergeArticleList.tsx
```diff
@@ -7,10 +7,20 @@ const Container = styled.div`
...
+    description: "Alchemy",
+    link: "https://www.alchemy.com/overviews/migrate-from-ropsten-to-goerli",
+  },
+  {
+    title: "The Merge is Coming",
```
minimalsm, line 21: Can we pick one of these links? I think we should use this one as it seems to be the parent page of the 'Technical Aspects of Ethereum Merge' page.

minimalsm (changes requested):
Hey @shelleyolivia, thanks for the PR. Not sure why the build is failing, I've triggered a manual build to try to fix it.

src/components/MergeArticleList.tsx
```diff
@@ -7,10 +7,20 @@ const Container = styled.div`
 `
 
 const reads: Array<CardListItem> = [
+   {
+    title: "Technical Aspects of Ethereum Merge",
+    description: "Alchemy",
+    link: "https://www.alchemy.com/overviews/the-ethereum-merge",
```
samajammin, line 13: My 2 gwei: looks like we already cover pretty much all this content on: https://ethereum.org/en/upgrades/merge/

src/components/MergeArticleList.tsx
samajammin: Agreed, I think this link makes more sense to include in the "Dapp and smart contract developers" section.

src/components/MergeArticleList.tsx
samajammin: I'd say this one offers new content, i.e. testnet timelines but probably isn't needed here - this could be included in the "Dapp and smart contract developers" section given most users won't care about testnets.

shelleyolivia: Hey @minimalsm thanks for the comments, appreciate the speedy review! 

Completely see your point - would you mind if we kept the Migrate from Ropsten + Goerli article plus one of the other pages as it's more general? I recommend the "The Merge is Coming" page as the second article but I'm open. Thanks again!

minimalsm (approved):
Thanks again @shelleyolivia, feel free to open a PR to add the developer resource to the relevant section 😀!

minimalsm:
@all-contributors please add @shelleyolivia for docs, ideas

PR #6994, opened 2022-07-10
TypeScript: Search Component

SNikhill:
## Description

Refactor Search and Input Component to use TypeScript.

## Related Issue

#6392


src/components/Search/index.tsx
```diff
@@ -113,8 +119,9 @@ const StyledHighlight = styled(Highlight)`
   }
 `
 
+//FIXME: Add a strict type for `hit` prop
 const PageHit =
-  (clickHandler) =>
+  (clickHandler: MouseEventHandler): React.FC<{ hit: Record<string, any> }> =>
```
SNikhill, line 124: @pettinarip 

Please let me know if there is a better (& strict) type for the `hit` prop.

src/components/Search/index.tsx
```diff
@@ -158,7 +165,7 @@ const indices = [
 ]
 
 // Validate against basic requirements of an ETH address
-const isValidAddress = (address) => {
+const isValidAddress = (address): boolean => {
   return /^(0x)?[0-9a-f]{40}$/i.test(address)
 }
```
SNikhill, line 171: `connectStateResults` accepts a component as an argument. Currently, this argument is defined in the functional call. I shall prefer that I separate it out and then, pass the component as an Argument

SNikhill: That way I can define proper React.FC type for the same

src/components/Search/index.tsx
```diff
@@ -197,10 +204,18 @@ const Results = connectStateResults(
...
+}
+
+const Search: React.FC<ISearchProps> = ({
+  handleSearchSelect,
+  useKeyboardShortcuts,
+}) => {
   const intl = useIntl()
-  const containerRef = useRef()
-  const inputRef = useRef()
+  const containerRef = useRef<HTMLDivElement>(null)
```
SNikhill, line 217: A Ref needs to be explicitly initiated with `null` otherwise, it will lead to a TypeError as `undefined` is not a part of the type definition for a Mutuable RefObject

src/components/Search/Input.tsx
```diff
@@ -57,17 +57,31 @@ const SearchSlash = styled.p`
...
 
-  const handleInputChange = (event) => {
+  const handleInputChange = (event: ChangeEvent<HTMLInputElement>): void => {
     const value = event.target.value
     refine(value)
     setQuery(value)
   }
 
-  const handleSubmit = (event) => {
+  const handleSubmit = (event: FormEvent<HTMLFormElement>): void => {
```
SNikhill, line 84: This could have been a `SubmitEvent` if this handler was explicitly assigned to an `input` of type submit.
But, in this case, `handleSubmit` is assigned to the `Form` element.

src/components/Search/Input.tsx
```diff
@@ -57,17 +57,31 @@ const SearchSlash = styled.p`
   }
 `
 
-const Input = ({ query, setQuery, refine, inputRef, ...rest }) => {
+interface IInputProps {
+  query: string
+  setQuery: (arg: string) => void
+  refine: (arg: string) => void
+  inputRef: React.MutableRefObject<HTMLInputElement>
+  [key: string]: unknown
```
SNikhill, line 65: I have added `[key: string]: unknown` because the Component seems to support other unknown properties as denoted by the usage of the rest operator in the component's signature.

src/components/Search/Input.tsx
```diff
@@ -57,17 +57,31 @@ const SearchSlash = styled.p`
   }
 `
 
-const Input = ({ query, setQuery, refine, inputRef, ...rest }) => {
+interface IInputProps {
+  query: string
+  setQuery: (arg: string) => void
```
pettinarip, line 62:
```suggestion
  setQuery: (query: string) => void
```

src/components/Search/Input.tsx
```diff
@@ -57,17 +57,31 @@ const SearchSlash = styled.p`
...
+  refine: (arg: string) => void
```
pettinarip, line 63: To be more explicit and clear
```suggestion
  refine: (query: string) => void
```

src/components/Search/Input.tsx
pettinarip: Hmm yes, lets not do that. It seems that here we are spreading `styled` props. Lets extend the props with the styled component's props:
```ts
interface IInputProps extends React.ComponentProps<typeof StyledInput> {
  ...
}
```

src/components/Search/index.tsx
```diff
@@ -1,4 +1,10 @@
-import React, { useState, useRef } from "react"
+import React, {
+  useState,
+  useRef,
+  RefObject,
+  EventHandler,
```
pettinarip, line 5: ✂️  unused imports

src/components/Search/index.tsx
```diff
@@ -197,10 +204,18 @@ const Results = connectStateResults(
...
+  useKeyboardShortcuts,
+}) => {
   const intl = useIntl()
-  const containerRef = useRef()
-  const inputRef = useRef()
+  const containerRef = useRef<HTMLDivElement>(null)
+  const inputRef = useRef<HTMLInputElement>(null)
   const [query, setQuery] = useState(``)
   const [focus, setFocus] = useState(false)
   const algoliaClient = algoliasearch(
```
pettinarip, line 221: Lets install the types for algolia. `yarn add -D @types/react-instantsearch-dom`

src/components/Search/Input.tsx
pettinarip: Make sense. Thanks for the clarification.

src/components/Search/index.tsx
```diff
@@ -158,7 +165,7 @@ const indices = [
 ]
 
 // Validate against basic requirements of an ETH address
-const isValidAddress = (address) => {
+const isValidAddress = (address): boolean => {
```
pettinarip, line 168:
```suggestion
const isValidAddress = (address: string): boolean => {
```

src/components/Search/index.tsx
pettinarip: I just checked it out and we need to define the component's props without `React.FC` (I was having a conflict).

Lets define the props in the next line as
```ts
({ hit }: { hit: Hit<Record<string, any>> }) => {
```
You can import `Hit` from `@algolia/client-search`

src/components/Search/index.tsx
pettinarip: Yes please! lets do that. Use `import type { StateResultsProvided } from "react-instantsearch-core"` as it props 👍🏼

pettinarip (changes requested):
Looking good @SNikhill I've left a couple of comments there.

When you are done, pls run again `yarn type-check`. I'm having the following error:
image

pettinarip (approved):
Great @SNikhill! works perfectly 💯 thanks for the PR. I've only merged `dev` to fix some merge conflicts.

PR #7010, opened 2022-07-11
Glossary hash link patch

wackerow:
## Description
- Patches recent update to remove glossary icon from menu glossary link.
- Only adds to links that include 'glossary' and '#' in the link string, no necessarily only `glossary#`

### DIFF
--- a/src/components/Link.tsx
+++ b/src/components/Link.tsx
@@ -97,7 +97,7 @@ const Link: React.FC<IProps> = ({
 
   const isExternal = to.includes("http") || to.includes("mailto:")
   const isHash = isHashLink(to)
-  const isGlossary = to.includes("glossary#")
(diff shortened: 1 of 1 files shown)


src/components/Link.tsx
```diff
@@ -97,7 +97,7 @@ const Link: React.FC<IProps> = ({
 
   const isExternal = to.includes("http") || to.includes("mailto:")
   const isHash = isHashLink(to)
-  const isGlossary = to.includes("glossary#")
+  const isGlossary = to.includes("glossary" && "#")
```
minimalsm, line 100: `String.includes` only accepts 1 argument so this will ignore "glossary" match any link including "#".

```suggestion
  const isGlossary = to.includes("glossary") && to.includes("#")
```

PR #7022, opened 2022-07-12
Change intl gatsby plugin

pettinarip:
## Description

Replace `gatsby-plugin-intl` with `gatsby-theme-i18n` to fix memory issues.

Test:
https://ethereumorgwebsitedev01-gatsbythemei18n.gtsb.io/

Context:
- `gatsby-plugin-intl` was populating the `pageContext` with all the translation strings. Not good since Gatsby put that data on each `page-data.json` file generated.
- `gatsby-theme-i18n` avoids that by doing a `require` of the messages on the `wrapPageElement` hook. There are 2 packages:
  - https://github.com/gatsbyjs/themes/tree/master/packages/gatsby-theme-i18n
  - https://github.com/gatsbyjs/themes/tree/master/packages/gatsby-theme-i18n-react-intl

TODO:
- [x] fix localized links
- [x] fix banner's logic
- [x] check redirects


gatsby-browser.tsx
```diff
@@ -15,12 +16,52 @@ import "@formatjs/intl-locale/polyfill"
...
+      })
+
+    if (!isLang(detected)) {
+      detected = defaultLanguage
+    }
+
+    const queryParams = search || ""
+    const newUrl = withPrefix(`/${detected}${originalPath}${queryParams}`)
+    window.localStorage.setItem("eth-org-language", detected)
+    window.location.replace(newUrl)
```
pettinarip, line 60: As the comment says, this is only used in dev envs. This was extracted from what the previous plugin was doing https://github.com/wiziple/gatsby-plugin-intl/blob/master/src/wrap-page.js#L74

The new plugin doesn't do this for us.

gatsby-node.ts
```diff
@@ -271,27 +269,22 @@ export const createPages: GatsbyNode<any, Context>["createPages"] = async ({
           const splitSlug = slug.split("/")
           splitSlug.splice(1, 1, lang)
           const langSlug = splitSlug.join("/")
-          createPage({
+          createPage<Context>({
             path: langSlug,
             component: path.resolve(`src/templates/${template}.tsx`),
             context: {
+              language: lang,
```
pettinarip, line 276: Keeping most of these context values for backward compatibility. I'll probably refactor this in next iterations.

gatsby-node.ts
```diff
@@ -271,27 +269,22 @@ export const createPages: GatsbyNode<any, Context>["createPages"] = async ({
...
               isOutdated: false,
               isContentEnglish: true,
-              relativePath: relativePath, // Use English path for template MDX query
-              // Create `intl` object so `gatsby-plugin-intl` will skip
-              // generating language variations for this page
-              intl: {
-                language: lang,
-                defaultLanguage,
-                languages: supportedLanguages,
-                messages: getMessages("./src/intl/", lang),
```
pettinarip, line 290: THE only reason why we are changing the plugin. This is causing memory issues.

i18n/config.json
```diff
@@ -0,0 +1,370 @@
+[
+  {
+    "code": "en",
+    "hrefLang": "en",
+    "name": "English",
+    "localName": "English",
+    "langDir": "ltr",
+    "dateFormat": "MM/DD/YYYY"
```
pettinarip, line 8: In next iterations, we need to start using these values.

package.json
```diff
@@ -62,6 +64,7 @@
     "react-helmet": "^6.1.0",
     "react-icons": "^4.3.1",
     "react-instantsearch-dom": "^6.6.0",
+    "react-intl": "^3.12.1",
```
pettinarip, line 67: Current version is v6. We keep using v3 to avoid changing too many things.

src/components/Link.tsx
```diff
@@ -222,4 +224,17 @@ const Link: React.FC<IProps> = ({
   )
 }
 
+export function navigate(
```
pettinarip, line 227: New plugin doesn't provide this function. Implemented a similar approach as the previous one was doing https://github.com/wiziple/gatsby-plugin-intl/blob/master/src/link.js#L42

wackerow (approved):
@pettinarip This is awesome! To the team, hopped on a call with Pablo to walk through this and test it out. Redirects seems to be working appropriately and all the languages are loading as expected. 

Build this from scratch locally with 16GB RAM with all languages and it built in under 11 minutes, and `node` never used over 2-3 GB RAM at any given time. Just one test case, but this is orders of magnitude faster than before.

Given our `master` branch deploy build is currently failing as a result of immense RAM demands, going to bring this in and get a patch deploy out immediately following.

SOLID WORK!! This is going to help tremendously 🎉 💪

wackerow:
Edit on above: `master` build is _not_ currently failing fortunately, since the re-triggered build ended up passing. This buys some time from rushing a deploy to `master`, but still going to bring this into `dev` so we can start taking advantage of its efficiencies and test it out before it goes live.

gatsby-browser.tsx
```diff
@@ -15,12 +16,52 @@ import "@formatjs/intl-locale/polyfill"
...
+  any,
+  Context
+>["wrapPageElement"] = ({ element, props }) => {
+  const { location, pageContext } = props
+  const { pathname, search } = location
+  const { originalPath } = pageContext
+
+  const [, pathLocale] = pathname.split("/")
+
+  // client side redirect on paths that don't have a locale in them. Most useful
```
samajammin, line 43: @nhsz @wackerow should we take a similar approach on the blog?

gatsby-browser.tsx
```diff
@@ -15,12 +16,52 @@ import "@formatjs/intl-locale/polyfill"
...
+  if (!isLang(pathLocale)) {
+    let detected =
+      window.localStorage.getItem("eth-org-language") ||
+      browserLang({
+        languages: supportedLanguages,
+        fallback: defaultLanguage,
+      })
+
+    if (!isLang(detected)) {
+      detected = defaultLanguage
```
samajammin, line 54: Should we always fallback to `defaultLanguage`? Better option might be to inspect the user's browser preference (that's what `gatsby-plugin-intl` did).

i18n/config.json
```diff
@@ -0,0 +1,370 @@
...
+  },
+  {
+    "code": "ar",
+    "hrefLang": "ar",
+    "name": "Arabic",
+    "localName": "العربية",
+    "langDir": "rtl",
+    "dateFormat": "MM/DD/YYYY"
```
samajammin, line 16: Shouldn't date formatting vary by language?

samajammin: Not sure if we need to rely on these? We have the `getLocaleTimestamp` helper for displaying date formats.

i18n/config.json
```diff
@@ -0,0 +1,370 @@
+[
+  {
+    "code": "en",
+    "hrefLang": "en",
+    "name": "English",
+    "localName": "English",
+    "langDir": "ltr",
```
samajammin, line 7: Doesn't appear that we actually use `langDir` for anything - should we?

samajammin: I know we use the `isLangRightToLeft` helper for that currently.

gatsby-browser.tsx
samajammin: Looks like we already added `browser-lang` as a dependency in package.json anyway?

gatsby-browser.tsx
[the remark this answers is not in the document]
pettinarip: That's a good idea. I think we could do that, yes.

i18n/config.json
pettinarip: Yes, I've added these fields because they were required by the `gatsby-theme-i18n` plugin. For now, we are not using `langDir` and `dateFormat` from here. Those things are still working as before.

- I agree we should not use `dateFormat` from this file. We should keep using what we currently have.
- However, I think we could consume `langDir` from this config file instead of what we are currently doing (function `isLangRightToLeft`).

i18n/config.json
pettinarip: Exactly. Related to prev comment. I haven't done that yet because I didn't want to add too many changes but I think that for `langDir` we should use it from here as the source of truth.

i18n/config.json
samajammin: Gotcha - I'm on board.

PR #7024, opened 2022-07-12
fix arxiv link formatting and add Paul Kohlhaas Desci podcast appearance

ademidun:
## Description

- Fix link formatting for ArXiv
- add Paul Kohlhaas Desci podcast appearance
    - Chose the anchor.fm podcast link because it was the first search result on Google and it links to other podcast providers (Apple, Spotify etc.)


corwintines (approved):
Thanks @ademidun! I think this looks good, thanks for fixing that bug!

PR #7044, opened 2022-07-13
Update: inclusion of SenseiNode

NachoRoizman:
Senseinode is a node operator located in Latin America, we are currently validating on rocketpool, verified on SSV and running our own validators in Brasil,Chile, Mexico and Argentina.


minimalsm:
@all-contributors please add @NachoRoizman for docs

PR #7048, opened 2022-07-13
Remove paragraphs about PoS

timbeiko:
Paragraphs about PoS in the "reduce gas costs" sections lead to confusion for people who aren't familiar with Ethereum's roadmap. For example: https://twitter.com/allenf32/status/1547205465267920896

I think we should aim for clarity, as we do elsewhere, e.g.: https://ethereum.org/en/upgrades/merge/#misconceptions


wackerow (approved):
@timbeiko Thanks for raising this. Fully agree this shouldn't be on the gas page and is likely only adding to confusion.

jmcook1186 (approved):
agree!

Issue #7020, opened 2022-07-12
Spelling mistake and provided link to Sharding doesn't work.


naveen106:
**Describe the bug**
While I was reading this page (shown in the screenshot), I found a hyperlink to _sharding_ on the paragraph for **Scalibility**. The link gives 404 Not Found error in response.
Edit : I have created a Pull Request for spelling mistake, this issue here is for only the link not working correctly.

**Expected behavior**
The link should work or the page should tell **WE'RE WORKING ON IT**

**Screenshots**

![0322](https://user-images.githubusercontent.com/61573478/178497306-ae26d2c4-e486-4db6-907f-24ff6b0553e8.jpg)

**Desktop (please complete the following information):**
Chrome

**Smartphone (please complete the following information):**
Android

naveen106:
@minimalsm  I found out why it was happening, the link provided was not corrected, I've created a pull request which provides the correct link, please look into it.

wackerow:
Appears to be resolved, thanks folks! closing out

PR #7050, opened 2022-07-13
link fixes for layer-2

corwintines:
## Description
- Fix internal links for layer 2 page

## Related Issue
https://github.com/ethereum/ethereum-org-website/issues/7020


wackerow (approved):
Just switched the "More on ___" links to be on their own line without a trailing period to keep more consistent with the rest of the site. 

Separate issue, but we may want to consider changing the wording on the "More on bridges" link to something like "What is a bridge?" (doesn't need to be a blocker here though imo)

PR #7066, opened 2022-07-14
add nav links for zkp page, add a youtube resource, and acronym

corwintines:
## Description
- Add dropdown and footer nav items for zero-knowledge proof page
- Add YouTube resource in further reading section
- Add reference to CRS beside Common Reference String to tie acronym to it.


wackerow (approved):
Thanks @corwintines, good catch on linking out to it! 😆

PR #7088, opened 2022-07-15
Add a markdown checker

minimalsm:
## Description
Testing a generalised markdown checker

## Related Issue
None


src/scripts/markdown-checker.js
```diff
@@ -0,0 +1,219 @@
...
+      }
+    } else {
+      languages[lang].push(file)
+    }
+  }
+
+  return languages
+}
+
+function processFrontmatter(path, lang) {
```
SNikhill, line 89: @minimalsm 

We can directly call this function with the path of the staged markdown to process its frontmatter.

SNikhill: We can either use:
- `git diff`
- [A Node Wrapper over `git diff`](https://www.npmjs.com/package/staged-git-files)
- [nodegit](https://github.com/nodegit/nodegit)

to get the staged files along with their paths.

src/scripts/markdown-checker.js
```diff
@@ -0,0 +1,219 @@
...
+    console.warn(`Missing 'title' frontmatter at ${path}:`)
+  }
+  // Description commented out as there are a lot of them missing :-)!
+  // if (!frontmatter.description) {
+  //   console.warn(`Missing 'description' frontmatter at ${path}:`)
+  // }
+  if (!frontmatter.lang) {
+    console.error(`Missing 'lang' frontmatter at ${path}: Expected: ${lang}:'`)
+  } else if (!(frontmatter.lang === lang)) {
+    console.error(
```
SNikhill, line 103: The only change being that the function will need to `throw` an exception or return `false` so that our caller can end the process with failure.

wackerow (approved):
@minimalsm Nice work! As we discussed, this of course has room to be iterated and improved upon. Given this is an adjunct to an already manual process, I think this is good to pull in and we can continue to update and improve this in time.

PR #7100, opened 2022-07-18
Update tokenpocket wallet info

chendatony31:
## Description
Update TokenPocket wallet's info

### Hardware
hardware

### ENS / eip1559 / custom gas
ens

### swaps
swaps

### rpc
rpc

### scam_protection
scam_protection


src/data/wallets/wallet-data.ts
```diff
@@ -1593,26 +1593,26 @@ const walletData: WalletData[] = [
     macOS: true,
     firefox: false,
     chromium: true,
-    hardware: false,
+    hardware: true,
```
minimalsm, line 1596: TokenPocket isn't a hardware wallet. It has `hardware_supported` already (see line 1602)

```suggestion
    hardware: false,
```

minimalsm (approved):
Thanks @chendatony31 👌 

We've just released our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) also 🥳!

src/data/wallets/wallet-data.ts
chendatony31: Ok, actually TokenPocket has its own hardware wallet, not named as TokenPocket though.

PR #7095, opened 2022-07-16
Upgrading contracts nav

samajammin:
## Description

- Add [Upgrading smart contracts page](https://ethereum.org/en/developers/docs/smart-contracts/upgrading-smart-contracts/) to documentation side nav
- Update page URLs to match conventions (see https://github.com/ethereum/ethereum-org-website/pull/6575/files#r918936013 for details)
- Add redirects for old URLs

## Related Issue
Create the page here #6472 but didn't add it to the site nav 😄

PR #7106, opened 2022-07-18
Remove extra H1 from upgrading smart contracts

minimalsm:
## Description

Noticed we had duplicate H1s here:

Screenshot 2022-07-18 at 10 16 07

## Related Issue
#7095


samajammin:
Nice catch!

Issue #6589, opened 2022-06-07
Add Aztec Network


critesjosh:
Before suggesting a layer 2, make sure you've read [our listing policy](/contributing/adding-layer-2/).

Only continue with the issue if your layer 2 meets the criteria listed there.

If it does, complete the following information which we need to accurately list the layer 2.

**Name**

Aztec

**Logo**

![Aztec_horizontal_lock_up_black](https://user-images.githubusercontent.com/18372439/172247766-ac519d4a-6bc8-4f78-a5d0-cda933c9b7ea.svg)

**Description**

Aztec Network is the first private zk-rollup on Ethereum, enabling decentralized applications to access privacy and scale.

**Website**
https://aztec.network/

**Developer documentation**
https://developers.aztec.network/

**L2Beat page**
https://l2beat.com/projects/aztec/

**Is the project open source?**
Yes, https://github.com/AztecProtocol/aztec-connect

**What category of layer 2 is it? (Optimistic, ZK)**
ZK

**Does the layer 2 use Ethereum Mainnet for data availability?**
Yes, here is the rollup processor https://etherscan.io/address/0x737901bea3eeb88459df9ef1be8ff3ae1b42a2ba

**Which bridges are used to onboard?**
https://zk.money/

**Which exchanges offer support for the project?**
None right now, work in progress

**Link to dapp ecosystem (if applicable)**

**Token contract list (if applicable)**

**What date did this project go live?**
March 3rd 2021, first rollup processor tx --> https://etherscan.io/tx/0x699eb365b1c797084d87ac685bd0276e1d5eed713fe95a0e1e2f2b0b99eb32bd

**Is this project a fork? If yes, which project was forked?**
no

**What wallets support this layer 2?**

you can log in to zk.money with metamask or wallet connect

**How can users and/or developers connect to this layer 2?**

via zk.money, using the sdk or directly via an api endpoint (like https://api.aztec.network/falafel)

**Has this project undergone an external security audit and/or bug bounty?**

Here is the audit report for the upcoming upgrade to the protocol, https://github.com/AztecProtocol/aztec-connect/blob/master/audits/Security%20Audit%20%20Results%20-%20Aztec%20Protocol.pdf

**Has the project sustained its users over time?**

Yes, you can see the transaction history here, https://explorer.aztec.network/

You can see the rollup contract balance over time here, https://etherscan.io/address/0x737901bea3eeb88459df9ef1be8ff3ae1b42a2ba#analytics

[...]

PR #7122, opened 2022-07-18
Add Aztec Network

corwintines:
## Description
- Add Aztec Network to zk-rollup data

## Related Issue
- Addresses: https://github.com/ethereum/ethereum-org-website/issues/6589


src/pages/layer-2.js
```diff
@@ -230,10 +230,8 @@ const Layer2Page = ({ data }) => {
           "https://api.cryptostats.community/api/v1/l2-fees/feeTransferEth?metadata=false"
         )
 
-        // Temporary, but filtering out L2's we arent listing
-        feeData = feeData.data.filter(
-          (l2) => l2.id !== "aztec-network" && l2.id !== "hermez"
-        )
+        // Filter out layer 2 networks not listed
+        feeData = feeData.data.filter((l2) => l2.id !== "hermez")
```
minimalsm, line 234: We should probably explicitly get the fee data we want here instead of filtering out what we don't want. E.g. Metis was recently added to this list, which I believe we don't intend on listing on the layer 2 page.

minimalsm (approved):

src/pages/layer-2.js
corwintines: Good call, will make an issue for this.

PR #7127, opened 2022-07-18
fix: 51% attackers cannot force fraudulent changes

eccentricexit:
The docs incorrectly claims a 51% attacker can force through fraudulent changes, which is a common misconseption.

All a 51% attacker can do is reorder and/or censor transactions. They cannot push fraudulent txes or invalid state transitions.

## Description

This PR makes a small clarification on what is the power of a 51% attacker.


jmcook1186 (approved):
Thanks for catching this!

corwintines (approved):

Issue #6488, opened 2022-05-27
Deep URLs not forwarding to default language


wackerow:
**Describe the bug**
When going to a URL with one directory in the path, the language locale does not need to be included, allowing a link to be language agnostic when sharing and it will redirect accordingly. If the directory is deeper than one level, this results in a 404.

**To Reproduce**

Compare the following:
- https://ethereum.org/upgrades/ (forwards to https://ethereum.org/en/upgrades/ for me at least)
- https://ethereum.org/upgrades/merge (results in 404)
or
- https://ethereum.org/developers/ ✅
- https://ethereum.org/developers/docs/ ❌

**Expected behavior**
Ideally all pages undergo the same default language redirect as a fallback, not just pages in the top path directory.

hursittarcan:
I'm not sure if this will help but: 
I had a quick look, I couldn't find the exact location of the bug. But I believe I might have found the code to look at (not 100% sure). I included a screenshot. 

- https://github.com/ethereum/ethereum-org-website/blob/f73abdab628fbad734225c30f38e9ba3d758e3bb/gatsby-config.ts#L48-L50

hursittarcan:
I'm not sure if this will help but: 
I had a quick look at this and I believe I found the problem. I'm not sure if we can fix this bug programmatically.

- Almost all redirects happen in [redirects.json](https://github.com/ethereum/ethereum-org-website/blob/dev/redirects.json)
- Example given in issue: https://github.com/ethereum/ethereum-org-website/blob/f73abdab628fbad734225c30f38e9ba3d758e3bb/redirects.json#L95-L96
- There is no `"fromPath": "/developers/docs/"` in this file. That's why it gives an error. 
- Adding: `"fromPath": "/developers/", "toPath": "/en/developers/"` will solve the example given in the issue. 

I'm not sure if this is a good way to fix this, since there are many more cases. But I don't see a different approach.

Issue #7097, opened 2022-07-17
Pages 404 if no language path is specified


samajammin:
**Describe the bug**

Noticing that now pages 404 when no languages path is specified, e.g. https://ethereum.org/wallets/find-wallet/

**To Reproduce**

1. Go to https://ethereum.org/wallets/find-wallet/
2. See 404

**Expected behavior**

We should redirect those requests to the appropriate URL 
A) (more complex, better UX) based on language preference of the user's browser
B) (simpler, worse UX) based on the default locale (English)

**Additional context**

I believe this is a result from changing out intl plugin in https://github.com/ethereum/ethereum-org-website/pull/7022

The previous plugin (`gatsby-plugin-intl`) had a nice feature to support automatic redirection based on the user's preferred language in browser provided by [browser-lang](https://github.com/wiziple/browser-lang) - perhaps we should look to mimic that implementation?

samajammin:
FYI @pettinarip assigning to you for now.

PR #7135, opened 2022-07-19
Redirect nested routes to default lang

pettinarip:
1. For each `defaultLang` page we create, lets also create a redirect from the root paths. E.g. for `/en/nft` create a redirect from `/nft` to `/en/nft`.
2. Cleanup the `redirects.json` file. Leave only the custom redirects. The rest are dynamically generated by the previous logic.

Fixes: #6488 #7097


corwintines (approved):
Works well for me! Nice job. 

Screen Shot 2022-07-20 at 4 32 58 PM
Sent a message, but noticed this in the console. Could have always been there and im just noticing it now as well.

pettinarip:
@corwintines Thanks for the review and nice catch! I'll try to fix them before merging it.

PR #6356, opened 2022-05-17
404 redirect bug

corwintines:
## Description
- Fix redirects on pages that dont exist to properly 404

https://6295310882fe9e2a5c53ad57--ethdotorg-corwin-test.netlify.app/en/testsestse
Can test with the above link, personal Netlify deploy to test production behaviour. 

Screen Shot 2022-05-30 at 2 27 39 PM

PR #7137, opened 2022-07-19
Resolve 404 pages

### LINKED ISSUE #5682: Incorrect response codes
**Describe the bug**

Seeing odd behaviour in our site response codes. Seems the site is not always returning 404 response code when it should.

**To Reproduce**

All of these I'd expect a 404 response:
- Go to https://ethereum.org/derp (404 response)
- Go to https://ethereum.org/en/upgrades/asdfasdf/ (200 response if 1st request, 304 on subsequent request)
- Go to https://ethereum.org/en/derp (200 response if 1st request, 304 on subsequent request)

**Expected behavior**

URLs should return a 404 response if no page exists.

**Screenshots**

Image 2022-03-15 at 2 36 24 PM

All

All

**Additional context**

May be related to #5681 but I'm not sure.

**Want to contribute?**

pettinarip:
**IMPORTANT**: this PR was branched from #7135. Merge that PR first.

---
Testing site: https://pettinarip-eth-org.netlify.app/

We are implementing the Netlify docs: https://docs.netlify.com/routing/redirects/redirect-options/#custom-404-page-handling

Fix #5682 
Also cover #6356

### DIFF
--- a/gatsby-browser.tsx
+++ b/gatsby-browser.tsx
@@ -22,6 +22,7 @@ import {
   defaultLanguage,
   isLang,
 } from "./src/utils/languages"
+import { IS_DEV } from "./src/utils/env"
 import { Context } from "./src/types"
 
 // Default languages included:
@@ -42,7 +43,7 @@ export const wrapPageElement: GatsbyBrowser<
 
   // client side redirect on paths that don't have a locale in them. Most useful
   // on dev env where we don't have server redirects
-  if (!isLang(pathLocale)) {
+  if (IS_DEV && !isLang(pathLocale)) {
     let detected =
       window.localStorage.getItem("eth-org-language") ||
       browserLang({
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -230,9 +230,19 @@ const config: GatsbyConfig = {
       },
     },
     // Needed for Gatsby Cloud redirect support
-    `gatsby-plugin-gatsby-cloud`,
+    {
+      resolve: `gatsby-plugin-gatsby-cloud`,
+      options: {
+        generateMatchPathRewrites: false,
+      },
+    },
     // Creates `_redirects` & `_headers` build files for Netlify
-    `gatsby-plugin-netlify`,
+    {
+      resolve: `gatsby-plugin-netlify`,
+      options: {
+        generateMatchPathRewrites: false,
+      },
+    },
   ],
   // https://www.gatsbyjs.com/docs/reference/release-notes/v2.28/#feature-flags-in-gatsby-configjs
   flags: {
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -260,6 +260,14 @@ export const createPages: GatsbyNode<any, Context>["createPages"] = async ({
     // e.g. English file: "src/content/community/index.md"
     // e.g. corresponding German file: "src/content/translations/de/community/index.md"
     if (language === defaultLanguage) {
+      createRedirect({
+        fromPath: slug.slice(3),
+        toPath: slug,
+        isPermanent: true,
+        ignoreCase: true,
+        force: true,
+      })
+
       for (const lang of supportedLanguages) {
         const splitPath = relativePath.split("/")
         splitPath.splice(2, 0, `translations/${lang}`)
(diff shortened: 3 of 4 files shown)


gatsby-browser.tsx
```diff
@@ -42,7 +43,7 @@ export const wrapPageElement: GatsbyBrowser<
 
   // client side redirect on paths that don't have a locale in them. Most useful
   // on dev env where we don't have server redirects
-  if (!isLang(pathLocale)) {
+  if (IS_DEV && !isLang(pathLocale)) {
```
pettinarip, line 46: On prod we have server side redirects. We don't need this. In dev is handy to avoid receiving a 404 page at `/`.

gatsby-node.ts
```diff
@@ -387,6 +416,20 @@ export const onCreatePage: GatsbyNode<any, Context>["onCreatePage"] = async ({
   }
 }
 
+export const onPostBootstrap: GatsbyNode["onPostBootstrap"] = ({ actions }) => {
+  const { createRedirect } = actions
+
+  supportedLanguages.forEach((lang) => {
```
pettinarip, line 422: Creating these 404 redirects once all the pages and other redirects were created. We need them at the end of the `_redirects` file.

gatsby-config.ts
```diff
@@ -230,9 +230,19 @@ const config: GatsbyConfig = {
...
+      options: {
+        generateMatchPathRewrites: false,
+      },
+    },
     // Creates `_redirects` & `_headers` build files for Netlify
-    `gatsby-plugin-netlify`,
+    {
+      resolve: `gatsby-plugin-netlify`,
+      options: {
+        generateMatchPathRewrites: false,
```
pettinarip, line 243: ref https://www.gatsbyjs.com/plugins/gatsby-plugin-gatsby-cloud/#configuration

redirects.json
```diff
@@ -171,30 +95,10 @@
     "fromPath": "/nft/",
     "toPath": "/en/nft/"
   },
-  {
-    "fromPath": "/nfts/",
```
corwintines, line 175: Would we want to keep this redirect? I imagine the nft one above is more likely what was supposed to be removed.

redirects.json
```diff
@@ -47,90 +47,14 @@
...
-  {
-    "fromPath": "/cookie-policy/",
-    "toPath": "/en/cookie-policy/"
-  },
-  {
-    "fromPath": "/languages/",
-    "toPath": "/en/languages/"
-  },
   {
     "fromPath": "/enterprise/",
```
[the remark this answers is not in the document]
corwintines: Or is this one kept because there was 2 😅

gatsby-node.ts
corwintines: So if understand, the _redirects file that is created when building contains all redirects for the pages on the site. The site tries to resolve one of those initial redirects for the url, but if it can't find it then falls back to these 404 redirects being created here? Hopefully, this question makes sense. If it doesn't let me know.

corwintines (approved):
Nice! Glad you got this working!

Tested in both english and spanish. 

Just had one question to double check I understand how this works: https://github.com/ethereum/ethereum-org-website/pull/7137#discussion_r926118747

gatsby-node.ts
pettinarip: Good question.

>  the _redirects file that is created when building contains all redirects for the pages on the site

Yes. Including the 404s. These files (_redirects & redirects.json) are only created for servers. They don't play any role in the local env.

> The site tries to resolve one of those initial redirects for the url, but if it can't find it then falls back to these 404 redirects being created here?

Yes, I think the complete order is:
1. `url/path` match file => display the file
2. `url/path` match redirect => display the redirected file
3. `url/path` doesn't match anything => display default 404 (in English)
    - unless it matches the 404 fall backs for each locale (the ones created here), e.g. `/es/something`, will display the `/es/404` page

gatsby-config.ts
pettinarip: A bit more context on this. Our i18n plugin is generating these `matchPath` routes for each locale, for client side routing purposes:
```js
if (newPage.path.match(/^\/[a-z]{2}\/404\/$/)) {
  // Match all paths starting with this code (apart from other valid paths)
  newPage.matchPath = `/${locale.code}/*`
}
```

If this flag is on, it would write in the `_redirects` file, something like:
```
/en/*  /en/404  200
/es/*  /es/404  200
```

Note the 200 status code. We don't want that. That will create a redirect. We want a 404 status code in there. That is why we are disabling this and creating those rules ourselves.

redirects.json
pettinarip: Gooood catch. Yes! exactly that, thanks. I've removed the wrong one. Fixing..

redirects.json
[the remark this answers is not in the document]
pettinarip: Good catch again! yes, any page that is created by us shouldn't have its declaration here. It is created automatically.

redirects.json
```diff
@@ -171,30 +95,10 @@
...
   {
     "fromPath": "/daos/",
     "toPath": "/en/dao/"
   },
-  {
-    "fromPath": "/defi/",
-    "toPath": "/en/defi/"
-  },
-  {
-    "fromPath": "/layer2/",
```
pettinarip, line 191: This one should also exists. Will revert these last two redirects.

PR #7148, opened 2022-07-20
delete header link markdown

tvanepps:
looks like extra text for a header link that is no longer a header

removes {#full-sync}


minimalsm (approved):
Thanks Trent :-)

Issue #7075, opened 2022-07-15
Epic: post-merge content updates


jmcook1186:
**Epic Description**

The Merge is almost upon us. Ethereum.org has recently been updated to ensure that the content is up-to-date and correct for the final months of pre-merge Ethereum. Now, we will create a branch of ethereum.org with content appropriate for post-merge Ethereum, with the intention to merge this branch in once The Merge is complete.

**Involves**

- Scoping out what content changes, additions, or subtractions are required related to the Merge upgrade (done, tracked in Excel sheet)
- Preparing appropriate content changes,
- Coordinating with external contributors to ensure accuracy
- Timing releases appropriately to provide the community with accurate information about the Merge during and immediately after transition

**Goals**

- Ensure references to proof-of-work are updated to either be historical references, or removed
- Update developer documentation to ensure all fundamentals are accurately represented under the proof-of-stake model
- Update /upgrade/merge and to be reflective of past tense and latest information
- Update /energy-consumption page with updated latest information

**Checkpoints**

[Page Audit Spreadsheet](https://docs.google.com/spreadsheets/d/1Nisi3_FY8nwOlLwPsTPm0jlgtAhypH0-pjatGDMSz7M/edit#gid=0)

The following pages have been identified as requiring post-merge adjustments (tick off when PR merge to `post-merge` branch):

- [ ] /nft/
- [ ] /developers/docs/gas/
- [ ] /developers/docs/consensus-mechanisms/pow/mining/
- [ ] /staking/
- [ ] /developers/
- [ ] /upgrades/
- [ ] /upgrades/merge/
- [ ] /developers/docs/web2-vs-web3/
- [ ] /developers/docs/consensus-mechanisms/pos/
- [ ] /developers/docs/intro-to-ethereum/
- [ ] /developers/docs/evm/
- [ ] /developers/docs/nodes-and-clients/
- [ ] /learn/
- [ ] /eth/
- [ ] /developers/docs/accounts/
- [ ] /developers/docs/dapps/
- [ ] /upgrades/beacon-chain/
- [ ] /developers/docs/transactions/
- [ ] /developers/docs/networks/
- [ ] /developers/docs/consensus-mechanisms/pow/
- [ ] /developers/docs/mev/
- [ ] /developers/docs/blocks/
- [ ] /developers/docs/consensus-mechanisms/
- [ ] /developers/docs/intro-to-ether/
- [ ] /upgrades/shard-chains/
- [ ] /glossary/
[...]

PR #7154, opened 2022-07-21
post-merge updates: intro to ether page

jmcook1186:
## Description

updates `developers/docs/intro-to-ether/index.md` ready for post-merge

## Related Issue

#7075

### DIFF
--- a/src/content/developers/docs/intro-to-ether/index.md
+++ b/src/content/developers/docs/intro-to-ether/index.md
@@ -34,7 +34,7 @@ It is [common](https://www.reuters.com/article/us-crypto-currencies-lending-insi
 
 Minting is the process in which new ether gets created on the Ethereum ledger. The underlying Ethereum protocol creates the new ether, and it is not possible for a user to create ether.
 
-Ether is minted as a reward for each block proposed, and also at every epoch checkpoint for other validator activity related to reaching consensus. The total amount issued depends on the number of validators and how much ether they have staked. This total issuance is divided equally among validators in the ideal case that all validators are honest and online, but in reality it varies based on validator performance. About 1/8 of the total issuance goes to the block proposer, the remainder is distributed across the other validators. Block proposers also receive tips from transaction fees and MEV related income but these come from recycled ether, not fresh issuance.
+Ether is minted as a reward for each block proposed and at every epoch checkpoint for other validator activity related to reaching consensus. The total amount issued depends on the number of validators and how much ether they have staked. This total issuance is divided equally among validators in the ideal case that all validators are honest and online, but in reality, it varies based on validator performance. About 1/8 of the total issuance goes to the block proposer; the remainder is distributed across the other validators. Block proposers also receive tips from transaction fees and MEV-related income, but these come from recycled ether, not new issuance.
 
 ## Burning ether {#burning-ether}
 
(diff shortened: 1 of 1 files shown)


src/content/developers/docs/intro-to-ether/index.md
```diff
@@ -34,7 +34,7 @@ It is [common](https://www.reuters.com/article/us-crypto-currencies-lending-insi
 
 Minting is the process in which new ether gets created on the Ethereum ledger. The underlying Ethereum protocol creates the new ether, and it is not possible for a user to create ether.
 
-Ether is minted when a new block is created on the Ethereum blockchain. As an incentive to build blocks, the protocol grants a reward in each block, incrementing the balance of an address set by the block producer. The block reward has changed over time, and today it is 2 ETH per block. After the merge issuance to each validator depends upon the amount of ether they have staked and their performance.
+Ether is minted in each slot (every 12 seconds). The total amount issued depends on the number of validators and how much ether they have staked. This total issuance is divided equally among validators in the ideal case that all validators are honest and online, but in reality it varies based on validator performance. About 1/8 of the total issuance goes to the block proposer, the remainder is distributed across the other validators. Block proposers also receive tips from transaction fees and MEV related income but these come from recycled ether, not fresh issuance.
```
wackerow, line 37: > Ether is minted in each slot (every 12 seconds). 

Would push back on this a bit. Correct me if I'm wrong, but ether isn't necessarily minted each "slot"... Each slot is an opportunity for a block, but is not guaranteed (ie, block proposer is offline), and IIRC, attestation rewards are not credited to the validator balance until the epoch checkpoint.

Perhaps:
```suggestion
Ether is minted as a reward for each block proposed, and also at every epoch checkpoint for other validator activity related to reaching consensus. The total amount issued depends on the number of validators and how much ether they have staked. This total issuance is divided equally among validators in the ideal case that all validators are honest and online, but in reality it varies based on validator performance. About 1/8 of the total issuance goes to the block proposer, the remainder is distributed across the other validators. Block proposers also receive tips from transaction fees and MEV related income but these come from recycled ether, not fresh issuance.
```

Not sure if we should get into block/slot timing in this section, any thoughts?

wackerow (changes requested):
@jmcook1186 Left a comment...

Also, I'm assuming it wasn't intentional to include changes to the `data/translation-reports/**/*.json` files—any chance we can revert these?

jmcook1186:
incorporated your suggestion and reverted changes to `/intl` in forced push.

src/content/developers/docs/intro-to-ether/index.md
```diff
@@ -34,7 +34,7 @@ It is [common](https://www.reuters.com/article/us-crypto-currencies-lending-insi
...
 
-Ether is minted when a new block is created on the Ethereum blockchain. As an incentive to build blocks, the protocol grants a reward in each block, incrementing the balance of an address set by the block producer. The block reward has changed over time, and today it is 2 ETH per block. After the merge issuance to each validator depends upon the amount of ether they have staked and their performance.
+Ether is minted as a reward for each block proposed, and also at every epoch checkpoint for other validator activity related to reaching consensus. The total amount issued depends on the number of validators and how much ether they have staked. This total issuance is divided equally among validators in the ideal case that all validators are honest and online, but in reality it varies based on validator performance. About 1/8 of the total issuance goes to the block proposer, the remainder is distributed across the other validators. Block proposers also receive tips from transaction fees and MEV related income but these come from recycled ether, not fresh issuance.
```
minimalsm, line 37:
```suggestion
Ether is minted as a reward for each block proposed and at every epoch checkpoint for other validator activity related to reaching consensus. The total amount issued depends on the number of validators and how much ether they have staked. This total issuance is divided equally among validators in the ideal case that all validators are honest and online, but in reality, it varies based on validator performance. About 1/8 of the total issuance goes to the block proposer; the remainder is distributed across the other validators. Block proposers also receive tips from transaction fees and MEV-related income, but these come from recycled ether, not new issuance.
```

minimalsm (approved):

PR #7159, opened 2022-07-21
Change the checkbox to radio on the profile chooser box

nloureiro:
We got a suggestion from "RomoPuma" (if you have a Github profile, please add it) on discord (https://discord.com/channels/714888181740339261/951053072258453574/999409307953143932)to change The checkboxes on the personas box to radio-style buttons.

Here are some screenshots of the change.

Screen Shot 2022-07-21 09 42 19 PM

Screen Shot 2022-07-21 09 42 09 PM

### DIFF
--- a/src/components/FindWallet/WalletPersonasSidebar.tsx
+++ b/src/components/FindWallet/WalletPersonasSidebar.tsx
@@ -89,47 +89,6 @@ const Persona = styled.div<{
 
   h3 {
     color: ${(props) => props.theme.colors.text};
-    position: relative;
-    padding-left: 2rem;
-    &:before {
-      content: "";
-      display: block;
-      border-radius: 100%;
-      width: 1.4rem;
-      height: 1.4rem;
-      left: 0;
-      top: 0.1rem;
-      transition: all 0.5s;
-      border: 1px solid;
-      position: absolute;
-      border-color: ${({ selected, isDark, theme }) =>
-        selected
-          ? isDark
-            ? theme.colors.primary
-            : theme.colors.primary
-          : isDark
-          ? theme.colors.white
-          : theme.colors.black400};
-    }
-    &:after {
-      content: "";
-      display: block;
-      border-radius: 100%;
-      width: 0.9rem;
-      height: 0.9rem;
-      left: 0.25rem;
-      top: 0.35rem;
-      transition: all 0.5s;
-      position: absolute;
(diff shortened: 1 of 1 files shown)


filtelescu:
RomoPuma here (also updated my 'about me' on Discord as a verification) -- Looks great! 

I noticed there's a slight animation time between states on the deploy (when selecting 'New to crypto' for example), not sure if that's supposed to be there or not considering on live the transition is instant. 

Super dope :)

nloureiro:
> 

yes. actually was on purpose. I think it improves awareness of the "different" radio button. I think the only downside of this change is awareness. You were right, this is a radio button behavior but in a different look/style/appearance. The check mark would try to make it more obvious. with the animation, I'm trying to gain a bit of the sense that something happen without removing the focus or usability.

I can reduce the time, it's a 1,5s transition.

filtelescu:
Understandable, you're definitely right that the check mark makes the selection more obvious. I do think reducing the time _slightly_ could help keep it consistent with the snappiness of the rest of the interface while still acting as a clear indicator. 

The best of both worlds 😎

corwintines:
I'm going to do some playing around to bring this in as an actual radio input instead of CSS styling if that's ok @nloureiro

minimalsm (changes requested):
Nice :-) agree with the change on the design here 💯

The implementation currently does not allow screen readers to select/unselect or know what is selected from the profiles. To fix this, I think we should break these styles out of the h3 and into a radio button-styled component. What do you think?

corwintines:
@minimalsm I think this should fix up the regression in screen reader functionality. @nloureiro @filtelescu any feedback welcome as well.

samajammin (approved):
LGTM! Though I'd like @minimalsm to confirm he doesn't see any issues w/ screen readers.

minimalsm (approved):
Sorry for blocking :-)

minimalsm:
For context, this was some discussion behind the scenes:

- Not blocking this PR but wanted to highlight we’re doing something unusual with radio buttons here
    - A radio button typically can only be selected, and after one radio button has been selected, another must be selected to unselect the previous (the default state of none selected cannot be re-achieved)
    - In the current implementation, only one option can be chosen, but we also deselect it (so it essentially behaves kinda like a checkbox, kinda like a radio button)
    - This might be a headache if we decide to implement Chakra’s Checkbox component

Issue #7150, opened 2022-07-21
Source homepage code examples from data files


samajammin:
**Is your feature request related to a problem? Please describe.**

Refactor request...

Currently we display code examples on the homepage that are stored as large strings in the component:
https://github.com/ethereum/ethereum-org-website/blob/dev/src/pages/index.tsx#L603

This is pretty gnarly 😅😬 

**Describe the solution you'd like**

We also store these code examples in data files which at least makes them easy to format:
- https://github.com/ethereum/ethereum-org-website/blob/dev/src/data/CreateWallet.js
- https://github.com/ethereum/ethereum-org-website/blob/dev/src/data/SimpleDomainRegistry.sol
- https://github.com/ethereum/ethereum-org-website/blob/dev/src/data/SimpleToken.sol
- https://github.com/ethereum/ethereum-org-website/blob/dev/src/data/SimpleWallet.sol

It'd be great if we could pull the content for these examples directly from these data files 😄 

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

vdusart:
Hey @samajammin, can I do this one please ?

samajammin:
Sure thing, thanks for volunteering @vdusart! Assigning to you 👍

samajammin:
Appreciate all your help with the TS migration BTW! ❤️

pettinarip:
Issue solved by @vdusart. Thanks!

PR #7164, opened 2022-07-22
Load content from data files

vdusart:
Source homepage code examples from data files

## Description

Load the content of the code examples (on homepage) from the corresponding files in `/src/data`

## Related Issue

#7150

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+# The content is used as a code example in /src/pages/index.js
+src/data/CreateWallet.js
--- a/overrides.d.ts
+++ b/overrides.d.ts
@@ -13,7 +13,7 @@ declare module "*.mp4" {
   export default src
 }
 
-declare module "raw-loader*" {
+declare module "!!raw-loader!*" {
   const content: string
   export default content
 }
--- a/overrides.d.ts
+++ b/overrides.d.ts
@@ -14,6 +14,6 @@ declare module "*.mp4" {
 }
 
 declare module "raw-loader*" {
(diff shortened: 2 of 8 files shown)


vdusart:
@samajammin I don't know why but sometimes webpack tries to bundle the `CreateWallet.js` file even if it is in the exclude list inside the `webpack.config.ts` so it creates an error :cry: 
Do you have an idea ?

pettinarip:
> I don't know why but sometimes webpack tries to bundle the `CreateWallet.js` file even if it is in the exclude list inside the `webpack.config.ts` so it creates an error cry Do you have an idea ?

Nice job @vdusart! I don't know the answer to that but I'll play with it a bit and see if I can help you out.

pettinarip:
@vdusart can you explain why do we need/want to exclude this file? I removed the webpack config file you added and everything worked as expected. IMO I don't think we have to exclude it since it is a module that needs to be imported in the page file.

Anyway, here is how you can extend the webpack config in Gatsby https://www.gatsbyjs.com/docs/how-to/custom-configuration/add-custom-webpack-config/

vdusart:
I'm not sure but sometimes it was not working because webpack was trying to bundle the file and as we don't have some modules (like ethers) it was not compiling...
I thought that putting this file in the exclude files from webpack will arrange it but apparently not 😢

vdusart:
In fact you were right @pettinarip the `webpack.config.ts` file is useless, I just removed it :smile:

samajammin (approved):
👏

PR #7167, opened 2022-07-23
Ts migrate selectablecard

Mousticke:
Migrate component to TS.

Set the value of the selected props to false by default


pettinarip (approved):
Thanks @Mousticke!! LGTM

PR #7178, opened 2022-07-25
Refactor/ts scripts

SNikhill:
## Description

Migrate generate-header-ids script to TypeScript.

## Related Issue

- #6392


src/scripts/generate-heading-ids.ts
```diff
@@ -13,16 +13,27 @@ const GitHubSlugger = require("github-slugger")
 
 // TODO we should auto-run this script when new markdown files are added to the project
 
-const toc = {}
+interface IGitHubSlugger {
+  slug: (headingText: string) => string
```
SNikhill, line 17: refactor: I should install `@types/github-slugger`

src/scripts/generate-heading-ids.ts
```diff
@@ -121,12 +136,23 @@ const addHeaderIDs = (lines, write = false) => {
...
+    const headerTextWithSlug = addHeaderID(line, slugger, write)
+    if (headerTextWithSlug) {
+      results.push(headerTextWithSlug)
+    }
   })
   return results
 }
 
-const traverseHeaders = (path, doc = "", write = false) => {
+type TraverseHeadersFunction = (
```
SNikhill, line 147: Defined a `config` object to specify optional parameters.

src/scripts/generate-heading-ids.ts
```diff
@@ -135,27 +161,31 @@ const traverseHeaders = (path, doc = "", write = false) => {
...
 
-const addHeaderIDsForDir = (path) => {
+const addHeaderIDsForDir = (path: string) => {
   if (path.includes("translations")) {
     throw new Error(`Heading ID generation is intended for English files only.`)
   }
   const fullPath = `src/content/${path}`
-  traverseHeaders(fullPath, null, false)
-  traverseHeaders(fullPath, null, true)
+  traverseHeaders(fullPath, { write: false })
```
SNikhill, line 181: **question**: should there be functionality to modify all the markdown files in a directory?
Currently, `headersToChange` just keeps track of header-number in a single markdown file.

src/scripts/generate-heading-ids.ts
```diff
@@ -104,11 +119,11 @@ const addHeaderID = (line, slugger, write = false) => {
   }
 }
 
-const addHeaderIDs = (lines, write = false) => {
+const addHeaderIDs = (lines: Array<string>, write = false): Array<string> => {
   // Sluggers should be per file
-  const slugger = new GitHubSlugger()
+  const slugger: IGitHubSlugger = new GitHubSlugger()
```
SNikhill, line 124: todo: use correct type for GitHub slugger

src/scripts/generate-heading-ids.ts
pettinarip: Hmm good question, that could be interesting. @minimalsm what do you think? (as I don't use very often this script). Could be handy

---
Note: either way, let's not include that potential feature in this PR.

pettinarip (commented):
Good job @SNikhill. Not sure if this was ready for review as there are a couple of TODO comments.

src/scripts/generate-heading-ids.ts
minimalsm: I never really use this either.

src/scripts/generate-heading-ids.ts
SNikhill: @pettinarip 

Are Definitely Typed types detected with a CommonJS require syntax?
I can use `typeof gitHubSlugger` to specify its type, but, can't use `BananaSlug` which is the [name of its type class definition from the @types/github-slugger package](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/github-slugger/index.d.ts#L10)

SNikhill: Or an alternate solution: declare the script as ES6 Module and use the `import` syntax.

pettinarip: Yes, lets try that. Lets use it as an es6 module where you can do these imports using `import`.

Reading a bit, I think you will need to override the tsconfig for `ts-node` to use commonjs module, like this:
```
// in tsconfig.json add
"ts-node": {
  "compilerOptions": {
    "module": "commonjs"
  }
}
```

pettinarip (changes requested):
As mentioned in the last comment, lets try to use es6 imports by configuring the `tsconfig` file.

In addition, lets
- install the `github-slugger` types
- rename this file to `generateHeadingIds.ts` if everything works as expected

package.json
```diff
@@ -100,7 +102,7 @@
     "crowdin-clean": "rm -rf .crowdin && mkdir .crowdin",
     "crowdin-import": "ts-node src/scripts/crowdin-import.ts",
     "format": "prettier --write \"**/*.{js,jsx,json,md}\"",
-    "generate-heading-ids": "node src/scripts/generate-heading-ids.js",
+    "generate-heading-ids": "ts-node --esm src/scripts/generateHeadingIds.mts",
```
SNikhill, line 105: @pettinarip 

I am using `esm` flag to mark it as ES Module along with `mts` extension for the file

src/scripts/generateHeadingIds.mts
```diff
@@ -41,11 +48,15 @@ const walk = (dir, doc) => {
...
-const stripLinks = (line) => {
-  return line.replace(/\[([^\]]+)\]\([^)]+\)/, (match, p1) => p1)
+const stripLinks = (line: string): string => {
+  return line.replace(/\[([^\]]+)\]\([^)]+\)/, (match, p1: string) => p1)
 }
 
-const addHeaderID = (line, slugger, write = false) => {
+const addHeaderID = (
+  line: string,
+  slugger: BananaSlug,
```
SNikhill, line 57: @pettinarip 

Type for GitHub Slugger is now inferred. Class Name: BananaSlug

package.json
```diff
@@ -90,6 +91,7 @@
     "prettier": "^2.2.1",
     "pretty-quick": "^3.1.0",
     "react-test-renderer": "^17.0.1",
+    "ts-node": "^10.9.1",
```
SNikhill, line 94: I had to list ts-node as a `devDependency`.

package.json
SNikhill: I had to manually set the syntax to `typescript` in neovim for this `mts` file to get the syntax highlighting.

src/scripts/generate-heading-ids.ts
SNikhill: @pettinarip 

For now, I am using the `.mts` extension but, if you want then, I think [we can use `moduleTypes`](https://typestrong.org/ts-node/docs/module-type-overrides) to define the module type for this script explicitly.

SNikhill: Also, since module type is inferred from the nearest package.json, we can add a `package.json` in the scripts folder and [specify the type as `module` for all the scripts](https://nodejs.org/docs/latest-v13.x/api/esm.html#esm_enabling).

But, this option would require converting the other scripts to use the ESM import and will go out of the scope of this PR.

pettinarip: Yes, I guess I like to have everything in `.ts` instead of mixing them with different extensions to make it simpler to other developers. But I don't have a strong opinion about that, I think we are fine with using it as `.mts`.

pettinarip (approved):
lgtm. Ty @SNikhill

PR #7179, opened 2022-07-25
Implement emotion & chakra

pettinarip:
## Description

The first step towards adopting Chakra UI is to migrate from `styled-components` to `emotion`.

**UPDATE:** chakra PR was merged #7195 **now** this PR includes chakra + emotion setup.

## Related Issue

Epic #6374
Fixes #7054

Issue #6374, opened 2022-05-17
Epic: Implement UI library


pettinarip:
*This is an epic that makes up part of [the ethereum.org Q3 roadmap](https://github.com/ethereum/ethereum-org-website/issues/7166)*

### **Description**

A major insight from our research into accessibility and our website audits last quarter was that many of the issues we face, like lack of keyboard navigation, support for assistive technology and consistent W3C design patterns can be solved out-of-the-box by adopting an open source UI library.

Ultimately we’ve come to believe that rolling our own robust components is not what we should focus our time on. Leveraging an existing UI library will save our development and design team time to focus more energy on what matters - shipping products for our Ethereum users. This effort also ties in well with our work on the open design system and will serve as the foundation to our component library.

### **Todo**

- [x]  Decide which UI lib to use
    - We are going to use [Chakra](https://chakra-ui.com/) ([see a summary of our research](https://www.notion.so/Research-options-and-select-UI-library-babae0a343e24c92be0d6bd486e35c1c))
- [x]  Finish TS migration of the codebase (#6392)
- [x] #7054
- [x] #7196
- [x] #7374
- [x] #7369
- [x] Bump Chakra to v2 #7617 
- [x] Migrate each one of our UI components
  - [x] First wave
  - [x] #8632
- [x] #9353
- [x] #10562
- [ ] cleanup tasks: clean `SharedStyledComponents`, update docs, and remove deprecated components #10575

### **How to contribute**

If you’d like to help with this migration, please follow these steps:

1. Leave a comment here asking which file you would want to work on.
2. We assign you to the file (by adding your handle next to the file name in the list).
    - Read and follow our [UI migration guide](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/chakra-migration-guide.md) for some tips & guidance.
    - **If you were assigned to multiple files** then create a separate PR for each component. It will make the testing process much simpler and more organized.
3. Once you finish it (PR merged), we’ll mark it as done.
4. Repeat!

----

<details>
  <summary>First wave (completed)</summary>
  
- [x]  `src/components`
    - [x] ~Accordion.tsx @SNikhill~
    - [x]  ActionCard.tsx @SNikhill #8009
    - [x]  AdoptionChart.tsx @soheil555 #9088
    - [x]  AssetDownload.tsx @benlazzero #9103
    - [x]  BannerNotification.tsx @Mousticke #7778
    - [x]  BoxGrid.tsx @soheil555 #9094
[...]

PR #7195, opened 2022-07-26
Chakra UI setup

### LINKED ISSUE #7196: Install Chakra UI
Dependency: `emotion` #7054

pettinarip:
**Note:** branched off from #7179 (depends on `emotion`)

## Description

Initial Chakra UI setup.

Things addressed in this PR:
- Installed Chakra (with Gatsby plugins)
- Overridden default Chakra theme with an initial custom theme
- Reorganized legacy global styles in a new `GlobalStyle` component
- Synced old dark mode state with Chakra color mode state

## Related Issue

Epic #6374 
Fixes #7196


src/@chakra-ui/gatsby-plugin/styles.ts
```diff
@@ -0,0 +1,12 @@
+const styles = {
+  global: (_props) => ({
+    // TODO: remove these overrides as we adopt the new Design System
```
samajammin, line 3: A bit more documentation/context on why this is needed to begin with would be helpful.

samajammin: e.g. why is both this file & GlobalStyle.tsx needed?

src/@chakra-ui/gatsby-plugin/theme.ts
```diff
@@ -0,0 +1,25 @@
...
+// Global style overrides
+import styles from "./styles"
+
+// Foundational style overrides
+import foundations from "./foundations"
+
+const config: ThemeConfig = {
+  cssVarPrefix: "eth",
+  initialColorMode: "light",
+  useSystemColorMode: false,
```
corwintines, line 16: I think we should use the system color mode personally, I enjoy this feature when I use websites.

pettinarip: Agreed. I've changed it 👍🏼

corwintines (approved):
LGTM. Tried to break a bunch of pages with no success yet.

PR #7197, opened 2022-07-26
Create NetworkUpgradeSummary component for history page

minimalsm:
## Description

This PR refactors repeating elements in the History markdown page into a 'NetworkUpgradesComponent'.

Some issues with the previous implementation.

- Even for a minor update, the Crowdin progress gets reset to near zero every time we upload changes to this file due to the complex nature of the syntax.
- Translators must translate the same information many times.
- [The first Emoji component breaks almost every time it gets uploaded from Crowdin](https://github.com/ethereum/ethereum-org-website/pull/6986/files).
- It uses HTML within markdown, which we'd like to avoid.

## Additional context

Currently, I've only updated a few to test there are no visual regressions. We should update the entire history page to use the component if we want to use this approach. 

We could also quite quickly create these JSON strings manually for the languages that already have the history page translated if we felt that would be worthwhile.

## Related Issue
Ongoing translation efforts


src/components/History/NetworkUpgradeSummary.tsx
```diff
@@ -0,0 +1,131 @@
...
+        </Link>
+      </UnstyledListItem>
+    )
+  }
+
+  return (
+    <UnstyledList>
+      <UnstyledListItem>
+        <FormattedEmoji text=":calendar:" />
+        <code>{dateTimeAsString}</code>
```
minimalsm, line 107: There is definitely a better way to do this, but given how we format the dates currently, I felt like this was fine for now.

src/components/History/NetworkUpgradeSummary.tsx
```diff
@@ -0,0 +1,131 @@
+// Libraries
+import React from "react"
+import styled from "styled-components"
+import { FormattedNumber } from "react-intl"
```
minimalsm, line 4: As we're using `react-intl` now, I used FormattedNumber here instead of [the common approach we're using elsewhere](https://github.com/ethereum/ethereum-org-website/blob/43c8b5709da532247c5f7668da25d56daf42e384/src/components/StatsBoxGrid.tsx#L344) as it seemed a lot cleaner. 

Now noticing we don't handle Farsi numbers by doing it this way.

wackerow (commented):
@minimalsm I like this overall! One thought I had, what about storing all of the data we're passing as props in a historical data file, then import that into the `NetworkUpgradeSummary.tsx` component, and just pass some type of id (upgrade `name` would work) to then grab the available data and render the component?

ie: 
```tsx
<NetworkUpgradeSummary name="altair" />
```
...then all of this is fetched within `NetworkUpgradeSummary.tsx`:
```
dateTimeAsString="Oct-27-2021 10:56:23 AM +UTC"
epochNumber={74240}
ethPriceInUSD={4024}
waybackLink="https://web.archive.org/web/20211026174951/https://ethereum.org/en/"
```

I think it would help clean up the props and reduce chances of markdown formatting issues with translations.

minimalsm:
> @minimalsm I like this overall! One thought I had, what about storing all of the data we're passing as props in a historical data file, then import that into the `NetworkUpgradeSummary.tsx` component, and just pass some type of id (upgrade `name` would work) to then grab the available data and render the component?
> 
> ie:
> 
> ```tsx
> <NetworkUpgradeSummary name="altair" />
> ```
> 
> ...then all of this is fetched within `NetworkUpgradeSummary.tsx`:
> 
> ```
> dateTimeAsString="Oct-27-2021 10:56:23 AM +UTC"
> epochNumber={74240}
> ethPriceInUSD={4024}
> waybackLink="https://web.archive.org/web/20211026174951/https://ethereum.org/en/"
> ```
> 
> I think it would help clean up the props and reduce chances of markdown formatting issues with translations.

Awesome idea @wackerow! 

This approach is much cleaner and lets us keep the markdown component to one line ❤️

wackerow:
@minimalsm Any update here? Seems to have gone stall, perhaps we open this up for the community?

src/data/NetworkUpgradeSummaryData.ts
```diff
@@ -0,0 +1,231 @@
...
+}
+
+const NetworkUpgradeSummaryData: NetworkUpgradeSummaryDataProps = {
+  shanghai: {
+    dateTimeAsString: "2023-04-12T22:27:35.000Z",
+    ethPriceInUSD: undefined,
+    waybackLink: undefined,
+    blockNumber: undefined,
+    epochNumber: undefined,
+    slotNumber: undefined,
```
wackerow, line 21: I would suggest we just remove the `undefined` declarations... The data schema is laid out above marking these as optional. @pettinarip Any thoughts/preferences?

wackerow (commented):
@corwintines nice job cleaning up! Left a couple comments but overall lgtm!

src/data/NetworkUpgradeSummaryData.ts
corwintines: I would advocate keeping it just to make copy-pasting for new entries easier.

src/components/History/NetworkUpgradeSummary.tsx
```diff
@@ -0,0 +1,119 @@
...
+      </Flex>
+    )
+  }
+
+  if (epochNumber) {
+    blockTypeTranslation = (
+      <Flex>
+        <Emoji fontSize="sm" mr={2} text=":bricks:" />
+        <Translation id="page-history-epoch-number" />
+        :&nbsp;
```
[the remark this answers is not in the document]
corwintines: I tried using `{' '}` but the space is not there when I do. Needed to use `:nbsp;` in place to get it working in this component.

src/components/History/NetworkUpgradeSummary.tsx
```diff
@@ -0,0 +1,119 @@
...
+      <Flex>
+        <Emoji fontSize="sm" mr={2} text=":bricks:" />
+        <Translation id="page-history-slot-number" />
+        :&nbsp;
+        <Link to={`https://beaconscan.com/slot/${slotNumber}`}>
+          {new Intl.NumberFormat(localeForStatsBoxNumbers).format(slotNumber)}
+        </Link>
+      </Flex>
+    )
+  }
```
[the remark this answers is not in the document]
corwintines: done!

src/components/History/NetworkUpgradeSummary.tsx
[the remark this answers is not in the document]
wackerow: That's odd... we don't rely on this syntax anywhere else in the codebase. Not critical of course, but curious why that wouldn't work.

src/data/NetworkUpgradeSummaryData.ts
wackerow: Hrm, not critical of course, but not a big fan. Adds 53 lines just to leave a template for users. I think the interface should be the source people go to when seeing what properties to declare.

wackerow (approved):
Looks good to merge!

PR #7146, opened 2022-07-20
update Besu link

tvanepps:
previous pegasys link redirects to consensys quorum. this PR updates it to the hyperledger page

PR #7214, opened 2022-07-27
Fix Besu links

minimalsm:
## Description
previous pegasys link redirects to consensys quorum. this PR updates it to the hyperledger page

## Related Issue
#7146


corwintines (approved):
Nice 🚀

Issue #2277, opened 2021-01-19
Integrate kb.beaconcha.in content


samajammin:
**Is your feature request related to a problem? Please describe.**

The folks at beaconcha.in have built out some incredible technical resources on Eth2:
https://kb.beaconcha.in/

The content is open source & free to use:
https://github.com/gobitfly/eth2-knowledge-base/blob/master/LICENSE

Chatted with @Buttaa & confirmed that we're welcome to add their resources on the site. As always, we should link out to them w/ credit as the original source.

**Describe the solution you'd like**

We should discuss & decide the best place for this content. 2 main options

1. Within the Eth2 section: https://ethereum.org/en/eth2/
2. Within the developer docs section: https://ethereum.org/en/developers/docs/

Curious to hear what folks think. My take: given their technical focus & that eventually, "Eth2" resources will eventually become just "Ethereum" dev resources, I learn toward putting them within the developer docs. On the side navigation we currently have an "Advanced" section. We could store them there or create a new "Eth2" section.

**Describe alternatives you've considered**
Just link out to kb.beaconcha.in pages where appropriate

**Context**

Part of this epic: #7075

samajammin:
Some thoughts on specific pages:
- https://kb.beaconcha.in/glossary : add individual terms to https://ethereum.org/en/glossary/ (if they don't yet exist) with the preface that it's related to Eth2 & link to https://ethereum.org/en/eth2/
- https://kb.beaconcha.in/ethereum-2.0-blocks : add terms to Eth2 block explorer (if they don't yet exist) https://ethereum.org/en/developers/docs/data-and-analytics/block-explorers/#eth2-data
- Port over pages for new Eth2 dev docs section:
   - https://kb.beaconcha.in/staking-and-hardware
   - https://kb.beaconcha.in/ethereum-2-keys
   - https://kb.beaconcha.in/ethereum-2.0-depositing
   - https://kb.beaconcha.in/attestation

Pages I'm not sure about:
- https://kb.beaconcha.in/the-genesis-event (is this important anymore? Perhaps useful knowledge to understand?)
- https://kb.beaconcha.in/depositing-to-ethereum-2.0 (should we just refer folks to https://launchpad.ethereum.org/)
- https://kb.beaconcha.in/rewards-and-penalties (great topic but needs content)

Buttaa:
Hi @samajammin , 

here are additional ideas which I did not have time for yet

- How rewards work
- improving inclusion delays
[...]

PR #7215, opened 2022-07-27
add page on validator keys from kb.beaconchain

jmcook1186:
## Description

adds a new page to `/developers/docs/consensus-mechanisms/pos/` that explains types of keys in pos Ethereum.
This is part of the migration of material from kb.beaconchain, although the content is significantly modified.

## Related Issue

#2277


minimalsm (commented):
Not sure if it's your linter or just copying from the other markdown file, but we should ensure that markdown lines do not have a limit on line length. [See example.](https://github.com/ethereum/ethereum-org-website/pull/7215/commits/4269571efc54151fb590ab78bb5726ba4f9264d7)

If we upload markdown files with fixed line lengths to Crowdin, it makes it harder on translators and reviewers because it fragments the string.

minimalsm (changes requested):
Ran out of time reviewing this but just need to double check we don't have any spelling mistakes in here.

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,90 @@
...
+
+Each branch is separated by a `/` so `m/2` means start with the master key and follow branch 2. In the schematic below a single mnemonic phrase is used to store three withdrawal keys, each with two associated validators.
+
+![validator key logic](multiple-keys.png)
+
+## Further Reading {#further-reading}
+
+[Ethereum Foundation blog post by Carl Beekhuizen](https://blog.ethereum.org/2020/05/21/keys/)
+
+[EIP-2333 BLS12-381 key generation](https://eips.ethereum.org/EIPS/eip-2333)
```
minimalsm, line 90: Further reading should always be in a markdown list

```suggestion
- [Ethereum Foundation blog post by Carl Beekhuizen](https://blog.ethereum.org/2020/05/21/keys/)
- [EIP-2333 BLS12-381 key generation](https://eips.ethereum.org/EIPS/eip-2333)
```

src/data/developer-docs-links.yaml
```diff
@@ -72,9 +72,11 @@
           to: /developers/docs/consensus-mechanisms/pos/
           items:
             - id: docs-nav-gasper
-              to: /developers/docs/consensus-mechanisms/pos/gasper/
+              to: /developers/docs/consensus-mechanisms/pos/gasper
```
minimalsm, line 75:
```suggestion
              to: /developers/docs/consensus-mechanisms/pos/gasper/
```

src/data/developer-docs-links.yaml
```diff
@@ -72,9 +72,11 @@
...
             - id: docs-nav-weak-subjectivity
-              to: /developers/docs/consensus-mechanisms/pos/weak-subjectivity/
+              to: /developers/docs/consensus-mechanisms/pos/weak-subjectivity
+            - id: docs-nav-keys
+              to: /developers/docs/consensus-mechanisms/pos/keys
```
minimalsm, line 79:
```suggestion
              to: /developers/docs/consensus-mechanisms/pos/weak-subjectivity/
            - id: docs-nav-keys
              to: /developers/docs/consensus-mechanisms/pos/keys/
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
...
+Before the switch to proof-of-stake, Ethereum users only had a single elliptic-curve-based private key to access their funds. With the introduction of proof-of-stake, users that wished to be solo stakers also required a **validator key** and a **withdrawal key**.
+
+### The validator key {#validator-key}
+
+The validator signing key consists of two elements:
+
+- Validator **private** key
+- Validator **public** key
+
+The purpose of the validator private key is to actively sign on-chain operations such as block proposals and attestations. Therefore these keys have to be held in a hot wallet.
```
minimalsm, line 27:
```suggestion
The purpose of the validator private key is to sign on-chain operations such as block proposals and attestations. Because of this, these keys must be held in a hot wallet.
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
...
+The **validator public key** is included in the _deposit data_ which allows Ethereum to identify the validator.
+
+### The withdrawal key {#withdrawal-key}
+
+The withdrawal key will be required to move the validator balance after this is enabled in the upcoming Shanghai upgrade. Just like the validator keys, the withdrawal keys also consist of two components:
+
+- Withdrawal **private** key
+- Withdrawal **public** key
+
+Losing this key means losing access to the validator balance. However, the validator can still sign attestations and blocks since these actions require the validator private key, but there is little to no incentive to do so if the keys are lost.
```
minimalsm, line 46:
```suggestion
Losing this key means losing access to the validator balance. However, the validator can still sign attestations and blocks since these actions require the validator's private key, but there is little to no incentive if the keys are lost.
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+This flexibility has the advantage to move validator signing keys very quickly from one device to another, however, if they have gotten lost or stolen, a thief may be able to **act maliciously** in the following ways:
+
+- Get the validator slashed) by:
+  - Being a proposer and signing two different beacon blocks for the same slot
+  - Being an attester and signing an attestation that "surrounds" another one.
+  - Being an attester and signing two different attestations having the same target.
+- Force a voluntary exit, which stops the validator from staking, and grants access to its ETH balance to the withdrawal key owner.
+
+The **validator public key** is included in the _deposit data_ which allows Ethereum to identify the validator.
```
minimalsm, line 37:
```suggestion
The **validator public key** is included in the _deposit data_, which allows Ethereum to identify the validator.
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
...
+### The validator key {#validator-key}
+
+The validator signing key consists of two elements:
+
+- Validator **private** key
+- Validator **public** key
+
+The purpose of the validator private key is to actively sign on-chain operations such as block proposals and attestations. Therefore these keys have to be held in a hot wallet.
+
+This flexibility has the advantage to move validator signing keys very quickly from one device to another, however, if they have gotten lost or stolen, a thief may be able to **act maliciously** in the following ways:
```
minimalsm, line 29:
```suggestion
This flexibility has the advantage of moving validator signing keys very quickly from one device to another. However, if they have gotten lost or stolen, a thief may be able to **act maliciously** in several ways:
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
...
+### The withdrawal key {#withdrawal-key}
+
+The withdrawal key will be required to move the validator balance after this is enabled in the upcoming Shanghai upgrade. Just like the validator keys, the withdrawal keys also consist of two components:
+
+- Withdrawal **private** key
+- Withdrawal **public** key
+
+Losing this key means losing access to the validator balance. However, the validator can still sign attestations and blocks since these actions require the validator private key, but there is little to no incentive to do so if the keys are lost.
+
+Separating out the validator keys from the Ethereum account keys enables multiple validators to be run by a single user.
```
minimalsm, line 48:
```suggestion
Separating the validator keys from the Ethereum account keys enables multiple validators to be run by a single user.
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+![validator key schematic](validator-key-schematic.png)
+
+## Deriving keys from other keys {#deriving-keys-from-keys}
+
+If every 32 ETH staked required a new set of 2 completely independent keys to be created, key management would quickly become unwieldy especially for users running multiple validators. To get around this problem, multiple validator keys can be derived from a single common secret. Storing that single secret allows access to multiple validator keys.
```
minimalsm, line 54:
```suggestion
If every 32 ETH staked required a new set of 2 completely independent keys, key management would quickly become unwieldy, especially for users running multiple validators. Instead, multiple validator keys can be derived from a single common secret and storing that single secret allows access to multiple validator keys.
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+[Mnemonics](https://en.bitcoinwiki.org/wiki/Mnemonic_phrase) and paths are a well-known feature that users often encounter when [they access](https://ethereum.stackexchange.com/questions/19055/what-is-the-difference-between-m-44-60-0-0-and-m-44-60-0) their hardware wallets. The mnemonic is a sequence of words that act as an initial seed for a private key. When combined with additional data, the mnemonic is used to generate a hash known as the 'master key'. This can be thought of as the root of a tree. Branches from this root can then be drived using a heirarchical path, so that child nodes can exist as combinations of their parent node's hash and their index in the tree.
```
minimalsm, line 56:
```suggestion
[Mnemonics](https://en.bitcoinwiki.org/wiki/Mnemonic_phrase) and paths are prominent features that users often encounter when [they access](https://ethereum.stackexchange.com/questions/19055/what-is-the-difference-between-m-44-60-0-0-and-m-44-60-0) their hardware wallets. The mnemonic is a sequence of words that act as an initial seed for a private key. When combined with additional data, the mnemonic generates a hash known as the 'master key'. This can be thought of as the root of a tree. Branches from this root can then be derived using a hierarchical path so that child nodes can exist as combinations of their parent node's hash and their index in the tree.
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+These paths have the following structure which will be familiar to users who have interacted with hardware wallets:
```
minimalsm, line 58:
```suggestion
These paths have the following structure, which will be familiar to users who have interacted with hardware wallets:
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
...
+m/44'/60'/0'/0`
+```
+
+The slashes in this path separate components of the private key as follows:
+
+```
+master_key / purpose / coin_type / account / change / address_index
+```
+
+This logic enables a user to attach as many validators to a single **withdrawal key** as they want, because the root of the tree can be common and differentiation can happen at the branches. The user can **derive any number of keys** from the Mnemonic phrase.
```
minimalsm, line 70:
```suggestion
This logic enables users to attach as many validators to a single **withdrawal key** as possible because the tree root can be common, and differentiation can happen at the branches. The user can **derive any number of keys** from the Mnemonic phrase.
```

minimalsm (changes requested):

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
wackerow: Agree with Josh's suggestions, but also, I'm not sure I understand the way this is worded.

My understanding is each validator has its own set of priv/pub signing keys + priv/pub withdrawal keys. For example, two validators would need two different withdrawal keys to eventually withdraw funds. Both of these can be derived from the seed, but are still unique.

But, with how it's worded here, seems like you're saying a single withdrawal key can be shared... I assume you're referring to a key higher up in the tree, but this threw me off a little. May be nitpicking, but should we maybe refer to this as the "mnemonic phrase" to be consistent with the image?

```suggestion
This logic enables users to attach as many validators as possible to a single **mnemonic phrase** because the tree root can be common, and differentiation can happen at the branches. The user can **derive any number of keys** from the mnemonic phrase.
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
...
+The withdrawal key will be required to move the validator balance after this is enabled in the upcoming Shanghai upgrade. Just like the validator keys, the withdrawal keys also consist of two components:
+
+- Withdrawal **private** key
+- Withdrawal **public** key
+
+Losing this key means losing access to the validator balance. However, the validator can still sign attestations and blocks since these actions require the validator private key, but there is little to no incentive to do so if the keys are lost.
+
+Separating out the validator keys from the Ethereum account keys enables multiple validators to be run by a single user.
+
+![validator key schematic](validator-key-schematic.png)
```
wackerow, line 50: ❤️

wackerow (changes requested):
@jmcook1186 Love this... was also thinking recently how we should consider this topic for an ELI-5 page somewhere. I think having an easy-to-understand place on the site that explains simply how mnemonics work (and the power behind them) would be a beneficial resource. 

Left a few thoughts =)

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
minimalsm: > English question, wouldn't wrapping "however" in commas be the more appropriate way to use it here? Always tried to avoid using it to start sentences where possible.

I changed it to a new sentence because the use of 'however' created a [run-on sentence](https://www.grammarly.com/blog/run-on-sentence-basics/). Using a semi-colon instead would also be correct.

re: your aversion to beginning sentences with 'However', I've not really heard of that before 😛 Maybe a British/American English difference?

> Also, not sure I'd switch to "several" here... It's really just two things, they could get you slashed or they could exit the validator.

I'm good with using 'in a few ways' here. My edit was to remove the implication that the list was a complete list of potentially malicious activities.

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
...
+
+Ethereum's keys are generated using [elliptic-curve cryptography](https://en.wikipedia.org/wiki/Elliptic-curve_cryptography).
+
+However, when Ethereum switched from [proof-of-work](/developers/docs/consensus-mechanisms/pow) to [proof-of-stake](/developers/docs/consensus-mechanisms/pos) a new type of key was added to Ethereum. The original keys still work exactly the same as before - there were no changes to the elliptic-curve-based keys securing accounts. However, users needed a new type of key for participating in proof-of-stake by staking ETH and running validators. This need arose from scalability challenges associated with many messages passing between large numbers of validators that required a cryptographic method that could easily be aggregated to reduce the amount of communication required for the network to come to consensus.
```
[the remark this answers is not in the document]
minimalsm: Agree. I don't love this stylistically either. I'd maybe drop the first use of 'however' in this paragraph altogether, as it's a little monotonous and confusing.

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
+---
+title: Keys in proof-of-stake Ethereum
+description: An explanation of keys used in Ethereum since the transition to proof-of-stake.
```
minimalsm, line 3: I think we can explain (most) things as how they are now (post-merge) vs relative to how they were.

```suggestion
description: An overview of keys used in Ethereum.
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
```diff
@@ -0,0 +1,89 @@
+---
+title: Keys in proof-of-stake Ethereum
```
minimalsm, line 2: Same comment as below. Any reason this can't just be Keys on Ethereum?

wackerow: I think including proof-of-stake makes sense since this page focuses on validator keys. With "Keys on Ethereum" I would assume this to focus more on regular wallet keys. There is a ton of overlap in the topics though, which is why I suggested we potentially expand on this in the future with more content about keys, but this PR is under `consensus-mechanisms/pos/` and I think this makes sense as is.

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
wackerow: Similar to above, I think keeping it dialed into PoS is appropriately here.
```suggestion
description: An explanation of keys used in Ethereum's proof-of-stake consensus mechanism
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
minimalsm: Would it be correct just to call them validator keys?

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
jmcook1186:
```suggestion
[Mnemonics](https://en.bitcoinwiki.org/wiki/Mnemonic_phrase) and paths are prominent features that users often encounter when [they access](https://ethereum.stackexchange.com/questions/19055/what-is-the-difference-between-m-44-60-0-0-and-m-44-60-0) their wallets. The mnemonic is a sequence of words that act as an initial seed for a private key. When combined with additional data, the mnemonic generates a hash known as the 'master key'. This can be thought of as the root of a tree. Branches from this root can then be derived using a hierarchical path so that child nodes can exist as combinations of their parent node's hash and their index in the tree. Read about [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) and [BIP-19](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) standards for mnemonic-based key generation.
```

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
jmcook1186: eh - could do, but withdrawal keys are also included on the page.

wackerow: Yeah this is a little tricky.
I tend to think of the content here slightly differently, with "validator keys" consisting of both "signing keys" and "withdrawal keys". The signing keys are "hot" and stay within or available to your validator client.

Only nuance there, is _all_ keys can "sign", so it kinda adds a different naming conflict when trying to explain this stuff.

@jmcook1186 Am I incorrect in this line of thought? Would it be improper to change the "The validator key" section below to "The signing key", and use that terminology instead?

We can then refer to the whole set as "validator keys".

---

Suggestion:
> `title: Keys in proof-of-stake Ethereum`

`title: "Validator keys"`

> `also required a **validator key** and a **withdrawal key**.`

`also required a validator **signing key** and a **withdrawal key**.`

> `### The validator key `

`### The signing key`

And also updating other instances of "validator" key to "signing" where appropriate, or optionally "validator signing keys" occasionally.

wackerow (approved):
@jmcook1186 Left a reply in the review comment thread above... Overall I think this is fine as is for the first go. 

Let me know what you think about my suggestion... if you agree I'm happy to make the changes, otherwise I'm fine with just pulling it in as is.

minimalsm:
> @jmcook1186 Left a reply in the review comment thread above... Overall I think this is fine as is for the first go.
> 
> Let me know what you think about my suggestion... if you agree I'm happy to make the changes, otherwise I'm fine with just pulling it in as is.

@jmcook1186 ping on this. Happy to merge either way just LMK.

src/content/developers/docs/consensus-mechanisms/pos/keys/index.md
jmcook1186: idk i don't love this bc the withdrawal key is not functionally different from the validator key. "Validator" and "withdrawal" keys separate them by their intended uses - i.e. validator keys are used for validator duties, withdrawal kays are used to withdraw stake - "signing  keys" suggests the alternative is "non-signing keys" which is not true.

jmcook1186: it's also consistent with previous docs, so I say keep validator/withdrawal for now and update later if necessary?
@minimalsm

PR #7222, opened 2022-07-27
Changed 'euros' to 'EUR' [bridges]

hursittarcan:
## Description
* Changed three cases of `euros` to `EUR`. I believe this looks better, because we are also using `USD` instead of `dollars` or `American dollars`. 

## Related Issue
* Not related to an issue.


corwintines (approved):
Makes sense. Thanks for the suggestion @hursittarcan!

PR #7231, opened 2022-07-28
Mention go-ethereum rather than core geth

timbeiko:
geth is by far the most used distribution of the codebase, and sending new users to another codebase, especially given all the changes with the merge can be misleading/dangerous.


jmcook1186 (approved):
Thanks!

PR #7233, opened 2022-07-29
Add optional 'locales' array to learning tools

wackerow:
## Description
- Adds `locales?: Array<Lang>` to `ILearningTool` type for optional array of locales that can be applied to any given learning tool.
- If no array (or empty array), tool will be shown on all locales. If array of locale(s) is present, will only be shown on declared locales.
- Sorted these to the top (open for feedback on if we actually want this sorting or not)
- Extracts reusable component to remove redundancy of ProductCard mapping

## Related Issue
Surfaced from PR #7199 where we did not have a proper way to handle a resource that is only in a non-English language.


src/pages/developers/learning-tools.tsx
```diff
@@ -226,6 +228,40 @@ const LearningToolsPage = ({
     },
   ]
 
+  interface ILearningToolsCardGridProps {
+    category: Array<ILearningTool>
+  }
+
+  const LearningToolsCardGrid: React.FC<ILearningToolsCardGridProps> = ({
```
pettinarip, line 235: Lets create a new file for this component

pettinarip (changes requested):
Nice job. Nice refactor as well 👍🏼

src/pages/developers/learning-tools.tsx
wackerow: @pettinarip Just committed a patch for this. 
Two interfaces were needed, one depending on the other, so I tried just pulling these into the `./types.ts` file to keep it more organized instead of exporting the type interface from the component file. Let me know what you think.

src/components/LearningToolsCardGrid.tsx
```diff
@@ -0,0 +1,50 @@
...
+          ({ locales }) =>
+            !locales?.length || locales.includes(intl.locale as Lang)
+        )
+        .sort(({ locales }) => (locales?.length ? -1 : 0))
+        .map(({ name, description, background, url, alt, image, subjects }) => (
+          <ProductCard
+            key={name}
+            background={background}
+            url={url}
+            alt={translateMessageId(alt, intl)}
```
pettinarip, line 38: Minor comment (nice to have).


I think that this component (as a bunch of others) should be dumb components. Meaning that they should not have much logic inside besides rendering something.

In the case of `alt`, I think we should receive a `string` instead of a `TranslationKey`. The translation (the usage of `translateMessageId`) should be the responsibility of a higher order component (in this case the page).

pettinarip (commented):
Cool job @wackerow.

#7199 was merged. Should we update this branch and add this new filter to that Spanish resource (to test how this works)?

pettinarip (approved):
LGTM

src/components/LearningToolsCardGrid.tsx
```diff
@@ -0,0 +1,50 @@
...
+  category,
+}) => {
+  const intl = useIntl()
+  return (
+    <StyledCardGrid>
+      {category
+        .filter(
+          ({ locales }) =>
+            !locales?.length || locales.includes(intl.locale as Lang)
+        )
```
minimalsm, line 31:
```suggestion
```

corwintines (approved):
Nice 🚀

PR #7245, opened 2022-07-29
Update Balancer Grants link

johngrantuk:
## Description

The page is currently linking to an older version of the Balancer Grants application. This change has updated the link which now points to the Balancer Notion home page which has more background info and further details on how to apply.


minimalsm (approved):
Thanks @johngrantuk :]

I've opened a PR in #7247 to fix the non-English pages that have this link too.

And we've just released our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) also 🥳!

minimalsm:
@all-contributors please add @johngrantuk for docs

PR #7256, opened 2022-07-30
Replace link with client spec

samajammin:
## Description

Subjective opinion but I think first linking to [the Ethereum RPC spec](https://github.com/ethereum/execution-apis) is more relevant to visitors than info on generic JSON RPC specs.

## Related Issue
None 😄


minimalsm (approved):
LGTM

PR #7268, opened 2022-08-01
Use Hashrate consistency

minimalsm:
## Description
Noticed we use hashrate / hash rate inconsistently in some places. Hashrate seems to be more common so defaulting to that for now.

## Related Issue
None


jmcook1186 (approved):
lgtm

PR #7286, opened 2022-08-03
Adding VyperExamples.com do docs

miohtama:
Another Vyper education site


minimalsm (approved):
Nice resource! Ty @miohtama <3

PR #7319, opened 2022-08-05
Update EIP-1559 label to "Fee optimization" for find-wallet

corwintines:
## Description
- Update label and description for `EIP-1559 support` to `Fee optimization`

### DIFF
--- a/src/data/wallets/wallet-filters.ts
+++ b/src/data/wallets/wallet-filters.ts
@@ -131,8 +131,9 @@ const walletFilterData: { [key: string]: WalletFilter } = {
     filterKey: "erc_20_support",
   },
   eip_1559_support: {
-    title: "EIP-1559 support",
-    description: "Lower gas fees with type-2 transactions",
+    title: "Fee optimization",
+    description:
+      "Support type 2 transactions for optimized gas fees, and fee refunds for unused gas",
     filterKey: "eip_1559_support",
   },
   buy_crypto: {
--- a/src/components/FindWallet/WalletPersonasSidebar.tsx
+++ b/src/components/FindWallet/WalletPersonasSidebar.tsx
@@ -288,7 +288,7 @@ const filterLabels = {
     icon: <ERC20Support />,
   },
   eip_1559_support: {
-    label: "EIP-1559 support",
+    label: "Fee optimization",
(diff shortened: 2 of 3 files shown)


src/data/wallets/wallet-filters.ts
```diff
@@ -131,8 +131,9 @@ const walletFilterData: { [key: string]: WalletFilter } = {
     filterKey: "erc_20_support",
   },
   eip_1559_support: {
-    title: "EIP-1559 support",
-    description: "Lower gas fees with type-2 transactions",
+    title: "Fee optimization",
+    description:
+      "Support type 2 transactions for optimized gas fees, and fee refunds for unused gas",
```
samajammin, line 136: Questions (that could be solved in a separate PR):
- Would it still be illuminating/valuable to mention EIP 1559 anywhere?
- Do we define "type 2 transactions" anywhere? If not we probably should... e.g. on the tx page: https://ethereum.org/en/developers/docs/transactions/#typed-transaction-envelope

src/data/wallets/wallet-filters.ts
```diff
@@ -131,8 +131,9 @@ const walletFilterData: { [key: string]: WalletFilter } = {
     filterKey: "erc_20_support",
   },
   eip_1559_support: {
```
samajammin, line 133: If we remove mention of EIP 1559, should we update this `eip_1559_support` key, e.g. to `type_2_tx_support`?

src/data/wallets/wallet-filters.ts
samajammin: Created an issue around the 2nd point: https://github.com/ethereum/ethereum-org-website/issues/7372

corwintines: I think we could add something to the description for this. I was going to do a general pass over the copy for these next sprint so I think it could be done then.

src/data/wallets/wallet-filters.ts
corwintines: Been looking at this code again and I think there should be a refactor on the filters, I think for myself keeping it as 
`eip_1559_support` helps as it is clear what it is as a variable, but open to suggestions

samajammin (approved):

PR #7320, opened 2022-08-05
Update types of wallets section of wallets page

corwintines:
## Description
- Update title `Types of wallet` to `Types of wallet interfaces`
    - Context on this, thought it was worth differentiating that these are the types of ways to interface with an account. Types of wallets to me is more Private key, smart contract, MPC. Not necessarily the device which this is providing information on.
- Add a short description
- Add Brower Extensions as a type of wallet interface.


src/intl/en/page-wallets.json
```diff
@@ -53,8 +53,10 @@
   "page-wallets-try-dapps": "Try some dapps",
   "page-wallets-try-dapps-alt": "An illustration of Ethereum community members working together",
   "page-wallets-try-dapps-desc": "Dapps are applications built on Ethereum. They’re cheaper, fairer and kinder on your data than most traditional applications.",
-  "page-wallets-types": "Types of wallet",
-  "page-wallets-web-browser": "Web wallets that let you interact with your account via a web browser",
+  "page-wallets-types": "Types of wallet interfaces",
```
samajammin, line 56: Hmmm so what is a "wallet" vs. what is a "wallet interface"? This now confuses me a bit.

samajammin: To me these are still "types of wallets". Not convinced we should introduce yet more jargon.

src/intl/en/page-wallets.json
```diff
@@ -53,8 +53,10 @@
...
+  "page-wallets-types-desc": "There are a few ways to interface and interact with your account:",
+  "page-wallets-web-browser": "Browser wallets let you interact with your account directly in the browser",
```
samajammin, line 58:
```suggestion
  "page-wallets-web-browser": "Browser wallets are web applications that let you interact with your account directly in the browser",
```

src/intl/en/page-wallets.json
samajammin, line 56:
```suggestion
  "page-wallets-types": "Types of wallets",
```

samajammin (changes requested):
Suggested a couple changes. Curious what other folks think here.

konopkja:
> ## Description
> * Update title `Types of wallet` to `Types of wallet interfaces`
>   
>   * Context on this, thought it was worth differentiating that these are the types of ways to interface with an account. Types of wallets to me is more Private key, smart contract, MPC. Not necessarily the device which this is providing information on.
> * Add a short description
> * Add Brower Extensions as a type of wallet interface.

this assumes others also have this background knowledge about smart contract, private key etc. I agree with Sam here, that this level of detail maybe unnecessary to differentiate.

corwintines:
Thanks for the input @samajammin @konopkja . Brought in the change requests. I think this should be good now!

samajammin (approved):

PR #7321, opened 2022-08-05
Change buy/withdraw crypto labels to on ramp and off ramp

corwintines:
## Description
- Change `Buy crypto` and `Withdraw crypto` to `On ramp crypto` and `Off ramp crypto` based on user testing feedback

### DIFF
--- a/src/data/wallets/wallet-filters.ts
+++ b/src/data/wallets/wallet-filters.ts
@@ -136,15 +136,15 @@ const walletFilterData: { [key: string]: WalletFilter } = {
     filterKey: "eip_1559_support",
   },
   buy_crypto: {
-    title: "Buy crypto",
+    title: "On ramp to crypto",
     description:
-      "Buy crypto directly in the wallet \n *Note: buying crypto may be region specific",
+      "Buy crypto with fiat directly in the wallet \n *Note: buying crypto may be region specific",
     filterKey: "buy_crypto",
   },
   withdraw_crypto: {
-    title: "Withdraw crypto",
+    title: "Off ramp to fiat",
     description:
-      "Withdraw to fiat directly in the wallet \n *Note: withdrawing crypto may be region specific",
+      "Sell crypto to fiat directly in the wallet \n *Note: withdrawing crypto may be region specific",
     filterKey: "withdraw_crypto",
   },
   multisig: {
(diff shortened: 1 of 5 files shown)


minimalsm (commented):
I agree in principle (it is more accurate to say 'on-ramp' saying 'buy crypto'), but I'm also a little concerned that this phrasing is very colloquial and won't be widely understood. 

@konopkja was this phrasing suggested by someone in user testing? And were they North American?

konopkja:
@minimalsm  no, the user testing only pointed at the problem that "withdraw crypto" was confusing terminology in this context. The user actually suggested using the term "Sell for fiat" or just "Sell crypto". Knowing we have many non native speakers on the website I would also suggest maybe easier terms.

corwintines:
@minimalsm @konopkja So more preferred terms would be:

"Buy crypto" and "Sell for fiat" ?

samajammin:
> "Buy crypto" and "Sell for fiat" 

My 2 gwei: these make sense and are simpler to understand. "Withdraw" & "off ramp" are less clear.

corwintines:
Great, will make these changes!

src/data/wallets/wallet-filters.ts
```diff
@@ -136,15 +136,15 @@ const walletFilterData: { [key: string]: WalletFilter } = {
...
-      "Buy crypto directly in the wallet \n *Note: buying crypto may be region specific",
+      "Buy crypto with fiat directly in the wallet \n *Note: buying crypto may be region specific",
     filterKey: "buy_crypto",
   },
   withdraw_crypto: {
-    title: "Withdraw crypto",
+    title: "Off ramp to fiat",
     description:
-      "Withdraw to fiat directly in the wallet \n *Note: withdrawing crypto may be region specific",
+      "Sell crypto to fiat directly in the wallet \n *Note: withdrawing crypto may be region specific",
```
minimalsm, line 147: @konopkja from your user testing on this page, do you think it's fair to assume people will understand the what fiat currency is? or should we explain it a little here?

konopkja: for many people who have invested in crypto fiat might be a known term. but i mean if there can be even simpler language than that id welcome it. Any ideas?

minimalsm: I'm not so convinced they would. I'll do some testing on friends at the weekend.

Maybe "Sell crypto for fiat (government backed currency) directly in the wallet" or linking to a glossary item in the text?

corwintines:
For now updated to "Buy crypto" and "Sell for fiat". Let me know what you think after the weekend @minimalsm

samajammin (approved):

Issue #7271, opened 2022-08-01
Add detail on MEV after The Merge


emmanuel-awosika:
**Is your feature request related to a problem? Please describe.**

The [current page](https://ethereum.org/en/developers/docs/mev/) on Maximal Extractable Value (MEV) doesn't have a lot of information about MEV after The Merge (except that PoS validators will extract MEV, not miners).

**Describe the solution you'd like**

Seeing as MEV is [predicted](https://hackmd.io/@flashbots/mev-in-eth2) to [factor heavily in validator incomes](https://github.com/flashbots/eth2-research/blob/main/notebooks/mev-in-eth2/eth2-mev-calc.ipynb) (a consequence of reduced ether issuance), I think it'd be great to mention how Ethereum is working on mitigating the impact of MEV on consensus-layer security and decentralization. Specifically, explaining how concepts, like [Proposer-Builder Separation](https://ethresear.ch/t/proposer-block-builder-separation-friendly-fee-market-designs/9725), and current solutions (e.g., [MEV Boost](https://ethresear.ch/t/mev-boost-merge-ready-flashbots-architecture/11177)) will help curb MEV's negative effects after the transition to proof-of-stake. 

**Describe alternatives you've considered**

Putting it as an FAQ on The Merge page (e.g., "How does The Merge affect MEV?") is possible, but I think having an actual explainer would be better. 

**Additional context**

If this sounds good, I can volunteer to start things off by creating an "MEV after The Merge" section on the MEV page. 

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

jmcook1186:
Hi Emmanuel

thanks for this - I personally would support this idea - I simply haven't had time to get to it myself among the other merge updates but acknowledge it is definitely a gap and would appreciate you working on it!

Please go ahead and take ownership of this - I'll look forward to reviewing!

samajammin:
Just noting this open PR: https://github.com/ethereum/ethereum-org-website/pull/7193

Not sure if there's info there worth grabbing/taking into account.

samajammin:
I agree with this issue overall! I think a FAQ on MEV in relation to the merge makes sense. I'd just suggest let's be wary of writing too much content within the FAQ. Let's answer the question but push visitors to the dedicated MEV page (https://ethereum.org/en/developers/docs/mev/) for most of the MEV content.

emmanuel-awosika:
[...]

PR #7333, opened 2022-08-08
Add section on MEV after The Merge

emmanuel-awosika:
Describes the risks that MEV extraction poses to Ethereum under proof-of-stake and evaluates two solutions to curbing negative impacts of MEV in post-Merge Ethereum (Proposer-Builder Separation and MEV Boost).

## Related Issue
#7271


ThreeMuskets12 (commented):
MEV Boost isn't necessarily one of two solutions to "MEV after The Merge" as much as it is one implementation (albeit, likely the only implementation that will be widely adopted) of the Ethereum Builder API. (https://github.com/ethereum/builder-specs) I think this might be confusing to readers who might confuse Flashbot's software implementation of the Builder API (named MEV-boost) with the "MEV after The Merge" solution itself, the Builder API.

emmanuel-awosika:
> MEV Boost isn't necessarily one of two solutions to "MEV after The Merge"

I described MEV Boost as "one of two solutions" because it made sense to include a solution that people could use *now*, as opposed to in-protocol proposer-builder separation which is far off into the future. Yes, it implements Builder API specs, but I doubt that changes anything (also, fwiw, the Builder API specs were based off of the original MEV Boost proposal). 

> I think this might be confusing to readers who might confuse Flashbot's software implementation of the Builder API (named MEV-boost) with the "MEV after The Merge" solution itself, the Builder API.

Is the Builder API the main "MEV after The Merge" solution? Here's what I got from the link you included: *"The Builder API is a temporary solution that requires higher trust assumptions than PBS, but can be fully implemented without modification to the base protocol."*

emmanuel-awosika:
Hey @ThreeMuskets12, I reworked the page to better emphasize the Builder API. Let me know what you think about the additions. And thanks for pointing out the issue in the earlier version.

src/content/developers/docs/mev/index.md
```diff
@@ -116,6 +116,92 @@ While many searchers are still making good money from MEV, as opportunities beco
 
 As DeFi grows and increases in popularity, MEV may soon significantly outweigh the base Ethereum block reward. With that comes a growing possibility of selfish block remining and consensus instability. Some consider this to be an existential threat to Ethereum, and disincentivizing selfish mining is an active area of research in Ethereum protocol theory. One solution currently being explored is [MEV reward smoothing](https://ethresear.ch/t/committee-driven-mev-smoothing/10408).
 
+## MEV after The Merge {#mev-after-the-merge}
```
jmcook1186, line 119: Feel like this wording worked when TM was still looming --now it is in the past maybe we refer to "MEV since The Merge" or "PoS-specific MEV" or similar?

jmcook1186 (commented):
seems good to me overall, but I'm no expert on this so have asked some researchers to take a look to confirm details.

src/content/developers/docs/mev/index.md
emmanuel-awosika: Does "MEV in Ethereum Proof-of-Stake (PoS)" sound good? I like "MEV since the Merge", but it may not be ideal for SEO.

emmanuel-awosika:
> seems good to me overall, but I'm no expert on this so have asked some researchers to take a look to confirm details.

Alright, then. Thanks for reviewing.

minimalsm (approved):
Struggling to get an expert review, but I'm fine with merging this as is.

PR #7367, opened 2022-08-09
Remove WalletCsv.ts and accompanying code

corwintines:
## Description
- Remove WalletCsv.ts
- Remove any code using WalletCsv.ts


samajammin:
While this does remove this /builders/ code that is no longer used, I'm left wondering if we could take a similar approach with the current `walletData` implementation?

I feel like the /builders/ approach is much cleaner than this massive GraphQL query to define every wallet image:
https://github.com/ethereum/ethereum-org-website/blob/dev/src/pages/wallets/find-wallet.tsx#L548

Curious your thoughts @pettinarip. Or is this a downside of using a TS file vs. using a CSV/JSON file via GraphQL to query this wallet data?

pettinarip:
> Curious your thoughts @pettinarip. Or is this a downside of using a TS file vs. using a CSV/JSON file via GraphQL to query this wallet data?

Hmmm I'm not sure if you could load the TS file with `gatsby-source-filesystem` and keep doing the same...I don't think that would work because you will also need a transformer to understand that file. I don't think it is worth it.

From the data perspective, I would keep doing what @corwintines is doing, basically importing the file directly. With the images...we could investigate a bit if we can load all the images inside a specific folder (`/wallets`) at once. Perhaps we could do something like:
```
{
  allFile(filter: {absolutePath: {regex: "/(wallets)/"}}) {
    edges {
      node {
        childImageSharp {
          gatsbyImageData(layout: FULL_WIDTH, placeholder: BLURRED, quality: 100)
        }
      }
    }
  }
}

```

samajammin (approved):

PR #7371, opened 2022-08-09
Find a wallet table: fade icons in and out when dropdown select is changed

corwintines:
## Description
- Adds functionality in for fading out and in the feature icons for a wallet in the WalletTable component.


corwintines:
Open to changing the timing of this animation. Some user feedback from user testing suggested that the compare features didn't seem like they were doing anything because the icons weren't updating (they were, just didn't look like it since they didn't change). This adds a fade-in and out animation to give the appearance of things changing, even if it's just going from true for one feature to true for another.

src/components/FindWallet/WalletTable.tsx
```diff
@@ -1,6 +1,7 @@
 // Libraries
-import React, { useState } from "react"
+import React, { useState, SVGProps } from "react"
 import { getImage, GatsbyImage } from "gatsby-plugin-image"
+import { keyframes } from "@emotion/react"
```
minimalsm, line 4: Should we decide on a convention of how we animate (maybe framer-motion?) on ethereum.org?

We now have at least three ways to animate:

- motion components
- CSS animations
- keyframes

src/components/FindWallet/WalletTable.tsx
```diff
@@ -250,11 +251,34 @@ const FlexInfo = styled.div`
...
+
 const FlexInfoCenter = styled(FlexInfo)`
   justify-content: center;
   cursor: pointer;
   height: 100%;
   display: flex;
+  animation: ${fadeIn} 0.5s;
+
+  &.fade {
+    animation: ${fadeOut} 0.5s;
```
minimalsm, line 280: It's a little too much. Maybe try half of this?

```suggestion
  animation: ${fadeIn} 0.25s;

  &.fade {
    animation: ${fadeOut} 0.25s;
```

minimalsm (commented):
Nice :-)

Agree on changing the timing... it feels a little too dramatic as is 😅.

Also left a comment about us picking an animation convention and but not necessarily a blocker to this PR.

src/components/FindWallet/WalletTable.tsx
corwintines: I was thinking of shortening the time to something like this, will bring this in :)

src/components/FindWallet/WalletTable.tsx
corwintines: Agree. I know that we use framer for a lot of other animations so will refactor to that.

minimalsm (approved):
LGTM. 

@nloureiro or @konopkja to review?

konopkja:
looks good to me

nloureiro:
> LGTM.
> 
> @nloureiro or @konopkja to review?

Just to be sure, are we working on this fading?

https://user-images.githubusercontent.com/1120748/193794839-735f879e-98db-4d5d-b06f-3493d51c591c.mov

If yes, it looks good!

minimalsm:
@nloureiro yes :-)! thanks guys, merged!

PR #7375, opened 2022-08-10
UI lib - new theme colors

### LINKED ISSUE #7374: Define base theme colors
To start the transition to Chakra components, lets define the new base theme colors in order to consume them in the new created components.

pettinarip:
Defines base theme colors.

## Description

1. Implements new Design System palette
2. It also defines the old colors used in the old theme so that we can migrate the components to chakra and keep the same styles as before. In next iterations, we are going to remove them and only use colors from the Design System palette.

Design System colors: https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System

## Related Issue

Fixes #7374 
Epic #6374


src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -0,0 +1,46 @@
...
+    [color]: { _light: lightColor, _dark: darkColor },
+  }
+}, {})
+
+const semanticTokens = {
+  colors: {
+    // define old colors from the old theme as semanticTokens to transition
+    // from emotion components to chakra
+    // TODO: remove these colors as we migrate away from them
+    ...oldColors,
```
pettinarip, line 30: This is to do the migration as seamless as possible. Detailed in the migration guide under the `## Theme colors` header. https://github.com/ethereum/ethereum-org-website/pull/7368
The idea is that we keep using the same color variable names.

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    // Design System colors
+    primary: { _light: "blue.500", _dark: "orange.500" },
+    primaryDark: { _light: "blue.700", _dark: "orange.800" },
+    primaryHover: { _light: "blue.300", _dark: "orange.300" },
+    body: { _light: "grey.700", _dark: "grey.100" },
+    bodyLight: { _light: "grey.500", _dark: "grey.100" },
+    disabled: { _light: "grey.300", _dark: "grey.500" },
+    background: { _light: "white", _dark: "black" },
```
corwintines, line 39:
```suggestion
    background: { _light: "white", _dark: "grey.700" },
```
I think dark is supposed to be this based on the figma

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -0,0 +1,46 @@
...
+  colors: {
+    // define old colors from the old theme as semanticTokens to transition
+    // from emotion components to chakra
+    // TODO: remove these colors as we migrate away from them
+    ...oldColors,
+
+    // Design System colors
+    primary: { _light: "blue.500", _dark: "orange.500" },
+    primaryDark: { _light: "blue.700", _dark: "orange.800" },
+    primaryHover: { _light: "blue.300", _dark: "orange.300" },
```
corwintines, line 35:
```suggestion
    primaryHover: { _light: "blue.300", _dark: "orange.300" },
    primaryLight: { _light: "blue.100", _dark: "orange.100"},
```

Should we also add the primaryLight in this PR as well?

pettinarip: Good catch, yes, definitely.

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
pettinarip: Another great catch, yes, it is a slightly different black. TY

corwintines (approved):
LGTM! 🚀

Issue #7383, opened 2022-08-11
Transactions section shows "Error, please refresh."


Aldizh:
**Describe the bug**

The data for transactions is not being displayed at all in the `TRANSACTIONS TODAY` section

**To Reproduce**

1. Scroll down to `TRANSACTIONS TODAY` section

**Expected behavior**

We should see the latest transaction volume instead of the error

**Screenshots**

<img width="1440" alt="Screen Shot 2022-08-10 at 11 05 48 PM" src="https://user-images.githubusercontent.com/2282120/184063183-98e7bd13-8be0-40fa-bdae-3a68e55d76f9.png">

**Desktop (please complete the following information):**

- OS: MacOS
- Browser: Chrome
- Version: 103.0.5060.134

**Additional context**

The issue is clearly visible in the developer console.
```
TypeError: a.map is not a function
    at StatsBoxGrid.tsx:466:12
```

**Want to contribute?**

[See this PR](https://github.com/ethereum/ethereum-org-website/pull/7384)

Gitwiline:
> **Describe the bug**
> 
> 
> 
> The data for transactions is not being displayed at all in the `TRANSACTIONS TODAY` section
> 
> 
> 
> **To Reproduce**
> 
> Steps to reproduce the behavior:
> 
> 
> 
> 1. Scroll down to `TRANSACTIONS TODAY` section
> 
> 4. See error
> 
> 
> 
> **Expected behavior**
> 
> 
> 
> We should see the latest transaction volume instead of the error
> 
> 
> 
> **Screenshots**
> 
> 
> 
> <img width="1440" alt="Screen Shot 2022-08-10 at 11 05 48 PM" src="https://user-images.githubusercontent.com/2282120/184063183-98e7bd13-8be0-40fa-bdae-3a68e55d76f9.png">
> 
> 
> 
> 
> 
> **Desktop (please complete the following information):**
> 
> 
> 
> - OS: MacOS
> 
> - Browser: Chrome
> 
> - Version: 103.0.5060.134
> 
> 
> 
> **Additional context**
> 
> 
> 
> The issue is clearly visible in the developer console.
> 
> ```
> 
> TypeError: a.map is not a function
> 
>     at StatsBoxGrid.tsx:466:12
> 
> ```
> 
> 
> 
> **Want to contribute?**
> 
> 
> 
> [See this PR](https://github.com/ethereum/ethereum-org-website/pull/7384)

[...]

PR #7384, opened 2022-08-11
Fix loading error for transactions API

Aldizh:
## Description

This change fixes an issue with transactions not being loaded on the main page

## Related Issue

Fixes: https://github.com/ethereum/ethereum-org-website/issues/7383


minimalsm (commented):
Hey @Aldizh, nice catch 👌😀

Looks like the Etherscan API might have changed on us 😕? Not really sure; it was working a few days ago (and had been for months). 

Would it make more sense to make this change in [the lambda function that calls the API?](https://github.com/ethereum/ethereum-org-website/blob/dev/src/lambda/etherscan.ts) We aren't currently using this anywhere else, but we might want to in the future.

minimalsm:
@all-contributors please add @Aldizh for bugs, code

wackerow (approved):
Nice catch @Aldizh! Thanks for the help =)

Issue #7281, opened 2022-08-02
Stablecoin card buttons need margin


hursittarcan:
# Description
The buttons inside the stablecoin cards do not have a margin-bottom. This makes both buttons stick to eachother (it doesn't look good imo). 

The problem is present in both desktop and mobile view. I included screenshots below. 

The first screenshot is the current version. The second screenshot is the result of adding a bottom margin of 5px. Only applied to the left card (with chrome's dev tools).  

![ethereum_desktop_margin - kopie](https://user-images.githubusercontent.com/75273616/182434325-59c1c654-fad5-49f8-8516-4d21cc058714.png)
![image](https://user-images.githubusercontent.com/75273616/182435369-c35d66a3-5782-4616-b1bc-52bcb34d794e.png)

samajammin:
Indeed - thanks for surfacing this bug!

nloureiro:
Great Catch, Thank you.

My take: make the Margin bottom 0.5rem (8px) and I would add the same size for both buttons
![Screen Shot 2022-08-03 09 42 15 AM](https://user-images.githubusercontent.com/1120748/182565359-6c3cc778-b57b-451d-8529-6997da244ca2.png)

vdusart:
Does someone work on this ? If not I can do it :slightly_smiling_face:

wackerow:
@vdusart All yours =). If possible would be nice to use a solution like `gap` here to add a gap between the items despite ordering or flow direction (row/column)

minimalsm:
Closed by #7385

PR #7385, opened 2022-08-11
Fix/stablecoin card

vdusart:
Fixing style of stablecoin button margin

## Description

I created a styled button with mb and mr params.

Hope that was what you meant @wackerow when you talked about a `gap` [here](https://github.com/ethereum/ethereum-org-website/issues/7281#issuecomment-1210773349)

## Related Issue

Fix: #7281

### DIFF
--- a/src/pages/stablecoins.tsx
+++ b/src/pages/stablecoins.tsx
@@ -269,6 +269,15 @@ const PaddedContent = styled(Content)`
   padding-bottom: 3rem;
 `
 
+const StyledButtonLink = styled(ButtonLink)<{
+  mb?: string
+  mr?: string
+}>`
+  margin-bottom: ${(props) => props.mb};
+  margin-right: ${(props) => props.mr};
+  width: 50%;
+`
+
 const tooltipContent = (
   <div>
     <Translation id="data-provided-by" />{" "}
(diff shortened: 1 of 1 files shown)


minimalsm (changes requested):
Hey @vdusart, nice work on this 💪 

Not blocking but I think it'd be a lot cleaner if we applied these styles from the parent. E.g. currently if we want to get rid of the second Button, we'd have to alter the styling on the first. But this is behaviour we'd probably want to use elsewhere too, so maybe pulling this into its own component would make sense at a future date.

My personal preference would be to have these buttons full width, I think it looks better at every screen size. @nloureiro any thoughts? 

DAI is full width. USDC is current width.

Screenshot 2022-08-11 at 12 48 47

Screenshot 2022-08-11 at 12 54 11

Screenshot 2022-08-11 at 12 54 24

vdusart:
@minimalsm I agree with you for the full size, it's nicer !
I'm not sure to understand what you mean with the first part of the comment... :cry:

minimalsm:
> I'm not sure to understand what you mean with the first part of the comment... 😢

I mean that ideally, the styling of these elements would come from the parent element. We're duplicating styles by putting the same margins on different buttons. If we want to make style changes, the code must be edited in both places.

Andy-Waine:
Hi @vdusart - I'm interested in tackling this issue, but I see you've already done a lot of the legwork. Perhaps I could contribute to your PR? or are you currently working on this?

vdusart:
Hey @Andy-Waine I'm not sure but I think it's almost done (maybe some quick changes) 🫤
But if you like design stuff you should definitely join the discord and especially the `ui-library-migration`! There will be for sure work to do there !! 😁

Andy-Waine:
@vdusart Awesome - I'll definitely do that. Thanks!

PR #7386, opened 2022-08-11
Found unnecessary boolean casts - bug risk

LucidSamuel:
## Description

Found unnecessary boolean casts - which could lead to a potential bug risk.


wackerow (approved):
Looks good, thanks @LucidSamuel!

corwintines (approved):

PR #7424, opened 2022-08-12
Add placeholders for token extensions

Pandapip1:
## Description

There are quite a few token standards that build on other token standards. All the standards listed here are final.

## Related Issue

N.A.

### DIFF
--- a/src/content/developers/docs/standards/index.md
+++ b/src/content/developers/docs/standards/index.md
@@ -41,7 +41,11 @@ More detailed information on these different types and categories can be found i
 ### Token standards {#token-standards}
 
 - [ERC-20](/developers/docs/standards/tokens/erc-20/) - A standard interface for fungible (interchangeable) tokens, like voting tokens, staking tokens or virtual currencies.
+  - [ERC-1363](https://eips.ethereum.org/EIPS/eip-1363) - Defines a token interface for ERC-20 tokens that supports executing recipient code after transfer or transferFrom, or spender code after approve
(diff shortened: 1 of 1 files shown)


minimalsm (commented):
Hey @Pandapip1, thanks for the PR :-) 

For a while, I've had the question: _what other standards should we include on ethereum.org?_

Looks like now I've got my answer! 😀 Feel free to suggest others, if you think they are important to highlight.

I like adding these, but I have a slight qualm about the new architecture it creates. If someone were to learn about standards, they should learn progressively from the most simple, used, and impactful to more complex, more utility-type NFTs. (i.e. ERC-20 -> ERC-721 -> ?? -> ERC-1363)... I hope my explanation makes sense here 😅

One thought (just brainstorming, feel free to suggest alternatives) is that we could have a 'Main' standards subheading and an 'Advanced' standards subheading to create this kind of architecture artificially. The heading names are up for debate, but the main idea is to separate the essentials from the niche. What do you think? 🤔

Pandapip1:
I would argue that EIP-20, EIP-721, and EIP-1155 are all roughly on par with each other in terms of understandability. The reason I used this particular structure is that it's clear which are the most basic standards, and which standards build on other standards. I would definitely be down for putting an "Extensions" heading on the relevant pages, though!

corwintines (approved):
Thanks @Pandapip1! I think this is good for now, and we can look at improving the info architecture soon. I don't think that needs to block this right now though.

PR #7428, opened 2022-08-13
Improving text understanding on verifying smart contracts page

OdairTrujillo:
Continuing with the last discussion about the confusing wording

## Description
The second part of the text is better understood with the use of the staring expression "On the other hand".


src/content/developers/docs/smart-contracts/verifying/index.md
```diff
@@ -7,7 +7,7 @@ sidebar: true
...
+It is important to make the distinction between "source code verification" and "[formal verification](/developers/docs/smart-contracts/formal-verification/)". Source code verification, which will be explained in detail below, refers to verifying that the given source code of a smart contract in a high-level language (e.g. Solidity) compiles to the same bytecode to be executed at the contract address. On the other hand, formal verification describes verifying the correctness of a smart contract, meaning the contract behaves as expected. Although context-dependent, contract verification usually refers to source code verification.
```
minimalsm, line 10: I agree that this would be improved by further simplification to make it more readable. However, I'm not the biggest fan of using metaphors in technical documentation as they can be unfamiliar to non-Native English speakers. Maybe it would be better to separate this section into three paragraphs?

cc: @emmanuel-awosika any thoughts?

emmanuel-awosika: > However, I'm not the biggest fan of using metaphors in technical documentation as they can be unfamiliar to non-Native English speakers.

I try to avoid them in writing for this reason. 

> Maybe it would be better to separate this section into three paragraphs? 

This is okay imo.

OdairTrujillo: Ok, let's try it.

emmanuel-awosika: We could use "while", but that would likely involve joining the two sentences (starting that other sentence with "while" reads a bit awkwardly):

> Source code verification, which will be explained in detail below, refers to verifying that the given source code of a smart contract in a high-level language (e.g. Solidity) compiles to the same bytecode to be executed at the contract address, **while** formal verification describes verifying the correctness of a smart contract (meaning the contract behaves as expected). 

Since that sentence would be really long, maybe we can try "however"?:

> Source code verification, which will be explained in detail below, refers to verifying that the given source code of a smart contract in a high-level language (e.g. Solidity) compiles to the same bytecode to be executed at the contract address. **However**, formal verification describes verifying the correctness of a smart contract, meaning the contract behaves as expected. 

I think the second one achieves what you have in mind w/o complicating things.

OdairTrujillo: I agree to use "however"

corwintines, line 10:
```suggestion
It is important to make the distinction between "source code verification" and "[formal verification](/developers/docs/smart-contracts/formal-verification/)". Source code verification, which will be explained in detail below, refers to verifying that the given source code of a smart contract in a high-level language (e.g. Solidity) compiles to the same bytecode to be executed at the contract address. However, formal verification describes verifying the correctness of a smart contract, meaning the contract behaves as expected. Although context-dependent, contract verification usually refers to source code verification.
```

Pulling in change from discussion in comment thread

corwintines (approved):
Thanks, @OdairTrujillo! I made the change to `However` based on how I understood the comments. If you or @emmanuel-awosika disagree feel free to open this back up with a new change.

corwintines:
@all-contributors can you add @OdairTrujillo for docs

PR #7431, opened 2022-08-14
Fix tense for withdraw misconception

corwintines:
## Description
- Fix the tense from `upgraded` to `upgrades` when referring to Shanghai upgrade


minimalsm (approved):
Nice eyes 🕵️‍♂️ that is a pretty bad typo to have 😬

PR #7438, opened 2022-08-15
[UI lib] Button, Link & ButtonLink components

pettinarip:
Migration of Button, Link & ButtonLink components.

## Description

This PR includes the migration of some core components that are a bit related between them. The PR does not only migrate the components to chakra, it also includes **refactors** on each one of them.

**Notes for the reviewer:**
- The main files to review are:
  - in the theme configuration folder (`src/@chakra-ui`)
  - and the mentioned components on the title
- The rest are little adjustments to satisfy the newer components.
- As this PR touches all the pages, lets try to test as many pages as we can.
- The Button styles are the new styles from the design system.

**Things to test or consider:**
- Button & Link styles
  - in light & dark mode
  - states (hover, active, disabled)
- Link functionality
  - Intl support
  - RTL support
  - External, hashed, static links

-----

Test env: https://ethereumorgwebsitedev01-chakrabuttonlinkcomponent.gtsb.io/

Figma: https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System

## Related Issue

Epic #6374


src/components/Button.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react"
+import { Button as ChakraButton, ButtonProps } from "@chakra-ui/react"
+
+import { scrollIntoView } from "../utils/scrollIntoView"
+
+export interface IProps extends ButtonProps {
+  toId?: string
+}
+
+const Button: React.FC<IProps> = ({ toId, children, ...props }) => {
```
pettinarip, line 10: I'm calling this component `Button` bc it is the same Chakra button but with the scrolling feature. This way, you will just use this button component in the codebase.
```ts
import Button from 'components/Button'
```

On the other hand, we could break them out by calling this component `ButtonScroll` and using it only where we need to use the auto-scrolling.
```ts
import ButtonScroll from 'components/ButtonScroll'

// for the rest, we use the native chakra button
import Button from 'chakra/Button'
```


I just wanted to check if you have a preference.

src/components/FeedbackWidget.tsx
```diff
@@ -257,29 +257,38 @@ const FeedbackWidget: React.FC<IProps> = ({ className }) => {
               )}
               <ButtonContainer>
                 {feedbackSubmitted ? (
-                  <ButtonPrimary
+                  <Button
+                    variant="outline"
+                    color="primary"
+                    borderColor="primary"
```
pettinarip, line 263: - [ ] TODO: create new variant for this case (is defined in the design system)

src/components/Link.tsx
```diff
@@ -1,228 +1,126 @@
...
 }) => {
-  // markdown pages pass `href`, not `to`
-  to = to || href
+  const theme = useTheme()
 
-  if (!to) {
-    throw new Error("Either 'to' or 'href' props must be provided")
-  }
+  // TODO: in the next PR we are going to deprecate `to` prop and just use `href`
+  const to = (toProp || href)!
```
pettinarip, line 43: in a following PR we are going to just use `href` for everything

gatsby-config.ts
```diff
@@ -194,7 +194,7 @@ const config: GatsbyConfig = {
     {
       resolve: "@chakra-ui/gatsby-plugin",
       options: {
-        resetCSS: false,
+        resetCSS: true,
```
pettinarip, line 197: We have enabled the reset from Chakra. It does seem to be working fine so far. I haven't seen any conflict yet.

src/components/Link.tsx
samajammin: SGTM

src/components/Link.tsx
```diff
@@ -1,228 +1,143 @@
...
+    eventAction: `Clicked`,
+    eventName: to,
+  }
+
+  const commonProps = {
+    dir,
+    ...restProps,
+  }
+
   // Must use <a> tags for anchor links
```
samajammin, line 80: Remove these comments if we no longer need to use `<a>` tag?

src/components/Link.tsx
```diff
@@ -1,228 +1,143 @@
...
-      <a dir={dir} className={className} href={to} aria-label={ariaLabel}>
+      <Link href={to} {...commonProps}>
         {children}
-      </a>
+      </Link>
     )
   }
 
+  // Download link for internally hosted PDF's (ex: whitepaper)
   // Links to static image assets must use <a> to avoid
```
samajammin, line 92: Same here?

corwintines:
Screen Shot 2022-08-18 at 4 27 30 PM
The hover states don't match with the design system at the moment. Im noticing:
- The boxshadow for darkmode is too light
- The boxshadow of the outline variant should be an outline as well, but right now it is solid.

I'll try and find the suggestions in the PR

corwintines:
Screen Shot 2022-08-18 at 4 45 29 PM

Noticing a regression in the footer social links. Left is the live production version, and right is this PR version.

corwintines:
Screen Shot 2022-08-18 at 4 59 09 PM

Same boxshadow comment as above for this. My other question is should the icon be orange as well? The icon variation in the design system has the icon the same color as the text.

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -31,12 +31,12 @@ const semanticTokens = {
...
     primaryDark: { _light: "blue.700", _dark: "orange.800" },
     primaryHover: { _light: "blue.300", _dark: "orange.300" },
-    primaryLight: { _light: "blue.100", _dark: "orange.100" },
     body: { _light: "grey.700", _dark: "grey.100" },
     bodyLight: { _light: "grey.500", _dark: "grey.100" },
-    disabled: { _light: "grey.300", _dark: "grey.500" },
+    disabled: { _light: "grey.400", _dark: "grey.500" },
     background: { _light: "white", _dark: "grey.700" },
     success: "green.500",
     error: "red.600",
```
minimalsm, line 42: This red has changed to red.500 in design system

minimalsm:
The on-hover box-shadow of the primary button in light mode is not correct.

Current (too light, basically invisible):

Screenshot 2022-08-29 at 11 17 19

Desired (`4px 4px 0 0 #d2d2f9;`):

Screenshot 2022-08-29 at 11 17 37

minimalsm:
Noticing on `active` state for buttons (both light and dark mode), we're using the color `--eth-colors-primaryHover`. Should this be `--eth-colors-primaryActive`?

Screenshot 2022-08-29 at 11 23 53

minimalsm:
Potential regression:

On ethereum.org, the psuedo element arrow used to indicate an external link has `margin-left: 0.125em` and `margin-right: 0.3em;`

Screenshot 2022-08-29 at 11 32 28

In this PR: we appear to have no margins at all.

Screenshot 2022-08-29 at 11 32 21

minimalsm:
Similar issue with the glossary link: 

On ethereum.org: we have an on-hover animation.

Screenshot 2022-08-29 at 11 38 05

In this PR: we have no hover effects.

minimalsm:
Screenshot 2022-08-29 at 11 54 53

Breadcrumbs in dark mode are using light theme colour (no issue on light theme).

minimalsm:
When some buttons are both hovered and active, the resulting UI seems a little odd to me. I've pinged @nloureiro [on Figma to clarify](https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B?node-id=545:16075#255508580).

Screenshot 2022-08-29 at 12 00 05

minimalsm:
I'm seeing some strange behaviour with theming on hard refreshes/direct navigation to pages with the URL. 

As a user with dark mode selected, whenever I see a light mode flash, I'd expect all aspects of the light mode theme to be used (in this case, primary link colour).

[Image from Gyazo](https://gyazo.com/69b15f7ba09983f2f54b339ab063dc4e)

pettinarip:
Thanks for the review @minimalsm 💪🏼 

> The on-hover box-shadow of the primary button in light mode is not correct.
>
> Desired (4px 4px 0 0 #d2d2f9;):

Correct. That color doesn't exist on the color palette. Will ping @nloureiro and see if we can add it.

> Noticing on active state for buttons (both light and dark mode), we're using the color --eth-colors-primaryHover. Should this be --eth-colors-primaryActive?

Yes. @nloureiro we need to update that name in both places. In the design system it should be called `primaryActive` as well to keep consistency. Note: we don't have a `primaryHover` color anymore since we are using `opacity` to achieve that color.

> Not sure if related to this PR: breadcrumbs in dark mode are using light theme colour (no issue on light theme).

Uhmm thats weird. Could you clear the local storage and try this again? I couldn't reproduce that on my side.

> I'm seeing some strange behaviour with theming on hard refreshes/direct navigation to pages with the URL.

This is an issue I haven't fix yet. We need to add the color mode on SSR to avoid this flashing. https://chakra-ui.com/docs/styled-system/color-mode#add-colormodemanager-optional-for-ssr

I will tackle this problem in a different PR.

pettinarip:
@minimalsm about the regressions, good catches. I'll take a look at them.

nloureiro:
> I'm seeing some strange behaviour with theming on hard refreshes/direct navigation to pages with the URL.
> 
> As a user with dark mode selected, whenever I see a light mode flash, I'd expect all aspects of the light mode theme to be used (in this case, primary link colour).
> 
> [Image from Gyazo](https://gyazo.com/69b15f7ba09983f2f54b339ab063dc4e)

It happens on safari and it's not always.


Screen Shot 2022-08-30 09 32 18 AM
Screen Shot 2022-08-30 09 32 06 AM

https://user-images.githubusercontent.com/1120748/187390322-64f087f7-ffb7-4cc0-a0a9-2a430f99dabc.mov

nloureiro:
The focus on the links is too strong. We don't have this on production. 
probably we need to rethink the focus style on link.


https://user-images.githubusercontent.com/1120748/187391759-ea257500-8e1d-44d7-803d-3b0d3ef6bfe4.mov

nloureiro:
> The focus on the links is too strong. We don't have this on production. We probably need to rethink the focus style on the link.
> 
>  Screen.Recording.2022-08-30.at.09.41.21.mov


I got it; it's the box shadow that has to disappear.

https://user-images.githubusercontent.com/1120748/187393138-ce4e7273-9850-4d44-bde1-43b23b0513df.mov

pettinarip:
@nloureiro thanks for the review.

re: focus border, I agree that the border is too thick & doesn't look nice on the links. I don't like too much the idea of hiding it until we rethink it. I'll try to copy the previous version to have at least some focus indicator for a11y purposes.
image
What do you think?

-----

re: the other issues:
- The initial flashing will be covered in a different PR since it is an already known issue.
- The other problems where you have a mixed color mode in different sections, thats weird. I think is the same issue @minimalsm is having with the breadcrumbs colors. I couldn't reproduce it in my local or in the test env....I think it is related to some weird state in the localstorage. Would you be able to clear the localstorage and test it again?

nloureiro:
> @nloureiro thanks for the review.
> 
> re: focus border, I agree that the border is too thick & doesn't look nice on the links. I don't like too much the idea of hiding it until we rethink it. I'll try to copy the previous version to have at least some focus indicator for a11y purposes. image What do you think?
> 
> re: the other issues:
> 
> * The initial flashing will be covered in a different PR since it is an already known issue.
> * The other problems where you have a mixed color mode in different sections, thats weird. I think is the same issue @minimalsm is having with the breadcrumbs colors. I couldn't reproduce it in my local or in the test env....I think it is related to some weird state in the localstorage. Would you be able to clear the localstorage and test it again?


**regarding the :focus**
I'm not sure of your idea here, but that will affect all the <a> tags, right?
How would it work in the side navigation?

we should have :focus for accessibility affordance, what do you think about this for now? then I will revisit this on the Design system and open a PR to fix accordantly.
@minimalsm and @konopkja : any thoughts here?
Screen Shot 2022-08-30 02 25 19 PM

minimalsm:
@pettinarip yeah, I can't reproduce anymore (even without clearing my local storage). I'm not sure what's going on.

====

Re: focus. I don't think the focus style is too much. The weird part is that it is persistening on page transitions. I also Can we stop that? 

I also (personally) wouldn't have onClick focus as the most common convention seems to be for keyboard usage only (but I could be wrong here)

pettinarip:
@nloureiro I like that style. I'll implement that for now.

@minimalsm interesting. Yes, I can confirm it is a v2 improvement. But good for us, we have already a PR opened for that #7617 👍🏼

nloureiro:
> 

@minimalsm : to reproduce the theme bug

1. follow this link in safari > https://ethereumorgwebsitedev01-chakrabuttonlinkcomponent.gtsb.io/en/
2. change the theme

corwintines:
Was doing some testing on my end to see if I could replicate any of the themeing issues, but haven't been able to reproduce them on my side. Tested in Safari, Chrome, Brave, and Firefox.

docs/chakra-migration-guide.md
```diff
@@ -50,10 +50,24 @@ const Text = styled.p`
 ```tsx
 import { Icon } from "@chakra-ui/react"
 import { BsQuestionSquareFill } from "react-icons/bs"
-;<Icon as={BsQuestionSquareFill} />
+
+<Icon as={BsQuestionSquareFill} />
 ```
 
-- [Deprecated]`src/components/SharedStyledComponents` - we are not using this anymore, use Chakra components instead.
+- [Deprecated]`src/components/SharedStyledComponents` - we are not using this anymore, replace everything with Chakra components.
```
corwintines, line 57: Should we add an example of the new way to do this?

pettinarip:
@nloureiro @minimalsm @corwintines 

I’ve updated the PR with the new focus style on the Links in order to let you guys see the results.

The problem is the **inconsistency** with other focus styles we have.

- We are using links not just for text but sometimes to wrap icons or buttons

image

- You can see on the nav dropdowns that the focus style is different. This is because those navs are not links, so they are not inheriting the link styles

image

- Some cards are wrapped with a Link.

image

- Another example of icons

image

----

I would propose to keep the previous focus styles (the ones we currently have in prod) for the links. To keep it simple for the scope of this PR & for consistency.

If you agree then I can modify them to be the same as before. Then, once we have a clear style for the focus in links, we can dedicate a new PR just for that.

In summary, my idea is:
- Button: we apply the new focus styles (the ones that are in the Design System)
- Link: keep the current styles that are in prod

nloureiro:
> @nloureiro @minimalsm @corwintines
> 
> I’ve updated the PR with the new focus style on the Links in order to let you guys see the results.
> 
> The problem is the **inconsistency** with other focus styles we have.
> 
> * We are using links not just for text but sometimes to wrap icons or buttons
> 
> image
> 
> * You can see on the nav dropdowns that the focus style is different. This is because those navs are not links, so they are not inheriting the link styles
> 
> image
> 
> * Some cards are wrapped with a Link.
> 
> image
> 
> * Another example of icons
> 
> image
> 
> I would propose to keep the previous focus styles (the ones we currently have in prod) for the links. To keep it simple for the scope of this PR & for consistency.
> 
> If you agree then I can modify them to be the same as before. Then, once we have a clear style for the focus in links, we can dedicate a new PR just for that.
> 
> In summary, my idea is:
> 
> * Button: we apply the new focus styles (the ones that are in the Design System)
> * Link: keep the current styles that are in prod

errr... this is messy.
I've tried some variations and it's not easy to keep a consistent :focus beahiver with the actual code.
Probably we need to work this on a component level. 

For these reason I agree on keep the production :focus, that is basically no :focus at all. Then we need to discuss how to precess here.

src/components/GlobalStyle.tsx
```diff
@@ -755,14 +755,15 @@ const GlobalStyle: React.FC = () => {
         /* Anchor tag styles */
         /* Selected specifically for mdx rendered side icon link */
         .header-anchor {
-          position: relative;
+          position: relative !important;
```
pettinarip, line 758: Due to some sort of build order, these styles are being overridden by the default styles that come with the plugin https://github.com/gatsbyjs/gatsby/blob/master/packages/gatsby-remark-autolink-headers/src/gatsby-ssr.js#L15 thats why I had to put those `!important`.

I'll see if I can revert these changes in the #7564 PR.

pettinarip:
Cool! I've updated the PR leaving the `:focus-visible` as `outline: auto`, to be handled by the browser.

minimalsm (commented):
This is what I'm getting now for focus state.

Screenshot 2022-08-31 at 16 58 17

This does not seem like very good for accessibility. 

Given that the Chakra update removes the on-click focus and the page transition issue (which I found out is [actually strict WCAG](https://www.w3.org/TR/WCAG20/), but we don't need to follow IMO), would we be better just keeping the focus styling before the revisions here?

corwintines (approved):
Nice job @pettinarip!

minimalsm (approved):

src/components/FeedbackCard.tsx
```diff
@@ -40,26 +40,6 @@ const ButtonContainer = styled.div`
...
-  display: flex;
-  gap: 0.5rem;
-  color: ${({ theme }) => theme.colors.primary};
-  border-color: ${({ theme }) => theme.colors.primary};
-  line-height: 140%;
-  vertical-align: middle;
-  svg {
-    height: 1.5rem;
-    &.flip {
-      transform: scaleY(-1);
```
samajammin, line 53: I believe we still need this transform: https://discord.com/channels/714888181740339261/938069231570923550/1019727618666410034

PR #7456, opened 2022-08-16
Added new resource for scam prevention

emmanuel-awosika:
I added a great Twitter thread by (ex)-MyCrypto CEO Taylor Monahan showing examples of common phishing messages and emails targeted at crypto users.


minimalsm (approved):
Nice find

PR #7467, opened 2022-08-17
Update the Yellow Paper tutorial

qbzzt:
Add a reference to the (incomplete, I think) execution layer specs [here](https://ethereum.github.io/execution-specs).


minimalsm (approved):
Thanks @qbzzt 🥳

Issue #7464, opened 2022-08-16
Add `robots.txt` file


pettinarip:
We have a generated sitemap that is located under `/sitemap/sitemap-index.xml`. Regardless that this location is pretty standard, some search engines might not find it.

**Describe the solution you'd like**

Create and specify the sitemap location in a `robots.txt` file since this file is understood by all the major search engines.

**Additional context**

This seems to be the default plugin for Gatsby: https://www.gatsbyjs.com/plugins/gatsby-plugin-robots-txt/

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

vdusart:
Hey @pettinarip, I hope you're fine !
I can do that :smile:

pettinarip:
> Hey @pettinarip, I hope you're fine ! I can do that smile

I'm doing great! thanks for asking. Hope you are doing good as well.

Ok, I'm assigning it to you. Feel free to ping me if you are stuck or blocked with anything.

pettinarip:
Issue fixed in #7477

PR #7477, opened 2022-08-18
Feat/add robots.txt file

vdusart:
## Description
This PR adds the generation of the `robots.txt` file inside the `gatsby-config.ts` file.

After the build the generated file looks like that:
```
User-agent: *
Allow: /
Sitemap: https://ethereum.org/sitemap/sitemap-index.xml
Host: https://ethereum.org
```

I didn't know what policy do you want to have so I kept the default one (`policy: [{ userAgent: "*", allow: "/" }]`), tell me if you want to change that.

## Related Issue
Add feature requested in #7464

### DIFF
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -104,6 +104,15 @@ const config: GatsbyConfig = {
         },
       },
     },
+    // robots.txt creation
+    {
+      resolve: "gatsby-plugin-robots-txt",
+      options: {
+        host: "https://ethereum.org/",
+        sitemap: "https://ethereum.org/sitemap/sitemap-index.xml",
+        policy: [{ userAgent: "*", allow: "/" }],
+      },
+    },
     // Ability to set custom IDs for headings (for translations)
     // i.e. https://www.markdownguide.org/extended-syntax/#heading-ids
     `gatsby-remark-autolink-headers`,
--- a/gatsby-config.ts
+++ b/gatsby-config.ts
@@ -66,7 +66,7 @@ const config: GatsbyConfig = {
         icon: `src/assets/favicon.png`,
       },
     },
-    // Sitemap generator (ethereum.org/sitemap.xml)
+    // Sitemap generator (ethereum.org/sitemap/sitemap-index.xml)
     {
       resolve: `gatsby-plugin-sitemap`,
       options: {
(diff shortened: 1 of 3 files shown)


gatsby-config.ts
```diff
@@ -104,6 +104,15 @@ const config: GatsbyConfig = {
         },
       },
     },
+    // robots.txt creation
+    {
+      resolve: "gatsby-plugin-robots-txt",
+      options: {
+        host: "https://ethereum.org/",
+        sitemap: "https://ethereum.org/sitemap/sitemap-index.xml",
```
pettinarip, line 112: There is a local variable called `siteUrl` at the beginning of this file. Lets try to use it here just as a good practice. What do you think?

pettinarip (changes requested):
Nice @vdusart, thanks for doing this. Left a comment.

wackerow (changes requested):
@vdusart Nice! Thanks for doing this... I agree with Pablo, if we could reuse the `siteUrl` variable both for the `host` and for the beginning of the `sitemap` string.

Also, one small favor to ask, would you mind updating the comment in the section _above_ your `gatsby-config` changes? The comment says:
```
// Sitemap generator (ethereum.org/sitemap.xml)
```
Would you mind updating this to:
```
// Sitemap generator (ethereum.org/sitemap/sitemap-index.xml)
```
🙏 Thanks again!

gatsby-config.ts
vdusart: Of course I'm ok, I missed that !

vdusart:
> @vdusart Nice! Thanks for doing this... I agree with Pablo, if we could reuse the `siteUrl` variable both for the `host` and for the beginning of the `sitemap` string.
> 
> Also, one small favor to ask, would you mind updating the comment in the section _above_ your `gatsby-config` changes? The comment says:
> 
> ```
> // Sitemap generator (ethereum.org/sitemap.xml)
> ```
> 
> Would you mind updating this to:
> 
> ```
> // Sitemap generator (ethereum.org/sitemap/sitemap-index.xml)
> ```
> 
> pray Thanks again!



> @vdusart Nice! Thanks for doing this... I agree with Pablo, if we could reuse the `siteUrl` variable both for the `host` and for the beginning of the `sitemap` string.
> 
> Also, one small favor to ask, would you mind updating the comment in the section _above_ your `gatsby-config` changes? The comment says:
> 
> ```
> // Sitemap generator (ethereum.org/sitemap.xml)
> ```
> 
> Would you mind updating this to:
> 
> ```
> // Sitemap generator (ethereum.org/sitemap/sitemap-index.xml)
> ```
> 
> pray Thanks again!

Yes no problem, I'm on it :smile:

pettinarip (approved):

wackerow (approved):
Thanks @vdusart! 💪

PR #7485, opened 2022-08-18
post merge updates: deprecated software

jmcook1186:
## Description

Adds section for deprecated "proof-of-work". Not included in main softyware list because no clients were actually deprecated as part of merge, only components. Added separately because it was deprecation of a concept, implemented as a set of software upgrades.

## Related Issue

#7075


minimalsm (approved):
LGTM. There will probably be other things to add here that will get deprecated but we can do this later. We'll probably want to break this page out a little in the future.

PR #7488, opened 2022-08-18
Fix client diversity page link

ameliejyc:
## Description

Previous link was wrong; fixed it to point at right page.

## Related Issue

I didn't make one as it's so tiny.


jmcook1186 (approved):
cool - thanks!

PR #7495, opened 2022-08-19
Add "Leslie the rhino" to /assets page

corwintines:
## Description
- Add Leslie the rhino to the assets page to be downloaded
- Add Tomo Saito artist info


corwintines:
hmm, not sure whats going on with all these extra commits. Will look into this tomorrow

corwintines:
File changes are correct though

src/pages/assets.tsx
```diff
@@ -275,6 +282,24 @@ const AssetsPage = ({ data }: PageProps<Queries.AssetsPageQuery, Context>) => {
...
+            artistUrl="https://tomosaito.com/"
+            image={data.leslieTheRhino}
+          />
+          <AssetDownload
+            title={translateMessageId("page-assets-leslie-the-rhino", intl)}
+            alt={translateMessageId("page-assets-leslie-the-rhino", intl)}
+            artistName="Tomo Saito"
+            artistUrl="https://tomosaito.com/"
+            image={data.leslieTheRhino}
+            shouldHide={true}
```
minimalsm, line 299: Is this just for left aligning the image container? Seems weird but looks like we already implemented it...

minimalsm (approved):

src/pages/assets.tsx
corwintines: Without it the image is full width and huge. I just used another example of this from where the wallet robot is added.

PR #7501, opened 2022-08-19
Update Dapps and Shared styled Components 

Rakeshmonkee:
## Description

Updated the original tornado.cash website to the new link.

## Related Issue

Old website no longer works


Rakeshmonkee:
Before:

Grids are not inline

bfore

height: fit-content;

---

After:

Grids are inline 

after

height: auto;

PR #7520, opened 2022-08-21
Update content in accounts/index.md

othaime-en:
## Description
Added an important distinction between Externally Owned Accounts (EOA) and Contract accounts regarding the presence of private keys.


minimalsm (approved):
Sorry for the delay @othaime-en, thanks for the PR here 💪

Issue #7562, opened 2022-08-24
speedrunethereum.com


nkennethk:
Before submitting this suggestion, be sure to read our [listing policy for dev tools](https://ethereum.org/en/contributing/adding-developer-tools/)

**Product name:**

Speedrun Ethereum

**Product description:**

Toolkit and applied learning challenges for web2 devs to learn etheruem dapp development in 30 minutes!

**How long has it been around?**

Has been around for almost a year, but built by the wonderful Build Guild leader, Austin Griffith, and leverages a bunch of community contributooors who have been a part of the course. 

**Where do you think it should go?**

https://ethereum.org/en/developers/learning-tools/

**GitHub URL**

https://github.com/scaffold-eth/scaffold-eth

**Docs URL**

https://github.com/scaffold-eth/scaffold-eth/blob/master/README.md

**Is the product open source?**

yes, very open source and very much so a public good!

**Additional context**

PR #7563, opened 2022-08-24
Add Speed Run Ethereum to develop tooling

wackerow:
- Adds the Speed Run Ethereum site to the list of developer tools
- Placed under "bootcamp" to start... curious people thoughts on this. Could potentially also lit as a "sandbox"

Fixes https://github.com/ethereum/ethereum-org-website/issues/7562


corwintines:
I think this could be in both bootcamps and sandbox personally

corwintines (approved):
I think this could be in both sandbox and bootcamp, but dont see that as blocking

src/intl/en/page-developers-learning-tools.json
```diff
@@ -35,6 +35,8 @@
   "page-learning-tools-remix-logo-alt": "Remix logo",
   "page-learning-tools-sandbox": "Code sandboxes",
   "page-learning-tools-sandbox-desc": "These sandboxes will give you a space to experiment with writing smart contracts and understanding Ethereum.",
+  "page-learning-tools-speed-run-ethereum-description": "Use Scaffold-ETH to expedite your Solidity knowledge. When you are ready to test your knowledge, speed run Ethereum.",
```
minimalsm, line 38: This doesn't really explain what speed-run Ethereum is.

minimalsm (commented):
Yeah, we definitely need to rework this page's sections (too much in bootcamps) but I'm good with putting it here for now.

Should we tweak the description a little? It isn't immediately obvious what speed run Ethereum is from its description.

src/intl/en/page-developers-learning-tools.json
corwintines, line 38:
```suggestion
  "page-learning-tools-speed-run-ethereum-description": "SpeedRunEthereum is a set of challenges to start developing on Ethereum uring Scaffold-ETH",
```
Maybe something like this?

wackerow: yeah that sounds good, I just ripped some text from their website =)

PR #7564, opened 2022-08-25
Cleanup global styles

pettinarip:
## Description

1. Removed all previous reset css rules in `GlobalStyles` in favor of the reset rules applied by Chakra.
2. Refactor: moved all the _still_ necessary global styles remaining inside `GlobalStyles` into the Chakra theme folder. To have everything in one place.
3. Deleted the `GlobalStyles` component.

## Related Issue

#6374


src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -34,10 +34,10 @@ const semanticTokens = {
...
-    body: { _light: "grey.700", _dark: "grey.100" },
-    bodyLight: { _light: "grey.500", _dark: "grey.100" },
-    disabled: { _light: "grey.400", _dark: "grey.500" },
-    background: { _light: "white", _dark: "grey.700" },
+    body: { _light: "gray.700", _dark: "gray.100" },
+    bodyLight: { _light: "gray.500", _dark: "gray.100" },
+    disabled: { _light: "gray.400", _dark: "gray.500" },
+    background: { _light: "white", _dark: "gray.700" },
     success: "green.500",
     error: "red.600",
```
minimalsm, line 42: This has been changed to red 500 on the design system (red 600 was a typo)

pettinarip: Good catch, ty

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
```diff
@@ -1,7 +1,7 @@
 export type Colors = typeof colors
 
 const colors = {
-  grey: {
+  gray: {
```
corwintines, line 4: I see in the figma this is spelt `grey`. Should that be updated there to reflect this change. I have no strong preference on the spelling we use, just should be consistent.

src/components/Layout.tsx
```diff
@@ -7,7 +7,6 @@ import { IntlProvider } from "react-intl"
 import { LocaleProvider } from "gatsby-theme-i18n"
 
 import { lightTheme, darkTheme } from "../theme"
-import GlobalStyle from "./GlobalStyle"
```
corwintines, line 10: This was the only use of this component, so I think we can delete it :)

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
pettinarip: Good call, yes.

To clarify, Chakra is using `gray`. Calling @nloureiro to update those names in the design system and keep the consistency.

corwintines (approved):
LGTM 🚀

PR #7572, opened 2022-08-25
Fix Geth light client storage

Aracki:
MB instead of GB are mentioned for storage of geth light clients.


minimalsm (commented):
Hey @Aracki, appreciate the PR but this change is incorrect. A light client requires < 400MB of storage (the whole point of it being 'light' is that it doesn't need a lot of storage).

Aracki:
HI @minimalsm, thanks for the feedback. 🙂

I'm pretty sure my storage is 200+ Gb for geth light client started with `--syncmode=light`, but maybe docs are explaining how it would be with "ultra-light" mode or I am missing something. 🤔

PR #7585, opened 2022-08-26
delete dydx as it is leaving ethereum

EvanVanNessEth:


corwintines:
Agree with the PR, but need to do a bit more work to remove dydx. Will get that done this week.

corwintines:
@minimalsm @wackerow if you don't mind taking a second look at this, lots of changes to remove dydx.

wackerow (approved):
Looks good!... Only thing I'm noticing is the search index isn't updated yet, but that should resolve on its own.

corwintines:
Yeah, I think that happens on deploy

PR #7593, opened 2022-08-26
Added amazing talk on desci and for the funding

vincentweisser:


corwintines:
I think this looks good, but I see it added twice. I'm going to remove the first embed link and keep the second.

corwintines (approved):
Thanks for suggesting the video @vincentweisser!

PR #7605, opened 2022-08-28
Fixed error related to Arbitrum Nova description

emmanuel-awosika:
The previous edit described Arbitrum Nova as a cross between a validium and an optimistic rollup, which is incorrect--the Arbitrum Nova chain only posts proofs of data availability to Ethereum, not proofs of validity (as validiums do). I've chosen to describe it as a cross between a sidechain and an optimistic rollup instead, as that feels more accurate.


minimalsm (approved):
Thanks @emmanuel-awosika ❤️

As an aside: I feel like we're soon going to run into a wall with these boxes that we're trying to put scaling technology into... everything is trending towards hybrid systems 🤔.

Issue #7609, opened 2022-08-28
nonce is missing from transaction information in transactions page


nitbit1309:
**Describe the bug**

The transaction contains nonce field to avoid replay attack. This field is not included in the information section of transaction. https://ethereum.org/en/developers/docs/transactions/

**To Reproduce**

1. Go to ['transaction page'](https://ethereum.org/en/developers/docs/transactions/)
2. See 'submitted transaction includes the following information:' section

**Expected behavior**

nonce field need to be present.

minimalsm:
Closed in #7610, thanks @nitbit1309 :]

PR #7610, opened 2022-08-28
nonce field is missing from transaction information in transactions page 

nitbit1309:
adding Nonce as it is also part of the transaction which gets validated along with the transaction

## Description
Nonce is added in the transaction information section.

## Related Issue
[Issue link #7609](https://github.com/ethereum/ethereum-org-website/issues/7609#issue-1353469978)


wackerow (approved):
@nitbit1309 Looks good! Thanks for the concise addition

PR #7613, opened 2022-08-29
Fix typo: Hashamoto -> Hashimoto

vinhloc30796:
## Description

Change "Hashamoto" to "Hashimoto"


minimalsm (approved):
Ugh nice catch, we should probably change this page name also but that has wider implications. 

Thanks for the PR @vinhloc30796 <3

PR #7623, opened 2022-08-29
update response from l2beat api

corwintines:
## Description
- Updates the response code from the change in the l2beat api


minimalsm (approved):
LGTM

PR #7630, opened 2022-08-30
Added info about how the bounty is up to 1 mil

fredriksvantes:
All Merge-related bounties submitted before September 8th receive a 4x bonus multiplier (up to 1 million USD).


fredriksvantes:
Not needed

PR #7632, opened 2022-08-30
Added project examples to DeSci page

vincentweisser:


jmcook1186 (commented):
Don't mind this, but overall would prefer to keep direct links to specific projects in the dedicated list at the end of the page rather than in the table. The reason I say that is that it makes it much more difficult to audit external links and can date the core info on the page if details about projects change.

Opinion @minimalsm ?

minimalsm:
> Don't mind this, but overall would prefer to keep direct links to specific projects in the dedicated list at the end of the page rather than in the table. The reason I say that is that it makes it much more difficult to audit external links and can date the core info on the page if details about projects change.
> 
> Opinion @minimalsm ?

Agree, I'd prefer to separate explaining benefits and showing projects.

minimalsm:
Closing this out as per these comments

> Don't mind this, but overall would prefer to keep direct links to specific projects in the dedicated list at the end of the page rather than in the table. 

If we want to add these to the current list, let's open a separate PR to do so.

Issue #7283, opened 2022-08-02
Tutorial: All you can cache


qbzzt:
**Tutorial title**

All you can cache

**Tutorial description**

When using rollups the cost of a byte in the transaction is a lot more expensive than the cost of a storage slot. Therefore, it makes sense to cache as much information as possible on chain.

In this article the developer learns how to create and use a caching contract in such a way that any parameter value that is likely to be used multiple times will be cached and available for use (after the first time) with a much smaller number of bytes, and how to write off chain code that uses this cache.

**Tag your tutorial**

layer-2, caching, storage

**Publication**

None, yet

**Skill level**

Intermediate

**Hosted on Ethereum.org or hosted elsewhere?**

Ethereum.org

**For tutorials to be hosted on Ethereum.org: Tutorial Content**

Not written yet.

PR #7634, opened 2022-08-31
Tutorial: All you can Cache

qbzzt:
## Description

When using rollups the cost of a byte in the transaction is a lot more expensive than the cost of a storage slot. Therefore, it makes sense to cache as much information as possible on chain.

In this article the developer learns how to create and use a caching contract in such a way that any parameter value that is likely to be used multiple times will be cached and available for use (after the first time) with a much smaller number of bytes, and how to write off chain code that uses this cache.

## Related Issue

Closing: #7283


jmcook1186 (approved):
This great - thank you!

qbzzt:
> This great - thank you!

Thank you. I took a lot of work to write, but I think it will help people make the most of roll ups.

PR #7645, opened 2022-09-01
Adding a youtube button & link

omahs:
Adding a link to the Ethereum Foundation youtube channel


corwintines:
I don't think we should be linking out to the Ethereum Foundation youtube channel like this as this is ethereum.org and not the Ethereum Foundation. Feel free to reopen and push back.

omahs:
@corwintines I proposed to add the same youtube link that's present on Ethereum.org, here:

A

PR #7651, opened 2022-09-01
Unmark status.im as a hardware wallet

Pandapip1:
## Description

Status is a desktop and mobile application that has built-in hot wallet and cold wallet support. It is **NOT** a hardware wallet.

Status.im and [Keycard](https://keycard.tech/) appear to be related, but since they are two very different products, they should likewise be distinct entries.

## Related Issue

N.A.


corwintines (approved):
Makes sense! Thanks for the PR on this @Pandapip1!

PR #7694, opened 2022-09-05
Update security.txt

fredriksvantes:
## Description

Updated security.txt with new expire date and updated URLs.


minimalsm (approved):
Thanks @fredriksvantes

PR #7702, opened 2022-09-05
TS - Fix Gatsby image type

pettinarip:
Implement correct types for Gatsby images.

The native `getImage` and `getSrc` Gatsby functions do not play very well with the types generated by GraphQL Typegen.

The Gatsby team [has improved a bit these functions](https://github.com/gatsbyjs/gatsby/issues/35748) but they still do not support null values.

In this PR, I've created a custom wrapper to deal with the null cases.

---

Most of the changes you see here are just:
- Usage of the new `getImage` and `getSrc` functions
- Usage of the Gatsby image interfaces where we had a type `string` before. These are:
  - `ImageDataLike` => data shape returned by the graphql query
  - `IGatsbyImageData` => data shape returned by the `getImage` function


src/components/ActionCard.tsx
```diff
@@ -89,15 +89,16 @@ const ActionCard: React.FC<IProps> = ({
   isRight,
   isBottom = true,
 }) => {
-  const isImageURL = typeof image === "string" && image.includes("http")
+  const isImageURL = typeof image === "string"
```
pettinarip, line 92: Simplifying this condition. Any sort of string must be passed to an html `img` tag.

src/components/ProductCard.tsx
```diff
@@ -184,18 +184,12 @@ const ProductCard: React.FC<IProps> = ({
...
   const hasRepoData = data && data.repository && !error
 
-  // Check if image is an svg as gatsby-plugin-image doesn't support svg
-  let isSvg = false
-  if (typeof image === "string") {
-    if (image.includes("svg")) {
-      isSvg = true
-    }
-  }
+  const isImgSrc = typeof image === "string"
```
pettinarip, line 187: If `image` is a string just pass it to the `img` tag since `GatsbyImage` can't process it anyway. 

Furthermore, in all the instances where we are using `ProductCard`, we are not sending a string. Only Gatsby image instances.

minimalsm (approved):

Issue #7703, opened 2022-09-06
Spelling mistake in documentation


srv-smn:
**Describe the bug**

There is a spelling mistake in the documentation. Once you reach to this article(https://ethereum.org/en/developers/docs/nodes-and-clients/#snap-sync) in the documentation. In the snap-sync section the spelling of "Merkle trie" is wrong.

**To Reproduce**
Go to https://ethereum.org/en/developers/docs/nodes-and-clients/#snap-sync 

**Expected behavior**

The spelling should be "Merkle tree"

**Screenshots**

![image](https://user-images.githubusercontent.com/47235134/188560856-692bc24c-8e36-4287-b337-b973beb787ac.png)

srv-smn:
I love contributions from the Ethereum community!.

minimalsm:
Hey folks, thanks for highlighting this.

This isn't a mistake, a 'trie' is a type of tree. Maybe not the best naming! ([More on that here](https://stackoverflow.com/questions/4737904/difference-between-tries-and-trees)). 

Pretty common occurrence but going to close this out as no bug here :]

minimalsm:
@all-contributors please add @srv-smn for ideas

PR #7705, opened 2022-09-06
fix typo #7703

parotax:
fix typo #7703


minimalsm:
Hey @parotax, thanks for the PR.

This isn't a mistake, a 'trie' is a type of tree. Maybe not the best naming! ([More on that here](https://stackoverflow.com/questions/4737904/difference-between-tries-and-trees))

Alas, you weren't the first and won't be the last #4087 to make this PR :P

PR #7722, opened 2022-09-06
Fix English link in Translation Banner

pettinarip:
The English-only link in the Translation Banner was being built incorrectly. It was prepending `en` and the default lang that the user had selected, for example, `/en/es/path`.

## Description

Use the new `language` prop to specify the lang of the link.

**Bonus:** I've change the `Link` name from `LinkWrapper` to just `Link` to have a better autocomplete exp.

### DIFF
--- a/src/components/TranslationBanner.tsx
+++ b/src/components/TranslationBanner.tsx
@@ -159,7 +159,8 @@ const TranslationBanner: React.FC<IProps> = ({
               <div>
                 <SecondaryButtonLink
                   variant="outline"
-                  to={`/en${originalPagePath}`}
+                  to={`${originalPagePath}`}
+                  language="en"
                 >
                   <Translation id="translation-banner-button-see-english" />
                 </SecondaryButtonLink>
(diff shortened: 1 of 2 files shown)


src/components/TranslationBanner.tsx
```diff
@@ -159,7 +159,8 @@ const TranslationBanner: React.FC<IProps> = ({
               <div>
                 <SecondaryButtonLink
                   variant="outline"
-                  to={`/en${originalPagePath}`}
+                  to={`${originalPagePath}`}
+                  language="en"
```
pettinarip, line 163: Main change

src/components/TranslationBanner.tsx
```diff
@@ -159,7 +159,8 @@ const TranslationBanner: React.FC<IProps> = ({
               <div>
                 <SecondaryButtonLink
                   variant="outline"
-                  to={`/en${originalPagePath}`}
+                  to={`${originalPagePath}`}
```
pettinarip, line 162:
```suggestion
                  to={originalPagePath}
```

minimalsm (approved):

PR #7723, opened 2022-09-06
Refactor YouTube component to use ChakraUI

minimalsm:
## Description
Migrate YouTube from styled-components to ChakraUI

## Related Issue
#6374

### DIFF
--- a/src/components/YouTube.tsx
+++ b/src/components/YouTube.tsx
@@ -22,7 +22,7 @@ const YouTube: React.FC<IProps> = ({ id, start = "0", title = "YouTube" }) => {
   const baseUrl = "https://www.youtube.com/embed/"
   const src = baseUrl + id + startQuery
   return (
-    <Box as="figure" display="block" marginY="1rem">
+    <Box as="figure" display="block" my={4}>
       <iframe
         width="100%"
         height="315"
--- a/src/components/YouTube.tsx
+++ b/src/components/YouTube.tsx
@@ -1,10 +1,5 @@
 import React from "react"
-import styled from "@emotion/styled"
-
-const Figure = styled.figure`
-  display: block;
-  margin: 1rem 0;
-`
+import { Box } from "@chakra-ui/react"
 
 /**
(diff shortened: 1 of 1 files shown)


src/components/YouTube.tsx
```diff
@@ -27,7 +22,7 @@ const YouTube: React.FC<IProps> = ({ id, start = "0", title = "YouTube" }) => {
   const baseUrl = "https://www.youtube.com/embed/"
   const src = baseUrl + id + startQuery
   return (
-    <Figure>
+    <Box as="figure" display="block" marginY="1rem">
```
pettinarip, line 25:
```suggestion
    <Box as="figure" my={4}>
```
- `figure` is by default `block` displayed, we can ignore it
- there is a short version of `marginY` => `my`
- lets try to use the chakra spacing scale. 1 rem = 4 (more about this [here](https://chakra-ui.com/docs/styled-system/theme#spacing))

src/components/YouTube.tsx
```diff
@@ -27,7 +22,7 @@ const YouTube: React.FC<IProps> = ({ id, start = "0", title = "YouTube" }) => {
...
       <iframe
```
minimalsm, line 26: I don't think we'd want to refactor the iframe to Chakra. LMK if you disagree

src/components/YouTube.tsx
minimalsm, line 25: Keeping display block is necessary for Crowdin formatting reasons.

pettinarip (changes requested):
Good first Chakra PR @minimalsm 🚀 I've pointed out only a few improvements.

src/components/YouTube.tsx
pettinarip: No, thats ok 👍🏼

src/components/YouTube.tsx
pettinarip: Yes, we are going to use the default values. In the Design System too.

pettinarip (approved):
🚀

PR #7724, opened 2022-09-06
Refactor VisuallyHidden component to use ChakraUI

minimalsm:
## Description

Migrate VisuallyHidden from styled component to ChakraUI

## Related Issue
#6374

### DIFF
--- a/src/components/VisuallyHidden.tsx
+++ b/src/components/VisuallyHidden.tsx
@@ -1,18 +1,22 @@
-import React from "react"
-import { VisuallyHidden as VisuallyHiddenSpan } from "@chakra-ui/react"
+import React, { useContext } from "react"
+import { VisuallyHidden as ChakraVisuallyHidden } from "@chakra-ui/react"
+
+import { ZenModeContext } from "../contexts/ZenModeContext"
 
 // Todo: Refactor to from isHidden boolean
 
 export interface IProps {
   children?: React.ReactNode
-  isHidden?: boolean
 }
 
-const VisuallyHidden: React.FC<IProps> = ({ isHidden = false, children }) =>
-  isHidden ? (
-    <VisuallyHiddenSpan>{children}</VisuallyHiddenSpan>
+const VisuallyHidden: React.FC<IProps> = ({ children }) => {
+  const { isZenMode } = useContext(ZenModeContext)
+
+  return isZenMode ? (
+    <ChakraVisuallyHidden>{children}</ChakraVisuallyHidden>
   ) : (
     <>{children}</>
   )
+}
 
 export default VisuallyHidden
--- a/src/components/VisuallyHidden.tsx
+++ b/src/components/VisuallyHidden.tsx
@@ -1,36 +1,7 @@
 import React from "react"
-import styled from "@emotion/styled"
+import { VisuallyHidden as VisuallyHiddenSpan } from "@chakra-ui/react"
 
-// "Accessibility/SEO Friendly CSS Hiding"
-// Source: https://css-tricks.com/snippets/css/accessibilityseo-friendly-css-hiding/
-const Container = styled.span`
-  border: 0;
-  clip: rect(1px, 1px, 1px, 1px);
-  clip-path: inset(50%);
-  height: 1px;
-  margin: -1px;
-  overflow: hidden;
-  padding: 0;
-  position: absolute;
-  width: 1px;
-  word-wrap: normal !important;
-  &:focus {
-    background-color: #eeeeee;
-    clip: auto !important;
-    clip-path: none;
-    color: #444444;
-    display: block;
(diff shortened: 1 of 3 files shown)


src/components/VisuallyHidden.tsx
```diff
@@ -1,43 +1,18 @@
 import React from "react"
-import styled from "@emotion/styled"
+import { VisuallyHidden as VisuallyHiddenSpan } from "@chakra-ui/react"
```
minimalsm, line 2: Alias here as conflicted with component name

src/components/VisuallyHidden.tsx
```diff
@@ -1,43 +1,18 @@
...
-    left: 5px;
-    line-height: normal;
-    padding: 15px 23px 14px;
-    text-decoration: none;
-    top: 5px;
-    width: auto;
-    z-index: 100000;
-  }
-`
+// Todo: Refactor to from isHidden boolean
```
minimalsm, line 4: Obviously out of scope for this refactor, but I disagree with how this component is architected.

If we wrap something as VisuallyHidden, we shouldn't also need to pass a boolean to hide the item we're wrapping.

minimalsm: && we might not even need a component for this at all if all we're doing is importing Chakra's primitive 🤔?

pettinarip: Yes, this one is one of those components that we should remove completely and replace with the Chakra one.

My personal approach to this would be to use this PR to remove it and refactor the usage of this component in the entire codebase in favor of the Chakra one.

I'm fine as well if you want to take this middle ground approach. We could do the refactor later.

minimalsm: Can do, the only place we use it is in Layout.tsx.

Are we fine with just conditionally rendering, or is there a smarter way to go about it 🤔 

```js
{isZenMode ? (
  <VisuallyHidden>
    <Nav path={path} />
    {shouldShowSideNav && <SideNavMobile path={path} />}
  </VisuallyHidden>
) : (
  <React.Fragment>
    <Nav path={path} />
    {shouldShowSideNav && <SideNavMobile path={path} />}
  </React.Fragment>
)}
```

pettinarip: Hmm seeing the code, I think it would be worth abstracting the zen mode state in its own component.

```tsx
<ZenMode>
  <Nav />
  ...
</ZenMode>
```

Inside `ZenMode`, you will have all of the logic

```tsx
function ZenMode({children}) {
  const [isZenMode, setIsZenMode] = useState(false)

  ...

  if(isZenMode) {
    return (
      <VisuallyHidden>{children}</VisuallyHidden>
    )
  }

  return chilren
}
```

minimalsm: Good idea. I will get to this when I have some free evening time this week.

src/components/VisuallyHidden.tsx
pettinarip: One thing on this, can you call it `ChakraVisuallyHidden`? since we are calling those aliases with the `Chakra` prefix in other files

pettinarip (changes requested):

corwintines:
@minimalsm to update this

src/components/VisuallyHidden.tsx
corwintines: I think this is easier said than done given the `handleZenModeChange` in `Layout.tsx` is passed to a provider. Will do some thinking here today.

corwintines: I think we can just import this context though instead of this. Still ways to clean it up.

pettinarip (approved):
lgtm!

src/components/VisuallyHidden.tsx
```diff
@@ -1,43 +1,22 @@
...
 
 export interface IProps {
   children?: React.ReactNode
-  isHidden?: boolean
 }
 
-const VisuallyHidden: React.FC<IProps> = ({ isHidden = false, children }) =>
-  isHidden ? <Container>{children}</Container> : <>{children}</>
+const VisuallyHidden: React.FC<IProps> = ({ children }) => {
```
pettinarip, line 12: I think we need to change the name of this component. This is not a generic component anymore. This is strictly related to zen mode. Would suggest calling this component `ZenMode`.

pettinarip (changes requested):

src/components/VisuallyHidden.tsx
wackerow: Agree... just pushed a commit for this. I think this should be good to go now.

wackerow (approved):

pettinarip (approved):
🚀

PR #7725, opened 2022-09-06
Refactor UpgradeTableOfContents component to use ChakraUI

minimalsm:
## Description

Migrate UpgradeTableOfContents from styled component to ChakraUI

## Related Issue
#6374

### DIFF
--- a/src/components/UpgradeTableOfContents.tsx
+++ b/src/components/UpgradeTableOfContents.tsx
@@ -1,5 +1,6 @@
 import React from "react"
-import { Box, Link, ListItem, UnorderedList } from "@chakra-ui/react"
+import { Box, ListItem, UnorderedList } from "@chakra-ui/react"
+import Link from "../components/Link"
 
 import type { Item as ItemTableOfContents } from "./TableOfContents"
 
@@ -86,9 +87,7 @@ const ItemsList: React.FC<IPropsItemsList> = ({
     <>
       {items.map((item, index) => (
         <ListItem margin={0} key={index}>
-          <div>
-            <TableOfContentsLink depth={depth} item={item} />
-          </div>
+          <TableOfContentsLink depth={depth} item={item} />
         </ListItem>
       ))}
     </>
@@ -98,14 +97,9 @@ const ItemsList: React.FC<IPropsItemsList> = ({
 export interface IProps {
   items: Array<Item>
   maxDepth?: number
-  className?: string
 }
 
-const UpgradeTableOfContents: React.FC<IProps> = ({
-  items,
-  maxDepth = 1,
-  className,
-}) => {
+const UpgradeTableOfContents: React.FC<IProps> = ({ items, maxDepth = 1 }) => {
   if (!items) {
     return null
   }
@@ -117,7 +111,6 @@ const UpgradeTableOfContents: React.FC<IProps> = ({
   return (
     <Box
       as="aside"
-      className={className}
       p={0}
       mb={8}
       textAlign="end"
@@ -129,7 +122,7 @@ const UpgradeTableOfContents: React.FC<IProps> = ({
         py={0}
         ps={4}
         pe={1}
-        fontSize="1.25rem"
+        fontSize="xl"
         fontWeight="normal"
         lineHeight="1.6"
         styleType="none"
--- a/src/components/UpgradeTableOfContents.tsx
+++ b/src/components/UpgradeTableOfContents.tsx
@@ -1,46 +1,9 @@
 import React from "react"
-import { motion } from "framer-motion"
-import { Link } from "gatsby"
-import styled from "@emotion/styled"
+import { Box, Link, ListItem, UnorderedList } from "@chakra-ui/react"
 
 import type { Item as ItemTableOfContents } from "./TableOfContents"
 
 const customIdRegEx = /^.+(\s*\{#([A-Za-z0-9\-_]+?)\}\s*)$/
-
-const Aside = styled.aside`
-  padding: 0rem;
-  text-align: right;
-  margin-bottom: 2rem;
-  overflow-y: auto;
-`
-
-const OuterList = styled(motion.ul)`
-  list-style-type: none;
-  list-style-image: none;
-  padding: 0;
-  margin: 0;
-  font-size: 1.25rem;
-  text-align: right;
-  line-height: 1.6;
-  font-weight: 400;
-  padding-right: 0.25rem;
-  padding-left: 1rem;
-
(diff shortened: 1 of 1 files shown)


src/components/UpgradeTableOfContents.tsx
```diff
@@ -1,47 +1,10 @@
...
 const customIdRegEx = /^.+(\s*\{#([A-Za-z0-9\-_]+?)\}\s*)$/
 
-const Aside = styled.aside`
-  padding: 0rem;
-  text-align: right;
-  margin-bottom: 2rem;
-  overflow-y: auto;
-`
-
-const OuterList = styled(motion.ul)`
```
minimalsm, line 17: Motion ul wasn't doing anything? I've removed it.

src/components/UpgradeTableOfContents.tsx
```diff
@@ -1,47 +1,10 @@
...
-  list-style-type: none;
-  list-style-image: none;
```
minimalsm, line 19: I have no idea what this was supposed to do, but it also appeared to be doing nothing when I tested removing it—so I just left it out as I wasn't sure the best way to approach this in Chakra.

src/components/UpgradeTableOfContents.tsx
```diff
@@ -145,11 +115,28 @@ const UpgradeTableOfContents: React.FC<IProps> = ({
...
+      mb={8}
+      textAlign="end"
+      overflowY="auto"
+      display={{ base: "none", l: "block" }}
+    >
+      <UnorderedList
+        m={0}
+        py={0}
+        ps={4}
+        pe={1}
```
minimalsm, line 131: Is always using the shorthand our preference?

src/components/UpgradeTableOfContents.tsx
```diff
@@ -145,11 +115,28 @@ const UpgradeTableOfContents: React.FC<IProps> = ({
...
+        fontSize="1.25rem"
```
minimalsm, line 132: @pettinarip I couldn't find anything in the design system for 20px fonts.

src/components/UpgradeTableOfContents.tsx
```diff
@@ -145,11 +115,28 @@ const UpgradeTableOfContents: React.FC<IProps> = ({
...
+        fontWeight="normal"
+        lineHeight="1.6"
```
minimalsm, line 134: No variables for this, Chakra defaults don't cover 1.6.

src/components/UpgradeTableOfContents.tsx
```diff
@@ -1,47 +1,10 @@
 import React from "react"
-import { motion } from "framer-motion"
-import { Link } from "gatsby"
-import styled from "@emotion/styled"
+import { Box, Link, ListItem, UnorderedList } from "@chakra-ui/react"
```
minimalsm, line 2: Am I correct in thinking that we import `Link` directly from Chakra as our `Link` component is already extending it?

src/components/UpgradeTableOfContents.tsx
```diff
@@ -1,47 +1,10 @@
...
-
-const ListItem = styled.li`
-  margin: 0;
-`
-
-const StyledTableOfContentsLink = styled(Link)`
-  position: relative;
-  display: inline-block;
-  color: ${(props) => props.theme.colors.text300};
-  margin-bottom: 0.5rem !important;
```
minimalsm, line 42: !important wasn't needed here, but I'm not clear how we'd implement it if it was.

src/components/UpgradeTableOfContents.tsx
pettinarip: 🤔 yeah, I don't see that we are doing any animation with it. Fine with removing it.

src/components/UpgradeTableOfContents.tsx
pettinarip: I would say so but I don't have a strong preference.

As a side note, in this particular case, we are going to style the native Chakra component (on the theme definition), as we did with the Button and the Link.

src/components/UpgradeTableOfContents.tsx
pettinarip: In this case, we should use `xl`.

`xl === 1.25rem` Here you can see the different values for `fontSize` https://chakra-ui.com/docs/styled-system/theme#typography

src/components/UpgradeTableOfContents.tsx
pettinarip: In this case, we should just use the raw value: `1.6rem`.

src/components/UpgradeTableOfContents.tsx
```diff
@@ -145,11 +115,28 @@ const UpgradeTableOfContents: React.FC<IProps> = ({
   }
 
   return (
-    <Aside className={className}>
-      <OuterList>
+    <Box
+      as="aside"
+      className={className}
```
pettinarip, line 120: I don't think we are passing any `className` to this component in any of the uses. I would recommend removing it since we should handle all the styles using props.

src/components/UpgradeTableOfContents.tsx
```diff
@@ -115,7 +85,7 @@ const ItemsList: React.FC<IPropsItemsList> = ({
   return (
     <>
       {items.map((item, index) => (
-        <ListItem key={index}>
+        <ListItem margin={0} key={index}>
           <div>
```
pettinarip, line 89: Could we remove those `div`s?

pettinarip (changes requested):
Good job! I've replied to your comments with some recommendations. LMK what do you think 👍🏼

src/components/UpgradeTableOfContents.tsx
pettinarip: Another thought I had, could you try to remove it and test if it works? because we are defining in [the global styles that we are using `1.6rem` as the default](https://github.com/ethereum/ethereum-org-website/blob/dev/src/%40chakra-ui/gatsby-plugin/styles.ts#L20).

src/components/UpgradeTableOfContents.tsx
pettinarip: We shouldn't need to use `!important` with Chakra anymore since now, all the styles are props, and the way you override styles is by overriding those props.

Example:
```tsx
// our custom Link component
function Link(props) {
  return <ChakraLink p={4} {...props} />
}

// overriding styles
<Link p={0} />
```

corwintines (approved):

Issue #7690, opened 2022-09-05
Read address aloud functionality on the staking deposit address reads address incorrectly


minimalsm:
**Describe the bug**

The 'read aloud' functinonality on the staking deposit contract page does not work correctly. When following a number the letter 'C' is read as '_degrees Celsius_' and the letter '**F**' as '_degrees Fahrenheit_'.

**To Reproduce**

1. Go to 'https://ethereum.org/en/staking/deposit-contract/'
2. Click on 'Read address aloud'
3. See error

**Expected behavior**

It should read the characters exactly as they appear (characters should not be relational to one another)

**Smartphone (please complete the following information):**

- Device: iMac
- OS: MacOS 12.5.1
- Browser: Brave
- Version: v1.43.89

**Want to contribute?**

We love contributions from the Ethereum community! Please comment on an issue if you're interested in helping out with a PR.

hursittarcan:
I could not reproduce this issue. It works ok on my device. (Win 10, Google Chrome v.104.0)

sadiejay:
I can recreate the issue on my device:
 
    OS: MacOS 12.5.1
    Browser: Firefox
    Version: v104.0.1

sadiejay:
May I be assigned this issue? I'm not sure how to tackle this but would like to try!

minimalsm:
Closed by #7736

sadiejay:
Oops did I miss something? I thought a comment was necessary before submitting a PR. Has that changed @minimalsm ?

PR #7736, opened 2022-09-07
Fix read address aloud functionality [#7690]

nagabhiru:
Fixed the read address aloud issue in #7690

## Description

We were using the Web Speech API and setting the [SpeechSynthesisUtterance.text](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesisUtterance/text) to DEPOSIT_CONTRACT_ADDRESS.split("").join(" ")

When following a number the letter 'C' is read as 'degrees Celsius' and the letter 'F' as 'degrees Fahrenheit'.

I have added commas between the characters so that each character is read separately

### DIFF
--- a/src/pages/staking/deposit-contract.tsx
+++ b/src/pages/staking/deposit-contract.tsx
@@ -229,9 +229,9 @@ const DepositContractPage = ({
     // Create textToSpeechRequest
     let speech = new SpeechSynthesisUtterance()
     speech.lang = "en-US"
-    speech.text = DEPOSIT_CONTRACT_ADDRESS.split("").join(" ")
+    speech.text = DEPOSIT_CONTRACT_ADDRESS.split("").join(",")
     speech.volume = 1
-    speech.rate = 0.5
+    speech.rate = 1
     speech.pitch = 1
     // Add event listeners
     // Explicity set state in listener callback


src/pages/staking/deposit-contract.tsx
```diff
@@ -229,7 +229,7 @@ const DepositContractPage = ({
     // Create textToSpeechRequest
     let speech = new SpeechSynthesisUtterance()
     speech.lang = "en-US"
-    speech.text = DEPOSIT_CONTRACT_ADDRESS.split("").join(" ")
+    speech.text = DEPOSIT_CONTRACT_ADDRESS.split("").join(",")
```
minimalsm, line 232: Nice. One note: this rapidly decreases the speed at which this is read, maybe we increase `speech.rate` to compensate?

minimalsm (commented):
Thanks for the PR @nagabhiru, this solves the problem 🥳 Left one small comment about the reading speed that we could probably iron out.

src/pages/staking/deposit-contract.tsx
```diff
@@ -229,9 +229,9 @@ const DepositContractPage = ({
...
     speech.volume = 1
-    speech.rate = 0.5
+    speech.rate = 1.5
```
minimalsm: This is better, thanks!

minimalsm (approved):
Thanks @nagabhiru, and thanks for explaining the solution. TIL :-)!

PR #7741, opened 2022-09-08
Fix lighthouse beacon node command

coincashew:
Reference: https://lighthouse-book.sigmaprime.io/merge-migration.html#connecting-to-an-execution-engine


minimalsm (approved):
Nice catch. Thanks @coincashew 💪

PR #7755, opened 2022-09-08
Feat/convert card chakra

vdusart:
Convert the Card component to Chakra

## Related Issue

Epiec: #6374

### DIFF
--- a/src/components/Card.tsx
+++ b/src/components/Card.tsx
@@ -1,9 +1,6 @@
 import React, { ReactNode } from "react"
-import styled from "@emotion/styled"
 import Emoji from "./Emoji"
 import { Flex, Text } from "@chakra-ui/react"
-
-const TopContent = styled.div``
 
 export interface IProps {
   children?: React.ReactNode
@@ -30,7 +27,7 @@ const Card: React.FC<IProps> = ({
     p={6}
     className={className}
   >
-    <TopContent>
+    <div>
       {emoji && <Emoji fontSize="5xl" text={emoji} mb={4} />}
       {title && (
         <Text as="h3" mt={0}>
@@ -42,7 +39,7 @@ const Card: React.FC<IProps> = ({
           {description}
         </Text>
       )}
-    </TopContent>
+    </div>
     {children}
   </Flex>
 )
--- a/src/components/Card.tsx
+++ b/src/components/Card.tsx
@@ -28,7 +28,9 @@ const Card: React.FC<IProps> = ({
     className={className}
   >
     <div>
-      {emoji && <Emoji fontSize="5xl" text={emoji} mb={4} />}
+      {emoji && (
+        <Emoji fontSize="5xl" text={emoji} mb={4} display="inline-block" />
+      )}
       {title && (
         <Text as="h3" mt={0}>
           {title}
--- a/src/components/Card.tsx
+++ b/src/components/Card.tsx
@@ -3,11 +3,6 @@ import styled from "@emotion/styled"
 import Emoji from "./Emoji"
 import { Flex, Text } from "@chakra-ui/react"
 
-const Description = styled.p`
-  opacity: 0.8;
-  margin: 0;
-`
-
 const TopContent = styled.div``
 
 export interface IProps {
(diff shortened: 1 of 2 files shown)


src/components/Card.tsx
```diff
@@ -1,30 +1,6 @@
...
-  margin: 0;
-`
-
-const StyledEmoji = styled(Emoji)`
-  margin-bottom: 1rem;
-`
-
-const TopContent = styled.div``
+import Emoji from "./Emoji"
+import { Flex, Text } from "@chakra-ui/react"
```
pettinarip, line 3: import order
```suggestion
import { Flex, Text } from "@chakra-ui/react"
import Emoji from "./Emoji"
```

src/components/Card.tsx
```diff
@@ -41,14 +17,33 @@ const Card: React.FC<IProps> = ({
...
-    </TopContent>
+  <Flex
+    direction="column"
+    bg="ednBackground"
+    borderRadius="2px"
+    border="1px"
+    borderStyle="solid"
+    borderColor="lightBorder"
+    p={6}
+    className={className}
```
pettinarip, line 28: Same as in the other PR, I think it is a good moment to deprecated the usage of `className` in our components.

src/components/Card.tsx
```diff
@@ -41,14 +17,33 @@ const Card: React.FC<IProps> = ({
...
+  >
+    <div>
+      {emoji && (
+        <Emoji fontSize="5xl" text={emoji} mb={4} display="inline-block" />
+      )}
+      {title && (
+        <Text as="h3" mt={0}>
```
pettinarip, line 35:
```suggestion
        <Heading as="h3" mt={0}>
```

src/components/Card.tsx
```diff
@@ -41,14 +17,33 @@ const Card: React.FC<IProps> = ({
...
-  <StyledCard className={className}>
-    <TopContent>
-      {emoji && <StyledEmoji size={3} text={emoji} />}
-      {title && <Title>{title}</Title>}
-      {description && <Description>{description}</Description>}
-    </TopContent>
+  <Flex
+    direction="column"
+    bg="ednBackground"
+    borderRadius="2px"
```
pettinarip, line 23:
```suggestion
    borderRadius="sm"
```

src/components/Card.tsx
```diff
@@ -41,14 +17,33 @@ const Card: React.FC<IProps> = ({
...
+    border="1px"
+    borderStyle="solid"
+    borderColor="lightBorder"
+    p={6}
+    className={className}
+  >
+    <div>
+      {emoji && (
+        <Emoji fontSize="5xl" text={emoji} mb={4} display="inline-block" />
```
pettinarip, line 32: Emoji is already displayed as `inline-block`
```suggestion
        <Emoji fontSize="5xl" text={emoji} mb={4} />
```

pettinarip (changes requested):
Thanks @vdusart, its looking good! Left a few comments.

src/components/Card.tsx
vdusart: I'm not sure about this because if I don't force it, it's not there.
*I did my tests on this page https://ethereum.org/en/community/*

This one is when I force the `display: inline-block`:

image

And this one when I remove the `display` option:

Screenshot from 2022-09-09 20-39-24

pettinarip: Interesting, I think I know what is going on.

In the `Emoji` component I've defined the display as `d="inline-block"` but the short `d` prop version is not supported any more in v2, so it is not taking effect.

Could you try changing that to `display="inline-block"` in the `Emoji` component? and test if that works?

vdusart: It's working :+1:

pettinarip: Awesome, thanks!

src/components/Card.tsx
pettinarip: As commented in a different PR, we are going to keep these `className`s until we remove the remaining styled components who use this prop.

pettinarip (approved):
🚀

PR #7768, opened 2022-09-09
Fix typo: remove duplicate is

z3zzz:
## Description

There was a typo that "is" used twice, therefore removed one "is".

## Related Issue

There is no related issue.


z3zzz:
found one more typo, will make new pull request after including that typo also.

PR #7721, opened 2022-09-06
UI migrate `CardList`

pettinarip:
## Description

Migrate `CardList` component to Chakra.

It also includes a refactor on the URL util functions that used to live inside the `Link` component. I've moved all of those related functions to its own file (`src/utils/url.ts`).

## Related Issue

#6374

PR #7780, opened 2022-09-10
Refactor DocLink component to use ChakraUI

Master7130:
## Description

Converted DocLink.tsx from styled-components to ChakraUI

## Related Issue

#6374


src/components/DocLink.tsx
```diff
@@ -1,94 +1,66 @@
...
-      <TextCell>
-        <Title>{children}</Title>
-      </TextCell>
-      <Arrow
-        isExternal={isExternal}
-        color={theme.colors.text}
-        name="arrowRight"
-      />
-    </Container>
+    <Link to={to} isExternal={isExternal} textDecoration="none" hideArrow>
```
pettinarip, line 18: Lets try to use https://chakra-ui.com/docs/components/link-overlay/usage. I have used in this other PR #7721

src/components/DocLink.tsx
```diff
@@ -1,94 +1,66 @@
...
+      <Container
+        position="relative"
+        zIndex={1}
+        textDecoration="none"
+        display="flex"
```
pettinarip, line 23: We can use the `Flex` component to simplify this. https://chakra-ui.com/docs/components/flex

src/components/DocLink.tsx
```diff
@@ -1,94 +1,66 @@
...
+        borderStyle="solid"
+        borderColor="border"
+        _hover={{
+          textDecoration: "none",
+          background: "tableBackgroundHover",
+          borderRadius: "base",
+        }}
+        role="group"
+      >
+        <Box display="flex" alignItems="center">
```
pettinarip, line 41:
```suggestion
        <Flex align="center">
```

src/components/DocLink.tsx
```diff
@@ -1,94 +1,66 @@
...
-    border-radius: 4px;
-
-    svg {
-      fill: ${(props) => props.theme.colors.primary};
-      transition: transform 0.1s;
-      transform: scale(1.2)
-        rotate(${({ isExternal }) => (isExternal ? "-45deg" : "0")});
-    }
-  }
-  @media (max-width: ${({ theme }) => theme.breakpoints.m}) {
```
pettinarip, line 41: We are not declaring this media query. We could use https://chakra-ui.com/docs/hooks/use-media-query for this.

pettinarip (changes requested):
Thanks for the PR @Master7130 🎉 

This is looking really good. I've left a few comments to use a bit better the Chakra components. LMK what you think.

Master7130:
> Thanks for the PR @Master7130 🎉
> 
> This is looking really good. I've left a few comments to use a bit better the Chakra components. LMK what you think.

Thanks for the suggestions! I'll start making the changes in the areas you pointed out.

src/components/DocLink.tsx
Master7130: Wouldn't it be better to use the ChakraUI <Link> component instead of <LinkBox> and <LinkOverlay> since there aren't going to be multiple links inside the component?

pettinarip: Not sure if I follow you on this one. We don't have multiple links in this component.

What `LinkOverlay` does is avoid wrapping the component with a `<a>` tag (as we were doing) which is not semantically correct. It would be an improvement on what we had.

Master7130: Ohhhhh ok. That makes sense. Sorry if my questions seem silly. It's my first time using ChakraUI actually so thanks for being patient with me and giving me suggestions.

pettinarip: That is totally fine! I didn't explain myself correctly. Glad that you brought that concern up 💯

src/components/DocLink.tsx
Master7130: Hey @pettinarip, can you take a look at my recent modifications to the pull request. Specifically the modifications I made in my latest change for the useMediaQuery. I'm not sure whether I did it the right way.

pettinarip (changes requested):

src/components/DocLink.tsx
```diff
@@ -1,94 +1,81 @@
...
-        isExternal={isExternal}
-        color={theme.colors.text}
-        name="arrowRight"
-      />
-    </Container>
+    <LinkBox>
+      <LinkOverlay href={to} isExternal={isExternal}>
+        <Container
+          textDecoration="none"
+          width={isLarger ? "" : "100%"}
```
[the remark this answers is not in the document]
Master7130: Oh ok.

src/components/DocLink.tsx
```diff
@@ -1,94 +1,67 @@
...
-        <Title>{children}</Title>
-      </TextCell>
-      <Arrow
-        isExternal={isExternal}
-        color={theme.colors.text}
-        name="arrowRight"
-      />
-    </Container>
+    <LinkBox>
+      <LinkOverlay href={to} isExternal={isExternal}>
```
pettinarip, line 15: This overlay should only wrap the main text of the DocLink (see suggestion below).

src/components/DocLink.tsx
```diff
@@ -1,94 +1,67 @@
...
+            flex={1}
+            justifyContent="space-between"
+          >
+            <Flex align="center">
+              <Emoji fontSize="md" mr={4} text=":page_with_curl:" />
+            </Flex>
+            <Box flex={1} flexDirection="column" color="text">
+              <Text color="text300" fontWeight="semibold" margin={0}>
+                {children}
+              </Text>
```
pettinarip, line 45:
```suggestion
            <LinkOverlay href={to} isExternal={isExternal}>
              <Text color="text300" fontWeight="semibold" margin={0}>
                {children}
              </Text>
            </LinkOverlay>
```

src/components/DocLink.tsx
```diff
@@ -1,94 +1,67 @@
...
-      <Arrow
-        isExternal={isExternal}
-        color={theme.colors.text}
-        name="arrowRight"
-      />
-    </Container>
+    <LinkBox>
+      <LinkOverlay href={to} isExternal={isExternal}>
+        <Box
+          textDecoration="none"
```
pettinarip, line 17: Should be part of the `LinkOverlay`.

src/components/DocLink.tsx
```diff
@@ -1,94 +1,67 @@
...
+              <Emoji fontSize="md" mr={4} text=":page_with_curl:" />
+            </Flex>
+            <Box flex={1} flexDirection="column" color="text">
+              <Text color="text300" fontWeight="semibold" margin={0}>
+                {children}
+              </Text>
+            </Box>
+            <Icon
+              as={AiOutlineArrowRight}
+              color="text"
```
pettinarip, line 49: Do we need to declare `color="text"` in each child? given that we are already declaring it on the wrapper component?

src/components/DocLink.tsx
pettinarip: We also need to use our custom `Link` in this `LinkOverlay` in order to make it work correctly for internal navigation.

Something like:
```tsx
import Link from "./Link"

<LinkOverlay
  as={Link}
  href={to}
  hideArrow
  color="text"
  textDecoration="none"
  _hover={{ textDecoration: "none" }}
>
```

You can see a similar implementation in this open PR #7721

pettinarip (changes requested):
Thanks for the recent changes @Master7130 this is looking great. I left a few more changes that we need to make this work well.

src/components/DocLink.tsx
Master7130: No we don't. Thanks for pointing that out!

Master7130:
Hey @pettinarip can you take a look at my recent changes?

src/components/DocLink.tsx
```diff
@@ -1,94 +1,72 @@
...
+        >
+          <Flex align="center">
+            <Emoji fontSize="md" mr={4} text=":page_with_curl:" />
+          </Flex>
+          <Box flex={1} flexDirection="column">
+            <LinkOverlay
+              href={to}
+              as={Link}
+              isExternal={isExternal}
+              _hover={{ textDecoration: "none" }}
```
pettinarip, line 46: image

we need to override the `textDecoration` for the default state as well.
```suggestion
              textDecoration="none"
              _hover={{ textDecoration: "none" }}
```

pettinarip (changes requested):
I've detected a few style issues. With those things fixed we should be fine to merge this. Good job @Master7130 Thanks for the work you did here 💯

src/components/DocLink.tsx
```diff
@@ -1,94 +1,72 @@
...
+        >
+          <Flex align="center">
+            <Emoji fontSize="md" mr={4} text=":page_with_curl:" />
+          </Flex>
+          <Box flex={1} flexDirection="column">
+            <LinkOverlay
+              href={to}
+              as={Link}
+              isExternal={isExternal}
+              textDecoration="none"
```
pettinarip, line 46: I meant to keep both `textDecoration`s (in both states). Adding the hover one back 👍🏼 
```suggestion
              textDecoration="none"
              _hover={{ textDecoration: "none" }}
```

pettinarip:
@all-contributors please add @Master7130 for code

Master7130:
> Great job @Master7130!
> 
> @Master7130 be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And we've just released our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) also 🥳!

Great! Thanks for guiding me through this!

PR #7785, opened 2022-09-10
refactor `EthVideo.tsx` component with chakra UI

MahendraBishnoi29:
## Description
Migrate `EthVideo.tsx` component to Chakra.

Includes refactor of `div` component.

## Related Issue
#6374

### DIFF
--- a/src/components/EthVideo.tsx
+++ b/src/components/EthVideo.tsx
@@ -1,6 +1,6 @@
 import React from "react"
 import { useTheme } from "@emotion/react"
-import { AspectRatio } from "@chakra-ui/react"
+import { Box } from "@chakra-ui/react"
 
 import darkVideo from "../assets/ethereum-hero-dark.mp4"
 import lightVideo from "../assets/ethereum-hero-light.mp4"
@@ -18,7 +18,7 @@ const EthVideo: React.FC<IProps> = ({ className, videoSrc }) => {
   const src = videoSrc ? videoSrc : isDarkTheme ? darkVideo : lightVideo
 
   return (
-    <AspectRatio ratio={1 / 1} className={className}>
+    <Box className={className}>
       <video
         id="hero-video"
         width="100%"
@@ -29,7 +29,7 @@ const EthVideo: React.FC<IProps> = ({ className, videoSrc }) => {
         loop
         muted
       />
-    </AspectRatio>
+    </Box>
   )
 }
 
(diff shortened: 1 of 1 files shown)


pettinarip (changes requested):
@MahendraBishnoi29 thanks for the PR.

For some reason, I don't see the video. You can check that in the preview deploy generated here https://ethereumorgwebsitedev01-mahendrabishnoi29ethereum00379.gtsb.io/en/eth

image

You can compare it with the production version. https://ethereum.org/en/eth/

I think it is because of how the parent container is positioned. I would recommend avoiding using AspectRatio in this case since I don't see an easy solution. One thing we could do here is to wrap the video tag with a Box and use the same styles we had.

```tsx
<Box as="video" ....>
```

src/components/EthVideo.tsx
```diff
@@ -3,6 +3,7 @@ import { useTheme } from "@emotion/react"
 
 import darkVideo from "../assets/ethereum-hero-dark.mp4"
 import lightVideo from "../assets/ethereum-hero-light.mp4"
+import { AspectRatio } from "@chakra-ui/react"
```
pettinarip, line 6: import order
```suggestion
import { AspectRatio } from "@chakra-ui/react"

import darkVideo from "../assets/ethereum-hero-dark.mp4"
import lightVideo from "../assets/ethereum-hero-light.mp4"

```

MahendraBishnoi29:
> @MahendraBishnoi29 thanks for the PR.
> 
> For some reason, I don't see the video. You can check that in the preview deploy generated here https://ethereumorgwebsitedev01-mahendrabishnoi29ethereum00379.gtsb.io/en/eth
> 
> image
> 
> You can compare it with the production version. https://ethereum.org/en/eth/
> 
> I think it is because of how the parent container is positioned. I would recommend avoiding using AspectRatio in this case since I don't see an easy solution. One thing we could do here is to wrap the video tag with a Box and use the same styles we had.
> 
> ```tsx
> <Box as="video" ....>
> ```

alright lemme fix it.

MahendraBishnoi29:
@pettinarip I did your suggested changes have a look. 🛠

pettinarip (approved):
🚀 thanks @MahendraBishnoi29 

@MahendraBishnoi29 be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And we've just released our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) also 🥳!

pettinarip:
@all-contributors please add @MahendraBishnoi29 for code

PR #7786, opened 2022-09-11
Update index.md

UniWrighte:
update to current username and password on img


corwintines:
Going to close this PR. I don't think this change needs to be made. @UniWrighte if you disagree, please reopen and provide a rationale for the change :)

PR #7815, opened 2022-09-12
Fix embed YouTube width

pettinarip:
Context: https://discord.com/channels/714888181740339261/727898649006309377/1018969307344031845

## Description

Use `YouTube` component instead of the fixed width iframe.


samajammin (approved):
👍

PR #7816, opened 2022-09-12
Migrate Checkbox from styled component to ChakraUI

marcellamalune:
## Description

Migrate Checkbox from styled component to ChakraUI

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/6374

### DIFF
--- a/src/components/Checkbox.tsx
+++ b/src/components/Checkbox.tsx
@@ -1,32 +1,6 @@
 import React from "react"
-import {
-  Checkbox as ChakraCheckbox,
-  Icon,
-  VisuallyHiddenInput,
-  Text,
-  VStack,
-} from "@chakra-ui/react"
+import { Box, VisuallyHiddenInput, Text } from "@chakra-ui/react"
 
-// Checkbox Container
-const containerProps = {
-  display: "inline-block",
-  verticalAlign: "middle",
-}
-
-// Hidden input
-
-// StyledCheckbox
-
-// Icon
-const svgProps = {
-  fill: "none",
-  stroke: `${(props) => props.theme.colors.white}`,
-  strokeWidth: "2px",
-  visibility: `${(props) => (props.checked ? "visible" : "hidden")}`,
-}
-// Label
-
-// Props
 export interface IProps {
   children?: React.ReactNode
   callback?: () => void
@@ -35,7 +9,6 @@ export interface IProps {
   size?: number
 }
 
-// Checkbox
 const Checkbox: React.FC<IProps> = ({
   callback,
   checked,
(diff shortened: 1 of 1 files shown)


src/components/Checkbox.tsx
```diff
@@ -75,21 +22,64 @@ const Checkbox: React.FC<IProps> = ({
...
-      <StyledCheckbox
-        aria-hidden="true"
-        checked={checked}
-        className="styled-checkbox"
-        size={size}
-      >
-        <Icon checked={checked} viewBox="0 0 24 24">
+    <Box {...checkboxContainerAttributes}>
+      <VisuallyHiddenInput {...hiddenCheckboxAttributes} />
+      <Box {...styledCheckboxAttributes}>
```
minimalsm, line 66: When using a screen reader, I'd expect the screen reader to announce that something is checked when it has been selected. Currently, it announces everything as 'unticked' in every state.

I think we should also pass `checked={checked}` to this Box to fix that.

pettinarip: @minimalsm looking at the previous code, we were using the `checked` prop just for style matters. I don't think that would change anything else.

src/components/Checkbox.tsx
```diff
@@ -75,21 +22,64 @@ const Checkbox: React.FC<IProps> = ({
...
-    <CheckboxContainer className={className} onClick={handleClick}>
-      <HiddenCheckbox type="checkbox" checked={checked} readOnly {...rest} />
-      <StyledCheckbox
-        aria-hidden="true"
-        checked={checked}
-        className="styled-checkbox"
-        size={size}
-      >
-        <Icon checked={checked} viewBox="0 0 24 24">
+    <Box {...checkboxContainerAttributes}>
```
pettinarip, line 64: For consistency matters across the codebase, do not wrap the styles in an object. Lets just declare the props in the tag itself.

src/components/Checkbox.tsx
```diff
@@ -75,21 +22,64 @@ const Checkbox: React.FC<IProps> = ({
...
+      <VisuallyHiddenInput {...hiddenCheckboxAttributes} />
+      <Box {...styledCheckboxAttributes}>
+        <svg
```
pettinarip, line 67: Wondering if we could use the Chakra `Icon` component in here instead of the <svg>. Check [this](https://chakra-ui.com/docs/components/icon/usage#using-the-icon-component) for reference.

pettinarip (changes requested):
Hey @marcellamalune good job on this one 👏🏼 

Regardless that I think these changes are fine for now, I think in the next iteration (different PR) we should delete this component and use the [native Chakra checkbox](https://chakra-ui.com/docs/components/checkbox/usage).

I encourage you if you want to help us with that. Basically, we should:
- delete this file/component
- replace all the uses of the old component with the Chakra checkbox
- style the Chakra checkbox with our custom styles in `src/@chakra-ui/gatsby-plugin/components` as we are doing for example with the `Button`

src/components/Checkbox.tsx
marcellamalune: @minimalsm @pettinarip I'm not sure how to proceed on this. I tried [this screen reader](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn) on the Gatsby preview build and it announces the checkbox as "ticked" if it is. Though I don't know if I'm testing it properly.

Looking at [this accessibility docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role) seems like we should use `role="checkbox"` and `aria-checked={checked}`.

src/components/Checkbox.tsx
marcellamalune: I struggled with the Chakra Icon a few hours but it turned out that component comes with an amount of CSS that I couldn't figure out how to override properly... The svg was moved a little bit lower and polyline even lower.
This was the result regardless of my attempts: 
Screenshot 2022-09-13 at 22 17 12

I propose to keep temporarily the `<svg>` until the next iteration with the Chakra Checkbox. Or if you have any suggestions on how to use the Chakra Icon I'll try again.

src/components/Checkbox.tsx
marcellamalune: In the last commit I removed style objects and declared the props as recommended.

marcellamalune:
@pettinarip I would be happy to help with the migration from this component to the native Chakra checkbox!

pettinarip (approved):
🚀

src/components/Checkbox.tsx
pettinarip: Since we are planning on using the Chakra native checkbox, we can skip this for now.

pettinarip:
> @pettinarip I would be happy to help with the migration from this component to the native Chakra checkbox!

Awesome @marcellamalune 💪🏼 LMK if you need help with that (on discord or in the GH issue).

src/components/Checkbox.tsx
pettinarip: @minimalsm let me know if you have any concerns about this before merging it.

pettinarip: As chatted with @minimalsm we agree on this. We will cover these things on the implementation of the native Chakra checkbox.

pettinarip:
@marcellamalune be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And we've just released our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) also 🥳!

pettinarip:
@all-contributors please add @marcellamalune for code

PR #7832, opened 2022-09-14
fix(content): Fixes the link to bordel

tarunbatra:
## Description
The link without the protocol is pointing to https://ethereum.org/enbordel.wtf which is the current address + the link. In order for it to be a standalone external link, the protocol can be added to it.

See it here at: https://ethereum.org/en/upgrades/merge/#wen-merge


minimalsm (approved):
Hey @tarunbatra, nice catch and thanks for the fix 💪

PR #7853, opened 2022-09-15
Fix thumbsdown icon [Fixes #7851]

### LINKED ISSUE #7851: Feedback card shows "thumbsup" emoji for both buttons
### Describe the bug

Feedback card shows "thumbsup" emoji for both buttons

### To reproduce

1. Go to any page
2. View feedback card at bottom

### Expected behavior

The "No" button should display a "thumbsdown" emoji.

### Screenshots

Image 2022-09-14 at 11 51 37 PM

All

All

### Additional context

_No response_

samajammin:
## Description

Fix regression to flip thumb icon

## Related Issue

Fixes #7851


samajammin:
I suspect there's a better Chakra-style approach to this. I tried using `transform` style prop but realized we're using just a vanilla SVG for the thumb icon.

On that note - should we? Seems we should always use our `Icon` component vs. importing raw SVG files into components? cc @wackerow @pettinarip @nhsz?

pettinarip:
> I suspect there's a better Chakra-style approach to this. I tried using `transform` style prop but realized we're using just a vanilla SVG for the thumb icon.

Oops, my bad. Yes, in this case you could achieve the same by using the `Icon` component from Chakra:
```tsx
<Icon as={ThumbsUp} transform="scaleY(-1)" />
```
Will create a PR soon to adjust that.

> On that note - should we? Seems we should always use our `Icon` component vs. importing raw SVG files into components? cc @wackerow @pettinarip @nhsz?

No, we are deprecating the usage of our custom `Icon` component in favor of just using the Chakra `Icon` component. In this case, importing just the svg worked fine. But in case you need to add styles to it then you need to warp it with the `Icon` component.

src/components/FeedbackCard.tsx
```diff
@@ -107,7 +111,7 @@ const FeedbackCard: React.FC<IProps> = ({
               </Button>
               <Button
                 variant="outline-color"
-                leftIcon={<ThumbsUp className="flip" />}
+                leftIcon={<ThumbsDown />}
```
pettinarip, line 114:
```suggestion
                leftIcon={<Icon as={ThumbsUp} transform="scaleY(-1)" fontSize="25px" />}
```

pettinarip (changes requested):

src/components/FeedbackCard.tsx
minimalsm: @pettinarip going to take this over and get it over the line.

pettinarip (approved):

PR #7869, opened 2022-09-15
Feat/convert callout banner chakra

vdusart:
Convert the CalloutBanner component to Chakra

## Related Issue

Epic: #6374


src/components/CalloutBanner.tsx
```diff
@@ -1,72 +1,11 @@
...
-`
-
-const Content = styled.div`
-  padding-left: 2rem;
-  flex: 1 0 50%;
-  display: flex;
-  flex-direction: column;
-  justify-content: center;
-  @media (max-width: ${(props) => props.theme.breakpoints.l}) {
-    margin-top: 2rem;
```
pettinarip, line 35: This one is missing

src/components/CalloutBanner.tsx
```diff
@@ -87,23 +26,51 @@ const CalloutBanner: React.FC<IProps> = ({
...
   id,
 }) => (
-  <StyledCard className={className} id={id}>
+  <Flex
+    direction={{ base: "column", lg: "row-reverse" }}
+    bg="layer2Gradient"
+    p={{ base: 8, sm: 12 }}
+    mx={{ base: 8, lg: 4 }}
+    mt={{ base: 16, lg: 24 }}
+    mb={{ base: 16, lg: 4 }}
```
pettinarip, line 35:
```suggestion
    mb={{ base: 16, lg: 40 }}
```

src/pages/upgrades/get-involved/index.tsx
```diff
@@ -455,7 +445,12 @@ const GetInvolvedPage = ({
         {getClientCards("cl")}
       </Content>
       <Staking>
-        <StyledCalloutBanner
+        <CalloutBanner
+          bg="transparent"
+          w={{ base: "full", md: "null" }}
+          p={{ base: 0, md: "null" }}
+          pt={{ base: 16, md: "null" }}
+          ml={{ base: 0, md: "null" }}
```
pettinarip, line 453: `null` is not a valid css value.

I think it should be something like:
```suggestion
          w={{ base: "full" }}
          p={{ base: 0, sm: 12 }}
          pt={{ base: 16, sm: 12 }}
          ml={{ base: 0 }}
```

src/components/CalloutBanner.tsx
```diff
@@ -87,23 +26,51 @@ const CalloutBanner: React.FC<IProps> = ({
...
-    <Content>
-      <H2>
+    <Flex
+      flexGrow={1}
+      flexShrink={0}
+      flexBasis="50%"
+      direction="column"
+      justifyContent="center"
+      pl={{ base: 0, sm: 4, lg: 8 }}
+      w={{ base: "full", lg: "null" }}
```
pettinarip, line 58:
```suggestion
      w={{ base: "full", lg: "inherit" }}
```

pettinarip (changes requested):
Great @vdusart. Thanks for this refactor 👍🏼 I've left a few comments.

src/components/CalloutBanner.tsx
```diff
@@ -87,23 +26,51 @@ const CalloutBanner: React.FC<IProps> = ({
...
+    direction={{ base: "column", lg: "row-reverse" }}
+    bg="layer2Gradient"
+    p={{ base: 8, sm: 12 }}
+    mx={{ base: 8, lg: 4 }}
+    mt={{ base: 16, lg: 24 }}
+    mb={{ base: 16, lg: 4 }}
+    borderRadius="base"
+    className={className}
+    id={id}
+  >
```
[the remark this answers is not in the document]
vdusart: Oh I see it was for this reason then :+1:

src/components/CalloutBanner.tsx
vdusart: Yes I know but it's pretty weird because as you can see on the picture bellow from [here](https://ethereum.org/en/stablecoins)
image

It seems there is no margin top on the element...

pettinarip: Yeah ok, I think I understand what is going on. There seems to be too much overriding of styles in this component.

I think a more elegant way of fixing this is by removing all the default margin styles this component has and delegate that aspect of the styles to the consumer.

Later I'll jump into this PR and make some adjustments related to that. I'll:
- remove all the margins from the component container
- refactor all the consumers to pass the correct margins for each case

With this approach we will remove that "responsability" from the component and make it simpler.

corwintines (approved):

corwintines:
@corwintines 
- [ ] Go through codebase and add margins to any CalloutBanner implemenation

pettinarip (approved):
🚀

PR #7880, opened 2022-09-16
Delete index.md

chatchai-hub:


minimalsm (commented):
Hey... don't know what this is supposed to be but you're deleting the developer docs page in this PR (which, incase it isn't obvious, we don't want to do...)

Issue #7700, opened 2022-09-05
Please add instruction to client types on https://ethereum.org/en/developers/docs/nodes-and-clients for the Merge !


ytrezq:
How to connect execution to consensus client ? Are any consensus client compatible with any execution client ?
Is setting the consensus client to archive mode required for running aa full archive execution client or is it possible to set up an execution full archive node using a light consensus client ?

corwintines:
Thanks @ytrezq 

Since the merge has happened 2 years ago, and documentation on how to run Ethereum clients is now up to date on the site and with respective clients, going to close this out.

Issue #7401, opened 2022-08-11
"Nodes and Clients" page has outdated / missing information on clients' supported networks


ThreeMuskets12:
**Describe the bug**

On the ["Nodes and Clients"](https://ethereum.org/en/developers/docs/nodes-and-clients/) page, the 'networks' column of the execution clients section does not list Sepolia as a supported network for any clients.

Additionally, the 'networks' column of the consensus clients section does not list Sepolia's beacon chain or Ropsten's beacon chain as a supported network for any clients.

Also I feel like the way the 'networks' column of the consensus clients uses the beacon chain names for testnets instead of the execution chains for testnets is kind of confusing to end users, especially for networks that have already undergone The Merge. For example, it was discussed on Consensus Layer Call 91 that referring to the now-merged "Goerli + Prater" network should just be referred to as "Goerli". [Source](https://github.com/ethereum/pm/issues/566#issuecomment-1184148151) 

**To Reproduce**

1. Go to '["Nodes and Clients"](https://ethereum.org/en/developers/docs/nodes-and-clients/) page'
2. Scroll down to the table under 'EXECUTION CLIENTS (FORMERLY 'ETH1 CLIENTS')'
3. See outdated / missing information in the 'Networks' column

Also
1. Go to '["Nodes and Clients"](https://ethereum.org/en/developers/docs/nodes-and-clients/) page'
2. Scroll down to the table under 'CONSENSUS CLIENTS (FORMERLY 'ETH2 CLIENTS')'
3. See outdated / missing information in the 'Networks' column

**Expected behavior**

-Sepolia should be listed as a supported network for all execution clients that support it
-Sepolia and Ropsten should be listed as supported networks for all consensus clients that support it
-[Potentially] network names for consensus clients should not be the name of the networks' beacon chain, and should instead be the name of the network itself. (IE not 'Prater' but 'Goerli')

**Desktop (please complete the following information):**

- OS: All
- Browser: All
- Version: All

corwintines:
Closing with the above PR's merged in

PR #7884, opened 2022-09-16
minor edit in docs to correspond to post-Merge reality

Aelay:
## Description
Update NB on the https://ethereum.org/en/developers/docs/nodes-and-clients page to reflect that it is not possible to run an execution client without a consensus client.

## Related Issue
https://github.com/ethereum/ethereum-org-website/issues/7700
https://github.com/ethereum/ethereum-org-website/issues/7401


jmcook1186 (approved):
thanks for catching this!

PR #7887, opened 2022-09-16
Migrate Morpher to Chakra

laurentlucian:
## Description

- Migrates `Morpher` component from `emotion` to `chakra-ui`

## Related Issue

#6374

### DIFF
--- a/src/components/Morpher.tsx
+++ b/src/components/Morpher.tsx
@@ -145,9 +145,6 @@ const Morpher = () => {
       fontSize="1rem"
       color="text"
       _hover={{ color: "primary" }}
-      sx={{
-        "&.active": { fontWeight: "bold" },
-      }}
       to="/languages/"
     >
       <span>{state.text}</span>
--- a/src/components/Morpher.tsx
+++ b/src/components/Morpher.tsx
@@ -1,19 +1,5 @@
 import React, { useState, useEffect } from "react"
-import styled from "@emotion/styled"
 import Link from "./Link"
-
-const NavLink = styled(Link)`
-  text-decoration: none;
-  font-size: 1rem;
-  color: ${(props) => props.theme.colors.text};
-  &:hover {
-    text-decoration: none;
-    color: ${(props) => props.theme.colors.primary};
-  }
-  &.active {
-    font-weight: bold;
-  }
-`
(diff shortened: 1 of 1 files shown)


src/components/Morpher.tsx
```diff
@@ -1,20 +1,6 @@
...
 
-const NavLink = styled(Link)`
-  text-decoration: none;
-  font-size: 1rem;
-  color: ${(props) => props.theme.colors.text};
-  &:hover {
-    text-decoration: none;
-    color: ${(props) => props.theme.colors.primary};
-  }
-  &.active {
```
laurentlucian, line 13: I couldn't tell what the `.active` class is doing so I couldn't compare, but decided to leave as is as it's not in the scope of this PR to fix this

pettinarip: I think that is fine. We were not using it before. This class is related to the active link class that Gatsby allows you to pass. https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-link/#link-api-surface-area

src/components/Morpher.tsx
```diff
@@ -154,9 +140,18 @@ const Morpher = () => {
...
 
   return (
-    <NavLink to="/languages/">
+    <Link
+      textDecor="none"
+      fontSize="1rem"
+      color="text"
+      _hover={{ color: "primary" }}
+      sx={{
+        "&.active": { fontWeight: "bold" },
```
pettinarip, line 149: Lets remove this for now since it is not doing anything

pettinarip (changes requested):
Cool, thanks for the PR @laurentlucian 💪🏼

src/components/Morpher.tsx
```diff
@@ -154,9 +140,18 @@ const Morpher = () => {
   }, [])
 
   return (
-    <NavLink to="/languages/">
+    <Link
+      textDecor="none"
+      fontSize="1rem"
```
pettinarip, line 145: could we do
```suggestion
      fontSize="md"
```

pettinarip (changes requested):

pettinarip (approved):
🚀 great! ty @laurentlucian!

Issue #7866, opened 2022-09-15
Bug report


reidjs:
### Describe the bug

Bullet points are misaligned on https://ethereum.org/en/staking/saas/#saas-providers

![image](https://user-images.githubusercontent.com/16528136/190452178-92b6b316-e58c-4730-ae2c-7af9b5e8f345.png)

### To reproduce

1. Go to https://ethereum.org/en/staking/saas/#saas-providers
2. Look at the staking SaaS provider boxes

### Expected behavior

This is by adding `margin: 1rem auto 1rem 0` to the ul > li > p elements
![image](https://user-images.githubusercontent.com/16528136/190453346-fb35bddb-e16d-4bbe-bb04-6b10c9cde42c.png)

### Desktop (please complete the following information)

Chrome

martinyung:
I would like to take this, thank you

corwintines:
Closing, was fixed in the above PR

PR #7891, opened 2022-09-17
Fix bullet points misalign in staking - saas providers

martinyung:
Fix the bullet point for <p/> tag misalignment in issue #7866

## Description

add `margin: 1rem auto 1rem 0` into p tag as suggested

## Related Issue
#7866


corwintines (approved):
Thanks, @martinyung

PR #7906, opened 2022-09-18
Update feedback components for Chakra

wackerow:
## Description
- Update FeedbackWidget and FeedbackCard using Chakra-UI instead of emotion
- Patches FeedbackWidget button colors
- Patches FeedbackWidget to reinitialize component state when changing URL paths
- Patches FeedbackWidget to default keyboard focus to close button instead of "Yes" answer on widget open
- Update Nav icons to use IconButton

## Related Issue
#6374


src/components/FeedbackWidget.tsx
```diff
@@ -17,157 +16,63 @@ import { translateMessageId } from "../utils/translations"
 import { useOnClickOutside } from "../hooks/useOnClickOutside"
 import { useKeyPress } from "../hooks/useKeyPress"
 import { useSurvey } from "../hooks/useSurvey"
+import { MdClose } from "react-icons/md"
```
pettinarip, line 19: import order. Should be at the top with the external lib imports

src/components/NakedButton.tsx
```diff
@@ -1,13 +1,24 @@
-import styled from "@emotion/styled"
+import React from "react"
+import { Button, ButtonProps } from "@chakra-ui/react"
+
+const NakedButton: React.FC<ButtonProps> = ({ children, ...props }) => (
```
pettinarip, line 4: Have you check the `ghost` variant? https://chakra-ui.com/docs/components/button/usage#button-variants

I think we should remove this component and start using `<Button variant="ghost">`. Any specific style we want to do on top of the default variant styles, we could add them on the Chakra theme definition as we did with the other variants.

Check the `variants` key in [here](https://github.com/ethereum/ethereum-org-website/blob/dev/src/%40chakra-ui/gatsby-plugin/components/Button.ts).

src/components/FeedbackWidget.tsx
```diff
@@ -17,157 +16,63 @@ import { translateMessageId } from "../utils/translations"
...
-    transform: scale(1.1);
-    transition: transform 0.2s ease-in-out;
-  }
-  transition: transform 0.2s ease-in-out;
-`
-
-export interface IProps {
-  className?: string
+}
+const FixedDot: React.FC<FixedDotProps> = ({
```
pettinarip, line 24: Just want to mention that there is a Chakra component called `IconButton` https://chakra-ui.com/docs/components/icon-button that perhaps could be a replacement for this. Not saying that this is in the scope of this PR but perhaps it is a better solution in the long term.

src/components/Layout.tsx
```diff
@@ -213,7 +213,7 @@ const Layout: React.FC<IProps> = ({
               <VisuallyHidden isHidden={isZenMode}>
                 <Footer />
               </VisuallyHidden>
-              <FeedbackWidget />
+              <FeedbackWidget location={path} />
```
pettinarip, line 216: One common practice to deal with this is to do
```tsx
<FeedbackWidget key={path} />
```
that way, the component is going to remount everytime the path change and you will not have to do any manual reset in the inner component logic.

pettinarip (changes requested):
Nicely done @wackerow left a few comments

src/components/FeedbackWidget.tsx
wackerow: Yes this component has more changes to come in a separate PR, thanks for noting!

src/components/FeedbackWidget.tsx
```diff
@@ -220,82 +124,131 @@ const FeedbackWidget: React.FC<IProps> = ({ className }) => {
...
               fallbackFocus: `#dot`,
             }}
           >
-            <Container
-              bottomOffset={bottomOffset}
-              ref={containerRef}
-              className={className}
+            <Flex
               id="modal"
+              ref={containerRef}
```
pettinarip, line 148: I think we should aim to use the Chakra Popover in this component. https://chakra-ui.com/docs/components/popover we will have better a11y and will abstract us for a bunch of things we are doing here (close icon, modal styles, focus trap, etc).

Not saying this should be in the scope of this PR but I'll create a new issue to tackle that after we merge this.

src/@chakra-ui/gatsby-plugin/components/Button.ts
```diff
@@ -70,5 +70,19 @@ export const Button: ComponentStyleConfig = {
       color: "primary",
       borderColor: "primary",
     },
+    icon: {
+      appearance: "none",
+      display: "grid",
+      placeItems: "center",
```
pettinarip, line 76: The button content is centered by default. If you inspect the [Button](https://chakra-ui.com/docs/components/button/usage) you will see that it is flex displayed and is centered.
```suggestion
```

src/components/Nav/index.tsx
```diff
@@ -433,20 +415,28 @@ const Nav: React.FC<IProps> = ({ path }) => {
...
+                    as={isDarkTheme ? MdWbSunny : MdBrightness2}
+                    fontSize="2xl"
+                  />
+                }
+                variant="icon"
+                _hover={{ color: "primary" }}
+                onClick={toggleColorMode}
+              />
+              <ButtonLink to="/languages/" variant="icon">
+                <Flex flexWrap="nowrap" alignItems="center">
```
pettinarip, line 433: It is not semantically correct to place a div inside the button. I've removed it and is not wrapping.
```suggestion
```

src/components/FeedbackWidget.tsx
```diff
@@ -220,82 +124,131 @@ const FeedbackWidget: React.FC<IProps> = ({ className }) => {
...
                   <Translation id="feedback-widget-thank-you-timing" />
-                </p>
+                </Text>
               )}
-              <ButtonContainer>
+              <Flex flexWrap="nowrap" gap={6} width="full">
                 {feedbackSubmitted ? (
                   <Button
-                    variant="outline-color"
+                    variant="solid"
```
pettinarip, line 218: `solid` is the default variant, not need to define it.
```suggestion
```

pettinarip (changes requested):
Nice! I think we can still remove some styles from our custom `icon` variant and fix the markdown semantics on the `Languages` button.

pettinarip (approved):
Awesome! 🚀 

I've just merged `dev` to fix merge conflicts 👍🏼

PR #7921, opened 2022-09-19
update issuance page

jmcook1186:
## Description

updates ETH issuance page for post-merge world.

## Related Issue

#7075

### DIFF
--- a/src/content/upgrades/merge/issuance/index.md
+++ b/src/content/upgrades/merge/issuance/index.md
@@ -15,78 +15,24 @@ The **issuance** of ETH is the process of creating ETH that did not previously e
 emoji=":chart_decreasing:"
 title="ETH issuance tldr">
 
-- Mining rewards ~13,000 ETH/day pre-merge
-- Staking rewards ~1,600 ETH/day pre-merge
-- **After The Merge, only the ~1,600 ETH per day will remain, dropping total new ETH issuance by ~90%**
-- The burn: At an average gas price of at least 16 gwei, at least 1,600 ETH is burned every day, which effectively brings net ETH inflation to zero or less post-merge.
+- Before The Merge ~13,000 ETH/day was issued to pay miners
+- After The Merge, issuance depends upon the number of active validators on the network, so varies slot to slot.
+- The total issuance is approximaely 90% smaller now than it was pre-merge.
+- Ether is also burned in every slot, with a greater amount burned when gas is higher. This means ether can be inflationary or deflationary depending on network demand.
 
 </Card>
 
-How ETH gets issued will change at the time of The Merge. Currently, new ETH is issued from two sources: the execution layer (i.e. Mainnet) and the consensus layer (i.e. Beacon Chain). After The Merge, issuance from the execution layer will go to zero. Let's break this down.
+## Issuance {#issuance}
 
-[More on The Merge](/upgrades/merge/)
-
-## Pre-merge {#pre-merge}
-
-### Execution layer issuance {#el-issuance-pre-merge}
-
-Under proof-of-work, miners only interact with the execution layer and are rewarded with block rewards if they are the first miner to solve the next block. Since the [Constantinople upgrade](/history/#constantinople) in 2019 this reward has been 2 ETH per block. Miners are also rewarded for publishing [ommer](/glossary/#ommer) blocks, which are valid blocks that don't end up in the longest/canonical chain. These rewards max out at 1.75 ETH per ommer, and are _in addition to_ the reward issued from the canonical block. Mining is an economically intensive activity, requiring high levels of ETH issuance to sustain.
-
-### Consensus layer issuance {#cl-issuance-pre-merge}
-
-The [Beacon Chain](/history/#beacon-chain-genesis) went live in 2020. Instead of miners, it is secured by validators using proof-of-stake. This chain was bootstrapped by Ethereum users depositing ETH one-way into a smart contract on Mainnet, which the Beacon Chain listens to, crediting the user with an equal amount on the new chain. Until The Merge happens, the Beacon Chain's validators are not processing transactions and are essentially coming to consensus on the state of the validator pool itself.
-
-Validators on the Beacon Chain are rewarded with ETH for attesting to the state of the chain and proposing blocks. Rewards (or penalties) are calculated and distributed at each epoch (every 6.4 minutes) based on validator performance. The validator rewards are **significantly** less than the miner rewards issued on proof-of-work (2 ETH every ~13.5 seconds), as operating a validating node is not an economically intense activity and thus does not require or warrant as high a reward.
-
-### Pre-merge issuance breakdown {#pre-merge-issuance-breakdown}
-
-Total ETH supply: **~119,300,000 ETH** (as of Q2 2022)
-
-**Execution layer issuance:**
-
-- Estimating at 2.08 ETH per 13.3 seconds\*: **~4,930,000** ETH issued in a year
-- Currently inflating at **~4.13%** (4.93M per year / 119.3M total)
-- \*This includes the 2 ETH per canonical block, plus an average of 0.08 ETH over time from ommer blocks. Also uses 13.3 seconds, the baseline block time target without any influence from a [difficulty bomb](/glossary/#difficulty-bomb). ([See source](https://bitinfocharts.com/ethereum/))
+Ethereum validators are rewarded with ETH for attesting to the state of the chain and proposing blocks. Rewards (or penalties) are calculated and distributed at each epoch (every 6.4 minutes) based on validator performance. The validator rewards are **significantly** less than the miner rewards that used to be issued on proof-of-work (2 ETH every ~13.5 seconds), as operating a validating node is not an economically intense activity and thus does not require or warrant as high a reward.
 
 **Consensus layer issuance:**
 
-- Using 13,000,000 total ETH staked, the rate of ETH issuance is ~1600 ETH/day ([See source](https://ultrasound.money/))
-- Results in **~584,000** ETH issued in a year
-- Currently inflating at **~0.49%** (584K per year / 119.3M total)
+Small rewards are paid validators who attest to and propose blocks. Validator rewards accrue to _validator balances_ that are managed within the consensus layer. These are separate Ethereum accounts to the accounts we're used to on Mainnet, and until the Shanghai upgrade funds from validator accounts will not be withdrawable/transferrable. This means that although new ETH is still being issued, 100% of it will be locked from the market until this upgrade occurs. When the Shanghai upgrade is rolled out, this ETH will become available.
 
-<InfoBanner>
-<strong>Total annual issuance rate: ~4.62%</strong> (4.13% + 0.49%)<br/><br/>
-<strong>~89.4%</strong> of the issuance is going to miners on the execution layer (4.13 / 4.62 * 100)<br/><br/>
-<strong>~10.6%</strong> is being issued to stakers on the consensus layer (0.49 / 4.62 * 100)
-</InfoBanner>
-
-## Post-merge {#post-merge}
-
-### Execution layer issuance {#el-issuance-post-merge}
-
-Execution layer issuance after The Merge will be zero. Proof-of-work will no longer be valid under the rules of consensus. All execution layer activity will be included in "beacon blocks", which are published and attested to by proof-of-stake validators.
-
-### Consensus layer issuance {#cl-issuance-post-merge}
-
-Consensus layer issuance will continue as before The Merge, with small rewards for validators who attest to and propose blocks. Validator rewards will continue to accrue to _validator balances_ that are managed within the consensus layer. These are separate Ethereum accounts to the accounts we're used to on Mainnet, and until the Shanghai upgrade funds from validator accounts will not be withdrawable/transferrable. This means that although new ETH is still being issued, 100% of it will be locked from the market until this upgrade occurs. When the Shanghai upgrade is rolled out, this ETH will become available.
-
-When validator withdrawals are enabled, stakers will be incentivized to remove their _earnings/rewards (balance over 32 ETH)_ as these funds are otherwise not contributing to their stake weight (which maxes as 32).
+When validator withdrawals are enabled, stakers will be incentivized to remove their _earnings/rewards (balance over 32 ETH)_ as these funds are otherwise not contributing to their stake weight (which maxes out at 32).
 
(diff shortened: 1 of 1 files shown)


src/content/upgrades/merge/issuance/index.md
```diff
@@ -15,79 +15,25 @@ The **issuance** of ETH is the process of creating ETH that did not previously e
 emoji=":chart_decreasing:"
 title="ETH issuance tldr">
 
-- Mining rewards ~13,000 ETH/day pre-merge
-- Staking rewards ~1,600 ETH/day pre-merge
-- **After The Merge, only the ~1,600 ETH per day will remain, dropping total new ETH issuance by ~90%**
-- The burn: At an average gas price of at least 16 gwei, at least 1,600 ETH is burned every day, which effectively brings net ETH inflation to zero or less post-merge.
+- Before The Merge ~13,000 ETH/day was issued to pay miners
+- After The Merge, issuance depends upon the number of active validators on the network, so varies slot to slot.
+- The total issuance is approximaely 90% smaller now than it was pre-merge.
```
SebastianSupreme, line 20: There's a typo.
Change it to "approximately".

SebastianSupreme:
There's a typo in line 20 of your suggested text.
Change it to "approximately".

src/content/upgrades/merge/issuance/index.md
```diff
@@ -15,79 +15,25 @@ The **issuance** of ETH is the process of creating ETH that did not previously e
...
 title="ETH issuance tldr">
 
-- Mining rewards ~13,000 ETH/day pre-merge
-- Staking rewards ~1,600 ETH/day pre-merge
-- **After The Merge, only the ~1,600 ETH per day will remain, dropping total new ETH issuance by ~90%**
-- The burn: At an average gas price of at least 16 gwei, at least 1,600 ETH is burned every day, which effectively brings net ETH inflation to zero or less post-merge.
+- Before The Merge ~13,000 ETH/day was issued to pay miners
+- After The Merge, issuance depends upon the number of active validators on the network, so varies slot to slot.
+- The total issuance is approximately 90% smaller now than it was pre-merge.
+- Ether is also burned in every slot, with a greater amount burned when gas is higher. This means ether can be inflationary or deflationary depending on network demand.
```
wackerow, line 21: Being a tldr section, my preference would be to keep some numbers for daily issuance under PoS available here, currently ~1700 ETH/day... I think it's helpful for people just looking for some simple numbers.

src/content/upgrades/merge/issuance/index.md
```diff
@@ -15,79 +15,25 @@ The **issuance** of ETH is the process of creating ETH that did not previously e
...
-</InfoBanner>
-
-## Post-merge {#post-merge}
-
-### Execution layer issuance {#el-issuance-post-merge}
-
-Execution layer issuance after The Merge will be zero. Proof-of-work will no longer be valid under the rules of consensus. All execution layer activity will be included in "beacon blocks", which are published and attested to by proof-of-stake validators.
-
-### Consensus layer issuance {#cl-issuance-post-merge}
+Small rewards are paid validators who attest to and propose blocks. Validator rewards accrue to _validator balances_ that are managed within the consensus layer. These are separate Ethereum accounts to the accounts we're used to on Mainnet, and until the Shanghai upgrade funds from validator accounts will not be withdrawable/transferrable. This means that although new ETH is still being issued, 100% of it will be locked from the market until this upgrade occurs. When the Shanghai upgrade is rolled out, this ETH will become available.
```
wackerow, line 31:
```suggestion
Small rewards are paid to validators who attest to and propose blocks. Validator rewards accrue to _validator balances_ that are managed within the consensus layer. These are separate Ethereum accounts to the accounts we're used to on Mainnet, and until the Shanghai upgrade funds from validator accounts will not be withdrawable/transferrable. This means that although new ETH is still being issued, 100% of it will be locked from the market until this upgrade occurs. When the Shanghai upgrade is rolled out, this ETH will become available.
```

Also, until withdrawals are confirmed, we should phrase this as being proposed for the Shanghai upgrade

wackerow (commented):
I'm personally not sure about making such a large swing in the structure of this content and would love to hear what others think as well.
I know we're in a post-merge world, but this page is directly under `upgrades/merge/issuance` and titled How The Merge impacts ETH supply—I would advocate for us keeping more of a comparison to the numbers from PoW. 

image

Just peeking at the ToC structure from before (right) to after (this PR, left), I worry we've lost some valuable content that helps break down actually how much changed from before the merge to now. 

I guess in short, personally didn't see this being such a large overhaul, but maybe I'm just partial so would love to hear from others.

src/content/upgrades/merge/issuance/index.md
jmcook1186: idk I removed it implies the daily issuance is a fixed number - I find this to be misleading - its dependent upon N active validators and their performance. The original and your comment already vary by 100ETH/day.
I guess its ok with appropriate caveats, but I preferred the takeaway being a) issuance is variable, b) its about 1/10th what it was before merge.

src/content/upgrades/merge/issuance/index.md
jmcook1186:
```suggestion
Small rewards are paid to validators who attest to and propose blocks. Validator rewards accrue to _validator balances_ that are managed within the consensus layer. These are separate Ethereum accounts to the accounts we're used to on Mainnet, and until the Shanghai upgrade funds from validator accounts will not be withdrawable/transferrable. This means that although new ETH is still being issued, 100% of it will be locked from the market until this upgrade occurs. When withdrawals are enabled, this ETH will become available.
```

src/content/upgrades/merge/issuance/index.md
```diff
@@ -96,6 +42,10 @@ Fee burning went live with <a href="/history/#london">the London upgrade</a> in
 
 On top of the fee burn implemented by the London upgrade, validators can also incur penalties for being offline, or worse, they can be slashed for breaking specific rules that threaten network security. These penalties result in a reduction of ETH from that validator's balance, which is not directly rewarded to any other account, effectively burning it from circulation.
 
+## Issuance estimates
```
minimalsm, line 45:
```suggestion
## Issuance estimates {#issuance-estimates}
```

src/content/upgrades/merge/issuance/index.md
minimalsm, line 21:
```suggestion
- ETH is burned in every slot, with a greater amount burned when gas is higher. This means ETH can be inflationary or deflationary depending on network demand.
```

minimalsm: The ideal scenario here would be live numbers. Maybe a big numbers component + Dune API could be used in the future to solve this? Three numbers: an average PoW issuance at some point close to The Merge (maybe the month leading up to it?), previous days' PoS issuance, and percentage reduction.

As much as I dislike seeing things written this way, the trade-off here might be something like: "_The total issuance is approximately 90% smaller now than it was pre-merge (~1700ETH as of October 2022)._" 

What do you think @jmcook1186 @wackerow?

src/content/upgrades/merge/issuance/index.md
```diff
@@ -96,6 +42,10 @@ Fee burning went live with <a href="/history/#london">the London upgrade</a> in
 
 On top of the fee burn implemented by the London upgrade, validators can also incur penalties for being offline, or worse, they can be slashed for breaking specific rules that threaten network security. These penalties result in a reduction of ETH from that validator's balance, which is not directly rewarded to any other account, effectively burning it from circulation.
 
+## Issuance estimates {#issuance-estimates}
+
+So, how much ether is actually issued? The precise amount changes in each slot because it depends upon the amount of ether staked by validators and their performance. At the time of writing (September 2022), net issuance (`issuance-burn`) since The Merge has been ~3100 ETH (if we were still using proof-of-work, 54,000 ETH would have been issued). Extrapolating the current network usage into the future gives an estimated 0.22%/yr net issuance. However, current network usage is very low, and more transactions lead to more ETH being burned. The website [ultrasound.money](https://ultrasound.money/) allows you to make predictions for the future net issuance under a range of market conditions.
```
minimalsm, line 47:
```suggestion
So, how much ETH is issued? The precise amount changes in each slot because it depends upon the amount of ether staked by validators and their performance. At the time of writing (September 2022), net issuance (`issuance-burn`) since The Merge has been ~3100 ETH (if we were still using proof-of-work, 54,000 ETH would have been issued). Extrapolating the current network usage into the future gives an estimated 0.22%/yr net issuance. However, current network usage is very low, and more transactions means more burned ETH. The website [ultrasound.money](https://ultrasound.money/) allows you to make predictions for the future net issuance under a range of market conditions.
```

minimalsm:
> I know we're in a post-merge world, but this page is directly under upgrades/merge/issuance and titled How The Merge impacts ETH supply—I would advocate for us keeping more of a comparison to the numbers from PoW.

We'll probably move this page out of here (/merge/) eventually and under a broader category of Ethereum's monetary policy, but I'd be okay with re-adding some additional details. 

> Just peeking at the ToC structure from before (right) to after (this PR, left), I worry we've lost some valuable content that helps break down actually how much changed from before the merge to now.

> I guess in short, personally didn't see this being such a large overhaul, but maybe I'm just partial so would love to hear from others.

Do you agree that the biggest value here was the comparison of the issuance/inflation numbers of pre/post-merge, or is it something else? 

I.e. I don't think we need to explain details of **how** issuance worked (block fees, tips, etc), but I do see the value of keeping comparative numbers (inflation rate, daily issuance), at least for now. Not a strong take, but I'm picturing a short section (PoW vs PoS: A comparison in numbers) that throws out the numbers (could we use a table?) to give people a broad tl;dr overview—similar to what we had before, but concise and consolidated. 

Other than that, I'm inclined to agree with Joseph's take here on making this page more focused on the here and now.

If you feel strongly otherwise (i.e. we should always explain how PoW issuance worked as a historical artefact), then maybe one option would be to break this into two pages (at some point): 

1. Issuance on Ethereum (`/issuance/`) — concerned with ETH issuance today and how it affects the overall supply
2. Issuance under proof-of-work (`/issuance/proof-of-work/`) — concerned with how things used to be, and comparisons of the old and the new

src/content/upgrades/merge/issuance/index.md
```diff
@@ -15,79 +15,25 @@ The **issuance** of ETH is the process of creating ETH that did not previously e
...
-Total ETH supply: **~119,300,000 ETH** (as of Q2 2022)
-
-**Execution layer issuance:**
-
-- Estimating at 2.08 ETH per 13.3 seconds\*: **~4,930,000** ETH issued in a year
-- Currently inflating at **~4.13%** (4.93M per year / 119.3M total)
-- \*This includes the 2 ETH per canonical block, plus an average of 0.08 ETH over time from ommer blocks. Also uses 13.3 seconds, the baseline block time target without any influence from a [difficulty bomb](/glossary/#difficulty-bomb). ([See source](https://bitinfocharts.com/ethereum/))
+Ethereum validators are rewarded with ETH for attesting to the state of the chain and proposing blocks. Rewards (or penalties) are calculated and distributed at each epoch (every 6.4 minutes) based on validator performance. The validator rewards are **significantly** less than the miner rewards that used to be issued on proof-of-work (2 ETH every ~13.5 seconds), as operating a validating node is not an economically intense activity and thus does not require or warrant as high a reward.
 
 **Consensus layer issuance:**
```
minimalsm, line 29: Do we need this heading? If we've got rid of execution layer issuance then we no longer have a point of comparison 🤔

src/content/upgrades/merge/issuance/index.md
jmcook1186: fine by me

src/content/upgrades/merge/issuance/index.md
jmcook1186, line 29:
```suggestion
```

wackerow:
I am not against updating this page to remove the comparisons to proof-of-work, but right now I we're kinda mixing the topics.

The current outline for this page is as follows:

```md
# How The Merge impacts ETH supply
## Issuance
## Issuance estimates
## Further reading
```

..and is located at `/upgrades/merge/issuance/`

If we're going to remove most of the comparisons to PoW, then I think the title here is misleading and needs to be changed (along with the path).

I want to get this all updated and completed as well, but I think the current page (comparing the loose numbers surrounding a comparison of pre-merge and post-merge) is still helpful... and if we want to have a more focused page that does _not_ compare to PoW, we can do that, but I think we should also try to expand the outline structure of the page a bit to provide some more details. I don't think removing details is a good approach here for the sake of simplicity.

@samajammin and others, curious your take on this.

samajammin:
I agree with @wackerow's comment here:

> I'm personally not sure about making such a large swing in the structure of this content and would love to hear what others think as well.
I know we're in a post-merge world, but this page is directly under upgrades/merge/issuance and titled How The Merge impacts ETH supply—I would advocate for us keeping more of a comparison to the numbers from PoW.

To me the goal of this page is to explain ETH issuance in the context of pre & post Merge, so explaining the before & after is useful here. I wouldn't delete most of the content you did in this PR @jmcook1186.

If we want to create a more general resouce on "ETH issuance" (which I think is a great idea!) then I would advocate for creating a subpage under `ethereum.org/eth/` (e.g. `ethereum.org/eth/issuance`).

veridelisi:
Could we add the following mathematical calculations?

There are (60/12)*60*24 = 7200 block per day

Each block targets 15*10^6 gas. 16 gwei is 16*10^-9 ether. 
Therefore, the amount of ether burned each day at a 16 gwei gas price is 
7200*16*(10^-9)*15*10^6 = around 1700 ETH. 

thanks a**elowsson**

wackerow:
Thanks for the suggestion @veridelisi!

The only adjustment I would make to this would be that we'd want to be solving for the $gwei$ amount based on the current ETH issuance per day (~1700), so we can re-write as such:

$7200 blocks/day * 15(10^6) gas/block * Y gwei/gas * 1 ETH/ 10^9 gwei == 1700 ETH/day$

Solving for $Y$:

$Y = (1700(10^9)) / (7200 * 15(10^6)) = (1700(10^3)/(7200 * 15)) = 15.74 gwei$

Or $16 gwei$ when accounting for significant digits (only 2 here).

Another way to rearrange this last step would be to replace $1700$ with a variable, say $X$ that represents the daily ETH issuance, and to simplify the rest to:

$Y = (X(10^3)/(7200 * 15)) = X / 108$

or simply $f(X) = X/108$ where $X$ is daily ETH issuance.

So, for example, if $X$ (daily ETH issuance) rises to 1800 based on total ETH staked, $f(X)$ (gwei required to offset all of the issuance) would then be $17 gwei$ (using 2 significant digits)

---
Working on an update to this and will add a section about this.

wackerow:
Just put up PR #8329 as a potential substitute for this PR, which leaves the page at its current path location of `/upgrades/merge/issuance` and intentionally keeps the comparison to pre-merge PoW for users to reference. Commented the same in the PR description, but I think we should table this PR for now and circle back in time to build out a new page nested under the `/eth` path for something that cuts out the pre-merge comparison and only focuses on the state of the network today/moving forward.

hoanglamcoin (approved):

PR #7936, opened 2022-09-20
rm gridlines and update label on energy plot

jmcook1186:
## Description

Removes gridlines from plot (because they had confusing spacing) and update label for "gaming" to clarify that it is a US only, not global, estimate.


src/components/EnergyConsumptionChart.tsx
```diff
@@ -86,7 +86,7 @@ const EnergyConsumptionChart: React.FC = () => {
       breakpoint: smallBreakpoint,
     },
     {
-      name: "Gaming",
+      name: "Gaming in the US",
```
wackerow, line 89: This wraps to three lines... small suggestion if we want to compress a little, but I'm fine with it either way
```suggestion
      name: "Gaming (USA)",
```

wackerow (approved):

src/components/EnergyConsumptionChart.tsx
jmcook1186: sure - this is similar to my original but @minimalsm suggested it might not be a smooth for translators. Happy either way.

corwintines (approved):

PR #7943, opened 2022-09-20
Fix code splitting issue

pettinarip:
With the inclusion of a custom `.babelrc` (on [v4](https://github.com/ethereum/ethereum-org-website/tree/v4.0.0)) code splitting stopped working correctly. This is due to [a Gatsby bug](https://github.com/gatsbyjs/gatsby/issues/35731).

## Description

In order to fix it, we have deleted the custom `.babelrc` (since there seems to be no way of ignoring it from Gatsby) and we have created a custom webpack config file to be used by `netlify-lambda`.


pettinarip (commented):
[FIXED]

~~We have a regression with Chakra styles. The buttons lost the styles~~
image

src/@chakra-ui/gatsby-plugin/components/index.ts
```diff
@@ -1,3 +1,9 @@
...
-export * from "./Tag"
+import { Button } from "./Button"
+import { Link } from "./Link"
+import { Tag } from "./Tag"
+
+export default {
+  Button,
+  Link,
+  Tag,
+}
```
pettinarip, line 9: This took me a very long time to detect 😆 hate these kind of issues.

Once I removed the `.babelrc` for some reason these exports stop working as I was expecting. I had to export these in a more manual way, with the expected structure.

wackerow (approved):
Looks good... As discussed, we can research further potentially using `@netlify/functions` with TypeScript support and remove usage of `netlify-lambda` going forward

corwintines (approved):

PR #7950, opened 2022-09-21
Fixed #7904, Respect current language for filtering tutorials

### LINKED ISSUE #7904: Tutorials linking to broken routes
### Describe the bug

Seems we're appending language routes incorrectly on some tutorials

### To reproduce

1. Go to https://ethereum.org/en/developers/tutorials/
2. Filter by 'UI'
3. View results - same tutorial listed multiple times, some linking to pages that don't exist (e.g. https://ethereum.org/en/id/developers/tutorials/nft-minter/)

### Expected behavior

Should show results only for pages that exist 😄 - should display tutorials of the current language (in this case, only /en/ routes).

All

All

### Additional context

_No response_

minhoryang:
## Description

Filtering tutorials on `/developers/tutorials` wasn't considering a current language option, but default `filteredTutorials` was considered well. So I copied the logic for filtering with the current language respect.

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/7904

### DIFF
--- a/src/pages/developers/tutorials.tsx
+++ b/src/pages/developers/tutorials.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react"
+import React, { useEffect, useState } from "react"
 import styled from "@emotion/styled"
 import { graphql, PageProps } from "gatsby"
 import { useIntl } from "react-intl"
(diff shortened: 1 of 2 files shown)


corwintines:
Hey @minhoryang!

Thanks for opening this. Your fix addressed some of the tag filtering which will be helpful, but while I was reviewing this I noticed a deeper issue. Turns out from the start, we are filtering out tutorials in the locale, instead of only showing tutorials in a locale. 

As this is a pretty big bug we want to get fixed for tomorrows deploy we will fix that in this PR as well. 

Again, thank you for opening this and helping as you did fix the original issue :)

pettinarip:
@corwintines I've added the final filter when the user selects a tag.

As a bonus, I've also added the filtering on the tags but its experimental, check it out and let me know what do you think. We could do it differently without filtering the tags and only updating the tag numbers.

corwintines (approved):
LGTM!

PR #7978, opened 2022-09-23
SectionNav refactor to ChakraUI

Master7130:
## Description

Converted SectionNav.tsx from styled-components to ChakraUI

## Related Issue

#6374

### DIFF
--- a/src/components/SectionNav.tsx
+++ b/src/components/SectionNav.tsx
@@ -12,7 +12,7 @@ const SectionNav: React.FC<IProps> = ({ children }) => (
     direction="column"
     color="textTableOfContents"
     marginTop={12}
-    justifyContent="space-between"
+    justify="space-between"
     background="searchBackground"
     borderRadius="2xl"
     border="1px solid"
--- a/src/components/SectionNav.tsx
+++ b/src/components/SectionNav.tsx
@@ -9,7 +9,7 @@ export interface IProps {
 
 const SectionNav: React.FC<IProps> = ({ children }) => (
   <Flex
-    flexDirection="column"
+    direction="column"
     color="textTableOfContents"
     marginTop={12}
     justifyContent="space-between"
--- a/src/components/SectionNav.tsx
+++ b/src/components/SectionNav.tsx
@@ -20,7 +20,7 @@ const SectionNav: React.FC<IProps> = ({ children }) => (
     padding={6}
     paddingBottom={0}
   >
-    <Flex display="flex" alignItems="flex-start">
+    <Flex align="flex-start">
       <Emoji text=":point_right:" mb={4} />
       <Heading
         as="h2"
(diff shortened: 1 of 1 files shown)


Master7130:
The letter spacing for the heading was previously 0.04rem but ChakraUI doesn't have a default value for that. The closest one was "wider" which is 0.05rem. Is that close enough or should I manually set it to 0.04rem?

src/components/SectionNav.tsx
```diff
@@ -1,50 +1,41 @@
...
   children?: React.ReactNode
 }
 
 const SectionNav: React.FC<IProps> = ({ children }) => (
-  <Card>
-    <TopContent>
-      <Emoji text=":point_right:" mb={`1em`} />
-      <Title>
+  <Flex
+    display="flex"
```
pettinarip, line 12:
```suggestion
```

src/components/SectionNav.tsx
```diff
@@ -1,50 +1,41 @@
...
+    flexDirection="column"
```
pettinarip, line 13:
```suggestion
    direction="column"
```

src/components/SectionNav.tsx
```diff
@@ -1,50 +1,41 @@
...
+    color="textTableOfContents"
+    marginTop={12}
+    justifyContent="space-between"
```
pettinarip, line 16:
```suggestion
    justify="space-between"
```

src/components/SectionNav.tsx
```diff
@@ -1,50 +1,41 @@
...
+    background="searchBackground"
+    borderRadius="2xl"
```
pettinarip, line 18:
```suggestion
    borderRadius="base"
```

src/components/SectionNav.tsx
```diff
@@ -1,50 +1,41 @@
...
+    border="1px solid"
+    borderColor="lightBorder"
+    padding={6}
+    paddingBottom={0}
+  >
+    <Flex display="flex" alignItems="flex-start">
```
pettinarip, line 24:
```suggestion
    <Flex align="flex-start">
```

pettinarip (changes requested):
Good job @Master7130. Left a few comments 👍🏼

Master7130:
I noticed that some of your suggestions were to change the style names from longer names like justifyContent to justify. Is using the shorthand form a best practice?

pettinarip:
> I noticed that some of your suggestions were to change the style names from longer names like justifyContent to justify. Is using the shorthand form a best practice?

Good question. It is (at least try) to maintain as much consistency as possible in the code base. Since by using [`Flex`](https://chakra-ui.com/docs/components/flex/usage#usage) you benefit from these shorthand props, we tend to use them to keep it more concise.

Master7130:
@pettinarip The build is failing here but it's succeeding locally. Can you rerun the build check?

pettinarip (approved):
🚀

PR #7982, opened 2022-09-23
Minor changes to the OR page

BenedettiLucca:
Added link for the Data Availability page.
Removed a repeated link in the 'Further reading' section


corwintines (approved):
Nice catch on the duplicate link, and good idea for the linking out here. Thanks, @BenedettiLucca!

PR #7986, opened 2022-09-23
Update Italian (it) Homepage, Essentials, and Use Ethereum buckets

minimalsm:
## Description

Update Italian (it) Homepage, Essentials, and Use Ethereum buckets


minimalsm:
LGTM

PR #7993, opened 2022-09-23
Update Catalan (ca) Exploring bucket

minimalsm:
## Description

Update Catalan (ca) Exploring bucket


minimalsm:
Closing this out as it's a regression (no value added, regression to English content)

PR #7994, opened 2022-09-23
Add an upcoming event in Ghana to this portal

iamGeorgePro:

### DIFF
--- a/src/data/community-meetups.json
+++ b/src/data/community-meetups.json
@@ -322,5 +322,11 @@
     "emoji": ":us:",
     "location": "San Diego",
     "link": "http://cryptosandiego.org"
+  },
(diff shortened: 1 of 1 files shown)


minimalsm (changes requested):
Hey @iamGeorgePro, thanks for the PR 💪 

Looks like Africa DeFiCon is an event (not a meetup)? We should add it to [our community-events file](https://github.com/ethereum/ethereum-org-website/blob/dev/src/data/community-events.json) and remove it from here. Community events are one time events, meetups are recurring meetings ☺️ 

LMK if you need a hand doing this.

corwintines (commented):
This was already added to events, so going to close this

PR #8007, opened 2022-09-24
Migrate Layer2ProductCard to Chakra 

ImMiguelP:
## Description
Migrates `Layer2ProductCard` component from `emotion` to `chakra-ui`

## Related Issue
#6374

### DIFF
--- a/src/components/Layer2ProductCard.tsx
+++ b/src/components/Layer2ProductCard.tsx
@@ -2,7 +2,7 @@
 import React from "react"
 import { GatsbyImage, IGatsbyImageData } from "gatsby-plugin-image"
 import { useIntl } from "react-intl"
-import { Box, Flex, Heading, Image, Text } from "@chakra-ui/react"
+import { Box, Center, Flex, Heading, Image, Text } from "@chakra-ui/react"
 
 // Components
 import ButtonLink from "./ButtonLink"
@@ -47,17 +47,14 @@ const Layer2ProductCard: React.FC<IProps> = ({
       direction="column"
       justify="space-between"
       bg="searchBackground"
-      borderRadius={1}
+      borderRadius="base"
       border={"1px solid lightBorder"}
       textDecoration="none"
       padding={2}
       _hover={{ transition: "transform 0.1s", transform: "scale(1.02)" }}
     >
-      <Flex
+      <Center
         bg={background}
-        direction="row"
-        justify="center"
-        align="center"
         boxShadow="inset 0px -1px 0px rgba(0, 0, 0, 0.1)"
         minH="200px"
       >
@@ -70,16 +67,10 @@ const Layer2ProductCard: React.FC<IProps> = ({
           maxW={{ base: "311px", sm: "372px" }}
           maxH={"257px"}
         />
-      </Flex>
-      <Flex
-        p={6}
-        textAlign="left"
-        h="100%"
-        direction="column"
-        className="hover"
-      >
+      </Center>
+      <Flex p={6} h="100%" direction="column">
         <Box>
-          <Heading as="h3" mt="0" mb={3}>
+          <Heading as="h3" size="md" mt={0} mb={3}>
             {name}
           </Heading>
           {children && (
--- a/src/components/Layer2ProductCard.tsx
+++ b/src/components/Layer2ProductCard.tsx
@@ -1,8 +1,8 @@
 // Libraries
 import React from "react"
-import styled from "@emotion/styled"
 import { GatsbyImage, IGatsbyImageData } from "gatsby-plugin-image"
 import { useIntl } from "react-intl"
+import { Box, Flex, Heading, Image, Text } from "@chakra-ui/react"
 
 // Components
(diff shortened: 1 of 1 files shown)


src/components/Layer2ProductCard.tsx
```diff
@@ -110,21 +41,61 @@ const Layer2ProductCard: React.FC<IProps> = ({
...
+      direction="column"
+      justify="space-between"
+      bg="searchBackground"
+      borderRadius={1}
+      border={"1px solid lightBorder"}
+      textDecoration="none"
+      padding={2}
+      _hover={{ transition: "transform 0.1s", transform: "scale(1.02)" }}
+    >
+      <Flex
```
pettinarip, line 56: Use chakra `Center` component. https://chakra-ui.com/docs/components/center/usage

src/components/Layer2ProductCard.tsx
```diff
@@ -110,21 +41,61 @@ const Layer2ProductCard: React.FC<IProps> = ({
...
+          maxW={{ base: "311px", sm: "372px" }}
+          maxH={"257px"}
+        />
+      </Flex>
+      <Flex
+        p={6}
+        textAlign="left"
+        h="100%"
+        direction="column"
+        className="hover"
```
pettinarip, line 79: Does this do something? if not then remove it

src/components/Layer2ProductCard.tsx
```diff
@@ -110,21 +41,61 @@ const Layer2ProductCard: React.FC<IProps> = ({
...
+      >
+        <Box>
+          <Heading as="h3" mt="0" mb={3}>
```
pettinarip, line 82: image

this heading seems too big comparing it with the prod version

pettinarip (changes requested):

src/components/Layer2ProductCard.tsx
ImMiguelP: Hi @pabloped i removed the `textAlign` as it does nothing but kept the `h` and `direction` bc without it the styling with the links gets messed up. Here's a photo for reference 

image

src/components/Layer2ProductCard.tsx
ImMiguelP: So it seems chakra sizing for` h3 "lg"` is bigger than the sizing y'all are using. I messed around with the font size and used "24px" but when the screen size resolution goes to phone the font is smaller than what is being used in the prod version. The closest I got to it looking to the prod version is using `size="md"`.

pettinarip: Oh ok, good point. Lets use `3xl` instead of `md` that is slightly smaller than the one we were using here (which was `2rem`).

src/components/Layer2ProductCard.tsx
```diff
@@ -110,21 +41,52 @@ const Layer2ProductCard: React.FC<IProps> = ({
...
+          alt={alt}
+          objectFit="contain"
+          alignSelf="center"
+          maxW={{ base: "311px", sm: "372px" }}
+          maxH={"257px"}
+        />
+      </Center>
+      <Flex p={6} h="100%" direction="column">
+        <Box>
+          <Heading as="h3" size="md" mt={0} mb={3}>
```
pettinarip, line 73:
```suggestion
          <Heading as="h3" size="3xl" mt={0} mb={3}>
```

pettinarip (changes requested):

src/components/Layer2ProductCard.tsx
ImMiguelP: Apologies, you are indeed correct its not needed here

src/components/Layer2ProductCard.tsx
ImMiguelP: Hi are we sure we want to use `3xl` here? This will just make the heading even more bigger than what it originally had (`lg`)? Is so I will happily do so just making sure this is what you going for based on styling.

Here is a screenshot to showcase the heading with `3xl` instead of `md`
image

pettinarip: So, on the preview deploy I see this:
image

and this is production:
image

Production is using 24px and the preview 20px.
24px = 1.5rem = 2xl in the chakra scale

src/components/Layer2ProductCard.tsx
```diff
@@ -110,21 +41,52 @@ const Layer2ProductCard: React.FC<IProps> = ({
...
+          alt={alt}
+          objectFit="contain"
+          alignSelf="center"
+          maxW={{ base: "311px", sm: "372px" }}
+          maxH={"257px"}
+        />
+      </Center>
+      <Flex p={6} h="100%" direction="column">
+        <Box>
+          <Heading as="h3" size="2xl" mt={0} mb={3}>
```
pettinarip, line 73:
```suggestion
          <Heading as="h3" fontSize={{ base: "xl", m: "2xl" }} mt={0} mb={3}>
```

pettinarip (approved):
Thanks @ImMiguelP 💪🏼

PR #8040, opened 2022-09-27
update ropsten to goerli

millecodex:


jmcook1186 (changes requested):
Thanks, but this doesn't really make sense in isolation because the tutorial text and the other images are also for Ropsten. Could you also update the other resources?

millecodex:
Yes, that makes sense. I will close this and open a new PR when ready.

PR #8043, opened 2022-09-27
Migrate MeetupList to Chakra

ImMiguelP:
## Description
Migrates `MeetupList` component from `emotion` to `chakra-ui`

## Related Issue
#6374


src/components/MeetupList.tsx
```diff
@@ -1,7 +1,7 @@
 // Libraries
 import React, { useState } from "react"
-import styled from "@emotion/styled"
 import { sortBy } from "lodash"
+import { Box, Flex, Input, Text } from "@chakra-ui/react"
 
 // Components
 import Emoji from "./OldEmoji"
```
pettinarip, line 7: This one is deprecated. Lets use Emoji. https://github.com/ethereum/ethereum-org-website/blob/dev/docs/chakra-migration-guide.md#update-dependencies

src/components/MeetupList.tsx
```diff
@@ -115,24 +47,63 @@ const MeetupList: React.FC<IProps> = () => {
...
+        borderColor="searchBorder"
+        color="text"
+        p={2}
+        borderRadius={1}
+        w="100%"
+        _focus={{ outline: "primary solid 1px" }}
+        _placeholder={{ color: "text200" }}
       />
-      <Table>
+      <Box boxShadow="tableBoxShadow">
```
pettinarip, line 67: As mentioned in the prev PR, this `tableBoxShadow` shadow doesn't exist. You will need to create it in https://github.com/ethereum/ethereum-org-website/blob/dev/src/%40chakra-ui/gatsby-plugin/foundations/shadows.ts

src/components/MeetupList.tsx
```diff
@@ -115,24 +47,63 @@ const MeetupList: React.FC<IProps> = () => {
...
         {filteredMeetups.length ? (
           filteredMeetups.map((meetup, idx) => (
-            <Item key={idx} to={meetup.link}>
-              <LeftContainer>
-                <ItemNumber>{idx + 1}</ItemNumber>
-                <ItemTitle>{meetup.title}</ItemTitle>
-              </LeftContainer>
-              <RightContainer>
+            <Link
```
pettinarip, line 70: Instead of wrapping this component with a Link, lets use the LinkOverlay component that Chakra offers for this specific scenario https://chakra-ui.com/docs/components/link-overlay/usage

I did a similar implementation in #7721

src/components/MeetupList.tsx
```diff
@@ -115,24 +47,63 @@ const MeetupList: React.FC<IProps> = () => {
...
+              display="flex"
+              justifyContent="space-between"
+              color="text"
+              boxShadow="0 1px 1px tableItemBoxShadow"
+              mb={0.25}
+              p={4}
+              w="100%"
+              _hover={{
+                textDecoration: "none",
+                borderRadius: 1,
```
pettinarip, line 83:
```suggestion
                borderRadius: "base",
```

src/components/MeetupList.tsx
```diff
@@ -115,24 +47,63 @@ const MeetupList: React.FC<IProps> = () => {
...
+            >
+              <Flex flex="1 1 75%" mr={4}>
+                <Box mr={4} opacity="0.4">
+                  {idx + 1}
+                </Box>
+                <Box>{meetup.title}</Box>
+              </Flex>
+              <Flex
+                textAlign={"right"}
+                alignContent={"flex-start"}
```
pettinarip, line 96:
```suggestion
                textAlign="right"
                alignContent="flex-start"
```

pettinarip (changes requested):
Thanks for the PR @ImMiguelP. I've noticed a few differences with the prod styles.

Check this page where we are using this component: https://ethereum.org/en/community/events/#meetups

This PR:
image

Expected (prod):
image

src/components/MeetupList.tsx
```diff
@@ -110,29 +50,78 @@ export interface IProps {}
...
-                <ItemNumber>{idx + 1}</ItemNumber>
-                <ItemTitle>{meetup.title}</ItemTitle>
-              </LeftContainer>
-              <RightContainer>
-                <Emoji text={meetup.emoji} size={1} mr={`0.5em`} />
-                <ItemDesc>{meetup.location}</ItemDesc>
-              </RightContainer>
-            </Item>
+            <LinkBox key={idx}>
+              <LinkOverlay
```
pettinarip, line 83: So, the idea of this `LinkOverlay` is to wrap only the text you want to behave as the link text. In this case, it would be the `{meetup.title}`.

`LinkBox` => should have all the layout styles
`LinkOverlay` => is the `<a>` tag that should only wrap the text

pettinarip (changes requested):
Great job so far. I would like to fix the implementation of `LinkOverlay`.

src/components/MeetupList.tsx
ImMiguelP: Ok that's understandable. S0 even though if I wrap the text with `LinkOverlay` the `LinkBox`  will make the other Box  clickable correct? In a sense its kind of like wrapping all the components in a `Form`?

pettinarip: Yes exactly, it is useful when you need to create this clickable components (like Cards) which are more than just a text link, and you still want to keep the html semantics in a good shape.

pettinarip (approved):

PR #8060, opened 2022-09-27
Remove preMergeBanner, update postMergeBannerTranslation

wackerow:
## Description
Removes `preMergeBanner` front matter variable from markdown and type file. No component is being rendered for this flag, and translated content that is OOD should already show a banner. 

Also adds postMergeBannerTranslation type to Frontmatter as string, and removes this from the `docs` template as it is not utilized in the dev docs. 

## Related Issue
#7075


corwintines (approved):
Thanks!

PR #8061, opened 2022-09-28
Migrate MarkdownTable to Chakra

ImMiguelP:
## Description
Migrates `MarkdownTable` component from `emotion` to `chakra-ui`

## Related Issue
#6374

### DIFF
--- a/src/components/MarkdownTable.tsx
+++ b/src/components/MarkdownTable.tsx
@@ -1,22 +1,24 @@
 import React from "react"
-import { Table, TableContainer, Th } from "@chakra-ui/react"
+import { Box, Table } from "@chakra-ui/react"
 
 export interface IProps {
   children?: React.ReactNode
 }
 
 const MarkdownTable: React.FC<IProps> = ({ children }) => (
-  <TableContainer my={8}>
-    <Table>
-      <Th
-        borderBottom={"1px solid"}
-        borderColor={"border"}
-        whiteSpace={"nowrap"}
-      >
-        {children}
-      </Th>
+  <Box my={8}>
+    <Table
+      sx={{
+        th: {
+          borderBottom: "1px solid",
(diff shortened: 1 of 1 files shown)


src/components/MarkdownTable.tsx
```diff
@@ -1,23 +1,21 @@
...
 
 const MarkdownTable: React.FC<IProps> = ({ children }) => (
-  <TableContainer>
-    <table>{children}</table>
+  <TableContainer my={8}>
+    <Table>
+      <Th
+        borderBottom={"1px solid"}
+        borderColor={"border"}
+        whiteSpace={"nowrap"}
```
pettinarip, line 14:
```suggestion
        borderBottom="1px solid"
        borderColor="border"
        whiteSpace="nowrap"
```

pettinarip (changes requested):
Hey @ImMiguelP look at this pic
image

This is the nft page, https://ethereumorgwebsitedev01-immiguelpethereumorgwebsi26645.gtsb.io/en/nft

It seems that the table is wider than the container.

src/components/MarkdownTable.tsx
ImMiguelP: Hey @pabloped thank you for this I resolved this issue. Turns out this issue happened when I used the `TableContainer` component so replaced it with `Box` and since the children is being rendered within the `Table` I decided to add the `sx{} `option to it to style the `Th` this also fixed the issue with the `borderBottom` being under the whole table

wackerow (approved):
@ImMiguelP Nicely done! Yeah, using the Box without the Th looks like the approach to go here; seems to fix it:
image
Thanks!

pettinarip (approved):
Awesome @ImMiguelP LGTM! 💪🏼

PR #8069, opened 2022-09-28
Post-merge cleanup

wackerow:
## Description
- Removes merge-related announcement banners
- Reverts homepage hero image, renaming panda version to `hero-panda.png`
- Removes useConfetti hook and `react-rewards` package
- Adds isolated panda art, separated from the existing hero image, both png and svg forms
- Adds new assets to the /assets page

## Related Issue
#7075


src/components/Layout.tsx
```diff
@@ -180,7 +165,7 @@ const Layout: React.FC<IProps> = ({
                 <Nav path={path} />
                 {shouldShowSideNav && <SideNavMobile path={path} />}
               </VisuallyHidden>
-              <BannerNotification
+              {/* <BannerNotification
```
samajammin, line 168: I suspect we should remove the `"merge-complete"` string from common.json?

samajammin (changes requested):
Nice - LGTM other than removing that 1 string!

src/components/Layout.tsx
wackerow: Oops, good catch. Removing string and comment now.

samajammin (approved):

PR #8073, opened 2022-09-29
Replace codeblocks with tables on blocks page

wackerow:
## Description
Breakdown of fields on the block page using codeblocks causing them to be a bit difficult to read.
- Refactors lists contained within codeblocks into markdown tables. 

Currently:
image

Suggested in PR:
image

Not yet in this PR, but if we adjust the styling in `MarkdownTable.tsx` we can add a little better padding between these columns. The following css would add 0.5rem spacing:
```scss
  th:not(:last-of-type), td:not(:last-of-type) {
    padding-inline-end: 0.5rem;
  }
```
...would result in...
image

@nloureiro Without objection I could add this in. This component will still need to be migrated to Chakra, but shouldn't prevent us from fixing the tight k

## Related Issue
None filed


nloureiro:
> ## Description
> Breakdown of fields on the block page using codeblocks causing them to be a bit difficult to read.
> 
> * Refactors lists contained within codeblocks into markdown tables.
> 
> Currently: image
> 
> Suggested in PR: image
> 
> Not yet in this PR, but if we adjust the styling in `MarkdownTable.tsx` we can add a little better padding between these columns. The following css would add 0.5rem spacing:
> 
> ```css-scss
>   th:not(:last-of-type), td:not(:last-of-type) {
>     padding-inline-end: 0.5rem;
>   }
> ```
> 
> ...would result in... image
> 
> @nloureiro Without objection I could add this in. This component will still need to be migrated to Chakra, but shouldn't prevent us from fixing the tight k
> 
> ## Related Issue
> None filed

I'm ok with this. I will need to re-visit this on the design system side. for now, go for it

wackerow:
Actually, noting that #8061 fixes this issue by migrating the `MarkdownTable.tsx` component to Chakra (which has a small amount of spacing between the cells). Going to leave it as is here, should resolve on its own.

corwintines:
Merged `dev` in since https://github.com/ethereum/ethereum-org-website/pull/8061 has been merged in.

corwintines (approved):
Nice 🚀

PR #8078, opened 2022-09-29
Add CryptoJobster as a resource for Ethereum jobs

fr33dr4g0n:
Added link to Ethereum specific page of jobs at https://cryptojobster.com/tag/ethereum/ that currently has 98 Ethereum based jobs listed.


minimalsm (commented):
Hey @fr33dr4g0n, thanks for the suggestion 👍  

As part of [our listing criteria for new products](https://ethereum.org/en/contributing/adding-products/), we state:

> _A product that has been "live" for over 6 months – this is another indication of security. 6 months is a good time frame for critical bugs and exploitations to have been found._

[Looks like Cryptojobster was only created this month](https://www.whois.com/whois/cryptojobster.com), so it wouldn't meet fit this criteria. It also looks to be just scraping the same data that exists on all other job boards, so I'm not convinced it would add anything novel or useful for our users. Going to close this out but feel free to continue the conversation here.

minimalsm:
@all-contributors please add @fr33dr4g0n for ideas

Issue #8050, opened 2022-09-27
ReleaseBanner.tsx @pettinarip


pettinarip:

lawlesx:
I would like to work on this

pettinarip:
Thanks @lawlesx!

PR #8081, opened 2022-09-29
Migrate release banner component to chakra UI

lawlesx:
## Description
Migrated `Release Banner` component to Chakra UI

As mentioned in the **Chakra migration guide** I used icons from chakra.
So I had to install chakra icons using
```
yarn add @chakra-ui/icons
```

Created a local component `BannerWrapper` which includes common components for both `completed` and `not complete` state. (Just to make it reusable)

### Attaching screen shots for the changes in the UI
Dark mode
image
Light mode
image
Responsive
image

**Note**: I tested it by creating a new page `src\pages\test.tsx` and deleted the page after testing

## Related Issue

Resolves issue #8050

### DIFF
--- a/src/components/ReleaseBanner.tsx
+++ b/src/components/ReleaseBanner.tsx
@@ -1,7 +1,6 @@
 // Libraries
 import React, { useEffect, useState } from "react"
 import Countdown, { zeroPad } from "react-countdown"
-import styled from "@emotion/styled"
 
 // Components
 import BannerNotification from "./BannerNotification"
@@ -39,7 +38,7 @@ const BannerWrapper: React.FC<BannerWrapperProps> = ({
 }) => {
   return (
     <BannerNotification shouldShow={isBannerVisible} position="relative">
-      <Center position="relative" paddingRight="2.5rem" flexWrap="wrap">
+      <Center position="relative" paddingRight="2rem" flexWrap="wrap">
         <Emoji
           marginRight="1rem"
           flexShrink={0}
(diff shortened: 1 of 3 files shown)


src/components/ReleaseBanner.tsx
```diff
@@ -57,6 +25,42 @@ interface CountdownRendererProps {
...
+          flexShrink={0}
+          text=":tada:"
+          w="1rem"
+          h="1rem"
+        />
+        {children}
+      </Center>
+      <IconButton
+        aria-label="Close Banner"
+        icon={<CloseIcon />}
```
pettinarip, line 53: We should not use the Chakra icons, so lets revert that part. What the migration guide says is to use the Chakra `Icon` component (or in this case the `IconButton` component).

We should still use the same svgs we were using before. In this case we were using `MdClose` which you can import it by doing `import { MdClose } from "react-icons/md"`

And finally, you can do `<IconButton icon={<MdClose />}`

pettinarip (changes requested):
@lawlesx nice work on this. I think we need to change the way we are using the icon but the rest looks solid.

src/components/ReleaseBanner.tsx
lawlesx: Got it
Removed `@chakra-ui/icons` and used `<MdClose />`

pettinarip (approved):
Looking good @lawlesx Will push some minor changes & the implementation of `CloseButton` instead of using `IconButton`.

pettinarip:
@all-contributors please add @lawlesx for code

Issue #7559, opened 2022-08-24
Quiz content: GAS AND FEES page


konopkja:
This is a sub-task of: https://github.com/ethereum/ethereum-org-website/issues/7461

Prepare content for quiz on page:
https://ethereum.org/en/developers/docs/gas/#top

Each question should be related to the most important aspects of the page. There can be a unique explanation for each correct/incorrect answer.

qbzzt:
What is the purpose of gas?

**A. Measure the cost of the computation and storage a transaction requires**
B. Measure the cost of the energy required to process a block, only required on PoW (proof of work) 
C. Measure the cost of a transaction in a fiat currency, such as USD
D. Measure the cost of a block in a fiat currency, such as USD


**Explanations for incorrect answers:**

B. Ethereum gas has nothing to do with physical energy.
C. & D. Gas is purchased for ETH.

qbzzt:
What unit is typically used to measure the cost of gas?

A. wei, the smallest unit of ETH (10<sup>-18</sup> ETH)
B. wei, the smallest unit of ETH (10<sup>-9</sup> ETH)
C. **gwei, 10<sup>-9</sup> ETH**
D. ETH

**Explanations for incorrect answers:**

A. Gas typically costs at least 10,000,000,000 wei, so while we could use wei, it is inconvenient
B. Wrong for two reasons: 1. Gas typically costs at least 10,000,000,000 wei, so while we could use wei, it is inconvenient and 2. wei is actually 10<sup>-18</sup> ETH.
D. Gas typically costs around 0.000,000,01 ETH, so while we could use ETH, it is inconvenient

qbzzt:
What causes the base fee to increase?

A. Low demand for blockspace, which results in blocks smaller than the target size (15M gas)
B. Low demand for blockspace, which results in blocks larger than the target size (15M gas)
C. High demand for blockspace, which results in blocks smaller than the target size (15M gas)
**D. High demand for blockspace, which results in blocks larger than the target size (15M gas)**

**Explanations for incorrect answers:**

A. If there is low demand for blockspace we try to increase it by *reducing* the base fee to make transactions cheaper.
B. If there is low demand for blockspace, why would blocks be bigger than the target size? Blocks are large where there is *high* demand.
[...]

Issue #7558, opened 2022-08-24
Quiz content: Layer 2 page


konopkja:
This is a sub-task of: https://github.com/ethereum/ethereum-org-website/issues/7461

Prepare content for quiz on page:
https://ethereum.org/en/layer-2/

Each question should be related to the most important aspects of the page. There can be a unique explanation for each correct/incorrect answer.

qbzzt:
What service(s) does Ethereum (Layer 1) give to Layer 2 solutions?

A. Reliable execution
B. Internal storage
**C. Data integrity and data availability**
D. Data integration, data availability, and internal storage.

qbzzt:
What is the major feature provided by Layer 2 solutions that Ethereum, by itself, cannot provide?

A. Security
B. Decentralization
**C. Scalability**
D. Uptime (reliability)

**Explanation:**

Ethereum optimizes for security and decentralization at the base layer, and decentralization gives us uptime automatically. A decentralized system with thousands of nodes will have a large number of nodes running at any point in time. Layer 2 solutions typically provide scalability by relying on Ethereum for security, and being less decentralized (but relying on mechanisms running on Ethereum to ensure the execution and internal storage are correct).

qbzzt:
Which type of layer 2 solution relies on economic incentives to ensure transactions are executed correctly?

A. Zero-knowledge rollups
**B. Optimistic rollups**
C. Sidechains
D. Validiums

**Explanation:**

Optimistic rollups have sequencers that post the transactions' data and their results (in state root form) on Ethereum, and verifiers that challenge them. Sequencers that post incorrect results, and verifiers that dispute correct results, pay a penalty. Sequencers that post correct results, and verifiers that dispute incorrect results, get rewarded. This is using economic incentives to encourage honesty.

qbzzt:
Which type of layer 2 solution relies on mathematics to ensure transactions are executed correctly?

**A. Zero-knowledge rollups**
B. Optimistic rollups
C. Sidechains
D. Validiums

**Explanation:**

Zero-knowledge rollups post zero-knowledge proofs that they processed their transactions correctly. Producing those proofs requires a lot of computation, and is not feasible to do on chain. However, to verify the proof only requires a relatively simple computation that can be done on-chain easily.

minimalsm:
[...]

Issue #7461, opened 2022-08-16
Epic: learning quizzes 💯


minimalsm:
This epic is part of our Q3 roadmap #7166 & Q4 2022 roadmap: #8399

## Description

A lot of important about Ethereum is complex and can be difficult to grasp. A great way to ensure important content is understood is by immediately testing knowledge after it is presented. The learning quizzes epic will introduce optional quizzes to [ethereum.org](http://ethereum.org/)
 content. The aim of the quizzes is to ensure that readers grasp the essential concepts to allow them to be meaningful participants in the Ethereum ecosystem.

## Deliverables

[As part of our Q3 roadmap](https://github.com/ethereum/ethereum-org-website/issues/7166), we'd like to achieve the following: 

- [x] https://github.com/ethereum/ethereum-org-website/issues/7552
- [x] https://github.com/ethereum/ethereum-org-website/issues/7553
- [x] https://github.com/ethereum/ethereum-org-website/issues/7554
- [x] https://github.com/ethereum/ethereum-org-website/issues/7555
- [x] https://github.com/ethereum/ethereum-org-website/issues/7556
- [x] https://github.com/ethereum/ethereum-org-website/issues/7557
- [x] https://github.com/ethereum/ethereum-org-website/issues/7558
- [ ] https://github.com/ethereum/ethereum-org-website/issues/7559

- To be completed

qbzzt:
Do you already have software you're going to use? If so, which one? I've written plenty of quizzes in my life, so I can contribute here.

konopkja:
We are building our custom solutions: preliminary design can be seen on this prototype (bottom of the page) https://www.figma.com/proto/1HxSnapMdpgFFhIHjDUytP/Learning-quizzes?page-id=266%3A189&node-id=266%3A791&viewport=-2281%2C-1848%2C0.47&scaling=min-zoom&starting-point-node-id=266%3A791

qbzzt:
Cool, what format would you accept as input? I'd like to get started
writing quizzes.

konopkja:
> Cool, what format would you accept as input? I'd like to get started writing quizzes.

 Created sub-tasks for individual pages we identified for the v1.  If there are any other pages you would love to write quiz content for let me know.

hursittarcan:
Maybe it's also idea to award people that completed a specific quiz with a POAP?

qbzzt:
[...]

PR #8094, opened 2022-09-29
[WIP] Learn quizzes

corwintines:


src/@chakra-ui/gatsby-plugin/components/Button.ts
```diff
@@ -26,6 +26,28 @@ const commonOutline = {
   },
 }
 
+const quizButton = {
```
pettinarip, line 29: For me, this is a blocker. Calling @nloureiro or @konopkja to jump into the discussion.

Have we decided on creating a new button style just for this case? Can't we reuse a button style from the already existing styles? https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System

The problem I see with this new style is that we are falling into the problems we had before. We are creating individual styles for each implementation we are doing and we end up having a bunch of different button styles across the site. IMO we should try to use the styles that the Design System tells us.

Regardless of that, if we decide to implement this new style, I would:
1. Update the design system showing this new style
2. Change the name of this variant to something that is abstracted from the implementation. This maybe is our `ghost` variant
3. Do not use the suffix `Button` on the name as it is redundant

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -38,6 +38,7 @@ const semanticTokens = {
     bodyLight: { _light: "gray.500", _dark: "gray.100" },
     disabled: { _light: "gray.400", _dark: "gray.500" },
     background: { _light: "white", _dark: "gray.700" },
+    quizButton: { _light: "gray.100", _dark: "gray.700" },
```
pettinarip, line 41: Naming. The name of the color should not be coupled with the implementation name. Should we add this new color to the Design System?

cc @nloureiro

pettinarip (changes requested):

src/components/Quiz/QuizQuestion.tsx
```diff
@@ -0,0 +1,104 @@
...
+        {prompt}
+      </Text>
+      {answers.map(({ id, label }, index) => {
+        const active = selectedAnswer === id
+        const iconBackgroundDark = active ? "orange.800" : "gray.500"
+        const iconBackgroundLight = active ? "blue.300" : "gray.400"
+        const display =
+          !showAnswer || id === selectedAnswer ? "inline-flex" : "none"
+        return (
+          <Button
```
pettinarip, line 53: Shouldn't these be radio buttons?

Check this implementation example https://chakra-ui.com/docs/components/radio/usage#custom-radio-buttons

src/intl/en/quizzes.json
```diff
@@ -0,0 +1,3 @@
+{
+  "quiz-test-your-knowledge": "Test your knowledge"
```
samajammin, line 2: Is this just a placeholder? I'm guessing this PR doesn't include internationalization support to start?

src/data/learnQuizzes/index.ts
```diff
@@ -0,0 +1,32 @@
+// Import data types
+import { RawQuizzes } from "../../types"
+
+// Declare hash-map of quizzes based on slug key
+const quizzes: RawQuizzes = {
+  "what-is-ethereum": {
+    title: "What is Ethereum?",
+    questions: ["a001", "a002", "a003", "a004", "a005"],
```
samajammin, line 8: Why are these needed? Should we make these more intelligible to humans?

src/data/learnQuizzes/questionBank.ts
```diff
@@ -0,0 +1,754 @@
...
+
+// Declare hash map of question bank
+const questionBank: QuestionBank = {
+  // TODO: Use hash of English prompt to generate this key
+  // What is Ethereum?
+  a001: {
+    prompt: "The biggest difference between Ethereum and Bitcoin is:",
+    answers: [
+      {
+        id: "a001-a", // TODO: Use hash of English question label to generate this id
```
samajammin, line 12: Why does each answer need an ID - can we not rely on their index?

src/data/learnQuizzes/questionBank.ts
```diff
@@ -0,0 +1,754 @@
+// Import data types
```
samajammin, line 1: Let's use kebab-case for directory names please (i.e. `data/learn-quizzes`). I'd even suggest we just go with `data/quizzes`).

samajammin:
@wackerow @konopkja @minimalsm - I'm seeing 8 pages listed in #7461 but 6 pages in this PR (at least 6 page links in the PR description) - are we planning to add quiz content for L2 page (https://github.com/ethereum/ethereum-org-website/issues/7558) & gas page (https://github.com/ethereum/ethereum-org-website/issues/7559) in this PR?

wackerow:
@samajammin Our initial use testing focused on the six you see, but I'm assuming we'll want to bring in the others as well for launch... @minimalsm are the other two ready / think these will be in a finalized state soon? Let me know and I can pull them in 👍

src/data/learnQuizzes/questionBank.ts
wackerow: Agreed, can update

src/data/learnQuizzes/questionBank.ts
wackerow: We could potentially, but went with this approach to help prevent any issues is the ordering ever changes (ie, if we decide to shuffle answer choices in the future).

src/data/learnQuizzes/index.ts
wackerow: The benefit here (separating individual questions from quizzes) is so we have the option to use any particular question in multiple quizzes if desired. ie, we currently reuse the "Who should you give your seed phrase / private keys to?" question in both the "wallets" and "security" quizzes.

Using an ID also helps reduce issues with tracking if we ever need to adjust the copy of a question. I had originally left a TODO to auto-generate an ID using a hash of the prompt (or something along those lines), but for the reason just mentioned I think it may be best to just add a manual ID. The number of questions should be within a reasonable amount where this isn't going to add significant overhead imo.

src/intl/en/quizzes.json
wackerow: Looks like this was teed up during the initial draft, presumedly as a placeholder to prep for future internationalization. Happy to just revert this for now since it's the only string set up this way.

src/components/Quiz/QuizRadioGroup.tsx
```diff
@@ -0,0 +1,164 @@
...
+        </Flex>
+      </chakra.label>
+    )
+  }
+
+  const handleChange = (value: string): void => {
+    handleSelection(value)
+  }
+
+  const { getRadioProps, getRootProps } = useRadioGroup({
```
pettinarip, line 117: By convention and to avoid potential issues, its recommended to call hooks at the top level of the function. https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

src/components/Quiz/QuizRadioGroup.tsx
```diff
@@ -0,0 +1,164 @@
...
+            Explanation
+          </Text>
+          <Text m={0}>{explanation}</Text>
+        </Box>
+      )}
+    </Flex>
+  )
+}
+
+export default QuizRadioGroup
```
pettinarip, line 164: Awesome. Wondering now if we could cleanup and remove the `quizButton` variant from the `Button` theme.

src/components/Quiz/QuizSummary.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const percentCorrect = useMemo<number>(
+    () => correctCount / questionCount,
+    [correctCount, questionCount]
+  )
+  const options = {
+    style: "percent",
+    maximumFractionDigits: 0,
+  }
+  const numberToPercent = (num: number): string =>
+    new Intl.NumberFormat(locale, options).format(num)
```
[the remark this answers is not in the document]
wackerow: It looks like that function restricts to using either "fa" or "en"... Is there a reason we wouldn't just want to use the Intl library with whatever the users `locale` is, and let it render the percentage accordingly?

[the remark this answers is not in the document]
pettinarip: It only overwrites the `fa` locale with `en`. The rest of the locales are going to be kept the same.

I would use it to keep it consistent with the rest of the site (that is why having `numberToPercent` in `/utils` & shared it across the site would be helpful).

There is a bit of context on the function comment but wondering what @minimalsm thinks about this. If it is necessary to be used here.

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -34,12 +34,17 @@ const semanticTokens = {
...
+    bodyInverted: { _light: "gray.100", _dark: "gray.700" },
     bodyLight: { _light: "gray.500", _dark: "gray.100" },
     disabled: { _light: "gray.400", _dark: "gray.500" },
     background: { _light: "white", _dark: "gray.700" },
-    success: "green.500",
+    neutral: { _light: "white", _dark: "gray.900" },
+    success: { _light: "green.500", _dark: "green.400" },
+    successNeutral: { _light: "green.100", _dark: "green.900" },
     error: "red.500",
+    errorNeutral: { _light: "red.100", _dark: "red.900" },
```
wackerow, line 47: cc: @nloureiro @konopkja
Take a look and let me know if you disagree with anything here, or if you have suggestions on any naming improvements. If we keep these we may want to update the DS with the ones that are missing.

pettinarip:
@wackerow FYI we have merged #8301 so you should have the correct `disabled` style on the buttons if you update the branch.

nloureiro:
Found a little bug in mobile

the space in the results card is uneven 

Screen Shot 2022-10-26 04 17 59 PM

src/components/Quiz/QuizSummary.tsx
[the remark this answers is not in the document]
minimalsm: @wackerow @pettinarip 

The reason we fall back to English for Farsi is that the Persian numbering causes UX inconsistencies. Some user interviews we did indicated that universal numbering (0-9) is fine for Farsi speakers (even non-English speakers), but mixing them was confusing.

Screenshot 2022-10-26 at 20 13 50

minimalsm:
Small nice to have for future iterations: minimum height? (or something to combat layout shift?)

Screenshot 2022-10-26 at 20 10 57

minimalsm:
Screenshot 2022-10-26 at 20 18 48

@konopkja @jmcook1186 thoughts on the button copy here?

'Take quiz again' seems like unusual phrasing to me. What about just 'Try again'?

src/components/Quiz/QuizSummary.tsx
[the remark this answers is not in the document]
wackerow: My bad, misread the function code... will fix =)

src/components/Quiz/QuizWidget.tsx
```diff
@@ -0,0 +1,389 @@
...
+                  )}
+                {showResults ? (
+                  <Flex gap={6}>
+                    <Button
+                      leftIcon={<Icon as={FaTwitter} />}
+                      onClick={handleShare}
+                    >
+                      Share results
+                    </Button>
+                    <Button onClick={initialize}>Take quiz again</Button>
```
corwintines, line 355:
```suggestion
                    <Button onClick={initialize}>Try again</Button>
```

wackerow:
Not against changing the button copy to "Try again", but if we do, I think we should hide this in the event the user scores 100%. 

@nloureiro I can try adjusting the horizontal padding and the font size on that summary screen to try and avoid overflow, but we may want to consider stacking these number on mobile as well. Or, we (cc: @konopkja) were a little on the fence with displaying that third column (the "Total" column)... maybe we just remove this as a patch and address more responsive styling separately?

Another thing to note is the overflow of the buttons on mobile, as they are not wrapping. A common pattern we've used for pairs of buttons on mobile is to stack them instead, especially given variability of text length with translations.

wackerow:
Have some fixes/potential fixes locally and will push them in a moment. Changes include
- Stacking/stretched buttons on mobile
- "Try again" copy (which conditionally is hidden is the user got 100%)
- Hides the "Total" column on mobile sizes
- Shrinks the fontSize slightly on mobile to "xl" from "2xl"

I'm finding these changes add some nice polishing touches that improve mobile responsiveness, let me know what you think.

wackerow:
Also, @minimalsm, agree regarding some type of height-handling... This does have an awful lot of content shifting, ie, when the quiz loads, when you submit a question, and when you go to the next question... they all produce shifting to the content beneath the widget. Curious how @nloureiro and @konopkja would like to handle this since we generally will not know the height of the largest content ahead of time.

src/components/Quiz/QuizWidget.tsx
```diff
@@ -0,0 +1,410 @@
...
+                  <>
+                    <Button
+                      leftIcon={<Icon as={FaTwitter} />}
+                      onClick={handleShare}
+                    >
+                      Share results
+                    </Button>
+                    {score < 100 && (
+                      <Button onClick={initialize}>Try again</Button>
+                    )}
```
wackerow, line 376: @minimalsm If we decide we don't want to hide this button with 100%, just commit this code suggestion:
```suggestion
                    <Button onClick={initialize}>Try again</Button>
```

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
```diff
@@ -22,10 +22,15 @@ const colors = {
     800: "#451900",
   },
   red: {
+    100: "#F7C8C8",
     500: "#b80000",
+    900: "#1B0C0C",
   },
   green: {
+    100: "#C8F7D8",
+    400: "#48BB78",
```
corwintines, line 31: I know this is used for success in dark mode. but this breaks the design system currently. @nloureiro @konopkja just calling out that this should get added in design system colors as it seems like the direction we are going with designs.

corwintines: https://github.com/ethereum/ethereum-org-website/pull/8094#discussion_r1005036237

Duplicate comment, sorry. What @wackerow said.

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
```diff
@@ -22,10 +22,15 @@ const colors = {
     800: "#451900",
   },
   red: {
+    100: "#F7C8C8",
     500: "#b80000",
+    900: "#1B0C0C",
```
corwintines, line 27: I do find this red hard to see in dark mode

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
```diff
@@ -22,10 +22,15 @@ const colors = {
...
   },
   green: {
+    100: "#C8F7D8",
+    400: "#48BB78",
     500: "#109e62",
+    900: "#0A160E",
```
corwintines, line 33: I do think this green is hard to see in dark mode

src/components/Quiz/QuizWidget.tsx
```diff
@@ -0,0 +1,411 @@
...
+                    h="4px"
+                    w={width}
+                    maxW={`min(${width}, 2rem)`}
+                    marginInline={0}
+                    p={0}
+                  />
+                )
+              })}
+            </Center>
+            <Center>
```
corwintines, line 332:
```suggestion
            <Center maxW={'29.75rem'} m='auto'>
```

This brings it in line with figma style

konopkja:
> Screenshot 2022-10-26 at 20 18 48
> 
> @konopkja @jmcook1186 thoughts on the button copy here?
> 
> 'Take quiz again' seems like unusual phrasing to me. What about just 'Try again'?

try again is better

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
nloureiro: I saw your comment on Figma. We need to do a revision on the quizzes colours after the launch.
For now it's good on my end

wackerow (approved):

src/utils/numberToPercent.ts
```diff
@@ -0,0 +1,11 @@
+import { getLocaleForNumberFormat } from "./translations"
+import type { Lang } from "./languages"
+
+export const numberToPercent = (
```
pettinarip, line 4: 🚀

pettinarip (approved):
Great job @wackerow @corwintines! approoobig

minimalsm (approved):

src/utils/numberToPercent.ts
minimalsm: shipppppp it

Issue #8095, opened 2022-09-29
Content Bug - Broken link in Backend Page in developers/docs/apis/backend/


nmoreau:
### Describe the bug

This page https://ethereum.org/en/developers/docs/apis/javascript/ has a broken link to navigate to https://ethereum.org/en/developers/docs/apis/json-rpc/endpoints/

### To reproduce

1. Go to page https://ethereum.org/en/developers/docs/apis/javascript/ 
2. Click on "endpoint link"
3. Actual : 404
4. Expected : you should navigate to https://ethereum.org/en/developers/docs/apis/json-rpc/#json-rpc-methods

Also I fixed the use of "endpoint" for "methods" which is more technically correct.
Endpoint is more the http endpoint that the node provide will provide. That was confusing.
This wording also aligns with the language used in this page https://ethereum.org/en/developers/docs/apis/json-rpc/

### Expected behavior

Link is not broken

### Screenshots

<img width="818" alt="image" src="https://user-images.githubusercontent.com/5418582/193127706-7cf945aa-c05d-4dde-b5b4-bda04bcacce6.png">

### Additional context

The following PR is fixing the issue
https://github.com/ethereum/ethereum-org-website/pull/8100

nmoreau:
This issue is fixed in the following PR
https://github.com/ethereum/ethereum-org-website/pull/8100

PR #8100, opened 2022-09-29
Fixing broken link and language on /backend page

nmoreau:
Fixed the broken link to the JSON-RPC page in the backend page.

## Description
The link on the public website is broken. See : https://ethereum.org/en/developers/docs/apis/backend/ 
I also renamed endpoints to methods to align the language.
The GRPC endpoint is litteraly the URI where you can send you JSON-RPC requests. That previous language was confusing.
## Related Issue

This is a suggested addition tracked by Issue 
https://github.com/ethereum/ethereum-org-website/issues/8095


corwintines (approved):
Makes sense. I think this language is saying the same thing as a web dev, but given the [specs themselves](https://github.com/ethereum/execution-apis) are referred to as methods, should keep it consistent with those.

PR #8126, opened 2022-10-01
Card content aligned to the bottom [Fixes #8062]

### LINKED ISSUE #8062: align to the bottom the card content
### Describe the bug

On the developers page, the content of the cards should alight to the bottom.

### To reproduce

1. got to https://ethereum.org/en/developers/
2. scroll down to the "How would you like to get started?" section
3. the interior of the third card should be align to the bottom

### Expected behavior

Align the buttons to the bottom, and the interior text should start at the same point 

### Screenshots

This is how it is:
Screen Shot 2022-09-28 11 59 19 AM

This is how it should be:
Screen Shot 2022-09-28 01 58 25 PM

all OS

_No response_

### Additional context

_No response_

najeebkp:
Card content aligned to the bottom  [Fixes #8062]

## Description

There was an alignment issue with card content in /developers page.


src/pages/developers/index.tsx
```diff
@@ -176,6 +176,10 @@ const StyledCard = styled(Card)`
   box-shadow: ${(props) => props.theme.colors.tableBoxShadow};
   margin: 1rem;
   padding: 1.5rem;
+  justify-content: space-between;
+  h3 {
+    min-height: 67px !important;
+  }
```
minimalsm, line 182: Why do we need this 🤔?

minimalsm (commented):
Thanks for the PR @najeebkp, just left one question around the minimum height on the h3.

src/pages/developers/index.tsx
najeebkp: Its because, we are setting the minimum height for the h3 tag(card heading). 
as we are having one line and two line headings both are taking different heights. to equate these, we have set a minimum height as the height of a two line heading.

pettinarip: I don't understand what is the issue here and how this would fix it. Could you share a link or an image where we can see the problem?

src/pages/developers/index.tsx
```diff
@@ -176,6 +176,10 @@ const StyledCard = styled(Card)`
   box-shadow: ${(props) => props.theme.colors.tableBoxShadow};
   margin: 1rem;
   padding: 1.5rem;
+  justify-content: space-between;
```
pettinarip, line 179: This change is ok but we want that change to be applied to all the `Card`s that we are using across the site. This change has impact only on this `developers/index` page.

Could you apply this style to the `src/components/Card.tsx` component instead of doing it here?

pettinarip (changes requested):

src/pages/developers/index.tsx
najeebkp: ok

najeebkp: sssssss

Here the h3 means the title of card. 
if there is one line title -> the card title height is lets say 30px
if there us two line title -> the card title height will increase to lets say 67px.

so what im doing here is, setting the title height minimum 67px, so that even one line title will have the height of 67px. 
therefor the spacing will be even.

src/components/Card.tsx
```diff
@@ -10,6 +11,11 @@ export interface IProps {
   className?: string
 }
 
+const Container = styled.div`
+  h3 {
+    min-height: 67px !important;
```
pettinarip, line 16: We are deprecating the usage of `emotion` in favor of using Chakra components. So, lets do this in a Chakra way. Check my next comment.

src/components/Card.tsx
```diff
@@ -26,8 +32,9 @@ const Card: React.FC<IProps> = ({
     borderColor="lightBorder"
     p={6}
     className={className}
+    style={{ justifyContent: "space-between" }}
   >
-    <div>
+    <Container>
```
pettinarip, line 37: Lets use Chakra components to do this. No need for the `!important` here
```suggestion
    <Box minHeight="67px">
```

src/components/Card.tsx
```diff
@@ -26,8 +32,9 @@ const Card: React.FC<IProps> = ({
     borderColor="lightBorder"
     p={6}
     className={className}
+    style={{ justifyContent: "space-between" }}
```
pettinarip, line 35: `Flex` is a Chakra component. The correct way to express this would be:
```suggestion
    justify="space-between"
```
See more on https://chakra-ui.com/docs/components/flex/usage

pettinarip (changes requested):
Thanks for the explanation @najeebkp on the min height. I think that is fine for now.

corwintines:
Thanks, @najeebkp for the contribution. Apologies for this going stale.

There was a small fix to align content to the bottom here that got merged in: https://github.com/ethereum/ethereum-org-website/pull/8726. For now, I think we can close this PR given future plans for cards. Which I'll get into below.

As for long-term goals for the Card component. We are working on implementing a design system to unify components on ethereum.org. Would you be interested in getting involved with that?

Issue #8127, opened 2022-10-01
Eth R&D discord link invalid


ryankav:
### Describe the bug

The discord invite on https://ethereum.org/en/community/research/ appears to be no longer valid

### To reproduce

1) Try to use the discord invite to `eth R&D`: https://discord.gg/n7JxAeRu
2) see that invite has expired or is no longer valid

### Expected behavior

The invite allows you to join the channel

### Screenshots

![Screenshot from 2022-10-01 14-35-32](https://user-images.githubusercontent.com/39402882/193412167-a15ed01d-3756-4450-8a0e-cb96b4cefb66.png)

PR #8129, opened 2022-10-01
Fix: Eth R&D Discord channel link

geeknees:
## Description

fix: #8127 

I am not sure if this link is correct, but I found the link on the official blog and replaced it.
https://blog.ethereum.org/2022/03/23/finalized-no-34

I have confirmed that the link works.


minimalsm (approved):
Thanks @geeknees 👍

Issue #8031, opened 2022-09-26
Visual regression in markdown lists


minimalsm:
### Describe the bug

The lists are no longer displaying as intended.

- Spacing between list items is not correct

You can [see how they displayed before in this preview](https://ethereumorgwebsitedev01-networkupgradescomponent.gtsb.io/en/nft) versus [how they display now.](https://ethereum.org/en/nft/)

### To reproduce

1. Go to https://ethereum.org/en/nft/
2. Scroll down to [the NFT examples section](https://ethereum.org/en/nft/#nft-comparison)
3. See visual regression

### Expected behavior

The list should be correctly formatted ([as per the previous visual style](https://ethereumorgwebsitedev01-networkupgradescomponent.gtsb.io/en/nft)) with adequate space between list items.

### Screenshots

Current:

<img width="963" alt="Screenshot 2022-09-26 at 11 44 53" src="https://user-images.githubusercontent.com/62268199/192258002-59bccc2b-316e-48fd-97e4-5afe436ba17a.png">

Before (expected):

<img width="933" alt="Screenshot 2022-09-26 at 11 45 01" src="https://user-images.githubusercontent.com/62268199/192258009-49e140d1-c672-43c1-91b4-cb74b86a9999.png">

### Additional context

I suspect this happened as part of [the migration to Chakra.](https://github.com/ethereum/ethereum-org-website/issues/6374)

AidanPine:
Hello, I would like to take on the responsibility for this issue if possible.

Pro-Coder04:
@minimalsm could you help me to out to find the location where the source code for this file would be?

minimalsm:
@AidanPine @Pro-Coder04 

Hey folks. As noted [in this comment](https://github.com/ethereum/ethereum-org-website/issues/8030#issuecomment-1257998371), it looks like this was caused by us not pulling over some global styles in our migration to emotion/Chakra from styled-components.

@pettinarip are you working on this already, or is it something we can allocate?

NoireBlanc:
@minimalsm I've already done work moving over the global list item styles, and it seems to have fixed the visual regression issue. However, in the expected look of the webpage you've linked, the first list featured on the nft page does not have the same list item spacing as the rest of the page. Is this expected, or is a consistent spacing for all lists with no exceptions preferred?

pettinarip:
Thanks folks for the help with this. We are trying to fix it under #8139 by just using Chakra components.

PR #8139, opened 2022-10-02
Fix list & table styles

### LINKED ISSUE #8030: Visual regression in markdown tables
### Describe the bug

The markdown table has visually regressed and is no longer displaying as intended. 

- Lines that divide rows are missing
- Content isn't aligned correctly

You can [see how it displayed before in this preview](https://ethereumorgwebsitedev01-networkupgradescomponent.gtsb.io/en/nft) versus [how the markdown table displays now.](https://ethereum.org/en/nft/)

### To reproduce

1. Go to https://ethereum.org/en/nft/
2. Scroll down to [the NFT comparison table](https://ethereum.org/en/nft/#nft-comparison)
3. See visual regression

### Expected behavior

The table should be correctly formatted ([as per the previous visual style](https://ethereumorgwebsitedev01-networkupgradescomponent.gtsb.io/en/nft)) dividers between rows and left aligned.

### Screenshots

Current:

Screenshot 2022-09-26 at 11 37 24

Before (expected):

Screenshot 2022-09-26 at 11 37 33

### Additional context

I suspect this happened as part of [the migration to Chakra.](https://github.com/ethereum/ethereum-org-website/issues/6374)

pettinarip:
## Description

Use Chakra `Table` & `List` components in markdown & common pages.

## Related Issue

Fixes #8030 #8031


src/@chakra-ui/gatsby-plugin/components/Table.ts
```diff
@@ -0,0 +1,53 @@
+import { ComponentStyleConfig } from "@chakra-ui/react"
+
+export const Table: ComponentStyleConfig = {
```
minimalsm, line 3: Should we keep the old margins here @nloureiro?


Old:

Screenshot 2022-10-03 at 10 55 34

This PR:

Screenshot 2022-10-03 at 10 55 48

src/@chakra-ui/gatsby-plugin/components/Table.ts
```diff
@@ -0,0 +1,53 @@
+import { ComponentStyleConfig } from "@chakra-ui/react"
+
```
minimalsm, line 2: Table line height is too small here. 

Before:

Screenshot 2022-10-03 at 11 00 28

This PR: 

Screenshot 2022-10-03 at 11 00 25

minimalsm (changes requested):
Good start on this 👍 a few nitpicks on the UI details that our designers might be better suited to weigh in on.

src/@chakra-ui/gatsby-plugin/styles.ts
```diff
@@ -19,15 +19,25 @@ const styles = {
       bg: mode("white", "gray.700")(props),
       lineHeight: "1.6rem",
     },
+    // styles used by markdown pages - we can't use `ListItem` in them.
```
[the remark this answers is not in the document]
pettinarip: Correct. We were using `1.45rem` before. Now that we are using the Chakra scale system, we should ideally use one of these values, 5 or 6:
- 5: '1.25rem'
- 6: '1.5rem'
(from https://chakra-ui.com/docs/styled-system/theme#spacing)

src/@chakra-ui/gatsby-plugin/styles.ts
```diff
@@ -19,15 +19,25 @@ const styles = {
...
+    li: {
```
pettinarip, line 23: Ideally we should not use global styles at all for this. I'm investigating how to avoid it.

minimalsm:
Noting that we're merging #8144, and should address removing these global styles when we merge this one.

pettinarip:
Closing this PR in favor of the work being done for the implementation of the DS. New table: #9932

However, I'll create a new PR to address the refactoring done here to group all the md component in one shared file. I think that is still useful.

PR #8145, opened 2022-10-03
Fix broken Platzi link (Spanish developer resource) 

SavPhill:
on https://ethereum.org/en/developers/learning-tools/:

The link https://platzi.com/blockchain is no longer available, so I have replaced for the alternative link on the platzi domain.

## Description

Replaced broken link

### DIFF
--- a/src/pages/developers/learning-tools.tsx
+++ b/src/pages/developers/learning-tools.tsx
@@ -228,7 +228,7 @@ const LearningToolsPage = ({
     {
       name: "Platzi",
       description: "page-learning-tools-platzi-description",
-      url: "https://platzi.com/blockchain",
+      url: "https://platzi.com/startupsweb3/",
       image: getImage(data.platzi),
       alt: "page-learning-tools-platzi-logo-alt",
       background: "#121f3d",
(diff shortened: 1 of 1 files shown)


minimalsm (commented):
Hey @SavPhill, thanks for the PR.

I think https://platzi.com/escuela/escuela-blockchain/ is more in link with what the original link served (no archive link to double check, but this looks like a better page to me — less selling, more showing what their course entails). Maybe a Spanish speaker could help confirm which is more valuable to link to?

src/pages/developers/learning-tools.tsx
```diff
@@ -228,7 +228,7 @@ const LearningToolsPage = ({
     {
       name: "Platzi",
       description: "page-learning-tools-platzi-description",
-      url: "https://platzi.com/blockchain",
+      url: "https://platzi.com/startupsweb3/",
```
pettinarip, line 231:
```suggestion
      url: "https://platzi.com/escuela/escuela-blockchain/",
```

pettinarip (changes requested):
Thanks for the PR @SavPhill. I agree with @minimalsm on this.

The link you proposed is only focused on how to build a startup in the web3 ecosystem and seems to be paid.

Navigating from the home page, the first main link to enter the blockchain education is https://platzi.com/escuela/escuela-blockchain/ I think is what we want here.

src/pages/developers/learning-tools.tsx
corwintines: Agree with @pettinarip and @minimalsm. Will commit their suggestion and pull in.

corwintines (approved):

Issue #8143, opened 2022-10-02
L2BEAT api is not working


bizzyvinci:
### Describe the bug

L2BEAT api is used to get L2 TVL but it raises error 500.

### To reproduce

1. Go to [layer-2 page](https://ethereum.org/en/layer-2/)
2. Scroll down a bit to where L2 TVL and ETH transfer fee are displayed
3. See error "Error, please refresh."
4. Open developer console and you will see the error is because of 500 status code in response

### Expected behavior

The present TVL and 30 day TVL change ought to be displayed.

### Screenshots

<img width="1467" alt="Screenshot 2022-10-02 at 21 38 18" src="https://user-images.githubusercontent.com/22333930/193475315-41f96967-9262-49d5-b593-ed9ce58b54a2.png">

### Additional context

I believe the issue is in [layer-2 code](https://github.com/ethereum/ethereum-org-website/blob/v6.6.1/src/pages/layer-2.tsx#L213-L215). I couldn't find an official api documentation for L2BEAT but I discovered this [url](https://l2beat.com/api/scaling-tvl.json) from the website. It works and it seems to fit. I assume L2BEAT changed their api url.

minimalsm:
@corwintines do we have a contact L2Beat that set us up with this endpoint?

LuisUrrutia:
Already make a PR for this issue. But there is one thing to discuss: L2Beat response for TVL now has four main keys: `layers2s`, `combined`, `bridges` and `projects`, I assumed that we needed the data from `layers2s`, but maybe we need the data from `combined`. @minimalsm, can u confirm this?

corwintines:
Hey @LuisUrrutia, your line of thinking was correct here. Thanks for looking into this!

@minimalsm I dont currently have one, can try and establish one. I think this came because they launched their new bridge risk framework, but would be nice to know about this stuff ahead of time.

PR #8166, opened 2022-10-04
Change L2Beat API

LuisUrrutia:
## Description
L2Beat changed its API URL and the response structure, so I modified the URL and adapted the calculations to work with the new structure of the response. They have 4 items on the JSON root; I assumed that what we need It's the data from `layers2s` key.

## Related Issue
https://github.com/ethereum/ethereum-org-website/issues/8143


corwintines (approved):
Really appreciate it @LuisUrrutia. Thank you :)

PR #8175, opened 2022-10-05
Migrate `OrderedList` to Chakra

TylerAPfledderer:
## Description

This migrates the [OrderedList](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/OrderedList.tsx)

For @pettinarip: In terms of completely replacing it with just using Chakra's `OrderedList`, I believe this component could be retained. Despite it being primitive in the logic, it is used in at least two places on the site with specific styling, unless there is interest in eventually creating styling variants.

There is an alternative approach that comes to mind, but I'll hear your feedback and thoughts first. 😃 

## Related Issue

#6374

### DIFF
--- a/src/components/OrderedList.tsx
+++ b/src/components/OrderedList.tsx
@@ -1,54 +1,65 @@
 // Libraries
 import React from "react"
-import styled from "@emotion/styled"
+import {
+  Box,
+  ListItem,
+  OrderedList as ChakraOrderedList,
+  SystemStyleObject,
+} from "@chakra-ui/react"
 
 export interface IProps {
   listData: Array<React.ReactNode>
   className?: string
 }
 
-// Styles
-const Content = styled.div`
-  margin-bottom: 1.45rem;
-
-  ol {
-    list-style: none;
-    counter-reset: li-counter;
-    padding-left: 2rem;
-    margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


src/components/OrderedList.tsx
```diff
@@ -1,54 +1,65 @@
...
 // `listData` should be a list of strings, or HTML components
 // ex: [<p>string<p>] or ['string']
 const OrderedList: React.FC<IProps> = ({ listData, className }) => {
   return (
-    <Content className={className}>
-      <ol>
+    <Box mb="1.45rem" className={className}>
+      <ChakraOrderedList
+        styleType="none"
+        pl="2rem"
```
pettinarip, line 40:
```suggestion
        pl={8}
```

src/components/OrderedList.tsx
```diff
@@ -1,54 +1,65 @@
...
+        mb="0"
```
pettinarip, line 41: We will need to add this margin on the start to keep it the same as we have it in prod

```suggestion
        mb="0"
        ms="1.45rem"
```

pettinarip (approved):
Amazing @TylerAPfledderer! this looks great & works perfect. I'm just adding one tiny margin.

PR #8176, opened 2022-10-05
Corrected external broken links

SavPhill:
On line 151 the url was duplicated, I have removed the duplicated text and now the page displays correctly.

On line 163 the url was no longer available, so I have added the url from waybackmachine and it can now be viewed

## Description
Correcting external urls which are no longer viewable.

### DIFF
--- a/src/content/enterprise/index.md
+++ b/src/content/enterprise/index.md
@@ -160,7 +160,6 @@ Here are some of the enterprise applications that have been deployed to the publ
 
 ### Credentials and certifications {#credentials}
 
-- [Utah Counties](https://web.archive.org/web/20210731052556/https://www.utahcounty.gov/Dept/ClerkAud/DigitalCertCopy.html) _issuing digital marriage certificates on Ethereum_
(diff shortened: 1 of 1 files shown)


src/content/enterprise/index.md
```diff
@@ -160,7 +160,7 @@ Here are some of the enterprise applications that have been deployed to the publ
 
 ### Credentials and certifications {#credentials}
 
-- [Utah Counties](http://www.utahcounty.gov/Dept/ClerkAud/DigitalCertCopy.html) _issuing digital marriage certificates on Ethereum_
+- [Utah Counties](https://web.archive.org/web/20210731052556/https://www.utahcounty.gov/Dept/ClerkAud/DigitalCertCopy.html) _issuing digital marriage certificates on Ethereum_
```
corwintines, line 163:
```suggestion
(https://web.archive.org/web/20210731052556/https://www.utahcounty.gov/Dept/ClerkAud/DigitalCertCopy.html) _issuing digital marriage certificates on Ethereum_
```

I think this should be removed if we are using a wayback link. Given it doesnt seem this is currently being used after looking at their page, we should remove this.

corwintines (approved):
Thanks @SavPhill

PR #8204, opened 2022-10-07
Migrate Checkbox to Chakra Native Component

marcellamalune:
Migrate Checkbox to Chakra Native Component

## Description

- Related previous PR: https://github.com/ethereum/ethereum-org-website/pull/7816
- Migrate Checkbox component
- Customize Chakra Checkbox style
- Update components and pages importing Checkbox

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/6374


src/@chakra-ui/gatsby-plugin/components/Checkbox.ts
```diff
@@ -0,0 +1,52 @@
...
+  _checked: {
+    bg: "primary400",
+    _hover: null,
+    borderColor: "black50",
+  },
+  border: "1px",
+  borderColor: "black50",
+  borderRadius: "3px",
+  transition: "all 150ms",
+  _focusVisible: null,
```
pettinarip, line 14:
```suggestion
  _focusVisible: {
     boxShadow: "none",
  },
```

src/@chakra-ui/gatsby-plugin/components/Checkbox.ts
```diff
@@ -0,0 +1,52 @@
+import type { ComponentMultiStyleConfig } from "@chakra-ui/theme"
+
+const baseStyleControl = {
+  bg: "background",
+  _checked: {
+    bg: "primary400",
+    _hover: null,
```
pettinarip, line 7:
```suggestion
    _hover: {
          bg: "primary400",
          borderColor: "primary600",
        },
```

src/@chakra-ui/gatsby-plugin/components/Checkbox.ts
```diff
@@ -0,0 +1,52 @@
...
+  parts: ["control"],
+  baseStyle: {
+    control: baseStyleControl,
+  },
+  sizes: {
+    md: {
+      control: {
+        h: "1.5rem",
+        w: "1.5rem",
+        minW: "1.5rem",
```
pettinarip, line 35:
```suggestion
```

src/components/Checkbox.tsx
```diff
@@ -1,78 +1,65 @@
...
       {children && (
-        <Text as="span" ml={2}>
+        <Text as="span" {...textProps}>
           {children}
         </Text>
       )}
-    </Box>
+    </ChakraCheckbox>
   )
 }
```
pettinarip, line 64: I think we need to get rid of all the custom props/logic that we had before. I would like to have this in a much simpler way.

```tsx
export interface IProps extends CheckboxProps {
  className?: string
}

const Checkbox: React.FC<IProps> = ({ className, ...rest }) => {
  return (
    <ChakraCheckbox className={className} icon={<CustomIcon />} {...rest} />
  )
}
```

This way we are basically just using the same Chakra component. Now, instead of passing a `callback` prop, we should refactor it to do a `onChange` (from the consumer's perspective).

pettinarip (changes requested):
Solid work on this @marcellamalune. Thanks so much for this PR. I left a few comments to cleanup the old props and logic we had before. The rest is great.

src/components/Checkbox.tsx
marcellamalune: @pettinarip Thank you for the feedback, I'm working to implement your suggestions. Should we keep the children as it is in the PR or not?

marcellamalune:
@pettinarip thank you for your feedbacks! I applied all your suggestions.

pettinarip (approved):
Awesome @marcellamalune! 🎉 I've just added some cleanup stuff. Will merge it soon.

PR #8224, opened 2022-10-11
Migrate `ExpandableInfo` to Chakra

TylerAPfledderer:
## Description

This migrates the [ExpandableInfo](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/ExpandableInfo.tsx)

[Per discussion in Discord](https://discordapp.com/channels/714888181740339261/1003686202257449111/1027648992416383117), there was the question about using the `Accordion` prop for this component, but did not make sense to do so for accessibility because of it's current usage as only a forced-open iteration on a single page. So going by the existing functionality was considered the best course of action.

See comments below for explanation on certain Chakra additions.

## Related Issue

#6374

### DIFF
--- a/src/components/ExpandableInfo.tsx
+++ b/src/components/ExpandableInfo.tsx
@@ -68,8 +68,8 @@ const ExpandableInfo = forwardRef<IProps, "div">(
         border="1px solid"
         borderColor="border"
         borderRadius="2px"
-        p="1.5rem"
-        mb="1rem"
+        p={6}
+        mb={4}
         spacing="0"
         background={background ?? "background"}
         position="relative"
@@ -85,25 +85,25 @@ const ExpandableInfo = forwardRef<IProps, "div">(
         <Stack
           justify="space-between"
           align="center"
-          gap={{ base: "2rem", md: "3rem" }}
+          gap={{ base: 8, md: 12 }}
           flexDirection={{ base: "column", md: "row" }}
           width="full"
         >
           {image && <GatsbyImage image={image} alt="" />}
-          <HStack gap="3rem" width="full">
-            <Box mr="1rem">
+          <HStack gap={12} width="full">
+            <Box mr={4}>
               <HStack
                 width="full"
-                m="1rem 0"
+                my={4}
                 sx={{
                   img: {
-                    mr: "1.5rem",
+                    mr: 6,
                   },
                 }}
               >
                 <Heading
-                  mt="0rem"
-                  mb="0.5rem"
+                  mt="0"
+                  mb={1}
                   fontSize={{ base: "2xl", md: "2rem" }}
                   fontWeight="600"
                   lineHeight="1.4"
@@ -146,10 +146,10 @@ const ExpandableInfo = forwardRef<IProps, "div">(
         >
           <Box
             color="text"
-            mt="2rem"
+            mt={8}
             borderTop="1px solid"
             borderColor="border"
-            paddingTop="1.5rem"
+            paddingTop={6}
           >
             {children}
           </Box>
--- a/src/components/ExpandableInfo.tsx
+++ b/src/components/ExpandableInfo.tsx
@@ -6,7 +6,6 @@ import {
   Center,
   ChakraProps,
   Collapse,
-  forwardRef,
   Heading,
   HStack,
   Icon,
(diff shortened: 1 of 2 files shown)


src/components/ExpandableInfo.tsx
```diff
@@ -1,216 +1,155 @@
...
-      transition: {
-        when: "afterChildren",
-      },
-    },
-    expanded: {
-      height: "100%",
-      transition: {
-        when: "beforeChildren",
-      },
+const ExpandableInfo = forwardRef<IProps, "div">(
```
TylerAPfledderer, line 31: Wrapped component in the Chakra `forwardRef` to allow Chakra props to be used on this component at the parent level.

See [run-a-node page](https://github.com/ethereum/ethereum-org-website/blob/d9f6693d92b66f81fa3dc0416fc26be6d8938156/src/pages/run-a-node.tsx#L539) when migrating the styled component to Chakra.

src/components/ExpandableInfo.tsx
```diff
@@ -1,216 +1,155 @@
...
     },
-  }
-  const chevronFlip = {
-    collapsed: {
-      rotate: 0,
-      transition: {
-        duration: 0.1,
+    ref
+  ) => {
+    const { isOpen, getButtonProps, getDisclosureProps } = useDisclosure({
```
TylerAPfledderer, line 45: This is used to provide accessibility for this component, should toggling expansion be used (`forceOpen` = false). `getButtonProps` holds all the necessary `aria-*` attributes for toggling, including `aria-expanded` and `aria-controls`.

`getDisclosureProps` Provides `id` prop for `aria-controls` and the `hidden` attribute, which is needed in conjunction with the accordion pattern from the [WAI-Aria Authoring Practices](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/)

This of course is on contingent on if accordion behavior should still exist.

src/components/ExpandableInfo.tsx
```diff
@@ -1,216 +1,155 @@
...
+              <HStack
+                width="full"
+                m="1rem 0"
+                sx={{
+                  img: {
+                    mr: "1.5rem",
+                  },
+                }}
+              >
+                <Heading mt="0rem" mb="0.5rem" size="md" fontWeight="600">
```
TylerAPfledderer, line 103: `Heading` component used for consistency of the heading typography in the theme config.

The `size` prop is used with the `md` value to be consistent with the current theming of the site. 
- `lineHeight` value is `1.2`
- `fontSize` is `xl`

However, it should be noted that the `xl` fontSize by default from Chakra is not the same as the one from the original config provided by the site. So currently the rendering of the font size will not be correct until the theme config is fully migrated.

src/components/ExpandableInfo.tsx
```diff
@@ -1,216 +1,155 @@
...
+                  transform: "scale(1.25)",
+                  transition: "transform 0.1s",
+                },
+              }}
+            >
+              <Icon as={MdExpandMore} boxSize="initial" size="36" />
+            </Center>
+          )}
+        </Stack>
+        <Collapse
```
TylerAPfledderer, line 140: Transition effect simplified to the single `Collapse` component, as all the fade-in/out effects are apart of the collapse transition.

src/components/ExpandableInfo.tsx
```diff
@@ -1,216 +1,162 @@
...
-export interface IProps {
+export interface IProps extends ChakraProps {
   children?: React.ReactNode
   image?: IGatsbyImageData
   title: ReactNode
   contentPreview: ReactNode
-  background: string
-  forceOpen: boolean
+  background: BackgroundProps["background"]
+  forceOpen?: boolean
```
pettinarip, line 27: 👍🏼

src/components/ExpandableInfo.tsx
```diff
@@ -1,216 +1,162 @@
...
+
+    const animateToggle = isOpen ? "expanded" : "collapsed"
+
+    return (
+      <VStack
+        border="1px solid"
+        borderColor="border"
+        borderRadius="2px"
+        p="1.5rem"
+        mb="1rem"
```
pettinarip, line 72: In general, lets try to use as much as we can the Chakra scale system (for font sizes & spacing). To keep it consistent with the rest of the codebase and future design system.

```suggestion
        p={6}
        mb={4}
```

pettinarip (changes requested):
This looks great @TylerAPfledderer nice job 💪🏼 

I haven't tested the collapsing feature yet but one thing I've noticed is that in the https://ethereum.org/en/run-a-node/ page we have a `width=90%` that is not being applied now. Would suggest replacing the emotion component that is wrapping this component in that page with a chakra approach.

```
// current run-a-node.tsx page implementation
const StyledExpandableInfo = styled(ExpandableInfo)`
  width: 90%;
  align-self: center;
  @media (max-width: ${({ theme }) => theme.breakpoints.m}) {
    margin: 0;
    width: 100%;
  }
`

// migrate this to chakra as well
// now it would be something like
<ExpandableInfo
  ...
  width="90%"
>
```

src/components/ExpandableInfo.tsx
TylerAPfledderer: A bad habit of mine. I'll be sure to keep this in mind for future changes and any other existing PRs! 😄

src/components/ExpandableInfo.tsx
TylerAPfledderer: I'm sorry @pettinarip . I made a grave error in my explanation and needed to review `forwardRef`!! 😅 

So yes, the `forwardRef` from Chakra is an enhanced version of React's `forwardRef` (Check out the [source code](https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/system/src/forward-ref.tsx))

However, the primary purpose of this version of `forwardRef` is to ensure the `as` prop is passed in at the parent level without type error should the given component need to be a different html element or consume a different component on a particular instance.

For example, if this component normally renders a `div` wrapping the contents but one of the parent components needs it to be an `article` in a given instance.

So doing this:
```tsx
<ExpandableInfo as="article">
```

Will not generate a type error.

See the extended documentation regarding the ["as" prop](https://chakra-ui.com/community/recipes/as-prop)

I now believe this is actually unnecessary here, because there is no negative impact to passing props at the parent level without `forwardRef` so long as the component props are typed correctly, no need for the `as` prop, and no need to pass reference at the parent level.

Now the only other thing related to this discussion is my familiarity with the argument on [discouraging the use of `React.FC`](https://github.com/facebook/create-react-app/pull/8177) and would like your opinion on this! 😄

TylerAPfledderer:
As to the width issue, I will go ahead an migrate the instance over in the `run-a-node` page! 👍🏼

TylerAPfledderer:
@pettinarip to add to the issue with the styling from the `run-a-node` page...

There is also a [`margin: 0` applied below the `md` breakpoint](https://github.com/TylerAPfledderer/ethereum-org-website/blob/7cb5c4476f47917de2ba7ab7f78ec6b0a787c871/src/pages/run-a-node.tsx#L178).

There are a couple of [cleaner] approaches to ensure this is rendered properly and not lose the `mb` styling that is in the component.
1. Explicit responsive values: `mb={{ base: 0, md: 4 }}` (you have to restate the original value of `4` else it will be overridden)
2. Use the `useQuery()` hook from Chakra and apply to the sx prop, like so...
```tsx
// Somewhere before the return
const mdBelowQuery = useQuery({ below: 'md' })

//Then on the component
<ExpandableInfo
  sx={{
    [`@media ${mdBelowQuery}`] : {
      mb: 0
    }
>
```

Which under the hood will render:
```css
@media (max-width: 48em) {
  margin-bottom: 0
}
```

The current problem I see is that the string from `useQuery` is not being returned correctly (returning `"@media (max-width: md)"`, and I'm trying to get some help to troubleshoot the issue). But if it was working correctly, might be useful for other components on this page.

Let me know what you think about this!

src/components/ExpandableInfo.tsx
```diff
@@ -1,216 +1,162 @@
...
+                sx={{
+                  img: {
+                    mr: 6,
+                  },
+                }}
+              >
+                <Heading
+                  mt="0"
+                  mb={1}
+                  fontSize={{ base: "2xl", md: "2rem" }}
```
TylerAPfledderer, line 107: If you replaced `md: "2rem"` with `md: 8` here, it will render `8px` because the fontSize prop does not like it.

pettinarip: Oh got it. My bad haha thought it was a spacing prop 🤪

pettinarip:
@TylerAPfledderer thanks for the explanation. I think I prefer the most common approach for now (`mb={{ base: 0, md: 4 }}`). The other one seems to verbose to me.

src/components/ExpandableInfo.tsx
pettinarip: Cool, thanks for the explanation & resource links again.

When we did the TS migration, I didn't know about that discussion about `React.FC`. Later, I saw that CRA was removing the usage of it and it grab my attention. I'm not super deep into the details but I can say that I don't see too much value on using it tbh and totally agree that the implicit `children` prop is not desired.

This is something we could discuss in a GH issue if you have a strong opinion about it and you feel that we should change it 👍🏼

TylerAPfledderer:
> @TylerAPfledderer thanks for the explanation. I think I prefer the most common approach for now (`mb={{ base: 0, md: 4 }}`). The other one seems to verbose to me.

I agree! 😅

src/components/ExpandableInfo.tsx
TylerAPfledderer: Thanks @pettinarip! I do not have a strong preference either way.

In fact current version of the FC does not type the `children`, so you have to explicitly type it either way!

The discussion is indeed old, and React 18 has helped out with this. The implementation of it for the purposes of these components has no impact either way, whether you do this:
```tsx
const someComponent: React.FC<Type> = (props) =>
```
or this:
```tsx
const someComponent = (props: Type) =>
```

But you would still have trouble using generic typing with it though, if any of the components were to use it.

I'm not aware if any of the components actually need that! 😄 

I will revert back to `React.FC` from `forwardRef` as it is not needed here and the component will still accept incoming Chakra Props.

TylerAPfledderer:
@pettinarip should be good to go now with all changes!

pettinarip (approved):
lgtm! thanks @TylerAPfledderer

PR #8209, opened 2022-10-08
chore: migrating the MergeInfographic component to ChakraUI

victorpatru:
## Description

Migrating the MergeInfographic component from emotion to Chakra-ui.

_Please note that I've had difficulties running the development server on my machine. Any feedback welcome and more than happy to make any adjustments as needed!_

## Related Issue

#6374

PR #8246, opened 2022-10-13
refactor: migrate the merge infographic component from emotion to chakraui #6374

victorpatru:
Hey @pettinarip, I've gone ahead and fixed the bugs that I encountered during the migration of the MergeInfographic component.

I wasn't sure about the best practices to fixing the bugs to a pull request so I closed the [previous pull request](https://github.com/ethereum/ethereum-org-website/pull/8209) and created a new one with my final version. (would love some insight on whether this was the right approach). 

_Related Issue #6374_

### DIFF
--- a/src/components/MergeInfographic.tsx
+++ b/src/components/MergeInfographic.tsx
@@ -80,40 +80,42 @@ const MergeInfographic: React.FC<IProps> = ({ className }) => {
         isolation: "isolate",
       }}
     >
-      <Box
-        position="absolute"
-        top="44%"
-        left="2%"
-        width="81%"
-        height="18%"
-        margin={0}
-        padding={0}
-        zIndex={2}
-        display="flex"
-        justifyContent="center"
-        alignItems="center"
-        lineHeight="1em"
-        textAlign="center"
-        maxHeight="2em"
-        color="background"
-        fontSize={{
-          base: "0.625em",
-          sm: "0.875em",
-          md: "1.125em",
-          lg: "1.375em",
-        }}
-        aria-hidden="true"
-      >
-        <Translation id="page-upgrades-merge-infographic-el" />
+      <Box>
+        <Box
+          position="absolute"
+          top="40%"
+          left="2%"
+          width="81%"
+          height="18%"
+          margin={0}
+          padding={0}
+          zIndex={2}
+          display="flex"
+          justifyContent="center"
+          alignItems="center"
+          lineHeight="1em"
+          textAlign="center"
+          maxHeight="2em"
+          color="background"
+          fontSize={{
+            base: "0.625em",
+            sm: "0.875em",
+            md: "1.125em",
+            lg: "1.375em",
+          }}
+          aria-hidden="true"
+        >
+          <Translation id="page-upgrades-merge-infographic-el" />
+        </Box>
+        <SvgText getString={getString} />
+        <Background
+          aria-hidden="true"
+          position="absolute"
+          width="100%"
+          height="100%"
+          zIndex={0}
+        />
       </Box>
-      <SvgText getString={getString} />
-      <Background
-        aria-hidden="true"
-        position="absolute"
-        width="100%"
-        height="100%"
-        zIndex={0}
-      />
     </AspectRatio>
   )
 }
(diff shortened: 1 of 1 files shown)


src/components/MergeInfographic.tsx
```diff
@@ -111,20 +84,55 @@ const MergeInfographic: React.FC<IProps> = ({ className }) => {
...
         "page-upgrades-merge-infographic-alt-text",
         intl
       )}
+      position="relative"
+      width="100%"
+      sx={{
+        isolation: "isolate",
+        aspectRatio: `${25 / 11}`,
+      }}
     >
```
TylerAPfledderer, line 100: Because the `aspectRatio` prop is used, I think it would be better to use the [AspectRatio](https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/layout/src/aspect-ratio.tsx) component instead, else `Box`. 

`VStack` doesn't make sense here because there are no visually stacking elements in the DOM flow (just about everything is absolute-positioned instead), and `VStack` by default renders a `marginTop` for each child element. In this case, it's simply rendering unnecessary space. :)

With `AspectRatio` you can use the `ratio` prop instead with the needed value.

**NOTE**: All the child components will need to be wrapped in a `div` because `AspectRatio` is looking for a single child. Else it'll break the SSR.

```suggestion
<AspectRatio
    className={className}
    role="img"
    aria-label={translateMessageId(
      "page-upgrades-merge-infographic-alt-text",
       intl
    )}
    position="relative"
    width="100%"
    ratio={25 / 11}
    sx={{
      isolation: "isolate",
    }}
>
```

This component will provide more robust positioning and sizing logic. I wonder if @pettinarip agrees on this approach instead of just using `Box`! 😄

src/components/MergeInfographic.tsx
```diff
@@ -111,20 +84,55 @@ const MergeInfographic: React.FC<IProps> = ({ className }) => {
...
-      <ExecutionLayer aria-hidden="true">
+      <Box
+        position="absolute"
+        top="44%"
```
TylerAPfledderer, line 103: `top` should have the value of `40%` here.

src/components/MergeInfographic.tsx
```diff
@@ -82,22 +31,46 @@ const SvgText: React.FC<SvgProps> = ({ getString }) => {
       xmlnsXlink="http://www.w3.org/1999/xlink"
       aria-hidden="true"
     >
-      <Text x="2%" y="35%" fontSize={lg}>
+      <text x="2%" y="35%" fontSize={lg} textAnchor="start" fill="currentColor">
```
TylerAPfledderer, line 34: With all of the `text` elements, you could do the same as the `Background` svg in using the Chakra factory, so you can keep the code DRY with the `textAnchor` and `fill` props.

```tsx
const Text = chakra("text", {
  baseStyle: {
    textAnchor: "start",
    fill: "currentColor",
  },
})
```

TylerAPfledderer (changes requested):
Hey @victorpatru ! I got a chance to take a look at your changes and provided some feedback for you 😄

src/components/MergeInfographic.tsx
pettinarip: Great suggestion, yes. Lets try to use as much as we can the Chakra components. And good point about the usage of `VStack`, I think its not necessary here.

src/components/MergeInfographic.tsx
pettinarip: That is weird. You are correct, in the code we have `40%` but in prod, inspecting the element using the dev tools, it says `44%` 🤷🏼 whatever...both values work fine at the end IMO.

pettinarip (changes requested):
Great job @victorpatru! this is looking great.

Lets implement the `AspectRatio` component in this refactor. With that, I think we can call this done.

victorpatru:
Committed the changes we've discussed. Thanks @pettinarip @TylerAPfledderer

victorpatru:
The build kept failing because the AspectRatio Chakra element expects a single child element. It should be fixed now 😄

pettinarip (approved):
LGTM! thanks for the great work you did here @victorpatru 🙏🏼 

Be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And we've just released our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) also 🥳!

pettinarip:
@all-contributors please add @victorpatru for code

PR #8260, opened 2022-10-15
Disable text selection of example ETH address

SavPhill:
In page https://ethereum.org/en/get-eth/ we have an example public ETH address, where we mark this as an example address only. 

eth-do-not-copy

My PR adds 3 lines of simple css to the CodeBox class in this page which will disable the copying of text in the browser, to ensure the user cannot copy this ETH address by mistake.


corwintines (approved):
Cool! Learned something new. Thanks, @SavPhill!

mmiguerodriguez:
Does it make sense to have incorrect hex characters in the example address?

PR #8263, opened 2022-10-15
grammatical error

farhandotdev:
Grammatical error
## Description
here is small grammatical error in the readme file

The changes is just only embraces to embrace


corwintines:
Thanks for the suggestion @farhandotdev, but embraces looks more correct.

Issue #8051, opened 2022-09-27
Migrate `Slider` component to Chakra UI


pettinarip:
This is part of our [Implement UI library](https://github.com/ethereum/ethereum-org-website/issues/6374) epic.

## Description
Migrate `src/components/Slider.tsx` from `emotion` styled components to `chakra`.

This component is used in https://ethereum.org/en/what-is-ethereum/
![image](https://user-images.githubusercontent.com/468158/192540772-0f55bcff-4fb5-4b30-9420-70e69afb3d97.png)

## Additional resources
Check our [epic](https://github.com/ethereum/ethereum-org-website/issues/6374) for more information & check our [migration guide](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/chakra-migration-guide.md).

lawlesx:
I will start with this

lawlesx:
I would like to work on this.

lawlesx:
@pettinarip Created a PR resolving this
Sorry for the delay, last week was hectic.

pettinarip:
Done! thanks t o @lawlesx

PR #8266, opened 2022-10-15
Migrate slider component to chakra UI

lawlesx:
## Description
Migrated `Slider.tsx` component to Chakra UI

### Screenshots
**Dark Mode**
Before
Screenshot 2022-10-16 at 12 02 45 AM
After
Screenshot 2022-10-16 at 12 02 34 AM

**Light Mode**
Before
Screenshot 2022-10-16 at 12 03 01 AM
After
Screenshot 2022-10-16 at 12 02 55 AM

## Related Issue

 Resolves Issue #8051


src/components/Slider.tsx
```diff
@@ -146,42 +52,89 @@ const Slider: React.FC<IProps> = ({ children, onSlideChange }) => {
...
+      borderWidth="1px"
+      borderStyle="solid"
+      borderRadius="0.3rem"
+      w="full"
+      bg="sliderBg"
+    >
+      <Box overflow="hidden" ref={emblaRef}>
+        <Flex>{children}</Flex>
+      </Box>
+      <Flex
```
pettinarip, line 67: I think we should position this as `absolute` as before. If not it looks weird.

image

The dots should be centered with the buttons. Take a look at [the prev implementation](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Slider.tsx#L52).

src/components/Slider.tsx
```diff
@@ -146,42 +52,89 @@ const Slider: React.FC<IProps> = ({ children, onSlideChange }) => {
...
+        justifyContent={{ sm: "left", base: "center" }}
+        mb={{ sm: "0", base: "1rem" }}
```
pettinarip, line 69:
```suggestion
        mb={{ sm: 0, base: 4 }}
```

src/components/Slider.tsx
```diff
@@ -146,42 +52,89 @@ const Slider: React.FC<IProps> = ({ children, onSlideChange }) => {
...
-                ? theme.colors.slider.btnColor
-                : theme.colors.slider.btnColorDisabled
-            }
-          />
-        </Button>
-      </Buttons>
-      <Dots>
+    <Box
+      position="relative"
+      p="2rem"
```
pettinarip, line 57:
```suggestion
      p={8}
```

src/components/Slider.tsx
```diff
@@ -146,42 +52,89 @@ const Slider: React.FC<IProps> = ({ children, onSlideChange }) => {
...
+      borderWidth="1px"
+      borderStyle="solid"
+      borderRadius="0.3rem"
```
pettinarip, line 60: Lets use `base` here (which is `0.25rem`) to be consistent with the rest of the site.

```suggestion
      borderRadius="base"
```

src/theme.ts
```diff
@@ -369,16 +367,14 @@ const darkColors = {
...
-    btnColorDisabled: "#737373",
-  },
+  sliderBg: "#191919",
+  sliderBorder: "#404040",
+  sliderDot: "#A4A4A4",
+  sliderDotActive: "#FF7324",
+  sliderBtnBg: "#404040",
+  sliderBtnColor: white,
+  sliderBtnBgDisabled: "#404040",
+  sliderBtnColorDisabled: "#737373",
```
pettinarip, line 377: Why do we need these changes? if not used please revert

pettinarip (changes requested):
Good job @lawlesx. The dots look a bit misaligned. I've left a possible solution.

src/theme.ts
lawlesx: Screenshot 2022-10-22 at 5 01 44 PM

In the previous implementation I was not able to apply the colors
Example:
```ts
bg="slider.bg"
```
This change was not reflected

I tried to reach out in [discord](https://discord.com/channels/714888181740339261/870313128242602004/1028259754360983562) but got no solution so the did the above change

src/components/Slider.tsx
lawlesx: Sure the buttons looks odd and I did try the prev implementation but there was some issues with the alignment so I made it in a presentable way.

Also as you said the button should be aligned center.
I pushed some changes that should fix it

## Attaching screen shots for new and old changes
### New Pushed changes as this comment date
Screenshot 2022-10-22 at 6 26 32 PM

### Old changes as in the live [website](https://ethereum.org/en/what-is-ethereum/)
Screenshot 2022-10-22 at 5 13 01 PM

lawlesx: Also I made sure it doesn't break for mobile view

pettinarip (changes requested):
Thanks for making the adjustments @lawlesx.

The last thing I see is that button's size is not the previous one. They should be `35px` instead of `40px`.

src/components/Slider.tsx
```diff
@@ -146,42 +52,93 @@ const Slider: React.FC<IProps> = ({ children, onSlideChange }) => {
...
+          aria-label="MdChevronRight"
+          onClick={scrollNext}
+          icon={<MdChevronRight fontSize={24} focusable={true} />}
+          isRound
+          _hover={{ boxShadow: "none" }}
+          _focus={{ boxShadow: "none" }}
+          bg={nextBtnEnabled ? "sliderBtnBg" : "sliderBtnBgDisabled"}
+          height="40px"
+          color={nextBtnEnabled ? "sliderBtnColor" : "sliderBtnColorDisabled"}
+        />
```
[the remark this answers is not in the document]
lawlesx: While appling `height` and `width` to `35px` the the buttons kind of breaks.
Screenshot 2022-10-25 at 9 18 23 PM

So I fixed it by applying `size="sm"`
Screenshot 2022-10-25 at 9 20 51 PM

I feel it looks close to the previous button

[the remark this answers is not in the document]
pettinarip: Nice! I like it even more (using standard Chakra values). Gj

pettinarip (approved):
Awesome @lawlesx this looks great.

PR #8269, opened 2022-10-16
typo fix for ethresear.ch link

benlazzero:
## Fixes the link typo for ethresearch that reads 'head toethereasr.ch'. 

Handles the link differently by not using the utility intl key 'page-upgrades-head-to' and using a new one 'page-upgrades-eth-research' which now includes the ethresear.ch link. 'page-upgrades-head-to' was only being used for this one link. I have changed every language that was using the key/value 'page-upgrades-head-to'.

1. change intl keys to 'page-upgrades-eth-research'
2. change intl values to include 'ethresear.ch' after translation


corwintines (changes requested):
Thanks for this @benlazzero! Only change I think needs to be made is the key name itself. Should be `page-upgrades-head-to-ethresearch` I think. Ill make this change.

corwintines (approved):
Great fix @benlazzero! Thanks for this.

Be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. We also have our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) 🥳!

corwintines:
@all-contributors can you add @benlazzero for code

PR #8287, opened 2022-10-18
Update index.md

knititwearit:
documentation - changed "subsequent" to previous as all previous blocks change

Proposed Change to Index.md

## Description

The text was changed as all "previous" blocks would change not all "subsequent" ones.


corwintines:
Hey @knititwearit. Thanks for the PR. Im going to close it out as its correct as is. If you were to change a block in history, it would mean you would have to change all blocks after that one, not before it.

corwintines:
@all-contributors can you add @knititwearit for idea

knititwearit:
> Hey @knititwearit. Thanks for the PR. Im going to close it out as its correct as is. If you were to change a block in history, it would mean you would have to change all blocks after that one, not before it.

Thanks :) learnt something here at least

Issue #8298, opened 2022-10-19
Feature request: Modify feedback widget behavior


konopkja:
### Is your feature request related to a problem? Please describe.

We want to modify the behavior of our feedback widget so its more prominent on desktop devices as right now we do not get enough desktop user feedback (most responses are coming from mobile devices).
<img width="216" alt="Screenshot 2022-10-19 at 12 10 55" src="https://user-images.githubusercontent.com/100724231/196758871-ee6976b9-5ce6-4bee-a7d7-860798661946.png">

### Describe the solution you'd like

**Add new behavior:**
<img width="206" alt="Untitled" src="https://user-images.githubusercontent.com/100724231/196759364-0ac17209-8021-4016-8470-76bdbcd135b1.png">

- Add new animation after 30 sec spent on page: change the default state to this new expanded state (there should be an animation of this expanding to the left 500MS)
- Once user clicks into it and closes the component, it should return to the small bubble
- This should happen once per session max.
- Animation is only for desktop devices.

### Describe alternatives you've considered

NA

marcellamalune:
Hey there, I'd like to help with this! @konopkja @pettinarip

pettinarip:
Assigned @marcellamalune thanks again!

corwintines:
@konopkja I think we can close this now

corwintines:
Go ahead and reopen if you disagree

PR #8312, opened 2022-10-20
modify feedback widget behaviour

marcellamalune:
## Description

- Add `isExpanded` state to FeedbackWidget component
- Set a 30sec timeout as a side effect to change `isExpanded` state to true and clear timeout on unmount
- Use media query to apply style changes only on desktop devices
- Apply 0.5 transition timing for width and border-radius
- Reset `isExpanded` to false when `isOpen` is reset too
- Import `Text` component from ChakraUI
- Truncate text while expanding to show a smooth animation

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/8298


src/components/FeedbackWidget.tsx
```diff
@@ -220,8 +240,23 @@ const FeedbackWidget: React.FC<IProps> = ({ className }) => {
...
     <>
-      <FixedDot onClick={handleOpen} bottomOffset={bottomOffset} id="dot">
+      <FixedDot
+        onClick={handleOpen}
+        bottomOffset={bottomOffset}
+        isExpanded={isExpanded}
+        id="dot"
+      >
         <StyledFeedbackGlyph />
+        <Text
```
pettinarip, line 250: I think we could wrap this `Text` with a transition like `ScaleFade` (https://chakra-ui.com/docs/components/transitions/usage#scalefade-transition) and delay it a bit.

IMO the animation should be:
1. Expand the button to the left (keeping the icon on the left)
2. Once the expand animation finish, we display the Text with a fade in or scale fade animation

pettinarip (changes requested):
Thanks for the PR @marcellamalune. I think the basic animation is working but there are some problems when the expanding animation starts. Take a look at this to see it clearly:
https://user-images.githubusercontent.com/468158/198043249-d6e3e1f4-5c69-4c76-bb2f-c46ed4ea9608.mp4

pettinarip:
This is looking nice so far @marcellamalune. Wanted to check with @konopkja what he thinks about it? should the animation be faster? or is it ok as it is?

konopkja:
> 

looks nice! agree, the animation could be twice as fast. other than that it looks good.

marcellamalune:
Thank you for your feedback @pettinarip @konopkja! I speeded up the animation.

konopkja:
> * Deploy preview

ok this is better. @pettinarip im good with this.

corwintines:
Screen Shot 2022-10-31 at 12 01 38 PM

Hey @marcellamalune!

Noticing this isn't working on mobile.

marcellamalune:
> Screen Shot 2022-10-31 at 12 01 38 PM
> 
> Hey @marcellamalune!
> 
> Noticing this isn't working on mobile.

Sorry about that, I screwed up something with the latest changes without realizing. Now it's fixed.

corwintines:
Thanks for fixing @marcellamalune. Will get to review this week, but I did add the `hacktoberfest-accepted` label to get you credit for this still.

src/components/FeedbackWidget.tsx
```diff
@@ -33,6 +36,17 @@ const FixedDot = styled(NakedButton)<{
...
   }
+  @media (min-width: ${({ theme }) => theme.breakpoints.l}) {
+    width: ${({ isExpanded }) => (isExpanded ? "15rem" : "3rem")};
+    border-radius: ${({ isExpanded }) => (isExpanded ? "50px" : "50%")};
+    #feedback-wrapper {
+      width: ${({ isExpanded }) => (isExpanded ? "13.5rem" : "3rem")};
+      position: ${({ isExpanded }) => (isExpanded ? "absolute" : "flex")};
+    }
+    #expanded-prompt {
+      display: ${({ isExpanded }) => (isExpanded ? "flex" : "none")};
```
pettinarip, line 47: I don't love the idea of using ids to achieve this. Why can't we implement this logic using style props?

Can't we do this like
```tsx
<Text ... width={{ base: "3rem", lg: isExpanded ? "13.5rem" : "3rem" }}>
  <Box
    as={Translation}
    id="feedback-card-prompt-page"
    display={{ base: "flex", lg: isExpanded ? "flex" : "none" }}
  />
</Text>
```

pettinarip (changes requested):
The overall result looks great right now @marcellamalune. My only concern is more about the implementation using ids, wondering if we could adjust a bit that.

konopkja:
@pettinarip @marcellamalune can we push this forward?

src/components/FeedbackWidget.tsx
marcellamalune: @pettinarip thanks for the feedback and for the improvements you've made 🙂

pettinarip (approved):

corwintines (approved):
Thanks @marcellamalune and @pettinarip!

Issue #8307, opened 2022-10-20
Bug report - Wrong information in the tutorial page: Understanding the Yellow Paper's EVM Specifications 


ali-bahjati:
### Describe the bug

In [the section 9.1: Basics](https://ethereum.org/en/developers/tutorials/yellow-paper-evm/#91-basics) the article mentions:

> The EVM also has a separate non-volatile storage model that is maintained as part of the system state - this memory is organized into word arrays (as opposed to word-addressable byte arrays in the stack). The stack is referred to as "memory" while the non-volatile storage is referred to as "storage".

1. To the best of my knowledge it is wrong, stack vs memory vs storage are all three different things.

### To reproduce

Go to [the article](https://ethereum.org/en/developers/tutorials/yellow-paper-evm/#91-basics)

### Expected behavior

It should state that memory is a separate thing and explain it. Ideally it should also explain call-data too.

corwintines:
Closed with above PR

PR #8322, opened 2022-10-21
update yellow paper -> clarify memory types

jmcook1186:
## Description

adds a more clear description of the EVM memory types 

## Related Issue

#8307

### DIFF
--- a/src/content/developers/tutorials/yellow-paper-evm/index.md
+++ b/src/content/developers/tutorials/yellow-paper-evm/index.md
@@ -32,19 +32,21 @@ The term [Turing-complete](https://en.wikipedia.org/wiki/Turing_completeness) me
 
 This section gives the basics of the EVM and how it compares with other computational models.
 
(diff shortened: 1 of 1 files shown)


src/content/developers/tutorials/yellow-paper-evm/index.md
```diff
@@ -32,19 +32,21 @@ The term [Turing-complete](https://en.wikipedia.org/wiki/Turing_completeness) me
...
 
 ```yul
 mstore(0, 0x60A7)
 ```
 
-It fills 32 memory locations - i.e. one word - with zeros in locations 0-29, 0x60 to 30, and 0xA7 to 31.
+`mstore` is one of three opcodes the EVM provides for interacting with memory - it loads a word into memory. The other two are `mstore8` which loads a single byte into memory, and `mload` which moves a word from memory to stack.
 
-The EVM also has a separate non-volatile storage model that is maintained as part of the system state - this memory is organized into word arrays (as opposed to word-addressable byte arrays in the stack). The stack is referred to as "memory" while the non-volatile storage is referred to as "storage".
+The EVM also has a separate non-volatile **storage** model that is maintained as part of the system state - this memory is organized into word arrays (as opposed to word-addressable byte arrays in the stack). This storage is where contracts keep persistent data - a contract can only interact with its own storage. Storage is organised in key-value mappings.
```
wackerow, line 45: ([Per style guide](https://ethereum.org/en/contributing/style-guide/#american-english))
```suggestion
The EVM also has a separate non-volatile **storage** model that is maintained as part of the system state - this memory is organized into word arrays (as opposed to word-addressable byte arrays in the stack). This storage is where contracts keep persistent data - a contract can only interact with its own storage. Storage is organized in key-value mappings.
```

wackerow (approved):
Nice improvements =)

corwintines (approved):

Issue #8323, opened 2022-10-21
Bug report - Uninformative link


JoshDavisLight:
### Describe the bug

"Correlation penalty" linked text on [Consensus Mechanisms](https://ethereum.org/en/developers/docs/consensus-mechanisms/pos/) directs to the Gasper paper, which doesn't have any immediately relevant information regarding correlation penalties. A search of the pdf for the string "correlation penalty" turns up 0 results.

### To reproduce

1. Go to [Consensus Mechanisms](https://ethereum.org/en/developers/docs/consensus-mechanisms/pos/)

3. Scroll down to Crypto-Economic Security'
4. Click on 'correlation penalty'

### Expected behavior

Clicking on the link should bring a user to a page with more information about what a correlation penalty is

### Additional context

Consider removing the link or finding another resource with more concise information regarding correlation penalties.

emmanuel-awosika:
Hi @JoshDavisLight. Thanks for pointing this out. I just replaced the link with a new one pointing to [this resource on correlation penalties](https://eth2book.info/bellatrix/part2/incentives/slashing#the-correlation-penalty). Turns out the inactivity leak links to points to the same paper, but it has new link pointing to a [more descriptive resource](https://eth2book.info/bellatrix/part2/incentives/inactivity). 

Let me know if this solves the issue.

corwintines:
Closed with above PR

PR #8327, opened 2022-10-22
Fixed links for inactivity leak and correlation penalties on PoS page

emmanuel-awosika:
[This issue](https://github.com/ethereum/ethereum-org-website/issues/8323) raised a concern about the links on correlation penalty pointing to the GASPER paper, which doesn't have information directly relevant to the topic (or perhaps it's hidden, as searching "correlation penalty" in the PDF doesn't bring up any results). The link on inactivity leaks also points to the same paper. 

I've replaced both with links pointing to sections in Ben Eddington's Eth2book since it explicitly describes correlation penalties and the inactivity leak.

## Related Issue
#8323


corwintines (approved):
Thanks, @emmanuel-awosika!

PR #8333, opened 2022-10-24
Refactor `DocsNav` with Chakra `LinkOverlay`, etc.

TylerAPfledderer:
## Description
- fix breakpoint token value naming (`m` to `md`)
- add props to parent `Flex` for aria pagination pattern
- rename `Card` to `CardLink`
- apply `LinkOverlay` tandem to `CardLink`
- simplify `CardLink` so it's content is reusable
- update styling logic for doc data passed to `CardLink`
- replace empty `Flex` components with `Spacer`

## Related Issue
None

## Additional Information
When providing dynamic values to Chakra props, instead of something like...
```tsx
<Box textAlign={ isTrue ? 'right' : undefined } >
```
I usually do this
```tsx
// Clears typecheck by not returning 'false' (will return the prop or `undefined`)
<Box { ...( isTrue && { textAlign: 'right' } ) } >
```

This is my first time using the latter approach in this repo. If this is not preferred, I will gladly fix it! 😅

### DIFF
--- a/src/components/DocsNav.tsx
+++ b/src/components/DocsNav.tsx
@@ -1,5 +1,13 @@
 import React from "react"
-import { Flex, FlexProps, Text } from "@chakra-ui/react"
+import {
+  Box,
+  Flex,
+  FlexProps,
+  LinkBox,
+  LinkOverlay,
+  Spacer,
+  Text,
+} from "@chakra-ui/react"
 
 import Link from "./Link"
 import Emoji from "./Emoji"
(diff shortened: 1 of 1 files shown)


src/components/DocsNav.tsx
```diff
@@ -81,50 +120,25 @@ const DocsNav: React.FC<IProps> = ({ relativePath }) => {
...
-          </TextDiv>
-          <Link to={nextDoc.to} textDecoration="none" p={4} h="100%">
-            <Emoji text=":point_right:" fontSize="5xl" />
-          </Link>
-        </Card>
-      ) : (
-        <Flex />
-      )}
+      {previousDoc ? <CardLink docData={previousDoc} /> : <Spacer />}
+      {nextDoc ? <CardLink docData={nextDoc} /> : <Spacer />}
```
pettinarip, line 141: nitpick: instead of changing the datastructure of `docData` I'd handle the `isNext` / `isPrev` as individual props.

```tsx
// you'd have
{previousDoc ? <CardLink docData={previousDoc} isNext /> : <Spacer />}
{nextDoc ? <CardLink docData={nextDoc} isPrev /> : <Spacer />}
```
I think its simpler and clearer than having it nested inside the `docData` object.

pettinarip (approved):
This refactor is great @TylerAPfledderer. Much cleaner than before. Thanks for doing this.

Left a small comment there to see what do you think 👍🏼

src/components/DocsNav.tsx
TylerAPfledderer: Makes sense. I'll make that adjustment! 😁

TylerAPfledderer:
@pettinarip pushed the suggested change!

pettinarip:
Awesome! thanks @TylerAPfledderer

PR #8334, opened 2022-10-24
refactor(eth-video): apply `useColorModeValue` for video source

TylerAPfledderer:
## Description

Replace dark mode check from emotion with `useColorModeValue` hook from Chakra.

This provides a way to not use a nested ternary to determine what source to use, while returning a local video asset based on color mode if no `videoSrc` provided.

## Related Issue
None


pettinarip (approved):
Nice! good catch also bc we were using the emotion `useTheme`.

PR #8338, opened 2022-10-24
Updated Bug Bounty Example

fredriksvantes:
## Description

As stated in the bug bounty program, the EF Bug Bounty Program bases its impact on the effect a vulnerability has on the Network: "The severity is calculated according to the OWASP risk rating model based on Impact on the Ethereum Network and Likelihood. [View OWASP method](https://www.owasp.org/index.php/OWASP_Risk_Rating_Methodology)"

Below the above line, there is an example of what could be described as a "High" vulnerability which states "There is a consensus bug between two clients, but it is difficult or impractical for the attacker to trigger the event." - This could indeed be accurate from a network perspective - if the issue was in a client being used by a large part of the network (say 45%), while if the consensus bug was in a client being used by 0.1% of the network it would most likely not have as big of an impact on the network itself.

This PR tries to make the example more clear.


minimalsm (approved):
Thanks @fredriksvantes

PR #8372, opened 2022-10-25
Refactor `CallToContribute` component

TylerAPfledderer:
## Description
Refactor `CallToContribute` post Chakra migration.

- Create `ContentColumn` reusable component for the image and the content.
- Use the `Show` component to only show the image above the `lg` breakpoint (see additional information below).
- Create `DescriptionParagraph` reusable component for the content paragraphs.

## Related Issue
N/A

## Additional Information
As of this PR, the Chakra UI migration is not complete. If I were to do the following with `Show`:

```tsx
<Show above="lg">
```

The token value would not be recognized. This is due to the under the hood functionality of `Show` using the `useTheme` hook from Chakra. `useTheme` looks through the emotion package to retrieve theme values from the Context, and I believe it is receiving the old theme config instead of the Chakra config. Indeed there is no issue when applying Chakra's token values in the style props (I believe it has to do with how the theme config is acquired different from `useTheme`)

`useToken` solves this issue temporarily until the complete merge occurs.

### DIFF
--- a/src/components/CallToContribute.tsx
+++ b/src/components/CallToContribute.tsx
@@ -1,4 +1,4 @@
-import { Flex, Heading, Icon, Text } from "@chakra-ui/react"
+import { Flex, Heading, Icon, Show, Text, useToken } from "@chakra-ui/react"
 import { FaGithub } from "react-icons/fa"
 import React, { ReactNode } from "react"
 import Link from "./Link"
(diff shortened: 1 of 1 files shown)


pettinarip (approved):
Nice! surprising to know that `Show` doesn't read the values directly from the Chakra theme 🤔 however it kind of makes sense if it is using `useTheme` under the hood. Thanks for sharing that information!

TylerAPfledderer:
@pettinarip certainly! I have not found anything yet that leads me to believe otherwise without deleting that old file.

Issue #8345, opened 2022-10-24
Epic: Change i18n Gatsby plugin/library


pettinarip:
This epic is part of our Q4 2022 roadmap: #8399

### Description

Not long ago, we detected that our final JS bundles (served to the clients) had a very big size which creates a slow load performance for our users impacting directly the UX and the accessibility of the site.

Since then we have run [some investigations](https://www.notion.so/efdn/Gatsby-i18n-issue-9ef381505b2448048c0ac1be90759f5d) and we have identified that the main problem is in [our i18n plugin](https://github.com/gatsbyjs/themes/issues/93). For that reason, we need to change it to one that scales well with larger sites like ours.

We need a solution that solves the 2 main issues we have been facing recently:
- Memory consumption on build stage
- Generation of big bundles

### TODO

- [x] Decide what i18n plugin/solution we want to use
 - └── We are going to use https://github.com/microapps/gatsby-plugin-react-i18next which will also change our underlying i18n lib to https://react.i18next.com/
- [x] Install the potential plugin/solution and see if it covers the 2 main issues we had #8344
- [x] Create a PR to change the i18n plugin #8406
- [x] Remove HTML from translation files #8592
- [x] ~~#8721~~

corwintines:
@pettinarip I think this can be close dnow

pettinarip:
Yeah! Done!

PR #8406, opened 2022-10-29
React i18next

pettinarip:
## Description

Implement `react-i18next` by using the next plugin https://github.com/microapps/gatsby-plugin-react-i18next

- GC preview: https://ethereumorgwebsitedev01-reacti18next.gatsbyjs.io/
- Netlify preview: [https://pettinarip-eth-org.netlify.app/](https://pettinarip-eth-org.netlify.app/en/)

Known issues:

- Custom 404 pages are not working on the GC env. Still under investigation but I don’t think it is a blocker. Its working fine on Netlify env.

## Related Issue

#8345


gatsby-browser.tsx
```diff
@@ -16,53 +12,6 @@ import "@formatjs/intl-locale/polyfill"
...
-    const queryParams = search || ""
-    const newUrl = withPrefix(`/${detected}${originalPath}${queryParams}`)
-    window.localStorage.setItem("eth-org-language", detected)
-    window.location.replace(newUrl)
-
-    return null
-  }
-
-  return <Layout {...props}>{element}</Layout>
-}
```
pettinarip, line 68: `gatsby-browser` & `gatsby-ssr` content was replaced by the usage of `gatsby-plugin-layout`.

src/components/Staking/StakingHierarchy.tsx
```diff
@@ -287,9 +290,11 @@ const StakingHierarchy: React.FC<IProps> = () => {
...
         </Glyph>
         <Content>
-          <p>
-            <Translation id="page-staking-hierarchy-solo-p1" />
-          </p>
+          <p
+            dangerouslySetInnerHTML={{
+              __html: t("page-staking-hierarchy-solo-p1"),
+            }}
+          />
```
pettinarip, line 297: This is a temporary implementation. In the next iteration we are planning in using the following approach: https://react.i18next.com/latest/trans-component#using-with-react-components where we will also need to modify our translation files.

src/pages-conditional/wallets.tsx
```diff
@@ -494,7 +478,16 @@ export const listImage = graphql`
 `
 
 export const query = graphql`
-  query WalletsPage {
+  query WalletsPage($language: String!) {
+    locales: allLocale(filter: { language: { eq: $language } }) {
```
pettinarip, line 482: On each page, we need to fetch the selected language strings.

minimalsm:
When reviewing I noticed a bug with our link styling for links embedded as HTML in JSON files - #8547.

wackerow:
Noticed our intl tutorials aren't loading properly
image

pettinarip:
> Noticed our intl tutorials aren't loading properly

Uhmm good catch @wackerow! I'm looking into it now. Thanks

minimalsm:
Noticed this on the other PR, but it's also happening here.

If a string is not available in a language, I would expect it to fall back to English strings. This is what happens in this PR:

Screenshot 2022-11-17 at 19 01 16

[English](https://ethereumorgwebsitereacti18n-i18nparsehtmljson62891.gtsb.io/en/layer-2)

[Chinese - A language with translated JSON strings](https://ethereumorgwebsitereacti18n-i18nparsehtmljson62891.gtsb.io/zh/layer-2)

[Slovenian - A language **without** translated JSON files](https://ethereumorgwebsitereacti18n-i18nparsehtmljson62891.gtsb.io/sl/layer-2)

Screenshot 2022-11-17 at 19 01 16

pettinarip:
> Noticed this on the other PR, but it's also happening here.
> 
> If a string is not available in a language, I would expect it to fall back to English strings. This is what happens in this PR:

Good catch @minimalsm I thought I had that covered! will take a look, thanks!

corwintines (approved):
LGTM, nice one

Noting, we need to update [/docs/best-practices.md](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/best-practices.md#%EF%B8%8F-translation-initiative)

minimalsm (approved):
LGTM :)

pettinarip:
@corwintines I'll create a new PR with those updates 👍🏼 thanks for pointing that out.

PR #8408, opened 2022-10-29
Updated Wallet Link

SavPhill:
on https://ethereum.org/en/wallets/find-wallet/ the link to Portis.io directs to https://portis.io/.

As we can see, https://portis.io has not configured the SSL certificate correctly when accessing directly without the  'www' subdomain. The looks unsecure to the visitor, so I have fixed the link to use the 'www' subdomain.

## Description
Fixing the external link to look more secure for the visitor.


minimalsm (approved):
Thanks for the fix @SavPhill ❤️

Issue #8210, opened 2022-10-08
Bug report - Dead Hyperlink


deshvin:
### Describe the bug

On: https://ethereum.org/en/developers/docs/nodes-and-clients/run-a-node/#getting-the-client

The hyperlink for "consensus client" points to a dead link: https://ethereum.org/endevelopers/docs/nodes-and-clients/#consensus-clients

The correct URL is: https://ethereum.org/en/developers/docs/nodes-and-clients/#consensus-clients

A '/' is missing between "en" and "developers"

### To reproduce

1. Go to https://ethereum.org/en/developers/docs/nodes-and-clients/run-a-node/#getting-the-client
2. Click on the "consensus client" hyperlink on the first line

### Expected behavior

The hyperlink for "consensus client" should open https://ethereum.org/en/developers/docs/nodes-and-clients/#consensus-clients

### Desktop (please complete the following information)

MacOS Monterey
Brave
106.0.5249.103

amit-ksh:
Can I fix this issue??

corwintines:
Closed with above PR

PR #8411, opened 2022-10-29
updated consensus client link

vanshika-srivastava:
Contributed to the issue - https://github.com/ethereum/ethereum-org-website/issues/8210

Consensus client link updated in the MD


minimalsm:
Hey @vanshika-srivastava, sorry about this. This issue has already been fixed and is not working as expected. Your solution hardcodes the language path to English, which we don't want as it will redirect non-English speakers to the English version of this page. Thanks for the PR and contribution though! ❤️

PR #8413, opened 2022-10-29
refactor: Migrate GhostCard from emotion to chakra-ui.

douglasmakey:
## Description

Migrate GhostCard from emotion to chakra-ui.

### DIFF
--- a/src/components/GhostCard.tsx
+++ b/src/components/GhostCard.tsx
@@ -1,34 +1,5 @@
 import React from "react"
-import styled from "@emotion/styled"
-
-const Container = styled.div`
-  position: relative;
-  margin-top: 0.5rem;
-  align-self: stretch;
-`
-
-const BaseCard = styled.div`
-  height: 100%;
-  width: 100%;
-  border-radius: 2px;
-`
-
-const Card = styled(BaseCard)`
-  z-index: 2;
-  padding: 1.5rem;
-  text-align: left;
-  border: 1px solid ${(props) => props.theme.colors.border};
-  background-color: ${(props) => props.theme.colors.ghostCardBackground};
-`
-
-const Ghost = styled(BaseCard)`
-  z-index: -1;
-  position: absolute;
-  bottom: 0.5rem;
-  left: 0.5rem;
-  border: 1px solid ${(props) => props.theme.colors.border};
-  background-color: ${(props) => props.theme.colors.ghostCardGhost};
-`
+import { Box } from "@chakra-ui/react"
 
 export interface IProps {
   children?: React.ReactNode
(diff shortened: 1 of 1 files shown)


src/components/GhostCard.tsx
```diff
@@ -1,45 +1,40 @@
...
+      zIndex="hide"
+      position="absolute"
+      backgroundColor="ghostCardGhost"
+      bottom="2"
+      left="2"
+      border="1px solid"
+      borderColor="border"
+      borderRadius="2px"
+      height="full"
+      width="full"
```
TylerAPfledderer, line 21: Two things can be improved with these three props.

Since they are used in both child components, maybe they should be in a reusable object and spread into each component?

Secondly, since both the `height` and `width` props in both instances have the same value (`full`), you can use the `boxSize` prop instead!

So in total...
```tsx
// Imported Type signature from Chakra totally optional, but great to help with the intellisense! :D
import { type ChakraProps } from '@chakra-ui/react'

const baseCardStyles: ChakraProps = {
  boxSize: "full",
  borderRadius: "2px"
}

return (
  // ... //
  <Box 
    {...baseStyles}
  // ... //
```

Purely semantics here, but I see this as a cleaner way to go (and utilizing a Chakra-specific prop)! 😅

TylerAPfledderer (changes requested):
Hey @douglasmakey! Good work! Thought I leave a comment with a possible change. 😄

src/components/GhostCard.tsx
```diff
@@ -1,45 +1,40 @@
...
+  <Box position="relative" mt="2" alignSelf="stretch" className={className}>
+    <Box
+      zIndex="hide"
+      position="absolute"
+      backgroundColor="ghostCardGhost"
+      bottom="2"
+      left="2"
+      border="1px solid"
+      borderColor="border"
+      borderRadius="2px"
```
pettinarip, line 19:
```suggestion
      borderRadius="sm"
```

src/components/GhostCard.tsx
pettinarip: Didn't know about `boxSize` 👍🏼  TIL

pettinarip (approved):
Nicely done @douglasmakey thanks for the PR.

pettinarip:
@douglasmakey be sure to [join the discord](https://discord.gg/CetY6Y4) if you are interested in contributing further to the project or have any questions for the team. And we've just released our 2022 POAPs so remember to [claim yours](https://ethereum.org/en/contributing/#poap) also 🥳!

pettinarip:
@all-contributors please add @douglasmakey for code

PR #8433, opened 2022-11-01
Fix link overlay in `CardList`

TylerAPfledderer:
## Description

Due to CSS stacking context, the description text in the `Card` component of `CardList` is rendered on top of the pseudo element rendered by `LinkOverlay`, because the component for this text applies the `opacity` prop with a value less than 1 which puts it higher in the z-stack than the `zIndex={0}` rendered by the overlay.

This results in an area above the description text where you would not be able to access the anchor element to click.

This PR converts the `opacity` prop to the `color` prop.

In addition, adds `useColorModeValue` to ensure the color of this text in both light and dark mode meets the color contrast requirements for accessibility.

## Related Issue

N/A

### DIFF
--- a/src/components/CardList.tsx
+++ b/src/components/CardList.tsx
@@ -7,6 +7,7 @@ import {
   LinkBox,
   LinkOverlay,
   StackProps,
+  useColorModeValue,
 } from "@chakra-ui/react"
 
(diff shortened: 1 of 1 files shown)


pettinarip (approved):
Nice catch! thanks for the PR

PR #8441, opened 2022-11-01
Update Matomo tracking eventValue for QuizWidget

wackerow:
## Description
- Send "1" for correct answer, "0" for incorrect

## Related Issue
None filed, request per @konopkja

### DIFF
--- a/src/components/Quiz/QuizWidget.tsx
+++ b/src/components/Quiz/QuizWidget.tsx
@@ -171,9 +171,7 @@ const QuizWidget: React.FC<IProps> = ({ quizKey, maxQuestions }) => {
       eventCategory: "Quiz widget",
       eventAction: "Question answered",
       eventName: `QID: ${questionId}`,
-      eventValue: answer.isCorrect
-        ? `Correct answer`
-        : `AID: ${answer.answerId}`,
+      eventValue: answer.isCorrect ? "1" : "0",
     })
     setShowAnswer(true)
   }


src/components/Quiz/QuizWidget.tsx
```diff
@@ -171,9 +171,7 @@ const QuizWidget: React.FC<IProps> = ({ quizKey, maxQuestions }) => {
       eventCategory: "Quiz widget",
       eventAction: "Question answered",
       eventName: `QID: ${questionId}`,
-      eventValue: answer.isCorrect
-        ? `Correct answer`
-        : `AID: ${answer.answerId}`,
+      eventValue: answer.isCorrect ? "1" : "0",
```
samajammin, line 174: Does string vs. integer matter here?

samajammin: Without looking at docs or other implementations we've done, my gut thinks these should be integers.

wackerow: The `eventValue` type is currently set up to [accept a string](https://github.com/ethereum/ethereum-org-website/blob/7410a42d41d0c63fa8ca0ad68d9e30a99c93705f/src/utils/matomo.ts#L6)... We could try updating the interface to accept a number, we should just make sure this passes appropriately to Matomo. 

This should still fix the issue on Matomo side of just being able to split these into two categories.

konopkja:
@wackerow  @samajammin can we push this forward?

corwintines (approved):

PR #8450, opened 2022-11-02
Add Hardhat+Solidity VSCode extension

fvictorio:
## Description

Added the Hardhat VSCode extension. I normally would've added it last, but I think it makes sense to have the VSCode extensions grouped, otherwise the list is like "VSCode extension, formatter, VSCode extension".

## Related Issue

No issue.


corwintines (approved):
Nice! Thanks, @fvictorio!

PR #8457, opened 2022-11-02
Fix fee explanation

ligi:
IMHO the sentence before made no sense - fixed it


corwintines (approved):
This makes more sense. Thanks, @ligi!

PR #8459, opened 2022-11-02
Quiz: Make answer more precise

ligi:
IMHO saying `always` here is wrong/not 100% precise - I have seen people giving away ETH (e.g. to test something out - especially in the early days)

### DIFF
--- a/src/data/quizzes/questionBank.ts
+++ b/src/data/quizzes/questionBank.ts
@@ -598,7 +598,7 @@ const questionBank: QuestionBank = {
       },
       {
         id: "e003-d",
-        label: "always a scam",
+        label: "extremely likely a scam",
         explanation:
           "ETH giveaways are always scams. Reporting and ignoring scammers is best.",
       },
--- a/src/data/quizzes/questionBank.ts
+++ b/src/data/quizzes/questionBank.ts
@@ -598,7 +598,7 @@ const questionBank: QuestionBank = {
       },
       {
         id: "e003-d",
-        label: "Extremely likely a scam",
+        label: "Are very likely to be a scam",
         explanation:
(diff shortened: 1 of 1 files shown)


corwintines (approved):
Thanks, @ligi.

Added a commit to capitalize the answers

Just wanted to leave my initial thoughts. I don't have too strong of a preference here, but I can see how this is more correct than what was there before. Wouldn't mind hearing what @konopkja @minimalsm and @wackerow have to say.

konopkja:
thank you for calling this out I think there is definitely some missed nuance. The goal here is to protect people and while EF might not be doing any giveaways now (dunno about the past), there could be other orgs that do (i know coinmarketcap was doing 1 eth giveaway to celebrate the merge for example)

**We could rephrase the correct answer to:**
Are very likely to be a scam

**Idea for discussion:**
perhaps we can also add to the explanation that Ethereum Foundation will not do any airdrops neither will vitalik

src/data/quizzes/questionBank.ts
```diff
@@ -581,24 +581,24 @@ const questionBank: QuestionBank = {
...
         id: "e003-c",
-        label: "commonly performed by prominent members of the community",
+        label: "Commonly performed by prominent members of the community",
         explanation:
           "Prominent community members do not do ETH giveaways. Scammers pretend well-know individuals, such as Elon Musk, are doing giveaways to give them the scam a sense of legitimacy.",
       },
       {
         id: "e003-d",
-        label: "always a scam",
+        label: "Extremely likely a scam",
```
ligi, line 601:
```suggestion
        label: "Are very likely to be a scam",
```

ligi:
> We could rephrase the correct answer to:
> Are very likely to be a scam

done

Issue #8310, opened 2022-10-20
Advanced / Networking Layer Suggests nodes use discv5


ryandotsmith:
### Describe the bug

Firstly, I'm not sure this is the right category to report this issues. It has to do with content not the technical implementation. Happy to post this elsewhere.

The docs [here](https://ethereum.org/en/developers/docs/networking-layer/#execution-discovery) indicate that discv5 is used to discover nodes on the ethereum network. However, nodes such as geth don't use discv5 as of today. They use discv4 while exploratory development happens on discv5.

I would like to suggest that we update this section of the docs to explicitly indicate that discv4 is used today and that discv5 is being developed and may be used in the future.

### To reproduce

1. Click https://ethereum.org/en/developers/docs/networking-layer/#execution-discovery

### Expected behavior

I expect to see that discv4 is currently being used

ryandotsmith:
I'd be happy to make these changes if / when we get to the bottom of this issue.

jmcook1186:
yes, you are right - the execution clients still use discv4 by default - but discv5 can be toggled on via CLI. 
Thanks for poiting this out - please lmk if you want to raise a PR, otherwise I can fix.

ryandotsmith:
> please lmk if you want to raise a PR, otherwise I can fix.

Thanks for the quick reply! I will open a PR.

ryandotsmith:
@jmcook1186 PR opened!

corwintines:
Hey @ryandotsmith, thanks for opening this. 

Just wanted to let you know I edited the description links from https://ethereum.org/en/developers/docs/networking-layer/#consensus-discovery to https://ethereum.org/en/developers/docs/networking-layer/#execution-discovery based on your PR.

ryandotsmith:
@corwintines brilliant!

PR #8462, opened 2022-11-02
Add discv4/v5 note to Execution Discovery section

ryandotsmith:
## Description

My goal is to provide clarity to other developers working on the discovery of execution nodes. The current version of Geth doesn't use v5 but instead uses v4. As a new networking developer, I learned the hard way that v5 isn't used by Geth and so this change makes it easier for other devs to learn that v4 is what is currently in use.

## Related Issue

#8310


corwintines (approved):
Thanks, @ryandotsmith!

jmcook1186 (approved):
thanks!

PR #8483, opened 2022-11-03
Update APR endpoint

wackerow:
## Description
- Updates StakingStatsBox with new endpoint from beaconcha.in that fetches a more accurate live APR, removing the need for the calculateStakingRewards utility. 
- Improves metric-specific loading and error handling states.

## Related Issue
None filed


src/components/Staking/StakingStatsBox.tsx
```diff
@@ -80,47 +92,64 @@ const StatsBoxGrid: React.FC<IProps> = () => {
...
 
-  // TODO: Improve error handling
-  if (error) return <StatErrorMessage />
-
   return (
     <Container>
       <Cell>
-        <Value>{totalEth}</Value>
+        {totalEth === ZERO ? (
+          <Spinner />
```
minimalsm, line 125: What color should the spinners here be @nloureiro ([link to page](https://ethereumorgwebsitedev01-aprendpoint.gtsb.io/en/staking))? The same color as the item it's rendering?

Screenshot 2022-11-04 at 09 16 31

Screenshot 2022-11-04 at 09 18 06

nloureiro: I think we can reduce the spinners a little bit to `--spinner-size: 1rem;`

For the color, I would go to `--eth-colors-gray-200` #E2E8F0

The other thing that loud improves is to give a fixed height to avoid the scroll jump in the pages when the numbers load and the spinners get out. dunno if it's within the scope of this PR

Screen Shot 2022-11-04 11 39 36 AM

wackerow: image

This is what I see on light mode when I use `gray.200` (#E2E8F0)

If we want a dimmed color on both light and dark mode, I would suggest just using the "disabled" semantic color token

Dark mode with first option using "disabled" (other two using "gray.200")
image
Light mode, same setup:
image

wackerow: I would address a minHeight separately, perhaps as part of the hero revamp that will include a subset for this type of hero

nloureiro: ok. deal. I will add this to the design system

minimalsm: Created an issue for this in #8516.

Merging this PR as it's a huge UX win over what we currently have.

minimalsm (approved):
Thanks for this @wackerow ❤️

PR #8505, opened 2022-11-04
Add section on "layer 3" scaling

Joeysantoro:
Add a short section explaining why scaling solutions can't just be stacked on top of each other, inspired by https://vitalik.ca/general/2022/09/17/layer_3.html.

### DIFF
--- a/src/content/developers/docs/scaling/index.md
+++ b/src/content/developers/docs/scaling/index.md
@@ -92,11 +92,6 @@ Learn more about [Validium](/developers/docs/scaling/validium/).
 - Not all solutions require utilizing the Ethereum consensus algorithm directly, and alternatives can offer benefits that would otherwise be difficult to obtain.
 - No one scaling solution is enough to fulfill the [Ethereum vision](/upgrades/vision/).
 
-## Why not just stack scaling solutions? {#why-not-stack-scaling}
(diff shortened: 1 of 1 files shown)


minimalsm:
Thanks for the PR @Joeysantoro, I'll try to get some other feedback here and get this reviewed this week :-)

src/content/developers/docs/scaling/index.md
```diff
@@ -92,6 +92,11 @@ Learn more about [Validium](/developers/docs/scaling/validium/).
 - Not all solutions require utilizing the Ethereum consensus algorithm directly, and alternatives can offer benefits that would otherwise be difficult to obtain.
 - No one scaling solution is enough to fulfill the [Ethereum vision](/upgrades/vision/).
 
+## Why not just stack scaling solutions? {#why-not-stack-scaling}
```
corwintines, line 95: I'd like to include the article for reading, but I think this section is a bit premature given we are still working on getting layer 2's online in their full form. As the post says "it seems premature to settle definitions while the architecture of the multi-rollup ecosystem is far from set in stone and most of the discussion is happening only in theory."

corwintines (approved):
Thanks @Joeysantoro!

corwintines:
@all-contributors can you add @Joeysantoro for content

src/content/developers/docs/scaling/index.md
Joeysantoro: Fair enough! I think the content itself answers an important point about Layer 2s if the naming can be pushed back for future discussions.

PR #8506, opened 2022-11-05
refactor(footer): migrate component to Chakra

TylerAPfledderer:
## Description

Migrate `Footer` component to Chakra

## Related Issue

#6374

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -351,7 +351,6 @@ const Footer: React.FC<IProps> = () => {
                         dir={isPageRightToLeft ? "auto" : "ltr"}
                         textDecor="none"
                         color="text200"
-                        _after={{ color: "text200" }}
                         _hover={{
                           textDecor: "none",
                           color: "primary",
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -1,105 +1,22 @@
+import {
+  Box,
+  Flex,
+  Heading,
+  Icon,
+  List,
+  ListItem,
+  SimpleGrid,
+  useToken,
+} from "@chakra-ui/react"
+import { graphql, StaticQuery } from "gatsby"
 import React from "react"
-import styled from "@emotion/styled"
+import { FaGithub, FaTwitter, FaYoutube, FaDiscord } from "react-icons/fa"
 import { useIntl } from "react-intl"
-import { StaticQuery, graphql } from "gatsby"
-import { Icon } from "@chakra-ui/react"
-import { FaDiscord, FaGithub, FaTwitter, FaYoutube } from "react-icons/fa"
-
-import { getLocaleTimestamp } from "../utils/time"
-import Translation from "./Translation"
-import Link from "./Link"
-import { isLangRightToLeft, TranslationKey } from "../utils/translations"
 import { Lang } from "../utils/languages"
-
-const StyledFooter = styled.footer`
-  padding-top: 3rem;
-  padding-bottom: 4rem;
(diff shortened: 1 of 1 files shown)


src/components/Footer.tsx
```diff
@@ -140,6 +57,8 @@ const Footer: React.FC<IProps> = () => {
 
   const isPageRightToLeft = isLangRightToLeft(intl.locale as Lang)
 
+  const [smallBp] = useToken("breakpoints", ["sm"])
```
TylerAPfledderer, line 60: See comment below at line 328

src/components/Footer.tsx
```diff
@@ -372,54 +291,92 @@ const Footer: React.FC<IProps> = () => {
...
+              "@media (max-width: 1300px)": {
+                gridTemplateColumns: "repeat(3, auto)",
+              },
+              [`@media (max-width: ${smallBp})`]: {
+                gridTemplateColumns: "repeat(2, auto)",
+              },
+              "@media (max-width: 500px)": {
+                gridTemplateColumns: "auto",
+              },
+            }}
```
TylerAPfledderer, line 338: Because some of the breakpoints used are not part of the theme config, there is not a concise way in handling the change of value in Chakra.

`useToken` at line 60 is used to retrieve the correct breakpoint value for `sm` so it can be interpolated here.

TylerAPfledderer (commented):
Provided a comment regarding responsive `gridTemplateColumns` prop for the link sections.

src/components/Footer.tsx
```diff
@@ -1,105 +1,22 @@
...
-    svg {
-      fill: ${(props) => props.theme.colors.primary};
-    }
-  }
-`
-
-const SocialIcons = styled.div`
-  margin: 1rem 0;
-`
-const SocialIcon = styled(Icon)`
```
TylerAPfledderer, line 94: This was unused and therefore not kept in the migration

minimalsm:
Thanks for the PR @TylerAPfledderer :-) we'll try to get to reviewing this soon.

src/components/Footer.tsx
pettinarip, line 60: I agree with this but I think this should be `md` instead of `sm`
```suggestion
  const [mediumBp] = useToken("breakpoints", ["md"])
```

pettinarip (changes requested):
Nice job @TylerAPfledderer.

For some reason, I don't see the little arrows on the external links. I couldn't realize yet why it is happening. Could take a closer look later if you can't figure it out.
image

src/components/Footer.tsx
```diff
@@ -1,105 +1,22 @@
...
-
-const FooterLink = styled(Link)`
-  text-decoration: none;
-  color: ${(props) => props.theme.colors.text200};
-  svg {
-    fill: ${(props) => props.theme.colors.text200};
-  }
-  &:after {
-    color: ${(props) => props.theme.colors.text200};
-  }
```
TylerAPfledderer, line 78: This was the culprit to the missing external icons when applied with Chakra. The object created with the `__after` prop on the `Link` component was overriding the styling applied [within the component](https://github.com/ethereum/ethereum-org-website/blob/2713f4cfdb4596f9d73ee0a9dc2a0661f5f58a0e/src/components/Link.tsx#L95-L99)

This wasn't needed anyway because the pseudo inherits color from the parent element.

pettinarip (approved):
That's awesome @TylerAPfledderer thanks for looking into that problem with the arrow as well :)

PR #8529, opened 2022-11-08
Refactor `Leaderboard` (Post Chakra Migration)

TylerAPfledderer:
## Description

- Apply `useColorModeValue` to tidy up color mode styles
  - Uses one instance with an object since there are multiple props
- Add hidden text around leaderboard names for improved screen reader accessibility
  - Before: when link is focused, screen reader reads out "username link"
  - After: when link is focused, screen reader reads out "In place number X with Y points username (see github profile) link"
  - Mentions github if user has one provided
- Convert to `List` and `ListItem` elements from Chakra, and provide an aria label so the user knows what this list is for
  - Label added for accessibility for blinds users with a screen reader and reducing cognitive load
  - Screen reader will now read out "Bug hunting leaderboard list with 5 items..."
- Replace `opacity` prop with `color` prop for the score text.
  - This will allow the pointer cursor to appear over this text and the area to be clickable, per [CSS Stacking Context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context)

> NOTE: Screen reader readouts come from NVDA. Results may vary with JAWS, VoiceOver, etc. 

## Related Issue
N/A


src/components/Leaderboard.tsx
```diff
@@ -26,14 +29,26 @@ export interface IProps {
...
 
   return (
-    <Box
+    <List
       bgColor="background"
-      boxShadow={colorMode === "dark" ? "tableBox.dark" : "tableBox.light"}
+      boxShadow={colorModeStyles.listBoxShadow}
       w="100%"
       mb={8}
+      aria-label="Bug hunting leaderboard"
```
TylerAPfledderer, line 51: Could be more generic, even though this component is only used for bug bounties.

src/components/Leaderboard.tsx
```diff
@@ -53,67 +68,77 @@ const Leaderboard: React.FC<IProps> = ({ content, limit = 100 }) => {
...
+                />
+                <Flex flex="1 1 75%" direction="column" mr={8}>
+                  <LinkOverlay
+                    as={Link}
+                    href={hasGitHub ? `${githubUrl}${username}` : "#"}
+                    textDecor="none"
+                    color="text"
+                    hideArrow
+                  >
+                    <VisuallyHidden>{`In ${PLACE_WORDS[idx]} place with ${score} points`}</VisuallyHidden>
```
TylerAPfledderer, line 116: This is a mistake with a small const list of words to use in saying the place (first, second, etc.) as the leaderboard component can have many more than just five in the list.

Could convert to...
`In place number X with Y points`

TylerAPfledderer (commented):
Provided my own comments on possible changes due to oversight on my part. Would like feedback on these comments.😄

src/components/Leaderboard.tsx
pettinarip: We should pass this prop from outside. Actually we are using the `Leaderboard` from [the acknowledge page](https://ethereum.org/en/contributing/translation-program/acknowledgements/) as well.

pettinarip: (and it should be wrapped by a translation function)

src/components/Leaderboard.tsx
```diff
@@ -26,14 +29,26 @@ export interface IProps {
...
+      listBoxShadow: "tableBox.light",
+      linkBoxShadow: "tableItemBox.light",
+      scoreColor: "blackAlpha.700",
+    },
+    {
+      listBoxShadow: "tableBox.dark",
+      linkBoxShadow: "tableItemBox.dark",
+      scoreColor: "whiteAlpha.600",
+    }
+  )
```
pettinarip, line 43: Nice refactor

src/components/Leaderboard.tsx
pettinarip: I like this improvement and think I prefer your second option here. I think its more worth having the second type of text than this one that is going to read `undefined` on the other links.

Another thing to have in mind is translations. This should be wrapped with the translation function & we should create a new entry for it in the `src/intl/en/common.json` file.

pettinarip (changes requested):
Thanks for always trying to improve the UX. I've left a few observations that I think we need to solve before merging it.

src/components/Leaderboard.tsx
TylerAPfledderer: Oh! I'm seeing the [TranslationLeaderboard](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/TranslationLeaderboard.tsx) component being used in the page you referenced.

Should this also be added to the `src/intl/en/common.json`?

src/components/Leaderboard.tsx
TylerAPfledderer: How would the translation function be used here for a string that includes dynamic values?

`translationMessageId`, correct?

pettinarip: That's a good question. Yes, you should use `translationMessageId` or the `Translation` component to do this.

But...since we don't have support for dynamic values in those functions yet, I would say: lets ignore the translations for this particular sentence for now.

pettinarip (approved):
Nice job @TylerAPfledderer. I've updated the translation functions to use the newer ones and I've removed some margins that are coming from the global styles.

PR #8539, opened 2022-11-08
Redirection after language choice

vdusart:
Redirection after language choice.

## Description

After choosing the language, we had to come back to the page we were reading.
With this feature, the page from which we come is saved in a parameter `from` in the url and the choice of the language redirects us to this page but in the new language.

See bellow a flow of url:
```
https://ethereum.org/en/developers/docs/programming-languages/
https://ethereum.org/en/languages/?from=/developers/docs/programming-languages/
https://ethereum.org/fr/developers/docs/programming-languages/
```

## Related Issue

None


src/components/Nav/index.tsx
```diff
@@ -97,25 +96,6 @@ const HomeLogo = styled(HomeIcon)`
...
-    svg {
-      fill: ${(props) => props.theme.colors.primary};
-    }
-  }
-`
-
-const NavIcon = styled(Icon)`
-  fill: ${(props) => props.theme.colors.text};
-`
-
```
pettinarip, line 118: Wondering why we have removed these, were they unused components?

pettinarip: If so, thanks :)

pettinarip (approved):
Cool feature @vdusart thanks for doing that

src/components/Nav/index.tsx
vdusart: Yes they were all 3 unused :smile:

PR #8570, opened 2022-11-11
Update index.md

cnn-rnn:
add a section on decentralized search and how it benefits DeSci


nhsz:
@jmcook1186 what do you think about these changes?

vincentweisser:
Imo too tangential from the core of what decentralized science is about, maybe you can rephrase and focus it much more on desci?

minimalsm:
Hey :) Thanks for the PR @cnn-rnn ⚡ 

Only just now getting the chance to look at this. Generally agree w/ @vincentweisser that this is tangential and shouldn't live on this page. Definitely interesting though! If this is something you believe should have its own page, please open an issue and we can discuss and find where it should live.

Closing this particular PR out, as we should start with an issue discussing where this content might live.

Thanks again 🙏

minimalsm:
@all-contributors please add @cnn-rnn for ideas

PR #8572, opened 2022-11-11
refactor: migrate expandable card to chakra

jb-0:
## Description
Migrates the expandable card to Chakra. Not 100% sure on the desired animation approach here, but I've gone with Chakra Fade - happy to revisit at a later date.

## Related Issue
#6374 

## Screenshots
image
image
image


TylerAPfledderer (changes requested):
Hey @jb-0!

Indeed, I agree the `Accordion` multi-component is best approach here. However, it was not intended to have a component like the `Fade` component wrap `AccordionPanel`. Based on the original components motion elements, it should actually be the other way around because `AccordionPanel` uses Chakra's `Collapse` component.

At the moment, I think it would get a little messy to litter the diff with comments versus stating all of my thoughts in this one comment, but this component can take advantage of `Accordion`'s unique feature to greatly reduce the code.

I would remove the need of `isVisible` custom state in favor of a render prop to obtain [AccordionItem internal state](https://chakra-ui.com/docs/components/accordion/usage#accessing-the-internal-state).

The `allowToggle` boolean prop applied to `Accordion` will also keep the toggling internal as there are no external triggers to make it a controlled component. In addition, the `AccordionButton` [per Chakra suggestion](https://chakra-ui.com/docs/components/accordion/usage#import) should be wrapped in an element with the `heading` role (if not a heading element) to support [Accessibility Patterns](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/).

There are also a couple of elements that were used in the original component that are no longer needed in `Accordion`, such as wrapping a "button-like" around `less / more` vs just using `Box` because `AccordionButton` is one whole selectable area. You can run the `trackCustomEvent` function through the render prop using the `isExpandable` internal state. 

My only other comment is if this component should have it's own `Accordion` and not make the `AccordionItem` component the parent, and use the `Accordion` component at the parent level? It should have no change in behavior aside from also using the `allowMultiple` prop. But this is a matter of preference and the component should work either way. 

I hope this helps in improving this migration!😄

jb-0:
Hi Tyler, thanks for the comment & Chakra insights. I've kept my replies inline to keep this somewhat trackable, committed some changes based on the comment.

> Hey @jb-0!
> 
> Indeed, I agree the `Accordion` multi-component is best approach here. However, it was not intended to have a component like the `Fade` component wrap `AccordionPanel`. Based on the original components motion elements, it should actually be the other way around because `AccordionPanel` uses Chakra's `Collapse` component.
>
The Chakra Accordion transition is a bit abrupt compared to the original component. So attempting to keep this as close to the original component as possible, which displayed the text via a fade. I've moved the fade to within the panel, but visually moving it down a level doesn't make a difference as far as I can tell?

> At the moment, I think it would get a little messy to litter the diff with comments versus stating all of my thoughts in this one comment, but this component can take advantage of `Accordion`'s unique feature to greatly reduce the code.
> 
> I would remove the need of `isVisible` custom state in favor of a render prop to obtain [AccordionItem internal state](https://chakra-ui.com/docs/components/accordion/usage#accessing-the-internal-state).
> 
> The `allowToggle` boolean prop applied to `Accordion` will also keep the toggling internal as there are no external triggers to make it a controlled component.

The render prop is interesting! I always find it wierd seeing function syntax within jsx like that, but it works for the most part. Unfortunately it means that the content can no longer be clicked to collapse. As we are doing a like for like component refactor I think this is quite important.

So instead I've opted to make it a controlled component through the use of the [index prop](https://github.com/chakra-ui/chakra-ui/discussions/2619). This removes the need for the `isVisible` conditional around the panel.

> In addition, the `AccordionButton` [per Chakra suggestion](https://chakra-ui.com/docs/components/accordion/usage#import) should be wrapped in an element with the `heading` role (if not a heading element) to support [Accessibility Patterns](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/).
>
Done


> There are also a couple of elements that were used in the original component that are no longer needed in `Accordion`, such as wrapping a "button-like" around `less / more` vs just using `Box` because `AccordionButton` is one whole selectable area. You can run the `trackCustomEvent` function through the render prop using the `isExpandable` internal state, triggering it if this state is false as it shown in the click handler for the parent element.
I've switched the more / less button (that was just a div) to Chakra Text now. As for the rest of the comment, I believe it's addressed by making the component controlled.

> My only other comment is if this component should have it's own `Accordion` and not make the `AccordionItem` component the parent, and use the `Accordion` component at the parent level? It should have no change in behavior aside from also using the `allowMultiple` prop. But this is a matter of preference and the component should work either way.
> 
That approach isn't permitted in Chakra, because of the way Chakra passes state between components internally, so an `AccordionItem` is required - see screenshot image


> I hope this helps in improving this migration!smile

src/components/ExpandableCard.tsx
```diff
@@ -1,93 +1,22 @@
...
-  background-color: transparent;
-  color: ${(props) => props.theme.colors.primary};
-`
+import {
+  Accordion,
+  AccordionButton,
+  AccordionItem,
+  AccordionPanel,
+  Box,
+  Button,
```
pettinarip, line 15: Cleanup unsed deps

src/components/ExpandableCard.tsx
```diff
@@ -107,111 +36,121 @@ const ExpandableCard: React.FC<IProps> = ({
...
-            "ExternalLink"
-          )
-        ) {
-          !isVisible && trackCustomEvent(matomo)
-          setIsVisible(!isVisible)
-        }
+    <Accordion
+      border="1px solid"
+      borderColor="border"
+      borderRadius="2px"
```
pettinarip, line 61:
```suggestion
      borderRadius="sm"
```

src/components/ExpandableCard.tsx
```diff
@@ -107,111 +36,121 @@ const ExpandableCard: React.FC<IProps> = ({
...
+      display="flex"
+      flex-direction="column"
+      marginBottom="1rem"
```
pettinarip, line 64:
```suggestion
      marginBottom="4"
```

src/components/ExpandableCard.tsx
```diff
@@ -107,111 +36,121 @@ const ExpandableCard: React.FC<IProps> = ({
...
-          variants={showHide}
-          animate={isVisible ? "expanded" : "collapsed"}
-          initial={false}
-        >
-          <Text
-            variants={fadeInOut}
-            animate={isVisible ? "expanded" : "collapsed"}
-            initial={false}
+      <AccordionItem borderTopWidth="0" flex="1">
+        <h3 style={{ margin: 0 }}>
```
pettinarip, line 73:
```suggestion
        <Heading as="h3" m={0}>
```

src/components/ExpandableCard.tsx
```diff
@@ -107,111 +36,121 @@ const ExpandableCard: React.FC<IProps> = ({
...
+          <AccordionButton
+            width="100%"
+            padding={`1.5rem 1.5rem ${isVisible ? "0" : "1.5rem"}`}
```
pettinarip, line 76: Lets keep this padding for the entire box and remove the extra bottom margin from the title section (see the following comment)
```suggestion
            p={6}
```

src/components/ExpandableCard.tsx
```diff
@@ -107,111 +36,121 @@ const ExpandableCard: React.FC<IProps> = ({
...
+                  }}
+                  margin="1rem 0"
+                >
+                  {!!Svg && <Svg />}
+                  <Text
+                    fontSize="xl"
+                    fontWeight="semibold"
+                    flex="1"
+                    marginTop="0rem"
+                    marginBottom="0.5rem"
```
pettinarip, line 107:
```suggestion
                    m="0"
```

src/components/ExpandableCard.tsx
```diff
@@ -107,111 +36,121 @@ const ExpandableCard: React.FC<IProps> = ({
...
+            </Box>
+          </AccordionButton>
+        </h3>
+        <AccordionPanel
+          paddingX="1.5rem"
+          paddingBottom="1.5rem"
+          paddingTop="0"
+          onClick={onClick}
+        >
+          <Fade in={isVisible}>
```
pettinarip, line 139: Appreciate the effort to keep the same type of animation we had but I think that in this transition we can start using the default animation from the Chakra's Accordion component which makes this simpler and cleaner.
```suggestion
```

src/components/ExpandableCard.tsx
```diff
@@ -107,111 +36,121 @@ const ExpandableCard: React.FC<IProps> = ({
...
-    </Card>
+            <Box
+              display="flex"
+              width="100%"
+              alignItems="center"
+              flexDir={{ base: "column", sm: "row" }}
+            >
+              <Box
+                marginBottom={{ base: "0.5rem", sm: 0 }}
+                marginRight={{ base: 0, sm: "1rem" }}
```
pettinarip, line 91:
```suggestion
                marginBottom={{ base: 2, sm: 0 }}
                marginRight={{ base: 0, sm: 4 }}
```

src/components/ExpandableCard.tsx
```diff
@@ -107,111 +36,121 @@ const ExpandableCard: React.FC<IProps> = ({
...
+              >
+                <Box
+                  display="flex"
+                  alignItems="center"
+                  sx={{
+                    svg: { marginRight: "1.5rem" },
+                  }}
+                  margin="1rem 0"
```
pettinarip, line 99:
```suggestion
                  my="4"
```

src/components/ExpandableCard.tsx
```diff
@@ -107,111 +36,121 @@ const ExpandableCard: React.FC<IProps> = ({
...
+          paddingBottom="1.5rem"
+          paddingTop="0"
+          onClick={onClick}
+        >
+          <Fade in={isVisible}>
+            <Box
+              fontSize="md"
+              color="text"
+              marginTop="2rem"
+              paddingTop="1.5rem"
```
pettinarip, line 144:
```suggestion
              marginTop="8"
              paddingTop="6"
```

pettinarip (changes requested):
Good job @jb-0. Left a couple of comments to simplify a bit the implementation & to be consistent with the codebase style.

src/components/ExpandableCard.tsx
```diff
@@ -107,111 +34,114 @@ const ExpandableCard: React.FC<IProps> = ({
...
+          onClick={onClick}
         >
-          <Text
-            variants={fadeInOut}
-            animate={isVisible ? "expanded" : "collapsed"}
-            initial={false}
+          <Box
+            fontSize="md"
+            color="text"
+            marginTop="8"
```
pettinarip, line 135:
```suggestion
```

pettinarip (approved):
lgtm! just removed an extra margin to make it look a bit nicer. Good job @jb-0

jb-0:
> lgtm! just removed an extra margin to make it look a bit nicer. Good job @jb-0 

Nice one, thanks for your help!

PR #8575, opened 2022-11-11
Use latest value from Beaconcha.in endpoint

wackerow:
## Description
- Fixes the endpoint URL for pulling latest Beaconcha.in metric related to staking APR.
- "/650" was pulling data from 650 days after Beacon Chain genesis, not the latest
- Removes old endpoint for Total ETH Staked and Total Validators, using new endpoint for these as well to consolidate API calls into one.
- Specifies ETH.STORE as the source for the APR calculation.

## Related Issue
Noted by @Buttaa offline 🙏

### DIFF
--- a/src/components/Staking/StakingStatsBox.tsx
+++ b/src/components/Staking/StakingStatsBox.tsx
@@ -16,6 +16,7 @@ import { getLocaleForNumberFormat } from "../../utils/translations"
 // Constants
 const NA_ERROR = "n/a"
 const ZERO = "0"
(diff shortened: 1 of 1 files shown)


wackerow:
Also noted was that we can fetch all the existing data needed from this same endpoint, and don't need the existing endpoint at all. Latest commit updates to use the total ETH staked from the new endpoint, and calculate the total validators from this.

Also updates to specify ETH.STORE as the source for the APR calculation.

corwintines (approved):

PR #8579, opened 2022-11-12
Update `DocLink` component (Post Chakra)

TylerAPfledderer:
## Description
- Remove `Box` wrapping all components and send its props to `LinkBox`
- Remove `position='relative'` from a `Flex` component that preventing the whole component from being clickable.
- Apply `useToken` to use a theme token in the `boxShadow` prop instead of a CSS Variable
- Replace one prop value with a spacing token
- Remove `role="group"` as it is invalid to the accessibility of this component (it's a single item, not a group of items)

## Related Issue
N/A

### DIFF
--- a/src/components/DocLink.tsx
+++ b/src/components/DocLink.tsx
@@ -26,7 +26,6 @@ const DocLink: React.FC<IProps> = ({ to, children, isExternal = false }) => {
       }}
     >
       <Flex
-        position="relative"
         zIndex={1}
         display="flex"
         flexDirection="row"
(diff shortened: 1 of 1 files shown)


pettinarip (approved):
Awesome! thanks @TylerAPfledderer 🚀

src/components/DocLink.tsx
```diff
@@ -11,62 +19,60 @@ export interface IProps {
...
+      padding={4}
+      borderRadius="sm"
+      color="text"
+      border="1px"
+      borderStyle="solid"
+      borderColor="border"
+      _hover={{
+        background: "tableBackgroundHover",
+        borderRadius: "base",
+        boxShadow: `0 0 1px ${linkBoxShadowColor}`,
```
pettinarip, line 35: I like a lot more this way than using the css var

PR #8585, opened 2022-11-14
i18n lib change - use `htmr` to parse html inside json files

pettinarip:
This PR removes the usage of `dangerouslySetInnerHTML` (for security reasons) to insert html content from the translation files. Instead, it uses the [htmr](https://github.com/pveyes/htmr) lib where we can also use our own custom components (similar to we do with markdown files).

## Description

- Revert the usage of `dangerouslySetInnerHTML`
- Use [`htmr`](https://github.com/pveyes/htmr) to parse the inner html that json files may have
- Use our own custom components in the parsing process

## Related Issue

#8345


src/components/Translation.tsx
```diff
@@ -1,12 +1,26 @@
 import React from "react"
```
pettinarip, line 1: Main change is in this file.

pettinarip:
Note: build is failing bc of this issue #8589

pettinarip:
Closing this PR since it got super messed up xD

Issue #8587, opened 2022-11-14
Remove FTX and FTX US


corwintines:
We should remove FTX from the get-eth page as they are no longer a functional exchange, and anyone interacting with that exchange is at risk of losing their assets.

Paused withdrawals
- https://ftx.com/

Bankrupt filing
- https://twitter.com/FTX_Official/status/1591071832823959552

minimalsm:
To clarify, we should remove FTX and FTX US, they are two different organisations, but both have filed for bankruptcy and are not accepting withdrawals or deposits (i.e. not acting as an exchange, and not helpful to users who are trying to get ETH).

PR #8588, opened 2022-11-14
remove ftx from get-eth

corwintines:
## Description
- Removes ftx from get-eth page, and from exchange data file

## Related Issue
- Fixes: https://github.com/ethereum/ethereum-org-website/issues/8587


src/components/EthExchanges.tsx
```diff
@@ -317,12 +313,6 @@ const EthExchanges = () => {
       easycrypto: file(relativePath: { eq: "exchanges/easycrypto.png" }) {
         ...cardListImage
       }
-      ftx: file(relativePath: { eq: "exchanges/ftx.png" }) {
```
samajammin, line 320: We should delete these asset files as well.

minimalsm (approved):

PR #8592, opened 2022-11-14
i18n lib change - use htmr to parse html inside json files #8585

pettinarip:
This PR removes the usage of `dangerouslySetInnerHTML` (for security reasons) to insert html content from the translation files. Instead, it uses the [htmr](https://github.com/pveyes/htmr) lib where we can also use our own custom components (similar to what we do with markdown files).

## Description

- Revert the usage of `dangerouslySetInnerHTML`
- Use [`htmr`](https://github.com/pveyes/htmr) to parse the inner html that json files may have
- Use our own custom components in the parsing process

## Related Issue

#8345

### DIFF
--- a/src/components/Translation.tsx
+++ b/src/components/Translation.tsx
@@ -1,12 +1,26 @@
 import React from "react"
-import { Trans } from "gatsby-plugin-react-i18next"
+import { useTranslation } from "gatsby-plugin-react-i18next"
+import htmr from "htmr"
 
-interface Props extends React.ComponentProps<typeof Trans> {
+import Link from "./Link"
+
+interface Props {
   id: string
+}
+
+const transform = {
+  a: Link,
 }
 
 // Wrapper on <FormattedHTMLMessage /> to always fallback to English
 // Use this component for any user-facing string
-const Translation = ({ id, ...rest }: Props) => <Trans i18nKey={id} {...rest} />
+const Translation = ({ id, ...rest }: Props) => {
+  const { t } = useTranslation()
+
+  const translatedText = t(id)
+
+  // @ts-ignore
+  return <>{htmr(translatedText, { transform })}</>
+}
 
 export default Translation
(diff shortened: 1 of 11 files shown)


pettinarip:
The last build (triggered manually) was successful. https://www.gatsbyjs.com/dashboard/5e02f963-3fe5-4e37-a1e7-ac05f7753a86/sites/cd906ab6-d953-4705-ba62-6d4128ac2d8a/pull-requests/84607095-38e5-445a-b917-7942e45fa276/builds

src/components/Translation.tsx
```diff
@@ -1,12 +1,26 @@
...
+
+interface Props {
   id: string
 }
 
+const transform = {
+  a: Link,
+}
+
 // Wrapper on <FormattedHTMLMessage /> to always fallback to English
```
samajammin, line 15: Seems this explanation is now out of date (given we swapped i18n libs)? Does it still fallback to English?

src/components/Translation.tsx
```diff
@@ -1,12 +1,26 @@
...
 
-interface Props extends React.ComponentProps<typeof Trans> {
+import Link from "./Link"
+
+interface Props {
   id: string
 }
 
+const transform = {
+  a: Link,
```
samajammin, line 12: Is this the only element we'd want a custom component for?

samajammin (commented):
A couple small questions but overall this seems like a much cleaner way to accomplish this!

src/components/Translation.tsx
pettinarip: Correct. Yea, I'll update that comment. And yes, it fallback to English.

src/components/Translation.tsx
pettinarip: So far I haven't seen any other html element on the translation files where we would want to use a custom React component.

corwintines (approved):

PR #8596, opened 2022-11-15
fix type

LuozhuZhang:
## Description

Nakamoto consensus is strictly a "protocol" rather than a "mechanism".
Refer to the definition of consensus mechanism on [ethereum.org](https://ethereum.org/en/developers/docs/consensus-mechanisms/#what-is-a-consensus-mechanism).

### DIFF
--- a/src/content/developers/docs/consensus-mechanisms/pow/index.md
+++ b/src/content/developers/docs/consensus-mechanisms/pow/index.md
@@ -16,7 +16,7 @@ To better understand this page, we recommend you first read up on [transactions]
 
 ## What is Proof-of-work (PoW)? {#what-is-pow}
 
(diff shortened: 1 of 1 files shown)


minimalsm:
Asked for some feedback on this:

> Generally has to be "mechanism" for ethereum's proof of stake consensus because there are critical components of that are explicitly out of protocol - e.g. out of band social coordination, weak subjectivity etc. "Protocol" refers just to the bits coded into clients. 
> 
> Nakamoto consensus is basically pow plus the longest chain rule - you could implement that in several distinct ways. I would think the protocol is the specific implementation.

I tend to agree. This isn't the first time this has come up: [see prior context](https://github.com/ethereum/ethereum-org-website/pull/4371#issuecomment-1061019487)

I think we should stick to our guns on this one and keep this as mechanism.

minimalsm:
@all-contributors please add @LuozhuZhang for ideas

PR #8601, opened 2022-11-16
Convert `Pill` component to using Chakra `Badge` Component

### LINKED ISSUE #8551: Replace `Pill` component with the Chakra `Badge` component

Currently we are using [a custom component (`Pill`)](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Pill.tsx) to display a badge like component.

Since we are migrating over using Chakra components, we need to replace this custom component with the Chakra one.

`Badge` component https://chakra-ui.com/docs/components/badge/usage

1. Identify all the different variants we have across the site
2. Create custom theme for the `Badge` component under https://github.com/ethereum/ethereum-org-website/tree/dev/src/%40chakra-ui/gatsby-plugin/components
3. Remove the custom `Pill` component
4. Replace all the uses of `Pill` with the `Badge` component

None.

### Additional context

_No response_

TylerAPfledderer:
## Description

Replaces the custom `Pill` component with Chakra UI's `Badge` component

- Creates custom theming
- Applies replacements in compound components, pages, and templates.
- Takes pill-like child components and applies `Badge` while retaining logic specific to these components

## Related Issue
Closes #8551


TylerAPfledderer:
Hey @pettinarip I am running into a pain point consolidating styles to a custom `Badge` theme config.

I listed all the styles that are being used in each component I see that would use a badge (minus general overrides from the default badge theming)

Check out what I am working with so far in the custom theming.
code

I can consolidate padding to base styles and `Sizes` options, but I can't easily consolidate the rest to just one variant without leaving some duplication (border and borderColor in particular). 

In other words, I would naturally want have `color` and `textTransform` as a variant, but the borderColor is not always applied where these other two props are needed.

Maybe if I can allow for the basic border styling to be applied by default to all applicable components, then I can make a more manageable variant and base to apply.

Is this OK here? It should be a minimal impact and I believe this is more about keeping from any potential confusion in abstracting away styling.

What do you see that can make further consolidation possible? I understand that this might be handled differently than how some secondary styling has been applied, and we need to consider time needed to be spent versus awaiting the new DS that should alleviate any issue like this.

src/@chakra-ui/gatsby-plugin/components/Badge.ts
```diff
@@ -0,0 +1,34 @@
...
+  baseStyle: {
+    borderRadius: "base",
+    border: "1px solid",
+    borderColor: "transparent",
+    color: "black300",
+    fontWeight: "initial",
+    py: 1,
+    px: 2,
+    textTransform: "uppercase",
+  },
```
TylerAPfledderer, line 18: To be able to consolidate styles for instances that include a background color, it was a choice between either ensuring that all instances also included the "primary" border color (which all did not), or not include it at all. I chose not to include it here.

Having the minor border in all instances is not showing to impact the layouts.

TylerAPfledderer:
@pettinarip I believe this is good to go for your review! 🤘🏼

src/templates/docs.tsx
```diff
@@ -41,6 +40,7 @@ import { ZenModeContext } from "../contexts/ZenModeContext"
 import { isLangRightToLeft } from "../utils/translations"
 import { Lang } from "../utils/languages"
 import { Context } from "../types"
+import { Badge } from "@chakra-ui/react"
```
pettinarip, line 43: import order

src/pages/layer-2.tsx
```diff
@@ -40,6 +39,7 @@ import { getImage } from "../utils/image"
 
 // Constants
 import { GATSBY_FUNCTIONS_PATH } from "../constants"
+import { Badge } from "@chakra-ui/react"
```
pettinarip, line 42: import order

src/pages/layer-2.tsx
```diff
@@ -773,7 +773,9 @@ const Layer2Page = ({ data }: PageProps<Queries.Layer2PageQuery>) => {
                   tokenLists={l2.tokenLists}
                 >
                   {l2.purpose.map((purpose, index) => (
-                    <Pill key={index}>{purpose}</Pill>
+                    <Badge key={index} me={2}>
+                      {purpose}
+                    </Badge>
```
pettinarip, line 778: These badges for some reason don't have a background color.
image

Should the default variant be transparent? I'd expect for this iteration to have the same default color that the Chakra Badge uses.

src/@chakra-ui/gatsby-plugin/components/Badge.ts
```diff
@@ -0,0 +1,34 @@
...
+    sm: {
+      py: 0,
+    },
+    lg: {
+      px: 3,
+    },
+  },
+  defaultProps: {
+    // Remove the default from Chakra
+    variant: "",
```
pettinarip, line 32: I'm wondering if we should have a default variant. IMO we should, since that is what we implicitly have right now on the designs. I think we should give it a name and put some defaults like the default background color.

I think that that variant should be the `solid` variant, having something like this:

```tsx
baseStyle: {
  borderRadius: "base",
  border: "1px solid",
  borderColor: "transparent",
  fontWeight: "initial",
  py: 1,
  px: 2,
  textTransform: "uppercase",
},
variants: {
  solid: {
    color: "black300",
    backgroundColor: "gray.300",
  },
  ...
},
defaultProps: {
  variant: "solid"
}
```

LMK what do you think.

Think should also fix the transparent bg we have on the l2 page:
image

pettinarip (changes requested):
As far as I tested, this is working really nice. Thanks for the significant refactor :) pills everywhere

TylerAPfledderer:
@pettinarip should be good for another check.

Found another styling discrepancy in the product cards on the developers learning tools page 👍

pettinarip (approved):
🚀

pettinarip (commented):
Wondering why I see the `StablecoinAccordion` changes in this PR 🤔 they were merged on #9004

PR #8606, opened 2022-11-16
Fix feedback widget copy

minimalsm:
## Description

On the feedback widget we switch tenses, "Was this page helpful?" slides out, then the pop-up shows "Is this page helpful?". This PR fixes that so that 'Is this page helpful?' is used consistently on the feedback widget.

https://user-images.githubusercontent.com/62268199/202189898-75371f9c-77c8-4312-98a2-06ec3a528e9e.mov

### DIFF
--- a/src/components/FeedbackWidget.tsx
+++ b/src/components/FeedbackWidget.tsx
@@ -171,7 +171,7 @@ const FeedbackWidget: React.FC<FeedbackWidgetProps> = ({ location = "" }) => {
                 alignItems="center"
                 display={{ base: "none", lg: isExpanded ? "flex" : "none" }}
               >
-                <Translation id="feedback-card-prompt-page" />
+                <Translation id="feedback-widget-prompt" />
               </Text>
             </ScaleFade>
           )}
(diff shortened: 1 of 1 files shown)


pettinarip (approved):

wackerow:
@minimalsm @konopkja  One other suggestion I had made would be to "animate out" the expanded button once it is pressed (to avoid having the button expanded + the modal showing at the same time). Any thoughts on including this? Could always address it separately as well, in which case go ahead and pull this one it =).

konopkja:
> @minimalsm @konopkja One other suggestion I had made would be to "animate out" the expanded button once it is pressed (to avoid having the button expanded + the modal showing at the same time). Any thoughts on including this? Could always address it separately as well, in which case go ahead and pull this one it =).

yes agree this would be better

minimalsm:
> @minimalsm @konopkja One other suggestion I had made would be to "animate out" the expanded button once it is pressed (to avoid having the button expanded + the modal showing at the same time). Any thoughts on including this? Could always address it separately as well, in which case go ahead and pull this one it =).

[Done here](https://github.com/ethereum/ethereum-org-website/pull/8606/commits/b840cc4e8aa621624cd65fbbf51e422ead2f258e)

minimalsm:
@wackerow @konopkja merging this. Feel free to loop back on the design question here after seeing its implementation.

PR #8629, opened 2022-11-17
Update markdown checker script

minimalsm:
## Description

Updates the markdown checker script:

- ensure skill level matching either 'beginner', 'intermediate', or 'advanced'
- ensure sidebar props are not present

### DIFF
--- a/src/scripts/markdown-checker.js
+++ b/src/scripts/markdown-checker.js
@@ -111,6 +111,10 @@ function processFrontmatter(path, lang) {
     )
   }
 
+  if (frontmatter.sidebar) {
+    console.error(`Unexpected 'sidebar' frontmatter at ${path}`)
+  }
+
   if (path.includes("/tutorials/")) {
     if (!frontmatter.published) {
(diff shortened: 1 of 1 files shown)


src/scripts/markdown-checker.js
```diff
@@ -130,6 +134,16 @@ function processFrontmatter(path, lang) {
         )
       }
     }
+
+    if (
+      frontmatter.skill !== "beginner" &&
+      frontmatter.skill !== "intermediate" &&
+      frontmatter.skill !== "advanced"
+    ) {
```
wackerow, line 142:
```suggestion
    if (!["beginner", "intermediate", "advanced"].includes(frontmatter.skill)) {
```
To trim down and limit repeated code =)

wackerow: One step further, could declare those strings as constant variables like
```ts
const BEGINNER = "beginner"
...
```
then reuse those below in the console log to ensure accuracy of the terms.

PR #8641, opened 2022-11-17
Breadcrumbs.tsx Refactor

Master7130:
## Description

Refactored Breadcrumbs.tsx from styled components to ChakraUI.

## Related Issue

#6374


src/components/Breadcrumbs.tsx
```diff
@@ -91,21 +50,51 @@ const Breadcrumbs: React.FC<IProps> = ({
...
-            <CrumbLink
+          <ListItem
+            key={idx}
+            margin={0}
+            marginRight={2}
+            fontSize="sm"
+            lineHeight="140%"
+            letterSpacing="wider"
+          >
+            <ButtonLink
```
TylerAPfledderer, line 78: This should just be the `Link` component, as the crumbs only need to be links and not also button. (There is also the solid background from the default button styling that should not be there)

TylerAPfledderer (changes requested):
Hey @Master7130!

Nice PR here, I have left only one comment with the changes you have made. 😄 

However, I believe this component can use Charka's [BreadCrumb multi-component]() and retain this component to provide additional needed logic.

The only tiny gotcha from what I can see is that the Chakra version currently deployed in this project is not up-to-date and therefore the `BreadCrumb` will not have the `listProps` prop for immediate access to the list element rendered for styling. So the workaround would be the following to get rid of the left margin from the global styles:
```tsx
<BreadCrumb
  sx={{
    ms: 0
  }}
/>
```

I would also like @pettinarip's feedback on this recommendation.

pettinarip:
Good suggestion @TylerAPfledderer. I agree with you, we should use the Chakra's component. However, I think we can merge this PR for now.

I've created a [new issue](https://github.com/ethereum/ethereum-org-website/issues/8772) to tackle the replacement. I haven't assigned anyone yet. Please, leave a comment on the issue if someone is interested in working on that. Thanks 🙏🏼

PR #8652, opened 2022-11-17
Update Khmer (km) homepage, and Essentials buckets

minimalsm:
## Description

Update Khmer (km) homepage, and Essentials buckets


minimalsm:
LGTM

Issue #6590, opened 2022-06-07
Update Sharding page


samajammin:
**Is your feature request related to a problem? Please describe.**

Our sharding page is outdated - we should revamp this to reflect the latest roadmap.

**Describe the solution you'd like**

  - Explain iterative approach (danksharding / protodanksharding)

Potential resources to link out to
  - [https://www.eip4844.com/](https://www.eip4844.com/)
  - [https://members.delphidigital.io/reports/the-hitchhikers-guide-to-ethereum](https://members.delphidigital.io/reports/the-hitchhikers-guide-to-ethereum)

**Describe alternatives you've considered**

For a quick fix (if we don't have bandwidth for a full rehaul of the page), we could keep the page content the same & just add a banner/disclaimer at the top, e.g. 

> "Execution Sharding has been deprioritized from the Ethereum roadmap in favour of Rollups [link to rollups page]. An update which refocuses this page on data sharding is expected in Q3" 

Hat tip to @timbeiko for the suggestion 😄 

**Additional context**

Somewhat related to #6523

samajammin:
@wackerow assigning to you for now but anyone: feel free to comment if you'd like to help out with this!

minimalsm:
Just gotta link to EIP-4488 now?

samajammin:
Removing stale label.

> Just gotta link to EIP-4488 now?

This would definitely be a great short-term fix but I do think the content of this page needs a overhaul. It's significantly outdated & doesn't reflect the latest roadmap.

Rnamita:
I would like to help with this issue can you please assign it to me?

ossayed:
Hi there I'd love to work on this issue!

corwintines:
I think we can close this issue out after the roadmap work done by @jmcook1186

PR #8656, opened 2022-11-17
Add sharding resources [#6590]

samajammin:
## Description

Adds a few sharding resources

## Related Issue

Related to #6590 but I don't think it solves it entirely.


minimalsm (approved):
Thanks @samajammin 👌 I agree with adding these links but want to note that because we don't mention protodanksharding//eip-4844 on this page, we now have further reading links that have absolutely zero context on the page.

We should prioritise revamping this page in Q1.

Issue #8632, opened 2022-11-17
Implement UI library - 2nd migration wave


pettinarip:
*This is part of our [Implement UI library](https://github.com/ethereum/ethereum-org-website/issues/6374) epic.*

This is our 2nd list of components to migrate all of our UI components to Chakra.

### **How to contribute**

If you’d like to help with this migration, please follow these steps:

1. Leave a comment here asking which file you would want to work on.
2. We assign you to the file (by adding your handle next to the file name in the list).
    - Read and follow our [UI migration guide](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/chakra-migration-guide.md) for some tips & guidance.
    - **If you were assigned to multiple files** then create a separate PR for each component. It will make the testing process much simpler and more organized.
3. Once you finish it (PR merged), we’ll mark it as done.
4. Repeat!

---

💡 Some of these components have some special requirements or considerations to take into account in the migration. Visit the issue attached to them to see the details.

- [x]  BannerGrid/
    - [x]  index.tsx @soheil555 #8800
- [x]  Banners/
    - [x]  PostMergeBanner.tsx @ImMiguelP #8845
- [x]  FindWallet/
    - [x]  WalletPersonasSidebar.tsx @soheil555 #8929
    - [x]  WalletFilterSidebar.tsx @TylerAPfledderer #9028
    - [x] #9982
- [x] Layer2/
    - [x] Layer2Onboard.tsx @TylerAPfledderer #10196
- [x]  Search/
    - [x] #8635 @amit-ksh 
- [x]  Staking/
    - [x]  StakingComparison.tsx @soheil555 #9046
    - [x]  StakingGuides.tsx @ImMiguelP #8821
    - [x]  StakingHierarchy.tsx @TylerAPfledderer #8686
    - [x]  StakingHomeTableOfContents.tsx @TylerAPfledderer #8703
    - [x]  StakingHowSoloWorks.tsx @ImMiguelP #8822
    - [x]  StakingProductsCardGrid.tsx @TylerAPfledderer #8679
    - [x]  StakingStatsBox.tsx @soheil555 #9017
    - [x] StakingCommunityCallout.tsx #10463
    - [x] StakingConsiderations.tsx @TylerAPfledderer #10203
    - [x] StakingLaunchpadWidget.tsx @TylerAPfledderer #10213
- [x]  BeaconChainActions.tsx @soheil555 #8720
[...]

PR #8678, opened 2022-11-20
Refactor: Migrate `TableOfContents` to Chakra

TylerAPfledderer:
## Description

- Migrates the `TableOfContents` component to Chakra UI
- Additionally fixes the `docs` template where a ternary check for the `DesktopTableOfContents` styled component needed to be applied as a Chakra prop for the given style

## Related Issue
#8632


src/components/TableOfContents.tsx
```diff
@@ -416,52 +335,96 @@ const TableOfContents: React.FC<IProps> = ({
...
+                  },
+                  "& .chakra-switch__thumb": {
+                    background: "ednBackground",
+                    outline: "2px solid",
+                    outlineColor: "secondary",
+                  },
+                }}
+                isChecked={isZenMode}
+                onChange={() => handleZenModeChange()}
+              />
```
TylerAPfledderer, line 409: Replace toggle icons in a `span` element with the Chakra [Switch](https://chakra-ui.com/docs/components/switch/usage) component for improved accessibility with a checkbox-like toggle. Creates a different sized element, readjusts space around it, and includes custom styling to closely mimic the original icons.

src/components/TableOfContents.tsx
```diff
@@ -322,50 +227,63 @@ const TableOfContentsMobile: React.FC<IPropsTableOfContentsMobile> = ({
...
+          as={MdExpandMore}
+          transform={isOpen ? "rotate(0)" : "rotate(-90deg)"}
+          boxSize={6}
+          transition="transform .4s"
+        />
+      </Flex>
+      <Fade
+        in={isOpen}
+        {...getDisclosureProps()}
+        transition={{ enter: { duration: 0.6 } }}
```
TylerAPfledderer, line 272: Removing the import of `motion` in favor of Chakra's `Fade` component. Since disclosure props are added to increase accessibility, collapse animation is different on close, though the transition duration is retained.

src/components/TableOfContents.tsx
```diff
@@ -322,50 +227,63 @@ const TableOfContentsMobile: React.FC<IPropsTableOfContentsMobile> = ({
...
-            },
-          },
-        }}
-        initial="closed"
+        </chakra.span>
+        <Icon
+          as={MdExpandMore}
+          transform={isOpen ? "rotate(0)" : "rotate(-90deg)"}
+          boxSize={6}
+          transition="transform .4s"
```
TylerAPfledderer, line 266: Simplify animation to a simple CSS transition.

src/templates/docs.tsx
```diff
@@ -60,12 +60,6 @@ const ContentContainer = styled.div<{ isZenMode: boolean }>`
   background-color: ${(props) => props.theme.colors.ednBackground};
 `
 
-const DesktopTableOfContents = styled(TableOfContents)<{
-  isPageIncomplete: boolean
-}>`
-  padding-top: ${(props) => (props.isPageIncomplete ? `5rem` : `3rem`)};
```
TylerAPfledderer, line 66: To restate from the PR description, this ternary used here will no longer work, and instead the `isPageIncomplete` prop on the component is replaced with the Chakra prop `pt` and place the ternary there.

src/components/TableOfContents.tsx
```diff
@@ -322,50 +227,63 @@ const TableOfContentsMobile: React.FC<IPropsTableOfContentsMobile> = ({
...
+      as="aside"
+      background="background"
+      border="1px"
+      borderColor="border"
+      borderRadius="4px"
+      py={2}
+      px={4}
+      /* TODO find better way - this accounts for huge header margin top */
+      /* but if doc DOESN'T start w/ header, it overlaps, e.g. /docs/accounts/  */
+      /* margin-bottom: -10rem; */
```
TylerAPfledderer, line 249: Does this comment still need to be here?

src/components/TableOfContents.tsx
```diff
@@ -258,14 +77,68 @@ const TableOfContentsLink: React.FC<IPropsTableOfContentsLink> = ({
   if (isNested) {
     classes += " nested"
   }
+
+  const $dotBg = cssVar("dot-bg")
```
pettinarip, line 81: Could you explain a bit about `cssVar`? how does this work?
Do you define it on `[$dotBg.variable]: "colors.background"` and then you can use the reference of that?

And wondering why you choose that name for it, shouldn't this be named `linkBg`?

src/components/TableOfContents.tsx
```diff
@@ -258,14 +77,68 @@ const TableOfContentsLink: React.FC<IPropsTableOfContentsLink> = ({
...
+      left: "-1.29rem",
+      top: "50%",
+      mt: -1,
+    },
+  }
+
   return (
-    <StyledTableOfContentsLink to={url} className={classes}>
+    <Link
+      as={GatsbyLink}
```
pettinarip, line 101: Could we use our custom `Link` component for this? `src/components/Link`

pettinarip (changes requested):
This is great @TylerAPfledderer 👏🏼 

Left a few comments but I would encourage you to create a folder for this component and break the multiple inner components out. I think that refactor could simplify the readability a lot. That is not a blocker for this PR, totally a nice to have or something we could also do in a different PR.

src/components/TableOfContents.tsx
TylerAPfledderer: You have the correct idea! `cssVar` takes a string to name the variable, and returns two strings to use:
1. The variable declaration (`--dot-bg`) as `$dotBg.variable` when setting a value to it. (i.e. `[$dotBg.variable]: <someValue>`)
2. The variable call (`var(--dot-bg)`)  as `$dotBg.reference` when retrieving the value (i.e. `backgroundColor: $dotBg.reference`)
> Aside: you can also supply a fallback value as a second argument, which is the same as the fallback in the `var()` function.

When calling a token name for CSS variables, you have to start with the name of the theme collection it belongs to, hence `colors.someToken`

I chose `dot-bg` as the name because it is for the background color of the dot that is rendered off to the left when hovering over the link or if the link is active, and is not applied to the link itself. Could make it `link-dot-bg` to make it clearer.

src/components/TableOfContents.tsx
TylerAPfledderer: It wasn't used originally, but I will go in and add it. No problem! 😄

src/components/TableOfContents.tsx
```diff
@@ -416,52 +335,96 @@ const TableOfContents: React.FC<IProps> = ({
...
-    <Aside className={className}>
-      <OuterList>
-        {!hideEditButton && (
-          <ButtonContainer>
-            <ButtonLink to={editPath} variant="outline" hideArrow mt={0}>
-              <ButtonContent>
-                <GithubIcon name="github" />{" "}
-                <span>
+    // TODO: Switch to `above="lg"` after completion of Chakra Migration
+    <Show breakpoint="(min-width: 1025px)">
```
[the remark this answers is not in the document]
TylerAPfledderer: In this case, the `Show` component is replacing a `display: none` prop that was used under the [lg breakpoint](https://github.com/ethereum/ethereum-org-website/blob/cb5f4cfdf95c6d8e7b6ba6ce01a6ec3efae3d84b/src/components/TableOfContents.tsx#L28-L30) and did not have an `isMobile` conditional check.

Would you say then that there should be `if / else` statement instead to control which layout is shown with `isMobile`?

src/components/TableOfContents.tsx
```diff
@@ -416,52 +335,96 @@ const TableOfContents: React.FC<IProps> = ({
...
+            <Flex
+              as={ListItem}
+              alignItems="center"
+              mb={2}
+              py="2px"
+              opacity={0.8}
+              fontSize="sm"
+            >
+              <Box as="span" me={2} id="zen-mode">
+                <Translation id="zen-mode" />
```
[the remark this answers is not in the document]
TylerAPfledderer: Yes! That's a simple addition.

src/components/TableOfContents.tsx
```diff
@@ -416,52 +335,96 @@ const TableOfContents: React.FC<IProps> = ({
   }
   if (isMobile) {
     return (
-      <TableOfContentsMobile
-        items={items}
-        maxDepth={maxDepth}
-        className={className}
-      />
+      // TODO: Switch to `below="lg"` after completion of Chakra Migration
+      <Show breakpoint="(max-width: 1024px)">
```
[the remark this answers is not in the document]
TylerAPfledderer: @pettinarip Right. This is an oversight on me! I'll remove this.

src/components/TableOfContents.tsx
```diff
@@ -322,50 +227,63 @@ const TableOfContentsMobile: React.FC<IPropsTableOfContentsMobile> = ({
...
+            maxDepth={maxDepth ? maxDepth : 1}
+          />
+        </List>
+      </Fade>
+    </Box>
   )
 }
 
-export interface IProps {
+export interface IProps extends ChakraProps {
```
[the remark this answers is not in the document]
TylerAPfledderer: `ChakraProps` is the basis of all the styling and pseudo prop types from the theme. `BoxProps` is pretty much the same thing in this context as it extends `ChakraProps`, but also extends the `As` prop type (not needed here) and gets all intrinsic HTML attributes associated with `Box`.

Both do not include and Chakra style prop shorthands (`direction` as a shorthand for `flexDirection`, for example). Again, not needed here.

So I can certainly switch out `ChakraProps` with `BoxProps` to be consistent with the parent-level component being used here and to obtain typing for html attributes should we apply them from the page level in the future. :)

TylerAPfledderer:
> This is great @TylerAPfledderer 👏🏼
> 
> Left a few comments but I would encourage you to create a folder for this component and break the multiple inner components out. I think that refactor could simplify the readability a lot. That is not a blocker for this PR, totally a nice to have or something we could also do in a different PR.

I'll go ahead and create a new folder.

src/components/TableOfContents.tsx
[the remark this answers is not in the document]
pettinarip: I understood this now! 😆 

Ok. We'll need this back. We are currently showing it on larger screens. My bad.
image

However, I would prefer that that is controlled by the consumer but nvm, things for a different PR/moment.

src/components/TableOfContents/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+            <Flex
+              as={ListItem}
+              alignItems="center"
+              mb={2}
+              py="2px"
+              opacity={0.8}
+              fontSize="sm"
+            >
+              <FormControl as={Flex} alignItems="center">
+                <FormLabel htmlFor="zen-mode" mb={0} me={2}>
```
pettinarip, line 127:
```suggestion
                <FormLabel htmlFor="zen-mode" mb={0} me={2} fontSize="sm">
```

src/components/TableOfContents/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+                  <Translation id="zen-mode" />
+                </FormLabel>
+                <Switch
+                  id="zen-mode"
+                  sx={{
```
pettinarip, line 132:
```suggestion
                  id="zen-mode"
                  size="sm"
                  sx={{
```

src/components/TableOfContents/TableOfContentsLink.tsx
```diff
@@ -0,0 +1,89 @@
...
+          },
+        },
+      }}
+    >
+      {trimmedTitle(item.title)}
+    </Link>
+  )
+}
+
+export default TableOfContentsLink
```
pettinarip, line 89: Lets remove the `TableOfContents` prefix now that we have a folder with that name. Rename the file as well.
```suggestion
export default Link
```

src/components/TableOfContents/TableOfContentsMobile.tsx
```diff
@@ -0,0 +1,79 @@
...
+            depth={0}
+            maxDepth={maxDepth ? maxDepth : 1}
+          />
+        </List>
+      </Fade>
+    </Box>
+  )
+}
+
+export default TableOfContentsMobile
```
pettinarip, line 79:
```suggestion
export default Mobile
```

src/components/TableOfContents/tableOfContents.utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    // TODO: Flip to object syntax with `lg` token after completion of Chakra migration
+    "@media (max-width: 1024px)": {
+      borderStart: 0,
+      borderTop: "1px",
+      borderTopColor: "primary300",
+      ps: 0,
+      pt: 4,
+    },
+  },
+}
```
pettinarip, line 58: rename file to `utils.ts`

pettinarip (changes requested):
Just a few super small requests. This is great, thanks for the big refactor here man.

TylerAPfledderer:
@pettinarip should be all good to go!

PR #8685, opened 2022-11-21
Refactor: Migrate `ProductCard` to Chakra

TylerAPfledderer:
## Description

Migrate the `ProductCard` component to Chakra.

## Related Issue
#8632

### DIFF
--- a/src/components/ProductCard.tsx
+++ b/src/components/ProductCard.tsx
@@ -1,5 +1,4 @@
 import React, { ReactNode } from "react"
-import styled from "@emotion/styled"
 import { GatsbyImage, IGatsbyImageData } from "gatsby-plugin-image"
 import { useQuery, gql } from "@apollo/client"
 
@@ -11,7 +10,6 @@ import {
   Flex,
   Heading,
   HStack,
-  Image,
   Img,
   Text,
   TextProps,
--- a/src/components/ProductCard.tsx
+++ b/src/components/ProductCard.tsx
@@ -5,118 +5,17 @@ import { useQuery, gql } from "@apollo/client"
 
 import GitStars from "./GitStars"
 import ButtonLink from "./ButtonLink"
-
-const ImageWrapper = styled.div<{
-  background: string
-}>`
-  display: flex;
-  flex-direction: row;
-  justify-content: center;
-  align-items: center;
-  background: ${(props) => props.background};
-  box-shadow: inset 0px -1px 0px rgba(0, 0, 0, 0.1);
-  min-height: 200px;
-`
-
-const Image = styled(GatsbyImage)`
-  width: 100%;
-  align-self: center;
-  max-width: 372px;
-  max-height: 257px;
-  @media (max-width: ${(props) => props.theme.breakpoints.s}) {
-    max-width: 311px;
-  }
-`
-
-const Card = styled.div`
-  color: ${(props) => props.theme.colors.text};
-  box-shadow: 0px 14px 66px rgba(0, 0, 0, 0.07);
-  display: flex;
-  flex-direction: column;
-  justify-content: space-between;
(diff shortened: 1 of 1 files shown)


src/components/ProductCard.tsx
```diff
@@ -156,7 +102,7 @@ export interface IProps {
 
 const ProductCard: React.FC<IProps> = ({
```
TylerAPfledderer, line 103: There is a TODO regarding adding loading state. What is the desired loading visual?

src/components/ProductCard.tsx
```diff
@@ -5,118 +5,17 @@ import { useQuery, gql } from "@apollo/client"
...
+  Box,
+  Center,
+  Flex,
+  Heading,
+  HStack,
+  Image,
+  Img,
+  Text,
+  TextProps,
+} from "@chakra-ui/react"
```
pettinarip, line 18: import order

src/components/ProductCard.tsx
```diff
@@ -139,6 +38,53 @@ const REPO_DATA = gql`
...
+      case "C#":
+        return "tagBlue"
+      case "Java":
+        return "tagPink"
+      default:
+        return "tagGray"
+    }
+  }
+  return (
+    <Box
```
pettinarip, line 72: These should be Badges in the upcoming refactor, right?

src/components/ProductCard.tsx
pettinarip: Hmm not sure 😆 wouldn't worry too much since that comment is 20 months old and probably we are going to refactor this with the DS.

pettinarip (changes requested):
Nice @TylerAPfledderer 

One small thing to look at. The pills/badges are not aligned:
image

src/components/ProductCard.tsx
TylerAPfledderer: Yes! I have made a note in #8601

src/components/ProductCard.tsx
TylerAPfledderer: @pettinarip curious to know if ESLint has ever been considered with a config that includes `import/order` to enforce a specific order of the different import types?

TylerAPfledderer:
> Nice @TylerAPfledderer
> 
> One small thing to look at. The pills/badges are not aligned: image

Whoops, didn't see that! On it.

src/components/ProductCard.tsx
pettinarip: That would be a useful addition, yes. Hasn't been discussed/considered before AFAIK.

src/components/ProductCard.tsx
```diff
@@ -186,46 +129,86 @@ const ProductCard: React.FC<IProps> = ({
...
+      <HStack mt={5} mb={2} px={6} spacing={3}>
         {subjects &&
           subjects.map((subject, idx) => (
             <SubjectPill key={idx} subject={subject}>
               {subject}
             </SubjectPill>
           ))}
         {hasRepoData &&
-          data.repository.languages.nodes.map(
+          data.repository.language.nodes.map(
```
pettinarip, line 200: Wondering why this has changed 🤔 

Currently I see an error breaking some pages. These two pages are loading but breaking after a second:
- https://ethereumorgwebsitedev01-tylerapfleddererethereumo08606.gatsbyjs.io/en/upgrades/get-involved
- https://ethereumorgwebsitedev01-tylerapfleddererethereumo08606.gatsbyjs.io/en/developers/local-environment/

pettinarip (changes requested):
Found a bug on some pages using this component.

pettinarip (approved):
Nicely done @TylerAPfledderer 🚀

PR #8714, opened 2022-11-22
fix incorrect info in glossary

jmcook1186:
## Description

The glossary claims that the zero address is the correct recipient for a contract creation transaction. This is repeated in the entries for `zero address` and `contract creation transaction`.

This info is not correct. The recipient for a contract creation transaction should be `null` not `0x0`. The former triggers contract creation at an address deterministically generated from the senders address/nonce, latter is a normal EOA that can receive ETH.

This PR corrects the info.

### DIFF
--- a/src/content/glossary/index.md
+++ b/src/content/glossary/index.md
@@ -205,7 +205,7 @@ An account containing code that executes whenever it receives a [transaction](#t
 
 ### contract creation transaction {#contract-creation-transaction}
 
-A special [transaction](#transaction), with the [zero address](#zero-address) as the recipient, that is used to register a [contract](#contract-account) and record it on the Ethereum blockchain.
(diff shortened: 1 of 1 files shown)


src/content/glossary/index.md
```diff
@@ -1095,7 +1095,7 @@ The smallest denomination of [ether](#ether). 10<sup>18</sup> wei = 1 ether.
 
 ### zero address {#zero-address}
 
-A special Ethereum address, composed entirely of zeros, that is specified as the destination address of a [contract creation transaction](#contract-creation-transaction).
+An Ethereum address, composed entirely of zeros, that is frequently used as a burn address for unwated funds.
```
ralexstokes, line 1098:
```suggestion
An Ethereum address, composed entirely of zeros, that is frequently used as a burn address for unwanted funds.
```

ralexstokes (approved):
nice!

minimalsm (approved):

PR #8734, opened 2022-11-24
Add CFI and EOF to glossary

Joeysantoro:
## Description

Add CFI and EOF as entries in the glossary.

CFI for its relevance around EIP and network upgrade discussions, and EOF given it is under significant development and several related EIPs for EOF are CFI.


corwintines:
Thanks @Joeysantoro!

Agree with adding CFI to the glossary. But I think EOF may be a bit premature given that hasn't been confirmed in a release yet. While EOF is a large topic in research, maintaining a glossary of these terms is a bit overwhelming. Ill make the changes. If you disagree feel free to open another issue/PR with EOF and we can discuss there :)

corwintines (approved):

PR #8744, opened 2022-11-27
Migrate `SimpleTable.tsx` to chakra

### LINKED ISSUE #8638: Migrate `SimpleTable.tsx` to Chakra
*This is part of our [2nd wave of migration](https://github.com/ethereum/ethereum-org-website/issues/8632) of the [Implement UI library](https://github.com/ethereum/ethereum-org-website/issues/6374) epic.*

**Things to consider**

- Check if we can use the [Chakra’s Table component](https://chakra-ui.com/docs/components/table/usage) for this

Check our [epic](https://github.com/ethereum/ethereum-org-website/issues/6374) & the [migration guide](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/chakra-migration-guide.md) for more information.

amit-ksh:
## Description

Migrate the `SimpleTable` component from emotion to chakra using Chakra's Table component.

## Related Issue

Closes #8638


src/components/SimpleTable.tsx
```diff
@@ -125,49 +28,93 @@ export interface IProps {
...
+          <Tr p={4}>
+            <Td colSpan={4}>
+              <Translation id="page-stablecoins-table-loading" />
+            </Td>
+          </Tr>
+        )}
+
+        {content.map(({ name, marketCap, image, type, url }, idx) => (
+          <Tr
+            as={url ? Link : Tr}
```
TylerAPfledderer, line 83: Supply `LinkBox` here if url is defined, but return `undefined` otherwise, as a cleaner alternative to a redundant `Tr`

```suggestion
            as={url ? LinkBox : undefined}
```

src/components/SimpleTable.tsx
```diff
@@ -125,49 +28,93 @@ export interface IProps {
...
+            }}
+            _focus={{
+              boxShadow: `0 0 1px ${textColor}`,
+              bg: "primary200",
+              color: "black300",
+            }}
+          >
+            <Td display="flex" alignItems="center">
+              {image && <Image src={image} alt="" mr={4} boxSize={6} />}
+              {name}
```
TylerAPfledderer, line 105: Provide `LinkOverlay` here and using the custom `Link` component instead of the Chakra version, with necessary styling overrides. (I like `inherit` here since most styling happens in the table components)

Also ensure that the overlay does not exist if `url` is undefined, else styling bugs can occur, or the anchor tag with no href will still exist (which is not accessible).
```suggestion
              { url ? (
                <LinkOverlay
                  as={Link}
                  hideArrow
                  to={url}
                  color="inherit"
                  textDecoration="inherit"
                  _hover={{
                    textDecoration: "inherit",
                  }}
                >
                  {name}
                </LinkOverlay>
              ) : (
                <>{name}</>
              )}
```

TylerAPfledderer (changes requested):
Hey @amit-ksh! 👋🏼 

I have a couple of suggestions to instantly make this more accessible with external links. 😄 (Using tables is an instant boost!

If we keep where the `Link` component is used for the row, then you will have an anchor tag rendered, wrapping `td` tags... which is not really approved in HTML styling validation, and might create issues with screen readers.

With the original intention of the whole row being clickable, the [LinkOverlay](https://chakra-ui.com/docs/components/link-overlay/usage) component tandem is a great approach!

@pettinarip I would like your thoughts here. 🤘🏼

amit-ksh:
Made all the changes as you said @TylerAPfledderer.

src/components/SimpleTable.tsx
amit-ksh: Yeah, I also thought of using the `LinkOverlay` component, but can't figure out how to use it in this scenario. Thanks for the help, I made all the changes as you said.

src/components/SimpleTable.tsx
```diff
@@ -125,49 +30,108 @@ export interface IProps {
...
+              color: "black300",
+              cursor: url ? "pointer" : "default",
+            }}
+            _focus={{
+              boxShadow: `0 0 1px ${textColor}`,
+              bg: "primary200",
+              color: "black300",
+            }}
+          >
+            <Td display="flex" alignItems="center">
```
pettinarip, line 105: I think we need to apply these styles to all the `Td`s since I don't see them aligned on the center.

image

src/components/SimpleTable.tsx
```diff
@@ -125,49 +30,108 @@ export interface IProps {
...
+          <Tr p={4}>
+            <Td colSpan={4}>
+              <Translation id="page-stablecoins-table-loading" />
+            </Td>
+          </Tr>
+        )}
 
+        {content.map(({ name, marketCap, image, type, url }, idx) => (
+          <Tr
+            as={url ? LinkBox : undefined}
```
pettinarip, line 85: @amit-ksh I think we need to do the inverse here.

```suggestion
          <LinkBox
            as={url ? Tr : undefined}
```

in order to render a `tr` element instead of a `div` (which is what the `LinkBox` renders).

image

I think that would fix the issue, right?

cc @TylerAPfledderer

pettinarip (changes requested):
Left a small question on how we are render the `LinkBox`.

Overall this is a great improvement on the a11y aspects. GJ

src/components/SimpleTable.tsx
pettinarip: Thats a great idea @TylerAPfledderer thanks for mentioning that!

src/components/SimpleTable.tsx
amit-ksh: In this case, `div` will appear if, the `url` is `undefined`
I think we should use like this  
```jsx
<LinkBox
    as={Tr}
```
This will definitely render `tr` and also work as a `LinkBox`

src/components/SimpleTable.tsx
amit-ksh: Done!

src/components/SimpleTable.tsx
pettinarip: Correct! good catch. Yeah I think we should always render a `Tr` here. Then if the `url` does not exist we don't show the `LinkOverlay` (something we are already doing) 👍🏼

amit-ksh:
Made all the changes as you suggested @pettinarip

src/components/SimpleTable.tsx
```diff
@@ -125,49 +31,111 @@ export interface IProps {
...
+              bg: "primary200",
+              color: "black300",
+              cursor: url ? "pointer" : "default",
+            }}
+            _focus={{
+              boxShadow: `0 0 1px ${textColor}`,
+              bg: "primary200",
+              color: "black300",
+            }}
+          >
```
TylerAPfledderer, line 105: A couple of props should be removed here, as they are ineffective or not needed.
`display`, `href`, and `textDecoration`
```suggestion
          <LinkBox
            as={Tr}
            key={idx}
            color="text"
            boxShadow={`0 1px 1px ${textColor}`}
            _hover={{
              textDecoration: "none",
              boxShadow: `0 0 1px ${textColor}`,
              bg: "primary200",
              color: "black300",
              cursor: url ? "pointer" : "default",
            }}
            _focus={{
              boxShadow: `0 0 1px ${textColor}`,
              bg: "primary200",
              color: "black300",
            }}
          >
```

src/components/SimpleTable.tsx
TylerAPfledderer: Yes this is fine, because `LinkBox` only really renders a `position: relative` which is used for `LinkOverlay`. :)

src/components/SimpleTable.tsx
```diff
@@ -125,49 +31,111 @@ export interface IProps {
...
+                    hideArrow
+                    to={url}
+                    color="inherit"
+                    textDecoration="inherit"
+                    _hover={{
+                      textDecoration: "inherit",
+                    }}
+                    isExternal
+                  >
+                    {name}
```
TylerAPfledderer, line 121: This is a personal recommendation, as it is a minor addition for A11y. (For @pettinarip approval 😄 )

Include visually hidden text for the `LinkOverlay` render only which reads `(Opens a new window)` (with the parens) since all the link that are ever passed into this table are assumed to be external.

> Maybe should have this under a translation function too and supplied in the `intl/en/common.json` file
```suggestion
                    {name}
                    <VisuallyHidden>
                      <Translation id="sr-link-opens-new-window" />
                    </VisuallyHidden>
```

And the json add:
`"sr-link-opens-new-window": "(Opens a new window)"`

TylerAPfledderer (changes requested):
@amit-ksh and @pettinarip: I've provided additional thoughts. :)

src/components/SimpleTable.tsx
amit-ksh: Thanks for the suggestions @TylerAPfledderer! 
I think we can also remove the `textDecoration` props from the hover style.

src/components/SimpleTable.tsx
amit-ksh: Hey @TylerAPfledderer, should I make these changes or wait for @pettinarip approval?

TylerAPfledderer: @amit-ksh I would wait to make sure it's OK since it is adding to the structure.

pettinarip: I think this is a worthy discussion we can have with the rest of the team on a different issue since I think that if we apply that change we should do the same with the rest of the external links we have on the site, to keep consistency. For that reason, I don't see that change as part of the scope of this PR/migration.

Happy to see this issue created and discussed with more people. I haven't seen this pattern a lot on other sites. I'm curious to hear what others think about it.

pettinarip (approved):
@amit-ksh great job! thanks for the refactor 💪🏼

PR #8752, opened 2022-11-28
Use systemic rather than systematic

aelmanaa:
## Description

The right word is systemic

- systematic = methodical  (= there is a plan/structure in place) 
- systemic = affects every part of an entire system


minimalsm:
@all-contributors please add @aelmanaa for docs

PR #8762, opened 2022-11-28
Add missing line continuation for lighthouse launch command

qiuhaohao:
## Description
Add missing line continuation for lighthouse launch command. The multi-line command does not have line continuation for the first line and therefore would fail.


corwintines (approved):
Nice catch! Thanks @qiuhaohao

corwintines:
@all-contributors can you add @qiuhaohao for content

PR #8788, opened 2022-11-29
Updated remix-deploy to reflect correct Enviroment

tconnolly11:
Changed references from "Javascript VM" to the currently supported "Remix VM (London)" show in Remix


corwintines (changes requested):
Thanks @tconnolly11! 

The image below this also needs to be changed it looks like.

Also, could export these to .png for the repo so that we aren't linking out to something that could break?

Thanks :)

corwintines:
Going to mark close this as abandoned. If you want to come back and resolve the comments @tconnolly11 feel free to reopen then :)

PR #8815, opened 2022-12-02
Replace Kanye/Yeezy reference NFT with Nike/Jordans

nuclearfalcon:
## Description
Kanye West has recently made many antisemitic comments and has [taken pro-nazi stances publicly](https://www.mediamatters.org/infowars/rapper-ye-goes-unhinged-antisemitic-rant-infowars). Earlier in his career the examples on this page were about Kanye the artist and the popular brand of shoes (Yeezys) that he created and marketed. Now Kanye is much more divisive and is a more dangerous name to use as an example. His brand of shoes has also been dropped by Adidas, so, while it is possible he takes his brand to Ethereum next, we can do better than to associate Ethereum with someone who spouts hate so openly.

When people search to learn about NFTs and land on ethereum.org, for many seeing Kanye's name and brand as examples might turn them off of the rest of the educational content that the page has to offer.

I propose we use a generic example and have replaced the text with one in this PR.

### DIFF
--- a/src/content/nft/index.md
+++ b/src/content/nft/index.md
@@ -14,7 +14,7 @@ summaryPoint3: Powered by smart contracts on the Ethereum blockchain.
 
 NFTs are currently taking the digital art and collectibles world by storm. Digital artists are seeing their lives change thanks to huge sales to a new crypto-audience. And celebrities are joining in as they spot a new opportunity to connect with fans. But digital art is only one way to use NFTs. Really they can be used to represent ownership of any unique asset, like a deed for an item in the digital or physical realm.
 
-If Andy Warhol had been born in the late 90s, he probably would have minted Campbell's Soup as an NFT. It's only a matter of time before your favorite brand has a run of NFTs on Ethereum. And one day owning your car might be proved with an NFT.
+If Andy Warhol had been born in the late 90s, he probably would have minted Campbell's Soup as an NFT. It's only a matter of time before Nike puts a run of Jordans on Ethereum. And one day owning your car might be proved with an NFT.
 
 ## What's an NFT? {#what-are-nfts}
 
--- a/src/content/nft/index.md
+++ b/src/content/nft/index.md
@@ -14,7 +14,7 @@ summaryPoint3: Powered by smart contracts on the Ethereum blockchain.
 
 NFTs are currently taking the digital art and collectibles world by storm. Digital artists are seeing their lives change thanks to huge sales to a new crypto-audience. And celebrities are joining in as they spot a new opportunity to connect with fans. But digital art is only one way to use NFTs. Really they can be used to represent ownership of any unique asset, like a deed for an item in the digital or physical realm.
 
-If Andy Warhol had been born in the late 90s, he probably would have minted Campbell's Soup as an NFT. It's only a matter of time before Kanye puts a run of Yeezys on Ethereum. And one day owning your car might be proved with an NFT.
+If Andy Warhol had been born in the late 90s, he probably would have minted Campbell's Soup as an NFT. It's only a matter of time before your favorite brand has a run of NFTs on Ethereum. And one day owning your car might be proved with an NFT.
 
 ## What's an NFT? {#what-are-nfts}
 


MicahZoltu:
While I agree with the idea of removing pop culture references from the website because they require too much maintenance to maintain, I think we should do a much more thorough pass over the whole site rather than handling them one-off like this.  For example, in this exact same line there is a reference to Andy Warhol that should be swapped out for something culturally neutral (are people in Africa going to get the reference?).

nuclearfalcon:
> While I agree with the idea of removing pop culture references from the website because they require too much maintenance to maintain, I think we should do a much more thorough pass over the whole site rather than handling them one-off like this. For example, in this exact same line there is a reference to Andy Warhol that should be swapped out for something culturally neutral (are people in Africa going to get the reference?).

A more thorough pass makes sense. I am not a fan of perfect being the enemy of good. I'm also not completely against pop culture references in explainers. Using fitting examples help people relate their learnings to their own life. 

However, when we see references that can be damaging to the Ethereum brand we should remove them, as I'm proposing here in my PR.

MicahZoltu:
I don't think Ethereum should get into the game of trying to decide who is a "good" pop culture reference and who is a bad one.  There is way too much politics involved in that.  If we do decide to stick with the pop-culture references in documentation, then I think we should just pick someone during initial writing and stick with it regardless of whether the person comes into or out of favor with the general public.  Otherwise we'll have never-ending debate over which references are socially acceptable today, and the results may not be valid tomorrow.

nuclearfalcon:
Changing our examples costs near nothing. This change specifically is removing a reference and not replacing it with a new pop culture reference, so should not be a problem. 

Ethereum isn't a static technology. We are already open to change in the core protocol, why hold ourselves to a different standard for the examples and educational content?

Kanye used to be a musician with a brand of shoes, now he is a political figure saying that a war criminal was a good person. I'm personally open to change and when I can avoid hateful rhetoric or can remove a hateful icon from our ecosystem's public documentation, I will take the steps to do so. I think we should merge this change and do a larger overhaul as the community sees fit with other proposals.

MicahZoltu:
> Changing our examples costs near nothing. This change specifically is removing a reference and not replacing it with a new pop culture reference, so should not be a problem.

The cost here isn't in the code change, it is in the political debate that ensues when someone tries to change an example for political reasons.  My argument here is that we shouldn't allow these debates to happen at all, because they will become a huge time sink.  To avoid this, we can either "set and forget" (don't change examples based on current events), or we can choose not to use any humans in any examples.  The path I think we should not go down is the one where we try to decide which example people are "good" and which are "bad" based on something they said recently.

nuclearfalcon:
I still attest that perfect should not be the enemy of good and this change doesn't perfectly fix all examples into generalized versions (which is the direction that you seem to be proposing in your first comment) but it does get us a step closer and its inclusion would benefit the community in the other ways expressed in the description of the PR.

MicahZoltu:
> its inclusion would benefit the community in the other ways expressed in the description of the PR.

I suspect not everyone would agree with this assertion, which is why I don't think we should make any changes based on political reasoning.  I *could* steel man a political argument against the specific removal you have proposed, but I really think we should avoid getting into such a discussion.  Even if the majority agree with your sentiment regarding this particular individual, it sets a precedent of it being appropriate to submit politically motivated changes to the website, rather than focusing on technical/UX.

src/content/nft/index.md
```diff
@@ -14,7 +14,7 @@ summaryPoint3: Powered by smart contracts on the Ethereum blockchain.
 
 NFTs are currently taking the digital art and collectibles world by storm. Digital artists are seeing their lives change thanks to huge sales to a new crypto-audience. And celebrities are joining in as they spot a new opportunity to connect with fans. But digital art is only one way to use NFTs. Really they can be used to represent ownership of any unique asset, like a deed for an item in the digital or physical realm.
 
-If Andy Warhol had been born in the late 90s, he probably would have minted Campbell's Soup as an NFT. It's only a matter of time before Kanye puts a run of Yeezys on Ethereum. And one day owning your car might be proved with an NFT.
+If Andy Warhol had been born in the late 90s, he probably would have minted Campbell's Soup as an NFT. It's only a matter of time before your favorite brand has a run of NFTs on Ethereum. And one day owning your car might be proved with an NFT.
```
minimalsm, line 17:
```suggestion
If Andy Warhol had been born in the late 90s, he probably would have minted Campbell's Soup as an NFT. It's only a matter of time before Nike puts a run of Jordans on Ethereum. And one day owning your car might be proved with an NFT.
```

minimalsm (approved):
Thanks again @nuclearfalcon

minimalsm:
@all-contributors please add @nuclearfalcon for ideas, docs

MicahZoltu:
> Adidas has discontinued its partnership with Kanye West and ceased production of Yeezys (Adidas own the IP); therefore, this reference no longer makes sense.

I agree with this reasoning (but not the original reasoning for the PR) so I'm OK with it.

PR #8822, opened 2022-12-04
refactor: stakinghowsoloworks to chakra

ImMiguelP:
## Description

Migrates `StakingHowSoloWorks` component from `emotion` to `chakra-ui`

## Related Issue
#8632

### DIFF
--- a/src/components/Staking/StakingHowSoloWorks.tsx
+++ b/src/components/Staking/StakingHowSoloWorks.tsx
@@ -1,5 +1,4 @@
 import React from "react"
-import styled from "@emotion/styled"
 import { graphql, useStaticQuery } from "gatsby"
 import { GatsbyImage } from "gatsby-plugin-image"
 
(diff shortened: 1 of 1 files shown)


src/components/Staking/StakingHowSoloWorks.tsx
```diff
@@ -56,10 +45,13 @@ const StakingHowSoloWorks: React.FC<IProps> = () => {
   ]
 
   return (
-    <Flex>
+    <Center
+      flexDirection={{ sm: "column", md: "row" }}
```
TylerAPfledderer, line 49: Recommend your first value use the `base` token instead of `sm` to ensure the value is applied at all screen widths below `md`

The layout is reverting back to row when the width gets below 480px.
```suggestion
      flexDirection={{ base: "column", md: "row" }}
```

TylerAPfledderer (changes requested):
Hey @ImMiguelP! I have only one tiny suggestion for you. 😸

src/components/Staking/StakingHowSoloWorks.tsx
ImMiguelP: @TylerAPfledderer sounds good ill commit the suggestion nice catch that's on me.

pettinarip (approved):
🚢 thanks @ImMiguelP

PR #8843, opened 2022-12-06
Migrate `StatsBoxGrid.tsx` to Chakra

victorpatru:
## Description

Migrate the `StatsBoxGrid.tsx` component to Chakra.

## Related Issue

#8632

### DIFF
--- a/src/components/StatsBoxGrid.tsx
+++ b/src/components/StatsBoxGrid.tsx
@@ -5,10 +5,13 @@ import axios from "axios"
 import { kebabCase } from "lodash"
 import { AreaChart, ResponsiveContainer, Area, XAxis } from "recharts"
 
+import { Grid, Icon } from "@chakra-ui/react"
+import { MdInfoOutline } from "react-icons/md"
+
 import Translation from "./Translation"
 import Tooltip from "./Tooltip"
 import Link from "./Link"
-import Icon from "./Icon"
+// import Icon from "./Icon"
 import StatErrorMessage from "./StatErrorMessage"
 import StatLoadingMessage from "./StatLoadingMessage"
 
@@ -46,22 +49,6 @@ const Title = styled.p`
   font-family: ${(props) => props.theme.fonts.monospace};
 `
 
-const Grid = styled.div`
-  display: grid;
-  grid-template-columns: repeat(2, 1fr);
-  margin: 2rem 2rem 0;
-  border-radius: 2px;
-  @media (max-width: ${({ theme }) => theme.breakpoints.l}) {
-    display: flex;
-    flex-direction: column;
-    width: 100%;
-    margin: 2rem 0 0;
-  }
-  @media (max-width: ${({ theme }) => theme.breakpoints.s}) {
-    margin: 0;
-  }
-`
-
 const Box = styled.div`
   position: relative;
   color: ${({ theme }) => theme.colors.text};
@@ -87,17 +74,17 @@ const StatRow = styled.div`
   flex-direction: column;
 `
 
-const StyledIcon = styled(Icon)`
-  fill: ${({ theme }) => theme.colors.text};
-  margin-right: 0.5rem;
-  @media (max-width: ${({ theme }) => theme.breakpoints.l}) {
-  }
-  &:hover,
-  &:active,
-  &:focus {
-    fill: ${({ theme }) => theme.colors.primary};
-  }
-`
+// const StyledIcon = styled(Icon)`
+//   fill: ${({ theme }) => theme.colors.text};
+//   margin-right: 0.5rem;
+//   @media (max-width: ${({ theme }) => theme.breakpoints.l}) {
+//   }
+//   &:hover,
+//   &:active,
+//   &:focus {
+//     fill: ${({ theme }) => theme.colors.primary};
+//   }
+// `
 
 const Lines = styled.div`
   position: absolute;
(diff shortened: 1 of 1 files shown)


src/components/StatsBoxGrid.tsx
```diff
@@ -233,18 +127,90 @@ const GridItem: React.FC<IGridItemProps> = ({ metric, dir }) => {
   )
 
   return (
-    <Box>
+    <Box
```
TylerAPfledderer, line 130: This should be a flex component. I would suggest `VStack` with `spacing={0}` and remove the `display="flex"` and `flexDirection="column"`

```tsx
<VStack
  spacing={0}
  // The other props
/>
```

src/components/StatsBoxGrid.tsx
```diff
@@ -233,18 +127,90 @@ const GridItem: React.FC<IGridItemProps> = ({ metric, dir }) => {
...
+        base: "-1px",
+        lg: "0",
+      }}
+      padding={{
+        base: "1rem",
+        lg: "1.5rem",
+      }}
+      paddingTop={{
+        base: "2rem",
+        lg: "0",
```
TylerAPfledderer, line 154: When this is applied with the `lg` value, it still overrides the value coming from the `padding` shorthand prop, so at `lg` and greater breakpoints, there is still `0` top padding.

This should be simplified to just using the shorthand `padding` prop only.
```jsx
padding={{
  base: "2rem 1rem 1rem",
  lg: "1.5rem"
}}
```
> Side note: I personally need to explore why it's not possible to use token values when providing multiple values in the shorthand. I should know this answer! 😅

src/components/StatsBoxGrid.tsx
```diff
@@ -233,18 +127,90 @@ const GridItem: React.FC<IGridItemProps> = ({ metric, dir }) => {
   )
 
   return (
-    <Box>
+    <Box
+      position="relative"
+      color="text"
+      height="20rem"
+      background={""}
```
TylerAPfledderer, line 134: I believe this is not needed as there is no default `background` color coming from anywhere else.
```suggestion
```

src/components/StatsBoxGrid.tsx
```diff
@@ -563,7 +542,23 @@ const StatsBoxGrid: React.FC<IProps> = () => {
   ]
   const dir = isLangRightToLeft(intl.locale as Lang) ? "rtl" : "ltr"
   return (
-    <Grid>
+    <Grid
```
TylerAPfledderer, line 545: Not blocking: I would see about exploring a different approach in going between `flex` and `grid` in a separate PR. Maybe just not doing that and sticking with Grid?

And I also like `SimpleGrid` here as a smaller component in this simple use case. Personal preference of course! 😄

src/components/StatsBoxGrid.tsx
```diff
@@ -170,14 +55,23 @@ const GridItem: React.FC<IGridItemProps> = ({ metric, dir }) => {
...
-      <span>
+    <VStack>
+      <Box>
         {state.value}{" "}
         <Tooltip content={tooltipContent(metric)}>
-          <StyledIcon name="info" />
+          <Icon
+            as={MdInfoOutline}
+            width="24px"
+            height="24px"
```
TylerAPfledderer, line 65: Nitpick: This can be simplified to the `boxSize` prop along with using the appropriate token value.
```suggestion
            boxSize={6}
```

TylerAPfledderer (changes requested):
Hey @victorpatru! After initial inspection, I have a few thoughts. 😸 

It is recommended to use Chakra token values instead of unit values.
> i.e. `2rem` --> `4`, say

You can view all the possible default tokens in the Chakra Docs [Default Theme Page](https://chakra-ui.com/docs/styled-system/theme)

I don't know if this is a consistent practice in this project, but I personally like curly brace notation with numerical token values.
```jsx 
margin={4}
```

src/components/StatsBoxGrid.tsx
TylerAPfledderer: cc @pettinarip here

victorpatru:
Hey @TylerAPfledderer. Thanks for the valuable help. I've gone ahead and committed some of the changes you proposed!

src/components/StatsBoxGrid.tsx
```diff
@@ -233,19 +126,82 @@ const GridItem: React.FC<IGridItemProps> = ({ metric, dir }) => {
...
+        base: "-1px",
+        lg: "0",
+      }}
+      padding={{ base: "2rem 1rem 1rem", lg: "1.5rem" }}
+    >
       <div>
-        <Title>{title}</Title>
+        <Text
+          fontSize="xl"
+          mb="0.5rem"
```
pettinarip, line 150:
```suggestion
          mb={2}
```

src/components/StatsBoxGrid.tsx
pettinarip: Yeah, agreed. With the `SimpleGrid` we could just do `<SimpleGrid columns={{ base: 1, lg: 2 }}>`, right?

src/components/StatsBoxGrid.tsx
```diff
@@ -563,7 +532,23 @@ const StatsBoxGrid: React.FC<IProps> = () => {
...
+        base: "flex",
+        lg: "grid",
+      }}
+      gridTemplateColumns="repeat(2, 1fr)"
+      margin={{
+        base: "0",
+        sm: "2rem 0 0",
+        lg: "2rem 2rem 0",
+      }}
+      borderRadius="2px"
```
pettinarip, line 546:
```suggestion
      borderRadius="sm"
```

pettinarip (approved):
LGTM! I'm just adding a few minor changes in favor of using more the Chakra values.

src/components/StatsBoxGrid.tsx
TylerAPfledderer: @pettinarip yes, this would work here.

I'll rollout a new PR to update this and any additional refactoring as needed.

Issue #8806, opened 2022-12-02
[BUG] BannerNotification text should be centered


corwintines:
### Describe the bug

Text should be center aligned in BannerNotification instead of left aligned

### To reproduce

1. Go to [staking page](https://ethereum.org/en/staking)
2. Notice banner at the top
3. See text left aligned

![Screen Shot 2022-12-01 at 5 16 10 PM](https://user-images.githubusercontent.com/15589226/205186593-43e2bc44-44c2-4be7-bf09-1c625f62f80d.png)

### Expected behavior

Text should be centered. BannerNotification could just be Stack instead of Flex

![Screen Shot 2022-12-01 at 5 21 35 PM](https://user-images.githubusercontent.com/15589226/205186761-654a8049-ff03-4594-8886-447739a945b4.png)

### Desktop (please complete the following information)

All

### Smartphone (please complete the following information)

All

ai-slave:
can't reproduce

TylerAPfledderer:
Banner not showing in prod, but showing in dev.

The wrapping flex element will need the `justifyContent="center"` here.

pettinarip:
TODO: once we fix this issue, remove the `Center` component from `PostMergeBanner` as it is not going to be needed anymore.

PR #8845, opened 2022-12-07
refactor: postmergebanner to chakra

ImMiguelP:
## Description

Migrates `PostMergeBanner` component from `emotion` to `chakra-ui`

## Related Issue
#8632

### DIFF
--- a/src/components/Banners/PostMergeBanner.tsx
+++ b/src/components/Banners/PostMergeBanner.tsx
@@ -3,14 +3,14 @@ import BannerNotification from "../BannerNotification"
 import Translation from "../Translation"
 
 import { TranslationKey } from "../../utils/translations"
-import { Flex, Text } from "@chakra-ui/react"
+import { Center, Text } from "@chakra-ui/react"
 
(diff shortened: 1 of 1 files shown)


TylerAPfledderer (commented):
LGTM! 🚀 

Not block: For #8806, might want to remove flexbox styling here for that issue's PR in favor of applying it directly to `BannerNotificaton`.

ImMiguelP:
> LGTM! 🚀
> 
> Not blocking: For #8806, might want to remove flexbox styling here for that issue's PR in favor of applying it directly to `BannerNotificaton`.

Sounds good so in this scenario the Center component will be used to keep all texts centered. I'll commit the new code rn.

TylerAPfledderer:
@ImMiguelP make sure to remove the `justifyContent` prop if `Center` is being used, because the component will render the center value.

pettinarip (approved):
Good job @ImMiguelP. Merging this for now and I agree that we shouldn't need the use of `Center` in this component once #8806 gets fixed.

Issue #8721, opened 2022-11-23
Investigate & create spike to extract translation strings from the code


pettinarip:

corwintines:
@pettinarip can this be closed now?

Issue #8586, opened 2022-11-14
Upgrade Gatsby to v5


pettinarip:
### Is your feature request related to a problem? Please describe.

No.

### Describe the solution you'd like

Upgrade Gatsby to v5 using the [official migration guide](https://v5.gatsbyjs.com/docs/reference/release-notes/migrating-from-v4-to-v5/).

### Describe alternatives you've considered

None.

### Additional context

More official resources:
- https://www.gatsbyjs.com/blog/gatsby-5-upgrade-say-no-to-yolo/
- Release notes https://www.gatsbyjs.com/docs/reference/release-notes/v5.0/

TylerAPfledderer:
FYI Chakra's Gatsby plugin in the current version might not support the upgrade. Currently checking if this has been addressed/investigated yet.

TylerAPfledderer:
Issue with Chakra's Gatsby Plugin has been fixed, and should be released by the time this project transitions to Gatsby 5

pettinarip:
Cool! great to hear that. Thanks for bringing that up.

PR #8866, opened 2022-12-10
React i18next - with translation strings extraction

pettinarip:
This PR reimplements the previously reverted PR #8406 for memory consumption issues on Gatsby Cloud.

## Description

All the work done in this PR is coming from: #9261

## Related Issue

#8721
#8345


docs/locales-process.md
```diff
@@ -0,0 +1,9 @@
+# Locales generation process
+
+Every time `yarn build` or `yarn start` is executed, the following process is
+going to be triggered as well:
+
+<img src="../static/locales.png">
+
+With this process, we reduce the amount of text we bundle on each page since we
+are querying only the necessary translations that each page needs.
```
pettinarip, line 9: A bit of context on how the whole process works. Would appreciate any feedback on this.

pettinarip: Also a recommended thing to read first that will help you understand and review the PR.

src/scripts/createLocales.ts
```diff
@@ -0,0 +1,205 @@
+import fs from "fs"
```
pettinarip, line 1: Main difference between this PR and the previous one #8406

docs/locales-process.md
```diff
@@ -0,0 +1,9 @@
+# Locales generation process
+
+Every time `yarn build` or `yarn start` is executed, the following process is
+going to be triggered as well:
+
+<img src="../static/locales.png">
```
samajammin, line 6: Diagrams for the win! 😍

src/components/ButtonDropdown.tsx
```diff
@@ -120,7 +119,7 @@ export interface IProps {
 
 const ButtonDropdown: React.FC<IProps> = ({ list, className }) => {
   const [isOpen, setIsOpen] = useState<boolean>(false)
-  const intl = useIntl()
+  const { t } = useTranslation()
```
corwintines, line 122: This isnt used anywhere now, I think we can just remove this from the file altogether

src/components/Roadmap.tsx
```diff
@@ -1,12 +1,11 @@
 import React, { useState, useEffect } from "react"
-import { useIntl } from "react-intl"
+import { useI18next, useTranslation } from "gatsby-plugin-react-i18next"
+import styled from "@emotion/styled"
```
corwintines, line 3:
```suggestion
```
Not needed anymore

src/components/Footer.tsx
```diff
@@ -53,225 +54,226 @@ export interface LinkSection {
...
 
-  const isPageRightToLeft = isLangRightToLeft(intl.locale as Lang)
+  const isPageRightToLeft = isLangRightToLeft(language as Lang)
 
   const [medBp] = useToken("breakpoints", ["md"])
 
   const linkSections: Array<LinkSection> = [
     {
-      title: "use-ethereum",
+      title: t("use-ethereum"),
```
wackerow, line 66: Do we want to apply this here?
Line 342 below passes this as a key to the Translation component:
```ts
            {linkSections.map((section: LinkSection, idx) => (
              <Box key={idx}>
                <Heading as="h3" fontSize="sm" lineHeight="1.6" my="1.14em">
                  <Translation id={section.title} />
                </Heading>
                <List fontSize="sm" lineHeight="1.6" fontWeight="400" m={0}>
```
@pettinarip Should we remove the use of the `<Translation/>` component below?

src/components/Nav/index.tsx
```diff
@@ -339,24 +334,24 @@ const Nav: React.FC<IProps> = ({ path }) => {
 
   const ednLinks: Array<IItem> = [
     {
-      text: "home",
+      text: t("home"),
```
wackerow, line 337: Line 433 below, we map this as a key... similar to a note elsewhere, should we remove the use of `<Tranlation/>` below?
```ts
          {ednLinks.map((link, idx) => (
            <NavLink
              key={idx}
              to={link.to}
              isPartiallyActive={link.isPartiallyActive}
            >
              <Translation id={link.text} />
            </NavLink>
          ))}
```

src/scripts/createLocales.ts
```diff
@@ -0,0 +1,217 @@
...
+      fileName: path.basename(file.path),
+    })
+
+    const namespace = createNamespaceFromFilename(file.path)
+
+    const customHandler = (key: string, options: any) => {
+      const ns = file.path.includes("components") ? "components" : namespace
+
+      // if the key is not found in the current lng then we store the
+      // defaultLng value instead
```
wackerow, line 83: @pettinarip Curious if there is a reason we're using "Lng" without the "a" for language in this script... we don't do this anywhere else in the codebase.

wackerow (commented):
This looks great @pettinarip!! Well done! 🎉 
Left a couple comments, but aside from this few small things this seems to be working great for me on my end.

minimalsm:
Bug: When I'm on the tutorials homepage in a non-English language, if I click on a tutorial, it prepends an additional lang path, causing a 404. Confirmed broken in French and Chinese, but works as expected in English.

So, instead of:

[/zh/developers/tutorials/run-node-raspberry-pi/](https://ethereumorgwebsitedev01-i18nextractstrings.gatsbyjs.io/zh/developers/tutorials/run-node-raspberry-pi/)

I get:

[/zh/zh/developers/tutorials/run-node-raspberry-pi/](https://ethereumorgwebsitedev01-i18nextractstrings.gatsbyjs.io/zh/zh/developers/tutorials/run-node-raspberry-pi/)

minimalsm:
We don't have any highlighting on active breadcrumbs. 

Note this sometimes happens on Gatsby preview deploys so might be nothing to do with this PR.

Screenshot 2023-01-05 at 12 55 49

src/pages/developers/tutorials.tsx
```diff
@@ -428,7 +425,7 @@ const TutorialsPage = ({
                 {/* TODO: Refactor each tutorial tag as a component */}
                 <Emoji text=":writing_hand:" fontSize="sm" mr={2} />
                 {tutorial.author} •
-                {published(intl.locale, tutorial.published ?? "")}
+                {published(language, tutorial.published ?? "")}
```
minimalsm, line 428: Not something to fix here, but noting we should require tutorials to have published dates on build.

src/components/ButtonDropdown.tsx
```diff
@@ -136,7 +135,7 @@ const ButtonDropdown: React.FC<IProps> = ({ list, className }) => {
     <Container
       className={className}
       ref={ref}
-      aria-label={`Select ${translateMessageId(list.text, intl)}`}
+      aria-label={`Select ${list.text}`}
```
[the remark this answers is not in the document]
pettinarip: Oh I guess that's why I had that import unused...I forgot to wrap this with the `t` function. Thanks!

src/components/Footer.tsx
pettinarip: Good catch. Yea, we don't need the `<Translation />` component here since we are already doing the translations above.

The reason I've changed it is because the scanner we are using can't detect translation keys from variables. In this case, the scanner will find this line `<Translation id={section.title} />` and it will not know what key `section.title` holds. Ideally, we should wrap all the translation key strings with the `t` function or the `Translation` component.

src/components/Nav/index.tsx
pettinarip: Yes, same case as the prev comment. Thanks.

src/scripts/createLocales.ts
pettinarip: Good catch. Didn't realize that. I guess I got confused with the terminology used by the scanner's code 🤦🏼 Will fix that.

src/components/ButtonDropdown.tsx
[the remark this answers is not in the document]
pettinarip: Oh no, now I remember. So, `text` now is not more a `TranslationKey`, its a `string`. Bc the translation is done by the consumer. Here we are expecting to receive an already translated string.

For the same reason I explained in the other comment, the scanner can't extract the translation keys from variables.

pettinarip:
As described in the PR description, the goal of this PR was to reduce memory consumption. At the moment of creation of the PR, it was building fine but after adding a few tweaks due to the code review, we started having memory issues again.

This means that the PR, at the moment of creation, was still consuming a lot of memory. To merge this PR, we want to be confident that we are far away from exceeding the memory limits and that we are not going to have future problems like this again.

I see 2 paths to go from here:

- Abandon the PR and the effort since we are planning on moving to NextJS in the near future. NextJS implementation can fix these problems that are directly related to how Gatsby's build system is designed
- Put the PR on hold and keep investigating what we can do to optimize the memory consumption and merge this PR later when we are in a good spot

Having said that, since the end goal of the i18n lib change is reducing the final JS bundles (which directly affect the UX), I think we should put this on hold and keep investing time in figuring out how to solve this problem.

There are other parts of the build process that may affect memory consumption as described [here](https://www.gatsbyjs.com/docs/how-to/performance/resolving-out-of-memory-issues/).

1. One major thing to optimize is image processing. There was [a PR](https://github.com/ethereum/ethereum-org-website/pull/6585) where we were trying to do this. We could recover that work and apply it to see how it impacts the overall memory consumption
2. Another thing is to [upgrade Gatsby to v5](https://github.com/ethereum/ethereum-org-website/issues/8586) since they say there are many optimizations implemented in this version
3. Keep investigating how we can break even more out the translations namespaces to reduce the amount of text we inject into bundles

Currently, Gatsby Cloud is [offering 8gb](https://support.gatsbyjs.com/hc/en-us/articles/360052326714-Platform-Limits) on our current plan. That should be enough. I understand that Vercel is also offering the same memory limits.

cc @samajammin @minimalsm @corwintines @wackerow

samajammin:
Thanks for the update @pettinarip!

> since the end goal of the i18n lib change is reducing the final JS bundles (which directly affect the UX), I think we should put this on hold and keep investing time in figuring out how to solve this problem

I agree that is what we ultimately should focus on solving, the JS bundle is the user-facing issue (vs. reducing memory consumption on builds).

I think all those ideas are decent options to explore. I would be curious to see a quick spike on upgrading to Gatsby v5 & seeing if that fixes any of our issues 🤔 if not, perhaps images are an area to optimize like you mention. Realistically I'm not sure we should say "let's just solve it with Next.js" as that feels like a massive undertaking & I'm not sure if that makes sense as a near-term priority... but if this issue persists & we don't see other options, perhaps it's an argument to prioritize it.

One more alternative to explore would be simply ask Gatsby how much to increase memory limits. It's certainly a bandaid solution but if it doesn't add much additional cost it might be the most reasonable approach in the short term.

src/components/Layer2/Layer2Onboard.tsx
```diff
@@ -262,30 +261,20 @@ const Layer2Onboard: React.FC<IProps> = ({
...
             <SelectedContainer>
               <p>
-                <b>{`${translateMessageId(
-                  "layer-2-onboard-wallet-selected-1",
-                  intl
-                )} ${selectedL2.name} ${translateMessageId(
-                  "layer-2-onboard-wallet-selected-2",
-                  intl
-                )}`}</b>
+                <b>{`${t("layer-2-onboard-wallet-selected-1")} ${
```
minimalsm, line 271: .

gatsby-ssr.tsx
```diff
@@ -3,34 +3,3 @@
  *
  * See: https://www.gatsbyjs.org/docs/ssr-apis/
  */
-
```
corwintines, line 6: I think we can just remove this file. I dont think this is doing anything anymore, unless there is a reason we need to keep it around for gatsby

corwintines (changes requested):
Just noticed this, otherwise I think we are good 🚀

corwintines (approved):
🚀

PR #8869, opened 2022-12-10
Update a Dapp Link

jhonnyvianello:
## Update a Dapp Link
Updated the CryptoPunks link since, after the acquisition of Yuga Labs, the web app will soon be moving from "https://www.larvalabs.com/cryptopunks" to "https://cryptopunks.app/" (note for example that, by selecting a CryptoPunks, you will already be redirected to the new site) 🆕


corwintines (approved):
Nice. Thanks, @jhonnyvianello

PR #8876, opened 2022-12-11
Fix descriptions for consensus mechanisms

jpanachida:
"Block creation" sections for Proof-of-work and Proof-of-stake were mixed up. I put the right descriptions.


jmcook1186 (approved):
thanks for catching this!

PR #8890, opened 2022-12-14
Refactor: Migrate `Trilemma` to Chakra UI

TylerAPfledderer:
## Description

Migrates the `Trilemma` component to Chakra UI

- Splits the triangle svg to its own file
- Abstracts the component logic to a hook
- DRY up code
- Boost typescript usage to accommodate the file splitting

## Related Issue
#8632


src/components/Trilemma/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+          </Text>
+        </Show>
+        {/* TODO: replace `l` with `lg` after UI migration */}
+        <Hide below="l">
+          <Card {...cardDetail} mt={8} minH="300px" />
+        </Hide>
+      </Flex>
+      {/* TODO: replace `l` with `lg` after UI migration */}
+      <Hide above="l">
+        <Drawer
```
pettinarip, line 75: Cool!

src/components/Trilemma/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+          <Translation id="page-upgrades-vision-trilemma-p" />
+        </Text>
+        <Text>
+          <Translation id="page-upgrades-vision-trilemma-p-1" />
+        </Text>
+        <Text>
+          <Translation id="page-upgrades-vision-trilemma-p-2" />
+        </Text>
+        {/* TODO: replace `l` with `lg` after UI migration */}
+        <Show below="l">
```
pettinarip, line 63: Can you remind me where this breakpoint is coming from? I don't think we are defining the `l` breakpoint in the chakra theme.

pettinarip (approved):
This is a really cool refactor @TylerAPfledderer good job! loved the way you broke it out.

src/components/Trilemma/index.tsx
TylerAPfledderer: @pettinarip I believe there is still a conflict with the original implementation of styled-components and emotion, where the theming sometimes gets pulled from the older config that still has to be applied.

I've only seen this issue really occur with the media query components because of the hooks that are called under the hood to retrieve the theme vals.

pettinarip: Right. Thanks for the reminder.

I was thinking that we can just start using the `lg` value here since:
- on the parent component we are already using the chakra breakpoints `flexDirection={{ base: "column", lg: "row" }}` 
- the impact of this new value is only going to be inside this component. This will not have a negative UX in any other component

TylerAPfledderer: @pettinarip right. A weird behavior!

What I should do then is keep implementing Chakra's `useToken` to extract the value from the Chakra theme and apply to the component, guaranteeing no breaking changes on migration completion (adding a TODO to remove later)

For consistency, there might be a couple of other migrated components that would need this update.

What do you think?

pettinarip: That's perfect, yes 💪🏼

TylerAPfledderer: Ok!

I'll make the change here, and if there are still others that need that change I'll create a new PR. 😁

TylerAPfledderer:
@pettinarip I have updated the breakpoint implementation for you to verify 🤘🏼

Issue #8923, opened 2022-12-15
Bug report: wrong color combination for tags on stablecoins page


konopkja:
### Describe the bug

Stablecoins page contains a section where tags are used.

The color combinations have very low contrast and the text inside the tag is unreadable.

### To reproduce

1. Go to (https://ethereum.org/en/stablecoins/)
3. Scroll down to the "How to get stablecoins" section
4. See the color combination

### Expected behavior

Contrasting colors so that text is readable. 

<img width="247" alt="Screenshot 2022-12-15 at 20 37 47" src="https://user-images.githubusercontent.com/100724231/207903042-84d761a2-7464-4218-8bb1-b7f7e00b30f2.png">
<img width="322" alt="Screenshot 2022-12-15 at 20 38 00" src="https://user-images.githubusercontent.com/100724231/207903044-ddf99fc9-5e03-474b-9b0b-59adb10e4df7.png">

### Screenshots of current situation

<img width="458" alt="Screenshot 2022-12-15 at 20 24 13" src="https://user-images.githubusercontent.com/100724231/207901516-fae8645d-b137-4044-9a6a-121b83651f27.png">
<img width="397" alt="Screenshot 2022-12-15 at 20 24 28" src="https://user-images.githubusercontent.com/100724231/207901520-74e00bc2-7c02-46b4-9225-3d7e417f0b85.png">

### Desktop (please complete the following information)

latest chrome, macbook

TylerAPfledderer:
@konopkja bug was introduced in #8774 (my mistake). I'll take this to make the correction!

uday03meh:
Hi, Can I start working on it?

TylerAPfledderer:
> Hi, Can I start working on it?

Hey @uday03meh! I have already addressed this with the referenced PR above. 😄

pettinarip:
Nice, thanks @uday03meh & @TylerAPfledderer (I've assigned this to you)

pettinarip:
This is issue is fixed in #8601

PR #8774, opened 2022-11-28
Refactor: Migrate `TutorialTags` to Chakra UI

TylerAPfledderer:
## Description

Migrate component to Chakra UI.

Includes update to `Pill` component to accept incoming Chakra props.

## Related Issue
#6374

PR #8927, opened 2022-12-15
Fix (`StablecoinAccordion`): change `color` prop in pills to `background`

TylerAPfledderer:
## Description

In PR #8774 a bug was introduced to the `StableAccordion` component where the pills were not rendering the text color correctly.

This PR changes the `color` prop to the `StyledPill` instances to `background` which was changed in the `Pill` component. This ensures that the background of the `StyledPill`'s receives the passed in color and not the text.

## Related Issue

Closes: #8923


pettinarip:
Hey @TylerAPfledderer sorry for the late review on this but I think that, considering that we are refactoring the Pill component in #8601, there is no need to merge this. Feel free to disagree and re-open.

TylerAPfledderer:
Nope that is good. Thank you!

Issue #8937, opened 2022-12-17
Clarifying Description Wording


machin3boy:

corwintines:
Look at PR for reason, closing.

PR #8938, opened 2022-12-17
fix: clarified wording regarding slasher rewards

machin3boy:
## Description

"A block proposer which includes slashing will be rewarded with the `slashed_validators_effective_balance / 512`." is less descriptive and more vague than "There is also a reward for a block proposer reporting validators that should be slashed, equal to 1/512 * effective balance for each slashed validator, `slashed_validators_effective_balance / 512`."

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/8937


corwintines:
Thanks for the suggestion @machin3boy. I reread the paragraph and this change seems a bit redundant to me. Im going to close it out, but feel free to reopen if you disagree and let us know why :)

PR #8953, opened 2022-12-19
Remove dangling code block

aviggiano:
Remove dangling code block

## Description

This code block seems to be outdated.

screenshot-2022-12-19-17-35-36


corwintines (approved):
Thanks for fixing this @aviggiano!

PR #8959, opened 2022-12-20
Fix: Replace `div` with fragment in `tutorial` template

TylerAPfledderer:
## Description

Currently, there is both left and right overflow on tutorial pages, which use the `tutorial.tsx` template:
Screen Shot 2022-12-20 at 11 39 02

This issue points to the `div` element wrapping all the content in the `tutorial.tsx` template. The flexbox applied to `main` in the `Layout.tsx` is stretching this `div` past the browser window. Replacing this `div` with a React fragment resolves the global overflow issues for the tutorial pages, and retains the styling in `Layout` to prevent creating other bugs.

This seems to only impact the tutorial pages and no other templates.

## Related Issue

N/A


wackerow (approved):
@TylerAPfledderer Thanks again! Nice work =)

PR #8999, opened 2022-12-23
Fix hydration issue on FeedbackWidget

### LINKED ISSUE #8996: Hydration issue on `FeedbackWidget`
### Describe the bug

When working on local dev mode, there is a `There was an error while hydrating` error when the `FeedbackWidget` is enabled.

### To reproduce

1. Using the local dev mode
2. Open any page in English
3. See hydration error
4. Now, go to the `Layout` component and hide/remove the `FeedbackWidget` component
5. Reload the page and you will no have the error

### Expected behavior

Probably there is something going on in the `useEffect` of the `FeedbackWidget` that generates a different output in server-side.

We need to have render the same output on server & client.

This component also depends on a prop called `location`. Perhaps that prop is not the same on server and on the client.

### Screenshots

_No response_

_No response_

_No response_

### Additional context

_No response_

pettinarip:
## Description

Get current language from the intl context and not from the url to avoid having hydration conflicts.

## Related Issue

Fixes #8996

### DIFF
--- a/src/components/FeedbackWidget.tsx
+++ b/src/components/FeedbackWidget.tsx
@@ -140,7 +140,7 @@ const FeedbackWidget: React.FC<FeedbackWidgetProps> = ({ location = "" }) => {
 
   useKeyPress(`Escape`, handleClose)
 
-  if (!location.includes("/en/")) return null
+  if (intl.locale !== "en") return null
   const closeButtonSize = "24px"
   return (
     <>


src/components/FeedbackWidget.tsx
```diff
@@ -140,7 +140,7 @@ const FeedbackWidget: React.FC<FeedbackWidgetProps> = ({ location = "" }) => {
 
   useKeyPress(`Escape`, handleClose)
 
-  if (!location.includes("/en/")) return null
```
pettinarip, line 143: `location` on server-side is not going to give us the same url. Since we are only interested in getting the current `locale`, we can get it from the `intl` context.

PR #9003, opened 2022-12-24
Docs: Update nonce description to account for front-running [Fixes #9002]

### LINKED ISSUE #9002: Docs: Accounts - Edit nonce description to account for front-running
Current description:
A counter that indicates the number of transactions sent from the account. This ensures transactions are only processed once. In a contract account, this number represents the number of contracts created by the account.

Proposed description:
A counter that indicates the number of transactions sent from the account. This ensures transactions are only processed once **and in the order that they were created by the account**. In a contract account, this number represents the number of contracts created by the account.

Reason:
Current description does not explicitly account for the possibility of accounts front-running their own transactions. For example, a user can submit a malicious tx with a higher nonce and pay a higher gas fee to get the later tx to be processed first.

0xKai27:
## Description
Current description does not explicitly account for the possibility of accounts front-running their own transactions. For example, a user can submit a malicious tx with a higher nonce and pay a higher gas fee to get the later tx to be processed first.

## Related Issue
Fixes #9002


src/content/developers/docs/accounts/index.md
```diff
@@ -42,7 +42,7 @@ Both account types have the ability to:
 
 Ethereum accounts have four fields:
 
-- `nonce` – A counter that indicates the number of transactions sent from the account. This ensures transactions are only processed once. In a contract account, this number represents the number of contracts created by the account.
+- `nonce` – A counter that indicates the number of transactions sent from the account. This ensures transactions are only processed once and in the order that they were created by the account. In a contract account, this number represents the number of contracts created by the account.
```
minimalsm, line 45: Is this true? You can manipulate the nonce to submit a new transaction before a pending one...

minimalsm (changes requested):

src/content/developers/docs/accounts/index.md
0xKai27: My description might have made it confusing with the double negatives. The nonce acts as a transaction counter at the account level and has 2 functions:

- Avoid double-spend: For each tx originating from an account, the protocol will only process unique nonces. That is, if a user submits multiple txs with the same nonce, only one will be processed by the network (likely the one with higher priority/gas fee). 
- Tx ordering: The protocol also processes txs from an account in the order dictated by the nonce. This ensures that a tx with a higher nonce is not mined before a tx with a lower nonce. As such, even if the higher nonce tx is submitted with a higher gas fee, the protocol will have to sequentially confirm the txs based on the account nonce.

Both the above guarantees that the system can not be manipulated via the gas market.

jmcook1186: I think this is basically ok, but the wording around the transaction ordering might be a little misleading because you can send multiple transactions with the same nonce and let their gas prices determine which one is executed- effectively overriding the pending transaction for a given nonce. 

How about this:

```suggestion
- `nonce` – A counter that indicates the number of transactions sent from an externally-owned account or the number of contracts created by a contract account. Only one transaction with a given nonce can be executed for each account, protecting against replay attacks where signed transactions are repeatedly broadcast and re-executed.
```

corwintines: Will pull in @jmcook1186 suggestion and merge.

corwintines (approved):

PR #9004, opened 2022-12-24
Refactor: Migrate `StablecoinAccordion` to Chakra UI

TylerAPfledderer:
## Description

Migrate component to Chakra UI

## Related Issue
#8632


src/components/StablecoinAccordion/use-stablecoin-accordion.ts
```diff
@@ -0,0 +1,311 @@
+import { useStaticQuery, graphql } from "gatsby"
```
pettinarip, line 1: For consistency, lets keep using camelCase names. This file should be named `useStablecoinAccordion.ts`

src/components/StablecoinAccordion/index.tsx
```diff
@@ -0,0 +1,274 @@
...
+  return (
+    <Accordion
+      borderRadius="base"
+      mt={12}
+      width="full"
+      allowToggle
+      reduceMotion
+    >
+      <AccordionCustomItem category="dapps">
+        {({ LeftColumnPanel, RightColumnPanel }) => (
```
pettinarip, line 107: Wondering why you have chosen this approach instead of just define those components in this file as you did with `StepBoxContainer` for example.

I mean, why we can't just define
```tsx
const LeftColumnPanel = (props: ChildOnlyType & Partial<BoxProps>) => ();
const RightColumnPanel = (props: ChildOnlyType) => ();
```
in this file?

I think that if there is no special reason, this would simplify the structure of the component.

pettinarip (changes requested):
This looks fantastic @TylerAPfledderer nice break out & simplification of this component.

I left a couple of questions/requests. LMK what do you think.

src/components/StablecoinAccordion/use-stablecoin-accordion.ts
TylerAPfledderer: You got it!

The [Trilemma component](https://github.com/ethereum/ethereum-org-website/tree/dev/src/components/Trilemma) would need an update with its similar file.

src/components/StablecoinAccordion/index.tsx
TylerAPfledderer: @pettinarip it was an approach to contain these to components within `AccordionCustomItem`, but I can understand the inconsistency you describe here.

No reason not to use the panel components directly instead of through a child function. I'll go ahead and make the change. 😄

TylerAPfledderer:
@pettinarip applied your change requests for re-review. 🤘🏼

src/components/StablecoinAccordion/use-stablecoin-accordion.ts
pettinarip: Oh did't see that one. Good catch, yea. Will do a quick refactor.

src/components/StablecoinAccordion/index.tsx
pettinarip: Cool. Yea, this is just a minor comment but I think simplicity will make it more readable for other devs. Thanks for doing that.

pettinarip (approved):
This is greaaaat! eager to see those `.stories` in the folder 😬

PR #9008, opened 2022-12-25
Chore: remove extra `Item` type export in `TableOfContents`

TylerAPfledderer:
## Description

This removes an extra type export of `Item` from the main `TableOfContents` file, as it is already exported from [utils.ts](https://github.com/ethereum/ethereum-org-website/blob/efa2d964d6715db1e1b5700ae6a513ade0f54faf/src/components/TableOfContents/utils.ts#L27-L30)


pettinarip (approved):
Cool, thanks!

PR #9038, opened 2023-01-03
Chore: Remove `Accordion` Component

TylerAPfledderer:
## Description

This PR removes the `Accordion` component from the project, as it is currently unused and not needed.

It appeared to come from PR #1938 as a component for experimentation, but was never implemented nor removed prior to the merge.

## Related Issue

#6374


pettinarip (approved):
Good catch. Thanks for the PR

PR #9056, opened 2023-01-04
Migrate the `SideNav` component to Chakra

soheil555:
## Description

Migrate the `SideNav` component to Chakra

## Related Issue
#8632


pettinarip (approved):
Good job @soheil555. lgtm.

Probably when we do the implementation of the DS and we refactor this component again we could redo part of this using the [Chakra's Accordion](https://chakra-ui.com/docs/components/accordion/usage). I think that will reduce the complexity of this component.

Issue #8640, opened 2022-11-17
Migrate `TranslationLeaderboard.tsx` to Chakra


pettinarip:
*This is part of our [2nd wave of migration](https://github.com/ethereum/ethereum-org-website/issues/8632) of the [Implement UI library](https://github.com/ethereum/ethereum-org-website/issues/6374) epic.*

**Things to consider**

- Lets see if we can replace the logic of the [`OptionContainer`](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/TranslationLeaderboard.tsx#L148) with the [radio buttons from chakra](https://chakra-ui.com/docs/components/radio/usage#custom-radio-buttons)

Check our [epic](https://github.com/ethereum/ethereum-org-website/issues/6374) & the [migration guide](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/chakra-migration-guide.md) for more information.

vdusart:
Hey @pettinarip,
I'm almost done with this component, but by looking at the documentation of the Radio Buttons on Chakra, I saw that:
![image](https://user-images.githubusercontent.com/43795504/205769270-1ac2e7ba-0487-48ed-bf19-173029e667b9.png)

In our case, it would make more sense to use `Tabs` as we are not collecting data, right ?

pettinarip:
Good question! 

We are filtering a data collection and we are displaying the result of that after the user clicks on one of the options. I think that the `Tabs` makes more sense if we were just displaying static information but it is not the case here.

LMK what do you think about that reasoning.

vdusart:
Humm 🤔 that's an other way to see things
For me what they meant when talking about `data collection purposes` was for sending POST data through forms

pettinarip:
@vdusart  you can have data collection in any environment, not only for sending data to a server. In this case, our data collection is the list of translators and we need to filter/sort them accordingly with the user's selection.

I still think this is a radio button group since you need to run a process each time you change the option. The radio buttons logic is going to fit better for that case. Otherwise, using Tabs you will need to implement an `onChange` function where you will need to create the logic to detect which tab was selected and perform the operation.

Again, you can do this with both. IMO Tabs are for static content and Radio (forms) for dynamic.

PR #9058, opened 2023-01-04
Convert `translationLeaderboard` to chakra

vdusart:
## Description

Convert `TranslationLeaderboard.tsx` file to Chakra

## Related Issue
Close: #8640
Epic: #8632

### DIFF
--- a/src/components/TranslationLeaderboard.tsx
+++ b/src/components/TranslationLeaderboard.tsx
@@ -46,6 +46,8 @@ const Button = (props) => {
         color: "primary",
         borderColor: "primary",
       }}
+      _focus={{}}
+      _active={{}}
       {...props}
     />
   )
--- a/src/components/TranslationLeaderboard.tsx
+++ b/src/components/TranslationLeaderboard.tsx
@@ -9,24 +9,6 @@ import { Option, OptionContainer, OptionText } from "./SharedStyledComponents"
 import Translation from "./Translation"
 import { Box, Flex, useColorModeValue } from "@chakra-ui/react"
 
-const Item = styled.div`
-  text-decoration: none;
-  display: flex;
-  justify-content: space-between;
-  align-items: center;
-  color: ${(props) => props.theme.colors.text} !important;
-  box-shadow: 0 1px 1px ${(props) => props.theme.colors.tableItemBoxShadow};
-  margin-bottom: 1px;
-  padding: 0.5rem 1rem;
-  width: 100%;
-  color: #000000;
-  &:hover {
-    border-radius: 4px;
-    box-shadow: 0 0 1px ${(props) => props.theme.colors.primary};
-    background: ${(props) => props.theme.colors.tableBackgroundHover};
-  }
-`
-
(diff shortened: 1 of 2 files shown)


src/components/TranslationLeaderboard.tsx
```diff
@@ -143,46 +101,66 @@ const TranslationLeaderboard: React.FC<IProps> = ({
...
+        </RadioCard>
+      </Flex>
+      <Box bg="background" boxShadow={tableBoxShadow} w="full" mb={8}>
+        <Flex
+          bg="grayBackground"
+          textDecoration="none"
+          justifyContent="space-between"
+          alignItems="center"
+          color="text"
+          mb="px"
```
pettinarip, line 143:
```suggestion
          mb="1px"
```

src/components/TranslationLeaderboard.tsx
```diff
@@ -214,45 +192,116 @@ const TranslationLeaderboard: React.FC<IProps> = ({
...
             }
             return (
-              <Item key={idx}>
+              <Flex
+                textDecoration="none"
+                justifyContent="space-between"
+                alignItems="center"
+                color="text"
+                boxShadow={tableItemBoxShadow}
+                mb="px"
```
pettinarip, line 201:
```suggestion
                mb="1px"
```

src/components/TranslationLeaderboard.tsx
```diff
@@ -214,45 +192,116 @@ const TranslationLeaderboard: React.FC<IProps> = ({
...
+          w={{ base: "full", lg: "initial" }}
+          justifyContent="center"
+          ml={{ base: "0", lg: "2" }}
+          mr={{ base: "0", lg: "2" }}
+          _hover={{
+            color: "primary",
+            borderColor: "primary",
+          }}
+          onClick={filterAmount === 10 ? showMore : showLess}
+        >
```
pettinarip, line 286: This should be a button. I know it was not a button before but I think we should fix this. I'd understand if you don't want to do that on this PR. I could create a new issue to tackle that refactor, up to you. LMK what do you think.

pettinarip: And we could share these styles with the ones we are using on the `RadioCard`.

I have in mind this:
```tsx
const Button = (props) => {
  return (
    <ChakraButton
      borderRadius="2rem"
      borderWidth="1px"
      borderStyle="solid"
      borderColor="text"
      ...
      ...props
    />
  )
}

// on the RadioCard component
const RadioCard = () => {
  ...
  
  return (
    <Button
      as="label"
      {...checkbox}
      ...
  )
}
```

pettinarip (changes requested):
Great job @vdusart. I've tested it and the component is working fine.

I've left a few small issues and an idea to refactor the buttons we are showing that I think are worth implementing.

vdusart:
Ok @pettinarip,
I just implemented the changes you proposed, tell me what you think about it :smile:

pettinarip (approved):
Thanks @vdusart lgtm!

PR #9068, opened 2023-01-05
update password managers

wslyvh:
Removing Lastpass due to recent security incidents.
They're also not open-source.

### DIFF
--- a/src/content/security/index.md
+++ b/src/content/security/index.md
@@ -55,8 +55,8 @@ Remembering strong, unique passwords for every account you have isn't ideal. A p
 
 - [Bitwarden](https://bitwarden.com/)
 - [KeePass](https://keepass.info/)
-- [LastPass](https://www.lastpass.com/)
 - [1Password](https://1password.com/)
+- Or check out other [recommended password managers](https://www.privacytools.io/secure-password-manager)
 
 ### Use Two-Factor Authentication {#two-factor-authentication}
 


wackerow (approved):
I don't disagree with this, but just note that being open source has not been a strict requirement on this page in the past, but perhaps we should discuss this more.

Do we keep listing 1Password due to it's popularity despite it not being open source?

On LastPass, I agree for the safety of users we should not be suggesting this given recent events. Others are free to push back. Thanks @wslyvh!

corwintines (approved):
Agree with removing LastPass given context. @wackerow I don't think they have to be open source at the moment, but something to consider for sure. For now, I think we can keep 1Password and just remove this.

isabelladebrito:
Hi @wslyvh, thanks again for contributing to ethereum.org!

Besides GitPOAP, you're also eligible to claim an "onchain achievement token" (also known as "OAT") on Galxe 🏅 

Here's how you can claim:

1. Head over to [https://galxe.com/ethereumorg/campaign/GCBctt4BkX](https://galxe.com/ethereumorg/campaign/GCBctt4BkX)
2. We have already assigned you a role on ethereum.org's Discord server, from where you should be able to claim your OAT on Galxe
3. Connect your wallet and Discord user
4. Check the OAT(s) you're eligible to claim et voilà! 

If you have any questions, don't hesitate to reach out to reply to this comment or reach out to me via ethereum.org's Discord server. Cheers!

PR #9081, opened 2023-01-06
Learn Page: Added description regarding smart contract accessibility …

### LINKED ISSUE #9078: Learn Page: Change "deploy code" to "deploy publicly executable code"
## Description

Propose additional wording to provide greater clarity regarding smart contract code accessibility. Current phrasing does not make explicit that anyone with a wallet can actually execute the network deployed code (i.e. smart contract). Possible misinterpretation whereby smart contract can only be accessed by the deployer and authorised parties (i.e. private network instead of permissioned).

## Suggested

Not only can Ethereum do what Bitcoin does (transfer money globally), it’s capable of much more – anyone can deploy **_publicly executable_** code onto the network. And because it’s so flexible, any computer program can run on Ethereum.

0xKai27:
## Description

Propose additional wording to provide greater clarity regarding smart contract code accessibility. Current phrasing does not make explicit that anyone with a wallet can actually execute the network deployed code (i.e. smart contract). Possible misinterpretation whereby smart contract can only be accessed by the deployer and authorised parties (i.e. private network instead of permissioned).

## Related Issue

Fixes #9078

### DIFF
--- a/src/intl/en/page-learn.json
+++ b/src/intl/en/page-learn.json
@@ -13,7 +13,7 @@
   "hero-button-lets-get-started": "Let's get started",
   "what-is-crypto-1": "You have probably heard a thing or two about cryptocurrencies, Bitcoin, and blockchain, but do you know what those actually are and how they relate to Ethereum?",
   "what-is-crypto-link-1": "And what is Ethereum anyway?",
-  "what-is-crypto-2": "Not only can Ethereum do what Bitcoin does (transfer money globally), it’s capable of much more – anyone can deploy code onto the network. And because it’s so flexible, any computer program can run on Ethereum.",
+  "what-is-crypto-2": "Not only can Ethereum do what Bitcoin does (transfer money globally), it’s capable of much more – anyone can deploy publicly executable code onto the network. And because it’s so flexible, any computer program can run on Ethereum.",
   "what-is-ethereum-card-title": "What is Ethereum?",
   "what-is-ethereum-card-description": "If you are new, start here to learn the basics of crypto and why Ethereum matters.",
   "what-is-ethereum-card-image-alt": "Illustration of a person peering into a bazaar, meant to represent Ethereum.",


src/intl/en/page-learn.json
```diff
@@ -13,7 +13,7 @@
   "hero-button-lets-get-started": "Let's get started",
   "what-is-crypto-1": "You have probably heard a thing or two about cryptocurrencies, Bitcoin, and blockchain, but do you know what those actually are and how they relate to Ethereum?",
   "what-is-crypto-link-1": "And what is Ethereum anyway?",
-  "what-is-crypto-2": "Not only can Ethereum do what Bitcoin does (transfer money globally), it’s capable of much more – anyone can deploy code onto the network. And because it’s so flexible, any computer program can run on Ethereum.",
+  "what-is-crypto-2": "Not only can Ethereum do what Bitcoin does (transfer money globally), it’s capable of much more – anyone can deploy publicly executable code onto the network. And because it’s so flexible, any computer program can run on Ethereum.",
```
minimalsm, line 16: Thanks for the thought here @0xKai27.

Whilst this is more specific, I worry that this might add unnecessary jargon to this page. For context, this page was purposely designed by @konopkja to be accessible to someone with zero Ethereum knowledge.

@konopkja I'll default to you on this decision, I know it's a tiny change but something about straw and camels 😬

konopkja: hm agree, making it simpler would be better and therefore i would prefer the original. Speaking of jargon - it did make me think whether we should use "deploy" or whether we should change it to something like "publish"

0xKai27: My two cents on "deploy" vs "publish".

Deploy has a more strategic connotation to it which fits the context better. For example, code has been deployed and is ready to be executed. 

Publish is more related to dissemination of information or making something known. This doesn't necessarily imply that the code that was published can actually be used. For example, code has been published and can be viewed at this address.

Loads of jargon and semantics but love how the ethereum.org pages are being maintained ;)

PR #9086, opened 2023-01-06
setup AdoptionChart for translation

corwintines:
## Description
- Setup AdoptionChart on [what-is-ethereum](https://ethereum.org/en/what-is-ethereum) for translation


src/intl/en/page-what-is-ethereum.json
```diff
@@ -96,5 +96,16 @@
   "page-what-is-ethereum-the-year-in-ethereum-2021-desc": "Jan 17, 2022 - Josh Stark and Evan Van Ness",
   "page-what-is-ethereum-atoms-institutions-blockchains": "Atoms, Institutions, Blockchains",
   "page-what-is-ethereum-atoms-institutions-blockchains-desc": "- Why blockchains matter?",
-  "page-what-is-ethereum-ethereum-in-numbers-title": "Ethereum in numbers"
+  "page-what-is-ethereum-ethereum-in-numbers-title": "Ethereum in numbers",
+  "adoption-chart-column-2010-label": "2010",
```
minimalsm, line 100: Do we need to translate years? cc: @lukassim

minimalsm: Context:

Screenshot 2023-01-06 at 09 59 27

lukassim: I think we can leave the years untranslated.
Having quickly looked through the top 5 non-Latin languages in Crowdin, all of them use Latin numbers for dates.

corwintines:
@minimalsm @lukassim fixed

minimalsm:
Build failed, rerunning

src/intl/en/page-what-is-ethereum.json
wackerow: I would use the Intl NumberFormat function to style numbers according to locale
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat
They shouldn't need to be translated.

wackerow (approved):

PR #9103, opened 2023-01-09
Migrate `AssetDownload.tsx` component to Chakra

benlazzero:
## Migrate the `AssetDownload.tsx` component to Chakra

#6374 

First migration I've done, I am happy to implement any change requests!


src/components/AssetDownload.tsx
```diff
@@ -102,34 +60,79 @@ const AssetDownload: React.FC<IProps> = ({
...
         {artistName && (
-          <Caption>
-            <ArtistSubtitle>
-              <Emoji text=":artist_palette:" mr={`0.5em`} />
+          <Flex
+            mb={4}
+            border="1px"
+            borderTop="none"
+            borderColor="white700"
+            p="0.5rem 1rem"
```
pettinarip, line 117:
```suggestion
            py={2}
            px={4}
```

src/components/AssetDownload.tsx
```diff
@@ -102,34 +60,79 @@ const AssetDownload: React.FC<IProps> = ({
...
+      display={{
+        base: shouldHide ? "none" : "flex",
+        lg: "flex",
+      }}
+      maxW="100%"
+      minW="170px"
+      flex="1 1 45%"
+      flexDirection="column"
+      justifyContent="space-between"
+      m="4"
```
pettinarip, line 73:
```suggestion
      m={4}
```

src/components/AssetDownload.tsx
```diff
@@ -102,34 +60,79 @@ const AssetDownload: React.FC<IProps> = ({
...
+      p={0}
+      opacity={shouldHide ? 0 : 1}
+    >
+      <Heading
+        as="h4"
+        fontSize={{ base: "1rem", md: "1.25rem" }}
```
pettinarip, line 79:
```suggestion
        fontSize={{ base: "md", md: "xl" }}
```

src/components/AssetDownload.tsx
```diff
@@ -102,34 +60,79 @@ const AssetDownload: React.FC<IProps> = ({
...
             {artistUrl && <Link to={artistUrl}>{artistName}</Link>}
-            {!artistUrl && <span>{artistName}</span>}
-          </Caption>
+            {!artistUrl && <Text m={0}>{artistName}</Text>}
+          </Flex>
         )}
-      </div>
-      <ButtonContainer>
+      </Box>
+      <Container mt={4} p={0} maxW="none">
```
pettinarip, line 129: I think there is no reason to use a `Container` here since we don't have any special width we want to keep on the content.
```suggestion
      <Box mt={4} p={0}>
```

src/components/AssetDownload.tsx
```diff
@@ -102,34 +60,79 @@ const AssetDownload: React.FC<IProps> = ({
   const Svg = svg
 
   return (
-    <Container shouldHide={shouldHide}>
-      <h4>{title}</h4>
-      <div>
-        {children && <ImageContainer>{children}</ImageContainer>}
+    <Container
```
pettinarip, line 63: Wondering why did you choose to use a `Container` in this case since the content width will always be 100%.

With `Box` you will not need to use `maxW` and `minW`.
```suggestion
    <Box
```

pettinarip (changes requested):
Good job @benlazzero thanks for the PR! 💪🏼 

The overall result looks great. I've left a few minor comments and a question about the using of the `Container` component.

src/components/AssetDownload.tsx
benlazzero: true, that makes sense.

src/components/AssetDownload.tsx
benlazzero: Me to! lol. Yeah Box is the cleaner component.

benlazzero:
Thanks glad to help! Tbh thinking back to when I wrote the code I wasn't all that sure about those differences. All the changes make total sense.

pettinarip (approved):
lgmt now @benlazzero thanks again 🚀 🚀

PR #9139, opened 2023-01-11
Add Correct package install

CamoCatX:
The change I made was simply in order to correctly install the package.

### DIFF
--- a/src/content/developers/tutorials/a-developers-guide-to-ethereum-part-one/index.md
+++ b/src/content/developers/tutorials/a-developers-guide-to-ethereum-part-one/index.md
@@ -95,7 +95,7 @@ $ pip install web3
 One more thing – we're going to simulate a blockchain later, which requires a couple more dependencies. You can install those via:
 
 ```bash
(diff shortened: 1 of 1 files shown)


CamoCatX:
> @CamoCatX what OS are you on? I'm wondering if this is OS specific problem, and if the change is a universally applicable one (at least on the big 3)
> 
>     * https://stackoverflow.com/questions/72822296/double-or-single-quotes-when-installing-specific-package-versions-with-pip
> 
>     * https://stackoverflow.com/questions/5226311/installing-specific-package-version-with-pip/33812968#33812968

I am on the Windows operation system.

corwintines:
@CamoCatX I think that would be best. This change is too OS specific for me to accept, but if you make that change I think that's reasonable :)

corwintines:
I was testing this change out locally, and this doesn't work on my machine (Linux and Mac). Going to close this out for now. Every machine is going to have its nuances, but I think for consistency sake, we should keep it how it was.

PR #9145, opened 2023-01-11
setup learn quizzes for translation

corwintines:
## Description
- Setup learn quiz question pool and components for translation


wackerow (approved):
Only suggestion/question here.. QuizWidget L222:
```ts
  const handleShare = (): void => {
    if (!quizData || !window) return
    trackCustomEvent({
      eventCategory: "Quiz widget",
      eventAction: "Other",
      eventName: "Share results",
    })
    const url = `https://ethereum.org${window.location.pathname}%23quiz` // %23 is # character, needs to added to already encoded tweet string
    const tweet =
      encodeURI(
        `I just took the "${quizData.title}" quiz on ethereum.org and scored ${correctCount} out of ${quizData.questions.length}! Try it yourself at `
      ) + url
    window.open(
      `https://twitter.com/intent/tweet?text=${tweet}&hashtags=${"ethereumquiz"}`
    )
  }
```

Is there a way we could set this up for translation as well? I think it could be tricky given it has variables in it and is then URI encoded before it's added to the link. @pettinarip Any thoughts here?

In the meantime I don't think that needs to block the rest of the PR. Nice job @corwintines!

corwintines:
Going to merge and we can resolve this separately like @wackerow mentioned

PR #9180, opened 2023-01-14
Change 'with' to 'for' for clarity

lukepuplett:
## Description

I'm changing "a user can trade Ethereum with our newly deployed ERC-20 token" to "a user can trade Ethereum _for_ our newly deployed ERC-20 token". It's a single, little word but it makes it much clearer.

In English we tend to say, "I'll trade my hat for that spade" and not "I'll trade my hat with that spade", which makes it sound like I'm using the spade to physically move the hat into someone else's possession. It's so subtle but when you're learning something completely new, you tend to analyse every word.

You could say, "a user can _buy_ Ethereum _with_ our newly deployed ERC-20 token" because you're buying it with ETH.

## Related Issue

N/A


minimalsm (approved):
Thanks @lukepuplett

PR #9197, opened 2023-01-17
Add docs on archive nodes

taxmeifyoucan:
Adds new doc with more details regarding archive node and running it. Also some updates and fixes in other nodes docs


src/content/developers/docs/nodes-and-clients/archive-nodes/index.md
```diff
@@ -0,0 +1,81 @@
+---
+title: Ethereum Archive Node
+description: 
+lang: en
+sidebar: true
+sidebarDepth: 2
+---
+
+An archive node is an instance of an Ethereum client with configuration to build an archive of all historical states. It is a useful tool for certain use cases but might be more tricky to run than a full node.
```
jmcook1186, line 9:
```suggestion
An archive node is an instance of an Ethereum client configured to build an archive of all historical states. It is a useful tool for certain use cases but might be more tricky to run than a full node. 
```

src/content/developers/docs/nodes-and-clients/archive-nodes/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+## What is an archive node
+
+To grasp the importance of an archive node, let's clarify the concept of "state."
+Ethereum can be referred to as *transaction-based state machine*. It consists of accounts and applications executing transactions which are changing their state. The global data with information about each account and contract is stored in trie database called state. This is handled by the execution layer (EL) client and includes:
+* Account balances and nonces 
+* Contract code and storage
+* Consensus-related data, e.g. Staking Deposit Contract
+
+To interact with the network, verify and produce new blocks, Ethereum clients have to keep up with chain tip and therfore the current state. EL client configured as a full node verifies and follows the latest state of the network. A full node temporarily caches only past few states, e.g. for last 128 blocks, for case of a chain reorg. But the recent state is all what client needs to verify incoming transactions and use the network.
```
jmcook1186, line 23:
```suggestion
To interact with the network, verify and produce new blocks, Ethereum clients have to keep up with the most recent changes (the tip of the chain) and therefore the current state. An EL client configured as a full node verifies and follows the latest state of the network but only caches the past few states, e.g. the state associated with the last 128 blocks, so it can handle chain reorgs and provide fast access to recent data. The recent state is all what client needs to verify incoming transactions and use the network.  
```

src/content/developers/docs/nodes-and-clients/archive-nodes/index.md
```diff
@@ -0,0 +1,81 @@
...
+* Account balances and nonces 
+* Contract code and storage
+* Consensus-related data, e.g. Staking Deposit Contract
+
+To interact with the network, verify and produce new blocks, Ethereum clients have to keep up with chain tip and therfore the current state. EL client configured as a full node verifies and follows the latest state of the network. A full node temporarily caches only past few states, e.g. for last 128 blocks, for case of a chain reorg. But the recent state is all what client needs to verify incoming transactions and use the network.  
+
+> You can imagine the state as a momentary network snapshot at a given block and the archive as a history replay 
+
+Historical states can be safely pruned because they are not necessary for the network to operate and it would be uselessly redundant for client to keep all out-of-date data. States that existed before recent block with all of their history are effectively thrown away. Only historical blockchain data are kept by a full node but they can be executed to generate a historical state.
```
jmcook1186, line 27:
```suggestion
Historical states can be safely pruned because they are not necessary for the network to operate and it would be uselessly redundant for client to keep all out-of-date data. States that existed before some recent block (e.g. 128 blocks before the head) are effectively thrown away. Full nodes only keep occasional historical snapshots they can use to regenerate older states on request. 
```

src/content/developers/docs/nodes-and-clients/archive-nodes/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+> You can imagine the state as a momentary network snapshot at a given block and the archive as a history replay 
+
+Historical states can be safely pruned because they are not necessary for the network to operate and it would be uselessly redundant for client to keep all out-of-date data. States that existed before recent block with all of their history are effectively thrown away. Only historical blockchain data are kept by a full node but they can be executed to generate a historical state. 
+
+However, this means that accessing a historical state on a full node consumes a lot of computation. The client might need to execute all past transactions and compute one historical state from genesis. Archive node solves this by storing not only the most recent states but every historical state created after each block. It basically makes a trade-off with bigger disk space requirement.
```
jmcook1186, line 29:
```suggestion
However, this means that accessing a historical state on a full node consumes a lot of computation. The client might need to execute all past transactions and compute one historical state from genesis. Archive nodes solve this by storing not only the most recent states but every historical state created after each block. It basically makes a trade-off with bigger disk space requirement. 
```

src/content/developers/docs/nodes-and-clients/archive-nodes/index.md
```diff
@@ -0,0 +1,81 @@
...
+Before starting your own archive node, learn about the differences between the clients and especially the various [hardware requirements](/developers/docs/nodes-and-clients/run-a-node/#requirements). Most clients are not optimized for this feature and their archives requirs more than 12TB of space. In contrast, implementations like Erigon and Akula can store the same data in under 3TB which makes them the most effective way of running an archive node.
+
+## Recommended practices
+
+Apart from general [recommendations for running](developers/docs/nodes-and-clients/run-a-node/), an archive node may be more demanding on hardware and maintanance. Considering its [key features](https://github.com/ledgerwatch/erigon#key-features), the most practical approach is using[Erigon](https://ethereum.org/en/developers/docs/nodes-and-clients/#erigon) client implementation.
+
+### Hardware 
+
+Always make sure to verify hardware requirements in given mode in client's documentation. 
+et
```
jmcook1186, line 64:
```suggestion
```

src/content/developers/docs/nodes-and-clients/archive-nodes/index.md
```diff
@@ -0,0 +1,81 @@
...
+The biggest requirement which comes with archive node is the disk space. Depending on client, it varies from 3 to 12TB. Even if HDD might be considered a better solution for large amounts of data, syncing it and constantly updating the head of the chain will require SSD drives. SATA drives are good enough but it should be a reliable quality, at least TLC. Disks can be fitted into a desktop computer or a server with enough slots. Such dedicated devices are ideal for running high uptime node. It's totally possible to run it on a laptop but the portability will come at an aditional cost.
```
jmcook1186, line 65:
```suggestion
The biggest requirement which comes with archive node is the disk space. Depending on client, it varies from 3 to 12TB. Even if HDD might be considered a better solution for large amounts of data, syncing it and constantly updating the head of the chain will require SSD drives. [SATA](https://www.cleverfiles.com/help/sata-hard-drive.html) drives are good enough but it should be a reliable quality, at least [TLC](https://blog.synology.com/tlc-vs-qlc-ssds-what-are-the-differences). Disks can be fitted into a desktop computer or a server with enough slots. Such dedicated devices are ideal for running high uptime node. It's totally possible to run it on a laptop but the portability will come at an additional cost.  
```

src/content/developers/docs/nodes-and-clients/archive-nodes/index.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Hardware 
+
+Always make sure to verify hardware requirements in given mode in client's documentation. 
+et
+The biggest requirement which comes with archive node is the disk space. Depending on client, it varies from 3 to 12TB. Even if HDD might be considered a better solution for large amounts of data, syncing it and constantly updating the head of the chain will require SSD drives. SATA drives are good enough but it should be a reliable quality, at least TLC. Disks can be fitted into a desktop computer or a server with enough slots. Such dedicated devices are ideal for running high uptime node. It's totally possible to run it on a laptop but the portability will come at an aditional cost.  
+
+All of the data needs to be fit in one volume, therfore disks has to be joint, e.g. with RAID0 or LVM. It might be also worth considering using [ZFS](https://en.wikipedia.org/wiki/ZFS) as it supports "Copy-on-write" which ensures data is correctly written to the disk without any low level errors.
```
jmcook1186, line 67:
```suggestion
All of the data needs to be fit in one volume, therfore disks have to be joined, e.g. with [RAID0](https://en.wikipedia.org/wiki/Standard_RAID_levels#RAID_0) or [LVM](https://web.mit.edu/rhel-doc/5/RHEL-5-manual/Deployment_Guide-en-US/ch-lvm.html). It might be also worth considering using [ZFS](https://en.wikipedia.org/wiki/ZFS) as it supports "Copy-on-write" which ensures data is correctly written to the disk without any low level errors. 
```

src/content/developers/docs/nodes-and-clients/archive-nodes/index.md
```diff
@@ -0,0 +1,81 @@
...
+et
+The biggest requirement which comes with archive node is the disk space. Depending on client, it varies from 3 to 12TB. Even if HDD might be considered a better solution for large amounts of data, syncing it and constantly updating the head of the chain will require SSD drives. SATA drives are good enough but it should be a reliable quality, at least TLC. Disks can be fitted into a desktop computer or a server with enough slots. Such dedicated devices are ideal for running high uptime node. It's totally possible to run it on a laptop but the portability will come at an aditional cost.  
+
+All of the data needs to be fit in one volume, therfore disks has to be joint, e.g. with RAID0 or LVM. It might be also worth considering using [ZFS](https://en.wikipedia.org/wiki/ZFS) as it supports "Copy-on-write" which ensures data is correctly written to the disk without any low level errors. 
+
+For more stability and security in preventing accidental database corruption, especially in a professional setup, consider using ECC memory if your system supports it. The size of RAM is generally adviced the same as for a full node but more can help speed up the synchronization.
```
jmcook1186, line 69:
```suggestion
For more stability and security in preventing accidental database corruption, especially in a professional setup, consider using [ECC memory](https://www.systools.in/blog/fix-ecc-error-hard-drive/) if your system supports it. The size of RAM is generally advised to be the same as for a full node but more RAM can help speed up the synchronization. 
```

jmcook1186 (changes requested):
This is great - thank you! I think this will be a really useful page. I've made some small suggestions, mostly adding links and fixing a few typos, but also some small text adjustments for readability/clarity - please take a look and commit if you agree.

src/content/developers/docs/nodes-and-clients/archive-nodes/index.md
taxmeifyoucan: Thanks for adding this, it's explained nicely but I would keep the point that full nodes keep the blockchain data so it doesn't give impression it's just snapshots. 

"Full nodes only keep historical blockchain data (blocks and transactions) and occasional historical snapshots.."

src/content/developers/docs/nodes-and-clients/archive-nodes/index.md
taxmeifyoucan: This link seems wrong, we can use wiki https://en.wikipedia.org/wiki/ECC_memory

taxmeifyoucan:
Thank you for such a quick review and fixes! Changes applied and I commit the rest I commented if you agree

jmcook1186 (approved):
great - thanks!

corwintines (approved):
Thanks, @taxmeifyoucan! I just cleaned up some of the text, and added links in the docs nav.

PR #9214, opened 2023-01-18
Fix typo in the glossary causing broken link

ezrannetttty:
as described in the title


minimalsm (approved):
Good catch! Thanks for the PR @ezrannetttty ❤️

PR #9217, opened 2023-01-18
Added beaconcha.in block explorer

Sanketh-Maturi:
## Description

Added the new open source beaconcha.in block explorer to the block explorers and services list.

### DIFF
--- a/src/content/developers/docs/data-and-analytics/block-explorers/index.md
+++ b/src/content/developers/docs/data-and-analytics/block-explorers/index.md
@@ -13,12 +13,12 @@ You should understand the basic concepts of Ethereum so you can make sense of th
 
 ## Services {#services}
 
-- [Beaconcha.in](https://beaconcha.in/)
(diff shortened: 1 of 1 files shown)


wackerow (commented):
Thanks @Sanketh-Maturi! Agree we should get Beaconchain.in listed.

Only question this raises for me is the ordering of projects. @corwintines any thoughts on this? Should we just make these alphabetical order by default?

minimalsm:
We should probably separate these by chain, but we can do that later. I'd say let's keep the most used (therefore most useful?) blockexplorer (Etherscan) at the top for now, but no strong opinion here.

corwintines:
I don't have a strong opinion. Maybe Etherscan then alphabetical for now.

corwintines:
@wackerow @minimalsm ill push up a change for that tomorrow and merge if theres no opposition.

corwintines (approved):

PR #9231, opened 2023-01-18
update image name in flash command ethonarm

juliencarponcy:
Also make device naming more consistent with the above: >> sdxxx


corwintines (approved):
I think this makes sense. 

@jmcook1186 could you confirm as well?

juliencarponcy:
I was about to say we could also provide a way for the user to find out which device they should dd on. As it is stated above that it is important to get the correct device.

At the very least, the proposed change with the <> and the xxx will prevent any accidental write on another drive by the user, since there is no longer /dev/mmcblk0 and /dev/<sdxx> will just throw an error forcing them to think twice. 

For the above, is it worth it adding a couple of lines about the df command like df -h to have a quick idea on which drive to write on based on capacity?

jmcook1186 (approved):
lgtm, thanks!

PR #9261, opened 2023-01-21
Implement react i18next

pettinarip:
This PR implements react-i18next and creates & loads the locales in a different way than #8866 to avoid having memory consumption problems.

## Description

_Preview deploy:_ https://ethereumorgwebsitei18nrework.gatsbyjs.io/en/

In short, instead of `mergeTranslations` in one big file, all the json files under `src/intl` are loaded as they are, keeping the same namespaces. Then, on each page, we query only the necessary namespace that that page needs.

To do this efficiently, we had to reorder some translations from one namespace to another. Basically to avoid loading an entire namespace just for 1 translation. For example, adding new strings on `common.json` is very expensive since that file is loaded on all the pages we have (at the moment 12k pages), so I tried my best to reduce that file to the minimum.

Other things done in this PR:
- `createLocales` script was completely redone. Check `docs/locale-process.md` for more information
- `mergeTranslations` script is not more needed and it was deleted
- new utility `unusedTranslations` script added
- new utility `reorgIntlFiles` script added
- `yarn.lock` was regenerated with latest packages versions (this one mitigates a wall of warnings we had on dev env)
- refactored page queries to fetch locales using the correct namespaces

----

Note: for ease to review purposes, I didn't run the `reorgIntl` script yet. I'll include that on a different PR. What that script is going to do is to apply the same reorgs I did on `src/intl/en` to all the other languages.

----

Note2: stress tests were executed here #9455


src/templates/docs.tsx
```diff
@@ -243,7 +243,7 @@ export const query = graphql`
     locales: allLocale(
       filter: {
         language: { in: $languagesToFetch }
-        ns: { in: ["src-templates-docs", "components", "common"] }
+        ns: { in: ["page-developers-docs", "common"] }
```
pettinarip, line 246: Now the namespaces have the same name as the files under `src/intl`.

src/templates/static.tsx
```diff
@@ -189,8 +189,6 @@ const StaticPage = ({
   const { editContentUrl } = siteData.siteMetadata || {}
   const absoluteEditPath = `${editContentUrl}${relativePath}`
 
-  const slug = mdx.fields?.slug || ""
```
pettinarip, line 192: This `slug` is incorrrect. This `slug` is always the `/en/` version. To fix this, now we are getting it from the `pageContext`.

src/intl/en/page-developers-docs.json
```diff
@@ -131,5 +131,13 @@
   "page-calltocontribute-link": "documentation template",
   "page-calltocontribute-link-2": "Discord server",
   "page-calltocontribute-span": "Edit page",
-  "page-calltocontribute-title": "Help us with this page"
+  "page-calltocontribute-title": "Help us with this page",
+  "layer-2-arbitrum-note": "Fraud proofs only for whitelisted users, whitelist not open yet",
+  "layer-2-boba-note": "State validation in development",
+  "layer-2-metis-note": "Fraud proofs in development",
+  "layer-2-optimism-note": "Fault proofs in development",
```
corwintines, line 138: I think this PR needs these translation files strings looked at. Some of these files don't make sense for some of these strings. Let me know if you want to sort through this together

src/pages/upgrades/get-involved/index.tsx
```diff
@@ -541,9 +541,7 @@ export const query = graphql`
     locales: allLocale(
       filter: {
         language: { in: $languagesToFetch }
-        ns: {
-          in: ["src-pages-upgrades-get-involved-index", "components", "common"]
-        }
+        ns: { in: ["page-upgrades-get-involved", "common"] }
```
corwintines, line 544: Screen Shot 2023-02-15 at 11 02 11 PM

This missing translation string is in `page-upgrades-index` which isn't imported here. Looks like it's only used in this file as well, so we could consider just moving this string. This string probably isn't needed anyways, Lodestar is now in production. 

I think 2 best options are:

1 - Move this `consensus-client-under-review` into `page-upgrades-get-involved`
2 - Make a PR to dev, which can be merged back into here, which removes this string and its use. Created an issue for this: https://github.com/ethereum/ethereum-org-website/issues/9498

corwintines (changes requested):
Nice one! This really improves load times from some testing. Noticed a couple things. Happy to sync up and talk about this.

src/intl/en/page-developers-docs.json
corwintines: Note this is in a few json files, just calling out the pattern.

src/pages/upgrades/get-involved/index.tsx
pettinarip: Good catch. Thanks for creating that issue. Probably, it will be easier to implement that PR and clean this after merging dev (option 2).

Once we are done with the review here, I'll update this branch with dev.

src/intl/en/page-developers-docs.json
pettinarip: Not sure if I follow you on this. LMK and we can get on a call.

src/pages/upgrades/get-involved/index.tsx
corwintines: https://github.com/ethereum/ethereum-org-website/pull/9537

Heres a PR for this

corwintines (approved):
LGTM, approving since dev will be merged into this branch that will include https://github.com/ethereum/ethereum-org-website/pull/9537

src/intl/en/page-developers-docs.json
corwintines: Ignore this, I understand why this is here now after a quick chat.

pettinarip:
Thanks @konopkja for the review 💪🏼 

1. Good catch.
2. This branch is pretty old. Soon I'll update it and we should see the same buttons as in prod 👍🏼 
3. Good catch.

tldr, I'll fix those missing translations & update the branch.

Issue #9307, opened 2023-01-26
Staking page hero banner isn't full width on XL displays


corwintines:
### Describe the bug

![Screen Shot 2023-01-25 at 8 30 10 PM](https://user-images.githubusercontent.com/15589226/214753003-a30e43ec-37cf-4c97-a47f-68becae96d1f.png)

The banner should be full width even above 1400px width banner

### To reproduce

1 - Go to [staking page](https://ethereum.org/en/staking) above 1400px width

You will see the banner isn't full-width on a display larger than 1400px width

### Expected behavior

1 - Go to [staking page](https://ethereum.org/en/staking) above 1400px width

You will see the banner isn't full-width on a display larger than 1400px width

### Screenshots

![Screen Shot 2023-01-25 at 8 30 10 PM](https://user-images.githubusercontent.com/15589226/214753380-ee48fd13-15bd-4763-a932-b9162dbab364.png)

MrJithil:
Added pull request. Please check.

PR #9308, opened 2023-01-26
Fix stakes width issues on xl

MrJithil:
## Description
Stake page hero banner is not showing fullwidth in xl screens

## Related Issue
#9307


corwintines (approved):
Thanks @MrJithil!

I removed an unnecessary style, but works now!

Issue #9313, opened 2023-01-26
`Translate page` button text is empty on mobile device


MrJithil:
### Describe the bug

`Translate page` button text is empty on `TranslationBanner ` mobile device

### To reproduce

1. Open in Safari browser on iPhone
2. Go to https://ethereum.org/de/contributing/#how-to-update-content
3. Observe that the `View Translation` button text is invisible or empty

### Expected behavior

`View Translation` button text should be visible

### Screenshots

https://user-images.githubusercontent.com/26359740/214821081-a9b1b2f1-0590-4cdc-86d9-b49644caaaab.MOV

### Smartphone (please complete the following information)

Same as above

MrJithil:
Working on it

MrJithil:
Fixes #9315

MrJithil:
Changes merged. Closing the issue.

Issue #9312, opened 2023-01-26
`Translate page` button text is empty after click


MrJithil:
### Describe the bug

Once we clicked the `Translate page` button in `TranslationBanner` component is empty or not visible after a click.

### To reproduce

1. Go to https://ethereum.org/de/contributing/translation-program/
2. Click on `Seite übersetzen` button on the `TranslationBanner` popup
3. Observe that the button text become invisible

### Expected behavior

Button should be visible after click/highlight

### Screenshots

https://user-images.githubusercontent.com/26359740/214819089-307bb76f-4038-4868-8608-dcda62091c3b.mov

MrJithil:
Working on it

MrJithil:
Changes merged. Closing the issue.

PR #9315, opened 2023-01-26
fix the translate page button not showing the text

MrJithil:
## Description
The language which loads to should be `en` as we are requesting the user to see the latest `en` version and seeking their help to translate the page. It will fix the empty text in the button issue as well.

Please advice, we should not load the `en` version on `Translate Page` click. So that I can make changes accordingly.

## Related Issue
#9312 #9313

### DIFF
--- a/src/components/ButtonLink.tsx
+++ b/src/components/ButtonLink.tsx
@@ -7,7 +7,7 @@ export interface IProps extends ILinkProps, ButtonProps {}
 
 const ButtonLink: React.FC<IProps> = ({ children, ...props }) => {
   return (
-    <Button as={Link} textDecoration="none" {...props}>
+    <Button as={Link} textDecoration="none" activeStyle={{}} {...props}>
       {children}
     </Button>
   )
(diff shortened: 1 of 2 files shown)


pettinarip (changes requested):
I think this issue is an issue from the `ButtonLink` component. The `ButtonLink` component is a Link but is styled as a Button. Having said that, a Button styled component should not have an `active` style or state.

So, I think the correct way to fix this problem is to override the `activeStyles` from the `ButtonLink` component, making all the `ButtonLink`s the same.

LMK what do you think.

@corwintines this is related to the problem we saw on the guides PR.

MrJithil:
> I think this issue is an issue from the `ButtonLink` component. The `ButtonLink` component is a Link but is styled as a Button. Having said that, a Button styled component should not have an `active` style or state.
> 
> So, I think the correct way to fix this problem is to override the `activeStyles` from the `ButtonLink` component, making all the `ButtonLink`s the same.
> 
> LMK what do you think.
> 
> @corwintines this is related to the problem we saw on the guides PR.

There are many instance of `ButtonLink` all are working well. So, removing the `activeStyle` from the `Link` may introduce many changes. To test and ensure all these instance is hard task.

corwintines:
Thanks so much @MrJithil!

corwintines:
@all-contributors can you add @MrJithil for code

Issue #9323, opened 2023-01-27
Upgrade Storybook to v7 and Component Story Format to v3


TylerAPfledderer:
_The following description and details is related to these upgrades as of `v7.0.0-beta.35` of Storybook and subject to updates and changes post preliminary findings._

## Overview
This issue is to track discussion in upgrading Storybook to v7 and the Story Component formatting to v3. This may be to resolve with a PR immediately, or to be kept as record for implementation at a later time.

## Setup

The following steps can be taken (in order):
1. Run `npx storybook@next upgrade --prerelease`, and select `yes` to all prompts
    - Run `storybook-binary` migration which adds `storybook` as a dev dependency. (Different than `@storybook/*`)
    - Run `sb-scripts` to update the build and dev scripts for Storybook
    - Run `newFrameworks` to replace storybook webpack deps
    - Run `autodocsTrue` to opt in on having autodocs generated for every story.
    - Run `missing-babelrc` to add a babel config as the settings in SB are no longer out of the box, and now looks at a project's config.
2. The final request to add babel config has its own prompts which can be given a yes to each along with installation of `@babel/preset-env`
    - Installation of `@babel/preset-typescript`
    - Installation of `@babel/preset-react`
3. With No. 1, the `main.js` file in `.storybook` will be updated. Additional changes need to happen manually.
    - The `webpack` option will need to be removed as it breaks compiling and also does not appear to be needed.
    - Under the `features` option, `emotionAlias` is no longer needed
4. Run `npx storybook@next migrate csf-2-to-3 --glob="**/*.stories.tsx"` to update the current story files to CSF3

## Component Story Format 3
In essence, this new version allows for each story to be an object.

For example, a story from [Button.stories.tsx](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Button/Button.stories.tsx) will go from this:
```tsx
const Template: ComponentStory<typeof Button> = (args) => <Button {...args} />

export const Solid = Template.bind({})
Solid.args = {
  children: "What is Ethereum?",
}
```
to this:
```tsx
// No Template!
export const Solid: StoryObj<typeof Button> = {
  args: {
    children: "What is Ethereum?",
  },
}
```
Or with [Card.stories.tsx](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Card/Card.stories.tsx):
```tsx
const defaultProps: IProps = {
  emoji: ":woman_student:",
  title: "Learn Ethereum development",
[...]

PR #9324, opened 2023-01-27
Upgrade storybook version to 7.0.0-beta.51

MrJithil:
## Description

Upgrade storybook version to 7.0.0-beta.36

Replaced
* `ComponentMeta` with `Meta`
* `ComponentStory` with `StoryFn` and `StoryObj`
 
## Related Issue

#9323


pettinarip:
Thanks for the PR @MrJithil. I've converted it to a Draft to avoid someone merging it until we are completely sure that we want to make this upgrade.

.babelrc.json
```diff
@@ -0,0 +1,16 @@
...
+        "targets": {
+          "chrome": 100
+        }
+      }
+    ],
+    "@babel/preset-typescript",
+    "@babel/preset-react"
+  ],
+  "plugins": []
+}
```
pettinarip, line 16: Would it be possible to rename this file to `.babelrc.storybook.json` to avoid confusion? and specifically call it from the `stroybook`s scripts?

The docs seems to be outdated for the CLI since they keep calling it `start-storybook` https://storybook.js.org/docs/7.0/react/api/cli-options

.storybook/main.js
```diff
@@ -16,30 +15,43 @@ module.exports = {
...
     "@chakra-ui/react": {
       disable: true,
     },
   },
-  core: {
-    builder: "webpack5",
-  },
   features: {
     storyStoreV7: true,
     emotionAlias: false,
```
pettinarip, line 29: Are these two options needed on v7?

src/components/BannerNotification/BannerNotification.stories.tsx
```diff
@@ -1,18 +1,18 @@
...
 /**
  * Story taken from PostMergeBanner component
  * and content from `../../content/developers/tutorials/hello-world-smart-contract-fullstack/index.md`
  */
-export const PostMergeBanner: ComponentStory<typeof BannerNotification> = (
-  args
-) => <BannerNotification {...args} />
+export const PostMergeBanner: StoryObj<typeof BannerNotification> = (args) => (
+  <BannerNotification {...args} />
+)
```
pettinarip, line 15: This shouldn't be necessary anymore, right?

pettinarip (commented):
Thanks @MrJithil for the PR 💪🏼 

As @TylerAPfledderer mentioned on the issue, would you mind bumping the storybook version to the latest one?

.storybook/main.js
TylerAPfledderer: @pettinarip

Both can be removed. Per migration docs `storyStoreV7` is set to true by default, and `emotionAlias` is no longer needed.

MrJithil:
Working on it.

.babelrc.json
TylerAPfledderer: A babel config file is now needed for the parsing to work in storybook, as there is no longer a configuration out of the box and uses the project's configuration instead.

This config file is the default generated with the migration process run in the CLI. Without it and the added deps, an error throws pointing to `./node_modules/@storybook/builder-webpack5/node_modules/babel-loader/lib/index.js`

Here are the [details from the migration docs](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#babel-mode-v7-exclusively)

I have not seen anything in the Gatsby discord or in their GH issues regarding any suggested alternatives to this generated config.

src/components/BannerNotification/BannerNotification.stories.tsx
TylerAPfledderer: This is for typing in an object context instead of JSX. So this should actually look like...
```tsx
export const PostMergeBanner: StoryObj<typeof BannerNotification> = {
  args: {
    shouldShow: true,
    justify: "center",
    textAlign: "center",
    sx: {
      "& p": {
        maxWidth: "100ch",
        m: 0,
        p: 0,
      },
      "& a": {
        textDecor: "underline",
      },
    },
    children: (
      <p>
        This tutorial is out of date after the merge and may not work. Please
        raise a PR if you would like to contribute.
      </p>
    ),
  },
}
```

.babelrc.json
pettinarip: That's ok. Yeah, I was asking just to add clarity. At least wouldn't be clear to me that this new `babelrc` file is only used for storybook and not for Gastby for example.

TylerAPfledderer: Right. Renaming the file would be the way to go here.

MrJithil: Changed the way of loading it.

src/components/BannerNotification/BannerNotification.stories.tsx
MrJithil: Great.

.storybook/main.js
MrJithil: Removed.

MrJithil:
All suggestions done.

.storybook/main.js
```diff
@@ -1,5 +1,5 @@
 const { propNames } = require("@chakra-ui/react")
-
+const { babelConfig } = require("./babel-storybook-config")
```
pettinarip, line 2: Nice! thanks for doing that

pettinarip (approved):
GJ @MrJithil

JithilNCA:
Thank You.

PR #9331, opened 2023-01-28
Additional free providers + JS library clarification

shelleyolivia:
Hey there, 

I added Moralis and Alchemy as another free option for a hosted node. Also, web3.js is used along with ethers.js today so I clarified the language so that beginning developers would not read the bottom section and think that web3.js is the predominant JS library today as this isn't the case.


corwintines (approved):
Thanks, @shelleyolivia!

PR #9340, opened 2023-01-29
Update community-events.json

yonathanavila:
Add Hash House Medellín at Ethereum community events

## Description
- We want to add Hash house Medellin as a community inside the webpage of Ethereum community events 
- https://ethereum.org/en/community/events/
- https://www.meetup.com/hash-house-medellin/
## Related Issue
N/A


corwintines:
Thanks, @yonathanavila. Generally, we leave the events for conferences as this list is hard to maintain for each individual event like this. But, we have added this meetup to the meetups list, so still shown on this page :)

PR #9341, opened 2023-01-29
Fix missing syntax

coincashew:
## Description
Missing \ at EOL, otherwise command will error.


corwintines (approved):
Thanks @coincashew!

PR #9346, opened 2023-01-29
update aztec links

critesjosh:
I do dev rel with aztec. Updating links to docs and the explorer

## Description

Updating aztec links


corwintines (approved):
Thank you @critesjosh!

PR #9349, opened 2023-01-30
fix typo

marcopedrinazzi:


Fqc301033 (approved):

corwintines:
Thanks @marcopedrinazzi. Both `broadcast` or `broadcasted` are acceptable in this situation. I think broadcast is preferred, and this would also potentially knock back translations, so i'm going to close this in favor of broadcast.

Issue #9353, opened 2023-01-30
Implement UI library - Migrate pages & templates


pettinarip:
*This is part of our [Implement UI library](https://github.com/ethereum/ethereum-org-website/issues/6374) epic.*

Our final step on the UI migration! time to migrate the pages & templates.

### **How to contribute**

If you’d like to help with this migration, please follow these steps:

1. Leave a comment here asking which file you would want to work on.
2. We assign you to the file (by adding your handle next to the file name in the list).
    - Read and follow our [UI migration guide](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/chakra-migration-guide.md) for some tips & guidance.
    - **If you were assigned to multiple files** then create a separate PR for each component. It will make the testing process much simpler and more organized.
3. Once you finish it (PR merged), we’ll mark it as done.
4. Repeat!

- [x] `src/pages/`
  - [x] wallets/find-wallet.tsx @soheil555 #9390
  - [x] contributing/translation-program/acknowledgements.tsx @amit-ksh #9397
  - [x] contributing/translation-program/contributors.tsx @amit-ksh #9369
  - [x] community.tsx @soheil555 #9410
  - [x] staking/index.tsx @mlCode111 #9398
  - [x] staking/deposit-contract.tsx @mlCode111 #9549
  - [x] stablecoins.tsx @alex1092 #9635
  - [x] learn/index.tsx @soheil555 #9422
  - [x] assets.tsx @MrJithil #9406
  - [x] bug-bounty.tsx @Seek4samurai #9475
  - [x] languages.tsx @MahendraBishnoi29 #10014
  - [x] index.tsx @TylerAPfledderer #9364
  - [x] 404.tsx #10318
  - [x] get-eth.tsx @vdusart #9627
  - [x] run-a-node.tsx #10534
  - [x] upgrades/get-involved/index.tsx @amit-ksh #9572
  - [x] upgrades/vision.tsx @amit-ksh #9524
  - [x] upgrades/index.tsx @amit-ksh #9473
  - [x] developers/local-environment.tsx @soheil555 #9583
  - [x] developers/index.tsx @Seek4samurai #9933
  - [x] developers/tutorials.tsx @vdusart #10104
  - [x] developers/learning-tools.tsx @Seek4samurai #10348
  - [x] layer-2.tsx @TylerAPfledderer #9404
  - [x] studio.tsx @rishisundar #9518

- [x] `src/pages-conditional/`
[...]

PR #9369, opened 2023-02-01
refactor: contributors page to chakra

amit-ksh:
## Description

Migrate the `contributors` page from emotion to Chakra.

## Related Issue

#9353


src/pages/contributing/translation-program/contributors.tsx
```diff
@@ -10,37 +18,13 @@ import Breadcrumbs from "../../../components/Breadcrumbs"
...
-  @media (max-width: ${(props) => props.theme.breakpoints.m}) {
-    grid-template-columns: 1fr 1fr;
-  }
-
-  @media (max-width: ${(props) => props.theme.breakpoints.s}) {
-    grid-template-columns: 1fr;
-  }
-`
+const Content = (props: BoxProps) => <Box py={4} px={8} w="full" {...props} />
+const ContentHeading = (props: TextProps) => <Text as="h2" {...props} />
```
TylerAPfledderer, line 27: Chakra provides a `Heading` component that should be used here.

This does mean that you will have to explicitly set the fontSizes, fontWeights, and lineHeights set in the [global styles](https://github.com/ethereum/ethereum-org-website/blob/ae5df7a0997720039d9124fb89c3a16c63af59b4/src/%40chakra-ui/gatsby-plugin/styles.ts#L67-L111) to override the defaults. But this will help with the transition to the new Design System and updating the theme config. 

By default, the `h2` element is rendered in `Heading` so that can be removed. And since the `lineHeight` value is the same for all headings, it can be defined here. :)
```suggestion
const ContentHeading = (props: HeadingProps) => <Heading lineHeight={1.4} {...props} />
```

src/pages/contributing/translation-program/contributors.tsx
```diff
@@ -97,56 +81,69 @@ const Contributors = ({
 
       <Content>
         <Breadcrumbs slug={location.pathname} />
-        <h1>
+        <ContentHeading as="h1">
```
TylerAPfledderer, line 84: To go with the comment on the [ContentHeading component](https://github.com/ethereum/ethereum-org-website/pull/9369/files#diff-fea7d7fee9efeaf79d0def83b648b3851e5ae6ef52069d898bcb37b54fb24341R27)

```suggestion
        <ContentHeading
          as="h1"
          fontSize={{ base: "2.5rem", md: "3rem" }}
          fontWeight={700}
        >
```

src/pages/contributing/translation-program/contributors.tsx
```diff
@@ -97,56 +81,69 @@ const Contributors = ({
...
           <Translation id="page-contributing-translation-program-contributors-title" />
-        </h1>
-        <h4>
-          <strong>
+        </ContentHeading>
+        <ContentHeading as="h4">
```
TylerAPfledderer, line 87: To go with the comment on the [ContentHeading component](https://github.com/ethereum/ethereum-org-website/pull/9369/files#diff-fea7d7fee9efeaf79d0def83b648b3851e5ae6ef52069d898bcb37b54fb24341R27)

```suggestion
        <ContentHeading
          as="h4"
          fontSize={{ base: "md", md: "xl" }}
          fontWeight={500}
        >
```

src/pages/contributing/translation-program/contributors.tsx
```diff
@@ -97,56 +81,69 @@ const Contributors = ({
...
+        <Text>
           <Translation id="page-languages-interested" />{" "}
           <Link to="/contributing/translation-program/">
             <Translation id="page-languages-learn-more" />
           </Link>
           .
-        </p>
-        <h2>
+        </Text>
+        <ContentHeading as="h2">
```
TylerAPfledderer, line 109: To go with the comment on the [ContentHeading component](https://github.com/ethereum/ethereum-org-website/pull/9369/files#diff-fea7d7fee9efeaf79d0def83b648b3851e5ae6ef52069d898bcb37b54fb24341R27)

```suggestion
        <ContentHeading fontSize={{ base: "2xl", md: "2rem" }} fontWeight={600}>
```

src/pages/contributing/translation-program/contributors.tsx
```diff
@@ -97,56 +81,69 @@ const Contributors = ({
...
+        <UnorderedList
+          display="grid"
+          gridTemplateColumns={{
+            base: "1fr",
+            sm: "repeat(2, 1fr)",
+            md: "repeat(3, 1fr)",
+            lg: "repeat(4, 1fr)",
+            xl: "repeat(6, 1fr)",
+          }}
+        >
```
TylerAPfledderer, line 121: We can utilize the `as` prop here to combine two Chakra Props. I like using `SimpleGrid` here to give you some shortcut props to simplify the readability. And because we are changing the number of columns through many breakpoints from the smallest with no skips, we can use array bracketing to simplify the syntax further.

Also, `UnorderedList` by default renders `margin-inline-start` with its own value which adds greater specificity then `margin-left`. This is overriding a `margin-left` value of `1.45rem`  from global styles with `1em`. 

```suggestion
        <SimpleGrid as={UnorderedList} columns={[ 1, 2, 3, 4, 6 ]} ms="1.45rem">
```

TylerAPfledderer (changes requested):
Hey @amit-ksh!

I have a few thoughts here for improvement. :)

src/pages/contributing/translation-program/contributors.tsx
pettinarip: Interesting approach! much cleaner, thanks for sharing this.

pettinarip (approved):
Gj @amit-ksh! 💪🏼 thanks for the PR

Issue #7504, opened 2022-08-19
Add layerswap.io to "How to get onto a layer 2" section


arkoc:
Is your feature request related to a problem? Please describe.

On the Layer2 page on ethereum.org, under the "How to get onto a layer 2" section, I noticed that the drop-down menus to select the layer 2 network have limited options, when in fact almost all CEXs can be bridged to almost any L2 Exchange. 

Describe the solution you'd like

With Layerswap.io, all of those layer 2 networks can be reached by nearly any CEX. Layerswap.io is instant and gasless, users only pay a small fee up to 10x cheaper than a traditional transfer. It would be nice to list Layerswap as an option in "Check exchanges that support L2" under the "How to get onto a layer 2" section.

Describe alternatives you've considered

Additional context
<img width="440" alt="image" src="https://user-images.githubusercontent.com/106245135/185391506-18d79078-a7a9-4881-8c21-e1ca1fdf2380.png">
<img width="420" alt="image" src="https://user-images.githubusercontent.com/106245135/185391569-801dd602-d51a-4421-a6d3-d2b9a1c162a6.png">

minimalsm:
Hey @arkoc, sorry for the delay here. I think adding Layerswap in some capacity makes sense, but it will require some design thinking. 

However, I do think we should differentiate native exchange support via simple transfers from interacting with smart contracts/bridges. Any thoughts @corwintines?

PR #9371, opened 2023-02-01
Add layerswap.io to "How to get onto a layer 2" section

corwintines:
## Description
- Adds support for third party onboarding options from cex to layer 2
- Adds data for Layerswap

## Related Issue
Fixes: https://github.com/ethereum/ethereum-org-website/issues/7504


src/components/Layer2/Layer2Onboard.tsx
```diff
@@ -224,6 +238,51 @@ const Layer2Onboard: React.FC<IProps> = ({
...
+      return {
+        label: cexOnboard.name,
+        value: cexOnboard.name,
+        cexOnboard: cexOnboard,
+      }
+    }
+  )
+
+  const formatGroupLabel = (data) => {
+    console.log(data)
```
wackerow, line 252:
```suggestion
```

wackerow (approved):
LGTM!

Design/layout comment, it's a little funky when the LayerSwap block comes down on mobile with the ETH glyph tucked right next to it. Doesn't need to block this, but just noting maybe we could find a way to tidy this layout up.

PR #6585, opened 2022-06-06
[Build perf] Optimize images

pettinarip:

PR #9374, opened 2023-02-02
Preoptimize source images under `/assets`

pettinarip:
Preoptimize source images to reduce cpu and memory consumption on build process. Partial work taken from #6585.

## Description

Idea taken from https://www.gatsbyjs.com/docs/preoptimizing-images/

Things achieved in this PR:
- Added script to optimize images
- Resized most of the images that are used in the site
- Fixed queries to stop using unnecessarily `FULL_WIDTH` images

### New source images sizes

To determine what size the source images should have, I followed this logic:
1. Look for all the usages of `image.png` in the codebase
2. Check all the different resolutions being used in each case
3. Pick the highest resolution used
4. Resize the image to: `highestResolutionUsed * 2` to support 2x displays


wackerow:
@pettinarip Noticing that a handful of these ended up with larger files. In general, do we still want to update these if the optimization led to a larger file?

image

pettinarip:
Good catch @wackerow! that totally breaks the purpose of this haha. I'll double-check those and revert if needed...uhm 🤔 wondering why that happened...

corwintines:
Overall in favor, but yeah maybe not the ones the got bigger 😅 wonder why though, some are smaller image sizes.

pettinarip:
I've rerun the optimizations on `png` images where we were not really optimizing anything (I've fixed the script).

nhsz:
> Good catch @wackerow! that totally breaks the purpose of this haha. I'll double-check those and revert if needed...uhm 🤔 wondering why that happened...

my theory is that it should depend on the optimization algorithm, e.g.: the original image has a quality preset of '75' and the optimizer has a defined value of '80', the resul will be a larger file, despite the final image being smaller in dimensions

pettinarip:
Interesting @nhsz.

So, the actual problem we had with the heavier images was related only to PNG images. We were not optimizing anything for that type (I would assume that `sharp` defaults to `quality=100` in that case). That is what I've fixed in the script, check [this commit](https://github.com/ethereum/ethereum-org-website/pull/9374/commits/e9e084ac89087c94f78147f896e635846808735a).

Re: the second part of your comment. Indeed, we are running the two optimizations but there is a difference:
- The first optimization is the one we are doing here and the purpose is to alleviate the image resizing process that happens on the build.
- The second optimization is done by Gatsby but the purpose of this step is to create different image sizes for different devices.  We are running this process with [`quality: 100` in most queries](https://github.com/ethereum/ethereum-org-website/blob/master/src/pages-conditional/dapps.tsx#L1416), so we shouldn't lose quality there.

I think that Gatsby is not going to optimize or compress again the image unless we indicate that on the queries.

corwintines (approved):

PR #9376, opened 2023-02-02
Update community-meetups.json

Sergey1997:


corwintines:
Hey @Sergey1997. This seems more like an event than a meetup. However, there haven't been any dates announced for this yet it looks like. Once that's been announced, I encourage you to open another PR to add this to events instead :)

Issue #9387, opened 2023-02-03
Fix typo on 'What is Ethereum' page


lukassim:
### Describe the bug

There is a typo on the 'What is Ethereum' page, specifically:
_The Merge was Ethereum's biggest uprade..._

### To reproduce

1. Go to https://ethereum.org/en/what-is-ethereum/
2. Scroll down to the 'What about Ethereum's energy consumption?' section
3. Read the 2nd sentence

### Expected behavior

'uprade' should be changed to 'upgrade'

### Screenshots

<img width="845" alt="Screenshot 2023-02-03 at 13 07 12" src="https://user-images.githubusercontent.com/37338979/216600452-af4aa21c-4e1a-4b3a-952e-8abb52d50b56.png">

### Desktop (please complete the following information)

All

### Smartphone (please complete the following information)

All

SoumyadeepOSD:
Hey @lukassim , I'm new to this repo. Can you please assign me this issue?

MahendraBishnoi29:
@SoumyadeepOSD I've already fixed this :)

SoumyadeepOSD:
> @SoumyadeepOSD I've already fixed this :)

Okay no problem

f3n4nd0m3:
<!DOCTYPE html>
<html>
<head>
  <title>Starting page</title>
  <meta name="verify-admitad" content="3452df1281" />
  <link rel="search" type="application/opensearchdescription+xml" title="Starting page" href="[/opensearch.xml](view-source:https://internet-start.net/opensearch.xml)">
  <link rel="stylesheet" href="[/landing.css](view-source:https://internet-start.net/landing.css)" type="text/css">
</head>
<body>
<script type="text/javascript">
(function (d, w, c) {
    (w[c] = w[c] || []).push(function() {
        try {
            w.yaCounter1347623 = new Ya.Metrika({"id":33064023,"webvisor":true,"clickmap":true,"trackLinks":true,"params":{"version":"old"},"userParams":{"version":"old"}});
        } catch(e) { }
    });

    var n = d.getElementsByTagName("script")[0],
        s = d.createElement("script"),
        f = function () { n.parentNode.insertBefore(s, n); };
    s.type = "text/javascript";
    s.async = true;
    s.src = (d.location.protocol == "https:" ? "https:" : "http:") + "//mc.yandex.ru/metrika/watch.js";

    if (w.opera == "[object Opera]") {
        d.addEventListener("DOMContentLoaded", f, false);
[...]

Issue #8960, opened 2022-12-20
Bug report: ExpandableCard component text alignment on upgrades FAQ


wackerow:
### Describe the bug

Seeing an odd mixture of aligning text center vs left (in English) within expandable cards, FAQ section of Upgrades page.

### To reproduce

1. Go to https://ethereum.org/en/upgrades/
2. Scroll down to Frequently Asked Questions section near the bottom
3. See alignment of text, and how some are centered and some are left aligned.

### Expected behavior

All of these should consistently aligned as start (left) or center, but not both.

### Screenshots

<img width="1530" alt="image" src="https://user-images.githubusercontent.com/54227730/208769419-314e02ea-f95a-400b-9a2b-fd9fe402a564.png">

### Additional context

I'm guessing this bug might have arisen via the Chakra-UI migration. In the past I believe these were all start-aligned, not centered. @nloureiro Can you confirm if centering text was intended here?

TylerAPfledderer:
@wackerow Chakra's `Accordion` component applies `textAlign: center` in the default theme to `AccordionButton`.

Can at the very least apply the value `unset` which also reverts the value of this property back to `left`. 

Better for future consideration with the new DS is to apply this to the base styles in the Chakra theme config.

TylerAPfledderer:
@wackerow This is coming from Chakra's `AccordionButton`, which serves as the toggle-able header. Because it renders a `button` wrapping all of that content, it is also applying `textAlign: center` from the user agent styles for `button`.

Can at the very least apply `textAlign: left` or `unset` to `AccordionButton`

For future consideration if content like this in all components that will use the `Accordion` components in the new DS, would look to apply this change as a base style in the Chakra theme config.

wackerow:
@TylerAPfledderer Thanks for looking into this... `text-align: start` could be one approach this, alternatively if we actually do want this to be centered, we could likely find a way to just force the width of both texts to fill the accordion, so the centered text is actually centered, rather than dependent on the length of the text.

@nloureiro Thoughts on this? In the past these were left-aligned, but not necessarily against reconsidering this. As mentioned, I'm more looking for consistency in how it appears.. the above screenshot is lookin a bit messy as-is.

nloureiro:
Yes, this is an alignment bug.

All should be to the left.
[...]

PR #9391, opened 2023-02-03
fix text alignment bug in ExpandableCard.tsx

MahendraBishnoi29:
## Description
This change addresses a text alignment bug in the `ExpandableCard.tsx` component. The text was not properly aligned. This change corrects the alignment and ensures that the text is displayed correctly.

## Related Issue
#8960 & #9387


MahendraBishnoi29:
@wackerow I've also fixed a small typo mistake.

wackerow (approved):
Looks good! Thanks @MahendraBishnoi29 

There seems to be some issue with the preview not using Node >18, but I built this locally and it's working fine... nothing here that should break anything. 

Also just a house-keeping note, given these are unrelated fixes that address separate issues, would suggest just keeping them separated as two PRs in the future. These are small so definitely not a big deal, just mentioning =)

MahendraBishnoi29:
> Looks good! Thanks @MahendraBishnoi29
> 
> There seems to be some issue with the preview not using Node >18, but I built this locally and it's working fine... nothing here that should break anything.
> 
> Also just a house-keeping note, given these are unrelated fixes that address separate issues, would suggest just keeping them separated as two PRs in the future. These are small so definitely not a big deal, just mentioning =)

@wackerow thank you for suggestion paul :) since it was only one word typo so fixed in the same PR, would make a separate PR for sure in future 🚀

also should I create a new branch for each of these small typographical errors?

Issue #9275, opened 2023-01-24
Navbar close button visible when opening the Search bar in smaller screen.


amit-ksh:
### Describe the bug

A non-working navbar close button is visible when the Search bar opened on a smaller screen.

### To reproduce

1. Open any page of **ethereum.org** with a screen width of less than 1000px 
2. Click on the Search bar
3. Click the close button on the top-left of the page
4. See the error

### Expected behavior

Navbar close button should not be visible.

### Screenshots

[Home _ ethereum.org.webm](https://user-images.githubusercontent.com/91947037/214236004-c47f409d-4d1c-4c0f-8776-1aa1d25489b4.webm)

### Desktop (please complete the following information)

OS: Windows 10
Browser: Brave

rohitguptaa22:
Hi team ,
Please assign this issue to me I resolve this issue.

rohitguptaa22:
@corwintines merged a PR for this issue, kindly check.

corwintines:
Closing this as it was merged

PR #9399, opened 2023-02-04
Navbar close button visible when opening the Search bar in smaller screen

rohitt-gupta:
## Description

This pull request is a pull request for the mentioned issue.

## Related Issue

when we open the website in the mobile form. 
Open the search container the menu button turns into a cancel button which is not the requirement and hence a bug.
fixes:-  #9275


corwintines:
Hey @rohitt-gupta!

Thanks for opening this. It seems like something happened while merging in dev, and the build no longer works. Could you take a look at what's going on and I can review after? Thanks :)

PR #9404, opened 2023-02-05
refactor(pages): migrate layer 2 page to chakra

TylerAPfledderer:
## Description

Migrate the main Layer 2 page to Chakra UI

## Related Issue
#9353


src/pages/layer-2.tsx
```diff
@@ -40,153 +51,92 @@ import { getImage } from "../utils/image"
...
-    flex-direction: column;
-  }
-`
+const StatDivider = () => {
+  const responsiveOrientation = useBreakpointValue<DividerProps["orientation"]>(
+    { base: "horizontal", md: "vertical" }
+  )
+  return (
+    <Divider
+      orientation={responsiveOrientation}
```
pettinarip, line 104: Probably you've already explained this to me xD but why we can't just do
```suggestion
      orientation={{ base: "horizontal", md: "vertical" }}
```

pettinarip: Also, I noted this bug when you load the page on desktop res:

https://user-images.githubusercontent.com/468158/219204060-5648b762-b51b-4d3f-8a77-4fbf6c3244bf.mp4

- we don't have that horizontal divider on high resolutions
- the vertical divider do not display

src/pages/layer-2.tsx
```diff
@@ -720,39 +681,10 @@ const Layer2Page = ({ data }: PageProps<Queries.Layer2PageQuery>) => {
         <p>
           <Translation id="layer-2-use-layer-2-generalized-1" />
         </p>
-        <CardGrid>
-          {layer2DataCombined
-            .filter((l2) => !l2.purpose.indexOf("universal"))
```
pettinarip, line 725: We are missing this entire section which is dedicated to the "Generalized layer 2s".

From prod version:
image

pettinarip (changes requested):
Looks solid @TylerAPfledderer thanks for the useful comments for each section of the page 😄

src/pages/layer-2.tsx
TylerAPfledderer: @pettinarip

- To the first: What you are suggesting is creating a responsive object. Unfortunately, the `orientation` prop does not accept this, only strictly typed string values of `horizontal` or `vertical` if not `undefined`. `useBreakpointValue` will return a dynamic string based on breakpoint to be passed in and pass the type check.
- To the second: What is missed for the divider to work vertically with `height='100%` is the parent container receiving an explicit height. In this case, it looks like the height can be set to `100px` above the `md` breakpoint and the layout should still work.

src/pages/layer-2.tsx
TylerAPfledderer: Whoops! Adding section comments was supposed to avoid this! 🤪

TylerAPfledderer: I will also update the heading and paragraph tags like you requested for the Home page

src/pages/layer-2.tsx
```diff
@@ -40,153 +53,109 @@ import { getImage } from "../utils/image"
...
-  background: ${(props) => props.theme.colors.ednBackground};
-  border-radius: 2px;
-  border: 1px solid ${(props) => props.theme.colors.lightBorder};
-  padding: 1.5rem;
-  flex: 50%;
-  @media (max-width: ${({ theme }) => theme.breakpoints.m}) {
-    flex: 100%;
-  }
-`
+const StyledInfoIcon = () => (
```
pettinarip, line 95: Noting a difference in size on this icon

image

vs prod version
image

src/pages/layer-2.tsx
pettinarip: The columns are stretching to fill the space.

image

we want to preserve the width of those columns. Check the prod version
image


I guess we want to have the `280px` as the max width.

pettinarip (changes requested):

src/pages/layer-2.tsx
TylerAPfledderer: Replaced `templateColumn` prop with `columns` with a value of `4` to take care that. So no need for an explicit max width! :)

TylerAPfledderer:
@pettinarip should be good with the above changes. I also simplified the stat box group into an iterative and did some further quick tweaks.

pettinarip:
Not sure what happened here @TylerAPfledderer. Seems that you merged `dev`... but now I see > 200 files changes. Could you take a look at that? maybe revert those merges and try it again?

------
On the other hand, I still see an issue with the cards. On smaller resolutions, we might need to only show 2 columns. and 1 column as the base.
image

TylerAPfledderer:
@pettinarip Looks like I solved the massive merge issue!

With the card grids, the minWidth was missing from the grid items, so I added that back in.

src/pages/layer-2.tsx
```diff
@@ -40,153 +53,114 @@ import { getImage } from "../utils/image"
...
+const StatSpan = (props: ChildOnlyType) => (
+  <Flex justifyContent="center" gap={2} {...props} />
+)
+
+const StatDescription = (props: ChildOnlyType) => (
+  <Text opacity={0.8} m={0} {...props} />
+)
+
+const Layer2CardGrid = (props: ChildOnlyType) => (
+  <SimpleGrid columns={4} gap={8} minChildWidth="280px" {...props} />
```
pettinarip, line 161: It seems that by using `minChildWidth` it ignores the `columns` and uses `auto-fit`. <https://chakra-ui.com/docs/components/simple-grid#auto-responsive-grid>

image

I think we should need to use the `Grid` component in this case.

pettinarip (changes requested):

src/pages/layer-2.tsx
TylerAPfledderer: @pettinarip oh shoot! 🤦🏼 

Are you suggesting to go use `Grid` because of the `gridTemplateColumns` prop?

I can go back to using the original value for `gridTemplateColumns` and still use `SimpleGrid` via `templateColumns`

```tsx
const Layer2CardGrid = (props: ChildOnlyType) => (
  <SimpleGrid
    // This value below is in prod
    templateColumns="repeat(auto-fill, minmax(min(100%, 280px), 1fr))"
    gap={8}
    {...props}
  />
)
```

pettinarip: Oh no, go with your approach if that solves the problem. I was just giving an idea. Didn't know that you could use `templateColumns` on the `SimpleGrid`.

pettinarip: However, do you feel that `auto-fill` should be used here? seems that we need the exact number of columns here. That would be `4` => `templateColumns="repeat(4, minmax(min(100%, 280px), 1fr))"`

TylerAPfledderer: Using `4` instead of `auto-fill` is good here. If the container was any wider at the desktop view, it would look odd to me having more than 4 in a row.

pettinarip: We needed the `auto-fill` in this case, it was not wrapping correctly.

image

Sorry for all the back and forth. Now it looks nice.

PR #9406, opened 2023-02-06
migrate assets to chakra UI

MrJithil:
Migrating the assets to Chakra UI
## Description
There are some margin difference of Hero Container. I am not sure, whether we need to customise it each page, or we are following the new Chakra design. Please comment after checking the PR website deployment. 

## Related Issue
#9353


src/pages/assets.tsx
```diff
@@ -93,245 +32,253 @@ const AssetsPage = ({ data }: PageProps<Queries.AssetsPageQuery, Context>) => {
     ? data.ethDiamondPurpleHero
     : data.ethDiamondBlackHero
   return (
-    <Page>
+    <Flex direction={"column"} width="full">
```
pettinarip, line 35: To follow conventions, lets not wrap the value if it is a string literal (this applies also to the rest of the changes).
```suggestion
    <Flex direction="column" width="full">
```

src/pages/assets.tsx
```diff
@@ -93,245 +32,253 @@ const AssetsPage = ({ data }: PageProps<Queries.AssetsPageQuery, Context>) => {
...
       <PageMetadata
         title={translateMessageId("page-assets-meta-title", intl)}
         description={translateMessageId("page-assets-meta-desc", intl)}
       />
-      <Content>
-        <HeroContainer>
-          <Header>
-            <Image
+      <Box padding={"1rem 2rem"}>
```
pettinarip, line 40: To follow conventions, lets use as much as we can, chakra values. You can see them [here](https://chakra-ui.com/docs/styled-system/theme#spacing).
```suggestion
      <Box py={4} px={8}>
```

src/pages/assets.tsx
```diff
@@ -93,245 +32,253 @@ const AssetsPage = ({ data }: PageProps<Queries.AssetsPageQuery, Context>) => {
...
+        <Flex direction={"column"} margin="2rem 0rem">
```
pettinarip, line 41:
```suggestion
        <Flex direction="column" my={8}>
```

src/pages/assets.tsx
```diff
@@ -93,245 +32,253 @@ const AssetsPage = ({ data }: PageProps<Queries.AssetsPageQuery, Context>) => {
...
+          <Center>
+            <Img
               image={getImage(assetPageHeroImage)!}
+              as={GatsbyImage}
```
pettinarip, line 45:
```suggestion
              as={GatsbyImage}
              image={getImage(assetPageHeroImage)!}
```

src/pages/assets.tsx
```diff
@@ -93,245 +32,253 @@ const AssetsPage = ({ data }: PageProps<Queries.AssetsPageQuery, Context>) => {
...
-          />
-          <AssetDownload
-            title={translateMessageId("page-assets-enterprise", intl)}
-            alt={translateMessageId("page-assets-enterprise", intl)}
-            image={data.enterprise}
-            artistName="William Tempest"
-            artistUrl="https://cargocollective.com/willtempest"
-          />
-        </Row>
+          <Flex>
```
pettinarip, line 99: `Row` has more styles that are missing here for mobile responsiveness.

pettinarip (changes requested):
Thanks for the PR @MrJithil 💪🏼 

> There are some margin difference of Hero Container. I am not sure, whether we need to customise it each page, or we are following the new Chakra design. Please comment after checking the PR website deployment.

What we try to achieve at this stage is to just migrate the existing emotion components to Chakra components, keeping the same styles we had. So, in theory, we should see the exact same page we currently have in production (same styles).

Later, when we implement our new Design System, our plan is to consolidate a lot of these repeated components that we have across different layouts/pages.

pettinarip (approved):
Thanks @MrJithil 

I've unified the `Row` components and add some missing styles on the headings.

JithilNCA:
Thank You.

PR #9414, opened 2023-02-07
Translated EVM Opcodes Page to German

SebastianSupreme:
I didn't change the actual opcodes obviously...


corwintines:
Hey @SebastianSupreme , thanks for the PR and the translation!
All of our translations are handled in Crowdin, since it allows for crowdsourcing and contains some useful localization tools. The translations also get reviewed by professional translators before being added to the website, which is done in Crowdin. I believe you already have a Crowdin account setup, so if you could bring these translations over there that would be great!

SebastianSupreme:
Hey @corwintines , sorry for the late reply.
I forgot to mention that Crowdin does not list the opcodes page.
So there is literally no way to translate the opcodes page through Crowdin.
That's why I thought a PR might do the trick.
I definitely get the point that there's no review process through GitHub but I've seen other translation PRs get approved before. That's why I wanted to give it a shot.
So we need to either reconsider this PR or add the opcodes page to Crowdin...

corwintines:
Hmm, ok let me bring this up with Luka and see whats going on. Thanks for bringing that up @SebastianSupreme.

corwintines:
Sounds like we will be adding this soon to the list. Ill add this to GH grooming to discuss as well to confirm what we will do with this PR. Thanks again @SebastianSupreme

PR #9415, opened 2023-02-07
Tutorial tags cleanup

corwintines:
## Description
- Remove project specific tutorial tags

## Related Issue
None


wackerow (approved):
This looks good, but curious given we're removing tags like MetaMask about removing tags such as "Alchemy" or "ethers.js" in favor of more generic tagging?

wackerow:
We can address other tag changes separately, going to pull this in

PR #9416, opened 2023-02-07
Improve the Ethereum Stack Page

SebastianSupreme:
I've added some information and corrected a few grammar mistakes.

### DIFF
--- a/src/content/developers/docs/ethereum-stack/index.md
+++ b/src/content/developers/docs/ethereum-stack/index.md
@@ -10,11 +10,11 @@ There are, however, core components of Ethereum that help provide a mental model
 
 ## Level 1: Ethereum Virtual Machine {#ethereum-virtual-machine}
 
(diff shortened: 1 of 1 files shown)


jmcook1186 (changes requested):
Hi - sorry but I don't think we should add any of this new information. There is no need to talk about `view` and `pure` functions in solidity in this introduction to the stack, it will only make it difficult for non-technical readers to understand. The section is also not Solidity-specific, it applies to any language that compiles to EVM bytecode. Same goes for discussing the different clients - it is orthogonal to the point of the section and only adds friction for a novice reader.

IMO only the typo fixes should be merged. 

Thanks

SebastianSupreme:
That's fine. Then, let's just do that.

jmcook1186 (approved):
cool, thanks

SebastianSupreme:
@jmcook1186 @minimalsm @corwintines 
Can this be merged too in today's master branch merge?

corwintines (approved):

Issue #9067, opened 2023-01-05
Bad redirect from /upgrades/shard-chains/


seishun:
### Describe the bug

Requesting this path redirects to //upgrades/sharding/.

### To reproduce

1. Go to https://ethereum.org/upgrades/shard-chains/

### Expected behavior

Landing on a page related to sharding or an error page.

corwintines:
Thanks for reporting this @seishun! Definitely not the intended experience. Ill open a PR to add this redirect in.

PR #9442, opened 2023-02-08
[Fix] Bad redirect from /upgrades/shard-chains/ 

corwintines:
## Description
- Adds default redirect for `/upgrades/shard-chains` so that it redirects to `/en/upgrades/sharding`

## Related Issue
Fixes https://github.com/ethereum/ethereum-org-website/issues/9067

### DIFF
--- a/redirects.json
+++ b/redirects.json
@@ -205,7 +205,7 @@
   },
   {
     "fromPath": "/upgrades/shard-chains/",
-    "toPath": "/en/upgrades/sharding/"
+    "toPath": "/en/roadmap/danksharding/"
   },
   {
     "fromPath": "/*/eth2/staking/",
--- a/redirects.json
+++ b/redirects.json
@@ -204,6 +204,10 @@
     "toPath": "/:splat/upgrades/sharding/"
   },
   {
+    "fromPath": "/upgrades/shard-chains/",
+    "toPath": "/en/upgrades/sharding/"
+  },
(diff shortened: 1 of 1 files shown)


wackerow (changes requested):
We have since merged in the roadmap updates, which removed the `/upgrades/sharding/` page, replacing it with `/roadmap/danksharding/`

Challenge is that this content only exists in English right now, and several languages still have the `/upgrades/sharding/` page available.

Is there a solution to redirect users to the `/roadmap/danksharding/` page only if it's available in that language? 

- We could prioritize migrating the translated forms of the old `/upgrades/sharding/` page to the `/roadmap/danksharding/` folder, using the new template.
- We could only force `/en/upgrades/sharding/` to redirect (specifically the English path only), leaving the translated ones alone
- Other ideas?

redirects.json
```diff
@@ -203,6 +203,10 @@
     "fromPath": "/*/upgrades/shard-chains/",
     "toPath": "/:splat/upgrades/sharding/"
   },
+  {
+    "fromPath": "/upgrades/shard-chains/",
+    "toPath": "/en/upgrades/sharding/"
```
corwintines, line 208:
```suggestion
    "toPath": "/en/roadmap/danksharding/"
```

Update with roadmap release

wackerow (approved):

PR #9453, opened 2023-02-09
Typo in withdrawals (principle -> principal)

sandakersmann:
Typo in withdrawals (principle -> principal)


wackerow (approved):
Thanks for the patch @sandakersmann! 🙏

PR #9457, opened 2023-02-09
Fix ETH.STORE URL

Buttaa:
We just added ETHSTORE to beaconcha.in. This PR forwards the user to the correct subpage :)

### DIFF
--- a/src/components/Staking/StakingStatsBox.tsx
+++ b/src/components/Staking/StakingStatsBox.tsx
@@ -69,10 +69,11 @@ const BeaconchainTooltip = ({ isEthStore }: { isEthStore?: boolean }) => (
     content={
       <div>
         <Translation id="data-provided-by" />{" "}
-        {isEthStore && (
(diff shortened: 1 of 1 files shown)


Fqc301033 (approved):

Fqc301033 (approved):

wackerow (approved):

wackerow (changes requested):
@Buttaa As I look at this again, I'm not sure we need to link out to two different places for eth.store... We were splitting this before when the beaconcha.in/ethstore page wasn't ready yet.

I think we should just update this to just show ETH.STORE (or Beaconcha.in ETH.STORE) and link to the new URL, removing the github link. 

We can keep the `isEthStore` check here, and if true use the ETH.STORE label/link, and if not use the regalar Beaconcha.in label/link. How does that sound?

wackerow (approved):
Per our chat @Buttaa, updated to simplify this so there is only one link to the new beaconcha.in/ethstore page. Good to go from my end!

Issue #8877, opened 2022-12-12
Include External Link Icon in a Child Element Instead of `:after` Pseudo


TylerAPfledderer:
### Describe the bug

For some screen readers, they have the potential to read out any content that can be read within the pseudo elements. The external link character in the `Link` component is one of them (Checked with NVDA).

Allowing this character to be conditionally rendered in a child element instead of a pseudo will allow the ability to apply `aria-hidden` on said element to ensure it is visible, but can not be read by a screen reader.

```tsx
{ !hideArrow && (
  <Box as="span" ms={0.5} aria-hidden>↗</Box>
)}
```

### To reproduce

1. Go to any page that renders `Link` or `ButtonLink` with external. (i.e. https://ethereum.org/en/layer-2/)
2. Use screen reader (`NVDA`, in particular) and navigate to a given link.
3. Reader will read out the text and end with "upright arrow"

### Expected behavior

Screen reader should not read out this repetitive icon

TylerAPfledderer:
Eh, this actually shouldn't be a bug, but an A11y improvement! 😅 ♿

corwintines:
Updated the tags to match your comment :)

Let me know if you're interested in working on this and ill assign this

TylerAPfledderer:
@corwintines assign away! 🤘

PR #9469, opened 2023-02-10
a11y(link): update indicators for external links

TylerAPfledderer:
## Description
Moves the external link icon in the `Link` component from the after pseudo to a child element. This allows for the icon to be hidden from screen readers.

Also adds hidden indicator text that the link will open a new tab. This dissolves possible confusion for a blind user on where they are at in the browser. (Are they still on the website to be able to go back a page, or someplace else? It might not always be clear depending on the link text.)

## Related Issue
#8877

### DIFF
--- a/src/components/Link.tsx
+++ b/src/components/Link.tsx
@@ -5,6 +5,7 @@ import {
   Link as ChakraLink,
   LinkProps,
   useTheme,
+  VisuallyHidden,
 } from "@chakra-ui/react"
 import { navigate as gatsbyNavigate } from "gatsby"
 import { LocalizedLink as IntlLink } from "gatsby-theme-i18n"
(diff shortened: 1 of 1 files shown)


src/components/Link.tsx
```diff
@@ -113,6 +115,12 @@ const Link: React.FC<IProps> = ({
         {...commonProps}
       >
         {children}
+        <VisuallyHidden>(opens in a new tab)</VisuallyHidden>
+        {!hideArrow && (
+          <Box as="span" ml={0.5} mr={1.5} aria-hidden>
```
corwintines, line 120: The addition of the `VisuallyHidden` content is great!

I'm still having the initial problem that this set out to solve, the arrow is being read out by the screen reader.

This is the DOM structure being created, so I think it has to do with the children still being accessible by the screen reader.
Screen Shot 2023-03-22 at 12 44 33 PM

Will look a bit into how to resolve this.

corwintines (changes requested):

src/components/Link.tsx
TylerAPfledderer: @corwintines hmmm what screen reader did you use?

corwintines: @TylerAPfledderer this is what I have at the moment. https://getpericles.com/

If this is a screen reader issue, I can look for a new one.

PR #9489, opened 2023-02-14
Adding QuickNode Faucet links

Sahilsen:
## Description

Added QuickNode faucet under Sepolia and Goerli faucet's list


corwintines (changes requested):
Thanks @Sahilsen 

If you can fix the ordering so this is at the bottom, ordered by time added, then this is good to go. I tried to push this up, but dont have write access here.

corwintines (approved):

PR #9490, opened 2023-02-15
Update index.md

miyamok:
Technical wordings corrected.

## Description

The explanation, "Logical equivalence" is also called "logical implication," is not true. I suggest "logical bi-implication" instead of "logical implication" which is true.

The undefined wordings "statement A" and "statement B" should be avoided.
The expression "can only be" concerning the truth of the first statement (statement A) makes the matter complicated, hence I suggest simply saying "is."


corwintines (approved):
Thanks @miyamok

PR #9491, opened 2023-02-15
Clear wording for logical conjunction

miyamok:
correction

## Description

Saying "but" for logical conjunction is confusing and should be avoided.
I suggest replacing "but" by "and also."


jmcook1186 (approved):
ok, thanks

PR #9493, opened 2023-02-15
Simplify PoS explanation

samajammin:
## Description

I can't help but feel like this paragraph gets too deep into the weeds for an "intro to Ethereum" page. Explaining that "32 ETH" is required for the "deposit contract" seems like too much advanced detail. Same with content covering fork-choice rules - that doesn't seem needed on an intro page.

## Related Issue

None :)


samajammin:
@jmcook1186 curious to hear your take on this. I won't be offended if you disagree 😄

jmcook1186 (approved):
agree, the simplified explanation is much better!

wackerow (approved):
👍 Yeah much cleaner, agree that got too into the weeds too quick

PR #9502, opened 2023-02-16
Update index.md

luaroncrew:


vj-codes (commented):
I think it is correct grammatically the way it is already.

luaroncrew:
You apply the construction (there is)/(there are) on a word "set" in singular. So gramatically it will be correct to use 3rd person singular for the verb. You cannot say "there are a set". It's correct to use "there is a set" 

https://www.grammar.cl/Present/ThereIsThereAre.htm

corwintines (approved):

PR #9509, opened 2023-02-17
DocSearch MVP implementation

### LINKED ISSUE #9429: [Design system V2] New doc search component.
This issue is a part of the https://github.com/ethereum/ethereum-org-website/issues/9431

The goal is to adopt doc search from Algolia to be the search component for ethereum.org

The component is a pre-built one we only get to change some styles to blend better into the Design system.

We have a Design proposal ready to be implemented by @wackerow, but open to getting feedback.

Here is the [Figma file](https://www.figma.com/file/wkPuZW7kbXuEiawA4A2isT/V2%2Fdoc-search-component?node-id=0%3A1&t=hK4RRlKeld9Lef2e-1) with the design.

The header with the new search input
Screen Shot 2023-02-08 08 30 00 PM

The search modal
Screen Shot 2023-02-08 08 29 47 PM

wackerow:
## Description
- Implements DocSearch component from `@docsearch/react` library
- Uses existing Algolia index
- Uses `Styles.tsx` to declare a `Global` styling element using css to override DocSearch class styling
- Mobile: Updates logic for mobile menu to handle new DocSearch _modal_, instead of the search panel being embedded in the menu drawer

TODO:
- [x] Finish setting up i18n support
- [x] Fix mobile menus
- [x] Design review and sign-off
- [x] Dev review

## Preview link
<a target="_blank" rel="noopener" href="https://ethereumorgwebsitedev01-docsearch.gatsbyjs.io/en/">https://ethereumorgwebsitedev01-docsearch.gatsbyjs.io/en/</a>

## Related issue
- Fixes #9429


src/components/Search/Styles.tsx
```diff
@@ -0,0 +1,242 @@
+import React from "react"
+import { Global, css } from "@emotion/react"
+
+const styles = css`
```
pettinarip, line 4: As this is a component
```suggestion
const Styles = css`
```

pettinarip: Oh its not...my bad. Ignore this.

wackerow: :-) Gonna refactor using approach here anyway:
https://github.com/chakra-ui/chakra-ui-docs/blob/main/src/components/search.styles.tsx

That will remove this variable, and allow us to replace all the css `var()` instances with JS.

wackerow: Per our discussion, pushed an update to use returned `theme` variable, but did not use this for colors as it created issues with color mode responsive semantic tokens.

wackerow:
Would be nice to also include the existing ETH address recognition that our current search has.

src/components/Nav/index.tsx
```diff
@@ -362,13 +362,31 @@ const Nav: React.FC<IProps> = ({ path }) => {
...
+      ) {
+        setIsSearchOpen(false)
+        setIsMenuOpen(false)
+      }
+    }
+    document.addEventListener("click", closeSearchListener)
+    // remove event listener on dismount
+    return () => {
+      document.removeEventListener("click", closeSearchListener)
+    }
```
wackerow, line 381: Need to debug this still... having trouble getting this to work appropriately on mobile.

Clicking the top search icon on mobile opens the modal, but then clicking/tapping outside of the modal (along the 1rem padding) will close the search but leaves the backdrop from the menu open.

image

Tapping/clicking this then removes it, but if you click the menu icon a few times it will toggle the menu without removing this backdrop.

TylerAPfledderer:
Hey @wackerow and @pettinarip!

Pablo had requested I check out a better way to apply styles using Chakra tokens instead of CSS variables through emotion.

**tldr**: [CodeSandbox Mock Solution](https://codesandbox.io/s/chakra-w-docsearch-mock-2flf11?file=/src/EthDocSearch/index.tsx)

This ended up being a little bit trickier than I was expecting, but my solution results in having to import `DocSearchButton` and `DocSearchModal` instead of `DocSearch`.

Why? In the `DocSearch` component, the modal is rendered in a portal. If I infuse Chakra props to `DocSearch`, the styling would be scoped to a container I would create, and the modal will be outside this scope. This would mean placing the modal styles as high as the global object in the theme config (not desirable).

## 

So I am using the Chakra Factory to allow chakra props to be available on the components. In addition, because the factory is using emotion to generate a class name hash to attach the props to, I'm ensuring the class name is at least forwarded to the component.

Then, using the `sx` prop to declare the needed className selectors and apply Chakra tokens to the styles where possible.

In the case of a modal, I have to add a wrapper element to pass `className` to because the modal does not directly accept it.

With the button, instead of the factory, because it is using CSS flex, I am consuming it with the Chakra `Flex` component, which I think is more appropriate here in the additional level of abstraction.

I copied all the disclosure logic from `DocSearch` and paste it here. This can now allow for further modification of the logic if you want.

This is open to moving around logic between the components for something that would be more desirable. And the styling might have some room for simplification.

Also, can still consider in the Modal to use it's `hitComponent` and `footerComponent` props to customize the search result styles and footer styles directly with Chakra components instead of through CSS selectors at the higher level.

## Alternative approaches
After some thought while making this Sandbox, I can not think of a different or better solution here. I will gladly accept any thoughts or observations to explore any changes here. 😄

wackerow:
@TylerAPfledderer Awesome! Thanks a lot for digging into this. :tada:

My thought here would be to implement this as a "v2". I just retitled this PR as an MVP. I think since we have a working PR, and this refactor has some complexities to it, if @pettinarip  is cool with pulling this in (after reviews) then I think your suggestions would make great additions for v2.

Might be worth hoping on a call to chat over some details here, and try to tie in the design updates @nloureiro has been working on.

TylerAPfledderer:
@wackerow good with me! I figured it would be a good follow-up once you were satisfied with the intended result. 😁

nloureiro:
I did a last review from my side, and all is good for the V1 🚢

pettinarip:
That is great @TylerAPfledderer! thanks for looking into that and for the detailed explanation.

Agree with you @wackerow on the planning. Lets first implement this version and then we can incorporate the improvements Tyler suggested as a v2. I've created a new issue to track that here #9573.

====
Note: @TylerAPfledderer I've assigned the issue to myself because for some reason, I can't assign it to you until you make a comment on the issue 🤷🏼 So, please leave a comment there if you want to be assigned to it 🙏🏼

pettinarip (commented):
Would rerun the install of the new dep. We are installing just 1 new dep but I see +- 7k changes on `yarn.lock`.

I'd suggest doing:
1. `git checkout dev -- yarn.lock`
2. install everything again `yarn`

We should see only a few modifications on the `yarn.lock`

pettinarip (approved):
lgtm! 🚀 🚀 🚀 nicely done @wackerow

PR #9510, opened 2023-02-17
Update theme foundations for new DS

TylerAPfledderer:
## Description

_This PR is part of Implementing the New Design System_

This PR adds or modifies foundational styles in the custom Chakra theme config.

- Adds or modifies colors, shadows, and typography outside of Chakra's default theming according to the new design system specs.
- Adds the `Inter` variable font as a static file, with font-face implementation via the emotion `Global` component. (Theme mods for this font commented out to not affect current prod)
- Adds comments, primarily noting tokens that are to be deprecated once components and layouts are converted.

Related issue(s): #9548


src/@chakra-ui/gatsby-plugin/foundations/colors.ts
```diff
@@ -10,30 +10,35 @@ const colors = {
...
     500: "#ff7324",
-    800: "#451900",
+    600: "#c95d20",
+    800: "#6a3301",
   },
   red: {
-    100: "#F7C8C8",
+    100: "#f7c8c8",
     500: "#b80000",
     900: "#1B0C0C",
```
pettinarip, line 29: I think we are deprecating this color as well, right? I don't see that on the DS anymore. cc @nloureiro

src/@chakra-ui/gatsby-plugin/foundations/typography.ts
```diff
@@ -0,0 +1,19 @@
+const typography = {
+  // TODO: Uncomment when the `Inter` font is included in the project
+  // fonts: {
+  //   heading: "Inter, sans-serif",
+  //   body: "Inter, sans-serif",
+  // },
```
pettinarip, line 6: Cool. Curious, is there any difference between defining it here than on the top-level (as it were before)?

Anyway, I like more placing this here than in the old place :)

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -31,21 +35,41 @@ const semanticTokens = {
 
     // Design System colors
     primary: { _light: "blue.500", _dark: "orange.500" },
+    primaryHighContrast: { _light: "blue.700", _dark: "orange.100" },
+    primaryLowContrast: { _light: "blue.100", _dark: "orange.800" },
     primaryLight: { _light: "blue.100", _dark: "orange.100" },
     primaryDark: { _light: "blue.700", _dark: "orange.800" },
```
pettinarip, line 41: wondering if we should deprecated these tokens as well @nloureiro

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
```diff
@@ -10,30 +10,35 @@ const colors = {
...
-    500: "#109e62",
+    500: "#0a7146",
+    // ! Deprecating 900
     900: "#0A160E",
   },
   yellow: {
     200: "#fff8df",
+    500: "#bd8400",
   },
 }
```
pettinarip, line 43: I think we should also add the brand colors: https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?node-id=71%3A516

Not sure if we should follow the `{colorName}.{scaleNumber}` convention here. Maybe we could add it like this:
```tsx
const colors = {
  brand: {
    jordyBlue: "...",
    pantoneBlue: "...",
    ....
  }
}
```
what do you think?

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -31,21 +35,41 @@ const semanticTokens = {
...
+    bodyLight: { _light: "gray.500", _dark: "gray.300" },
+    disabled: { _light: "gray.300", _dark: "gray.500" },
     background: { _light: "white", _dark: "gray.700" },
+    backgroundHighlight: { _light: "gray.100", _dark: "gray.900" },
+    // ! Deprecating neutral
     neutral: { _light: "white", _dark: "gray.900" },
-    success: { _light: "green.500", _dark: "green.400" },
+    success: "green.500",
+    successLight: "green.100",
+    successOutline: { _light: "success", _dark: "successLight" },
```
pettinarip, line 57: where is this coming from? the `Alert`s?

src/components/Fonts.tsx
```diff
@@ -0,0 +1,19 @@
...
+const Fonts = () => (
+  <Global
+    styles={`
+      /* latin */
+      @font-face {
+        font-family: 'Inter';
+        font-style: normal;
+        font-weight: 300 800;
+        font-display: swap;
+        src: url('/fonts/Inter.ttf') format('truetype');
```
pettinarip, line 13: Wondering is we should also add `woff`/`woff2` formats for modern browsers. cc @nloureiro

pettinarip (commented):
This is great @TylerAPfledderer! thanks 💪🏼 Would love also the eyes from @nloureiro on PR.

Where did you get the font file? is that coming from any official site?

TylerAPfledderer:
@pettinarip the font file came straight from the font authors repo. (Looks like a smaller file size for the variable version than from Google fonts)

https://github.com/rsms/inter

src/@chakra-ui/gatsby-plugin/foundations/typography.ts
TylerAPfledderer: No difference, just for code splitting! 😄

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
TylerAPfledderer: This color is being recognized as `Error Light` in the DS

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
TylerAPfledderer: @pettinarip I did not have a chance yet to go back in and mark these since Nuno last updated the palette.

src/components/Fonts.tsx
TylerAPfledderer: @TylerAPfledderer yes! There is a file available in the author's repo I linked that I didn't see previously.

gatsby-ssr.tsx
```diff
@@ -34,3 +34,16 @@ export const wrapPageElement: GatsbySSR<any, Context>["wrapPageElement"] = ({
...
+    <link
+      rel="preload"
+      href="/fonts/Inter-roman.var.woff2"
+      as="font"
+      type="font/woff2"
+      crossOrigin="anonymous"
+      key="interFont"
+    />,
+  ])
+}
```
TylerAPfledderer, line 49: @pettinarip added this preload per [Gatsby docs](https://www.gatsbyjs.com/docs/how-to/styling/using-local-fonts/#preload-your-fonts)

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
pettinarip: I'm talking about the `red.900`. It is not on the DS anymore.
image

src/@chakra-ui/gatsby-plugin/foundations/typography.ts
pettinarip: Cool haha didn't notice that you were spreading the props on the other side 👍🏼 this is better

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
pettinarip: Yea, not a blocker. We can always mark them later.

pettinarip (changes requested):

gatsby-ssr.tsx
pettinarip: Nice! not sure how this works if we define multiple fonts.

TylerAPfledderer: @pettinarip Since this is making a request ahead of time, there would have to be consideration on what font would be the most noticeable by the user, as doing this multiple times could impact performance.

src/components/Fonts.tsx
```diff
@@ -0,0 +1,27 @@
...
+        font-family: 'Inter';
+        font-style: normal;
+        font-weight: 300 800;
+        font-display: swap;
+        font-named-instance: 'Regular';
+        src: url("/fonts/Inter-roman.var.woff2") format("woff2");
+      }
+      `}
+  />
+)
```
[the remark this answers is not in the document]
TylerAPfledderer: Sounds good. I can update this!

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
TylerAPfledderer: It is only being used in the `errorNeutral` semantic which I have already marked for deprecation, so this should also be marked.

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
Inatiye0911905991: From Ethiopia I'm sintayew Gashaw

pettinarip (approved):
lgtm 💪🏼 gj @TylerAPfledderer 

Would wait for @nloureiro 👀 on this as well.

TylerAPfledderer:
@pettinarip @nloureiro would recommend upgrading Chakra to the latest for this PR.

[@chakra-ui/react v2.5.4](https://chakra-ui.com/changelog/2.5.4) introduces the ability tokenize gradient styles so we can now use semanticTokens for them! 😁

nloureiro (commented):
It all sounds on my side to add this.
Will need to have another PR to test everything on the browser.
excellent work @TylerAPfledderer

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
pettinarip: @nloureiro not sure. Are we planning on using those colors on other parts of the site?

- If **no** then I wouldn't add them on the theme yet.
- If **yes** then I would add them as I mention previously. Following the same name you used on the DS.

Would be something like

```tsx
const colors = {
  gray: {
    100: "#f7f7f7",
    ....
  },
  blue: {
    ...
  },
  orange: {
    ....
  },
  // the 5 brand colors
  brand: {
    jordyBlue: "...",
    celesteGreen: "...",
    babyPink: "..."
  }
}
```

The naming for those colors would be the names you used on the DS
image

src/components/Fonts.tsx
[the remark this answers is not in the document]
pettinarip: If we are planning on using them in the future then lets add it, yes.

The browser will only load those fonts that are used on the page. If no element uses that font, it will not load it. We should be fine on that aspect.

pettinarip (changes requested):
To summarize what we need to add before merging:

- [ ] add bold italic font
- [ ] bump Chakra to latest version
- [ ] add gradients as semantic tokens
- [ ] add brand colors? (to be confirmed by @nloureiro)

TylerAPfledderer:
> To summarize what we need to add before merging:
> 
>     * [ ]  add bold italic font
> 
>     * [ ]  bump Chakra to latest version
> 
>     * [ ]  add gradients as semantic tokens
> 
>     * [ ]  add brand colors? (to be confirmed by @nloureiro)

@pettinarip @nloureiro I made the updates except for the last.

I forgot: should I add the stories covering the color and spacing tokens? Or save them for a separate PR? (I would think the latter since this PR got lengthy!)

Could also include the gradients for a story under the same group (`Atoms / Design Tokens / Gradients`), and also still need to figure out where to put them in the directory structure since they are not connected to a component.

pettinarip:
@TylerAPfledderer awesome!

> I made the updates except for the last.

I think we can decide and implement this later. Its ok.

> I forgot: should I add the stories covering the color and spacing tokens? Or save them for a separate PR? (I would think the latter since this PR got lengthy!)

Yea, lets do that on a different PR.

> Could also include the gradients for a story under the same group (Atoms / Design Tokens / Gradients), and also still need to figure out where to put them in the directory structure since they are not connected to a component.

Correct. I think that wasn't added to the figma yet https://www.figma.com/file/Ne3iAassyfAcJ0AlgqioAP/DS-to-storybook-structure?node-id=42-92&t=OJG0sQ0QHfrWglFb-0

pettinarip (approved):
Amazing job @TylerAPfledderer 👏🏼 👏🏼

PR #9513, opened 2023-02-17
Add gulag distinction

Bread-VeSea:


Bread-VeSea:
Re-opening because I do think a distinction needs to be made and this was not acknoweldged

Bread-VeSea:
Addressed elsewhere

PR #9519, opened 2023-02-17
Deprecate NakedButton component

wackerow:
## Description
- Deprecate `NakedButton` component, and replace with use of Chakra components
- Uses `IconButton` for icons used as buttons without any text
- Adds "Clear" string to `common.json` for aria label
- Removes use of a button for hash link on `run-a-node` page

## Related Issue
- #6374


pettinarip (approved):
Awesome! nicely done @wackerow

PR #9527, opened 2023-02-20
New Job Board

CryptoDev89668487:
Requesting to add a new job board.


corwintines:
This looks like it's less than 6 months old, and largely shows the same jobs as any other job board here. Going to close this out for those reasons, but mainly the 6-month guideline which we have for products.

PR #9551, opened 2023-02-22
Add ETHVenice Event

jhonnyvianello:
## Add ETHVenice Event
Added ETHVenice 2023 Event 🇮🇹


corwintines (approved):
Thanks @jhonnyvianello

PR #9572, opened 2023-02-24
Migrate the `upgrades/get-involved` page to Chakra

amit-ksh:
## Description

Migrate `upgrades/get-involved` page to Chakra

## Related Issue

#9353

### DIFF
--- a/src/pages/upgrades/get-involved/index.tsx
+++ b/src/pages/upgrades/get-involved/index.tsx
@@ -196,11 +196,20 @@ const ReverseRow = (props: ChildOnlyProp) => (
   />
 )
 
-const Column = (props: ChildOnlyProp) => (
+const LeftColumn = (props: ChildOnlyProp) => (
   <Box
     w={{ base: "full", md: "50%" }}
-    ml={16}
+    mr={16}
+    ml={{ base: 16, md: 0 }}
+    {...props}
+  />
+)
+
+const RightColumn = (props: ChildOnlyProp) => (
+  <Box
+    w={{ base: "full", md: "50%" }}
     mr={{ base: 16, md: 0 }}
+    ml={16}
     {...props}
   />
 )
@@ -542,7 +551,7 @@ const GetInvolvedPage = ({
       <StyledGrayContainer>
         <PageContent>
           <Row>
-            <Column>
+            <LeftColumn>
               <H2 id="#bug-bounty">
                 <Translation id="page-upgrades-get-involved-bug-hunting" />
               </H2>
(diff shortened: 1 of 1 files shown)


src/pages/upgrades/get-involved/index.tsx
```diff
@@ -15,153 +33,207 @@ import PageMetadata from "../../../components/PageMetadata"
...
+  <Box
+    as={CalloutBanner}
+    bg="transparent"
+    w={{ base: "full", md: "auto" }}
+    p={{ base: 0, md: 12 }}
+    pt={{ base: 16, md: 12 }}
+    sx={{
+      "&": {
+        ml: 0,
+      },
```
amit-ksh, line 232: At the moment, `CalloutBanner` is written in emotion and the *margins* of the component can't be overridable using `ml=0` because `@media (max-width)` query gets higher precedence over the Chakra class. But Chakra `sx` attribute allows us to override the `@media (max-width)` query, that's why I used it here.

src/pages/upgrades/get-involved/index.tsx
```diff
@@ -15,153 +33,207 @@ import PageMetadata from "../../../components/PageMetadata"
...
+    direction={{ base: "column-reverse", md: "row" }}
+    {...props}
+  />
+)
+
+const Column = (props: ChildOnlyProp) => (
+  <Box
+    w={{ base: "full", md: "50%" }}
+    ml={16}
+    mr={{ base: 16, md: 0 }}
```
TylerAPfledderer, line 203: These should be reversed.
```suggestion
    ml={{ base: 16, md: 0 }}
    mr={16}
```

TylerAPfledderer (changes requested):
Hey @amit-ksh! Nice work here. I've left some thoughts. :)

src/pages/upgrades/get-involved/index.tsx
```diff
@@ -396,10 +467,10 @@ const GetInvolvedPage = ({
...
       />
-      <Content>
+      <PageContent>
         <HeroCard>
           <HeroContainer>
-            <StyledBreadcrumbs slug={location.pathname} startDepth={1} />
+            <Breadcrumbs slug={location.pathname} startDepth={1} />
             <SloganGradient>
               <Translation id="page-upgrades-get-involved" />{" "}
               <Emoji size={1} text=":wave:" />
```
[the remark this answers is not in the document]
amit-ksh: Done 👍

src/pages/upgrades/get-involved/index.tsx
amit-ksh: Thanks for pointing this out, I created two separate components for `LeftColumn` and `RightColumn` as previously. Now it should work as expected.

pettinarip (approved):
Nicely done @amit-ksh! 💪🏼 

🚀

Issue #9546, opened 2023-02-22
Epic: Implementation of the Design System


pettinarip:
_This is an epic that makes up part of the [ethereum.org Q1 roadmap](https://github.com/ethereum/ethereum-org-website/issues/9090)_

## Description

Now that the first version of our Design System is complete and we have already migrated most of the UI components to Chakra, we can proceed with implementing the DS on [ethereum.org](https://ethereum.org/). Our DS is a set of guidelines and reusable components that will help ensure a consistent and user-friendly experience across the website.

To implement the DS, we will follow the same structure we used in its development. We’ll break down the implementation process into three stages:

- [x] #9548 - In the first stage, we will focus on integrating core components and the foundation styles.
- [ ]  #11117 - The second stage will involve adding more complex or composed components, and adapting existing ones as needed.
- [ ]  **Implement DS v3** - Finally, in the third stage, we will focus on special pages that have custom designs. For instance, the Layer 2 and Find a Wallet pages.

For more information about our DS and the migration to Chakra, check out the links below:

- Design System: https://github.com/ethereum/ethereum-org-website/issues/6284
- Migration of UI components: https://github.com/ethereum/ethereum-org-website/issues/6374

### New folder structure for new components

Since we've incorporated [Storybook and Chromatic](https://github.com/ethereum/ethereum-org-website/issues/8490) into our workflow, from now on we need to create stories for every component we create or modify based on the DS.

```sh
// at the end of the implementation, all components should have
// the following structure

components/
  NewComponent/
    index.tsx
    NewComponent.stories.tsx
```

Going forward, any changes made to the UI will require approval or denial on [Chromatic](https://www.chromatic.com/builds?appId=63b7ea99632763723c7f4d6b), which will become an additional step in our review process.

konopkja:
@pettinarip  should we close this one now?

wackerow:
@konopkja My take here is we should probably sync up to reevaluate our plans for the design system since much time has passed since the original initiative. 

Related open issues:
- #9545 (This one)
- #9548 
- #11117
- #10744
- #10886
- Maybe others I missed?

Given the many migrations since then, e.g., Next.js 14 (and soon™️ v15), shadcn/ui with TailwindCC v3 (soon™️ v4), and recent discussion about updating the DS again, we may want to consider calling this complete to the extent it's been implemented, and abandoning the rest while we reevaluate.

pettinarip:
[...]

PR #9575, opened 2023-02-24
docs(applying-storybook): update storybook docs for v7

TylerAPfledderer:
_This is an update for #9546 to provide better direction in implementation for Storybook for components._

This PR updates the [applying-storybook.md](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/applying-storybook.md) in the docs directory of the project to be inline with Storybook v7

- Updates codeblocks and notation for the new Story Object syntax
- Adds a quick ref table for the different selections in the UI Dashboard

### DIFF
--- a/docs/applying-storybook.md
+++ b/docs/applying-storybook.md
@@ -18,20 +18,20 @@ It's as easy as running `yarn storybook` to boot up a dedicated localhost to see
 
 ## Setting up a component's stories
 
+> 🚨 NOTE: This project has version 7, which is currently still in beta. The following documentation outlines preferences in setup as it relates to this version. You can refer to the [7.0 beta docs](https://storybook.js.org/docs/7.0/react/get-started/introduction) if you need any additional details
+
 A Storybook "story" is an instance of a component in a certain state or with certain parameters applied to show an alternative version of the component.
 
-Each component will only need one file containing all the stories, and should follow the naming convention of the component.
+Each component will only need one file containing all the stories.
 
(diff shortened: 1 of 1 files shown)


TylerAPfledderer:
@pettinarip I would love to know if there is anything I should add or remove. :)

TylerAPfledderer:
@pettinarip I have made those requested changes.

One final thing I suppose that is missing. Should I add on here notation regarding what the directory structure should be in Storybook per Nuno's suggestions, and include how to write them in each stories meta `title` option?

It's currently only set up to do an auto generate based on the current structure which will not have the intended result.

Disclaimer: I think there is a way we can auto-generate the correct structure too, but might require known variables to exist. And I haven't explored this kind of feature before.

pettinarip:
@TylerAPfledderer that is a good point.

I don't think we need to define the concrete structure here but perhaps it is worth to mention that our naming convention is coming from the DS. That we follow the naming declared in the DS.

That way, we avoid the maintenance of this doc.

TylerAPfledderer:
I'll ping you when I've added a blurb for this.

TylerAPfledderer:
@pettinarip I've added the note for the display name preference.

I've also added a note regarding [single-story hoisting of the display name](https://github.com/ethereum/ethereum-org-website/pull/9575/files#diff-be6f8c53bd1da59e25e89779449738be4716663b7c000e38745c7534e435e2d3R109).

pettinarip (approved):
Awesome! great job, thanks @TylerAPfledderer

PR #9587, opened 2023-02-27
Add another DeSci use case

hung-genetica:
## Description

- Add one more DeSci use case: Precision medicine for underrepresented populations 
- And add a link to GenomicDAO project


corwintines:
This content doesnt really add much to the page, and this project is not an Ethereum project its a BSC project. Going to close this out.

Issue #7828, opened 2022-09-13
Update Beacon Chain terminology after The Merge?


jmcook1186:
This issue aims to determine the right way to refer to the Beacon Chain after The Merge. It will be helpful to define a correct naming convention to be used across ethereum.org after the merge.

The Beacon Chain (TBC) went live in November 2020 and ran in parallel to Ethereum Mainnet (EM) right up to the moment of The Merge in September 2022. Having independent terms to describe the two independent chains made a lot of sense because TBC and EM were developed separately and fulfilled different functions - they were different blockchains. However, since The Merge, there is only one Ethereum blockchain formed by combining pre-merge EM and pre-merge TBC. 

Does it makes sense to persist the "Beacon Chain" terminology or to deprecate it in favour of "Ethereum Mainnet". 

**Arguments for persisting TBC terminology are:**
- Since it is a chain of `beacon_blocks` it makes sense to refer to the Beacon Chain.
- there is Beacon terminology baked into clients and tooling, such as `beacon_state`, `beacon_block`, `beacon_node` etc which doesn't make much sense if we deprecate the term Beacon Chain
- It makes a useful distinction between pre- and post-merge Ethereum and EL/CL.

**Arguments against TBC terminology are:**
- The Beacon Chain was a development artefact that did not handle real transactions. It is better for the term to keep this specific definition and leave it behind in the pre-merge world. Let EM refer to the "live" chain for continuity with pre-merge world.
- it will be confusing to have multiple terms for the same chain (Ethereum blockchain, Ethereum Mainnet, Beacon Chain). We have already seen that the ETH2 terminology confused people and was surprisingly sticky. To non-experts multiple terms suggests multiple chains and this misconception will be reinforced by the abundant pre-merge documentation that will persist online.

It would be great to hear people's opinions on this issue

samajammin:
My 2 gwei: given that there's now only one "Ethereum blockchain", I think it makes sense to keep terminology consistent with that, particularly to avoid confusion for newcomers. So I'd vote to deprecate use of "Beacon Chain" across the site.

I suspect it may be useful to keep the [Beacon Chain page](https://ethereum.org/en/upgrades/beacon-chain/) intact as a place to explain this history (i.e. that it used to be a standalone blockchain that was merged with Ethereum Mainnet) but in the rest of e.org site content & documentation, my hunch is that it'd be best to avoid "Beacon Chain" references.

[...]

PR #9589, opened 2023-02-28
Remove mention of deprecated Beacon Chain

SebastianSupreme:
Since The Merge, the term "Beacon Chain" is deprecated and what was once known as the "Beacon Chain" has now become the consensus layer of Ethereum.

We no longer have two separate chains but rather one single Ethereum Mainnet with an execution layer and consensus layer.

Even the page about the Beacon Chain describes it in the past tense (https://ethereum.org/en/upgrades/beacon-chain/#main-content).


wackerow:
Hey @SebastianSupreme, thanks for dropping this! To my knowledge, despite using the terms "consensus" and "execution" to refer to these layers and client types, we do still use "Beacon Chain" in the [consensus specs](https://github.com/ethereum/consensus-specs/blob/dev/specs/capella/beacon-chain.md).
@djrtwo Any thoughts here? Is this a term the devs are trying to deprecate entirely?

jmcook1186:
btw there was some discussion around this at merge-time. It was inconclusive, but the arguments are at least laid out in the issue description: https://github.com/ethereum/ethereum-org-website/issues/7828

SebastianSupreme:
Considering that the page about the Beacon Chain speaks of it in past tense, indicating that it doesn't exist anymore, we should be consistent with this and remove the name from pages which are meant to explain Ethereum in simple terms in order to avoid any confusion for new readers (e.g. the staking pages).
Pages which do include a more technical terminology may continue to refer to the Beacon Chain for now because I'm aware that the consensus specs contain remains of Beacon Chain terminology as well.

SebastianSupreme:
@corwintines can we merge this now? According to the linked discussion, most people are in favor of this change.

SebastianSupreme:
@corwintines Can we merge this now? According to the linked discussion, most people are in favor of this change.

wackerow:
I personally disagree with removing the Beacon Chain terminology, but I do agree that the wording presented on the `/roadmap/beacon-chain/` pages is poor. To my knowledge, the intent of rewording the "tense" on that page after upgrades were checked off was not to imply that the Beacon Chain no longer existed. The intro appears to try to tell a story of when the Beacon Chain started, but it definitely gives a tone that it no longer exists.

I'd be more in favor of correcting the tense surrounding this copy than to remove the name "Beacon Chain", which the page has always been named, in included in the URL, and is still used in the specs.

corwintines:
I agree with @wackerow. @SebastianSupreme would you be interested in updating this instead?

SebastianSupreme:
Correct me if I'm wrong but just to get the definition right, what you're saying is that the Beacon Chain still exists as the main part of the  consensus layer, and that the Beacon Chain wasn't altered during the merge, which is why it still exists like it did before, just not as a separate thing. And the consensus layer is basically the module that consists of the Beacon Chain and some other stuff that adds additional like

SebastianSupreme:
Correct me if I'm wrong but just to get the definition right, what you're saying is that the Beacon Chain still exists as the main part of the  consensus layer, and that the Beacon Chain wasn't altered during the merge, which is why it still exists like it did before, just not as a separate thing. And the consensus layer is basically the module that consists of the Beacon Chain and some other stuff that adds additional functionality like communicating with other layers e.g. the execution layer.

I could get behind this definition and if you agree, I'd be happy to rewrite the Beacon Chain page.

SebastianSupreme:
Ok so I've  been thinking and I'd like to propose the following definition: The Beacon Chain still exists as the main part of the  consensus layer, which is why it still exists, just not as a separate thing. And the consensus layer is basically the module that consists of the Beacon Chain and some other stuff that adds additional functionality like communicating with other layers e.g. the execution layer.

If you can get behind this definition I'd be happy to rewrite the Beacon Chain page accordingly.

SebastianSupreme:
Ok so I've been thinking and I'd like to propose the following definition: The Beacon Chain still exists as the main part of the consensus layer, which is why we can still call it that, just not to address the entire consensus layer, which means that both terms aren't interchangeable because the consensus layer is a module of Ethereum that consists of the Beacon Chain and some other stuff that adds additional functionality that is not inherently part of the Beacon Chain itself, like communicating with other layers e.g. the execution layer.

If you can get behind this definition I'd be happy to rewrite the Beacon Chain page accordingly.

SebastianSupreme:
So, it seems like I was wrong and the Beacon Chain does still exist as Ethereum's consensus layer but has been modified compared to the pre-merge Beacon Chain. But even so, it still keeps its name after those updates and being given the purpose to be Ethereum's consensus layer.

I'd be happy to rewrite the Beacon Chain page accordingly and remove the confusion that is caused by the past tense wording.

jmcook1186:
yea i agree - I feel like the window for updating terms has kind of closed now being so far post-merge. BC has persisted and is pretty well embedded now. No need to purge the term from the site generally. That said, I still prefer @SebastianSupreme 's updates using `consensus layer` in most cases tbh.

@SebastianSupreme if you want to try to clarify the page then please go ahead and raise a PR, happy to go along with the consensus here.

corwintines (approved):

corwintines:
@SebastianSupreme could you fix conflicts here, ill pull in after :) Thanks!

SebastianSupreme:
@corwintines should be good to go now

minimalsm:
@jmcook1186 @wackerow please review and confirm we're happy here 😀

wackerow (approved):
Lgtm, just left one small suggestion but not critical
cc: @corwintines

PR #9598, opened 2023-02-28
Update /enterprise content

tasdienes:
## Description

* Remove list of L2s, since we now have better content elsewhere 
* Removed some obsolete/dead enterprise projects and updated others


minimalsm (approved):
Thanks @tasdienes ❤️

PR #9624, opened 2023-03-03
Making it even more clear that publicly exposed APIs are excluded fro…

fredriksvantes:
## Description

Making it even more clear that publicly exposed APIs are excluded from Bug Bounty Program


minimalsm (approved):
Thanks @fredriksvantes

PR #9627, opened 2023-03-03
Chakra - Get Eth page

vdusart:
Converting the `get-eth` page to Chakra

I don't know why but I can't get the `InfoBanner` [here](https://github.com/vdusart/ethereum-org-website/blob/60e8e0ab1192ec2bdee5f9228eb7ab16675bdb78/src/pages/get-eth.tsx#L209) to take the `mt={8}` in account

## Related Issue

Epic: #9353


pettinarip (approved):
Thanks @vdusart this looks great

I've fixed the `InfoBanner` issue by using the component directly wo using the `as` prop. I've also use `Text` and `Heading` for the `p` and `h2`/`h3` tags we had.

PR #9632, opened 2023-03-04
Update community-events.json

Sergey1997:


corwintines:
Hey @Sergey1997

I think this might be a bit early given there is not a lot of information on the website or this event. It is largely just showing 2022 stuff, with not much info on 2023. When there is more info on what to expect for 2023, this can be reopened to look at again.

PR #9634, opened 2023-03-05
refactor(templates/use-cases): migrate to Chakra

TylerAPfledderer:
## Description
Migrate the `use-cases` template to Chakra UI

## Related Issue
#9353


src/templates/use-cases.tsx
```diff
@@ -25,135 +39,66 @@ import Logo from "../components/Logo"
...
+import UpgradeTableOfContents from "../components/UpgradeTableOfContents"
+import TableOfContents from "../components/TableOfContents"
 import Translation from "../components/Translation"
 import SectionNav from "../components/SectionNav"
 import {
   Divider,
   Paragraph,
-  Header1,
-  Header4,
   ListItem,
```
pettinarip, line 50: We need to deprecate these components as well.

- Paragraph => `Text`
- ListItem => `<chakra.li>`

src/templates/use-cases.tsx
```diff
@@ -25,135 +39,66 @@ import Logo from "../components/Logo"
...
-  h1: Header1,
+  h1: H1,
   h2: H2,
   h3: H3,
-  h4: Header4,
+  h4: H4,
   p: Paragraph,
   li: ListItem,
-  pre: Pre,
+  // pre: Pre,
```
pettinarip, line 101: Isn't this used?

src/templates/use-cases.tsx
```diff
@@ -173,134 +118,145 @@ const components = {
...
-    display: none;
-  }
-`
+const HeroContainer = (props: ChildOnlyProp) => (
+  <Flex
+    bg="cardGradient"
+    boxShadow="inset 0px -1px 0px rgba(0, 0, 0, 0.1)"
+    flexDirection={{ base: "column-reverse", lg: "row" }}
+    justifyContent="flex-end"
+    minH="608px"
```
pettinarip, line 127: I think we need to force this height. In prod we have this defined height:
image


```suggestion
    h="608px"
```

pettinarip (changes requested):
gj @TylerAPfledderer I've pointed only a few things.

For example, the page Hero, has a height much bigger than in prod. Take a look at the defi page:
https://ethereumorgwebsitedev01-tylerapfleddererethereumo62717.gatsbyjs.io/en/defi/
vs
https://ethereum.org/en/defi/

src/templates/use-cases.tsx
TylerAPfledderer: Whoops! Not sure what I did here. Adding it back in. 😅

TylerAPfledderer:
@pettinarip this should be good to go!

pettinarip (approved):
lgtm! 💯

PR #9637, opened 2023-03-06
Fix grammar error in scaling page

mikebolt:
"new network designs that enable efficiently confirm" -> "new network designs that enable the network to efficiently confirm"

I added some missing words.


corwintines (approved):
Thanks @mikebolt !

corwintines (approved):
Thanks @mikebolt !

PR #9670, opened 2023-03-08
refactor: migrate template/upgrade to Chakra

amit-ksh:
## Description

Migrate the `upgrade` template to Chakra UI

## Related Issue

#9353


src/templates/upgrade.tsx
```diff
@@ -358,49 +411,57 @@ const UpgradePage = ({
           <DesktopBreadcrumbs slug={slug} startDepth={1} />
           <MobileBreadcrumbs slug={slug} startDepth={1} />
```
amit-ksh, line 412: Why we are using 2 separate Breadcrumbs having the style?  Can anybody explain to me why we are doing this?

amit-ksh:
A little heads up, the Pages are horizontally scrollable around the `1000px` screen width and I couldn't able to figure out why this is happening, Can you check it out?

src/templates/upgrade.tsx
```diff
@@ -185,130 +249,125 @@ const components = {
...
+    display: { base: "none", lg: "block" },
+  },
+})
+
+const MobileBreadcrumbs = chakra(Breadcrumbs, {
+  baseStyle: {
+    mt: 2,
+    display: { lg: "none" },
+  },
+})
```
pettinarip, line 279: Good point. Not sure why we had it like that. But anyway, lets refactor it a bit.

We can remove these two since we are always going to display the Breadcrumbs to the user. We can just use the Breadcrumbs component and apply the necessary styles on it.

```tsx
// remove these
<DesktopBreadcrumbs slug={slug} startDepth={1} />
<MobileBreadcrumbs slug={slug} startDepth={1} />

// just use
<Breadcrumbs slug={slug} startDepth={1} mt={2} />
```

src/templates/upgrade.tsx
```diff
@@ -185,130 +249,125 @@ const components = {
...
+      direction="column"
+      justify="flex-start"
+      position={{ base: "relative", lg: "absolute" }}
+      bg={{ base: "ednBackground", lg: "background" }}
+      border="1px"
+      borderColor="border"
+      borderRadius="base"
+      boxShadow={{ lg: cardBoxShadow }}
+      maxW={{ base: "full", lg: "640px" }}
+      mt={{ base: 0, lg: 12 }}
```
pettinarip, line 362:
```suggestion
```

src/templates/upgrade.tsx
```diff
@@ -185,130 +249,125 @@ const components = {
...
-`
+  return (
+    <Flex
+      direction="column"
+      justify="flex-start"
+      position={{ base: "relative", lg: "absolute" }}
+      bg={{ base: "ednBackground", lg: "background" }}
+      border="1px"
+      borderColor="border"
+      borderRadius="base"
```
pettinarip, line 359:
```suggestion
      borderRadius="sm"
```

pettinarip (changes requested):
Nice @amit-ksh! thanks for the work done here.

Good catch on the Nav issue (on `1000px` screen resolutions). As you mentioned, it is an existing issue so lets not focus on that here. I'll create a separated issue for that.

pettinarip (approved):
🚀 LGTM! thanks @amit-ksh

PR #9681, opened 2023-03-09
refactor(theme): update config for `extendBaseTheme`

### LINKED ISSUE #9679: Use `extendBaseTheme` instead of `extendTheme`
With this approach, we apply the theme for just the components we need or those that we have defined in our custom theme. The main benefit is that this will reduce the amount of JS we bundle, resulting in improved performance.

As a consequence of this change, we’ll need to check and update all the existing component theme configs to ensure that they maintain the same styles and avoid any unexpected visual changes.

TylerAPfledderer:
## Description

Applies Chakra's `extendBaseTheme` function for the custom config and updates the config to reflect this change. 

- The purpose of this function versus the previously used `extendTheme` is to not immediately include all of the default component themes.
- Default themes are imported as needed to ensure the current UI is retained
- A small reusable function is created to allow for recursive merging (deep nested merging) of default and custom styling to prevent any unnecessary overrides.

Certainly open for trimming any of this, either now or in a future PR.

## Related Issue
Closes #9679


src/@chakra-ui/gatsby-plugin/components/components.utils.ts
```diff
@@ -0,0 +1,70 @@
...
+  List: listDefaultTheme,
+  Menu: menuDefaultTheme,
+  Modal: modalDefaultTheme,
+  Select: selectDefaultTheme,
+  Spinner: spinnerDefaultTheme,
+  Switch: switchDefaultTheme,
+  Table: tableDefaultTheme,
+  Tabs: tabsDefaultTheme,
+  Tag: tagDefaultTheme,
+} = theme.components
```
TylerAPfledderer, line 29: Import aliases used for consistency, as some of these defaults are exported to custom component configs.

The rest are imported to retain the current UI with all Chakra components used that have themes, and includes themes for components that may be used later as the UI migration continues and other refactoring needs.

It is possible that this list could be trimmed down with any discovery of themes that are not actually needed.

src/@chakra-ui/gatsby-plugin/components/components.utils.ts
```diff
@@ -0,0 +1,72 @@
...
+ * @param defaultTheming - The related object that comes from the Charka default themes
+ *
+ * @param styleObjs - The following style objects to be merged
+ */
+export function defineMergeStyles(
+  defaultTheming?: SystemStyleObject,
+  ...styleObjs: SystemStyleObject[] | undefined[]
+) {
+  return merge(defaultTheming, ...styleObjs)
+}
```
TylerAPfledderer, line 72: This function was created for re-usability to handle merging of default styles with custom stylings. It uses the lodash `merge` function to handle recursive merging to prevent unexpected overrides.

Though this works as intended, I believe this could be improved, as there are multiple circumstances that this would be used in.

Ex. 1: Simple merge
```tsx
// From custom Link config
baseStyle: defineMergeStyles(linkDefaultTheme.baseStyle, {
    color: "primary",
    textDecoration: "underline",
    _focus: {
      boxShadow: "none",
    },
    _focusVisible: {
      boxShadow: "none",
      outline: "auto",
    },
  },
})
```

Ex. 2: Style function props needed to be passed to a default theme that uses some of those props (type error throws otherwise)
```tsx
// From custom Button config
const variantOutline = defineStyle((props) =>
  defineMergeStyles(defaultVariants?.outline(props), commonOutline)
)
```

Ex. 3: There may be a circumstance where a default baseStyle would be combined with a default variant because the component may not use other variants in the project, thereby compacting the custom config.
```tsx
const baseStyle = defineMergeStyles(
  defaultBaseStyle,
  defaultVariant.outline,
  {
    // Custom styles
  }
)
```

pettinarip (approved):
Amazing @TylerAPfledderer!

A few minor observations but this looks really nice. GJ

src/@chakra-ui/gatsby-plugin/components/components.utils.ts
TylerAPfledderer: @pettinarip 

> Awesome. Would love to have this documented somewhere...I feel bad to lose this doc in this PR.
I can put this up in the PR description as `Additional Information`

With your second suggestion, this approach will return a type equivalent only to the first object, which would be the default theme, and therefor lock down what the second object could have in its properties.

With the first suggestion, this will throw a type error in configs where the function is used in a `defineMultiStyleConfig` property.

For example, the `sizes` property in the [Button theme config](https://github.com/ethereum/ethereum-org-website/blob/1573a8c0b8646016a356c1113c992371cccfd3ba/src/%40chakra-ui/gatsby-plugin/components/Button.ts#L114-L118). The error thrown here:
```bash
Type 'SystemStyleObject' is not assignable to type 'Dict<SystemStyleInterpolation> | undefined'
```

`Dict` is currently a private type in Chakra, but it's simply
```ts
type Dict<T = any> = {
  [ x: string ]: T
}
```

We could apply this assertion on the return, written as `Record<string, SystemStyleObject>` which will work everywhere.
```ts
export function defineMergeStyles(
  defaultTheming?: SystemStyleObject,
  ...styleObjs: SystemStyleObject[] | undefined[]
) {
  return merge(defaultTheming, ...styleObjs) as Record<string, SystemStyleObject>
}
```

pettinarip:
Not a big deal but I was analyzing the final bundles and noticed that the size had an unexpected result. They have basically the same size.

`dev` build (`extendTheme`):
@chakra-ui bundle size 277.01 kb
image

This PR build (`extendBaseTheme`):
@chakra-ui bundle size 286.39 kb
image

I was expecting a smaller size overall but we got an even slightly heavier one. Do you know what could be the reason? maybe is that we are not excluding a lot of Chakra components after all? 🤔 

But again, not a big deal, regardless of the bundle size, I think this is a great improvement to reduce the number of theme styles we have.

pettinarip:
@TylerAPfledderer cool. Then, are you ok if I merge this PR? lgtm.

After that, you can play around with `ChakraBaseProvider` and see if that has any real impact. LMK what you think

TylerAPfledderer:
@pettinarip yes, go ahead and merge! 🚀

PR #9682, opened 2023-03-09
Fix energy consumption copy on staking homepage

minimalsm:
## Description

- Made copy present tense to reflect post-Merge Ethereum

### DIFF
--- a/src/intl/en/page-staking.json
+++ b/src/intl/en/page-staking.json
@@ -5,7 +5,7 @@
   "page-staking-benefits-2-title": "Better security",
   "page-staking-benefits-2-description": "The network gets stronger against attacks as more ETH is staked, as it then requires more ETH to control a majority of the network. To become a threat, you would need to hold the majority of validators, which means you'd need to control the majority of ETH in the system–that's a lot!",
   "page-staking-benefits-3-title": "More sustainable",
-  "page-staking-benefits-3-description": "Stakers don't need energy-intensive computers to participate in a proof-of-stake system–just a home computer or smartphone, making Ethereum environmentally sustainable.",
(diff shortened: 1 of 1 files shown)


src/intl/en/page-staking.json
```diff
@@ -5,7 +5,7 @@
   "page-staking-benefits-2-title": "Better security",
   "page-staking-benefits-2-description": "The network gets stronger against attacks as more ETH is staked, as it then requires more ETH to control a majority of the network. To become a threat, you would need to hold the majority of validators, which means you'd need to control the majority of ETH in the system–that's a lot!",
   "page-staking-benefits-3-title": "More sustainable",
-  "page-staking-benefits-3-description": "Stakers don't need energy-intensive computers to participate in a proof-of-stake system–just a home computer or smartphone. This will make Ethereum better for the environment.",
+  "page-staking-benefits-3-description": "Stakers don't need energy-intensive computers to participate in a proof-of-stake system–just a home computer or smartphone, making Ethereum environmentally sustainable.",
```
jmcook1186, line 8: Looking back at this, i don't love it. In reality the storage requirements are way too high for running nodes on a mobiles, at least until mature light clients roll out. Yes you can run a validator on a Raspberry Pi, but only if you attach an expensive SSD to it! The real benefit of the merge was just that you don't have to waste CPU cycles doing pow computations => low energy consumption.

```suggestion
  "page-staking-benefits-3-description": "Stakers don't need to do energy-intensive proof-of-work computations to participate in securing the network meaning staking nodes can run on relatively modest hardware using very little energy .",
```

wackerow: I do agree with this... I think saying you just need a smartphone to participate in PoS is misleading at this point in time. The rephrasing seems good to me.

wackerow: (I know this wasn't the initial intent of the PR, but I think this is an easy enough fix to just include here if we're okay with it)

corwintines (approved):

PR #9688, opened 2023-03-10
refactor(Search): improve Chakra implementation

### LINKED ISSUE #9573: DocSearch implementation v2
_After the MVP for DocSearch is merged, iterate it one more time, and apply suggestions given by @TylerAPfledderer in [this comment](https://github.com/ethereum/ethereum-org-website/pull/9509#issuecomment-1442373827)._

_Pasting the entire comment here for readers' convenience. Credits to @TylerAPfledderer_ 

=============================

**tldr**: [CodeSandbox Mock Solution](https://codesandbox.io/s/chakra-w-docsearch-mock-2flf11?file=/src/EthDocSearch/index.tsx)

This ended up being a little bit trickier than I was expecting, but my solution results in having to import `DocSearchButton` and `DocSearchModal` instead of `DocSearch`.

Why? In the `DocSearch` component, the modal is rendered in a portal. If I infuse Chakra props to `DocSearch`, the styling would be scoped to a container I would create, and the modal will be outside this scope. This would mean placing the modal styles as high as the global object in the theme config (not desirable).

## 

So I am using the Chakra Factory to allow chakra props to be available on the components. In addition, because the factory is using emotion to generate a class name hash to attach the props to, I'm ensuring the class name is at least forwarded to the component.

Then, using the `sx` prop to declare the needed className selectors and apply Chakra tokens to the styles where possible.

In the case of a modal, I have to add a wrapper element to pass `className` to because the modal does not directly accept it.

With the button, instead of the factory, because it is using CSS flex, I am consuming it with the Chakra `Flex` component, which I think is more appropriate here in the additional level of abstraction.

I copied all the disclosure logic from `DocSearch` and paste it here. This can now allow for further modification of the logic if you want.

This is open to moving around logic between the components for something that would be more desirable, along with logic that is not here but is in the PR. And the styling might have some room for simplification.

Also, can still consider in the Modal to use it's `hitComponent` and `footerComponent` props to customize the search result styles and footer styles directly with Chakra components instead of through CSS selectors at the higher level.

After some thought while making this Sandbox, I can not think of a different or better solution here. I will gladly accept any thoughts or observations to explore any changes here. 😄 

### Additional context

_No response_

TylerAPfledderer:
## Description

This PR improves the DocSearch to better utilize Chakra.

- Imports the `DocSearchButton` and `DocSearchModal` with event logic from `DocSearch`
  - Supercharge these components with Chakra to directly accept props.
- Split all the custom styling into two for these component and place the two styling objects in a utils file.
- Add a story of the `Search` component for isolated visual testing of the component pair
  - Add the `storybook-react-intl` addon to prevent erroring with the `IntlProvider` since translation functionality is included in the component.
  - Include default credentials from the [DocSearch docs](https://docsearch.algolia.com/docs/DocSearch-v3#testing) to be used in development (especially with Storybook) which allows testing filtering using the DocSearch content.

## Related Issue
Closes #9573


TylerAPfledderer:
Closing in favor of a new PR from an updated `dev` branch.

PR #9693, opened 2023-03-10
✍️ Clarification in section on ETH value

z0r0z:
Remove investment language that confuses ether utility.


wackerow:
Thanks @z0r0z! Honestly think this is probably a good adjustment here. Couple things:

- This just removes the English string, but does not remove where the text is trying to be rendered. You'll need to remove the paragraph with `<Translation id="page-eth-has-value-desc-5" />` inside it in the `pages-conditional/eth.tsx` file to actually remove it's usage.
- Would also remove this string from the other language `.json` files, as they wouldn't be needed anymore (identifiable by `"page-eth-has-value-desc-5"` key)
- Mind giving this PR a more descriptive title to help us stay organized?

z0r0z:
thanks @wackerow ! Just made those changes if you want to take a look 👍

minimalsm:
Hey! Thanks for the PR :)

I disagree that the page's current language "confuses ether utility" and don't think we need to change this page. Appreciate the suggestion, but I'm going to close this out

Thanks again!

PR #9701, opened 2023-03-10
fix: replace `disabled` with `isDisabled` for button components

TylerAPfledderer:
## Description

When updating the Chakra UI dep, it introduced [a change in the `Button` component props](https://github.com/chakra-ui/chakra-ui/commit/0f376650efce9eb2d2cba3484ffe4e3a73364504) where `disabled` became `isDisabled`. This addresses that change.


wackerow (approved):
Nice, thanks @TylerAPfledderer!

PR #9712, opened 2023-03-12
Update community-events.json

UrbanHak:


UrbanHak:
Thank you so much for your attention. We are Ethereum Miami. We have launched ETH Mobile at main stage of ETH Denver. https://ethmiami.ooo

corwintines (changes requested):
@UrbanHak should this be called Eth Mobile? This redirects to a different site, so little confused as to this event. Also, it's getting redirected to https://www.ethmobile.io/, so would suggest changing this too.

PR #9723, opened 2023-03-14
Added animation to Find Wallet filter panel in mobile [Fixes #8172]

### LINKED ISSUE #8172: Find wallet filter panel animation

On mobile, when a user opens the filter panel, right now it appears instantly. And when the user closes the panel disappears instantly. This makes for a weird user experience.

When a user opens the filter panel in mobile view, the filter panel should animate in from the left.

When a user closes the filter panel in mobile view, the filter panel should animate out to the left.

None

### Additional context

_No response_

vishal590:


corwintines:
Thanks @vishal590 

The animation looks a bit awkward, but I think this has more to do with this section of the site needing a larger overall refactor. I think you did the best with what you were given, but im going to close this out for now and create an issue for refactoring the find-wallets page table in general.

I do think you are on the right track here with how you implemented it, this is just a bigger issue in the implementation I did last year before we were using Chakra.

PR #9736, opened 2023-03-17
Update index.md

ChaosGod:
in china，java stack called “java 技术栈”，go stack called “go 技术栈”。so ethereum stack called “以太坊技术栈”


nhsz (approved):
Thanks for the fix!

PR #9742, opened 2023-03-17
change the hints order

devas123:
As I understand slots are related to slots and seconds are related to epochs


corwintines:
Hey @devas123 

Appreciate the PR. However, what is here already is correct. 

More information on [Blocks](https://ethereum.org/en/developers/docs/blocks/#proof-of-work-protocol), but each block and slot are every 12 seconds.

Every 32 blocks/slots is an epoch, which finalizes the slots.

PR #9743, opened 2023-03-17
patricia merkle trie: specify array length

ryandotsmith:
## Description

Looks like the sentence was chopped off. Added back '17' with a few grammar tokens.


jmcook1186 (approved):
thanks for catching this!

PR #9753, opened 2023-03-19
Update index.md: Fix some words

shungo0222:
## Description

- producers -> producer's
- recipient -> recipient's


corwintines:
I don't think this is actually correct: https://www.dictionary.com/browse/producer

Thanks for the suggestion @shungo0222, but going to close this.

PR #9768, opened 2023-03-21
Remove redundant `={true}` from boolean attributes

wackerow:
## Description
Boolean element attributes that are set to `true` can be shortened to simply declaring the attribute name by itself. This PR greps the repo for any usage of `attribute={true}` and shortens to `attritbute`

## Related Issue
None filed


nhsz (approved):
Thanks, LGTM!

corwintines (approved):

PR #9769, opened 2023-03-21
Add Code4rena to the Bug bounty platforms list

Simon-Busch:
## Description
Add Code4rena to the Bug bounty platforms list


wackerow (approved):
LGTM, thanks @Simon-Busch !

PR #9219, opened 2023-01-18
Add community lead job posting

corwintines:
## Description

- Add link to community lead job posting on /about
- Add page for community lead job posting at /about/community-lead
- Add back job.tsx template for posting

PR #9773, opened 2023-03-22
Remove community lead job posting

wackerow:
## Description
- Applications have been closed. Removes job posting and link on `/about/` page.
- Also removes `job` markdown template as there are no additional markdown files using this template.

(Essentially reverts #9219)


nhsz (approved):
LGTM, question, is it neccesary to remove the `job` template or could we keep it for future reference in case of other job posts @wackerow?

corwintines (approved):
Thanks @wackerow

@nhsz I know we removed this template before with the design posting, I think it was causing build errors at the time but trying to recall. Either way, I think we can remove it and just add it back when we need again, I had to do that for this initially and really didnt create any work.

PR #9795, opened 2023-03-24
[A11y] refactor(ProductList): improve structure and accessibility

### LINKED ISSUE #8878: A11y Refactor: ProductList Component
## Objective
Improving accessibility and UI markup for the [ProductList](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/ProductList.tsx) component, including UX with screen readers.

## Description
Now that this component has been migrated to ChakraUI for the UI Epic, there might be further improvements that can be made.
- Apply the Chakra `List` component wrapping the mapped components.
- Apply the `ListItem` component to parent component in the map
- Use `aria-labelledby` to connect the list with the heading content (ensuring that if the user simply navigated to the list via screen reader shortcut, they know what type of list this is.
- Add visually hidden text in button link to extend the meaning of the link
  - It currently just renders the text "Go"
  - Add dynamic text with title prop, like `"to { title } website (opens a new window)"` (all links in all instances of this component are external)

TylerAPfledderer:
## Description

Improves accessibility navigating the `ProductList` component with a screen reader.

- Applies the use of Chakra's `List` and `ListItem` components.
- Connects the list with it's heading via `aria-labelledby` to ensure the user knows what the list should they skip the heading
- Adds hidden text for the button link to let the user know that it's a link to an external site. (This component will already read out "opens a new tab")

> NOTE: Using the screen reader NVDA for testing

### Secondary addition to the PR
This PR also updates the layout to provide more consistency with spacing.
- Keeps the button link the same width in all instances.
- Keeps the each item an equal height if two lists are rendered side-by-side.

cc @nloureiro for approval on this

## Related Issue

Closes #8878


src/components/ProductList.tsx
```diff
@@ -22,10 +31,13 @@ export interface IProps {
 const ProductList: React.FC<IProps> = ({ content, category }) => {
   const shadow = useColorModeValue("tableBox.light", "tableBox.dark")
 
+  const ariaLabelledby = "cat-name"
```
corwintines, line 34: I just had a question abut this label. Im a little confused what "cat-name" is. Should the be "category-name", or should we actually be passing the category prop instead?

TylerAPfledderer: This is just an id because it is being used in two places (lines 40 and 51) for the `aria-labelledby` feature.

I can be clearer here in using a more verbose id, along with uppercase variable naming to denote this as a const variable 😁

corwintines:
```suggestion
  const ariaLabelledby = "CATEGORY-NAME"
```

Sure, lets go with this :)

corwintines: Sure! I pushed those changes up. This should be good to go now :)

corwintines (approved):

PR #9796, opened 2023-03-25
Update index.md

saurabhguptasg:
update fromWei and toWei to the new from_wei and to_wei syntax


nhsz (approved):
Thanks for fixing this, LGTM!

PR #9807, opened 2023-03-28
Added taiko to the list of projects working on zkEVMs

seunbayo:
## Description

I added taiko labs to the list of projects working on zkEVMs


d1onys1us (commented):
just some small comments. i think we can just open the PR with only the markdown diff, or is that not the case?

seunbayo:
i will open another pr to this effect

PR #9810, opened 2023-03-28
Update min. requirements for SSD from 1TB to 2TB

cheeky-gorilla:
- Many client pairs require more than 1TB today (e.g. Teku + Besu require +1.1GB)

- EthStaker has been recommending 2TB since before the merge: https://www.reddit.com/r/ethstaker/comments/rizos9/hardware_for_staking_2022_edition/

- eth.org itself recommends 2TB here: https://ethereum.org/en/run-a-node/#build-your-own

- As of right now the verkle trie migration will require ELs to temporarily store a copy of both the merkle patricia tree and verkle trie. Currently that would be just under 2TB, but depending on how soon the verkle trie upgrade is launched (unlikely before Q1 2024), it could pass 2TB in the meantime. (This could be mitigated if state expiry / EIP 4444 is launched in the interim.)

### DIFF
--- a/src/content/developers/docs/nodes-and-clients/run-a-node/index.md
+++ b/src/content/developers/docs/nodes-and-clients/run-a-node/index.md
@@ -89,14 +89,14 @@ All clients support major operating systems - Linux, MacOS, Windows. This means 
 
 - CPU with 2+ cores
 - 8 GB RAM
-- 700GB free disk space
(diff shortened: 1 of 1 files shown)


jmcook1186 (approved):
yes - thanks!

corwintines (approved):
Thanks @cheeky-gorilla

PR #9832, opened 2023-03-29
Fix vision roadmap translation banner bug and rename json keys

minimalsm:
## Description

Previously, /upgrades/vision => was matched to page-upgrades-vision

Now, we have /roadmap/vision => is matching page-roadmap-vision, which doesn't exist. So, the function checkIsPageOutdated does not work as expected https://github.com/ethereum/ethereum-org-website/blob/dev/gatsby-node.ts#L100

- Renaming page-upgrades-vision -> page-roadmap-vision fixes this bug.
- Also cleaned up JSON key strings across the codebase to use this same naming convention

## Related Issue 

None


jmcook1186 (approved):
lgtm!

pettinarip (approved):
🚀

PR #9844, opened 2023-03-30
February Chinese update

corwintines:
## Description
- Crowdin update for Chinese simplified


src/intl/zh/page-upgrades-vision.json
corwintines, line 1: @pettinarip should this have just been renamed to pages-roadmap-vision.json instead of deleted?

pettinarip (approved):

Issue #9843, opened 2023-03-30
Energy consumption chart translation label bug


corwintines:
### Describe the bug

Chart labels on the energy consumption page aren't being translated

<img width="650" alt="Screen_Shot_2023-03-28_at_10 50 57" src="https://user-images.githubusercontent.com/15589226/228870277-0db31827-5b4a-4843-9fff-9036a9754d87.png">

### To reproduce

1. https://ethereum.org/en/energy-consumption
2. Scroll to energy consumption chart
3. See broken labels

### Expected behavior

Labels are translated

### Would you like to work on this issue?

- [ ] Yes
- [ ] No

corwintines:
Closing with above PR merged. https://github.com/ethereum/ethereum-org-website/issues/9857 has also been created from this

PR #9847, opened 2023-03-30
Fix energy consumption chart translations

pettinarip:
## Description

Fixes translations on the energy consumption chart: https://ethereum.org/en/energy-consumption/

Adding the translations in the `page-about` namespace as is the one used for the `static` template (the template used for this page).

## Related Issue

#9843


nhsz:
I have a question, we have the labels translated in german only right?

Screen Shot 2023-03-31 at 15 23 10

They still overlap a bit, maybe we should modify the design? e.g.: adding references on the graph and translated labels below cc @nloureiro @konopkja

corwintines:
@nhsz ive created a separate issue to address what you reported here. 

https://github.com/ethereum/ethereum-org-website/issues/9857

Id like to get this merged in so that we have the labels again, and address this in that issue

corwintines (approved):

pettinarip:
@nhsz good point. Yes, currently we have translations only for `en` and `de`. We can follow up the conversation on #9857.

Thanks @corwintines

PR #9849, opened 2023-03-30
February French updates

corwintines:
## Description
- Crowdin updates for French in February


src/intl/fr/page-upgrades-vision.json
corwintines, line 1: @pettinarip should this have just been renamed to pages-roadmap-vision.json instead of deleted?

corwintines:
- [ ] @corwintines remove consensus mechanism pages from here

pettinarip (approved):

PR #9854, opened 2023-03-30
February Spanish updates

corwintines:
## Description
- Crowdin updates for Spanish in February


src/intl/es/page-upgrades-vision.json
corwintines, line 1: @pettinarip should this have just been renamed to `pages-roadmap-vision.json` instead of deleted?

pettinarip (approved):

Issue #9548, opened 2023-02-22
Implementation of the DS v1


pettinarip:
_This is part of our [Implementation of the DS](https://github.com/ethereum/ethereum-org-website/issues/9546) epic._

This issue will track the progress to implement the [v1 of the DS](https://github.com/ethereum/ethereum-org-website/issues/6284).

## How to contribute

Currently, we're in the planning stage and creating a GitHub issue for each item on the list below. If you'd like to contribute, please visit the relevant issue when it's created and leave a comment to get assigned.

## TODO

1. Create stories for existing components that have an equivalent in the DS
    - [ ]  Detect and create a list of these components
    - [ ]  Create new issue to create `stories.tsx` for them
2. Setup the foundation styles from the DS
    1. Brand colors
        - [ ]  Setup colors on theme
        - [ ]  Implement new colors
    2. Base colors
        - [ ]  Setup colors on theme
    3. Typography
        - [ ]  Setup & load new fonts
        - [ ]  Define with designers
            1. Text styles to be used from now on
            2. Default `fontSizes` & `lineHeight` for the Heading component
        - [ ]  Create new [textStyles](https://chakra-ui.com/docs/styled-system/text-and-layer-styles#text-styles)
        - [ ]  Refactor codebase as much as we can to use these new text styles
3. Create base components
    - [ ]  Social media links
        - [ ]  IconLink
    - [ ]  Button
        - [ ]  Button
        - [ ]  ButtonLink
        - [ ]  IconButton
    - [ ]  Forms
        - Codesandbox: [https://codesandbox.io/s/eth-form-components-mock-o2fwdg](https://codesandbox.io/s/eth-form-components-mock-o2fwdg)
        - [ ]  Input
        - [ ]  Radio buttons
        - [ ]  Switch
        - [ ]  Checkbox
        - [ ]  Dropdown
    - [ ]  Tooltip
    - [ ]  Avatar
        - Codesandbox: [https://codesandbox.io/s/eth-avatarlink-mock-o0ivnz](https://codesandbox.io/s/eth-avatarlink-mock-o0ivnz)
    - [ ]  Navigation
    - [ ]  Accordion
    - [ ]  Tag
        - Codesandbox: [https://codesandbox.io/s/eth-tag-mock-5nlzuy](https://codesandbox.io/s/eth-tag-mock-5nlzuy)
    - [ ]  Table
[...]

PR #9856, opened 2023-03-31
feat(theme): add/update form component themes

### LINKED ISSUE #8630: Create a Chakra `Select` component
*This is part of our [2nd wave of migration](https://github.com/ethereum/ethereum-org-website/issues/8632) of the [Implement UI library](https://github.com/ethereum/ethereum-org-website/issues/6374) epic.*

Right now we have a [`Select` component](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/SharedStyledComponents.tsx#L443) that is styled using styled components and uses `react-select`.

- Deprecate the usage of the current component
- ~~Create a new component called `Select` under `src/components` that is wrapped and styled using Chakra.~~
- Create the theme for the [Chakra Select](https://chakra-ui.com/docs/components/select/usage) component
- Replace the old with the Chakra Select component in the entire codebase

### Additional context

Be aware of this: there are cases where we are adding more styles to the base styles of the Select.
- [`FindWallet`](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/FindWallet/WalletTable.tsx#L143)

TylerAPfledderer:
## Description

_This is part of the [Implementation of the DS v1](https://github.com/ethereum/ethereum-org-website/issues/9548)_

Add or updates theming configs for form components to adhere to the new design system.

- Adds/updates theming for Radio, Switch, Checkbox, Select (Dropdown), and Input
- Creates custom component for Input
- Adds Storybook stories for each component
- Update other components or pages where necessary for immediate inclusion
  - This is includes adding `Select` implementation as a replacement for the `react-select` lib.

## Related Issue
Closes #8630


TylerAPfledderer:
@pettinarip I have not included replacing any instance of `react-select` here as I felt that would be better served in a separate PR at a later time if not immediately following this one. There seems to be a fair amount of logic (i.e. in the [WalletTable](https://github.com/ethereum/ethereum-org-website/blob/9637d84066c5c5cab3b660aca585dff17a663db0/src/components/FindWallet/WalletTable.tsx#L963-L1011)) being used where a game-plan is needed to tackle them if a custom component is not going to be created.

src/@chakra-ui/gatsby-plugin/components/Select.ts
```diff
@@ -0,0 +1,68 @@
...
+  [$borderBaseWidth.variable]: "1px",
+  borderColor: "currentColor",
+  cursor: "pointer",
+  py: 2,
+  ps: 2,
+  pe: 8,
+  _active: {
+    bg: "primaryLowContrast",
+    boxShadow: "none",
+  },
```
TylerAPfledderer, line 20: Adding active styling is creating a side-effect where on open of the dropdown, a lot of flickering can occur on the options.

src/components/Select/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+} from "@chakra-ui/react"
+
+interface SelectProps
+  extends Omit<ChakraSelectProps, "defaultValue">,
+    Pick<ThemingProps<"Select">, "variant"> {
+  defaultValue?: { label: string; value: string; [x: string]: any }
+  onChange(selectedOption: any | ""): void
+  placeholder?: string
+  options: Record<string, any>[]
+}
```
TylerAPfledderer, line 15: I was hoping to create a generic type to be able to take the type passed into the `options` array and apply to the payload for `onChange`. But I do not believe I have enough knowledge here to know how to correctly apply it since there are different types of data being sent to the `options` prop based on the four different components that use this component.

TylerAPfledderer (commented):
@pettinarip added the custom `Select` component to replace the `react-select` lib. A couple of thoughts on this and applying the DS styling.

src/@chakra-ui/gatsby-plugin/components/Select.ts
```diff
@@ -0,0 +1,68 @@
+import { selectAnatomy } from "@chakra-ui/anatomy"
+import { createMultiStyleConfigHelpers, cssVar } from "@chakra-ui/react"
+import { defineMergeStyles, selectDefaultTheme } from "./components.utils"
+
+const { defineMultiStyleConfig, definePartsStyle } =
+  createMultiStyleConfigHelpers(selectAnatomy.keys)
+
+const $borderBaseWidth = cssVar("border-base-width")
+
+const baseStyleField = defineMergeStyles(selectDefaultTheme.baseStyle?.field, {
```
pettinarip, line 10: Question.

Noticed that the native select is inheriting all the field styles from the input component. https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/theme/src/components/select.ts#L15

Do you think we should merge our Input theme in here instead of the default select?

Something like:
```suggestion
const baseStyleField = defineMergeStyles(OutInputTheme.baseStyle?.field, {
```

The problem I see is that our select & input designs do not share some things, like the focus border. Probably we need to align on that first.

src/@chakra-ui/gatsby-plugin/components/Select.ts
```diff
@@ -0,0 +1,68 @@
...
+  pe: 8,
+  _active: {
+    bg: "primaryLowContrast",
+    boxShadow: "none",
+  },
+  _hover: {
+    boxShadow: "buttonHover",
+    color: "primary",
+  },
+  _focusVisible: {
```
pettinarip, line 25: For some reason, this focus is not working 100% as I would expect. Same happens to the buttons.

When you click on them, we are showing these focus styles, and I wouldn't expect that to happen. Let me show you:

https://user-images.githubusercontent.com/468158/234983124-f1aa49fc-e721-4a98-89a8-f9c2212348fc.mp4

I would not expect to have the thick border on it when you click on it. Same with the button.

Perhaps we need to override the `_focus: null` here as well. What do you think?

pettinarip (commented):
@TylerAPfledderer this is an amazing job! 👏🏼 

Left a couple of thoughts, questions, and feedback.

Would love to have @nloureiro eyes on this as well.

src/@chakra-ui/gatsby-plugin/components/Select.ts
TylerAPfledderer: > The problem I see is that our select & input designs do not share some things, like the focus border. Probably we need to align on that first.

That was the original concern in pulling the input theme. 

Indeed, the default Chakra theming for Select also inherits styling from the Input, so I would've expected doing the same.

src/@chakra-ui/gatsby-plugin/components/Input.ts
```diff
@@ -0,0 +1,109 @@
...
+        _placeholder: {
+          color: "disabled",
+          opacity: 1,
+        },
+        _focusVisible: {
+          borderColor: fc,
+          boxShadow: "none",
+        },
+        _hover: null, // override default
+        _invalid: {
```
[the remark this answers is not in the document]
TylerAPfledderer: @pettinarip Hmmm no. I thought I asked about Nuno about it before though but I do not see a comment. This can be removed if preferred.

src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from "react"
+import { CheckboxGroup, VStack } from "@chakra-ui/react"
+import { Meta, StoryObj } from "@storybook/react"
+import CheckboxComponent from "."
```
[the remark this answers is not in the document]
TylerAPfledderer: Simple enough!

src/@chakra-ui/gatsby-plugin/components/Select.ts
TylerAPfledderer: Interesting! This might be browser specific, because this is not an issue in Firefox.

In Chrome, this is coming from `focus-visible`, not `focus`. Might need to find an override for Chromium browsers?

TylerAPfledderer:
@pettinarip looks like the Chromatic checks are lost after I made my latest push 🤔

src/@chakra-ui/gatsby-plugin/components/Input.ts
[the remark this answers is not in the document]
pettinarip: @nloureiro do you have an opinion about these styles when the form is invalid? do you think it is worth having those now? perhaps we can decide this later

src/@chakra-ui/gatsby-plugin/components/Select.ts
pettinarip: All right, lets not do that then yet. We can polish this later.

src/@chakra-ui/gatsby-plugin/components/Select.ts
pettinarip: Uhmm...I'd expect it to work as in the Chakra docs examples. These [buttons](https://chakra-ui.com/docs/components/button/usage) don't display the focus border when I click them. And yes, if you look at the [theme file](https://github.com/chakra-ui/chakra-ui/blob/main/packages/components/theme/src/components/button.ts#LL11C10-L11C10) they are using the `focusVisible` event as well 🤔 that is why I was thinking on overriding the `focus` event to check if that fixed the issue.

pettinarip:
> @pettinarip looks like the Chromatic checks are lost after I made my latest push thinking It should've ran again because of the `synchronize` action type

That's weird 🤷🏼 not seeing that happening on other PRs. Perhaps it was just a temporary issue.

pettinarip (commented):
@TylerAPfledderer I want to call out that the select and the input should be aligned in size with the buttons.

That means that they should have:
- `md` => 42px height => currently they are 40px
- `sm` => 30px height => currently they are 32px

If you check the Button theme file, I have [hardcoded the 42px value](https://github.com/ethereum/ethereum-org-website/blob/dev/src/%40chakra-ui/gatsby-plugin/components/Button.ts#L116), which is not ideal.

Ideally, we would have a `rem` values for them:
- for `md` 42px => would be `2.625rems`
- for `sm` 30px => would be `1.875rems`

Questions:
- Do you know how we could define those custom sizes in the theme?
- Once we figure out the prev question, could we apply those sizes to all the previously mentioned components?
- What do you think about it?

TylerAPfledderer:
> @TylerAPfledderer I want to call out that the select and the input should be aligned in size with the buttons.
> 
> That means that they should have:
> 
>     * `md` => 42px height => currently they are 40px
> 
>     * `sm` => 30px height => currently they are 32px
> 
> 
> If you check the Button theme file, I have [hardcoded the 42px value](https://github.com/ethereum/ethereum-org-website/blob/dev/src/%40chakra-ui/gatsby-plugin/components/Button.ts#L116), which is not ideal.
> 
> Ideally, we would have a `rem` values for them:
> 
>     * for `md` 42px => would be `2.625rems`
> 
>     * for `sm` 30px => would be `1.875rems`
> 
> 
> Questions:
> 
>     * Do you know how we could define those custom sizes in the theme?
> 
>     * Once we figure out the prev question, could we apply those sizes to all the previously mentioned components?
> 
>     * What do you think about it?

### Short Answer
This will be fine if new spacing tokens are added. The design concern is more about the spacing between the left/right icons and the text field with the larger size input. Maybe a minor thing, but it is there.

### Long Answer
The easiest approach here would be to simply add two new spacing tokens: `7.5` and `10.5` to apply.

However, for the input, note that the positioning of the icons around the text field depends on the height. This is applied in the [InputGroup component](https://github.com/chakra-ui/chakra-ui/blob/997e3083f282f73c8a85bb1e612e1062f1f6d715/packages/components/input/src/input-group.tsx#L41-L74). This means that the left/right padding for the input field will be equal to the height, so there is adequate space for those icons (which the width of their wrapping `div` will also be equal to the height.

Won't be able to easily override this because the elements are being cloned with this updated styling from the InputGroup.

I made some adjustments locally with this in mind, and put focus on the size of the right element in the DS, and got this:
image

In fact, the smaller size is more inline with how `InputGroup` handles the spacing as I mentioned above (the distance of the placeholder text to the left side of the component in the DS is equal to the height). It's the bigger size that does not show this consistency.

src/@chakra-ui/gatsby-plugin/components/Select.ts
TylerAPfledderer: @pettinarip 
Hmmm I think it is a difference in pseudo state behavior. With a button, you would get the similar behavior with the focus styling on click if styling was applied to `:focus` and not `:focus-visible`. This is consistent between Firefox and Chrome.

I would not expect an inconsistency in browser heuristic here with `:focus-visible`.

But check out [W3 Specs for :focus-visible](https://www.w3.org/TR/selectors-4/#the-focus-visible-pseudo)

And specifically the bullet point for when the User Agent may or may not indicate focus with `:focus-visible`:

`If the user interacts with the page via a pointing device (mouse, touchscreen, etc.) and the focused element does not support keyboard input, don’t indicate focus`
(Maybe??)

pettinarip:
@TylerAPfledderer good observation. I don't think the padding difference between the sizes is going to be a big problem.
What's your opinion with that issue @nloureiro?

> The easiest approach here would be to simply add two new spacing tokens: 7.5 and 10.5 to apply.

Yes! lets do that. I don't think we are going to use those values other than in the inputs and buttons.

TylerAPfledderer:
@pettinarip need assistance getting the conflicts resolved 😄

nloureiro:
Question from my side: should we add the error state? I don´t have it in the DS, but the "is required" should have some visual difference.

What do you think?

Screen Shot 2023-06-09 11 30 21 AM

nloureiro:
The read-only should not have hover behavior because it's read-only.

https://github.com/ethereum/ethereum-org-website/assets/1120748/e09d0501-47ae-426a-8b16-899ad2e9087c

nloureiro:
We now use this search "button", and on this PR we have the old one. maybe we should  update it if possible

Screen Shot 2023-06-09 11 21 52 AM

nloureiro:
Is the spacing to the label relevant for this PR?


Screen Shot 2023-06-09 12 00 52 AM
Screen Shot 2023-06-09 11 30 21 AM

pettinarip:
> We now use this search "button", and on this PR we have the old one. maybe we should update it if possible

I'll update the branch now but I see the same search button here and in prod 🤔 I'm a bit confused on what you see as the difference.

pettinarip:
> The read-only should not have hover behavior because it's read-only.

@TylerAPfledderer is that something you know how to add?

nloureiro:
> 

The cursor should not indicate the is an action there.

In the case of the disabled should have the not-allowed
```
 .not-allowed {
  cursor: not-allowed;
}
```

nloureiro:
After a second review, I want to request a small change on the hover color for the checkbox and radio, the blue was too intense.
changed to body medium

Screen Shot 2023-06-15 09 45 01 AM

nloureiro:
On the switch,  are we using the strong grey as a default?
Maybe that's me but I could not be sure on chromatic preview

Screen Shot 2023-06-15 09 39 22 AM

TylerAPfledderer:
@pettinarip @nloureiro should be good now for another review.

src/@chakra-ui/gatsby-plugin/components/components.utils.ts
```diff
@@ -70,3 +72,62 @@ export function defineMergeStyles(
...
+    [_notDisabledReadOnly]: {
+      // Hovering over the label triggers the style for the control
+      "*:hover > &": {
+        bg: "body.light",
+        borderColor: "primary.highContrast",
+      },
+    },
+    _invalid: {
+      // TODO: Investigate inconsistency in prop rendering order (possible Chakra bug)
+      // border: "2px",
```
TylerAPfledderer, line 116: For whatever reason, using `border` and `borderColor` here works fine for Checkbox, but rendered in reverse order for Radio, cause the `borderColor` styling to be overriden in the CSS by the `border`.
image

Now this is probably due to using the default theming which includes an `invalid` styling, but I think a refactor in how I am applying these common props might fix it (send  these props to the checkbox and then import the checkbox styles to the radio and switch). Otherwise, this might be a compiling bug.

Either way, I think the reorganizing and addressing the prop issue should be saved for a separate PR to get this PR over the finish line... unless you would like for it to be resolved here.

pettinarip: Ok, thanks @TylerAPfledderer.

src/@chakra-ui/gatsby-plugin/components/components.utils.ts
```diff
@@ -70,3 +72,66 @@ export function defineMergeStyles(
...
+    [_notDisabledReadOnly]: {
+      // Hovering over the label triggers the style for the control
+      "*[data-checked]:hover > &": {
+        bg: "primary.hover",
+        borderColor: "primary.highContrast",
+      },
+      "*:not([data-checked]):hover > &": {
+        bg: "body.light",
+        borderColor: "primary.highContrast",
+      },
```
pettinarip, line 116: @TylerAPfledderer I've updated the hover colors to match the DS. Not sure if there is a more concise way for these selectors but it works as expected.

We want to display `primary.hover` on checked and `body.light` on unchecked ones.

src/@chakra-ui/gatsby-plugin/components/Switch.ts
```diff
@@ -0,0 +1,66 @@
...
+    bg: "body.medium",
+    borderColor: "transparent",
+    _invalid: { borderColor: "error.outline" },
+    _focusVisible: { bg: "gray.500", borderColor: "transparent" },
+    ["&:not([data-checked])[data-disabled]"]: {
+      bg: "transparent",
+      borderColor: "disabled",
+    },
+    [_notDisabledReadOnly]: {
+      "*[data-checked]:hover > &, *:not([data-checked]):hover > &": {
```
pettinarip, line 50: @TylerAPfledderer I had to do this to catch the two possible states and override the common styles. The switch is slightly different in terms of hover styles than the radio and checkbox (check DS).

pettinarip (approved):
Looks good. Lots of good stuff here, thanks @TylerAPfledderer

nloureiro (approved):
Looks good!!!

src/@chakra-ui/gatsby-plugin/components/Input.ts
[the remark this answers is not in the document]
pettinarip: This was before you created the styles for the invalid state. Its outdated now, will mark it as resolved.

Issue #8955, opened 2022-12-20
Design Developers docs - The fire starter 


nloureiro:
Usually, the title "Developers docs" refers to technical documentation to developers to follow and learn new technologies or guide them on building new applications.

At etheteum.org, this documentation is a reference to the Ethereum community, targeting beginners and intermediate users who want to learn more about Ethereum.

All new technologies bring new design challenges, and it's clear that design documentation on web3 is fundamental and lacking in the ecosystem.

This issue aims to start the discussion around what this content should be like. 

The fire starter was the contribution from [Elizabeth Bassey](https://github.com/Dev-Liz) made on a comment in the [Design system epic](https://github.com/ethereum/ethereum-org-website/issues/6284) that couldn't be more aligned with the internal roadmap discussion for Q1 2023.

A clear action item to this issue will be an EPIC dedicated to this subject, but for now, this issue will be the starting point of the discussion with the community.

The original comment is [here](https://github.com/ethereum/ethereum-org-website/issues/6284#issuecomment-1358302715), and this is a copy of the text to make it easy to have the discussion on the same page.

-----------

```
Getting started in web3 as a designer, can be very daunting because there are little to no resources available for designers (unlike developers)

ethereum.org could cover topics that would easily onboard designers from web2 into web3. Such as:

Introduction to web3 Design

Difference between web1, web2 and web3 Designs (what makes web3 different)
Web3 Fundamentals

Web3 Terminologies
Web3 tokens (and why we need them) ?
Web3 wallets
Dapps
DAOs (How they work)
Web3 Design principles
Analyzing web3 products

Ui Elements and components design
Interaction and flow
Web3 Design Basics

Wallet interaction
Interaction with blockchain protocols
Defi Product Design
NFT Product Design
Designing for web3 Governance
User Research and Usability testing in web3

Glossary

Resources

Contributing to web3 Design projects

_We could further break each topic down into subtopics._
```

Dev-Liz:
Thank you for this! @nloureiro 

I suggest we include a navigation link - **Designers** on Ethereum.org landing page (like there is for Developers).

This section would be a documentation for web3 designers which would include:

- Tutorials
- How-to-guides
- Explanation &
- References

(Using [diataxis](https://diataxis.fr/) as a framework for proper documentation).

All content topics can be grouped under each of the framework's categories (Tutorials, How-to-guides, Explanations & References).


In the meantime, we could start by:
- Creating a timeline for the project
[...]

PR #9858, opened 2023-03-31
Design and UX in web3

konopkja:
## Description

Introduction to research and design in web3.  This is supposed to be a one  stop repositorz of useful links for designers and researchers coming to web3.

[link](https://ethereumorgwebsitedev01-konopkjadesignanduxdocs.gatsbyjs.io/en/developers/docs/design-and-ux/) to preview

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/8955


nloureiro:
@konopkja do have an idea where will this page be linked? from the dev docs left-side navigation?

konopkja:
> @konopkja do have an idea where will this page be linked? from the dev docs left-side navigation?

either in foundational topics before its robust enough, or as a standalone

nhsz:
Having this under `/developers` is a bit weird to me and will hurt discoverability for sure, maybe we should have a `/design` route? IMO needs a bit more discussion

konopkja:
> Having this under `/developers` is a bit weird to me and will hurt discoverability for sure, maybe we should have a `/design` route? IMO needs a bit more discussion

i agree. In an ideal solution it should not be under devs, but currently I dont see any quick solution for this. @nhsz do you have any ideas? how do you imagine "/design" to be implemented in the navigation?

nhsz:
> > Having this under `/developers` is a bit weird to me and will hurt discoverability for sure, maybe we should have a `/design` route? IMO needs a bit more discussion
> 
> i agree. In an ideal solution it should not be under devs, but currently I dont see any quick solution for this. @nhsz do you have any ideas? how do you imagine "/design" to be implemented in the navigation?

I see no need to rush this . Don't know what could be the best approach as we currently don't have this kind of resources on the site, but I'd say maybe a separated hub could work better, sounds like it could be its own epic, is this planned? cc @minimalsm

konopkja:
There are no deadlines, but I also don't see any risks with publishing this in dev docs. There are a few benefits to having this live. (indexing, design communities pointing to this, designers reading... preliminary feedback was very good from the communities I contacted)  

There is no epic around this for foreseeable future, but if this gets traction I can see it as good evidence of having more commitment in the next quarters.

nhsz:
> There are no deadlines, but I also don't see any risks with publishing this in dev docs. There are a few benefits to having this live. (indexing, design communities pointing to this, designers reading... preliminary feedback was very good from the communities I contacted)
> 
> There is no epic around this for foreseeable future, but if this gets traction I can see it as good evidence of having more commitment in the next quarters.

Not the best for SEO hosting this `/developers`, that's my main pushback. Have you considered a subsection under [`/guides`](https://ethereum.org/en/guides/)?

nhsz:
> > 
> 
> "Guides" section is for different target audience: beginners learning how to use Ethereum. Devs actually is not that far from design, in some methodologies designers are considered to be a "dev" role too (eg in scrum you have 3 roles: PO, dev, scrum master ). If the main pushback is SEO, while I agree with the sentiment, I don't think its a big enough issue, also with that logic we already have foundational topics that should not be there. If anything for future restructuring, it helps to be aware of all different areas that should be considered.

Navigation and information hierarchy is a current issue we have, if we start to use `/developers` to host any slightly related content these issues are just going to keep increasing. As a long term solution, IMO it should have it's own hub, with proper categories and navigation. If not, I'd say at least should be hosted on a different route.

konopkja:
> 

I dont think so, that page is for ethereum.org design contributors while this list is for the designers/product owners/ researchers in the ethereum ecosystem.

corwintines:
I think this is fine to have in the developer docs. Developer can mean many things, and plenty of those docs dont have anything to do with development as it is. Overall, that section of the site needs its information architecture looked at and rethought, which I think would result in a design hub, developer hub, general docs, etc. But, at the moment I think its fine to bring in there. 

A note on this though that still needs to be done, we need to add this to the nav still, otherwise no one will be able to navigate to it. 

TLDR, I think this is fine to add in the developer docs for now, Q3 we should begin actually rethinking this IA of the docs though.

corwintines (approved):

PR #9868, opened 2023-04-03
feat(Avatar): create component, theme, and stories

TylerAPfledderer:
## Description

This PR creates the Avatar component per the new DS

- Builds the component and Chakra theme
- Adds a set of stories per the DS requirements
  - Avatars are always rendered as links
  - Story for Single Avatars
  - Story for a Group of Avatars
  - Story for single avatars with text (in the form of using the `LinkBox`/`LinkOverlay` component pair)

## Related Issue
This is a part of the implementation of the DS v1 #9548


.storybook/main.js
```diff
@@ -3,13 +3,7 @@ const { propNames } = require("@chakra-ui/react")
...
 
 module.exports = {
-  stories: [
-    {
-      directory: "../src/components",
-      titlePrefix: "Components",
-      files: "**/*.stories.tsx",
-    },
-  ],
+  stories: ["../src/components/**/*.stories.tsx"],
```
TylerAPfledderer, line 6: This changed can be placed in a separate PR if preferred (not going to impact title in the story file of this PR)

.storybook/preview.js
```diff
@@ -25,6 +25,9 @@ window.___navigate = (pathname) => {
 
 export const parameters = {
   actions: { argTypesRegex: "^on[A-Z].*" },
+  backgrounds: {
+    disable: true,
+  },
```
TylerAPfledderer, line 30: To remove the dark mode toggle button that Storybook renders.

This is so that only the toggle from the Chakra addon is visible.

This can also be added on a separate PR if desired.

.storybook/preview.js
```diff
@@ -34,6 +37,7 @@ export const parameters = {
   chakra: {
     theme,
   },
+  layout: "centered",
```
TylerAPfledderer, line 40: Global positioning of all the stories to be in the center of the preview container

Can also be added on a separate PR if desired.

src/@chakra-ui/gatsby-plugin/components/components.utils.ts
```diff
@@ -65,8 +65,11 @@ export {
...
-  defaultTheming?: SystemStyleObject,
-  ...styleObjs: SystemStyleObject[] | undefined[]
+  defaultTheming?: SystemStyleObject | unknown,
+  ...styleObjs: SystemStyleObject[] | unknown[]
 ) {
-  return merge(defaultTheming, ...styleObjs)
+  return merge(defaultTheming, ...styleObjs) as Record<
+    string,
+    SystemStyleObject | any
+  >
```
TylerAPfledderer, line 74: Looks like this change did not get added to #9681 as originally intended.

Can be added to a separate PR if desired.

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
```diff
@@ -11,7 +11,7 @@ const colors = {
   },
   blue: {
     100: "#dedeff",
-    300: "#ababfe",
+    300: "#8282ff",
```
TylerAPfledderer, line 14: This color value is part of the current version of the DS

.storybook/main.js
```diff
@@ -87,7 +81,4 @@ module.exports = {
       },
     },
   },
-  docs: {
-    autodocs: true,
-  },
```
TylerAPfledderer, line 92: Remove the auto-generated docs for each story set, and only add them as needed.

Can be done in a separate PR if desired

nloureiro:
@TylerAPfledderer and @pettinarip Made a few changes to the DS Figma file and commented on the changes there

https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?type=design&node-id=2402%3A45505&t=ggIw3B0PhQylUIH8-1

Let me know if you need more details?

src/@chakra-ui/gatsby-plugin/foundations/colors.ts
pettinarip: FYI this is already updated in `dev`

src/@chakra-ui/gatsby-plugin/components/Avatar.ts
```diff
@@ -0,0 +1,87 @@
...
+
+const $border = cssVar("avatar-border-color", "transparent")
+const $mlBySize = cssVar("ml-by-size")
+
+const baseStyleContainer = defineStyle((props) =>
+  defineMergeStyles(defaultBaseStyle?.(props).container, {
+    [$border.variable]: "transparent",
+    borderWidth: "1px",
+    "&:hover, [data-peer]:hover ~ &": {
+      boxShadow: "buttonHover",
```
pettinarip, line 23: Can't remember where we ended up with this but talking async with @nloureiro, we said that it would be nice if we can use a more dynamic shadow size for the Avatar.

```tsx
// we tested with 0.15em and its fine
boxShadow="0.15em 0.15em 0 color"
```
but I think this will only apply for the Avatar, the other components should keep using the `primary` shadow.

Let me know what do you think on this implementation @TylerAPfledderer

src/@chakra-ui/gatsby-plugin/components/Avatar.ts
```diff
@@ -0,0 +1,87 @@
...
+const pickedDefaultSizes: { [k in (typeof USED_SIZES)[number]]?: object } =
+  pick(defaultSizes, ...USED_SIZES)
+
+const sizes = defineMergeStyles(pickedDefaultSizes, {
+  xs: {
+    group: {
+      [$mlBySize.variable]: "space.-1",
+    },
+    excessLabel: {
+      fontSize: "0.563rem",
```
pettinarip, line 70: Curious about this value. How did you get it?

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -101,6 +101,12 @@ const semanticTokens = {
         "linear-gradient(102.7deg, rgba(185, 185, 241, 0.2) 0%, rgba(84, 132, 234, 0.2) 51.56%, rgba(58, 142, 137, 0.2) 100%)",
     },
   },
+  shadows: {
+    buttonHover: {
+      _light: "4px 4px 0 var(--eth-colors-blue-100)",
+      _dark: "4px 4px 0 #352313",
+    },
+  },
```
pettinarip, line 109: Nice, is the idea to delete the `shadows` file at some point? or should we still keep using that file to separate a bit the tokens.

TylerAPfledderer: @pettinarip this is here because a semantic color is not being declared in the DS for this style of shadow, so a semantic has to be made for each color based on color mode.

nloureiro:
The spacing and size of the shadow still is a blocker for me, I might be neat picking but would be cool if we could adjust this

Screen Shot 2023-06-29 09 23 18 AM

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
pettinarip: We need to use `primary.lowContrast` here but our `orange.800` is outdated.

We need to update `orange.800` to `#352313` and then we can use `primary.lowContrast`

pettinarip:
> The spacing and size of the shadow still is a blocker for me, I might be neat picking but would be cool if we could adjust this

What do you exactly mean with the spacing? the current spacing in the code is 0.25rem = 4px. On the other side, we are going to implement a dynamic shadow using `0.15em` but the spacing is going to be the same always.

pettinarip (changes requested):
Ok, in summary, I think what is left in this PR is:

- dynamic shadow using `em`
- fix shadow color using `primary.lowContrast` and updating the `orange.800` color
- fix the spacing from 4px to 8px

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
TylerAPfledderer: With the latest updates from the base branch, this has taken care of itself. Still have to push a commit to remove `buttonHover` from `semanticTokens`.

src/@chakra-ui/gatsby-plugin/components/Avatar.ts
TylerAPfledderer: In the DS for the `xs` size avatar, the excess label is showing a font size of `9px` which is this in rem.

This value is not a part of the default fontSize tokens. `2xs` in the default theme is `0.625rem`, and `3xs` is `0.45rem`.

src/@chakra-ui/gatsby-plugin/components/Avatar.ts
TylerAPfledderer: @pettinarip I've pushed this change for you to view.

Though it looks very minor. I'm seeing clipping which does not occur if pixel values are applied. 
image

Not sure what is causing this on my end.

TylerAPfledderer:
> Ok, in summary, I think what is left in this PR is:
> 
>     * dynamic shadow using `em`
> 
>     * fix shadow color using `primary.lowContrast` and updating the `orange.800` color
> 
>     * fix the spacing from 4px to 8px

All items have been addressed.

As to the spacing between avatar and link text, the DS shows padding around the text container, which contributes to the larger gap.

So there is a 4px gap between the avatar and text container, plus 4px padding in the text container.

pettinarip (approved):
I've a question about the font sizes used in the avatar text but not a blocker. Good job!

Issue #9840, opened 2023-03-30
Show execution layer client diversity of staking services


one-three-three-seven:
### Is your feature request related to a problem? Please describe.

I'm currently looking for a staking service which does not use Geth.

I have noticed that some of the staking services at https://ethereum.org/en/staking/saas/ have a check mark on the "DIVERSE CLIENTS" field. I have asked the support from one these services about this topic and the support responded with the info, that they **only** use Geth for the execution layer.

The "DIVERSE CLIENTS" field seems currently only used for the consensus layer. But the execution layer is actually also very important. The shown information is misleading.

### Describe the solution you'd like

I would suggest to add a new metric "DIVERSE EXECUTION CLIENTS" and rename "DIVERSE CLIENTS" to "DIVERSE CONSENSUS CLIENTS".

### Describe alternatives you've considered

We could also remove the "DIVERSE CLIENTS" check mark from services which do not have diversified the execution layer clients.

### Additional context

I think I could make a Pull Request for this. **But I would like to know which solution is preferred first.**

### Would you like to work on this issue?

- [X] Yes
- [ ] No

corwintines:
Thanks @one-three-three-seven. I agree, this seems like an important distinction that is missing.

Before assigning and looking at next steps, id like to see what @wackerow thinks about this suggestion. 

Just to check, I see that you checked off being interested in working on this issue. Are you open to researching and updating the data for this?

one-three-three-seven:
Hello @corwintines,

Thanks for your reply. I have already started to make a very basic website about this topic: https://execution-diversity.info
But I think ethereum.org is a better location to rise awareness.

It's not possible to verify the information which the services are providing, but I still believe it's better to have an official statement from the services about their (execution layer) diversity.

We may need a threshold at which we consider a service as diversified. (For example: Do we show Lido as diversified with 90% Geth usage?) In my opinion, <2/3 Geth usage would be a good threshold.

So, we would show a service/pool with **<2/3 Geth usage** as **diversified** and the other ones (with **>2/3 Geth usage**) as **not diversified**. (Depending on their official statement.)

Buttaa:
@one-three-three-seven, I completely agree that the current labeling of "Diverse Clients" is misleading and should be expanded to include "EL client diversity" and "CL client diversity, imo.

Initially, I was hesitant to advocate for diversity in execution clients, but over the past few months, most EL clients have made significant progress and are now suitable for staking.

wackerow:
[...]

PR #9871, opened 2023-04-03
Add Ethpool SaaS

Buttaa:
Adds Ethpool to https://ethereum.org/en/staking/saas/


recy21 (approved):

Buttaa:
Thanks @wackerow , ptal https://github.com/ethereum/ethereum-org-website/issues/9882

src/data/staking-products.json
```diff
@@ -381,6 +381,36 @@
...
+      "minEth": 32,
+      "additionalStake": null,
+      "additionalStakeUnit": null,
+      "monthlyFee": null,
+      "monthlyFeeUnit": null,
+      "isFoss": false,
+      "hasBugBounty": false,
+      "isTrustless": false,
+      "isPermissionless": false,
+      "pctMajorityClient": 15,
```
wackerow, line 400: I saw your comment that this was for the execution side... this currently refers to consensus clients, but [this issue](https://github.com/ethereum/ethereum-org-website/issues/9840) is discussing changing that. 

This also shines light on maybe needing to adjust this definition a little more, since it's a little gray now. It was intended to highlight services that were running a majority of their validators using the most popular CL client. Right now Lighthouse and Prysm are pretty much neck-and-neck (which is incredible compared to a year ago!)...

Given the current use and definition, I would suggest we change this to 0% for now, since you aren't using any Prysm clients (which are still slightly in the lead)... neither of these clients have a supermajority, but they each have the ability to halt finality. I might suggest (separately) that we update this to be "more than 50% of validators are using a client that holds > 33% network share" (both for execution clients and consensus clients)
```suggestion
      "pctMajorityClient": 0,
```

wackerow (changes requested):
@Buttaa Left a couple comments. If we're cool with the suggestions, I think this is good to go!

wackerow (approved):

PR #9872, opened 2023-04-03
update "sharding" text/links across site

jmcook1186:
## Description

various places across the site either make outdated references to sharding or include links to the now deprecated sharding page. This PR aims to fix both, although probably not completely comprehensively.


src/intl/en/page-layer-2.json
jmcook1186, line 1: @corwintines do we need to manually replicate these changes in the json file for each language?

corwintines (approved):
Nice!

Issue #8797, opened 2022-11-30
Suggest a staking product or service


totoCZ:
### Project name

Squid

### Product type

Staking as a service

### Logo

![squid Logo - Black with Transparent Background](https://user-images.githubusercontent.com/5316250/204906808-79539a06-37fe-4fb7-be99-bbebb58049d9.svg)

### Description

Squid specializes in building and operating infrastructure for Ethereum users and also offers a hosted staking service.

### Website

https://www.ohsquid.com

### If software is involved, is everything open source?

No, code is closed source

### Is the project a fork? If yes, which project was forked?

No

### Is the product out of beta development?

Yes

### If a service, what are the fees associated with using the service?

12 USD per validator per month.

### What date did the project or service go live?

Nov 2022

### Has the project undergone an external security audit?

No

### Has the project undergone any security bug bounties?

No

### Is the project being actively maintained?

Yes

### Is the product or service free of trusted/human intermediaries?

No, the service administrator has access to staking keys.

### If listing a staking-as-a-service, are users required to sign-up for an account?

Yes, at a minimum sign up with email is required. No other information is needed.

### If listing as staking-as-a-service, who holds the signing keys, and withdrawal keys?

Signing keys: User & Service
Withdrawal keys: User

corwintines:
@wackerow could you review when you have a moment

wackerow:
Thanks for filing this @totoCZ! Appreciate the thoroughness with the form. 

From your post, this seems fitting for an SaaS listing with the following indicators:

OPEN SOURCE: No
AUDITED: No
BUG BOUNTY: No
BATTLE TESTED: Nov 2022 - <6mo, newly released (will soon switch to >6mos)
PERMISSIONLESS: No
DIVERSE CLIENTS: No?
SELF CUSTODY: No

If this sounds correct, would welcome a PR to add it to the `staking-products.json` file, or let me know if you need a hand and I can help get this added.

wackerow:
Noticing here that I had labeled "SELF CUSTODY: No" in my comment above, but per your post:

> If listing as staking-as-a-service, who holds the signing keys, and withdrawal keys?
> Signing keys: User & Service
> Withdrawal keys: User

...it looks like this would meet the current definition. I see you already labeled this `true` in the PR 👍, just wanted to make note.

totoCZ:
Excellent, thank you.

PR #9934, opened 2023-04-08
Add squid to staking services

totoCZ:
Here is PR as per https://github.com/ethereum/ethereum-org-website/issues/8797#issuecomment-1489408546


wackerow (approved):
@totoCZ Thanks for the PR!

This all looks good, just noticed some bugginess with the icon. Pushed a commit that updates the `/icons/staking/index.ts` to export the `SquidGlyphIcon.tsx` file for use, and also updated that file itself to get the icon working appropriately. Switched to a monochrome icon in the process as well for consistency and simplicity; let us know if you have issues with this approach.

image

This looks good on my end... Let us know if you're okay with this adjustment or not and we can get this in!

PR #9952, opened 2023-04-12
Redirects in browser

pettinarip:
## Description

This PR enables the `redirectInBrowser` flag in Gatsby to also redirect the user to the correct destination on the client side.

## Related Issue

#9951

### DIFF
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -4,7 +4,7 @@ import path from "path"
 import util from "util"
 import child_process from "child_process"
 import { createFilePath } from "gatsby-source-filesystem"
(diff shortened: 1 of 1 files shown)


pettinarip:
Abandoning this for now since it is now working as expected for some reason. The redirects on client side are not working.

Also, there is a suggestion from the Gatsby docs to try to avoid using this if possible

> Redirects are generally for redirecting legacy URLs to their new configuration on the server. If you can’t update your UI for some reason, set redirectInBrowser to true and Gatsby will handle redirecting in the client as well. You almost never need this so be sure your use case fits before enabling.

And since we did the UI changes on our side in #9951, we don't need to do this anymore.

PR #9959, opened 2023-04-13
fix(TableOfContents): export `Item` type-sig from index file

TylerAPfledderer:
## Description

This addresses type errors in template files where the `Item` type originally imported from the `TableOfContents` component was no longer recognized. This comes from the restructuring of the component into multiple files where the type now resides in a utils file.

This PR exports the type through the index file of the component to retain the imports in the templates.

## Related Issue
N/A


pettinarip (approved):
Much needed :)

PR #9962, opened 2023-04-13
refactor: update storybook config and theme util

TylerAPfledderer:
- Update Storybook config to do the following:
  - Set the `stories` option in `.storybook/main.js` to only point to where the story files are located and not up global prefixed titles for the sidebar. Rely only on declaring the structure via the component stories for now.
  - Disable autodocs so that docs should only be added on a component-by-component basis with custom markdown
  - Disable the backgrounds toggle in `.storybook/preview.js`. This is a color mode toggle for the background of the component preview. However, the Chakra UI addon has it's own toggle that is connected to the Chakra provider and needs to be prioritized.
  - Set the `layout` option in `.storybook/preview.js` to `centered` so each rendered preview is in the center of the window. Only override this as needed in the story files.

- In the theme config `component.utils` file, updates the type signature of `defineMergeStyles` to allow for instances of objects imported from the default theming without type errors regarding object mismatch.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -3,13 +3,7 @@ const { propNames } = require("@chakra-ui/react")
 const { babelConfig } = require("./babel-storybook-config")
 
 module.exports = {
-  stories: [
-    {
-      directory: "../src/components",
-      titlePrefix: "Components",
-      files: "**/*.stories.tsx",
-    },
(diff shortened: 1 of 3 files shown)


pettinarip (approved):
LGTM! 🚀

PR #9980, opened 2023-04-17
fix: broken PageHero buttons on the Layer2 page

pettinarip:
## Description

- Fixes id attr on page's headers
- Fixes hidden overflow on PageHero buttons

The last fix covers the visual issue when you hover or focus the buttons.
image
Now:
image

## Related issues

Fixes https://github.com/ethereum/ethereum-org-website/pull/9980


src/pages/layer-2.tsx
```diff
@@ -72,20 +73,20 @@ const SectionHeading = (props: HeadingProps) => {
...
-const ContentBox = (
-  props: HTMLAttributes<"div"> & {
-    children: ReactNode
-    isLightGrayBg?: boolean
-  }
-) => (
+const ContentBox = ({
+  isLightGrayBg,
+  ...rest
+}: BoxProps & {
```
nhsz, line 79:
```
{
  children: ReactNode
  isLightGrayBg?: boolean
}
```

should be defined as an interface

nhsz (changes requested):
Thanks, asking for a small change

src/pages/layer-2.tsx
```diff
@@ -72,20 +73,20 @@ const SectionHeading = (props: HeadingProps) => {
...
+  children: ReactNode
+  isLightGrayBg?: boolean
+}) => (
```
wackerow, line 82:
```suggestion
interface ContentBoxProps extends BoxProps {
  isLightGrayBg?: boolean
}
const ContentBox: React.FC<ContentBoxProps> = ({ isLightGrayBg, ...rest }) => (
```

src/pages/layer-2.tsx
pettinarip: Agreed

src/pages/layer-2.tsx
pettinarip: Thanks!

nhsz (approved):
LGTM!

corwintines (approved):

PR #9990, opened 2023-04-18
Update index.md

Yoohoo-cf:
update the benefits and limitations of Web3


minimalsm:
Hey :) Thanks for the PR. 

I don't think these changes are appropriate for this particular page, as they add vague (somewhat ideological) assertions that are hard to justify on developer documentation (I'm aware there is an existing issue here). Maybe check out [the web3 page](https://ethereum.org/en/web3) and see if there are changes that would be more appropriate there?

minimalsm:
Closing out as per above

PR #9992, opened 2023-04-18
Remove console easter egg

pettinarip:
## Description

Removes console easter egg.


nhsz (approved):
LGTM, thanks for the cleanup

PR #9998, opened 2023-04-19
fix(learn-page): typo bug

soheil555:
## Description

Update `theme.breakpoints.l` to `theme.breakpoints.lg` in the learn page.


corwintines:
Thanks for opening this @soheil555 

This would actually be a breaking change, as the breakpoint this is referring to is `l`. You can check that out here: https://github.com/ethereum/ethereum-org-website/blob/dev/src/theme.ts#L401

Unless you were looking to use a chakra breakpoint, but that would be a different change than this. Going to close this out for that reason.

PR #10000, opened 2023-04-19
Add matomo events

corwintines:
## Description

- Add matomo events throughout the site


minimalsm:
@corwintines is there a way to test if these are working as expected outside of prod 🤔

src/templates/use-cases.tsx
```diff
@@ -346,22 +346,47 @@ const UseCasePage = ({
       {
         text: t("template-usecase-dropdown-defi"),
         to: "/defi/",
+        matomo: {
+          eventCategory: "use cases menuy",
```
minimalsm, line 350: Typo?

```suggestion
          eventCategory: "use cases menu",
```

corwintines:
> @corwintines is there a way to test if these are working as expected outside of prod 🤔

Not easily, unfortunately. You can edit the matomo.ts file to work locally if you want to test locally, but don't think that there's an easy way to do this on preview deploys

src/templates/use-cases.tsx
```diff
@@ -324,22 +324,47 @@ const UseCasePage = ({
       {
         text: t("template-usecase-dropdown-defi"),
         to: "/defi/",
+        matomo: {
+          eventCategory: "use cases menu",
+          eventAction: "click",
+          eventName: "/defi/",
```
pettinarip, line 330: Just noting that this differs from the requested eventName in the Notion card. Those don't have the `/`s. Maybe is not a problem and we can leave it like is.

```suggestion
          eventName: "defi",
```

pettinarip: Same with the other cases in this file.

pettinarip (approved):
LGTM!

Wanted to mention that perhaps for other iterations we could create a new data structure for events + its type to avoid some repetition.

```tsx
type MatomoEvent = {
  eventCategory: string
  eventAction: string
  eventName: string
}

// now you can extend or use this type in other components, instead of repeating the same 3 props every time
interface Props {
  ...
  matomoEvent: MatomoEvent
} 
```

src/templates/use-cases.tsx
corwintines: Pushed up changes for this

PR #10012, opened 2023-04-20
♻️ Propose a 'clearer' python code for RLP encoding page

wenceslas-sanchez:
Re-wrote some part of the python code on the page RLP encoding for a better understandinhg.

## Description

What I propose is to remove useless `else`; IDK if removing them is againts the **explicit is better than implicit** doctrine; I will let you judge.

Also, some `for` and `if` condition where wrote in one line, which is not really visible when you are know python and even other languages.


jmcook1186 (approved):
Thanks, I think this is more clear. At the end of the day this snippet only aims to illustrate to readers what is happening in an RLP encoding - I think your version benefits from being more concise and clearly organized.

PR #10019, opened 2023-04-20
refactor(staking-template): migrate to chakra

soheil555:
## Description

Migrate `staking` template to Chakra UI

## Related Issue
#9353


pettinarip (approved):
Thanks @soheil555. Looks great!

I took the opportunity to add a few refactors in the code to better resize on different resolutions and reused a few components that we had already defined in a different template.

Issue #9997, opened 2023-04-18
Input layout bug on staking withdrawals page


wackerow:
### Describe the bug

On the staking withdrawals page, the widget to check withdrawal credentials has an input for "validator index" which is not displaying at the proper size.

### To reproduce

1. Go to https://ethereum.org/en/staking/withdrawals/#how-do-i-prepare
2. See "Validdator index" input size

### Expected behavior

Input element should match the height of the buttons with proper padding to prevent the placeholder text from being too close to the left border.

### Screenshots

<img width="666" alt="image" src="https://user-images.githubusercontent.com/54227730/232922232-e22410d2-db1e-4dea-bbaf-f37fae31ed8f.png">

### Desktop (please complete the following information)

macOS Ventura 13.2.1
Brave latest (112.0.5615.121)

### Would you like to work on this issue?

- [ ] Yes
- [X] No

MrJithil:
This issus not exist in the latest `dev` branch. Seems already fixed and will be available from next release.

wackerow:
@MrJithil Mind sharing what browser/OS you're using? I'm still seeing this on Brave and Safari for macOS using the `dev` branch.

MahendraBishnoi29:
@wackerow it's not fixed I've checked, assign it to me I'll take care of it

MrJithil:
> @MrJithil Mind sharing what browser/OS you're using? I'm still seeing this on Brave and Safari for macOS using the `dev` branch.

I was using Chrome.

nhsz:
Fixed by #10027

PR #10027, opened 2023-04-21
fix: Input layout bug on staking withdrawals page

MahendraBishnoi29:
## Related Issue
#9997

### DIFF
--- a/src/components/Staking/WithdrawalCredentials.tsx
+++ b/src/components/Staking/WithdrawalCredentials.tsx
@@ -140,6 +140,8 @@ const WithdrawalCredentials: FC<IProps> = () => {
       <Flex alignItems="center" gap={2} flexWrap="wrap">
         <Input
           size="lg"
+          padding=".5rem"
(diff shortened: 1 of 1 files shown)


src/components/Staking/WithdrawalCredentials.tsx
```diff
@@ -140,6 +140,8 @@ const WithdrawalCredentials: FC<IProps> = () => {
       <Flex alignItems="center" gap={2} flexWrap="wrap">
         <Input
           size="lg"
+          padding=".5rem"
+          borderRadius=".25rem"
```
nhsz, line 144:
```suggestion
          borderRadius="base"
```

nhsz (changes requested):
Thanks for the contribution. We have the `borderRadius` value defined on the theme, so we should use `base` instead of hardcoding it

nhsz (approved):
LGTM, thanks for fixing this!

PR #10045, opened 2023-04-24
✏️ fix typo in base fee example

wenceslas-sanchez:
Hi team,

I guess there is a typo in the example, following the table above it should be an increase of 12.5% and not 112.5%.

Have a nice day ! 👍🏻


nhsz (approved):
Thanks for fixing this, LGTM!

Issue #9924, opened 2023-04-07
Bug report


vishal590:
### Describe the bug

In Explore dapp, Art and Collectibles 
super rare website redirect is not working properly 

### To reproduce

1. Go to this link - https://ethereum.org/en/dapps/?category=collectibles
2. Choose category Art and Collectibles
3. See Art and Fashion
4. Click on Superrare website
5. redirects on wrong location

### Expected behavior

Need to change redirect link from - https://www.superrare.co
Need to change redirect link to - https://superrare.com

### Screenshots

![adfas](https://user-images.githubusercontent.com/58502188/230660048-4b2192ed-9156-4fbc-aaf2-22d62dbce4b4.PNG)

### Desktop (please complete the following information)

OS [WINDOWS 10]
Browser : Chrome

### Would you like to work on this issue?

- [X] Yes
- [ ] No

Issue #9869, opened 2023-04-03
Diversify Links with Kernel Additions


andytudhope:
### Is your feature request related to a problem? Please describe.

After talking through some ideas with @minimalsm I'd like to add links to Kernel and some of the educational resources we have gathered in order to diversify what is already offered on e.org.

### Describe the solution you'd like

This will entail me going through the site page by page, thinking about user journeys, thinking about which specific Kernel links work well at which specific times, and adding them.

While doing this, I will also update any copy that I can and fix any bugs that come up.

### Describe alternatives you've considered

I do not plan to add copious amounts of new information. If anything, I will edit and remove whatever I can to make people's experience cleaner, clearer and less overwhelming.

### Would you like to work on this issue?

- [X] Yes
- [ ] No

minimalsm:
Merged both PRs @andytudhope, sorry for the delay. 

I'll leave this issue for now, feel free to close whenever you'd call this done.

andytudhope:
Ah, sorry friendly bot. I am still planning on working on this: it's just that travels and hackathons have taken priority for the time being. Still a long way to go.

wackerow:
Hey @andytudhope! Circling back... looks like this has run quiet over the past year. I'm going to go ahead and close this out, but primarily for organizational/clean-up purposes. If you'd like to continue these efforts please do us a favor and open a new issue and we can go from there, thanks!

PR #10048, opened 2023-04-24
Guides Pages Updates

andytudhope:
## Description

This PR continues work on updating copy across the e.org website. In this case, the focus is specifically the `Guides` section under the `Learn` tab. I have edited the copy, added just one link to Kernel, and generally tried to keep cutting away for clarity and brevity, while not sacrificing accuracy.

## Related Issue

#9869 

## Notes

It seems like a previous fix I made has been reverted on `dev`, so this corrects #9924 again.

### DIFF
--- a/src/content/guides/how-to-use-a-bridge/index.md
+++ b/src/content/guides/how-to-use-a-bridge/index.md
@@ -6,7 +6,7 @@ lang: en
 
 # How to bridge tokens to layer 2
 
-If everyone always uses Ethereum, it can get slow and expensive. One solution to this is to create new "layers": i.e. different networks which operate in similar ways to Ethereum itself. These so-called Layer 2s help reduce congestion and cost on Ethereum by processing many more transactions at lower fees, and only storing the result of these on Ethereum every so often. As such, these layers 2s enable us to transact with increased speed and decreased costs. Many popular crypto projects are moving to layer 2s because of these benefits. The simplest way to move tokens from Ethereum to layer 2 is to use a bridge.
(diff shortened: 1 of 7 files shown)


src/content/guides/how-to-use-a-bridge/index.md
```diff
@@ -6,7 +6,7 @@ lang: en
 
 # How to bridge tokens to layer 2
 
-Layer 2s are scaling solutions for Ethereum which allow users to transact with increased speed and decreased costs. Many popular crypto projects are moving to layer 2s because of these benefits. The simplest way to move tokens from Ethereum to layer 2 is to use a bridge project.
+If everyone always uses Ethereum, it can get slow and expensive. One solution to this is to create new "layers": i.e. different networks which operate in similar ways to Ethereum itself. These so-called Layer 2s help reduce congestion and cost on Ethereum by processing many more transactions at lower fees, and only storing the result of these on Ethereum every so often. As such, these layers 2s enable us to transact with increased speed and decreased costs. Many popular crypto projects are moving to layer 2s because of these benefits. The simplest way to move tokens from Ethereum to layer 2 is to use a bridge.
```
jmcook1186, line 9: Just a few nitpicks on the opening line - "if everyone always uses Ethereum" is pretty vague - not *everyone* has to use Ethereum for gas fees to get high. Not sure what "always" is meant to convey - it doesn't matter how often users are transacting as long as there are enough transactions competing for block space to push up the cost of inclusion. And I guess "slow and expensive" is a bit of a weird one, because it's kind of one or the other - you can either accept a sufficiently high cost to ensure inclusion (expensive), or you can wait until gas drops to your cost threshold (slow).

```suggestion
If there is a lot of traffic on Ethereum, it can become expensive. One solution to this is to create new "layers": i.e. different networks which operate in similar ways to Ethereum itself. These so-called Layer 2s help reduce congestion and cost on Ethereum by processing many more transactions at lower fees, and only storing the result of these on Ethereum every so often. As such, these layers 2s enable us to transact with increased speed and decreased costs. Many popular crypto projects are moving to layer 2s because of these benefits. The simplest way to move tokens from Ethereum to layer 2 is to use a bridge.
```

jmcook1186 (changes requested):
another really helpful set of updates - thanks. I just left one comment with some nits in the layer-2 section for you to review.

jmcook1186 (approved):
thanks!

PR #10053, opened 2023-04-24
Chinese Simplified April update

corwintines:
## Description
- Import Crowdin content buckets for Chinese Simplified from April


nhsz (approved):
LGTM

PR #10071, opened 2023-04-25
remove old grant pages and redirects for them

corwintines:
## Description
- Remove redirects and pages for `medalla-data-challenge` and `staking-community-grants` as they are outdated, not referenced anywhere on the site, and doesn't need to be kept around


nhsz:
Is it ok to remove these pages @pettinarip?

pettinarip (approved):
I'm ok with this. I don't see too much value of having these pages since they are from 2020.

nhsz:
> I'm ok with this. I don't see too much value of having these pages since they are from 2020.

I remember there was a discussion about keeping them or it was on esp site only maybe

pettinarip:
> > I'm ok with this. I don't see too much value of having these pages since they are from 2020.
> 
> I remember there was a discussion about keeping them or it was on esp site only maybe

Correct. That was on ESP side. That same thing allow us to stop creating these grant pages on e.org side, we can just reference them to the ESP website.

nhsz (approved):

PR #10075, opened 2023-04-26
refactor(Nav): use the Chakra `Menu` component

### LINKED ISSUE #10750: Header - Menu and Dropdowns
Implement the [new DS styles](https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?node-id=248%3A507) in the [Menu](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Nav/Menu.tsx) component.

image

Note: consider creating a new Menu component if it makes things easier.

TylerAPfledderer:
## Description

Updates the `Menu` component in `Nav` by implementing Chakra's `Menu` component. 💪🏼 

Required removing the `hasSubNav` prop for `NavDropdown`, which was not being used; the Chakra `Menu` component does not have the ability for nesting itself.

Fixes #10750


TylerAPfledderer:
@pettinarip minor formatting changes bled into the diff on the branch update.

TylerAPfledderer:
@pettinarip @nloureiro noting that I am updating this to make use of the new DS. Then go from here to the entire header with a storybook. Doing this would keep in scope and prevent closing this PR to start a new one unnecessarily.

src/components/Nav/Dropdown.tsx
```diff
@@ -1,203 +1,151 @@
...
+      bg: "none",
+    }}
+  >
+    <BaseLink
+      isPartiallyActive={false}
+      color="inherit"
+      textDecor="none"
+      _hover={{ textDecor: "none" }}
+      {...props}
+    />
```
pettinarip, line 121: Having `BaseLink` as a child is creating invalid html since `MenuItem` is rendering a button.

Do you think we can do `<MenuLink as={BaseLink}>` here?

pettinarip (commented):
Nice @TylerAPfledderer!

I've linked this PR with #10750 (didn't realize before that you were implementing the DS in this PR xD).

The only thing missing here is the active link styles:
image

Wondering if you see that doable in this PR.

nloureiro:
@TylerAPfledderer  Thank you for this.

I found two bugs for the design side, sorry... but when this hit the browser, I had to rethink these two details. I've changed the Figma.

Screen Shot 2023-09-08 03 35 18 PM


I used the dev docs to make this screen recording.

https://github.com/ethereum/ethereum-org-website/assets/1120748/e7432019-556a-480d-92e8-848e5d5bde1d


Again, sorry for the bugs from my side...

src/components/Nav/Dropdown.tsx
TylerAPfledderer: Ah, you mean `MenuItem`?

I must have overlooked that issue. `MenuItem` should be fine here to consume `BaseLink` via `as`. Will check on that asap!

pettinarip: Oh yes haha, `MenuItem` it is xD

TylerAPfledderer:
@nloureiro @pettinarip I have pushed the changes for padding, spacing, and colors. Also, the `MenuItem` now consumes `BaseLink`.

At the moment, I see two issues:
1. When we mean `active` styles here, do we mean "on click" or styles for the link that matches the current page?
2. There is an issue in keyboard interaction. For this component in Chakra, keyboard focus can only occur with arrow keys. Using the `Tab` key will have no effect. However, `BaseLink` is preventing the arrow keys from applying change of focus. I will need to explore further the approach here, and whether or not this is considered a bug.

pettinarip:
@TylerAPfledderer 

1. The latter. So, in theory the Link component should handle this already, since the Gatsby Link provides this functionality.  [Here](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Link.tsx#L61) you can see the expected prop. You shouldn't be doing anything else I believe.

2. Ok. Yes, I think I had the same issue before. I ended up doing [this](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/ButtonDropdown.tsx#L81) in the ButtonDropdown migration. Not sure if that is the best approach and not sure if there is some side-effect that I didn't noticed. You can see this dropdown being used in the [nft page](https://ethereum.org/en/nft/), on the sidenav on the left.

TylerAPfledderer:
> @TylerAPfledderer
> 
>     1. The latter. So, in theory the Link component should handle this already, since the Gatsby Link provides this functionality.  [Here](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Link.tsx#L61) you can see the expected prop. You shouldn't be doing anything else I believe.
> 
>     2. Ok. Yes, I think I had the same issue before. I ended up doing [this](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/ButtonDropdown.tsx#L81) in the ButtonDropdown migration. Not sure if that is the best approach and not sure if there is some side-effect that I didn't noticed. You can see this dropdown being used in the [nft page](https://ethereum.org/en/nft/), on the sidenav on the left.

1. Will review, because something is not working as intended then.
2. That is unfortunately breaking intended Screen Reader UX. ☹️ The `MenuItem` is given `role='menuitem'` (and `MenuList` is given `role='menu'`). So when you focus on the menu button, the screen reader (NVDA) refers to it as a "submenu". And then when you enter the list the SR might say "X button 1 of X". If `MenuItem` consumed a link component the reader might say "X link 1 of X". But if you wrap `MenuItem` in the link, the SR will not longer read out the number of items in the list, but just read out the element it is focused on.

I will be looking to spend some more time exploring both items.

pettinarip:
@TylerAPfledderer Oh interesting. Ok, thanks for looking into that 👍🏼

TylerAPfledderer:
@pettinarip now that I had a chance to review:

1. `_activeLink` should be preferred (targets the `aria-current="page"` attribute that is rendered by `BaseLink`) to have theme-aware props. The `activeStyle` prop could be used, but you couldn't use the semantic tokens (not theme-aware).
2. I'm still having difficulty understanding why keyboard interaction completely fails with `BaseLink` :(

fgm5858:
https://github.com/ethereum/ethereum-org-website/assets/146592193/475f6b6d-d547-48f1-b508-766bd746202d

fgm5858:
Robaron mi cueta de gmail

wackerow:
Hey @TylerAPfledderer, deeply appreciate all the work you put in on this. Unfortunate we couldn't get this merged in sooner, but with the recent epic to redesign the nav menu and implementation pending in #12125, I'm gonna go ahead and close this out. We can build the stories out and iterate from there
🙏

PR #10078, opened 2023-04-26
Writing cohort banner

corwintines:
## Description
- Adds a DismissableBanner to promote the ethereum.org writing cohort

### DIFF
--- a/src/components/Banners/Implementations/WritersCohortBanner.tsx
+++ b/src/components/Banners/Implementations/WritersCohortBanner.tsx
@@ -1,0 +1,21 @@
+import React from "react"
+import { Text } from "@chakra-ui/react"
+
+import DismissableBanner from "../DismissableBanner"
+
+const WritersCohortBanner = () => {
+  if (
+    typeof window !== `undefined` &&
+    window.location.pathname.includes("contributing")
+  ) {
+    return (
+      <DismissableBanner storageKey="writersCohort">
+        <Text m={0}>Hello world</Text>
+      </DismissableBanner>
+    )
+  }
+
+  return null
+}
+
+export default WritersCohortBanner
--- a/src/components/Banners/Implementations/WritersCohortBanner.tsx
+++ b/src/components/Banners/Implementations/WritersCohortBanner.tsx
@@ -4,10 +4,14 @@ import { Text } from "@chakra-ui/react"
 import DismissableBanner from "../DismissableBanner"
 import Link from "../../Link"
 
-const WritersCohortBanner = () => {
+interface IProps {
+  pathname: string
+}
+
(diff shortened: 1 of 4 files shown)


src/components/Banners/Implementations/WritersCohortBanner.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react"
+import { Text } from "@chakra-ui/react"
+
+import DismissableBanner from "../DismissableBanner"
+import Link from "../../Link"
+
+const WritersCohortBanner = () => {
+  if (
+    typeof window !== `undefined` &&
+    window.location.pathname.includes("contributing")
```
pettinarip, line 10: For SSR purposes, a better way of getting the `pathname` is from Gatsby itself (or specifically from reach router which is what they use underneath).

```tsx
// every page or template receives a `location` prop
const StaticPage = ({ data, location }) => {
  ...
}

// then you can pass that prop down to the components
<WritersCohortBanner pathname={location.pathname} />
```

This way you abstract yourself from the `window` object and you can render the same thing server and client side.

pettinarip (approved):
Looks good! I left a comment for improvement.

src/components/Banners/Implementations/WritersCohortBanner.tsx
```diff
@@ -0,0 +1,28 @@
...
+  ) {
+    return (
+      <DismissableBanner storageKey="writersCohort">
+        <Text m={0}>
+          Join the first-ever ethereum.org Writers Cohort, in collaboration with
+          Taptive, starting May 11th.{" "}
```
minimalsm, line 16: Can probably remove Taptive from here for simplicity. 

```suggestion
          Join the first-ever ethereum.org Writers Cohort, starting May 11th.{" "}
```

src/components/Banners/Implementations/WritersCohortBanner.tsx
corwintines: Makes sense, changed this PR over to this just now.

src/components/Banners/Implementations/WritersCohortBanner.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+interface IProps {
+  pathname: string
+}
+
+const WritersCohortBanner = ({ pathname }) => {
+  if (
+    typeof window !== `undefined` &&
+    (pathname.includes("contributing") || pathname.includes("community"))
+  ) {
```
pettinarip, line 15: My idea was to completely remove the `window` dep in order to render it in both places. Or does it break?

```suggestion
  if (pathname.includes("contributing") || pathname.includes("community")) {
```

pettinarip (approved):

PR #10080, opened 2023-04-26
Update community-meetups.json

sleonp-dev:
<Update Community MeetUps, Ecuador Community>

#Adding the information of the ETH Ecuador Community. 

#Adding information to community meetup, Eventbrite page!

### DIFF
--- a/src/data/community-meetups.json
+++ b/src/data/community-meetups.json
@@ -1,5 +1,11 @@
 [
   {
+    "title": "Ethereum Tricolor",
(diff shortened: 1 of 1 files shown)


nhsz:
Hi @sleonp-dev, thanks for the contribution! Seems these are more specific events for Blockchain Week Ecuador rather than a recurring meetup? If that's the case, the proper place for these kind of events is the [`community-events.json`](https://github.com/ethereum/ethereum-org-website/blob/dev/src/data/community-events.json) file

PR #10093, opened 2023-04-28
Fix image sizes

pettinarip:
## Description

Recover original images and resize them again to match what the queries are requesting.


f3theH:
error en el sistema por compañía regalan 10 000 000 00 NFT

PR #10100, opened 2023-04-30
Feat: Create New Tag Theme

TylerAPfledderer:
## Description

_This is a part of implementing [V1 of the New Design System](https://github.com/ethereum/ethereum-org-website/issues/9548)_

This PR modifies the existing `Tag` component and updates it's theme to match the new DS.

Also update the filter tags in the [Developer Tutorials list page](https://ethereum.org/developers/tutorials) to ensure they do not receive the new styling in prod for the moment.


src/@chakra-ui/gatsby-plugin/components/Tag/index.ts
```diff
@@ -0,0 +1,110 @@
...
+const baseStyleLabel = defineStyle({
+  ...tagTheme.baseStyle?.label,
+  fontSize: "xs",
+  textTransform: "uppercase",
+  textAlign: "center",
+})
+
+const baseStyleCloseButton = defineStyle({
+  ...tagTheme.baseStyle?.closeButton,
+  opacity: 1,
```
pettinarip, line 42: I see an extra margin on the close button that we don't have on the icons that makes it look weird.

```suggestion
  opacity: 1,
  m: 0,
```

src/@chakra-ui/gatsby-plugin/components/Tag/index.ts
```diff
@@ -0,0 +1,110 @@
...
+      outlineOffset: 0,
+    },
+  },
+})
+
+const baseStyleLabel = defineStyle({
+  ...tagTheme.baseStyle?.label,
+  fontSize: "xs",
+  textTransform: "uppercase",
+  textAlign: "center",
```
pettinarip, line 37: To better align it vertically.

```suggestion
  textAlign: "center",
  lineHeight: 1,
```

src/@chakra-ui/gatsby-plugin/components/Tag/index.ts
```diff
@@ -0,0 +1,110 @@
...
+  return {
+    container: {
+      ...defaultStyles?.container,
+      // Remove default dark mode styles
+      _dark: {},
+      ...statusStyles,
+    },
+    label: {
+      // ...warningLabelStyles
+    },
```
pettinarip, line 64: I think we also need to style the `closeButton`s pseudo classes with the same styles that the container has. I'm talking specifically about the focus state.

Right now we are displaying the default Chakra styles here. I think we should use the same border styles and colors that we use for the container.

This is how it looks now when you focus on the close button:
image

Not saying that this is the code but could be something like:
```tsx
closeButton: {
  _focusVisible: {
    ...statusStyles["&:any-link"]._focusWithin,
    bg: "transparent",
    boxShadow: "none",
  },
},
```

src/@chakra-ui/gatsby-plugin/components/Tag/utils.ts
```diff
@@ -0,0 +1,320 @@
...
+
+// Because the color scheme from the default theme
+// Goes through the `Badge` config, used the variables
+// created from that config.
+export const $badgeBg = cssVar("badge-bg")
+export const $badgeColor = cssVar("badge-color")
+
+export const $tagBoxshadowColor = cssVar("tag-boxshadow-color")
+
+// TODO: Can this get consolidated?
```
pettinarip, line 11: Good question. I don't have any better idea right now.

However, keeping this in its own file definitely helps, good job on doing that.

src/pages/developers/tutorials.tsx
```diff
@@ -155,11 +154,39 @@ const TagContainer = styled.div`
...
+    <chakra.button
+      ref={ref}
+      bg="none"
+      bgImage="radial-gradient(46.28% 66.31% at 66.95% 58.35%,rgba(127, 127, 213, 0.2) 0%,rgba(134, 168, 231, 0.2) 50%,rgba(145, 234, 228, 0.2) 100%)"
+      border="1px"
+      borderColor={props.isActive ? "primary300" : "white800"}
+      borderRadius="base"
+      boxShadow={!props.isActive ? "table" : undefined}
+      color="text"
+      fontSize="sm"
```
pettinarip, line 175: These buttons are too tall comparing them to the originals.

```suggestion
      fontSize="sm"
      lineHeight="1.2"
```

src/pages/developers/tutorials.tsx
```diff
@@ -155,11 +154,39 @@ const TagContainer = styled.div`
...
+      opacity={props.isActive ? 1 : 0.7}
+      p={2}
+      textTransform="uppercase"
+      _hover={{
+        color: "primary",
+        borderColor: "text200",
+        opacity: "1",
+      }}
```
pettinarip, line 183: We need to spread the props since the `onClick` is not working rn

```suggestion
      }}
      {...rest}
```

pettinarip (changes requested):
Nicely done @TylerAPfledderer. Left a few comments but overall this looks great.

src/@chakra-ui/gatsby-plugin/components/Tag/index.ts
TylerAPfledderer: Right!

src/@chakra-ui/gatsby-plugin/components/Tag/index.ts
TylerAPfledderer: @pettinarip This example you gave will actually work here.

I think what I am missing here is whether or not this button will actually be visible if the tag itself is a link. That might change how I approach this to ensure the focus ring for the button has enough contrast to the background of the tag (looking at the `normal` tag)

@nloureiro I don't believe we received that clarity in the DS yet. 😄 

Will the close button ever be used if the tag is a link?

src/@chakra-ui/gatsby-plugin/components/Tag/utils.ts
TylerAPfledderer: I can leave that comment in there just in case a lightbulb goes off later! 😅

src/@chakra-ui/gatsby-plugin/components/Tag/index.ts
TylerAPfledderer: @nloureiro  Oh! I understand.

I'm going to roll with what you suggested above to provide you a proper review.

I think it will be best to have three stories here:

1. Show all tag style variants as links with text and no icons to simply demonstrate the state changes (maybe they should include the + icon as the right-side element?).
2. Provide another set of the same tags, but not as a links and containing the close button to demonstrate the buttons styling in the context of the different style variants
3. The different variants of elements that are expected to be rendered in the component.

Would this be better?

nloureiro: hum... I think it might work, as far as I understand.

correct me if I'm wrong:
we will have a folder in storybook with "tags" > inside there will be 3 stories

is that it?

pettinarip: Correct @nloureiro here you will see another entry
image

@TylerAPfledderer yes, make sense 👍🏼

TylerAPfledderer: Sweet! I'll get that added asap

TylerAPfledderer:
@nloureiro @pettinarip update pushed with a third story added and found a cheap solution to get hover/focus styling for the close button.

src/@chakra-ui/gatsby-plugin/components/Tag/index.ts
```diff
@@ -0,0 +1,123 @@
...
+const baseStyleCloseButton = defineStyle({
+  ...tagTheme.baseStyle?.closeButton,
+  opacity: 1,
+  m: 0,
+  // Clear default
+  _focusVisible: null,
+  "&:focus-visible, &:hover": {
+    outline: "3px solid",
+    outlineOffset: "-2px",
+  },
```
TylerAPfledderer, line 50: Cheap solution to get focus/hover styling.
1. override default styling by nulling out (because I'm using a multi-selector instead)
2. Provide a more generous styling of an outline ring (this is taking advantage of the default `outlineColor` value of `currentColor`)

src/@chakra-ui/gatsby-plugin/components/Tag/index.ts
nloureiro: look good to me, great job!!

src/@chakra-ui/gatsby-plugin/components/Tag/index.ts
pettinarip: Nice! looks great now

pettinarip (commented):
I don't see this as a blocker but wanted to raise the issue to see if it has a simple fix. If not, we can tweak that on a separate PR.

The multi-line tag is not growing in height as in the design:
image

DS:
image

TylerAPfledderer:
@pettinarip I think a larger line height than `1` needs to be  established and have that factored into the overall height for a single-line tag

nloureiro:
Is there any URL where I can check this?

nloureiro:
> Is there any URL where I can check this?

got it, thank you @pettinarip 

@TylerAPfledderer one detail I missed in the Design system is the hover for the close button. with the border only is weird.

so I'm adding this behavior to the Figma but my idea is to go full background like this screenshot

what to you think?


https://github.com/ethereum/ethereum-org-website/assets/1120748/0474f8e4-8127-46a0-8983-b2239d259ab4

TylerAPfledderer:
@nloureiro this can be done, but will add a little complexity to our [large object of styles](https://github.com/ethereum/ethereum-org-website/blob/db32d0958709e7cd99e1e515f5cfd72f43eb8af4/src/%40chakra-ui/gatsby-plugin/components/Tag/utils.ts).

(I still need to figure out how to refactor this, but there are one or two styles that do not line up with the rest to create some iterator and taking advantage of color token names.)

Also, I was unaware that there were modifications to the color palette! 😅 (The `bodyLight` colors changed and the addition of `bodyMedium`)

nloureiro:
> 

Sorry about the color change. I thought I mentioned it somewhere... but I should have made it more clear 😅

I'm not sure I follow, but what I got it that we will do this change on separate PR. Is that it?

If yes, I'm ok with it.


My main concern is this behavior on hover, and it's something too strong. If you have some other suggestion, I'm open to it
Screen Shot 2023-05-19 09 31 37 AM

nloureiro (approved):

pettinarip (approved):
Ok, I think we are in a good state to merge this, great job again @TylerAPfledderer and @nloureiro for the review.

A few things I've taken note of and that we are going to fix in different PRs:
- Close button styles: Tyler is looking into how we can simplify the styling object structure to include these changes with ease #10265
- Fix multi-line tag height #10266

PR #10105, opened 2023-05-01
Preoptimize source images under `/assets`

pettinarip:
## Description
This PR applies the changes reverted in this PR #10094. The original PR is #9374 where we were resizing the images on the assets folder to optimize our build process.

In addition, this PR includes a few fixes for some images that were throwing warnings during the build process.


wackerow:
Images that are noticeably sharper on desktop _production_:
image
image

Side-by-side:
image

Although the below image appears sharper on _this PR_:
image

Somewhat subjective on how sharp we want these images to be... Personally really appreciate having the higher quality, but recognize there are trade-offs with image size.

nhsz:
> Images that are noticeably sharper on desktop _production_: image image
> 
> Side-by-side: image
> 
> Although the below image appears sharper on _this PR_: image
> 
> Somewhat subjective on how sharp we want these images to be... Personally really appreciate having the higher quality, but recognize there are trade-offs with image size.

the thing is if sharpness diffs are easily noticeable we're losing quality in the process, which shouldn't happen (this also depends on screen resolution)

wackerow:
Yeah, for reference, I was testing using a 4K screen. Was this compression method intended to be lossless? They do not appear to be if that is the case.

nhsz:
> Yeah, for reference, I was testing using a 4K screen. Was this compression method intended to be lossless? They do not appear to be if that is the case.

it's not compression but resizing of the images

nloureiro:
My review on this is that I could see minimal degradation on some images, but they neglectable the size benefits of loading times for pages.

here some examples
Screen Shot 2023-05-03 10 20 00 AM

Screen Shot 2023-05-03 10 17 34 AM

but in the end, the size of each image make a huge difference
Screen Shot 2023-05-03 10 26 05 AM


for me it's good to go!!!

pettinarip:
@wackerow @nhsz @nloureiro 

I resized the 800px & 1200px images again from the originals but this time with `quality=100` to try to lose the less quality as possible.

The idea of this preoptimization is to reduce the size of the source images before we perform the build. We are not trying to reduce the size of the final images served to the clients. This is an improvement to alleviate the load of the build process, not for the client load performance.

=======

FYI on the build process, Gatsby takes the source image and produces subsequent images in multiple sizes to perform and adapt well on different devices.

For example:

1. Input: source image (4000px) 
2. Gatsby build
3. Output: 3 images
    - 1 for lg res at 1200px
    - 1 for md res at 600px
    - 1 for sm res at 400px

What we want to improve by reducing the size is the **source image**. Instead of loading and processing a 4k image, we want to process the minimum possible resolution.

pettinarip:
As discussed, I'm reverting the optimizations for the big handmade drawing images to avoid any quality loss. These are:

```
- eth.png
- what-is-ethereum.png
- wallet.png
- doge-computer.png
- developers-eth-blocks.png
- enterprise-eth.png
- infrastructure_transparent.png
- impact_transparent.png
- wallet-cropped.png
- finance_transparent.png
- hackathon_transparent.png
- future_transparent.png

- upgrades/upgrade_eth.png
- upgrades/upgrade_doge.png
- upgrades/newrings.png
- upgrades/core.png
- upgrades/merge.png
- upgrades/oldship.png
- upgrades/upgrade_robot.png
- upgrades/upgrade_rhino.png

- home/hero.png
```

These images are going to be optimized in a separate PR.

src/assets/eth-gif-chalk.png
corwintines, line 1: Should we remove this change?

corwintines, line 1: Should this be removed since the file size is larger?

src/assets/wallets/blockwallet.png
corwintines, line 1: Should this be removed since the file size is larger?

corwintines (changes requested):
I imagine these should be the original image. Im noticing in the script that there isn't a condition to skip saving the optimized image if its file size is larger. Is this something that should be checked for?

src/scripts/optimizeImages.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    if (ext === "png") {
+      await resized.png({ quality: QUALITY }).toFile(optimizedName)
+    }
+
+    if (ext === "jpg" || ext === "jpeg") {
+      await resized.jpeg({ quality: QUALITY }).toFile(optimizedName)
+    }
+
+    return rename(optimizedName, match)
```
corwintines, line 41: I think adding this to the optimizeImages.ts function it would avoid the above

```
const optimizedInfo = await sharp(optimizedName).metadata()
  if (optimizedInfo.size < info.size) {
    await rename(optimizedName, match)
  }
  return
```

src/assets/layer-2/hero.png
wackerow, line 1: Similar here, should this be removed? although 704 Bytes is pretty insignificant

minimalsm:
@pettinarip is there anything blocking this now that we've removed the larger hand-drawn illustrations?

This looks ok to me...

corwintines (approved):
Bringing in. We can address if we need this script in the future after next migration.

src/assets/layer-2/hero.png
Inatiye0911905991: 👍

PR #10123, opened 2023-05-04
Update `#maxRedeem` description in `index.md`

misicnenad:
Update `#maxRedeem` description in `index.md` from "...that can be redeem from.." to "...that can be redeem**ed** from...".

## Description

The tense of the word "redeem" in the description should be in present perfect - "...that can be redeem from.." should be "...that can be redeem**ed** from...".


nhsz (approved):
LGTM, thanks!

PR #10125, opened 2023-05-04
Fix broken use case hero images in non-English languages

minimalsm:
## Description

Some non-English pages were using the English path for their frontmatter image, causing the image to 404 and not show.

E.g. using `../../assets/use-cases/dao-2.png` (which only works for English files) instead of `../../../../assets/use-cases/dao-2.png` (which is required for all non-English languages)

This PR fixes these for use case pages.

### DIFF
--- a/src/content/translations/de/decentralized-identity/index.md
+++ b/src/content/translations/de/decentralized-identity/index.md
@@ -5,7 +5,7 @@ lang: de
 template: use-cases
 emoji: ":id:"
 sidebarDepth: 2
-image: ../../assets/eth-gif-cat.png
+image: ../../../../assets/eth-gif-cat.png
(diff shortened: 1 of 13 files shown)


nhsz:
Maybe we could use absolute paths for assets to avoid this issues on translations? e.g.: `src/assets/use-cases/dao-2.png` should work in any case, cc @corwintines @pettinarip @wackerow

corwintines:
> Maybe we could use absolute paths for assets to avoid this issues on translations? e.g.: something like `src/assets/use-cases/dao-2.png` should work in any case, cc @corwintines @pettinarip @wackerow

That would be ideal, but unfortunately, I think that is easier said than done with Gatsby. I guess another point to the idea of moving to NextJS at some point.

corwintines (approved):

PR #10137, opened 2023-05-05
Remove markdown syntax from within React components on History page (en)

minimalsm:
## Description

Replaces the markdown syntax within React components on the History page with HTML. 

Limited to testing this one page (here and on Crowdin) before doing others.

### DIFF
--- a/src/content/history/index.md
+++ b/src/content/history/index.md
@@ -11,11 +11,12 @@ A timeline of all the major milestones, forks, and updates to the Ethereum block
 
 <ExpandableCard title="What are forks?" contentPreview="Changes to the rules of the Ethereum protocol which often include planned technical upgrades.">
 
-Forks are when major technical upgrades or changes need to be made to the network – they typically stem from [Ethereum Improvement Proposals (EIPs)](/eips/) and change the "rules" of the protocol.
+Forks are when major technical upgrades or changes need to be made to the network – they typically stem from <a href="/eips/">Ethereum Improvement Proposals (EIPs)</a> and change the "rules" of the protocol.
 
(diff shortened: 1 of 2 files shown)


minimalsm:
@nhsz yes, I believe it makes no difference what we use for as long as it's surrounded by angle brackets.

Would we prefer that?

src/content/history/index.md
```diff
@@ -36,11 +37,13 @@ The Shanghai upgrade brought staking withdrawals to the execution layer. In tand
 
 <ExpandableCard title="Shanghai EIPs" contentPreview="Official improvements included in this upgrade.">
 
-- [EIP-3651](https://eips.ethereum.org/EIPS/eip-3651) – _Starts the `COINBASE` address warm_
-- [EIP-3855](https://eips.ethereum.org/EIPS/eip-3855) – _New `PUSH0` instruction_
-- [EIP-3860](https://eips.ethereum.org/EIPS/eip-3860) – _Limit and meter initcode_
-- [EIP-4895](https://eips.ethereum.org/EIPS/eip-4895) – _Beacon chain push withdrawals as operations_
-- [EIP-6049](https://eips.ethereum.org/EIPS/eip-6049) - _Deprecate `SELFDESTRUCT`_
+<ul>
+<li><a href="https://eips.ethereum.org/EIPS/eip-3651">EIP-3651</a> – <em>Starts the <code>COINBASE</code> address warm</em></li>
```
minimalsm, line 41: Mmm our linter doesn't run on markdown html

nhsz:
> @nhsz yes, I believe it makes no difference what we use for as long as it's surrounded by angle brackets.
> 
> Would we prefer that?

IMO we should pick one for consistency but also Chakra components have built in a11y that helps so I'd lean to that if it works

nhsz:
> @nhsz yes, I believe it makes no difference what we use for as long as it's surrounded by angle brackets.
> 
> Would we prefer that?

IMO we should pick one for consistency and avoid mixing tags, but also Chakra components have built in a11y that helps so I'd lean to that if it works

nhsz:
> > @nhsz yes, I believe it makes no difference what we use for as long as it's surrounded by angle brackets.
> > Would we prefer that?
> 
> IMO we should pick one for consistency but also Chakra components have built in a11y that helps so I'd lean to that if it works

Didn't know we're already converting html tags into Chakra components, so ok to go with html 👍

minimalsm:
Closing in favour of #10357

FYI, experimenting with lining in this PR but it breaks readability on Crowdin

PR #10140, opened 2023-05-05
changed 0.5 ETH initial penalty to 1 ETH

nixorokish:
After Phase 0, initial penalty was raised to its full value of 1/32 of the slashed validator's effective balance, a maximum of 1 ETH, in the pre-Merge Bellatrix upgrade.


jmcook1186 (approved):
yep this looks good - for reference the change in spec is documented here: https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md#updated-penalty-values

PR #10142, opened 2023-05-06
Added ApeWorX SDK

brucey0x:
Apeworx is a Pythonic SDK for the EVM and the spiritual successor of Brownie. I noticed it wasn't added.

Added section for Apeworx and reference to its logo in `src/pages/developers/local-environment.tsx`. Uploaded its logo in  `src/assets/dev-tools/apeworx.png`.

Wasn't sure if there were any other pages that needed adjusting for this addition, so please pay careful attention to that when reviewing.

Thanks!


minimalsm (changes requested):
Thanks for the PR @brucey0x ⚡️And welcome to the community! 🎉

To get the description and alt to display properly, they need to be added to `src/intl/en/page-developers-local-environment.json`. I've left an example in the comments :)

Could we also reduce the size of the image? It only needs to be 100px in height, and this is 1000x1000 right now 😀

Also, can

corwintines:
Going to mark as abandoned and close this out.

PR #10170, opened 2023-05-10
Change Ethpool "permissionless"

Buttaa:
As previously discussed with @wackerow, email login is only necessary for urgent notifications. Users who don't mind can use throwaway emails.

### DIFF
--- a/src/content/developers/docs/nodes-and-clients/node-architecture/index.md
+++ b/src/content/developers/docs/nodes-and-clients/node-architecture/index.md
@@ -45,7 +45,7 @@ Node operators can add a validator to their consensus clients by depositing 32 E
 
 | Execution Client                                   | Consensus Client                                                 | Validator                    |
 | -------------------------------------------------- | ---------------------------------------------------------------- | ---------------------------- |
-| Gossips transactions over its p2p network          | Gossips blocks and attestations overs its p2p network            | Proposes blocks              |
+| Gossips transactions over its p2p network          | Gossips blocks and attestations over its p2p network             | Proposes blocks              |
 | Executes/re-executes transactions                  | Runs the fork choice algorithm                                   | Accrues rewards/penalties    |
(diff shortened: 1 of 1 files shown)


wackerow:
@Buttaa Just clarifying here, is there a way for the user to join the service without providing this? I understand users can use a throw away, but I wanted to clarify my Discord comments that "permissionless" in this context I see as no requirement to create an account with the service to stake. In this case it sounds like the email is only used for notifications, but appears to be a requirement before they're able to start the service at all.

Can you help clarify if there is any ability for Ethpool to prevent a user from starting or accessing their account? This is ultimately what matters here imo... whether or not a service can censor a user from using the service.

Buttaa:
> 

@wackerow, gotcha. Sorry for the miscommunication there!

Our users do need an email address to login and do the deposit (or upload the validator keystore). 
We could avoid this by adding Metamask login but it would not make it "permissionless" because the frontend and the service itself is still centralized.

We could add MM login on top, but it would lead to a potential loss of funds/worse UX. Let me explain:

- Our users can upload validator keystores to our service with 0x0 credentials. Thus, they have no ethereum address (0x01) address attached to them. Without an email address, it is a pain for us to verify the user in case the user wants to leave the service (e.g, switching pool)

- Some SaaS allows hosting Rocketpool nodes, where time-critical actions are required, otherwise, the funds may become stuck (= dissolved mini pools). Asking the user to login via MM + and email to notify them about the status, sounds worse UX to me.

Applying "permissionless" to SaaS is questionable because the service itself is still centralized if email login or not. 
It only makes sense to apply this to protocols (like Rocketpool) or LSD's where fractional staking is possible. SaaS are not permissionless by definition

Buttaa:
Closed by mistake. @wackerow could you please reopen?

corwintines:
@Buttaa im not able to reopen this, but there doesn't seem to be any changes in the files when I check the tab out, so maybe this is ok now. You'll likely need to open this up again if there are any changes, but feel free to open a new PR for that.

PR #10120, opened 2023-05-04
Update energy consumption page

jmcook1186:
## Description

Responding to issue #10109 .
Adds data from Cambridge source, removes estimate with less robust citation, and adds some detail to the text that clarifies uncertainty in energy consumption estimates. 

## Related Issue

#10109

PR #10172, opened 2023-05-10
Fix table overflow caused by VisuallyHidden width on external Links

minimalsm:
## Description

Noticed in #10120: the energy consumption has a horizontal scroll under ~680px caused by the table ([also visible in prod](https://ethereum.org/en/energy-consumption/)).

- The `VisuallyHidden` component within the `Link` component is causing the issue.
- Looks like the 1px width of the `VisuallyHidden` component is causing the overflow scroll not to work correctly, causing a horizontal scroll. 
- When I delete it the `VisuallyHidden` component or add `width: 0px` to it, it does not overflow

Suspect this might be a bug in Chakra itself 🤔?

### A11Y

I tested the a11y of this with the width set to 0px, and it works as expected using VoiceOver/JAWs screen readers. However, I isolated this change to links within markdown tables incase there are unintended consequences I'm not aware of...

cc: @TylerAPfledderer @pettinarip 

## Related Issue

#10120

Fixes: https://github.com/ethereum/ethereum-org-website/issues/10174

### DIFF
--- a/src/components/MarkdownTable.tsx
+++ b/src/components/MarkdownTable.tsx
@@ -14,6 +14,9 @@ const MarkdownTable: React.FC<IProps> = ({ children }) => (
           borderColor: "border",
           whiteSpace: "nowrap",
         },
+        ".chakra-link .visually-hidden": {
+          width: "0px",
+        },
       }}
     >
       {children}
--- a/src/components/MarkdownTable.tsx
+++ b/src/components/MarkdownTable.tsx
@@ -6,16 +6,13 @@ export interface IProps {
 }
 
 const MarkdownTable: React.FC<IProps> = ({ children }) => (
-  <Box my={8} overflowX="auto">
+  <Box position="relative" my={8} overflowX="auto">
     <Table
       sx={{
         th: {
           borderBottom: "1px solid",
           borderColor: "border",
           whiteSpace: "nowrap",
-        },
-        ".chakra-link .visually-hidden": {
-          width: "0px",
         },
       }}
     >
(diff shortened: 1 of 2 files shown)


nhsz:
mh, @TylerAPfledderer you think this could be a Chakra bug? 🤔

src/components/MarkdownTable.tsx
```diff
@@ -14,6 +14,9 @@ const MarkdownTable: React.FC<IProps> = ({ children }) => (
           borderColor: "border",
           whiteSpace: "nowrap",
         },
+        ".chakra-link .visually-hidden": {
+          width: "0px",
+        },
```
pettinarip, line 19: I think this issue is related to the absolute positioning of the hidden div and the `overflowX` of the table. For some reason, the absolute positioned div is overflowing the container.

In short, I think we can add a `position: relative` on the table's container and it will fix it without the need of adding a new class or style on the Link itself.

TylerAPfledderer:
Interesting! This might only be happening in Chrome, as there is no overflow in Firefox.

Yes to @pettinarip. `position: relative` to the table container works best here. 

For context: the styling currently applied to `VisuallyHidden` is a common approach.
- See the [A11y Project](https://www.a11yproject.com/posts/how-to-hide-content/)
- See [WebAim](https://webaim.org/techniques/css/invisiblecontent/) (They suggest not to use a `0` size value)

src/components/MarkdownTable.tsx
minimalsm: Fixed :)

nhsz (approved):
LGTM

PR #10180, opened 2023-05-11
Fix typo in zero knowledge proofs page

bskrksyp9:
Fixed grammatical errors


bskrksyp9:
Hi @minimalsm on which chain is this poap available? m not sure how can I add the network of it in metamask to view my poap?

PR #10181, opened 2023-05-11
fix description of merkle trie depth calc

jmcook1186:
## Description

the calculation of the generalized index in ssz encoding was reported as 

"
Each element in the serialized list has a generalized index equal to 2* *depth + idx where idx is its zero-indexed position in the serialized object and the depth is the number of levels in the Merkle tree, which can be determined as the **square root of the number of elements (leaves)**.
"

whereas it should actually be:

"
Each element in the serialized list has a generalized index equal to 2* *depth + idx where idx is its zero-indexed position in the serialized object and the depth is the number of levels in the Merkle tree, which can be determined as the **base-two logarithm of the number of elements (leaves)**".)
"

updated here.

## Related Issue

No issue, error pointed out by @gballet


gballet:
credit where it's due: thanks to @jsign for pointing out this discrepancy

nhsz (approved):
LGTM, thanks!

PR #10187, opened 2023-05-11
fix: inappropriate "mev" examples

sftchance:
Currently, a page on [ethereum.org](https://ethereum.org/en/developers/docs/mev/) defines MEV with an example of [Sandwhich Attack](https://ethereum.org/en/developers/docs/mev/#mev-examples-sandwich-trading). 

However, the included mentions and examples were nothing more than a state of having "potential profitability". By this logic, buying and selling any ERC20 should be on this list however, of course, that is not an example of genuine "MEV".

## Description

The original author of this section provided examples that are defined based on "running a transaction faster" and "finding money others didn't" rather than any level of miner inclusion or nuance and it is resulting in a significant amount of people believing that a sandwich attack is "MEV".

The term has been misappropriated to mean atomic instances of profit, which is not what MEV is. This PR removes the inclusion of the naive MEV examples to prevent further misdefinition and misunderstanding of what MEV is. The fact that Toxic Flow is not even included in this section is a visual display that things were fundamentally misunderstood at the time of writing.

There is potentially a discussion to be had about correcting the "definition" of MEV to reflect once again "miner extractable value" as it only recently took on the definition of "maximal" due to the lacking context and knowledge of those proposing and using the term.

The entire MEV page is currently baked with an immense set of inaccuracies not addressed in this PR.

> Is worth acknowledging that this misdefinition and misrepresentation does not only take place here but also in places like Chainlink marketing articles. This has become an ecosystem-wide misdefinition and misrepresentation, and addressing it here first would be wise.

### DIFF
--- a/src/content/developers/docs/mev/index.md
+++ b/src/content/developers/docs/mev/index.md
@@ -36,56 +36,6 @@ Rather than programming complex algorithms to detect profitable MEV opportunitie
 
 Flashbots is an independent project which extends execution clients with a service that allows searchers to submit MEV transactions to validators without revealing them to the public mempool. This prevents transactions from being frontrun by generalized frontrunners.
 
-## MEV examples {#mev-examples}
-
-MEV emerges on the blockchain in a few ways.
-
-### DEX arbitrage {#mev-examples-dex-arbitrage}
-
-[Decentralized exchange](/glossary/#dex) (DEX) arbitrage is the simplest and most well-known MEV opportunity. As a result, it is also the most competitive.
-
-It works like this: if two DEXes are offering a token at two different prices, someone can buy the token on the lower-priced DEX and sell it on the higher-priced DEX in a single, atomic transaction. Thanks to the mechanics of the blockchain, this is true, riskless arbitrage.
-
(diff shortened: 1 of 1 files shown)


corwintines:
Thanks @sftchance 

> By this logic, buying and selling any ERC20 should be on this list however, of course, that is not an example of genuine "MEV".

This is a form of MEV in the form of front running often. 

---

I do think what is on this page is correct and an accurate description of MEV. While some information could be added to this page, such as toxic flow in your suggestion, I don't think removing examples of MEV is the solution here. 

There are many instances of discussion around these types of MEV, including from flashbots themselves. 

https://medium.com/flashbots/frontrunning-the-mev-crisis-40629a613752
https://dune.com/chorus_one/ethereum-mev-data
https://docs.flashbots.net/new-to-mev (contains many references)

---
> There is potentially a discussion to be had about correcting the "definition" of MEV to reflect once again "miner extractable value" as it only recently took on the definition of "maximal" due to the lacking context and knowledge of those proposing and using the term.

Open to hearing reasons for this, but this change was made and often reflected across the industry after the transition from proof-of-work to proof-of-stake when miners became validators. 

---

Im going to close this out as it is, because removing these examples I don't believe to be the right move. But, if you would like to add more content on this topic I encourage you to open an issue or PR around that as it sounds like you have issues around the whole page. Open to how we can work on this together, let me know :)

PR #10188, opened 2023-05-11
refactor(MarkdownTable): apply `TableContainer` and header padding

TylerAPfledderer:
## Description

Refactors the `MarkdownTable` component by:
1. Replacing the `Box` component with `TableContainer` for improved alignment with Chakra features
    - Maintains x-axis overflow while hiding vertical overflow and ensuring 100% max width to a parent container (not currently an issue)
    - Applies `whiteSpace: nowrap` to prevent wrapping when overflow exists at the smaller screens.
2. Add padding to heading content when overflow exists at the smaller screens.

Before:
image

After:
image

## Related Issue
N/A


wackerow (changes requested):
Hey @TylerAPfledderer!

Just looked at this on desktop, and I'm not sure this is entirely what we want:

https://ethereumorgwebsitedev01-tylerapfleddererethereumo73284.gatsbyjs.io/en/bridges/

https://github.com/ethereum/ethereum-org-website/assets/54227730/aa029575-95da-440a-a4ee-8a60a3e7426a

Tables with longer copy should really be wrapped. 

On mobile this example doesn't show the column header at all, since it's center aligned and the container is much wider than the device.

image


Currently in production we get:
image
...which is obviously not great either, but at least makes it obvious there is a table there and makes all the info quickly visible.

TylerAPfledderer:
@wackerow thank you for taking a look!

Whoops, I didn't consider tables like this that would serve as multi-column "feature" lists. 😅 

Hmmm, that's tough. These tables can get pretty difficult to read on mobile (my example especially). This is good to know for the new DS!

In fact, I shouldn't have opened this due to PR #8139 where @pettinarip was intended to completely remove the component. I will close this PR for that reason. (The other PR was closed in favor of #9932 for the moment)

PR #10209, opened 2023-05-13
Enforce TS type check on builds

pettinarip:
## Description

This PR does two things:
- Forces type checking on build time
- Fixes current type errors


src/utils/mergeObjects.ts
pettinarip, line 1: Not used. We have `lodash` installed anyway.

src/components/Codeblock.tsx
```diff
@@ -231,30 +231,11 @@ const codeTheme = {
   },
 }
 
-const getValidChildrenForCodeblock = (child) => {
```
pettinarip, line 234: Not needed since in all of our use cases, we receive a `string` as the child. Forced this on the `IProps` definition.

nhsz:
> ## ❌ ethereum-org-website-dev deploy preview failed
> Your build failed. [View the build logs](https://gatsbyjs.com/dashboard/5e02f963-3fe5-4e37-a1e7-ac05f7753a86/sites/dc7c4340-a8e9-44ed-b7be-883c70d2f02d/builds/29e23a59-6c57-4baa-8e3a-5aeadf143dc0/details).
> 
> **Errors**
> 
> Your plugins must export known APIs from their gatsby-node.
> 
> See https://www.gatsbyjs.com/docs/reference/config-files/gatsby-node/ for the list of Gatsby node APIs.
> 
> - The plugin gatsby-plugin-mdx@3.20.0 is using the API "unstable_shouldOnCreateNode" which is not a known API.
> 
> Some of the following may help fix the error(s):
> 
> - Rename "unstable_shouldOnCreateNode" -> "shouldOnCreateNode"
> Your plugins must export known APIs from their gatsby-node.
> 
> See https://www.gatsbyjs.com/docs/reference/config-files/gatsby-node/ for the list of Gatsby node APIs.
> 
> - The plugin gatsby-plugin-mdx@3.20.0 is using the API "unstable_shouldOnCreateNode" which is not a known API.
> 
> Some of the following may help fix the error(s):
> 
> - Rename "unstable_shouldOnCreateNode" -> "shouldOnCreateNode"

to enforce the type checking we should fix the errors we currently have on the logs or builds will start to fail (and previews)

pettinarip:
> to enforce the type checking we should fix the errors we currently have on the logs or builds will start to fail (and previews)

@nhsz those errors are not related to TS, they are Gatsby related. The thing that broke the build was a bug on the `Codeblock` component (should be fixed now).

wackerow (approved):
LGTM 🎉

package.json
```diff
@@ -119,7 +119,8 @@
   },
   "scripts": {
     "postinstall": "yarn theme",
-    "build": "gatsby build",
+    "build": "yarn type-check && yarn build:gatsby",
```
corwintines, line 122: I was thinking about your comment earlier about the issue with this where it would break running the first time until someone has built the project. What if we used a conditional in that build command

```
"build": "if [ -f path/to/file.js ]; then yarn type-check && yarn build:gatsby; else yarn build:gatsby && yarn type-check; fi"
```

This way, we could check for the `gatsby-types.d.ts` file, if it exists run type-check first and then gatsby:build. If the types file isn't there, then we would build first then check types. 

I think this may address your concerns that we were talking about earlier.

wackerow: Is this something we're still looking to implement @pettinarip?

pettinarip: @wackerow @corwintines I'd say that given that we are in the transition to move to NextJS we can discard this change for now and enforce this again when the migration is done.

Will close this PR. Let me know if you disagree.

pettinarip:
Closing. Check [this message](https://github.com/ethereum/ethereum-org-website/pull/10209#discussion_r1340186648) for details.

PR #10211, opened 2023-05-14
update title of "how to "register" an Ethereum account"

umededoteth:
## Description

I think "create" is a more appropriate word than "register", even if "register" is in quotation marks.

If we're putting "register" in quotation marks, this means that we don't really mean "register", so why use it at all?

Just my thoughts!

### DIFF
--- a/src/content/guides/how-to-register-an-ethereum-account/index.md
+++ b/src/content/guides/how-to-register-an-ethereum-account/index.md
@@ -1,5 +1,5 @@
 ---
-title: How to "register" an Ethereum account
+title: How to create an Ethereum account
 description: A step-by-step guide on Ethereum account creation using a wallet.
 lang: en
 ---
--- a/src/content/guides/how-to-register-an-ethereum-account/index.md
+++ b/src/content/guides/how-to-register-an-ethereum-account/index.md
@@ -1,5 +1,5 @@
 ---
-title: How to create an Ethereum account
+title: How to "create" an Ethereum account
 description: A step-by-step guide on Ethereum account creation using a wallet.
 lang: en
 ---


corwintines:
@konopkja wondering what your take on this is. Its a small change so I could go either way, but seeing as you authored this will give you final say.

jmcook1186:
strongly in favour of changing it to "create"

konopkja:
just to give some context to understand what needs to be considered: 

the word register has been specifically used here as it is a phrase our users used in our feedback widget. The attempt here is to move closer towards our visitors and their language.  In quotation marks its of course because register isnt factually correct, however that is a common struggle on our website - often times to be factually correct or nuanced we use language and phrases our users do not use or have harder times understanding.

since the How to guides are focused on absolute newcomers to the space who might have limited technical skills, it is extremely important to use as similar language to their own as possible. 

All that being said I think create could be close enough to the language so we can change it, experiment and see how it influences number of visits to the page.

egypjoy:
So are you saying having a explorer account having a validator work station
on the Ethereum network are you saying having millions on my platform and
much Ethereum crypto in my Ethereum wallet that the Ethereum.org guide help
me to obtain that still I at this moment with millions staked and in pools
I still don't have a Ethereum account.

src/content/guides/how-to-register-an-ethereum-account/index.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: How to "register" an Ethereum account
+title: How to create an Ethereum account
```
corwintines, line 2:
```suggestion
title: How to "create" an Ethereum account
```

Just updating based on discussion

corwintines (approved):
Thanks @umededoteth. Going to approve and bring it in based on the discussion.

src/content/guides/how-to-register-an-ethereum-account/index.md
wackerow: @corwintines Looks like this was pulled in, but I might still push back on quoting this... the only reason we quoted "register" was because it was not technically accurate... I think "How to create an Ethereum account" is technically accurate enough to not warrant any quoting, and could add confusion by leaving it in.

PR #10219, opened 2023-05-15
Spike: refactor ExpandableCard to work with md translations

pettinarip:
**DO NOT MERGE**

----

Currently, the custom attrs that are set in a react component, do not show up in Crowdin for translations.

For example,
```tsx
// https://github.com/ethereum/ethereum-org-website/blob/dev/src/content/history/index.md
<ExpandableCard title="title" contentPreview="preview content">
  ...content
</ExpandableCard>
```
The attr `contentPreview` can't be translated in Crowdin.

## Description

Adds a refactor to the `ExpandableCard` to accept the preview content as a child component.

```tsx
// new proposal: removes the `contentPreview` attr
<ExpandableCard>
  <ExpandableCardPreview title="title">
    preview content
  </ExpandableCardPreview>

  <ExpandableCardContent>
    content
  </ExpandableCardContent>
</ExpandableCard>
```

You can see these changes on:
- A react page (/bug-bounty): https://ethereumorgwebsitedev01-expandablecardintl.gatsbyjs.io/en/bug-bounty/
- A md page (/history): https://ethereumorgwebsitedev01-expandablecardintl.gatsbyjs.io/en/history/

**Note:** if this test is accepted, the new structure for `ExpandableCard` will be applied to the rest of the codebase.

### DIFF
--- a/src/content/history/index.md
+++ b/src/content/history/index.md
@@ -9,13 +9,17 @@ sidebarDepth: 1
 
 A timeline of all the major milestones, forks, and updates to the Ethereum blockchain.
 
-<ExpandableCard title="What are forks?" contentPreview="Changes to the rules of the Ethereum protocol which often include planned technical upgrades.">
-
-Forks are when major technical upgrades or changes need to be made to the network – they typically stem from [Ethereum Improvement Proposals (EIPs)](/eips/) and change the "rules" of the protocol.
+<ExpandableCard>
+  <ExpandableCardPreview title="What are forks?">
+    Changes to the rules of the Ethereum protocol which often include planned technical upgrades.
+  </ExpandableCardPreview>
+  <ExpandableCardContent>
+  Forks are when major technical upgrades or changes need to be made to the network – they typically stem from [Ethereum Improvement Proposals (EIPs)](/eips/) and change the "rules" of the protocol.
 
 When upgrades are needed in traditional, centrally-controlled software, the company will just publish a new version for the end-user. Blockchains work differently because there is no central ownership. [Ethereum clients](/developers/docs/nodes-and-clients/) must update their software to implement the new fork rules. Plus block creators (miners in a proof-of-work world, validators in a proof-of-stake world) and nodes must create blocks and validate against the new rules. [More on consensus mechanisms](/developers/docs/consensus-mechanisms/)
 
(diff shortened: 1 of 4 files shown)


src/content/history/index.md
```diff
@@ -9,13 +9,17 @@ sidebarDepth: 1
...
 
-<ExpandableCard title="What are forks?" contentPreview="Changes to the rules of the Ethereum protocol which often include planned technical upgrades.">
-
-Forks are when major technical upgrades or changes need to be made to the network – they typically stem from [Ethereum Improvement Proposals (EIPs)](/eips/) and change the "rules" of the protocol.
+<ExpandableCard>
+  <ExpandableCardPreview title="What are forks?">
+    Changes to the rules of the Ethereum protocol which often include planned technical upgrades.
+  </ExpandableCardPreview>
+  <ExpandableCardContent>
+  Forks are when major technical upgrades or changes need to be made to the network – they typically stem from [Ethereum Improvement Proposals (EIPs)](/eips/) and change the "rules" of the protocol.
```
minimalsm, line 17: This works as expected in Crowdin 👍 nice workaround @pettinarip.

Screenshot 2023-05-17 at 16 22 02

One question, might we want to do the same with the title? This makes the context clearer in Crowdin, but mostly because it feels weird mixing the two methods of declaring the card's content.

```suggestion
<ExpandableCard>
<ExpandableCardTitle>
    What are forks?
  </ExpandableCardTitle>
  <ExpandableCardPreview>
    Changes to the rules of the Ethereum protocol which often include planned technical upgrades.
  </ExpandableCardPreview>
  <ExpandableCardContent>
  Forks are when major technical upgrades or changes need to be made to the network – they typically stem from [Ethereum Improvement Proposals (EIPs)](/eips/) and change the "rules" of the protocol.
```

pettinarip: Oh right, we can do that, yea. I agree that it will be cleaner, good point.

Just for curiosity, does the `title` attr content works with Crowdin? because I don't see it in your example. The text `"What are forks?"` does not appear on the screen that you shared 🤔

corwintines: I think what @minimalsm is proposing is create another component `ExpandableCardTitle` and pass the text in there instead of passing a prop to `ExpandableCard` or `ExpandableCardPreview`. Not sure if that clears that up, or if there was confusion before 😅

wackerow:
This looks great... Bummer we need to add this much syntax for these cards, especially with how often we use them, but I think it will help clear up many issues.

corwintines:
@pettinarip was wondering what the status of this PR or plans for it.

PR #10232, opened 2023-05-17
Update index.md

fushengcheng:
The original description of quote and getAmountOut was wrong. The difference between the two is not whether the transaction fee is considered or not. quote function considers the price to remain constant, while getAmountOut function considers the liquidity to remain constant, the scenarios and mathematical formulas based on which they are used are different.


corwintines:
I think the original here makes more sense. Going to close this out. @fushengcheng if you have come documentation to share around this to update please share :)

PR #10234, opened 2023-05-17
feat(storybook): add storybook-react-i18next addon

TylerAPfledderer:
## Description

Adds the [storybook-react-i18next addon](https://storybook.js.org/addons/storybook-react-i18next/) to the project storybook.

- Merges the namespace resources for each language under the default "translation" ns to easily access the ids without declaring a specific namespace
- Sets a default set of languages to test, plus the supported languages declared in the user's local `.env`
- Converts the storybook `main.js` file to TypeScript
- Converts the Storybook 'preview.js` file to storybook and updates the object per version 7 migration.

## Related Issue
Discussion #9879


TylerAPfledderer:
@pettinarip at the time of this comment, regarding the file conflicts:
- `.storybook/main.js`: the new ts file needs the `staticDirs: ["../static"]` option
- `.storybook/preview.js`: the new ts file needs `import "../static/fonts/inter-font-face.css"`

.storybook/i18next.ts
```diff
@@ -0,0 +1,42 @@
...
+import { supportedLanguages } from "../src/utils/languages"
+
+export const baseLocales = {
+  en: { title: "English", left: "En" },
+  zh: { title: "中国人", left: "Zh" },
+  ru: { title: "Русский", left: "Ru" },
+  uk: { title: "українська", left: "Uk" },
+}
+
+// Only i18 files named in this array are being exposed to Storybook. Add filenames as necessary.
```
pettinarip, line 12:
```suggestion
// Only i18n files named in this array are being exposed to Storybook. Add filenames as necessary.
```

pettinarip (commented):
Great addition @TylerAPfledderer

Would you mind if I add some translations to the existing stories in this PR?

TylerAPfledderer:
> Would you mind if I add some translations to the existing stories in this PR?

@pettinarip made the suggested changes. All good with me! 😄

pettinarip (approved):
FYI @TylerAPfledderer I ended up removing the "buildLangs" since it was adding too much languages to the SB which I don't think we really need right now. For testing purposes, the `baseLocales` are just enough and makes the build ligther. Later, if we consider that we need more langs we can add them.

TylerAPfledderer:
> FYI @TylerAPfledderer I ended up removing the "buildLangs" since it was adding too much languages to the SB which I don't think we really need right now. For testing purposes, the `baseLocales` are just enough and makes the build ligther. Later, if we consider that we need more langs we can add them.

@pettinarip I agree. All good with me! 😄

PR #10255, opened 2023-05-18
Add additional resources about account abstraction

0xolishell:
hey there -- adding a few recent articles about various aspects of account abstraction. I work at Alchemy and we've gotten lots of views + tons of positive feedback about these articles. Thought this would be of help to EF readers too as web3 in general delves deeper in account abstraction Lmk of any questions, thanks!


corwintines (approved):
Thanks @0xolishell 

Cleaned this up to just be part one. That way readers can hop in, and continue their journey there.

PR #10256, opened 2023-05-18
Rename Urdu language name

pettinarip:
## Description

- Renames lang codes from `ur-in` and `ur-pk` to `ur`
- Renames `"Urdu (India)"` and `"Urdu (Pakistan)"` names to `"Urdu"`

cc @lukassim


wackerow:
Should we add a redirect for the existing `/ur-in/` lang path to 301 to `/ur/`?

corwintines:
@wackerow I don't think we would need to do that. This was added pretty recently, I think it's pretty unlikely that this redirect would be used. But, open to pushback here as its pretty quick to add.

corwintines (approved):

Issue #10258, opened 2023-05-18
isLangRightToLeft function not using language config setting


corwintines:
### Describe the bug

Currently the `isLangRightToLeft` is implemented as follows

```
export const isLangRightToLeft = (lang: Lang): boolean => {
  return lang === "ar" || lang === "fa"
}
```
This problem here is the languages for rtl are hardcoded instead of looking at the languages config from `i18n/config.json`. Languages like Urdu are being rendered ltr instead of rtl because of this.

### To reproduce

1. Go to homepage in Urdu (lang code ur)
2. See page being rendered ltr

### Expected behavior

isLangRightToLeft should use the languages config settings instead of being hardcoded.

### Would you like to work on this issue?

- [X] Yes
- [ ] No

PR #10259, opened 2023-05-19
isLangRightToLeft function not using language config setting

corwintines:
## Description
- Fixes the `isLangRightToLeft` function to use a languages config for lang direction instead of being hardcoded values for langauges.

## Related Issue
Fixes https://github.com/ethereum/ethereum-org-website/issues/10258

### DIFF
--- a/src/utils/translations.ts
+++ b/src/utils/translations.ts
@@ -1,10 +1,15 @@
 import type { Lang } from "./languages"
 import type { Direction } from "../types"
 
+import i18nConfigs from "../../i18n/config.json"
+
 export type TranslationKey = string
 
 export const isLangRightToLeft = (lang: Lang): boolean => {
-  return lang === "ar" || lang === "fa"
+  const langConfig = i18nConfigs.filter(
+    (language) => language.hrefLang === lang
+  )
+  return langConfig[0].langDir === "rtl"
 }
 
 export const getDirection = (lang?: Lang): Direction => {
--- a/src/utils/translations.ts
+++ b/src/utils/translations.ts
@@ -7,7 +7,7 @@ export type TranslationKey = string
 
 export const isLangRightToLeft = (lang: Lang): boolean => {
   const langConfig = i18nConfigs.filter(
-    (language) => language.hrefLang === lang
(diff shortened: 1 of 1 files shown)


src/utils/translations.ts
```diff
@@ -1,10 +1,15 @@
...
 import type { Direction } from "../types"
 
+import i18nConfigs from "../../i18n/config.json"
+
 export type TranslationKey = string
 
 export const isLangRightToLeft = (lang: Lang): boolean => {
-  return lang === "ar" || lang === "fa"
+  const langConfig = i18nConfigs.filter(
+    (language) => language.hrefLang === lang
```
wackerow, line 10: I think we want to use `language.code` here, not `hrefLang`

ie: Filipino we use the internal code `ph`, but the formal language code is `fil`... in `/utils/languages.ts` we use `ph`, not `fil` (which is where we declare the `Lang` type and it's allowed options)

```suggestion
    (language) => language.code === lang
```

src/utils/translations.ts
```diff
@@ -1,10 +1,15 @@
...
+  )
```
wackerow, line 11: We may also want to guard (or throw an error) against the case where our filtered array is empty (no matches)

Maybe:

```ts
if (!langConfig.length) throw new Error("Language code not found")
```

or

```ts
if (!langConfig.length) return "rtl"
```

wackerow (changes requested):
@corwintines Pretty sure our build error is related to using the `hrefLang` instead of the `code` property... in the case of Filipino we'd be looking for `fil` but this doesn't exist and would result in the array being empty, and [0] wouldn't work.

src/utils/translations.ts
corwintines: Ah, this makes sense. Great catch!

src/utils/translations.ts
corwintines: I think ill do the first one with throwing an error. Think this would make it easier to debug as this would be an issue. Thanks for the suggestion

wackerow (approved):

PR #10261, opened 2023-05-19
Add docsearch:description meta tag

wackerow:
## Description
- Adds a `meta` tag for `docsearch:description` to gain access to the current metadata description for each page via Algolia DocSearch
- About: https://docsearch.algolia.com/docs/required-configuration/#introduce-global-information-as-meta-tags

### DIFF
--- a/src/components/PageMetadata.tsx
+++ b/src/components/PageMetadata.tsx
@@ -200,6 +200,10 @@ const PageMetadata: React.FC<IProps> = ({
           property: `og:site_name`,
           content: `ethereum.org`,
         },
+        {
+          name: `docsearch:description`,
+          content: desc,
+        },
       ].concat(meta)}
     >
       <script type="application/ld+json">


corwintines (approved):

wackerow:
This is in attempt to bring the new DocSearch index crawler config closer to what we currently have on the site, which utilizes the metadata description.
Currently:
```ts
    "lvl2": {
      "selector": "//meta[@name='description']/@content",
      "type": "xpath",
      "global": true
    },
```
...but the new crawler does not support that selector syntax.

pettinarip:
Nice @wackerow!

I was wondering if we should also add the `pathname` as a top level selector. For example, when you search "Get ETH" (which is the name of the page in the nav dropdown), that name does not appear in the `h1` or `title` of that page, ignoring it from the top search results you get.

Just an idea. A different workaround could be to do something like what we are doing in the Stablecoins page where we set the same name on the `h1` as in the nav's dropdown name.

PR #10267, opened 2023-05-19
Add chromatic script

pettinarip:
## Description

As suggested by [the Chromatic team](https://www.chromatic.com/docs/github-actions#forked-repositories), we are adding the `chromatic` script to work with forked repos.

The intent of this is to fix an issue with [some UI Reviews](https://www.chromatic.com/review?appId=63b7ea99632763723c7f4d6b&number=10131&type=linked) that were coming from forked repos as empty.


nhsz (approved):
LGTM

PR #10276, opened 2023-05-19
Add 3xpl to the list of block explorers

Har01d:
3xpl supports ERC-20, ERC-721, and ERC-1155 token transfers. It's fast, its core is open-source, and most importantly there are no ads.


wackerow (approved):
Hey @Har01d! Sorry for slowness here, but thank you for posting... 

Given we already have a few block explorers listed, just want to be careful not to overload users. Typically we're trying to look for things that offer something new that the others may not...

That being said, I had never heard of this explorer personally, so I checked it out, and think it could work. I like the option to be able to download data, the open source nature of it, and the lack of ads. Personally I see these qualities as something it offers that are missing at least in part by the others. 

Has my approval, but I'll let @corwintines take another look before merging in.

Har01d:
Thank you! So far we've been able to add a lot of L2s as well: Arbitrum One, Base, Gnosis Chain, Optimism, Polygon, and Polygon zkEVM. And we're coming with a good Beacon Chain explorer too.

3xpl also works with various naming systems, and the search is universal, so you can do this: https://3xpl.com/search?q=vitalik.eth (and it shows only addresses which have had some transactions on the relevant chains).

Har01d:
Hey @wackerow, @corwintines, any update on this? :)

Har01d:
So far 3xpl already fully supports Beacon Chain and features 7 different L2s. The cool things is that everything is in the same data format.

image

minimalsm (commented):
Not blocking, but I think in general is we should be raising the bar on the products we list here and not overloading users as @wackerow said. If this provides some value in ways that other products do not, fair enough.

One thing to note with 3XPL (not saying this isn't the case elsewhere), is that there is no filtering of spam tokens. 

Screenshot 2024-02-13 at 18 55 47

minimalsm:
Also, @Har01d noticed that we already list another block explorer, Blockchair, that you are also working on. Not saying there is anything particularly malicious at play here, but it seems a little unusual?

corwintines:
Thanks @Har01d 

I don't think this really adds anything that isn't covered by block explorers already on the list. This list is already getting a bit long and probably needs to be trimmed down as it is. Going to close this out for now, but if you disagree feel free to reopen and we can discuss again.

Har01d:
@minimalsm, @corwintines, thank you for your review, and let me respond.

> If this provides some value in ways that other products do not, fair enough.

It does. Unlike most explorers listed on ethereum.org (https://ethereum.org/ru/developers/docs/data-and-analytics/block-explorers) it
- does not have any ads
- has an .onion version (which is really the only way how everyone should use a block explorer)
- provides free data services for academia (name another explorer which allows you to download its database)
- has many L2s and Beacon Chain built-in into the same interface (and we plan to expand!)
- provides a way to run arbitrary SQL queries against its database (like Dune, but, ofc, simpler at this point)
- gives back to the open-source community

Etherscan is cool and has lots of features, but it has none from the list above.

> One thing to note with 3XPL (not saying this isn't the case elsewhere), is that there is no filtering of spam tokens.

3xpl actually has the reverse: there are "verified" tokens (ETH itself included), so when you navigate to someone's address (say vitalik.eth: https://3xpl.com/ethereum/address/0xd8da6bf26964af9d7eed9e03e53415d37aa96045) you see the list of non-spam/scam tokens first:

image

and they all have a checkmark. We thought about putting a warning to tokens containing "USD" or similar, but I'm not exactly sure that's a perfect solution too. On our LLM API we output verified tokens only (see https://3xpl.ai/)

> Not saying there is anything particularly malicious at play here, but it seems a little unusual?

Some people build several blockchains, some build several block explorers :)

It's just two different architectures with their own pros and cons.

> I don't think this really adds anything that isn't covered by block explorers already on the list.

Well, but it's a bit strange that there are options with fewer features listed. For example, rantom.app doesn't support ENS, doesn't have any data features, returns a server error when trying to find a transaction (try https://rantom.app/tx/0x2feb29bfa5a89b402ea9ca4e1a83b3f1166815ab9772bbdabd5b945d48788950) or a 404 error when searching for an address (https://rantom.app/address/0xae2fc483527b8ef99eb5d9b44875f005ba1fae13) and it's on the list.

Sirato (aka chainlens.com) doesn't even seem to be a public Ethereum block explorer? There's only a "book a demo" button that works.

Please reconsider adding 3xpl to the list as we're trying to build a fast feature-rich browsing experience and would appreciate community support.

wackerow:
Hey @Har01d! Appreciate your reply here... gave this a look through and I tend to agree with a lot of your points. 

Notably, I appreciate the privacy approach taken here with the availability of a `.onion` address, which I agree is a distinguishing feature here (only found one other explorer listed that has an onion address, and it did not end up loading correctly for me... had no issues with 3xpl). Also very-much appreciate the lack of ads, open source repo, and open API access.

Given these points, I'd like to re-open and bring in this PR. 🎉 

---

> Sirato (aka chainlens.com) doesn't even seem to be a public Ethereum block explorer? There's only a "book a demo" button that works.

To this point, I also tend to agree... I wasn't aware this was a closed product, and don't think this is an appropriate place to direct users to for something like a block explorer... thank you for raising this!

wackerow:
@all-contributors please add @Har01d for tool

corwintines:
Thanks for the pushback @Har01d. Agree with the changes made here.

Har01d:
Great stuff, thanks!

PR #10278, opened 2023-05-20
refactor(theme): implement nested semantic tokens

TylerAPfledderer:
🚨IMPORTANT: This is a breaking change, affecting any existing PRs handling UI changes that involve the changed theme tokens

## Description

Updates semantic token naming to apply them in nested structures for improved organization.

Also updates color values.

## Related Issue
N/A


src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -34,40 +34,59 @@ const semanticTokens = {
...
-    successLight: "green.100",
-    successOutline: { _light: "success", _dark: "successLight" },
-    // ! Deprecating successNeutral
-    successNeutral: { _light: "green.100", _dark: "green.900" },
+
+    // Complementary Set
+    attention: {
+      base: "yellow.500",
+      light: "yellow.200",
+      outline: { _light: "attention", _dark: "attention.light" },
```
pettinarip, line 71: Should it be?
```suggestion
      outline: { _light: "attention.base", _dark: "attention.light" },
```

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -34,40 +34,59 @@ const semanticTokens = {
...
-    errorOutline: { _light: "error", _dark: "errorLight" },
-    // ! Deprecating errorNeutral
-    errorNeutral: { _light: "red.100", _dark: "red.900" },
-    attention: "yellow.500",
-    attentionLight: "yellow.200",
-    attentionOutline: { _light: "attention", _dark: "attentionLight" },
+    error: {
+      base: "red.500",
+      light: "red.100",
+      outline: { _light: "error", _dark: "error.light" },
```
pettinarip, line 77:
```suggestion
      outline: { _light: "error.base", _dark: "error.light" },
```

src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -34,40 +34,59 @@ const semanticTokens = {
...
+      // ! Deprecating errorNeutral
+      neutral: { _light: "red.100", _dark: "red.900" },
+    },
+    success: {
+      base: "green.500",
+      light: "green.100",
+      outline: { _light: "success", _dark: "success.light" },
```
pettinarip, line 84:
```suggestion
      outline: { _light: "success.base", _dark: "success.light" },
```

src/components/Search/utils.ts
```diff
@@ -154,11 +154,11 @@ export const getSearchModalStyles = (): SystemStyleObject => ({
 
   ".DocSearch-Footer": {
     "--docsearch-footer-height": "space.10",
-    "--docsearch-logo-color": "colors.bodyLight",
+    "--docsearch-logo-color": "colors.body.light",
     "--docsearch-footer-shadow": "none",
-    bg: "primaryLowContrast",
+    bg: "primary.owContrast",
```
pettinarip, line 159:
```suggestion
    bg: "primary.lowContrast",
```

src/components/Search/utils.ts
```diff
@@ -154,11 +154,11 @@ export const getSearchModalStyles = (): SystemStyleObject => ({
 
   ".DocSearch-Footer": {
     "--docsearch-footer-height": "space.10",
-    "--docsearch-logo-color": "colors.bodyLight",
+    "--docsearch-logo-color": "colors.body.light",
```
pettinarip, line 157: I know this is from a prev implementation but curious about this `colors` variable. Do you know why is there? how is it mapped? shouldn't it be just `body.light`?

src/components/Quiz/QuizWidget.tsx
```diff
@@ -301,7 +301,7 @@ const QuizWidget: React.FC<IProps> = ({ quizKey, maxQuestions }) => {
           size="50px"
           bg={
             !showAnswer
-              ? "primary"
+              ? "primary.base"
               : currentQuestionAnswerChoice?.isCorrect
               ? "success"
               : "error"
```
pettinarip, line 307: image

```suggestion
              ? "success.base"
              : "error.base"
```

src/components/Search/index.tsx
pettinarip, line 1: We lost the bg color in the modal. Not sure where the issue is.

image

pettinarip (changes requested):
Nice, I like this new notation, much cleaner.

src/components/Search/utils.ts
TylerAPfledderer: I'll need to find it in the source code so you have better context, but the mapping from the root level here is specifically for values declared in CSS Variables to be resolved correctly.

TylerAPfledderer: @pettinarip so for context, it starts in the `getCSS()` method (the root of the theme compiling) at the [isCSSVariableToken check](https://github.com/chakra-ui/chakra-ui/blob/38acfe89c5d1f1edc67bbc44e2edd38980ca3e08/packages/core/styled-system/src/css.ts#L63-L72). And in this check is this [token resolver](https://github.com/chakra-ui/chakra-ui/blob/38acfe89c5d1f1edc67bbc44e2edd38980ca3e08/packages/core/styled-system/src/css.ts#LL19C1-L29C2) to search in a map object, hence the value being dot notation starting at the root of the given collection `colors`.

src/components/Quiz/QuizWidget.tsx
TylerAPfledderer: @pettinarip note that for the second item you point out in the screenshot ("Sidechains") this is unrelated as the rendering of `white` is coming from logic for [QuizRadioGroup](https://github.com/ethereum/ethereum-org-website/blob/279659c5fafb754900c04bedc180402900c3d026/src/components/Quiz/QuizRadioGroup.tsx#L100-L104). 

This indeed looks like an error in color choice in light mode!

src/components/Search/index.tsx
TylerAPfledderer: Missed updating a CSS variable here!

src/components/Search/utils.ts
pettinarip: Oh interesting.

So, would it be similar as doing
```suggestion
    "--docsearch-logo-color": "var(--eth-colors-body-light)",
```
?

src/components/Quiz/QuizWidget.tsx
pettinarip: I think the `white` color is correct. The thing missing here is the red background.

This is in [dev](https://ethereumorgwebsitedev01.gatsbyjs.io/en/what-is-ethereum/)
image

src/components/Search/utils.ts
TylerAPfledderer: It would be converted to that CSS variable, yes.

src/components/Quiz/QuizWidget.tsx
TylerAPfledderer: Ahh found it! Pushed the fix.

static/static/eth-diamond-black-gray.svg
pettinarip, line 1: Why did the images change? probably are related to the latest merge update...I see that the permissions were changed. We need to revert them.

pettinarip (changes requested):
Just saw some changes on the images permissions, we need to revert that part. The rest looks fine. Thanks for adding that new paragraph to the docs :)

pettinarip (approved):
:shipit: 🎉

PR #10294, opened 2023-05-22
Fix staking product icon color

wackerow:
## Description
- Force staking product icon to `white` for both color modes to match the neighboring header text

## Related Issue
image

After patch
image


corwintines (approved):
Nice!

PR #10299, opened 2023-05-23
Handle SVGs / PNGs in Crowdin import script

minimalsm:
## Description

- We recently added SVG files to Crowdin.
- When pulling in the updates from Crowdin, the script was trying to open the SVG files as if they were folders and erroring out.
- This PR adds a check for other file types (SVGs/PNGs) to the `scrapeDirectory` function

### DIFF
--- a/src/scripts/crowdin-import.ts
+++ b/src/scripts/crowdin-import.ts
@@ -1,6 +1,12 @@
 // Library requires
-const { copyFileSync, existsSync, mkdirSync, readdirSync } = require("fs")
-const { resolve, join } = require("path")
+const {
+  copyFileSync,
+  existsSync,
+  mkdirSync,
+  readdirSync,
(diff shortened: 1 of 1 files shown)


minimalsm:
Ping @wackerow :) having to manually delete these files atm

Screenshot 2023-06-20 at 12 52 36

wackerow:
Hey @minimalsm, do we still need this? Looks like the `crowdin-import` file has been updated to check for `svg` files as well... I see you take the approach of anything that isn't a `json` file here... are we anticipating any other file formats? If not, I think we can probably just close this.

corwintines:
Discussed with @minimalsm, but this import process will be outdated with how we are going to do images. Going to close this out, and we will improve the import script or make a new one for image imports after we iron out that process a bit more. Going to close this out.

PR #10308, opened 2023-05-23
Intl patches

wackerow:
## Description
- Fixes a handful of front matter `image` properties in `translations/` folder to reference canonical image using `../../../../`
- Updates remnant `language-ur-in` translation keys to `language-ur`, removing the `(India)` suffix from the string itself for consistency
- `i18n/config.json` -- Deprecates unused `hrefLang` property, replacing with `crowdinCode`. Updates each with the code used by Crowdin.
- `crowdin-imports.ts` -- Deprecates lang-code mapping, replacing with a lookup from the canonical `i18n/config.json` file to fetch the Crowdin lang code assigned for each of our internal lang codes
- Updates Filipino lang code to use `fil` now that the codebase fully supports 3-letter language codes instead of just the `XX` or `XX-XX` format. Adds redirect for existing `/ph/*` links to forward to `/fil/:splat`

## Related Issue
Ongoing translation efforts


i18n/config.json
```diff
@@ -1,439 +1,439 @@
 [
   {
     "code": "en",
-    "hrefLang": "en",
```
minimalsm, line 4: Why are we deprecating hrefLang 🤔?

Looks like we aren't using them properly, though we should:

Screenshot 2023-05-30 at 11 07 13

cc: @wackerow @pettinarip

pettinarip (approved):

i18n/config.json
pettinarip: Yea, we are not using it right now. I don't think its a problem to lose this.

On the other side, the plugin that we are using is [taking care of `hrefLang` in theory](https://github.com/microapps/gatsby-plugin-react-i18next/blob/master/src/Link.tsx#LL25C14-L25C14).

corwintines (approved):

PR #10328, opened 2023-05-25
Updated info about POAPs

bskrksyp9:
An info regarding GitPOAP have been added., as sourced from official discord server.


wackerow (approved):
Hey @bskrksyp9 similar to #10329, just made a couple small adjustments and moved this below the standard POAP explainer. Bringing in!

PR #10329, opened 2023-05-25
Updated info about GitPOAPs

bskrksyp9:
Added info on GitPOAP and it's procedure to claim. Source as per official discord server.

### DIFF
--- a/src/content/contributing/index.md
+++ b/src/content/contributing/index.md
@@ -83,19 +83,6 @@ We have low barrier to entry issues on our GitHub repository specifically design
 
 ## Claim your contributor POAP {#poap}
 
-{For GitPOAP} (one time contribution)
-
-GitPOAP will automatically recognizes your merged contribution and lets you mint a unique contributors POAP on their platform itself!
-
-### How to claim {#how-to-claim}
-
-1. Visit [GitPOAP](https://www.gitpoap.io). 
-2. Connect with your wallet or even with your email id through sign in option.
-3. Search for your GitHub username, ETH address, ENS names or any GitPOAP to check if you're eligibile.
-4. If your github account is eligible, then you would be able to mint a GitPOAP!
- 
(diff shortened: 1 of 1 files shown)


src/content/contributing/index.md
```diff
@@ -83,6 +83,19 @@ We have low barrier to entry issues on our GitHub repository specifically design
...
 
+{For GitPOAP} (one time contribution)
+
+GitPOAP will automatically recognizes your merged contribution and lets you mint a unique contributors POAP on their platform itself!
+
+### How to claim {#how-to-claim}
+
+1. Visit [GitPOAP](https://www.gitpoap.io). 
+2. Connect with your wallet or even with your email id through sign in option.
+3. Search for your GitHub username, ETH address, ENS names or any GitPOAP to check if you're eligibile.
```
clauBv23, line 94:
```suggestion
3. Search for your GitHub username, ETH address, ENS names or any GitPOAP to check if you're eligible.
```

bskrksyp9: thanks for correction 👍🏼

wackerow (changes requested):
Hey @bskrksyp9... I see two PRs for this, #10328 and this one... can you help clarify which you're looking to use by closing the other? Assuming you wanted to go with this one since it came after and has more details.

Few notes on this one, I would put the information about GitPOAP after our note about "If your contribution gets merged...", and also we'll need to clean up the `{}` usage... would use bolded text or headers as needed.

bskrksyp9:
Hello @wackerow !

Yes you could modify the info as you mentioned, but the 2 PRs are different, though the change in content is same

One appears on Contributing page of site, where the other is from the github Readme Page

wackerow (approved):
@bskrksyp9 Just pushed a patch to this... shifted it below our standard POAP explainer, removed what looked like an unfinished note about tiered POAPs which we should leave off. Bringing it in, thanks!

PR #10337, opened 2023-05-26
Update TVS on What is Ethereum page

minimalsm:
## Description

- Update the link to point more directly to the source
- Change from measuring in T to measuring in B


nhsz (approved):
LGTM

PR #10342, opened 2023-05-26
Update Rocketpool min ETH in staking-products.json

IslamAbdelRahman:
## Description

Rocketpool now supports 8 ETH minipool (after Atlas Upgrade), the 16 number is outdated
https://docs.rocketpool.net/guides/atlas/whats-new.html#_8-eth-bonded-minipools

## Related Issue
N/A


src/data/staking-products.json
```diff
@@ -20,7 +20,7 @@
           "url": "https://github.com/trailofbits/publications/blob/master/reviews/RocketPool.pdf"
         }
       ],
-      "minEth": 16,
+      "minEth": 8,
```
corwintines, line 23:
```suggestion
      "minEth": 10.4,
```

Going to change this to 10.4 as you need to have RPL as well will RocketPool. 

On their node operators page they also highlight 10.4 due to this requirement: https://rocketpool.net/node-operators

corwintines (approved):
Thanks @IslamAbdelRahman

PR #10348, opened 2023-05-29
migrated: pages/developers/learning-tools.tsx to Chakra UI

Seek4samurai:
## Description
Migrated pages/developers/learning-tools.tsx to Chakra UI.

## Related Issue
https://github.com/ethereum/ethereum-org-website/issues/9353


src/pages/developers/learning-tools.tsx
```diff
@@ -1,78 +1,103 @@
...
+  </Subtitle>
+)
+
+const ContentBox = (props: ChildOnlyProp) => (
+  <Box pt={4} pb={4} pl={8} pr={8} w={"100%"} {...props}></Box>
+)
+
+const StackContainer = (props: ChildOnlyProp) => (
+  <Box
+    border="1px solid border"
```
minimalsm, line 86:
```suggestion
    border="1px solid"
    borderColor="border"
```

src/pages/developers/learning-tools.tsx
```diff
@@ -1,78 +1,103 @@
...
-  font-size: 1.25rem;
-  line-height: 1.4;
-  font-weight: 400;
-  color: ${(props) => props.theme.colors.text300};
-  max-width: 55ch;
-  margin-bottom: 0.5rem;
-  margin-top: 1rem;
-`
+const H1 = (props: ChildOnlyProp) => (
+  <Flex
```
minimalsm, line 47: This should be an H1

```suggestion
  <Heading
    as="h1"
```

src/pages/developers/learning-tools.tsx
```diff
@@ -1,78 +1,103 @@
...
+    m={"8 0 0"}
+    mt={0}
+    color="text"
+    fontStyle="normal"
+    fontFamily={"mono"}
+    textTransform="uppercase"
+    fontWeight="semibold"
+    fontSize="4xl"
```
minimalsm, line 55: Let's stick to current design, which is 32px.

```suggestion
    fontSize="2rem"
```

src/pages/developers/learning-tools.tsx
```diff
@@ -1,78 +1,103 @@
...
-    margin-left: 0rem;
-    margin-right: 0rem;
-    border-radius: 0px;
-  }
-`
+const SubtitleTwo = (props: ChildOnlyProp) => (
+  <Subtitle>
+    <Box mt={0} {...props}></Box>
+  </Subtitle>
+)
```
minimalsm, line 78: Currently, this doesn't remove the margin on the Subtitle (which is what we want). It adds another element inside of it, but the parent keeps its margin-top.

We can do this:

```suggestion
const SubtitleTwo = (props) => (
  <Subtitle mt={0} {...props}/>
)
```

src/pages/developers/learning-tools.tsx
```diff
@@ -1,78 +1,103 @@
...
+  />
+)
 
-const SubtitleTwo = styled(Subtitle)`
-  margin-top: 0rem;
-`
+const Subtitle = (props: ChildOnlyProp) => (
+  <Flex
+    fontSize={"xl"}
+    lineHeight={"short"}
```
minimalsm, line 65: Just noting, this is a tiny bit different from what we have currently. Not sure if we care about this @pettinarip?

minimalsm (changes requested):
Hey @Seek4samurai, nice work on this 💪

I've Made some suggested tweaks to get this closer to the current design. 

The main thing we need to solve is that we are setting some breakpoints the wrong way round (e.g. 0px margin on big screens and 8px on small, instead of the opposite). I've left suggestions that I believe fixes all of these 🫡

minimalsm (dismissed):
Hey @Seek4samurai, nice work on this 💪

I've Made some suggested tweaks to get this closer to the current design. 

The main thing we need to solve is that we are setting some breakpoints the wrong way round (e.g. 0px margin on big screens and 8px on small, instead of the opposite). I've left suggestions that I believe fixes all of these 🫡

Seek4samurai:
Ok maybe I did messed up some breakpoints. Thanks!

src/pages/developers/learning-tools.tsx
pettinarip: Yea, we want to keep the current styles so this should be `lineHeight={1.4}`

pettinarip (approved):
Thanks for the PR @Seek4samurai.

I've applied the suggestions from the reviews. This is good to go now 💪🏼

PR #10320, opened 2023-05-24
feat: Learning Quizzes Hub

nhsz:
PR for the Quiz Hub page.

## New

- `/quizzes` page 
- `/utils`
- `/context` (Context API)
- `QuizItem` component
- `QuizzesList` component
- `QuizzesStats` component
- `QuizzesModal` component
- UI translations

## Updated/refactored

- `QuizWidget` component
- `QuizSummary` component

## Links

- [**Design**](https://www.figma.com/file/1HxSnapMdpgFFhIHjDUytP/Learning-quizzes?type=design&node-id=2472-10885&t=vJG6GB4YbSvzRK8k-0)
- [**Preview**](https://ethereumorgwebsitedev01-featlearningquizzeshub.gatsbyjs.io/en/quizzes/)

## QA list

- UI bugs
- UX bugs
- mobile, responsive design
- dark mode
-  standalone version of the quizzes (like the one at the bottom of [/layer-2/](https://ethereumorgwebsitedev01-featlearningquizzeshub.gatsbyjs.io/en/layer-2/)) still works as expected.

PR #10349, opened 2023-05-29
Localise number formatting in quiz stats

minimalsm:
## Description

Added enhanced localization and formatting for numbers on the quiz stats component:

- Formatted numbers and percentages based on user locale (We do this in quite a few places already, e.g. [the layer 2 page](https://github.com/ethereum/ethereum-org-website/blob/87050ee3d1ed311dbe6d1794624f91ce9675ea6c/src/pages/layer-2.tsx#L196))
- Handle placement of the in-text greater than indicator in RTL languages 

## Related Issue

#10320

### DIFF
--- a/src/components/Quiz/QuizzesStats.tsx
+++ b/src/components/Quiz/QuizzesStats.tsx
@@ -11,6 +11,7 @@ import {
   Text,
 } from "@chakra-ui/react"
 import { FaTwitter } from "react-icons/fa"
+import { useI18next } from "gatsby-plugin-react-i18next"
 
 import Button from "../Button"
 import Translation from "../Translation"
@@ -18,7 +19,13 @@ import { TrophyIcon } from "../icons/quiz"
 
 import { QuizzesHubContext } from "./context"
 
+// Utils
 import { getNumberOfCompletedQuizzes, getTotalQuizzesPoints } from "./utils"
+import {
+  isLangRightToLeft,
+  getLocaleForNumberFormat,
+} from "../../utils/translations"
+import { Lang } from "../../utils/languages"
 
 import { QuizShareStats } from "../../types"
 
@@ -50,6 +57,9 @@ const handleShare = ({ score, total }: QuizShareStats) =>
   })
 
 const QuizzesStats: React.FC = () => {
+  const { language } = useI18next()
+  const localeForQuizStats = getLocaleForNumberFormat(language as Lang)
+  const isRightToLeft = isLangRightToLeft(language as Lang)
   const { score: userScore, completed, average } = useContext(QuizzesHubContext)
   const totalQuizzesNumber =
     ethereumBasicsQuizzes.length + usingEthereumQuizzes.length
@@ -57,6 +67,38 @@ const QuizzesStats: React.FC = () => {
   const numberOfCompletedQuizzes = getNumberOfCompletedQuizzes(
     JSON.parse(completed)
   )
+
+  // Data from Matomo, manually updated
+  const collectiveQuestionsAnswered = 100000
+  const collectiveAverageScore = 67.4 / 100 // converted to fraction for percentage format
+  const collectiveRetryRate = 15.6 / 100 // converted to fraction for percentage format
+
+  const formatNumber = new Intl.NumberFormat(localeForQuizStats, {
+    style: "decimal",
+    minimumSignificantDigits: 2,
+    maximumSignificantDigits: 3,
+  })
+
+  const formatPercent = new Intl.NumberFormat(localeForQuizStats, {
+    style: "percent",
+    minimumSignificantDigits: 2,
+    maximumSignificantDigits: 3,
+  })
+
+  // formatted collective stats
+  const formattedCollectiveQuestionsAnswered = formatNumber.format(
+    collectiveQuestionsAnswered
+  )
+  // If language is RTL, more than indicator should go at start
+  const textForCollectiveQuestions = isRightToLeft
+    ? `+${formattedCollectiveQuestionsAnswered}`
+    : `${formattedCollectiveQuestionsAnswered}+`
+
+  const formattedCollectiveAverageScore = formatPercent.format(
+    collectiveAverageScore
+  )
+  const formattedCollectiveRetryRate = formatPercent.format(collectiveRetryRate)
+  const formattedUserAverageScore = formatPercent.format(average)
 
   return (
     <Box flex={1} order={{ base: 1, lg: 2 }} w="full">
@@ -125,7 +167,7 @@ const QuizzesStats: React.FC = () => {
                 <Text mr={10} mb={0} mt={{ base: 2, lg: 0 }} color="bodyLight">
                   <Translation id="average-score" />{" "}
                   <Text as="span" color="body">
-                    {average}%
+                    {formattedUserAverageScore}
                   </Text>
                 </Text>
 
@@ -163,7 +205,7 @@ const QuizzesStats: React.FC = () => {
                 <Translation id="average-score" />
               </Text>
               {/* Data from Matomo, manually updated */}
-              <Text color="body">67.4%</Text>
+              <Text color="body">{formattedCollectiveAverageScore}</Text>
             </Stack>
 
             <Stack>
@@ -172,7 +214,7 @@ const QuizzesStats: React.FC = () => {
               </Text>
 
               {/* Data from Matomo, manually updated */}
-              <Text color="body">100 000+</Text>
+              <Text color="body">{textForCollectiveQuestions}</Text>
             </Stack>
 
             <Stack>
@@ -181,7 +223,7 @@ const QuizzesStats: React.FC = () => {
               </Text>
 
               {/* Data from Matomo, manually updated */}
-              <Text color="body">15.6%</Text>
+              <Text color="body">{formattedCollectiveRetryRate}</Text>
             </Stack>
           </Flex>
         </Flex>


src/components/Quiz/QuizzesStats.tsx
```diff
@@ -172,7 +214,7 @@ const QuizzesStats: React.FC = () => {
               </Text>
 
               {/* Data from Matomo, manually updated */}
-              <Text color="body">100 000+</Text>
+              <Text color="body">{textForCollectiveQuestions}</Text>
```
wackerow, line 217: If you use this, you can skip the "+" logic above.
```suggestion
              <Text color="body" _after={{ content: `"+"` }}>
                {formattedCollectiveQuestionsAnswered}
              </Text>
```

src/components/Quiz/QuizzesStats.tsx
```diff
@@ -50,6 +57,9 @@ const handleShare = ({ score, total }: QuizShareStats) =>
   })
 
 const QuizzesStats: React.FC = () => {
+  const { language } = useI18next()
+  const localeForQuizStats = getLocaleForNumberFormat(language as Lang)
+  const isRightToLeft = isLangRightToLeft(language as Lang)
```
wackerow, line 62: See line 217
```suggestion
```

src/components/Quiz/QuizzesStats.tsx
```diff
@@ -11,14 +11,21 @@ import {
...
 
 import { QuizzesHubContext } from "./context"
 
+// Utils
 import { getNumberOfCompletedQuizzes, getTotalQuizzesPoints } from "./utils"
+import {
+  isLangRightToLeft,
+  getLocaleForNumberFormat,
+} from "../../utils/translations"
+import { Lang } from "../../utils/languages"
```
wackerow, line 28: See line 217
```suggestion
import { getLocaleForNumberFormat } from "../../utils/translations"
import { Lang } from "../../utils/languages"
```

src/components/Quiz/QuizzesStats.tsx
```diff
@@ -58,6 +68,38 @@ const QuizzesStats: React.FC = () => {
...
+  })
+
+  // formatted collective stats
+  const formattedCollectiveQuestionsAnswered = formatNumber.format(
+    collectiveQuestionsAnswered
+  )
+  // If language is RTL, more than indicator should go at start
+  const textForCollectiveQuestions = isRightToLeft
+    ? `+${formattedCollectiveQuestionsAnswered}`
+    : `${formattedCollectiveQuestionsAnswered}+`
```
wackerow, line 95: See line 217
```suggestion
```

src/components/Quiz/QuizzesStats.tsx
```diff
@@ -58,6 +68,38 @@ const QuizzesStats: React.FC = () => {
...
+    style: "decimal",
+    minimumSignificantDigits: 2,
+    maximumSignificantDigits: 3,
+  })
+
+  const formatPercent = new Intl.NumberFormat(localeForQuizStats, {
+    style: "percent",
+    minimumSignificantDigits: 2,
+    maximumSignificantDigits: 3,
+  })
```
wackerow, line 86: Wondering if it's worth pulling these out and importing them from the utils folder

wackerow (commented):
@minimalsm Left a few suggestions! If we use the `after` pseudoelement with content of `+`, the DOM will automatically render that inline and take direction into account (always placing it "after" by default). Should help simplify some logic here =)

src/components/Quiz/QuizzesStats.tsx
minimalsm: Agree! My bad, should have checked for that. 

We could refactor it to accept props for number of significant digits and use that across the codebase.

src/components/Quiz/QuizzesStats.tsx
minimalsm: Ahh nice thinking, I hadn't thought of this. Agree that the code is pretty convoluted and open to ideas on how to make it less so 😁

One potential problem with using `after` is that pseudo-element elements are ignored by screen readers.

nhsz:
@minimalsm I've done some updates to they way I calculate average, new variable is `parsedAverage`

minimalsm:
Updated @nhsz 🫡

nhsz:
@minimalsm I'd suggest to parse integer numbers to avoid redundant decimals when float is `0`, like `0.0%, 50.0%`, etc, IMO those should be displayed as `0%, 50%`, etc.

I was doing that with

```
const parsedAverage = Number.isInteger(computedAverage)
    ? computedAverage
    : computedAverage.toFixed(2)
```

src/components/Quiz/QuizzesStats.tsx
pettinarip: Good point. I'd say that if we care about sr then we could use a inner span element. This will get placed at the beginning if `ltr`.
```suggestion
              <Text color="body">
                {formattedCollectiveQuestionsAnswered}
                <Text as="span">+</Text>
              </Text>
```

src/components/Quiz/QuizzesStats.tsx
```diff
@@ -42,24 +49,65 @@ const handleShare = ({ score, total }: QuizShareStats) => {
...
-    : computedAverage.toFixed(2)
-
   // These values are not fixed but calculated each time, can't be moved to /constants
   const totalQuizzesNumber =
     ethereumBasicsQuizzes.length + usingEthereumQuizzes.length
   const totalQuizzesPoints = getTotalQuizzesPoints()
 
+  const computedAverage =
+    userAverageScores.length > 0
+      ? userAverageScores.reduce((a, b) => a + b, 0) / userAverageScores.length
```
pettinarip, line 69: fyi `_.mean(userAverageScores)`

pettinarip (approved):
Nice @minimalsm

-----
Nice to have: would recommend to isolate all this new logic into its own function.
```tsx
// to have something like this
const { language } = useI18next()
...
const formattedStats = getFormattedStats(language, userAverageScores)
```

minimalsm:
> @minimalsm I'd suggest to parse integer numbers to avoid redundant decimals when float part is `0`, like `0.0%, 50.0%`, etc, IMO those should be displayed as `0%, 50%`, etc.
> 
> I was doing that with
> 
> ```
> const parsedAverage = Number.isInteger(computedAverage)
>     ? computedAverage
>     : computedAverage.toFixed(2)
> ```
> 
> logic should be probably adapted a bit for this solution

@nhsz the number formatter already does this for us (unless I misunderstand you).

```javascript
const percentFormatter = new Intl.NumberFormat("en-US", {
    style: "percent",
    minimumSignificantDigits: 2,
    maximumSignificantDigits: 3,
  })

  console.log(percentFormatter.format(0.49)) // 49%
  console.log(percentFormatter.format(0.4900000000)) //49%
  console.log(percentFormatter.format(0.499)) //49.9%
  console.log(percentFormatter.format(0.4999)) // 50%
```

src/components/Quiz/QuizzesStats.tsx
minimalsm: Added both 👍

minimalsm:
> Nice @minimalsm
> 
> Nice to have: would recommend to isolate all this new logic into its own function.
> 
> ```tsx
> // to have something like this
> const { language } = useI18next()
> ...
> const formattedStats = getFormattedStats(language, userAverageScores)
> ```

Good point! [Done here 🫡](https://github.com/ethereum/ethereum-org-website/pull/10349/commits/6fbabe4a23e40efd803f5282c2f09a9ff83c9ce8)

src/components/Quiz/QuizzesStats.tsx
```diff
@@ -65,41 +107,23 @@ const QuizzesStats: React.FC = () => {
...
-  const collectiveRetryRate = 15.6 / 100 // converted to fraction for percentage format
-
-  // Todo: make global util function
-  const numberFormatter = new Intl.NumberFormat(localeForNumbers, {
-    style: "decimal",
-    minimumSignificantDigits: 2,
-    maximumSignificantDigits: 3,
-  })
+  const collectiveAverageScore = 67.4
+  const collectiveRetryRate = 15.6
```
minimalsm: Can we move these to constants?

nhsz:
> > @minimalsm I'd suggest to parse integer numbers to avoid redundant decimals when float part is `0`, like `0.0%, 50.0%`, etc, IMO those should be displayed as `0%, 50%`, etc.
> > 
> > I was doing that with
> > 
> > ```
> > const parsedAverage = Number.isInteger(computedAverage)
> >     ? computedAverage
> >     : computedAverage.toFixed(2)
> > ```
> > 
> > logic should be probably adapted a bit for this solution
> 
> @nhsz the number formatter already does this for us (unless I misunderstand you).
> 
> ```javascript
> const percentFormatter = new Intl.NumberFormat("en-US", {
>     style: "percent",
>     minimumSignificantDigits: 2,
>     maximumSignificantDigits: 3,
>   })
> 
>   console.log(percentFormatter.format(0.49)) // 49%
>   console.log(percentFormatter.format(0.4900000000)) //49%
>   console.log(percentFormatter.format(0.499)) //49.9%
>   console.log(percentFormatter.format(0.4999)) // 50%
> ```

@minimalsm I tried completing a quiz with 50% score and it's displayed as `50.0%` instead of `50%` (integer). Also initial average appears as `0.0%` instead of just `0%`

src/components/Quiz/QuizzesStats.tsx
pettinarip: That make sense, yes.

src/components/Quiz/QuizzesStats.tsx
```diff
@@ -41,25 +49,82 @@ const handleShare = ({ score, total }: QuizShareStats) => {
...
+  const numberFormatter = new Intl.NumberFormat(localeForNumbers, {
+    style: "decimal",
+    minimumSignificantDigits: 2,
+    maximumSignificantDigits: 3,
+  })
+
+  const percentFormatter = new Intl.NumberFormat(localeForNumbers, {
+    style: "percent",
+    minimumSignificantDigits: 2,
+    maximumSignificantDigits: 3,
```
minimalsm, line 71: - Instead of showing single digit percentages (0-9%) as x.0% (2 significant digits) show as 0%
- Instead of showing single digit numbers (0-9) as x.0 (2 significant digits) show an integer 

```suggestion
  const numberFormatter = new Intl.NumberFormat(localeForNumbers, {
    style: "decimal",
    minimumSignificantDigits: 1,
    maximumSignificantDigits: 3,
  })

  const percentFormatter = new Intl.NumberFormat(localeForNumbers, {
    style: "percent",
    minimumSignificantDigits: 1,
    maximumSignificantDigits: 3,
```

minimalsm:
> @minimalsm I tried completing a quiz with 50% score and it's displayed as `50.0%` instead of `50%` (integer). Also initial average appears as `0.0%` instead of just `0%`

@nhsz [should be fixed in this commit](https://github.com/ethereum/ethereum-org-website/pull/10349/files#r1219675628):

Can you confirm?

src/components/Quiz/QuizzesStats.tsx
minimalsm: Done https://github.com/ethereum/ethereum-org-website/pull/10349/commits/16e03f378b9fa3c47e54f1579345d17e51539274

nhsz:
> > @minimalsm I tried completing a quiz with 50% score and it's displayed as `50.0%` instead of `50%` (integer). Also initial average appears as `0.0%` instead of just `0%`
> 
> @nhsz [should be fixed in this commit](https://github.com/ethereum/ethereum-org-website/pull/10349/files#r1219675628):
> 
> Can you confirm?

LGTM now, thanks! I have moved `getFormattedStats` to `/utils` to clean the codebase a bit. Merging!

nhsz (approved):

PR #10356, opened 2023-05-31
Embracing the Beacon Chain 🤝

SebastianSupreme:
Improved version of https://github.com/ethereum/ethereum-org-website/pull/9589


src/content/roadmap/beacon-chain/index.md
```diff
@@ -13,17 +13,14 @@ summaryPoint3: The Beacon Chain introduced the consensus logic and block gossip
...
 
-The Beacon Chain was the name given to a ledger of accounts that conducted and coordinated the network of Ethereum [stakers](/staking/) before those stakers started validating real Ethereum transactions. It did not process transactions or handle smart contract interactions.
-
-It introduced the consensus engine (or "consensus layer") that took the place of proof-of-work mining on Ethereum and brought many significant improvements with it.
-
-The Beacon Chain was a foundational component for [the secure, environmentally friendly and scalable Ethereum we have now](/roadmap/vision/).
+The Beacon Chain is the name given to a ledger of accounts that conducted and coordinated the network of Ethereum [stakers](/staking/) before those stakers started validating real Ethereum blocks. It does not process transactions or handle smart contract interactions though because that is being done in the execution layer.
+The Beacon Chain is responsible for things like block and attestation gossip, running the fork choice algorithm, following the head of the chain, and [much more](/developers/docs/nodes-and-clients/node-architecture/#node-comparison).
```
jmcook1186, line 23:
```suggestion
The Beacon Chain is responsible for things like block and attestation handling, running the fork choice algorithm, and managing rewards and penalties. 

Read more on our [node architecture page](/developers/docs/nodes-and-clients/node-architecture/#node-comparison).
```

src/content/roadmap/beacon-chain/index.md
```diff
@@ -13,17 +13,14 @@ summaryPoint3: The Beacon Chain introduced the consensus logic and block gossip
...
+The Beacon Chain is the name of the original proof-of-stake blockchain that was launched in 2020. It was created to ensure the proof-of-stake consensus logic was sound and sustainable before enabling it on Ethereum Mainnet. Therefore, it ran alongside the original proof-of-work Ethereum. Switching off proof-of-work and switching on proof-of-stake on Ethereum required instructing the Beacon Chain to accept transactions from the original Ethereum chain, bundle them into blocks and then organize them into a blockchain using a proof-of-stake based consensus mechanism. At the same moment, the original Ethereum clients turned off their mining, block propagation and consensus logic, handing that all over to the Beacon Chain. This event was known as [The Merge](/roadmap/merge/). Once The Merge happened, there were no longer two blockchains; there was just one proof-of-stake Ethereum chain. The Merge also introduced a modular blockchain design, which is why Ethereum now consists of two layers that communicate through the Engine API. Those layers are the consensus layer (i.e. the imported Beacon Chain) and the execution layer (i.e. the original Ethereum blockchain).
```
jmcook1186, line 18: Modular blockchains generally refers to systems that use separate networks for their data availability, computation and/or consensus. I do understand where you are coming from calling out the modularity of the two clients, but I think it could confuse people because Ethereum is still monolithic (one chain handles execution, consensus and data availability), at least for now. 

```suggestion
The Beacon Chain is the name of the original proof-of-stake blockchain that was launched in 2020. It was created to ensure the proof-of-stake consensus logic was sound and sustainable before enabling it on Ethereum Mainnet. Therefore, it ran alongside the original proof-of-work Ethereum. Switching off proof-of-work and switching on proof-of-stake on Ethereum required instructing the Beacon Chain to accept transactions from the original Ethereum chain, bundle them into blocks and then organize them into a blockchain using a proof-of-stake based consensus mechanism. At the same moment, the original Ethereum clients turned off their mining, block propagation and consensus logic, handing that all over to the Beacon Chain. This event was known as [The Merge](/roadmap/merge/). Once The Merge happened, there were no longer two blockchains; there was just one proof-of-stake Ethereum chain.
```

jmcook1186 (changes requested):
left some nitpicks, once they are resolved this sgtm - thanks!

src/content/roadmap/beacon-chain/index.md
SebastianSupreme: Would it be ok to just swap modular for layered?
Because I do think it's important to create that connection between Beacon Chain and consensus layer here in this paragraph that introduces and defines the Beacon Chain.
And there's just no way of doing that without mentioning the two layers.

jmcook1186: Ok, fair point. How about just being very explicit about it? Something like:

"Since The Merge, Ethereum nodes have had to run two clients: one responsible for gossiping and executing transactions and managing Ethereum's state, and one that handles block gossip and consensus logic. Each of these clients has its own peer-to-peer network, which is why it is common to refer to the "execution layer" and "consensus layer".  The two layers can communicate with one another using the [Engine API](https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md). Up until The Merge, the Beacon Chain was a chain of 'empty' blocks created by the consensus layer, before it started accepting transaction data from execution clients."

SebastianSupreme:
@jmcook1186 Should be good to go now

PR #10360, opened 2023-06-01
Correct heading level on `eth_chainId`

feuGeneA:
I was seeing it rendered weirdly, and I found this mistake.

Here's what I was seeing:
Right side-bar:
image
Body:
image

Furthermore, the link in the right side bar was doing nothing when I clicked it.

I was using Firefox 102.4.0esr (64-bit) on Debian Linux.


minimalsm (approved):
Nice catch! Thanks for the PR @feuGeneA 💪

minimalsm:
@all-contributors please add @feuGeneA for bugs, content

PR #10365, opened 2023-06-02
add Node RPC docs to the list

xivanc:
Add Node RPC docs to the list

## Description

link to main page and docs page

## Related Issue
N/A


wackerow:
Hey @xivanc, apologies this PR hasn't been addressed yet.. thanks for the suggestion here...

Can't help but notice that this page is getting a bit unwieldy with this list growing quite large.... we typically try to strike a balance with user experience by not overwhelming readers with options, and tend to look for listings that are going to bring something new to the table that isn't already being served.

@konopkja Any thoughts here from a UX perspective? I'm not necessarily against listing this given it's in the develop docs which tends to have a bit more flexibility, but want to be careful.

@xivanc Care to elaborate on something that makes Node RPC may provide that isn't covered by the others listed?

xivanc:
> Care to elaborate on something that makes Node RPC may provide that isn't covered by the others listed?

@wackerow thank you for reaching out. We are an RPC provider focusing on Ethereum and L2s. We don't offer any special features that you can't find with other providers, but our value lies in being cost effective. We are 10x-100x cheaper than any other known RPC provider (details [here](https://www.noderpc.xyz/blog/posts/ethereum-evm-rpc-price-comparison)), and maybe there are devs out there who are tight on budget and looking for cheap solutions for their projects.

konopkja:
> Hey @xivanc, apologies this PR hasn't been addressed yet.. thanks for the suggestion here...
> 
> Can't help but notice that this page is getting a bit unwieldy with this list growing quite large.... we typically try to strike a balance with user experience by not overwhelming readers with options, and tend to look for listings that are going to bring something new to the table that isn't already being served.
> 
> @konopkja Any thoughts here from a UX perspective? I'm not necessarily against listing this given it's in the develop docs which tends to have a bit more flexibility, but want to be careful.
> 
> @xivanc Care to elaborate on something that makes Node RPC may provide that isn't covered by the others listed?

from UX perspective it is always better to list less choices, but idk if we should list only a single node RPC provider. Am i missing any other discussion point here @wackerow?

xivanc:
@konopkja 

If the list is too long, it might make sense to introduce categories, similar to what you do with wallets here https://ethereum.org/en/wallets/find-wallet/#main-content ?

For example, our provider only focuses on Ethereum and L2s. Other providers may offer additional APIs around erc20 tokens, NFTs. Some providers offer additional Account Abstraction APIs (ERC-4337), etc. 

This gives site visitors a more advanced UI to find what they're looking for.

xivanc:
@wackerow , thanks for reply. Super bad timing, our infra got down for literally a moment, it's back on again.

Daniel

wackerow (approved):
Hah, yeah bad timing I guess... looks good now! So looks like the only point of indecision here is that there are a fair amount of products on this page and is in need of some clean up.

I'm going to pull this in, and can post a new issue recommending we clean these up somehow, with my preference being to categorize these a bit if possible to help guide users.

Thanks @xivanc!

wackerow:
@all-contributors please add @xivanc for tool

PR #10375, opened 2023-06-04
refactor: template/docs.tsx to chakra

amit-ksh:
## Description
Migrate the `template/docs.tsx` to ChakraUI.

## Related Issue
#9353

### DIFF
--- a/src/templates/docs.tsx
+++ b/src/templates/docs.tsx
@@ -50,7 +50,7 @@ const Page = (props: ChildOnlyProp & Pick<FlexProps, "dir">) => (
   <Flex
     direction="column"
     w="full"
-    border="1px"
+    borderBottom="1px"
     borderColor="border"
     {...props}
   />
--- a/src/templates/docs.tsx
+++ b/src/templates/docs.tsx
@@ -1,9 +1,21 @@
-import React, { useContext } from "react"
+import React, { ComponentPropsWithoutRef, useContext } from "react"
 import { graphql, PageProps } from "gatsby"
 import { MDXProvider } from "@mdx-js/react"
 import { MDXRenderer } from "gatsby-plugin-mdx"
-import styled from "@emotion/styled"
-import { Badge } from "@chakra-ui/react"
+import {
+  Badge,
+  Divider as ChakraDivider,
+  Flex,
+  FlexProps,
+  Heading,
+  ListItem as ChakraListItem,
+  ListItemProps,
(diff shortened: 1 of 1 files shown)


src/templates/docs.tsx
```diff
@@ -21,117 +33,153 @@ import TableOfContents, {
...
-  display: flex;
-  flex-direction: column;
-  width: 100%;
-  border-bottom: 1px solid ${(props) => props.theme.colors.border};
-`
+const Page = (props: ChildOnlyProp & Pick<FlexProps, "dir">) => (
+  <Flex
+    direction="column"
+    w="full"
+    border="1px"
```
TylerAPfledderer, line 53:
```suggestion
    borderBottom="1px"
```

TylerAPfledderer (changes requested):
Hey @amit-ksh! I got a couple of tiny things for you, but otherwise LGTM. 🎉 

In the [components object](https://github.com/ethereum/ethereum-org-website/pull/10375/files#diff-ac052215d439cdda9c27e55f13d17cb4a52c422d14cecc7397adc934c651d5e8R186) for MDX, you'll need to add the `UnorderedList` component from Chakra as the `ul` component, because the use of `ListItem` from Chakra requires the provider supplied by the former.

```ts
const components = {
 // ... //
 ul: UnorderedList,
 // ... //
}

**Comment by amit-ksh** [comment]
Hey @TylerAPfledderer, I made all changes as you suggested.

File: src/templates/docs.tsx [hunk 2]
```diff
@@ -158,11 +218,9 @@ const components = {
...
 }
 
-const Contributors = styled(FileContributors)`
-  @media (max-width: ${(props) => props.theme.breakpoints.l}) {
-    padding-bottom: 2rem;
-  }
-`
+const Contributors = (
+  props: ComponentPropsWithoutRef<typeof FileContributors>
+) => <FileContributors pb={{ base: 8, lg: 2 }} {...props} />
```
pettinarip, line 223: We'll need to override the `p` with the same breakpoints we are using in the `pb` style, otherwise the `pb` is not going to be applied correctly.

```suggestion
) => <FileContributors p={{ base: 0, lg: 2 }} pb={{ base: 8, lg: 2 }} {...props} />
```

pettinarip (approved):
Looks great @amit-ksh, I've added some tiny changes. Thanks for the PR.

PR #10377, opened 2023-06-05
Improve QuizSummary copy

minimalsm:
## Description

The current quiz summary is a little confusing, as it isn't entirely clear what 'Total' means.

image

- To make this clearer, we can switch 'Total' with 'Questions'.
- We already have a 'questions' string in the JSON file, so simply replaced the string and removed the unused JSON keys

## Related Issue
#10320


nhsz (approved):
LGTM, thanks!

Issue #10266, opened 2023-05-19
[Tag] Update close button styles


pettinarip:
Update the styles for the close button inside the Tags.

![image](https://github.com/ethereum/ethereum-org-website/assets/468158/cfefc7cf-c5d9-4fb9-95b3-4b4ccb576bd0)

https://github.com/ethereum/ethereum-org-website/assets/1120748/0474f8e4-8127-46a0-8983-b2239d259ab4

DS: https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?type=design&node-id=1600-34738&t=thqLEloXh2D9Dd4H-0

TylerAPfledderer:
Commenting for assignment 😄

wackerow:
Hey @TylerAPfledderer! Just circling through... Any update? Happy to open this back up for another contributor

TylerAPfledderer:
@wackerow Hmm this is quite old. Don't know why this was missed!

Keep me on the. I believe I ran into an issue where I needed a consult from @pettinarip and @nloureiro, but not sure what it was here.

Thank you for the follow-up!

PR #10380, opened 2023-06-05
feat(theme): add Alert component and theme

TylerAPfledderer:
## Description
_This is part of the [Implementation of the DS v1](https://github.com/ethereum/ethereum-org-website/issues/9548)_

Adds the `Alert` component and theming per the [new DS](https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?type=design&node-id=2114-43571)


nloureiro:
@TylerAPfledderer the hover effect on the close button for the alerts needs to be the same as the tags, my bad,  I did not add this detail to the Figma file. sorry about that

pettinarip (changes requested):
Nice job @TylerAPfledderer 

Based on @nloureiro's comment, we need to apply the same styles on the close button as in the Tag component. https://dev--63b7ea99632763723c7f4d6b.chromatic.com/?path=/story/molecules-display-content-tags--style-variants-basic

FYI and context, we also have this other issue https://github.com/ethereum/ethereum-org-website/issues/10266 that is going to impact both components styles.

TylerAPfledderer:
@pettinarip right. Would it be fine to handle the close button as a separate PR as before, but one PR to cover both the tag and the alert?

pettinarip:
> @pettinarip right. Would it be fine to handle the close button as a separate PR as before, but one PR to cover both the tag and the alert?

Make sense. Will approve it then 👍🏼

pettinarip (approved):

PR #10389, opened 2023-06-07
KaTeX support for mathematical expressions [Closes #7653]

### LINKED ISSUE #7653: Mathematical expression support for markdown
Currently our markdown does not support syntax for mathematical expressions. Although these are not super frequent, having the ability to properly display mathematical expressions could expand our ability to convey certain concepts. I suspect we could use more math examples if we had the proper support for them. 

Surfaced while working through the new page related to proof-of-stake rewards and penalties. 

https://github.blog/2022-05-19-math-support-in-markdown/
GitHub recently rolled out a similar feature, which can be used with `$` or `$$` delimiters. ie. $e^{i \pi} = -1$ instead of `e^(i*pi) = -1`
Not sure exactly what library they use.
Couple possible libraries
https://www.gatsbyjs.com/plugins/gatsby-remark-katex/
https://www.mathjax.org/

Just leave the markdown as is, and type all math examples with inline text. :-(

**Additional context**
The method chosen should be sure to support dark mode color themes.
Ideally text is still selectable, rather than rendered as an image. 
Ideally also support RTL languages, but not certain how that would play in.

Example of mathematic expressions being used: https://eth2book.info/altair/annotated-spec/#reward-and-penalty-calculations

**Want to contribute?**

wackerow:
## Description
- Adds support for KaTeX mathematical expressions
- Markdown: Can be used with single $ wrappers for inline expressions, and double $$ on their own lines to make a centered block expression.
- React/HTML: When markdown isn't available, can use `<span className="math-inline">` for inline and `<div className="math-display">` for centerer block expressions.
- Markdown support currently implemented for the static, docs and tutorials templates
- Updated the `/roadmap/merge/issuance/` page to take advantage of the new syntax in the mathematical expressions at the bottom for demonstration purposes.

## Packages
- Uses `remark-math@3.0.1` to read markdown for related syntax and convert to `math-inline` and `math-display` elements for transforming (using old version intentionally for compatibility with `gatsby-plugin-mdx`)
- Uses `@matejmazur/react-katex` to transform elements into KaTeX elements
- Uses `katex` for styling of elements
[Gatsby implementation explainer guide](https://nickymeuleman.netlify.app/blog/math-gatsby-mdx)

## Preview link
- https://ethereumorgwebsitedev01-katexsupport.gatsbyjs.io/en/roadmap/merge/issuance/#calculating-average-gas-price-for-deflation

## Todo
- [x] Document in ~~README~~ repo docs

## Related Issue
- Closes #7653


nhsz:
Something to take into account, we currently have an [open issue in the blog](https://github.com/ethereum/blog/issues/1034) related to using `$` for inline expressions which causes text like `$10...` (talking about an amount of money) to render the rest of the paragraph as a LaTeX expression. I checked that some time ago but din't found a good solution to escape those cases

wackerow:
Right, this is something we would need to be aware of with content going forward, but you should be able to escape them with a `\$`...

I made this adjustment in [translators-guide/index.md](https://ethereumorgwebsitedev01-katexsupport.gatsbyjs.io/en/contributing/translation-program/translators-guide/#formats) where `Symbol placement: $100 or 100$` was causing $100 or 100$ to render, and needed to change to `Symbol placement: \$100 or 100\$` which allows the dollar signs to render: 
image

nhsz:
> Right, this is something we would need to be aware of with content going forward, but you should be able to escape them with a `\$`...
> 
> I made this adjustment in [translators-guide/index.md](https://ethereumorgwebsitedev01-katexsupport.gatsbyjs.io/en/contributing/translation-program/translators-guide/#formats) where `Symbol placement: $100 or 100$` was causing 100or100 to render, and needed to change to `Symbol placement: \$100 or 100\$` which allows the dollar signs to render: image

the thing is we can't expect other users (e.g. blog editors, publishers or here external content contributions) to be aware and remember that detail (using `\`) when writing content, currently they expect that to be rendered as text, not latex and that shouldn't change (this is what happened in the blog, most people don't know we support latex there), ideally should be something we could catch automatically for those cases, the hard part is to differentiate `$latex exp$` from `$100` in content

Inatiye0911905991 (approved):

wackerow:
For sure... We could try to build something in to check for this.

FWIW, when doing some regex searching on the site, the above example of `$100 or 100$` was the only one on the entire site that accidentally rendered as math. Also, many markdown editors such as HackMD, VSCode, Obsidian, and GitHub itself, already support this when rendering markdown.

If we need to, we could look into automated protections against this, but in either case I think supporting latex brings us closer to what users are used having available when working with markdown.

(todo)
- [ ] Update above string in intl files as well

wackerow:
One thing worth noting here though is the challenge with mobile responsiveness, as longer expressions really aren't. Can look into this more, but in the meantime just placed an `overflowX: scroll` on it, but it doesn't really provide an indication that there is more to view when it overflows.

Might be a plugin out there to rearrange the expressions, or perhaps we could add logic to scale the whole expression down to fit.

wackerow:
image

Noted one more that surfaced when looking at the intl files, the Catalan DeFi page: `- Veneu X $asset al mercat B per 1,10$ `

For it to trigger, there needs to be a `$` at the end of something, followed by a space, and also having a `$` earlier on the same line, without other elements breaking them up (a $ inside a link with a $ following a link will not match together). Escaping any trailing `$` that are followed by a space seems to be enough to avoid rendering math.

src/content/roadmap/merge/issuance/index.md
```diff
@@ -101,31 +101,50 @@ On top of the fee burn implemented by the London upgrade, validators can also in
...
-- `(1 block / 12 seconds) * (60 seconds/minute) = 5 blocks/minute`
-- `(5 blocks/minute) * (60 minutes/hour) = 300 blocks/hour`
-- `(300 blocks/hour) * (24 hours/day) = 7200 blocks/day`
+$$
+{1\space block \over 12\sec} \times 60 {\sec \over \min} = 5 {blocks \over \min} \\
+\downarrow \\
+5 {blocks \over \min} \times 60 {\min \over hour} = 300 {blocks \over hour} \\
+\downarrow \\
+300 {blocks \over hour} \times 24 {hours \over day} = 7200 {blocks \over day}
+$$
```
minimalsm, line 114: This seems horrific for translations... Is there a simpler way to do this?

cc: @lukassim

wackerow: Hm... yeah.

Not sure I have a great solution here... but feel like I'd still lean towards trying to implement. @lukassim Do you have any idea if it'd be possible to block all text in these segments aside from things like a unit, ie "sec" "min" "day" "blocks", etc?

lukassim: Not really, we can hide individual strings, but not parts of strings.

Having taking a look at this in Crowdin, I think this would be very confusing for translators and can see a lot of issues and incorrect translations.

A good enough solution could be to implement this in English but not other languages, though we don't really have a good way of doing that...

Screenshot 2023-07-20 at 16 12 54

wackerow:
Personally still think this would be a nice feature to have available, but sounds like we don't quite have consensus on whether this is worth it or not given complications with translations. Will go ahead and close this for now... perhaps we can re-address after the Next.js migration and after we've made further progress with Crowdin automation.

PR #10402, opened 2023-06-07
Import Japanese updates

wackerow:
## Description
- Import latest from Crowdin for Japanese

## Related Issue
Ongoing translation efforts

PR #10406, opened 2023-06-08
Add check for broken frontmatter images in translated files to markdown checker

minimalsm:
## Description

- Creates a warning if a file in the /translations/ path has an invalid frontmatter link (based on hardcoded relative path, should not be two or three levels)

## Related Issue

#10402


minimalsm:
I've retested this and feel good about pulling it :) merging.

PR #10412, opened 2023-06-08
Update JSON keys used for quiz titles on Quiz hub

minimalsm:
## Description

Uses already translated page titles, mostly in common.json, for quiz titles

## Related Issue

#10320


minimalsm:
Ignore. Pulled from dev branch instead of staging, please see #10413 instead.

PR #10414, opened 2023-06-08
Staking: break out EL + CL client diversity

wackerow:
## Description
Initial PR to update staking product indicators for more clarity around client diversity. This breaks the "Diverse Clients" indicator into both "execution client diversity" and "consensus client diversity" (previously only taking consensus layer into account)

This does not really attempt to update the actual definition for this indicator, but instead starts with simply breaking it into two more specific indicators.

The data has been entered as `null` to start for every product. In the meantime we can review, get some feedback, and see which values we could potentially fill in... or decide it we would want to just push it with `null` values and let project submit evidence to update the respective indicator. 

## Related Issue
- #9840 (Addresses, but does not necessarily close)


src/content/staking/pools/index.md
```diff
@@ -50,7 +50,7 @@ Attribute indicators are used below to signal notable strengths or weaknesses a
 There are a variety of options available to help you with your setup. Use the above indicators to help guide you through the tools below.
 
 <InfoBanner emoji="⚠️" isWarning>
-Please note the importance of choosing a service that takes <a href="/developers/docs/nodes-and-clients/client-diversity/">client diversity</a> seriously, as it improves the security of the network, and limits your risk. Services that have evidence of limiting majority client use are marked as <em style="text-transform: uppercase;">"diverse clients."</em>
+Please note the importance of choosing a service that takes <a href="/developers/docs/nodes-and-clients/client-diversity/">client diversity</a> seriously, as it improves the security of the network, and limits your risk. Services that have evidence of limiting majority client use are indicated with <em style="text-transform: uppercase;">"exeuction client diversity"</em> and <em style="text-transform: uppercase;">"consensus client diversity."</em>
```
corwintines, line 53:
```suggestion
Please note the importance of choosing a service that takes <a href="/developers/docs/nodes-and-clients/client-diversity/">client diversity</a> seriously, as it improves the security of the network, and limits your risk. Services that have evidence of limiting majority client use are indicated with <em style="text-transform: uppercase;">" execution client diversity"</em> and <em style="text-transform: uppercase;">"consensus client diversity."</em>
```

wackerow: Thanks for catching... found it in one other place as well, just pushed a patch

corwintines (approved):

PR #10427, opened 2023-06-09
i18n(intl/en): add missing translations for breadcrumb links

TylerAPfledderer:
## Description

Adds translations to the `en/common.json` file for breadcrumb links that were missing them.

Addresses issue where slug chunks were rendering as the link text with hyphens.

This covers all text issues that were discovered.

## Related Issue
N/A


src/intl/en/common.json
```diff
@@ -13,6 +22,8 @@
   "community-hub": "Community hub",
   "community-menu": "Community Menu",
   "contact": "Contact",
+  "content-buckets": "Content Buckets",
```
minimalsm, line 25: We don't translate this page @pettinarip... current breadcrumb implementation forces us to do it this way, which doesn't seem ideal

pettinarip: Not sure if I follow you on this one @minimalsm

Even though we don't translate this page, we still need the English version of it.

If our concern is about having to do the translations for the rest of langs, then I think that that is a problem from Crowdin. Isn't there a way we can specify that we don't want to translate a given string?

Curious to know whether you have something in mind as an ideal solution? perhaps you are thinking of hardcoding the value in English instead of using the i18n lib?

minimalsm: Yeah let's just specify in Crowdin, good idea.

I guess i was thinking that we could just exclude it entirely, but this PR is probably a precursor to an automatic check that should fail if these aren't included, so seems silly to make one exception

wackerow (approved):
Pulling this in, thanks @TylerAPfledderer!

Issue #9134, opened 2023-01-10
Suggest a dapp: Convex


0xKai27:
### Name

Convex

### Description

Convex allows Curve.fi liquidity providers to earn trading fees and claim boosted CRV without locking CRV themselves. Liquidity providers can receive boosted CRV and liquidity mining rewards with minimal effort.

### Logo

![convex-finance-cvx-logo](https://user-images.githubusercontent.com/104422522/211511695-1fd7405a-7ad4-4bd2-8ba2-41f1efb2b3da.png)

### Brand color

333333

### Is your dapp open source?

Yes

### If open source, what is the repo for the dapp?

https://github.com/convex-eth

### Please describe the measures taken to ensure the dapp's security and provide documentation wherever possible

Please see [Audit docs](https://docs.convexfinance.com/convexfinance/faq/audits) for list of formal audits performed on contracts for Convex Finance.

[Github](https://github.com/convex-eth/platform/tree/main/audit) of audit reports

### Can a user access the dapp without creating an account?

Yes, access is via standardised wallets. Please see screenshot of current [app](https://www.convexfinance.com/stake)

![image](https://user-images.githubusercontent.com/104422522/211508288-3e9c538a-5a42-474c-8ba4-c84c0f3190ee.png)

### Are dapp funds non-custodial?

Smart contract addresses can be found [here](https://docs.convexfinance.com/convexfinance/faq/contract-addresses)

[`Booster.sol`](https://github.com/convex-eth/platform/blob/main/contracts/contracts/Booster.sol) defines a `withdraw()` function which is publicly callable

### Do you offer community support?

Discord: https://discord.com/invite/TTEVTqY488

### Describe any geographical restrictions on usage of your dapp

N/A

0xKai27:
Relates to #9129

corwintines:
@0xKai27 what category do you see this under?

Also, would you like to lead adding Convex?

I think this project is fine to add.

PR #10452, opened 2023-06-13
Add convex to lending

corwintines:
## Description

- Add Convex app data

## Related Issue

Fixes https://github.com/ethereum/ethereum-org-website/issues/9134


minimalsm (changes requested):
We should optimise image here. Don't need 50kb for this.

Kinda questioning the utility of adding a dapp that is specific to Curve, but not a hill I'd die on. Another discussion for another time :)

src/intl/en/page-dapps.json
```diff
@@ -212,6 +212,8 @@
   "page-dapps-more-on-nft-gaming-button": "More on tokenised in-game items",
   "page-dapps-dapp-description-pwn": "Easy loans backed by any token or NFTs on Ethereum.",
   "page-dapps-pwn-image-alt": "PWN logo",
+  "page-dapps-dapp-description-convex": "Convex allows Curve.fi liquidity providers to earn trading fees and claim boosted CRV without locking CRV themselves. Liquidity providers can receive boosted CRV and liquidity mining rewards with minimal effort.",
```
minimalsm, line 215: Too long.

```suggestion
  "page-dapps-dapp-description-convex": "Convex allows Curve liquidity providers to earn trading fees and claim boosted CRV without locking their CRV.",
```

minimalsm (approved):

Issue #9953, opened 2023-04-12
Migrate `ButtonDropdown` to Chakra


pettinarip:
*This is part of our [2nd wave of migration](https://github.com/ethereum/ethereum-org-website/issues/8632) of the [Implement UI library](https://github.com/ethereum/ethereum-org-website/issues/6374) epic.*

**Things to consider**

- Replace the current styled components using the [Chakra Menu](https://chakra-ui.com/docs/components/menu/usage) component

Check our [epic](https://github.com/ethereum/ethereum-org-website/issues/6374) & the [migration guide](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/chakra-migration-guide.md) for more information.

PR #10462, opened 2023-06-14
Migrate `ButtonDropdown` component

pettinarip:
## Description

`ButtonDropdown` migration to Chakra UI using the [Menu](https://chakra-ui.com/docs/components/menu/usage) component.

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/8632
https://github.com/ethereum/ethereum-org-website/issues/9953


minimalsm (changes requested):

src/@chakra-ui/gatsby-plugin/components/Menu.ts
```diff
@@ -0,0 +1,38 @@
...
+import { menuDefaultTheme, defineMergeStyles } from "./components.utils"
+
+const { baseStyle: defaultBaseStyle } = menuDefaultTheme
+
+const { defineMultiStyleConfig, definePartsStyle } =
+  createMultiStyleConfigHelpers(menuAnatomy.keys)
+
+const baseStyle = definePartsStyle(
+  defineMergeStyles(defaultBaseStyle, {
+    list: {
```
pettinarip, line 12: @TylerAPfledderer do you know how can I override the button's styles?

I tried doing
```tsx
button: {
  _active: { bg: "transparent" },
},
```
but didn't work. I still see the default button active styles. In the `ButtonDropdown` implementation I'm using the `as` prop to pass the `Button` to the `MenuButton`.

TylerAPfledderer: Easiest solution would be to apply the style directly on the component so it can override via compiling than CSS specificity.
```tsx
<MenuButton
  as={Button}
  variant="outline"
  leftIcon={<MdMenu />}
  _active={{ bg: "transparent" }}
  {...rest}
>
```

I would note that the `Menu` component will also be used in the main navigation (see #10075), so do not want to bleed unnecessary styles for ButtonDropdown into that UI. Variants can also be unnecessary if certain styles are more for exceptions (or edge-cases) than truly for an intended variant.

However, the outline variant for the `Button` would not have a bg color for `_active` in #10315 so maybe this would no longer be a concern? 😅

sscornelius:
great,

src/@chakra-ui/gatsby-plugin/components/Menu.ts
pettinarip: Thanks @TylerAPfledderer 

> Easiest solution would be to apply the style directly on the component to ensure priority in compiling.

I tried that but the problem is that that overrides _all_ the active styles from the Button. I'd need to re-declare all the existing styles from the theme first, right? anyway I agree that would be the cleanest way to go.

> I would note that the Menu component will also be used in the main navigation (see https://github.com/ethereum/ethereum-org-website/pull/10075), so do not want to bleed unnecessary styles for ButtonDropdown into that UI.

Great point! Didn't think about that. I'll move these styles then, from the theme to the `ButtonDropdown` component to just use them in there.

> The outline variant for the Button would not have a bg color for _active in https://github.com/ethereum/ethereum-org-website/pull/10315 so maybe this would no longer be a concern?

Another good point. I'll try to get that PR merged first.

TylerAPfledderer: > I tried that but the problem is that that overrides all the active styles from the Button. I'd need to re-declare all the existing styles from the theme first, right?

@pettinarip Indeed that will happen. 😅 There are a couple of possible workarounds, but they would not be clean if the `Menu` component is not going to be using an outline variant for its button any place else in the page.

wackerow (approved):
LGTM! This brings keyboard support for these menus ❤️ (arrows and escape instead of just tab)... An a11y win

PR #10473, opened 2023-06-16
Update index.md with Dwellir details

benjichat:
## Description

We have updated the documentation with our company's information

## Related Issue

There is no issue per se. We are simply another high level node provider


corwintines:
Looking at this in comparison to other products, this site is lacking a lot of necessary documentation on how to use the services. If this gets added, feel free to reopen for another look.

PR #10490, opened 2023-06-19
refactor(Trilemma): update Chakra implementations

TylerAPfledderer:
## Description

Updates usage of Chakra UI for the `Trilemma` component.

- Replace an instance of `Flex` with `Stack`
  - This is minor. `Stack` should be the preferred choice over `Flex` if the default `flexDirection` should be `column` if there is no expectation in changing to `row` at a breakpoint.
- Apply `forwardRef` to the `CircleSelect` component for passing event props.
- Remove `children` type signature from svg `Text` component as it is already present in `HTMLChakraProps`
- Move spacing props higher up to a parent or ancestor for more streamlined control.
- Remove unneeded props from `DrawerContent` and the child `Card`
- Use string values for `hideBelow` and `hideFrom` as the compiled `lg` breakpoint values renders correctly.
- For the `Stack` component wrapping the text content, remove the base `flex` prop value as it does not need to be declared for the breakpoint object syntax to compile correctly.

## Preview link
- https://deploy-preview-10490--ethereumorg.netlify.app/en/roadmap/vision

## Related Issue
N/A


F3n4nd0gm:
Hello

F3n4nd0gm:
http://192.168.1.254

F3n4nd0gm:
https://coinmarketcap.com/

src/components/Trilemma/index.tsx
```diff
@@ -32,22 +32,20 @@ const Trilemma = () => {
...
 
-  // TODO: Replace with direct token implementation after UI migration is completed
-  const lgBp = useToken("breakpoints", "lg")
-
   return (
     <Flex
-      alignItems="flex-start"
+      alignItems="center"
       flexDirection={{ base: "column", lg: "row" }}
-      justifyContent="space-between"
```
wackerow, line 42: Are we looking to change the layout here, or match production? Noticing this causes a change in how this component lays out on larger screens.

src/components/Trilemma/index.tsx
```diff
@@ -32,22 +32,20 @@ const Trilemma = () => {
...
+      gap={12}
+      py={12}
+      pl={{ lg: 12 }}
+      pr={{ lg: 32 }}
```
wackerow, line 42: If we're adding this here, we'll need to also remove the margins on `<chakra.svg` inside `Triangle.tsx` (L115-117).. we shouldn't need `mt` since we're aligning `center`, and we shouldn't need `me` since we handle this here. I think we still need `my` with how it's set up to keep the height within reason on mobile.

We should make these RTL-friendly though:
```suggestion
      ps={{ lg: 12 }}
      pe={{ lg: 32 }}
```

src/components/Trilemma/index.tsx
```diff
@@ -32,22 +32,20 @@ const Trilemma = () => {
...
     >
-      <Flex
-        flexDirection="column"
-        my={{ base: 8, md: 12 }}
-        mx={{ md: 12 }}
-        py={8}
+      <Stack
+        gap={8}
```
wackerow, line 45: Do we want a gap here? We're using a text component with margin-bottom on it already

image

```suggestion
        gap={0}
```

wackerow (changes requested):
Hey @TylerAPfledderer, solid work here... gave it a run through and looks good, just a couple notes. From the PR description didn't quite sound like layout changes were part of this, so just left a couple notes of where it looked off compared to production and how I think we can get it closer. 

With those few adjustments I think we can get this in... I'll circle back on this and if you're too busy I'm happy to patch it up

src/components/Trilemma/index.tsx
TylerAPfledderer: Applied RTL-aware props.

At the very least, top margin should be considered, as even with larger screens, there are certain points where the text wrapping expands there container height be too close to the bounds of the section container compared to how much spacing there is consistently through the breakpoints. center alignment primary helps the triangle svg though.

Example is at 996px without top margin:
image

src/components/Trilemma/index.tsx
TylerAPfledderer: I recognized the necessity of the prop and added it back in.

src/components/Trilemma/index.tsx
TylerAPfledderer: I updated the structure and spacing to resolve this.

I brought in the updated `Heading` and `Text` components.

I also updated the structure where it is a top-down approach to how spacing is handled, where the parent primarily sets the spacing between its children, and the children set spacing exceptions depending on the breakpoint. (There is a little bit of that with the text content area from the mobile screen size).

So with what you highlighted above, the paragraphs (with the new `Text` component that provides no margin) are in their own `Flex` which defines the gap, which is set to `24px`. This is different that the spacing from the heading and these paragraphs, where the parent `Stack` which sets its gap (`spacing`) to `32px`.

src/components/Trilemma/index.tsx
```diff
@@ -32,55 +32,46 @@ const Trilemma = () => {
...
-            background="background.base"
-          >
-            <Card
-              {...cardDetail}
-              background="none"
-              border="none"
-              justifyContent="flex-start"
-              my={8}
-            />
+          <DrawerContent borderTopRadius="1rem" background="background.base">
```
wackerow, line 73: Could we use `borderTopRadius="2xl"` here?

image

```suggestion
          <DrawerContent borderTopRadius="2xl" background="background.base">
```

src/components/Trilemma/index.tsx
```diff
@@ -32,55 +32,46 @@ const Trilemma = () => {
...
+        </Heading>
+        <Flex flexDir="column" gap="6">
+          <Text>{t("page-roadmap-vision-trilemma-p")}</Text>
+          <Text>{t("page-roadmap-vision-trilemma-p-1")}</Text>
+          <Text>{t("page-roadmap-vision-trilemma-p-2")}</Text>
+          <Text fontWeight={600} hideFrom="lg">
+            {t("page-roadmap-vision-trilemma-modal-tip")}:
+          </Text>
+        </Flex>
+        <Card {...cardDetail} minH="300px" hideBelow="lg" />
```
wackerow, line 64: https://github.com/ethereum/ethereum-org-website/assets/54227730/70d48554-bf52-4a7f-b68a-d70a9ab56223

Only small thing I'm noticing is the smaller spacing now between the last paragraph and this card... mt="6" on this element should bring it to parity with production:

```suggestion
        <Card {...cardDetail} minH="300px" hideBelow="lg" mt="6" />
```

@nloureiro Will let you weigh in here if you have a preference

src/components/Trilemma/index.tsx
```diff
@@ -32,55 +32,46 @@ const Trilemma = () => {
...
+        mb={{ lg: "20" }}
+        ms={{ md: "12" }}
+        me={{ md: "12", lg: 0 }}
         flex={{
-          base: "1 1 100%",
           lg: "0 1 500px",
         }}
       >
-        <OldHeading fontSize="2rem" mt={0}>
+        <Heading fontSize="2rem">
```
wackerow, line 53: @nloureiro Brought this up elsewhere, but more use of a non-standard font size

image

^ These are the sizes we have assigned to tokens right now... not requesting anything for this PR, just noting... [card here](https://www.notion.so/efdn/Review-use-of-non-standard-font-sizes-713321312cd04b0cbd8d28eb37ebfe5b?pvs=4)

wackerow (approved):
Nice job @TylerAPfledderer! Approving this from a dev perspective but would like to have a designer approval before finally pulling this in. cc: @nloureiro @konopkja 

Not positive we can use "2xl" for a `borderTopRadius` but I'm fairly certain we can... if you agree, feel free to commit that suggestion, but not a blocker.

---

## TODO:
- [ ] Designer approval

src/components/Trilemma/index.tsx
TylerAPfledderer: Since the parent container is a flexbox, why not `alignItems="center"` instead? It looks like that was the point of using margin in prod. (The right margin is still the same as prod)
image

And then noting in the mobile view the stacked content doesn't have all this extra margin below the triangle (compared to prod)
image

wackerow: This looked fine imo, I was referring to the margin/spacing here:

image

nloureiro (approved):

src/components/Trilemma/index.tsx
TylerAPfledderer: @wackerow oh! I will review that spacing

wackerow (approved):

wackerow:
Thanks again @TylerAPfledderer!!

PR #10493, opened 2023-06-19
chore(deps): bump Chakra, Storybook, and Chromatic deps

TylerAPfledderer:
## Description
Bumps Chakra UI, Storybook, Chromatic and related dependencies to latest.

This will also remove the console warning for an out-of-date version of `@chakra-ui/storybook-addon` as it is now configured for Storybook v7

There are no known breaking changes.

## Related Issue
N/A


pettinarip (commented):
@TylerAPfledderer is it me or ~3.9k line changes in the `yarn.lock` seems too much?

For reviewing processes I think it would be worth updating the versions in the `package.json` as well, to clearly see what packages are we upgrading and what we are not. For example, its difficult to know if we upgrading just the Chakra react lib and cli, or just the react one. Same with SB and its addons.

yarn.lock
```diff
@@ -2,133 +2,141 @@
 # yarn lockfile v1
 
 
-"@algolia/autocomplete-core@1.7.4":
-  version "1.7.4"
-  resolved "https://registry.yarnpkg.com/@algolia/autocomplete-core/-/autocomplete-core-1.7.4.tgz#85ff36b2673654a393c8c505345eaedd6eaa4f70"
-  integrity sha512-daoLpQ3ps/VTMRZDEBfU8ixXd+amZcNJ4QSP3IERGyzqnL5Ch8uSRFt/4G8pUvW9c3o6GA4vtVv4I4lmnkdXyg==
+"@algolia/autocomplete-core@1.9.3":
```
pettinarip, line 5: I'll rerun this install bc I see that we are upgrading everything.

corwintines:
Screen Shot 2023-06-29 at 11 23 25 AM

Getting this build error when looking into this locally. Will continue to look, but calling it out in case either either @pettinarip or @TylerAPfledderer have any ideas

pettinarip:
> Getting this build error when looking into this locally. Will continue to look, but calling it out in case either either @pettinarip or @TylerAPfledderer have any ideas

Yea, I'll take a look now. It has to be related to some weird conflict after the yarn.lock generation.

pettinarip (commented):
@TylerAPfledderer had to downgrade `@chakra-ui/react` to v2.6.1 because v2.7.0 generates the following error. Not sure what is going on.

```
 ERROR  UNKNOWN

Cannot read properties of undefined (reading '__cssMap')



  TypeError: Cannot read properties of undefined (reading '__cssMap')
  
  - index.mjs:1703 getCSSVar
    [ethereum-org-website]/[@chakra-ui]/styled-system/dist/index.mjs:1703:1

```

TylerAPfledderer:
> @TylerAPfledderer I had to downgrade `@chakra-ui/react` to v2.6.1 because v2.7.0 generates the following error. Not sure what is going on.
> 
> ```
>  ERROR  UNKNOWN
> 
> Cannot read properties of undefined (reading '__cssMap')
> 
> 
> 
>   TypeError: Cannot read properties of undefined (reading '__cssMap')
>   
>   - index.mjs:1703 getCSSVar
>     [ethereum-org-website]/[@chakra-ui]/styled-system/dist/index.mjs:1703:1
> ```

@pettinarip I've double-checked a couple of things, along with updating this branch with the latest `dev`, and still no good. I've sent this to the team to see if I can get some insight.

It looks like this issue is only occurring with the `Input` theme, and specifically the use of the default outline variant object, even though everything looks fine with everything else like the `Button` theme that has a similar implementation. 🤔

pettinarip:
> @pettinarip I've double-checked a couple of things, along with updating this branch with the latest `dev`, and still no good. I've sent this to the team to see if I can get some insight.
> 
> It looks like this issue is only occurring with the `Input` theme, and specifically the use of the default outline variant object, even though everything looks fine with everything else like the `Button` theme that has a similar implementation. thinking

Ok, thanks for follow that up. I'll update this branch again and merge it with v2.6.1 for now. Once we figure out a solution we can bump again the chakra dep.

PR #10516, opened 2023-06-21
correct RandAO VRF explanation

### LINKED ISSUE #10135: Bug report
### Describe the bug

The explanation of `randao_reveal` in *Block proposal* is confusing and probably wrong.

> The `randao_reveal` field takes a verifiable random value that the block proposer creates by adding some of its own entropy to the accumulated RANDAO value from previous blocks.

It is a BLS signature of the epoch number. The overall goal is to mix in entropy from the private key. The explanation suggests

### To reproduce

1. Go to https://ethereum.org/en/developers/docs/consensus-mechanisms/pos/block-proposal/#how-is-a-block-created

### Expected behavior

The explanation should not cause the user to look for [additional resources](https://inevitableeth.com/home/ethereum/network/consensus/randao) in order to know that it's actually a BLS sig of the epoch number.

### Screenshots

_No response_

_No response_

_No response_

### Additional context

_No response_

- [X] Yes
- [ ] No

lourkeur:
## Description

As discussed in the issue, the sentence did not accurately describe the value in the RandAO field.

This addresses the why but not the why. Readers might ask why a BLS Signature is a good source of randomness.

## Related Issue

Closes #10135


jmcook1186 (approved):
lgtm - thanks for the PR!

PR #10520, opened 2023-06-21
Randomize exchange and wallet listing order on /get-eth

wackerow:
## Description
Reverts alphabetical order on exchange/wallet listings on /get-eth page, using random sort

### DIFF
--- a/src/components/EthExchanges/use-eth-exchanges.ts
+++ b/src/components/EthExchanges/use-eth-exchanges.ts
@@ -1,4 +1,5 @@
 import { useState } from "react"
+import { shuffle } from "lodash"
 import { useStaticQuery, graphql } from "gatsby"
 import { IGatsbyImageData } from "gatsby-plugin-image"
 import { useI18next, useTranslation } from "gatsby-plugin-react-i18next"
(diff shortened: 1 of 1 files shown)


Inatiye0911905991:
> ## Description
> Reverts alphabetical order on exchange/wallet listings on /get-eth page, using random sort

0x51370D6c01B997c24e806fd1D14cB2b2D658c092

src/components/EthExchanges/use-eth-exchanges.ts
```diff
@@ -492,16 +493,15 @@ export const useEthExchanges = () => {
...
-    data.exchangesByCountry.nodes
-      .map((node) => {
-        return {
-          value: node.country,
-          label: node.country,
-          exchanges: node,
-        }
-      })
-      .sort((a, b) => a.value.localeCompare(b.value))
+  const exchangesByCountry: Array<ExchangeByCountry> = shuffle(
```
corwintines, line 496: I think we should revert this since it shuffles the countries in the list. I think this part should remain alphabetical. But, everything else is good and shuffles the exchanges and wallets which I think is good. 

Screen Shot 2023-06-29 at 8 46 30 AM

minimalsm: @wackerow 👆

corwintines: Pushed up this change after talking to @wackerow

corwintines (approved):

PR #10532, opened 2023-06-23
Update index.md (old web3 python API was being used)

scrovy:
The web3 python library API has been updated and these calls were not working:
- `to_checksum_address` instead of `toChecksumAddress`
- `get_logs` instead of `getLogs`
- `block_number` instead of `blockNumber`
- `w3.keccak` instead of `w3.sha3`

Also cryptokitties does not have many transactions anymore. Recommended to increase the number backward searching blocks up from 120 if no events are found.

Added condition `recent_tx` not empty, since `recent_tx[0]` is accessed and this does not always exists (if no event is found)

### DIFF
--- a/src/content/developers/docs/standards/tokens/erc-721/index.md
+++ b/src/content/developers/docs/standards/tokens/erc-721/index.md
@@ -141,7 +141,7 @@ ck_extra_abi = [
     }
 ]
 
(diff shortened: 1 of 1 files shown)


jmcook1186 (approved):
thanks for the updates!

PR #10534, opened 2023-06-23
Migrate `run a node` page

pettinarip:
## Description

Migrates the Run a node page to Chakra UI.

## Related Issue

#9353


wackerow (approved):
image

I assume this is just part of adopting the design system, just making sure it's intentional we're removing this haze effect.

image

Also, I noticed the `<li>` indentation is reduced here... was that intentional?

Neither are blockers in my opinion, just wanted to check, otherwise looks good!

pettinarip:
@wackerow good catches. No, they were not intentional. I'll take a look at those and fix them. Ty

PR #10542, opened 2023-06-25
feat: build custom react-select component with Chakra

TylerAPfledderer:
## Description
_Part of Implementation of new DS v1 #9548_

This PR update the usage of the `react-select` package.

- Bumps the `react-select` dep to latest which includes TypeScript support
- Creates a custom select component, including custom internal parts using Chakra components.
- Create a custom theming config `ReactSelect` to be sent to the internal parts
- Update implementation with related page components.
- Update onChange handler functions to match new TypeScript support.


sammyfilly (commented):
build web

sammyfilly (approved):

src/components/EthExchanges/index.tsx
```diff
@@ -96,14 +90,15 @@ const EthExchanges = () => {
...
-      />
+      <Box w="full" maxW="container.sm">
+        <ReactSelect
+          aria-label={t("page-get-eth-exchanges-header")}
+          options={exchangesByCountry}
+          onChange={handleSelectChange}
+          placeholder={placeholderString}
+          variant="outline"
+        />
+      </Box>
```
TylerAPfledderer, line 101: This is the only instance where it was originally a combobox. Should it still be a combobox?

sammyfilly:
Can it be anything else other than the combo box

src/components/EthExchanges/index.tsx
pettinarip: By combobox you are referring to a dropdown where the user can type a value and get the list filtered? if that is the case, then yes, we should support that here, as it is one of the key things to maintain.

nloureiro:
All look great!
One remark that helps separate the group options from the others the line that divides the group needs to in all group separations. Probably this wasn't clear on Figma, sorry.

Screen Shot 2023-06-29 11 20 16 AM

pettinarip:
> All look great! One remark that helps separate the group options from the others the line that divides the group needs to in all group separations. Probably this wasn't clear on Figma, sorry.

@TylerAPfledderer just chatted about this with Nuno, we are good with this. We are displaying both lines. No change request there.

pettinarip (changes requested):
Marking this as blocked until we implement the combobox features on the EthExchange dropdown. Its the only remaining thing here.

TylerAPfledderer:
> Marking this as blocked until we implement the combobox feature on the EthExchange dropdown. Its the only remaining thing here.

This is now pushed.

Because there is only one known instance of a combobox, I would have `isSearchable` default to `false`, and then pass the prop in this instance to enable it.

I did not see a negative impact here on styling.

F3n4nd0gm (changes requested):
https://ethereum.org/es/

F3n4nd0gm:
Is virus en mi pc

TylerAPfledderer:
Closing this PR as it is very outdated. Will be creating a new PR to account for updates to environment and logic.

PR #10545, opened 2023-06-25
Fix spelling in proto-danksharding explanation

dhvogel:
## Description

This is a small spelling correction for the proto-danksharding explanation.


jmcook1186 (approved):
thanks!

PR #10551, opened 2023-06-26
Support to translatable image in mdx files

pettinarip:
Before this change, for every md file that was using a local image, we were updating those image paths to point to the corresponding English versions.

## Description

This PR modifies the gatsby plugin to support files that are colocated in the translation files. If the file exists in the given language folder then it will use that file instead of using the English one.

## Example
- web 3.0 image - EN https://ethereumorgwebsitedev01-translatableimagemdxsupport.gatsbyjs.io/en/web3/#web3
- web 3.0 image - ES https://ethereumorgwebsitedev01-translatableimagemdxsupport.gatsbyjs.io/es/web3/#web3


corwintines (approved):
Nice

PR #10552, opened 2023-06-27
Ethpool Permissionless

Buttaa:
## Description

Changes Ethpool's permissionless description to "true". 
The discussion has occurred in: https://github.com/ethereum/ethereum-org-website/pull/10170#issuecomment-1542323800

### DIFF
--- a/src/data/staking-products.json
+++ b/src/data/staking-products.json
@@ -465,7 +465,7 @@
       "isFoss": false,
       "hasBugBounty": false,
       "isTrustless": false,
-      "isPermissionless": false,
+      "isPermissionless": true,
       "pctMajorityExecutionClient": 0,
       "pctMajorityConsensusClient": 0,
       "isSelfCustody": true,


wackerow:
Hey @Buttaa, sorry I hadn't replied to the comment you linked to... I'm not sure that was resolved though. I see your arguments about most SaaS providers, but looking at Abyss as an example, they appear to use contracts to mediate everything, along with their own token. Users don't have any way of being blocked or turned down, and that was the point of the indicator. 

Looking at other products, realizing that [Sensei Node](https://www.senseinode.com/) is another marked as permissionless, when it really shouldn't be. Their website directs you straight to a "contact us" form... likely a mistake on my end, which I'll patch now that I've noticed it.

Does that make sense? We could debate if that type of indicator is valuable for the category of course, but with it's current definition I'm not sure the toggle for Ethpool should be flipped.

PR #10571, opened 2023-06-28
clarifies weak vs strong statelessness on roadmap

jmcook1186:
Based on feedback from Dankrad, this PR clarifies that string statelessness is not intended to be part of the protocol roadmap because weak statelessness is thought to be sufficient.


corwintines (approved):
LGTM. Will wait to see if there are any changes suggested before next release candidate.

Issue #10553, opened 2023-06-27
Change button style to our default main button


konopkja:
### Describe the bug

On https://ethereum.org/en/run-a-node/ there are custom style buttons that were supposed to match the branding of dappnode and avado, but since then both brandings changed and I think its better to keep there our button style instead anyway.

### To reproduce

1. Go to https://ethereum.org/en/run-a-node/
2. scroll down to choose your adventure

### Expected behavior

consistent button style usage with our main CTA style
<img width="619" alt="Screenshot 2023-06-27 at 19 06 11" src="https://github.com/ethereum/ethereum-org-website/assets/100724231/b07757e6-4f6f-4078-95a5-005f37e290c9">

### Screenshots

![Screenshot 2023-06-27 at 17 49 49](https://github.com/ethereum/ethereum-org-website/assets/100724231/8750e1cb-14d5-498e-80ff-e30c30645c8d)

### Would you like to work on this issue?

- [ ] Yes
- [X] No

hashnj:
hello, i would like to contribute to resolve this issue ,i am new to open source and git hub
please also guide me a bit

ItIsOHM:
Hello @konopkja, can I work on this issue too if there are any pending tasks other than the ones assigned to @hashnj?

M-A-Sattar:
Hi, I'm currently learning blockchain development, and would like to try and fix this issue, I'm very new to open source, can I participate?

PR #10574, opened 2023-06-28
Button style changed: Issue ethereum#10553

hashnj:
## Description

i have changed the style of buttons, the color to default and also the color on hover. 
please let me know if it is meeting your requirements.

Fixes: https://github.com/ethereum/ethereum-org-website/issues/10553#issuecomment-1610217271


corwintines (approved):
Thanks @hashnj 

Since we are just going to use the default button, it makes more sense to just refactor our the DappNodeButtonLink and AvadoButtonLink in favor of the default ButtonLink component. Ive pushed up these changes so that we are just using the ButtonLink component now.

PR #10606, opened 2023-07-04
feat: Update Text and Heading Themes for New DS

TylerAPfledderer:
## Description

Update theming of the `Heading` and `Text` components with stories.

## Related Issue
_Part of the Implentation of the DS v1 #9548_


TylerAPfledderer:
@nloureiro Please view the Chromatic UI Tests check and [storybook](https://63b7ea99632763723c7f4d6b-jfgtvscucc.chromatic.com/).

One thing outstanding on here: should a margin bottom be a part of the `Heading` styling -- to be specified for each size in the theme -- or be kept separate and only be added as the `Heading` is used?

pettinarip:
> @nloureiro Please view the Chromatic UI Tests check and [storybook](https://63b7ea99632763723c7f4d6b-jfgtvscucc.chromatic.com/). Of course, this is considered a priority to be able to properly work on other components that need it.
> 
> One thing outstanding on here: should a margin bottom be a part of the `Heading` styling -- to be specified for each size in the theme -- or be kept separate and only be added as the `Heading` is used?

Interesting point @TylerAPfledderer. IMO for this first iteration we should not add those margins to the Heading component.

Later, as we migrate the different layouts or pages (previously defined in the DS), we are going to start detecting patterns with these margins, and we should decide then if we add them to the Headings or if we do something else, like using `textStyles` (I'd go with adding them to the Headings and then overriding the other cases were we are not using the default margins).

src/@chakra-ui/gatsby-plugin/components/Heading.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Instead of rewriting the entire sizes object, take the existing value from the
+ * default theme and replace the lineHeight values.
+ */
+const sizes = Object.entries(defaultSizes || {}).reduceRight(
+  (acc, [key, value]) => {
+    return {
+      ...acc,
+      [key]: defineStyle({
+        ...value,
+        lineHeight: lineHeightScale[key],
```
pettinarip, line 27: Trying to understand how this mapping works @TylerAPfledderer 

On the Chakra default theme I see these possible values for `lineHeight`

```
lineHeights: {
    normal: "normal",
    none: 1,
    shorter: 1.25,
    short: 1.375,
    base: 1.5,
    tall: 1.625,
    taller: "2",
    "3": ".75rem",
    "4": "1rem",
    "5": "1.25rem",
    "6": "1.5rem",
    "7": "1.75rem",
    "8": "2rem",
    "9": "2.25rem",
    "10": "2.5rem",
  }
```

but in the mapping defined previously, we are using for example, `"xs"` for the `md` size. Where is this `xs` coming from in this case?

pettinarip: Ah! I see now that we have defined them on the `typography.ts` file, got it now! :)

src/@chakra-ui/gatsby-plugin/styles.ts
```diff
@@ -17,16 +15,18 @@ const styles = {
...
+        mb: 0,
+      },
     },
     a: {
       color: "primary.base",
       textDecoration: "underline",
     },
-    // should be replace with https://chakra-ui.com/docs/components/text
-    p: {
-      margin: "0px 0px 1.45rem",
```
pettinarip, line 28: Now, looking at the preview deploy, I see that we kind of depend on this margin to avoid having to refactor all the codebase. I'd say that for the `Text` component we add this button margin. What do you think @nloureiro and @TylerAPfledderer?

Note for context for @nloureiro. This `Text` component represents the `p` tag, the paragraphs in html.

TylerAPfledderer: @pettinarip fair point!

So for example, with the Homepage hero where this margin would be used against the CTA button "Explore Ethereum", when working on the new heroes, I took a different approach, whereby I structured the Header and text under one flex stack, and then that grouping with the button at a higher level stack, and used the `gap` or `spacing` prop to define the "margin" between the different texts and between the paragraph and the button.

I should ask then what would best practice be here? In my description above, this approach using the Flexbox `gap` property to define the margins between content instead of using the `margin` prop.

And I would consider in addition to the above the idea of using `margin` only when we are stacking paragraphs, with a defined :first- or :last-of-type selector so that only margin is defined between the paragraphs, but not added above the first or below the last of the stacked siblings.

So:
```ts
"p:not(:last-of-type)": {
  mb: "2" // or some other token value
}
```

I also understand that margin has to be well defined for the mdx file parsing because you have no ability to isolate groups of text elements into neatly packaged boxes. So I would want to figure out what the best practice should be to keep usage in MDX pages inline with the page layouts that do not use MDX.

pettinarip: Using `gap` and `spacing` is correct and expected as the best way to go looking ahead.

Now, talking about the current problem and thinking about how to support the old margins, I think that the best way to go is by leaving the `Text` component as it is and creating a new temporary one called `OldText` that would have the old margins.

Following this idea
- the `Text` component would be used in all new implementation
- the `OldText` would be marked as deprecated and would be used in the rest of the old/current implementations.

The `OldText` component might be something as simple as this
```tsx
function OldText(props) {
  return <Text margin={oldDefaultMargins} {...props} />
}
```

With that new component in place, I could go and refactor the entire codebase by importing the `OldText` and adding those changes to this PR.

Then, as you migrate the new components as the `Hero`, you can start using and importing the native `Text`.

Does it make sense?

TylerAPfledderer: @pettinarip Ok! So using the `OldText` component for the MDX files too, correct?

pettinarip: Correct.

TylerAPfledderer: @pettinarip I've pushed this new component here.

pettinarip: Thanks. I'll refactor the code to use that component instead of the `Text`.

TylerAPfledderer: Sounds good!

src/@chakra-ui/gatsby-plugin/components/Text.ts
```diff
@@ -0,0 +1,48 @@
+import { defineStyle, defineStyleConfig } from "@chakra-ui/react"
+
+const sizes = {
```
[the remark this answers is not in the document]
TylerAPfledderer: @pettinarip to clarify, this is just regarding the `Text` component? It looks like the sizes for the `Heading` component have not changed (which is to be expected since they are currently inline with the Chakra defaults)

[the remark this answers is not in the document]
TylerAPfledderer: Also need clarification regarding the body copy with this update. Marked a question [in the figma file](https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B?node-id=384:15485&mode=dev#520717779) as I found a discrepancy.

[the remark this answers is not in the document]
pettinarip: Correct @TylerAPfledderer. The `Heading` has not suffered any update.

TylerAPfledderer:
> Maybe late to this one. Where should I look to review?
> 
> here? Screen Shot 2023-08-14 03 35 43 PM

@pettinarip looks like Chromatic wasn't triggered with the new stories.

@nloureiro yes to this one as well as `Atoms / Typography / Text`

TylerAPfledderer:
@pettinarip @nloureiro I updated the branch recently, and now the chromatic review checks are now available!

TylerAPfledderer:
@nloureiro @pettinarip this should be good for a review again!

src/components/BaseStories/Heading.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+  ],
+} satisfies Meta<HeadingType>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+const headingScale: Array<HeadingProps> = [
+  {
+    as: "h1",
```
nloureiro, line 36: @TylerAPfledderer @pettinarip maybe this is normal, but having 3 times  `as="h1"` is correct?

nloureiro (approved):

src/components/BaseStories/Heading.stories.tsx
TylerAPfledderer: @nloureiro I could only assume that the three largest sizes would be reserved for the `h1` heading. If that is true, then I would make sure that the story reflected that preference.

pettinarip (approved):
🚢

PR #10609, opened 2023-07-05
Update community-events.json

eenagy:
Update event list

## Description

Add EthChiangMai to the event list


corwintines:
Thanks @eenagy 

I think this is lacking too much information on the event site at the moment. When there is more information for attendees and they can purchase tickets feel free to open a new PR for this.

PR #10616, opened 2023-07-06
Add Translators as contributors on non-English docs/tutorials

minimalsm:
## Description

Adds translators as contributors to docs pages and tutorials

## Behaviour:

- If a page's frontmatter is in English, display GitHub contributors
- If a page's frontmatter is not English, show Crowdin contributors
- In either case, if the page has no contributors, do not show the component.

## Notes:

- Note: Ignore GitGuardian. I burned the API keys that were committed before pushing this PR. cc: @Aldekein 
- I've added an API key under the secret CROWDIN_API_KEY to the repo
- Note to run this locally, you'll need a Crowdin API key.
- GitHub action example run on a fork: https://github.com/minimalsm/ethereum-org-website/pull/11


src/components/FileContributors.tsx
```diff
@@ -101,15 +103,27 @@ const ContributorList = ({ children }: { children: React.ReactNode }) => {
...
+        <Avatar
+          height="40px"
+          width="40px"
+          src={contributor.avatarUrl}
+          name={contributor.name}
+          mr={2}
+        />
+      ) : (
+        <ChakraSkeletonCircle size="10" mr={2} />
+      )}
```
minimalsm, line 126: Not working as expected... 

[Chakra's Avatar](https://chakra-ui.com/docs/components/avatar) has a fallback when the image is loading, but we want to show the skeleton until the user's avatar has been fetched (Maybe @TylerAPfledderer knows?)

TylerAPfledderer: @minimalsm I'm unfortunately unable to take a look at the PR locally. Something with my local does not like trying to show any of the `doc` pages that would render these avatars.

To be clear on intent: the fallback prop is only meant if the fetch returns an error, like the endpoint yields no result or if the url has a typo, etc. The skeleton should disappear when the loading is complete from the fetch regardless of the result. I think this would be good UX, else if the the skeleton stays visible when there is an error in the fetch the user might think there is something wrong with the server or browser, not the specific endpoint.

The best solution here would be to let the `Avatar` component control the loading state. Since the skeleton should disappear regardless of the outcome from the fetch, we can apply `() => setImageLoaded(true)` to both the `onLoaded` and `onError` props of the Avatar.

And to limit some of the styling duplications, the `Skeleton` can wrap the `Avatar` and pass the controlled loading state to `isLoaded`.

Check out this [Sandbox Demo](https://play.chakra-ui.com/playground?code=N4Igxg9gJgpiBcICWBbADhATgFwATFwFcBnGAZWwENsZcBfXAM0whVwHJMZKxt2AdAHaoMOfLjIBrGABsY2CIIA0uAIIA3apUz0mLNuwACYABaVJmSgFpCSAPRcefIUMiDieAMKLsmJACNCBR0AXlwACgI3XwCgrHoASlwQgD58IVxcNw9cAG0kYgBJFEoAcxgAGQhKWCgVUmxissrq2oBdZKJSCmoYcMZKGVIEl0FMrmxCTDHwjMzcAB4UufnFqVl5RSYkbG9BGn3cfyxYTAAlGqQSEP4QRkIZGVvcAqqamCgQ4AKm8rfaujLMareYLDRaTArEGZQSUFAwL7RPyBYIAOlh8LoUOhxwAHmQkAAvBG3ACMAAZbtiQcRMGBET5kXFMKjKJoqJgAKqYGRY4HQzIyVpIQSlG4gGSUQkATyp-IFdjsuAAEmzaEL3lAjrRfIRaFxStooHJiMRcBBGLgYJgWDoRUx5KYRaVcNgTLRUM1qatFP8vuEkqlcA1fi1NeFdTAEnyBfNFABRG1Yf2BtIhkp-VofCOYPXR724OxA6ELOzrOQKQTF0FFuYjQR8oQwXGiPDZPCqNBoToB5JpYBzCZTGZQpbUhZ7GIo+JI2LBL4ENkQ7kyeAcEzYbBoYjwRX+HaomTSuxQSiCKyUfyWFAQdj0Bi1+WZCeMudYAuz6eYBcFzJLjkrmu7AbluO57geR4nmeF5XnCt5KL+uAYjAQEACJnmosE3uo7AIU+8x0DGIKPiWk5MsEH6vl+P74as-7aIB66btuu52Pu2CHsep7npe14QDheGxjCcIoRw6FjKoWH8bhBaEdSJE1tW9Z8iAdBAA)

Note that I'm using `Skeleton` and not `SkeletonCircle`. The `fitContent` prop (which renders the styling `width: fit-content`) will be overridden in the `SkeletonCircle` in favor of a `boxSize` prop styling, which I think is a mistake.

wackerow: > Something with my local does not like trying to show any of the doc pages that would render these avatars.

For those encountering this, make sure `.env` doesn't have `**/docs` for `IGNORE_CONTENT=`... it's there by default in `.env.example` so this is likely common.

minimalsm: Let's just allow this behaviour; it will be the default when we migrate this component in the design system. I'll push a commit to remove this logic.

Inatiye0911905991:
From Ethiopia im sintayew Gashaw

Inatiye0911905991 (approved):

src/constants.ts
```diff
@@ -10,3 +12,18 @@ export const INITIAL_QUIZ = "what-is-ethereum"
...
 export const TOTAL_QUIZ_RETRY_RATE = 15.6
+
+// Crowdin
+export const CROWDIN_PROJECT_ID = 363359
+export const CROWDIN_API_MAX_LIMIT = 500
+export const FIRST_CROWDIN_CONTRIBUTION_DATE = "2019-07-01T00:00:00+00:00"
+export const REGULAR_RATES: ReportsModel.RegularRate[] = [
+  {
+    mode: "tm_match",
+    value: 1.01,
```
minimalsm, line 23: 1.01 because Crowdin API doesn't consider 1.00 to be a positive float 🫠

src/scripts/crowdin/getCrowdinContributors.ts
```diff
@@ -0,0 +1,17 @@
+import getAndSaveDirectories from "./source-files/fetchAndSaveDirectories"
+import getDirectoryIds from "./utils"
+import { getTranslatedMarkdownPaths } from "../markdownChecker"
+import fetchAndSaveFileIds from "./source-files/fetchAndSaveFileIds"
+import getTranslationCostsReport from "./reports/getTranslationCostsReport"
+
+async function main() {
+  await getAndSaveDirectories()
+  const directoryIds = getDirectoryIds()
+  await fetchAndSaveFileIds(directoryIds)
```
minimalsm, line 10: Inconsistent function naming here, bit of a smell but don't think it should block.

corwintines (approved):

src/scripts/crowdin/reports/reportsHelpers.ts
```diff
@@ -0,0 +1,121 @@
...
+  now.setDate(now.getDate() - 1)
+  return now.toISOString()
+}
+
+export async function fetchTranslationCostsReport(
+  fileId: number,
+  crowdinLangCode: string
+): Promise<void> {
+  const now = new Date()
+  now.setDate(now.getDate() - 1) // set the date to one day in the past (yesterday)
```
minimalsm, line 24: This code can be deleted

pettinarip (commented):
Amazing job @minimalsm! 💪🏼 

The only thing I'd call out in this implementation is that we should always try to avoid storing all the `.json` files in the `/data` folder since everything that lives there is going to be loaded in buildtime by Gatsby. And the only necessary file that we should load, should be the `combined-translators.json`.

But given that we are in the transition to Next, this is not needed. Just wanted to mention it to keep that in mind in case we face an issue during the transition.

Anyway, reviewing this helped me understand how we are going to migrate this portion of code to the Next imp.

PR #10667, opened 2023-07-12
add disclaimer banner to find wallets

corwintines:
## Description

- Adds a disclaimer banner to the top of the find wallets page with the same text as at the bottom. 

Reason: Users typically do not reach the bottom of the page to know that these are not endorsements. This will make it more apparent.


wackerow (approved):

wackerow:
Bringing in as a starting point, but think we should chat more about making this more noticeable and standardized across the product listings on the site.

PR #10670, opened 2023-07-13
refactor(local-environment): remove duplicate object

TylerAPfledderer:
## Description

Removes a duplicate object that is not being used in the `developers/local-environment` page.

The object being used in prod should be the correct one (inside the component function block), as it contains converted translations for image alt text.


minimalsm (approved):
Thanks @TylerAPfledderer ❤️

corwintines (approved):

PR #10678, opened 2023-07-13
Update index.md

EvilJordan:
The `withdrawals` header was somehow missing a carriage return, causing the table that follows not to display as markdown

## Description

Added a carriage return to fix markdown issues (that did not appear during preview!)


jmcook1186 (approved):
argh! thanks for fixing!

PR #10706, opened 2023-07-18
feat(Link): update component for new DS

TylerAPfledderer:
## Description

Updates the `Link` component for the new DS

## Related Issue

_Part of the Implementation of the new DS v1 #9548_


pettinarip (changes requested):
Good job @TylerAPfledderer 

I think this PR will need to wait a bit after we implement the other link components and/or override other styles since this change will impact every link we have on the page.

For example, we should implement first the FooterLink component. Right now, we have this:
image

Then we also need to override some styles on the `ButtonLink` component. Currently, we got this:
image

The header links is also impacted by this PR
image

TylerAPfledderer:
@pettinarip pushed style overrides for the areas mentioned above.

pettinarip (changes requested):
A few more places where we need to override:

dev docs
image

dev docs - table of content
image

breadcrumbs
image

table of contents used in the `use-cases` templates
image


======

Will keep testing after we fix these issues.

On the other hand, I'd like to have more eyes on this PR as it impacts the entire website. So, I'll ping the team and wait until someone else approves it.

cc @nloureiro would be nice if you can test this as well.

TylerAPfledderer:
> A few more places where we need to override:
> 
> ....
> 
> ======
> 
> Will keep testing after we fix these issues.
> 
> On the other hand, I'd like to have more eyes on this PR as it impacts the entire website. So, I'll ping the team and wait until someone else approves it.
> 
> cc @nloureiro would be nice if you can test this as well.

Ooop! I will take a look at the sections you have presented.

pettinarip (changes requested):
A few more issues detected.

https://build-47c87355-85f3-4e1a-a8b0-527e59f4370f.gatsbyjs.io/en/dapps/
image

(I think it is inside the use-cases template) https://build-47c87355-85f3-4e1a-a8b0-527e59f4370f.gatsbyjs.io/en/defi/
image

The links should not be bold
image

Tutorials page, the content is bold but should be normal
https://build-47c87355-85f3-4e1a-a8b0-527e59f4370f.gatsbyjs.io/en/developers/tutorials/
image

TylerAPfledderer:
> A few more issues detected.
> 
> https://build-47c87355-85f3-4e1a-a8b0-527e59f4370f.gatsbyjs.io/en/dapps/ image
> 
> (I think it is inside the use-cases template) https://build-47c87355-85f3-4e1a-a8b0-527e59f4370f.gatsbyjs.io/en/defi/ image
> 
> The links should not be bold image
> 
> Tutorials page, the content is bold but should be normal https://build-47c87355-85f3-4e1a-a8b0-527e59f4370f.gatsbyjs.io/en/developers/tutorials/ image

@pettinarip completed these overrides (I found a couple more in the Dapps page)

pettinarip (approved):
Approving @TylerAPfledderer looks good to me!

I've added 2 minor fixes but the rest is working fine for me. Great job.

======
Will wait a bit more to see if we get another pair of 👁️

corwintines:
Screen Shot 2023-08-08 at 9 20 56 PM

I don't know that this needs to be blocking, but I am noticing that the visited styles almost begin to imply a call to action in some instances.

corwintines:
Screen Shot 2023-08-08 at 9 23 35 PM

Another example on the `/get-eth/` page where the visited style is giving this call to action look

corwintines:
Screen Shot 2023-08-08 at 9 57 41 PM

Another example on the `/staking/solo/` page in the explore guides section. Looks like a similar base component to the `/get-eth/` from above

corwintines:
Screen Shot 2023-08-08 at 10 02 59 PM

Similar to home page, but this is on `/eth/`

I think this is happening in different Card components. In this case its:

```
<CentralActionCard
  to="/what-is-ethereum/"
  title={t("page-eth-whats-ethereum")}
  description={t("page-eth-whats-ethereum-desc")}
  image={getImage(data.ethereum)!}
/>
```

corwintines:
Screen Shot 2023-08-08 at 10 05 15 PM

Another example of a Card implementation 😅 

This one is on the `/wallets` page

```
<CardList content={guides} mb={{ base: 6, lg: 0 }} />
```

corwintines:
I think these are the only issues I have been able to find. Looks good to me though other than this. Great work!

corwintines:
@TylerAPfledderer I also added one change to the stories for link to have both external and internal links in the link list

pettinarip:
Good catch @corwintines 

@TylerAPfledderer I think that every use of the `LinkBox` and `LinkOverlay` is going to inherit this "problem" where we have the visited style on it. Wondering if we could override/force these styles in the components' themes. Instead of overriding the styles of each component that consumes them.

pettinarip (changes requested):
Marking this as blocked.

@TylerAPfledderer there is an ongoing discussion about having the `Link`s as bold vs normal. We might change that back to normal. Also, wondering if you have any opinion about this, do you have a preference in terms of the font weight for the links?

On the other hand, there is also the issue detected by Corwin, about the visited styles. Let's see if we can tweak that as well. Left a comment related to this.

TylerAPfledderer:
> Good catch @corwintines
> 
> @TylerAPfledderer I think that every use of the `LinkBox` and `LinkOverlay` is going to inherit this "problem" where we have the visited style on it. Wondering if we could override/force these styles in the components' themes. Instead of overriding the styles of each component that consumes them.

@pettinarip @corwintines 

The `LinkBox` / `LinkOverlay` tandem are by themselves not theme-able. They will only inherit the issue if `LinkOverlay` consumes `Link`, which I would think would be all the time for this project's purposes.

Which leads me to think that a custom `LinkOverlay` should be created to handle this... but does not address the other instances of Link where we don't want visited styles.

I think it would be difficult to attempt an override in the theme config; I'm imagining a multi-selector stack that would get too complex.

On the other hand, there could be a `Link` variant like "unvisited"? But that seems silly,

The one solution I would lean towards is exporting a version of the `Link` component from it's file that would include visited styles instead of applying them to the base theme.

So you would have your normal `Link` component exported, and then a Link component we could call `InlineLink` that would get the visited styles.

It's looking like the choice is to only used visited styles on links that are within body copy, correct? Which from the above mentions means that use of the visited styles is secondary so it should be included in the base styles.

---

> there is an ongoing discussion about having the Links as bold vs normal. We might change that back to normal. Also, wondering if you have any opinion about this, do you have a preference in terms of the font weight for the links?

I do not have a preference here. 

Example
image
Seems pretty negligible, nor should it be a factor in readability.

nloureiro:
@corwintines, you got the point in the sense that not all "links" should have visited the state.


From the UX perspective, this is what I stand for
Screen Shot 2023-08-09 04 37 41 PM

https://ux.stackexchange.com/questions/37998/when-to-display-a-link-as-a-previously-visited-link


But ethereum.org is a mesh of links > we use buttons as links.

My goal, probably naif take, was to make all links with a visited state > mainly the ones in the content = simplifying all default <a> tags.

All "button links" should not have visited the state (and probably should be buttons).


The problem (where my naif takes might come in): we use the same style for both, so it can create a sense that it is a bug.


Example on this PR:

the visited color here is great
Screen Shot 2023-08-09 04 22 43 PM


the visited color here is still great, but it might raise some questions like the ones you pointed out
Screen Shot 2023-08-09 04 38 28 PM

My take is that we should keep the visited as it is in this PR and gradually stop using button links and convert to "buttons" (primary, secondary, etc., depending on the case. In this last example, these cards should have buttons, not links)


does this make sense to you?

pettinarip:
> My take is that we should keep the visited as it is in this PR and gradually stop using button links and convert to "buttons" (primary, secondary, etc., depending on the case. In this last example, these cards should have buttons, not links)
> 

@nloureiro I agree that this should be gradual. For that reason, what is in the scope of this PR is to change the styles **only** for `a` tags that looks like links. Everything else (e.g. cards using links) should not use the visited styles.

I think that is the only thing missing here. After that, we should be good to go.

@TylerAPfledderer 
> So you would have your normal `Link` component exported, and then a Link component we could call `InlineLink` that would get the visited styles.
> 

That option sounds good to me 👍

However, I’d naturally choose the opposite way, having the default exported `Link` with the visited styles since that is how a common link behaves and everyone will understand that quicker.

Then IMO, the secondary exported Link component is the special case. Its special bc it doesn’t have the visited style and we could add some doc comment explaining why this component exists.

TylerAPfledderer:
> > My take is that we should keep the visited as it is in this PR and gradually stop using button links and convert to "buttons" (primary, secondary, etc., depending on the case. In this last example, these cards should have buttons, not links)
> 
> @nloureiro I agree that this should be gradual. For that reason, what is in the scope of this PR is to change the styles **only** for `a` tags that looks like links. Everything else (e.g. cards using links) should not use the visited styles.
> 
> I think that is the only thing missing here. After that, we should be good to go.
> 
> @TylerAPfledderer
> 
> > So you would have your normal `Link` component exported, and then a Link component we could call `InlineLink` that would get the visited styles.
> 
> That option sounds good to me 👍
> 
> However, I’d naturally choose the opposite way, having the default exported `Link` with the visited styles since that is how a common link behaves and everyone will understand that quicker.
> 
> Then IMO, the secondary exported Link component is the special case. Its special bc it doesn’t have the visited style and we could add some doc comment explaining why this component exists.

@pettinarip Understood! 😄 I am making the case from the standpoint of how often we have a Link used in files that need the visited styles vs not visited styles, where base styles lean toward the more common use case.

Otherwise, I guess rename the secondary export with the visited styles override to maybe `BaseLink`, say, because it is being consumed by other components? 

Could also rename the primary export to `InlineLink` to make the distinction clearer, since it's a link that will be rendered inline with other text or in lists.

(Again with this naming convention to avoid further confusion! 😆 )

Either way, I find this to still be easier than trying to find re-usability from the theme config.

Let me know what you think and I can finish up from here.

pettinarip:
@TylerAPfledderer 

So, your suggestion is to have:
- `InlineLink`: as the default export or as a named export?
- `BaseLink`: to be consumed by other components that don't want the visited style

I agree with that implementation/naming if the `InlineLink` is the default export.

```sh
- Link/
    index.ts
      export BaseLink
      export default InlineLink
```

IMO the most important thing is to have the common link as the default export. Because I think it increases the DX. Exporting something different as the default will generate confusion for other devs and will require them to visit and analyze the code or comments to understand what are they using.

pettinarip:
> Either way, I find this to still be easier than trying to find re-usability from the theme config.

👍🏼

pettinarip (changes requested):
To summarize the changes discussed here and in Discord:

- [ ] remove bold from links (more about this decision in [this discord comment](https://discord.com/channels/714888181740339261/938069231570923550/1140613479464710235))
- [ ] refactor Link component to export a version with no visited styles (BaseLink)
- [ ] refactor the codebase to use the new BaseLink

@TylerAPfledderer happy to help you on any of these points. Just lmk.

TylerAPfledderer:
> To summarize the changes discussed here and in Discord:
> 
>     * [ ]  remove bold from links (more about this decision in [this discord comment](https://discord.com/channels/714888181740339261/938069231570923550/1140613479464710235))
> 
>     * [ ]  refactor Link component to export a version with no visited styles (BaseLink)
> 
>     * [ ]  refactor the codebase to use the new BaseLink
> 
> 
> @TylerAPfledderer happy to help you on any of these points. Just lmk.

@pettinarip Will get rolling!

> InlineLink: as the default export or as a named export?
I guess this a question of best practice for other components in the project that might have multiple exports in a file.

We are either talking about:
```tsx
import InlineLink, { BaseLink } from "~/components/Link"
```
or 
```tsx
import { InlineLink, BaseLink } from "~/components/Link"
```

I've yet to ask what you would prefer here. 😄 I've not needed default exports in other projects and kept with named exports to make sure I'm importing the component name consistently.

pettinarip:
@TylerAPfledderer 

I'd go with option 1 `import Link, { BaseLink } from "~/components/Link"`

----

This is what I have in mind:

The new components have the following structure
```sh
Component/
  index.tsx // <--- default exports
  Component.stories.tsx
```

Once we have that, for the Link component, we'd have
```tsx
// in `components/Link/index.tsx`

// named export for the "special" case - no visited styles
export const BaseLink = () => { ... }

// normal Link used in most cases
const InlineLink = () => <BaseLink ... />

export default InlineLink
````

That way, the users can import it with the name they want and we can keep consistency with the rest of the code.

TylerAPfledderer:
@pettinarip ok! I'll take a look at this

pettinarip (approved):
Thanks for doing all this refactor @TylerAPfledderer, looks great now.

nloureiro:
> 

I did a review, and all is good from my side
Great work @TylerAPfledderer

corwintines (approved):

corwintines:
Nice one, bringing this in! Thanks @TylerAPfledderer!

PR #10716, opened 2023-07-18
fix typo in slashing doc

paul-leydier:
## Description

Unless I'm mistaken, this sentence is refering to the slashed validator's stake, not to the stake of several validators?


corwintines (approved):
Makes sense, nice catch. Thanks!

PR #10726, opened 2023-07-19
Home page: fix icon color inside outline button

pettinarip:
Right now in the home page, we have a button that is displaying an icon with a different color.
image

## Description

This PR removes that hardcoded color and lets it inherit the color of the default button styles as specified in the DS.

Bouns: refactored a bit the `ButtonLinkRow` component to simplify it usage.

Now:
image

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -143,22 +143,13 @@ const Row = (props: { children: ReactNode; isReversed?: boolean }) => (
   </Flex>
 )
 
-const ButtonLinkRow = (props: {
-  firstButton: { to: string; child: ReactNode }
-  secondButton?: { to: string; child: ReactNode }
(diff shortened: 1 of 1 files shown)


src/pages/index.tsx
```diff
@@ -590,28 +569,18 @@ const HomePage = ({
...
-              to: "/contributing/",
-              child: <Translation id="page-index-contribution-banner-button" />,
-            }}
-            secondButton={{
-              to: "https://github.com/ethereum/ethereum-org-website",
-              child: (
-                <>
-                  <Icon
-                    as={FaGithub}
-                    color="text"
```
pettinarip, line 604: Main change in this PR: remove this color

corwintines (approved):

PR #10737, opened 2023-07-20
Fix typo on zh hello-world-smart-contract

niuhp:
编写 ---> 编译


wackerow:
Hey @niuhp, thanks for posting! @lukassim Can you remind me how we handle a-la-carte corrections to translations? Should we be directing to Crowdin?

wackerow (approved):
image

Given this is a very small patch which Google verifies / agrees with, I'm going to pull this in.

cc: @lukassim since this will also need to be updated in Crowdin

wackerow:
@all-contributors please add @niuhp for bug fix

PR #10756, opened 2023-07-21
Update zero-knowledge-proofs page

setgree:
Mostly stylistic changes for clarity, a few big ones.

1. I reworked the intro section to make it less repetitive, including opening with a quote definition rather than a rephrasing. I know this is atypical, but the rephrasing was (in my opinion) not quite right, and also sets up a kind of repetitive read. Might as well just say exactly what it is at first, as the experts defined it.

2. I reworked the bit about Tornado Cash substantially. The Ethereum docs are not the appropriate venue for arguing that "privacy has to eventually become the default on public blockchains." Instead, I noted that Tornado Cash has been sanctioned by OFAC, which is something readers should probably know lest they think that the Ethereum Foundation is telling them it's a good idea to use it and potentially face legal action 😃 

3. I didn't change this much, but the section on bribery and collusion seems a little muddled. Bribery is *sort of* an example of collusion; the point of collusion is mostly that organizations/people are working together to secretly coordinate their interests in an undisclosed way. Whether or not money changes hands is kind of besides the point; for instance, if Apple and Google engage in illegal price-fixing, it is because they believe that the arrangement benefits both of them economically, even if neither pays upfront. And bribery is much broader than collusion (you might bribe a government official to do your paperwork promptly, but no one would call this collusion). I'd think about reworking this section to think of more examples that are outright collusion, or clarifying that off-chain bribery is the main thing ZK knowledge proofs address in this context.


corwintines:
Hey @setgree!

Im going to close this out as there was a major revision on this page during the next migration. If you still think there are changes to be made to this updated content, do you mind opening it in a new PR against the new revision? Sorry for the delay here, will be on top of your new set of revisions.

setgree:
all good

PR #10761, opened 2023-07-23
Hotfix: switch from googleapis to @googleapis/calendar

corwintines:
## Description

- Switch from googleapis to googleapis/calendar to reduce bundle size


pettinarip (approved):
Approving but I  think we should revert the `cross-env` change, right? since that was not the issue and that change gives support for windows devs.

corwintines:
@pettinarip Will open that for dev after if that's ok.

PR #10771, opened 2023-07-24
Update community-meetups.json

LingTian:
Add in Lebanon local meetups.
Web3 Lebanon is a local dev meetup in Beirut.


corwintines:
@LingTian is this a recurring meetup? It seems like this is the first and only event so far.

Issue #3746, opened 2021-08-29
Re-think social footer links


samajammin:
**Is your feature request related to a problem? Please describe.**

This is just a hunch that I'm not backing up with any data but I suspect our footer links create confusion for site visitors:
![Home ethereum org 2021-08-29 at 12 47 16 PM](https://user-images.githubusercontent.com/8097623/131263462-d7fc74f4-a229-4313-9828-ba75b265a6d4.jpg)

My hunch is that it's not clear what these icons represent. Are these the "official" Ethereum social links? We seem to give the impression that they are. Anecdotally, many people drop into the ethereum.org Discord server thinking it is the Ethereum Discord & ask questions irrelevant to ethereum.org. This is understandingly confusing for people new to Ethereum. Footer links are a common patterns for projects & organizations to list their official social channels on their website but ethereum.org is not a standard website & Ethereum not a typical project.

Currently we link to:
- ethereum.org Github repo: https://github.com/ethereum/ethereum-org-website
- ethereum.org Twitter: https://twitter.com/ethdotorg
- EF YouTube channel: https://www.youtube.com/channel/UCNOfzGXD_C9YMYmnefmPH0g
- ethereum.org Discord server: https://discord.com/invite/CetY6Y4

This is a somewhat strange mix of ethereum.org & EF social presences - e.g. why do we link to the EF YouTube but not the EF Twitter? Why not add popular community social presences, like r/ethereum?

**Describe the solution you'd like**

I think we should reconsider the goals of these social links - what is the main purpose? What problem are we trying to solve for users? How can we increase awareness that there is no "official" Ethereum social channels?

One option might to be: display a bunch of social icons (e.g. including Reddit & Facebook) that all link to the Community page. Then in the community page we explicitly explain that there is no official social channels - no one controls Ethereum, there is no official Ethereum team or support - but there are many communities within the Ethereum ecosystem. From there, we could link out to a variety of credible options & organize them by categories based on what the visitor might be looking for (e.g. R&D vs. trading vs. building).

Regarding ethereum.org social links, I think these would be most relevant listed on the [About Us page](https://ethereum.org/en/about/) where it's clear we're talking about ethereum.org, not Ethereum more broadly.

**Describe alternatives you've considered**

- Remove the footer social icons entirely
- Add a blurb that explicitly states these are ethereum.org social links, not Ethereum or EF more broadly

**Additional context**

[...]

PR #10775, opened 2023-07-24
Remove EF YouTube link from Footer

corwintines:
## Description

- Removes YouTube link for EF from footer. We haven't been great at posting our community call recordings, and pointing to the Ethereum Foundation YouTube channel may confuse users

## Related Issue
Fixes https://github.com/ethereum/ethereum-org-website/issues/3746


wackerow (approved):
Did we want to mark this as fixing #3746? We may still want to consider another layer that informs users that these are not "official" anything, other than the home for the community around this website. I'd be down to consider adding this more universally anytime an external link is clicked too.

PR #10798, opened 2023-07-27
fix:formatting is broken for "Available libraries" in the last

Amit0617:
Last examples of 'Available Libraries' section are not well formatted 
## Description
Things like asterisk were visible on website.


corwintines (approved):
Thanks!

PR #10802, opened 2023-07-27
Community Events widget

corwintines:
## Description

- Implements a new community events widget to see the previous, upcoming, and next events


src/lambda/calendarEvents.ts
pettinarip, line 1: Tested on the preview deploy. The lambda seems to be working fine but the result is empty. Not sure if this is correct. Do we have events on the calendar?

corwintines: This is correct at the moment, I had deleted the test data in anticipation for this going out. Ill add some of the future events on the calendar now to help with testing this out.

pettinarip (approved):
🚀 🚀 🚀

PR #10808, opened 2023-07-28
Fix icon of the edit button

pettinarip:
Simplify and use the new DS styles on the buttons with icons.
image

## Description

Changes the color of the Edit button's icon to the primary color.

~~This PR also sets `hideArrow` as always true in the `ButtonLink` as we never want to display the arrow for any link wrapped by this component.~~


corwintines:
Im wondering if we should be defaulting that hideArrow to true. There are many places that we use the ButtonLink for external links which now don't have the arrow (Layer 2 page, run a node page, staking pages, etc).

pettinarip:
Good point @corwintines. I think you are right. Let me revert that part.

corwintines (approved):

Issue #10478, opened 2023-06-16
Redesigning the homepage 


rohan9024:
### Is your feature request related to a problem? Please describe.

The feature is not related to any problem yet we can make the homepage look much better by adding animations.

### Describe the solution you'd like

We can add animations, add 3d models using three js and tailwind css for styling.

### Describe alternatives you've considered

-Animations using framer motion
-Rendering 3d models using three js
-Tailwind css for styling

Tailwind css--> https://tailwindcss.com/

### Additional context

Please assign me this issue. If the design team has already made some 3d models or designs, I would love to collaborate with them.

### Would you like to work on this issue?

- [X] Yes
- [ ] No

rohan9024:
@corwintines can you please guide me on this?

konopkja:
Hello @rohan9024, Im excited to see a community member start looking into animations. Thank you for taking the initiative. 

The core team has considered adding animations to the site few times too. They could be a great UX enhancer. While it is exciting, we would like to explore this on a smaller scale at fist to see the performance. 

My suggestion that would have buy-in is to:

1. animate menu icons: globe and sun/halfmoon using framer
2. inspiration for what we are thinking: https://m2.material.io/design/iconography/animated-icons.html

After that maybe we can move forward and explore other parts of the website that could be enhanced with animations. 

Would you be up to take this on?

rohan9024:
@konopkja Yeah sure.... I would love to.... Can I add tailwind css?

konopkja:
> @konopkja Yeah sure.... I would love to.... Can I add tailwind css?

@pettinarip fyi

bod0bod:
> @konopkja should I proceed with it?

Certainly

pettinarip:
Hi @rohan9024 exciting idea! thanks for sharing and offering your coder hands for it :)

We are not using Tailwind, lets not add it.

Our current css stack is Chakra UI + framer motion. We should be able to create the animations with those tools. You can see [here](https://chakra-ui.com/getting-started/with-framer) how to use it.

konopkja:
@rohan9024 do you still want to work on this? if so let us know and i will assign you the issue to start the work :)

rohan9024:
@konopkja yeah definitely! I would love to.
But I need some assistance since I'm new to gatsby and chakra ui.

pettinarip:
[...]

PR #10811, opened 2023-07-28
Added hover animations for nav icons

rohan9024:
## Description

I have used basic chakra ui hover animations

Here is the video of it-->

https://github.com/ethereum/ethereum-org-website/assets/64919642/0bd326d6-bd31-4795-a94c-c2c826d99aa6

This was the issue--> #10478

Can anyone from the core team assist me in design they have been thinking for the home page?


corwintines:
@nloureiro would like some input from you here. 

I know we initially wanted `language` to be there to make the button more explicit for users, so I do think that needs to be added back in. 

I'm not the biggest fan of the little rotation on the color mode. Don't think it fits much of whats on the site. But the hover effect making it larger seems fine to me. 

@rohan9024 I would also scope this outside of the homepage as this is specific to the nav. Open to other ideas you have for the homepage, but we have not spent time on redesigning the home page at the moment.

rohan9024:
@corwintines removed the rotation animation. Are there any more changes to be made?

nhsz:
I'd push against using Framer for simple animations as it adds unnecesary complexity and JS code to the bundle we're aiming to reduce, otherwise we'll start to bloat the site. We could use CSS instead. @konopkja @corwintines @pettinarip

pettinarip:
> I'd push against using Framer for simple animations as it adds unnecesary complexity and JS code to the bundle we're aiming to reduce, otherwise we'll start to bloat the site. We could use CSS instead. @konopkja @corwintines @pettinarip

Agree. However, he is not using framer motion for these animations. Probably you were confused because there is an import of `motion` but is not being used xD

konopkja:
> @corwintines removed the rotation animation. Are there any more changes to be made?

so how does it look now with that change? can i see?

pettinarip:
> > @corwintines removed the rotation animation. Are there any more changes to be made?
> 
> so how does it look now with that change? can i see?

@konopkja you can access the preview deploy [here](https://ethereumorgwebsitedev01-rohan9024ethereumorgwebsitedev.gatsbyjs.io/en/) or by clicking in [this auto-generated comment](https://github.com/ethereum/ethereum-org-website/pull/10811#issuecomment-1656019336) by the GatsbyCloud bot
image

konopkja:
> * Deploy preview

Yes the language should be added back in, languages are important part of our website.

I am actually personally for the spin as its more visible and is more engaging (which is something in small doses we want to add to the website). But lets see what other think and also ping @nloureiro 

Already better than nothing as we have now.

rohan9024:
@konopkja Thank you Jakub!!! Finally someone gets me🥰 I'll add the languages text back👍

rohan9024:
> > I'd push against using Framer for simple animations as it adds unnecesary complexity and JS code to the bundle we're aiming to reduce, otherwise we'll start to bloat the site. We could use CSS instead. @konopkja @corwintines @pettinarip
> 
> Agree. However, he is not using framer motion for these animations. Probably you were confused because there is an import of `motion` but is not being used xD

Oops, sorry for that😂.  I'll remove that.....

nloureiro:
Hey, @rohan9024 thank you for this.

There are various considerations on this PR, and from my stand, I think CSS it's the best way to go, if possible, for animations. Simpler and lighter for the browser.

I would add the text we have in production now and rotate 10º to the globe to be the same type of animation as the moon/sun.

The timing and speed look good to me.

Something like this.

https://github.com/ethereum/ethereum-org-website/assets/1120748/05da15b4-d4bc-44ad-97a6-2b10ff138845

rohan9024:
@nloureiro have you made a commit or should I?

.env.example
corwintines, line 1: This shouldn't be deleted

yarn.lock
corwintines, line 1: I don't think this needs to be added, we can update packages in a separate PR

src/components/Nav/index.tsx
```diff
@@ -78,22 +77,35 @@ const Nav: FC<IProps> = ({ path }) => {
...
                 }
+                _hover={{
+                  color: "primary.base",
+                  transform: "scale(1.1) rotate(20deg)",
+                }} // Scale and rotate the icon on hover
+                transition="transform 0.2s ease-in-out, color 0.2s ease-in-out"
                 icon={<Icon as={isDarkTheme ? MdWbSunny : MdBrightness2} />}
                 variant="icon"
-                size="sm"
+                size="md"
```
corwintines, line 87: We shouldn't be changing the size here

corwintines (changes requested):
Thanks @rohan9024 

The following bugs need to be addressed here
- [ ] Icon sizes should go back to original
- [ ] The gap between icons increased, this should be the same as before
- [ ] Add the .env.example file back
- [ ] The highlight of the language icon and text should be in sync

https://github.com/ethereum/ethereum-org-website/assets/15589226/ebc00f0b-b132-4709-8a97-eb52cbdc079d

- [ ] When highlighting the language text, it should also apply the rotation of the icon


https://github.com/ethereum/ethereum-org-website/assets/15589226/08934f43-7c6c-4315-9fb9-79b0a239fae7

corwintines:
@rohan9024 we merged some changes into dev for the Nav. You'll need to merge dev into your branch here to resolve conflicts.

rohan9024:
@corwintines Sure, I'll merge it 👍

Issue #10806, opened 2023-07-28
Sidebar not hiding when transitioning from dev docs to tutorials


pettinarip:
### Describe the bug

The dev docs [Sidebar](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/SideNav.tsx) is not hiding when the user transition from the dev docs page to the tutorials page.

### To reproduce

1. Go to https://ethereum.org/en/developers/docs/
2. Click on Tutorials in the secondary nav
3. See error

### Expected behavior

When the user is in the Tutorials page, the sidebar should not be visible

### Screenshots

![image](https://github.com/ethereum/ethereum-org-website/assets/468158/d438831c-fd85-4c2c-9a00-39eb17a81b68)

### Would you like to work on this issue?

- [ ] Yes
- [X] No

wackerow:
Fixed with #10812

PR #10812, opened 2023-07-28
[Hotfix] Fixes [#10806] added condition to setShouldShowSideNav as false

altinocoelho:
## Description

Fixes [#10806] added condition to setShouldShowSideNav as false

not sure what caused the issue, but since it affects all pages, this works at least as workaround 

## Related Issue

[#10806]

### DIFF
--- a/src/components/Layout.tsx
+++ b/src/components/Layout.tsx
@@ -52,7 +52,6 @@ const Layout: React.FC<IProps> = ({
     if (path.includes("/docs/")) {
       setShouldShowSideNav(true)
     } else {
-      !path.includes("/docs/")
       setShouldShowSideNav(false)
     }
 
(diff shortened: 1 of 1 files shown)


src/components/Layout.tsx
```diff
@@ -51,6 +51,9 @@ const Layout: React.FC<IProps> = ({
   useEffect(() => {
     if (path.includes("/docs/")) {
       setShouldShowSideNav(true)
+    } else {
+      !path.includes("/docs/")
```
pettinarip, line 55: What is this line for?
```suggestion
```

pettinarip (changes requested):
Thanks for the PR @altinocoelho

src/components/Layout.tsx
wackerow: Yeah looks like this may have been left over as a note, but isn't needed for the logic since the `else` statements covers it already.

pettinarip: Correct. Ok, I'll remove it for now.

pettinarip (dismissed):
@altinocoelho thanks for. I'll change the base of the PR to `master` to merge this as a hotfix. Will be deployed soon 👍🏼

PR #10815, opened 2023-07-30
Contextualize the "More on EIPs" link in the Governance page

Pandapip1:
## Description

Removes a fact that isn't particularly relevant to the topic, and adds context for the "More on EIPs" link that follows it.

CC @dtedesco1 

## Related Issue

*N.A.*


corwintines (approved):
Thanks @Pandapip1!

PR #10823, opened 2023-07-30
Fix Nav styles to match DS header

### LINKED ISSUE #10807: Incorrect font size in color mode and language buttons
### Describe the bug

The current font size for the color mode and language buttons is too small.

### To reproduce

1. Go to the home page
2. See error on the color mode & language buttons on the nav bar

### Expected behavior

Font size should be the `md` size (1rem or 16px).

### Screenshots

image

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes
- [X] No

pettinarip:
## Description

This PR fixes the styles issues reported in #10807 but also refactor the `Nav` component to match the DS.

## TODO
- [x] replace & remove deprecated `icon` variant
- [x] update docs if necessary
- [x] @TylerAPfledderer see if we can still use the previous api for the Button theme (using `isSecondary`) and refactor accordingly

## Related Issue

#9546 
Fixes #10807


src/components/Nav/index.tsx
```diff
@@ -62,52 +63,48 @@ const Nav: FC<IProps> = ({ path }) => {
...
-              </ButtonLink>
+              <HStack spacing={2} hideBelow="lg">
+                <Button
+                  aria-label={
+                    isDarkTheme
+                      ? "Switch to Light Theme"
+                      : "Switch to Dark Theme"
+                  }
+                  variant="secondaryGhost"
+                  px={1.5}
```
pettinarip, line 92: I'd like to have your 👀 on this PR @TylerAPfledderer, especially with the following changes:

1st change is in the `Button` component, where I had to remove the `sx` inner logic because otherwise, I wasn't able to override the `px` prop. Not sure exactly why (I think it is because the `sx` prop has more priority than the style props).

2nd change is a consequence of the 1st one. Since I removed the usage of `useStyleConfig` and `sx` from the Button component (and ButtonLink), I started to see the warning about the `isSecondary` flag. I ended up removing it and creating 2 new variants to cover the secondary cases. Not ideal I think, but it worked and is simple to understand.

Let me know what do you think about these changes or if you have any concerns. Also, feel free to test the rest of the changes as well and leave any comments :)

src/components/Nav/index.tsx
```diff
@@ -62,52 +63,48 @@ const Nav: FC<IProps> = ({ path }) => {
...
-              gap={{ base: 1, xl: 0 }}
+              gap={{ base: 2, xl: 4 }}
             >
-              <Search ref={searchRef} />
-              <IconButton
-                aria-label={
-                  isDarkTheme ? "Switch to Light Theme" : "Switch to Dark Theme"
-                }
-                icon={<Icon as={isDarkTheme ? MdWbSunny : MdBrightness2} />}
-                variant="icon"
```
pettinarip, line 82: Trying to remove as much as I can the `icon` variant to finally delete it from the theme.

src/components/Nav/index.tsx
```diff
@@ -25,9 +26,9 @@ const Nav: FC<IProps> = ({ path }) => {
     t,
     toggleColorMode,
     linkSections,
-    searchRef,
     mobileNavProps,
   } = useNav({ path })
+  const searchModalDisclosure = useDisclosure()
```
pettinarip, line 31: Moved the search modal logic to the parent to avoid using refs which was causing me a headache.

src/components/Search/index.tsx
```diff
@@ -27,142 +27,145 @@ import { trackCustomEvent } from "../../utils/matomo"
...
-    />
+    <Button ref={ref} variant="secondaryGhost" px={1.5} {...props}>
+      <MdSearch />
+    </Button>
   )
 )
 
-const Search = forwardRef<{}, "button">((_, ref) => {
-  const searchButtonRef = React.useRef<HTMLButtonElement>(null)
-  const { isOpen, onClose, onOpen } = useDisclosure()
```
pettinarip, line 43: Main change here was the removal of this line. It was moved to the parent component (the Nav).

src/components/Nav/index.tsx
TylerAPfledderer: To the first, I can see where the issue would happen, so this is definitely cleaner! 😄 

As to the second, it is certainly a matter of preference. I guess here it is readability versus overkill? I'm good with this though! Indeed still easy to roll with.

src/components/Nav/index.tsx
TylerAPfledderer: Good call here 😅

src/components/Nav/index.tsx
```diff
@@ -62,52 +63,48 @@ const Nav: FC<IProps> = ({ path }) => {
...
+                    isDarkTheme
+                      ? "Switch to Light Theme"
+                      : "Switch to Dark Theme"
+                  }
+                  variant="secondaryGhost"
+                  px={1.5}
+                  onClick={toggleColorMode}
+                >
+                  <Icon as={isDarkTheme ? MdWbSunny : MdBrightness2} />
+                </Button>
```
TylerAPfledderer, line 96: Using `IconButton` also works here, and still simplified from the original
```suggestion
                <IconButton
                  icon={isDarkTheme ? <MdWbSunny /> : <MdBrightness2 />}
                  aria-label={
                    isDarkTheme
                      ? "Switch to Light Theme"
                      : "Switch to Dark Theme"
                  }
                  variant="secondaryGhost"
                  px={1.5}
                  onClick={toggleColorMode}
                />
```

TylerAPfledderer (changes requested):
Only one additional thought I had here. 😄

src/components/Nav/index.tsx
pettinarip: Thanks @TylerAPfledderer 

This PR is a draft because I wanted to first check with you about it and see if you had a workaround to keep the same convention you built (with simpler variants + the `isSecondary` flag). Tbh I don't like having variants called `"secondaryOutline"` hahaha.

So, let me know if you prefer to hold on a bit here and see if you can add the necessary changes to make this work with the original convention you did.

pettinarip: Awesome thanks!

LMK if you can add commits to this PR (I doubt it) or create a new PR off this branch and we can merge it back.

src/components/Nav/index.tsx
TylerAPfledderer: I'm not sure why it would be doing that. I'm not seeing that issue when I try to apply it. Maybe something to follow up with later once this PR gets pushed?

pettinarip: Hmm probably I was just doing silly things and it works fine xD

Yea...lets leave it for now and we can tackle it later.

corwintines (approved):
Nice!

PR #10855, opened 2023-08-03
add Translatathon banner

corwintines:
## Description

- Adds banner for Translatathon


pettinarip (commented):
For some reason, the build was good but the site is crashing. I'll trigger a new build just in case
image

pettinarip (approved):

PR #10859, opened 2023-08-03
Button - Fix disabled styles in solid variant

pettinarip:
## Description

Adds the correct disabled styles to the button's solid variant as defined in the DS. It also shows the disabled state in the button's stories.


src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -60,7 +60,7 @@ const semanticTokens = {
       base: { _light: "white", _dark: "gray.700" },
       highlight: { _light: "gray.100", _dark: "gray.900" },
     },
-    disabled: { _light: "gray.300", _dark: "gray.500" },
+    disabled: { _light: "gray.400", _dark: "gray.500" },
```
pettinarip, line 63: Updated this token to match the DS spec

PR #10860, opened 2023-08-03
Fix quiz modal bg color

pettinarip:
Atm the bg color of the Quiz modal is hardcoded to `white`, which generates the following result in dark mode
image

## Description

This PR uses the semantic token `neutral` to display the correct color between light and dark mode.


minimalsm (approved):
Nice catch, LGTM :)

Issue #10828, opened 2023-07-31
Bug report: quizzes page crashes while loading


0x0OZ:
### Describe the bug

The [quizzes](https://ethereum.org/en/quizzes/) page crashes while loading and show a blank screen after everything is loaded

Nothing can describe it more than a video:

https://github.com/ethereum/ethereum-org-website/assets/52073989/1c00f310-09ac-4633-b4ff-8175d3446a7a

the error occurs in `./src/components/Quiz/QuizItem.tsx` line 24 `id` is undefined

### To reproduce

I couldn't reproduce this bug in other browsers including an incognito mode in chrome browser...

The only thing I did on this page was do the quizzes; doing them again didn't reproduce it.

I didn't try to clear the cache to fix it, because I might not be able to reproduce it and figure out what is the cause if possible.
### Expected behavior

/

### Screenshots

![image](https://github.com/ethereum/ethereum-org-website/assets/52073989/71b95ce6-4cac-44ae-90aa-0da117f677ff)
![image](https://github.com/ethereum/ethereum-org-website/assets/52073989/babb3a22-1436-4f4b-bd80-5bc4509d27d1)

_No response_

### Desktop (please complete the following information)

- OS: Manjaro Linux , Kernel: `6.1.41-1-MANJARO`
- Affected Browser: Chrome
- Browser version: `Google Chrome 115.0.5790.110`

### Would you like to work on this issue?

- [ ] Yes
- [X] No

MahendraBishnoi29:
I would love to work on this issue 🚀

corwintines:
Will assign to you @MahendraBishnoi29, thanks!

MahendraBishnoi29:
> Following the discussion from #10845, afaik it's not possible to 'sync' changes when using localStorage if we change the structure of the data we are using to track progress across questions (e.g.: adding new quizzes, changing the order, etc) so this kind of issues are expected for users with progress on a previous version when deploying new quizzes cc @pettinarip @corwintines

how should i proceed and tackle this issue then? @nhsz

pettinarip:
Thanks for the effort @MahendraBishnoi29. We've decided to revert these changes #10873 for now until we have a robust solution which may imply doing a refactor on how we store state on localstorage.

PR #10862, opened 2023-08-03
Added safe check to completed quiz item so it doesn't return undefined

checkomkar:
## Description
Added safe check so completed quiz item so it doesn't return undefined

## Related Issue

#10828

### DIFF
--- a/src/components/Quiz/QuizItem.tsx
+++ b/src/components/Quiz/QuizItem.tsx
@@ -21,7 +21,7 @@ const QuizItem: React.FC<QuizzesListItem> = (props) => {
     userStats: { completed },
   } = useContext(QuizzesHubContext)
   const numberOfQuestions = allQuizzesData[id].questions.length
-  const isCompleted = JSON.parse(completed)[id][0]
+  const isCompleted = JSON.parse(completed)?.[id]?.[0] || false
 
   const { t } = useTranslation()
 
--- a/src/components/Quiz/utils.ts
+++ b/src/components/Quiz/utils.ts
@@ -42,6 +42,14 @@ export const updateUserStats = ({
 }) => {
   // Read userStats from localStorage as quiz could be standalone (out of Quiz Hub page)
   const userStats = JSON.parse(localStorage.getItem(USER_STATS_KEY)!)
+  const oldCompletedQuizzes = JSON.parse(userStats.completed)
+  // Check if quiz key present in local storage, if not, create it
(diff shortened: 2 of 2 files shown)


pettinarip (changes requested):
Thanks for the PR @checkomkar.

Even though this change fixes the first issue and lets you load the page, it has another problem.

When you try to make one of the new quizzes like "Run a node", once you finish it, the logic tries to save the progress and it fails, crashing the page.

To simulate the error:
1. Manually add old quiz data in localstorage:

```
key: quizzes-stats
data: {"score":0,"average":[],"completed":"{\"what-is-ethereum\":[false,0],\"what-is-ether\":[false,0],\"web3\":[false,0],\"wallets\":[false,0],\"security\":[false,0],\"nfts\":[false,0],\"layer-2\":[false,0],\"merge\":[false,0]}"}
```

2. Try to complete the "Run a node" quiz

checkomkar:
> Thanks for the PR @checkomkar.
> 
> Even though this change fixes the first issue and lets you load the page, it has another problem.
> 
> When you try to make one of the new quizzes like "Run a node", once you finish it, the logic tries to save the progress and it fails, crashing the page.
> 
> To simulate the error:
> 
> 1. Manually add old quiz data in localstorage:
> 
> ```
> key: quizzes-stats
> data: {"score":0,"average":[],"completed":"{\"what-is-ethereum\":[false,0],\"what-is-ether\":[false,0],\"web3\":[false,0],\"wallets\":[false,0],\"security\":[false,0],\"nfts\":[false,0],\"layer-2\":[false,0],\"merge\":[false,0]}"}
> ```
> 
> 2. Try to complete the "Run a node" quiz

Hey @pettinarip, thank you for highlighting the issue. I have added a fix for the same, see if that work. Thanks, cheers!

wackerow:
Thanks for the help @checkomkar! Just tried this out, and the /quizzes page does load, but I'm getting strange results in the summary section... "Completed" quizzes only seems to increment if the user scores 100%, and the "Your total points" score stays at `0 / 54` no matter what. The green tick mark for completion also only shows up if you score 100%.

pettinarip:
FYI @checkomkar thanks for trying to fix this issue but we've decided to revert these changes #10873 for now until we have a more robust solution which probably is going to involve doing a refactor on the quizzes logic.

pettinarip:
Thanks again @checkomkar for the effort you put on this. I'll close this one in favor of #10873. FYI we will try to do a refactor to find a better solution #10887

PR #10863, opened 2023-08-04
fix(Button): reintroduce `isSecondary` prop

TylerAPfledderer:


src/components/Button/index.tsx
```diff
@@ -7,11 +7,37 @@ import {
...
+  isSecondary?: boolean
+}) => {
+  const { variant, isSecondary } = props
+  // These two variants do not have secondary styling, so prevent overrides
+  return {
+    "data-secondary":
+      !["solid", "link"].includes(variant || "solid") && isSecondary,
+  }
+}
+
```
TylerAPfledderer, line 21: This allows use of `data-secondary` under both the `Button` and `IconButton` components, and avoiding the need to do forced styling in the theme config (i.e. add `!important` flag to a style value, say)

src/@chakra-ui/gatsby-plugin/components/Button.ts
```diff
@@ -30,7 +31,16 @@ const baseStyle = defineStyle((props) => ({
...
-}))
+  "&[data-secondary='true']": {
+    color: "body.base",
+  },
+  "&.chakra-link": {
+    textDecoration: "none",
+    _hover: {
+      textDecoration: "none",
+    },
+  },
```
TylerAPfledderer, line 42: For any instance `Button` is a link, typically consuming a component that uses Chakra's `Link`

src/components/Button/Button.stories.tsx
```diff
@@ -113,3 +115,20 @@ export const MultiLineText: Story = {
...
+      </Text>
+      <VStack>
+        <IconButton aria-label="toggle" icon={<MdNightlight />} px="1.5" />
+        <ButtonLink to="#" borderRadius="full" px="0" py="0">
+          <Translation id="get-involved" />
+        </ButtonLink>
+      </VStack>
+    </>
+  ),
+}
```
TylerAPfledderer, line 134: Use this story to compile any visual testing where styles are overridden on the component render. Can get wacky here with edge cases to ensure the overrides will work.

src/components/Button/index.tsx
```diff
@@ -20,7 +46,14 @@ const Button = forwardRef<IProps, "button">(({ toId, onClick, ...props }) => {
     onClick?.(e)
   }
 
-  return <ChakraButton onClick={handleOnClick} {...props} />
+  return (
+    <ChakraButton
+      ref={ref}
+      onClick={handleOnClick}
+      {...checkIsSecondary({ variant: rest.variant?.toString(), isSecondary })}
```
TylerAPfledderer, line 53: `rest.variant?.toString()` because the `variant` prop type is a union.

src/components/ButtonLink.tsx
```diff
@@ -1,24 +1,13 @@
 import React from "react"
-import { Button, ButtonProps, useStyleConfig } from "@chakra-ui/react"
 
+import type { IProps as IButtonProps } from "./Button"
 import Link, { IBaseProps as ILinkProps } from "./Link"
+import Button from "./Button"
 
-export interface IProps extends ILinkProps, ButtonProps {}
+export interface IProps extends ILinkProps, Omit<IButtonProps, "toId"> {}
```
TylerAPfledderer, line 7: Improve type safety by emitting a TypeError for `toId`, since this component renders a link.

TylerAPfledderer:
@pettinarip added the above comments, and good to go for a review! 😄

src/@chakra-ui/gatsby-plugin/components/Button.ts
pettinarip: Nice simplification, this would cover also the cases where we apply a Link to an IconButton. Well...any instance as you said haha

src/components/Button/index.tsx
pettinarip: Works perfectly 💪🏼

src/components/Button/index.tsx
```diff
@@ -7,11 +7,37 @@ import {
...
+   * trigger a scroll to that element.
+   */
   toId?: string
+  /**
+   * Custom theme prop. If true, `body` color is used instead of
+   * `primary` color in the theming.
+   *
+   * `NOTE`: Does not apply to the `Solid` or `Link` variants
+   */
+  isSecondary?: boolean
```
pettinarip, line 35: Thanks for adding those comments 🙏🏼

pettinarip (approved):
Thanks @TylerAPfledderer this works much better now. Will merge it and test it more deeply in the Nav's refactor PR.

PR #10476, opened 2023-06-16
Add quiz: staking

wackerow:
## Description
- Adds an eight-question quiz for "staking" topic

## Preview
https://ethereumorgwebsitedev01-quizstaking.gatsbyjs.io/en/staking/solo/#quiz

PR #10475, opened 2023-06-16
Add quiz: run a node

wackerow:
## Description
- Adds a six-question quiz for "run a node" topic

## Preview
https://ethereumorgwebsitedev01-quizrunanode.gatsbyjs.io/en/run-a-node#quiz

PR #10873, opened 2023-08-05
Revert new quizzes

pettinarip:
Due to [recent issues](https://github.com/ethereum/ethereum-org-website/issues/10828) detected in the Quizzes hub page, we decided to revert the additions of the new quizzes until we refactor the code to avoid any crashes.

## Description

Reverts the following PRs: #10475, #10476, and #10477.

## Related Issue

#10828


corwintines:
Is there a way to remove the links to quizzes on /quizzes/ but keep these for the actual pages themselves?

pettinarip:
> Is there a way to remove the links to quizzes on /quizzes/ but keep these for the actual pages themselves?

Oh good idea @corwintines! let me try doing that instead.

wackerow:
Agree it would be nice to only remove from the quiz hub if possible, but also don't want this to sit too long since it's currently breaking the quiz hub.

pettinarip:
I pushed the revert again since it was missing a few things, now it should work as expected.

@corwintines @wackerow I tried to keep the individual quizzes on the specific pages but I don't think it is possible because the QuizWidget and the Hub are storing the results in localstorage, so it is messing with the final stats displayed in the hub.

wackerow:
Okay... bit of a bummer, but hopefully we can get this figured out eventually so we can add more quizzes. No question that this quiz hub needs to be flexible to account for new quizzes and questions in the future.

wackerow:
Noticing too while reviewing these, and commented similarly on https://github.com/ethereum/ethereum-org-website/pull/10862#issuecomment-1667393026 that the "Completed" count only increments if the user gets 100%, and same with the green tick.

image

^ After completing the "Ethereum security and scam prevention" quiz with one wrong answer.


I think we should reconsider how to display this... if we want to have an exclusive counter for perfect quizzes, I would rename this to something other than "Completed" although I don't think holding people to perfect scores is the best way to go. We currently use `PASSING_QUIZ_SCORE = 65`, so perhaps we update this term and include any quiz that is above that threshold instead. Also we may want to have some additional indicators aside from the green tick to indicate that the quiz has been 1- been attempted, 2- passed, 3- completed perfect.

wackerow (dismissed):
Let's just make sure we come back to this 🙏 
Thanks for addressing this @pettinarip!

wackerow (approved):

pettinarip:
> Noticing too while reviewing these, and commented similarly on [#10862 (comment)](https://github.com/ethereum/ethereum-org-website/pull/10862#issuecomment-1667393026) that the "Completed" count only increments if the user gets 100%, and same with the green tick.
> 
> image
> ^ After completing the "Ethereum security and scam prevention" quiz with one wrong answer.
> 
> I think we should reconsider how to display this... if we want to have an exclusive counter for perfect quizzes, I would rename this to something other than "Completed" although I don't think holding people to perfect scores is the best way to go. We currently use `PASSING_QUIZ_SCORE = 65`, so perhaps we update this term and include any quiz that is above that threshold instead. Also we may want to have some additional indicators aside from the green tick to indicate that the quiz has been 1- been attempted, 2- passed, 3- completed perfect.

cc ^ fyi @konopkja

konopkja:
> > Noticing too while reviewing these, and commented similarly on [#10862 (comment)](https://github.com/ethereum/ethereum-org-website/pull/10862#issuecomment-1667393026) that the "Completed" count only increments if the user gets 100%, and same with the green tick.
> > image
> > ^ After completing the "Ethereum security and scam prevention" quiz with one wrong answer.
> > I think we should reconsider how to display this... if we want to have an exclusive counter for perfect quizzes, I would rename this to something other than "Completed" although I don't think holding people to perfect scores is the best way to go. We currently use `PASSING_QUIZ_SCORE = 65`, so perhaps we update this term and include any quiz that is above that threshold instead. Also we may want to have some additional indicators aside from the green tick to indicate that the quiz has been 1- been attempted, 2- passed, 3- completed perfect.
> 
> cc ^ fyi @konopkja

good catch, the status "success" on final page should correspond to "completed" in the table and for the v2 we could think of enhancing this experience

PR #10885, opened 2023-08-07
Update types.ts

leleka-barbie:


corwintines:
This adds an extra space that isn't needed. Going to close this out.

PR #10894, opened 2023-08-09
Updates L2 page copy

jmcook1186:
## Description

Thought the original copy was a bit too opinionated and some parts felt a little out of date. These updates aim to make it more clear that there are trade-offs.


src/intl/en/page-layer-2.json
```diff
@@ -27,10 +27,10 @@
...
   "layer-2-why-do-we-need-layer-2-1": "Three desirable properties of a blockchain are that it is <strong>decentralized, secure, and scalable</strong>. The <a href=\"https://www.ledger.com/academy/what-is-the-blockchain-trilemma\" target=\"_blank\">blockchain trilemma</a> states that a simple blockchain architecture can only achieve two out of three. Want a secure and decentralized blockchain? You need to sacrifice scalability.",
-  "layer-2-why-do-we-need-layer-2-2": "Ethereum has reached the network's current capacity with <a href=\"https://etherscan.io/chart/tx\" target=\"_blank\">1+ million transactions per day</a> and high demand for each of these transactions. The success of Ethereum and the demand to use it has caused gas prices to rise substantially. Therefore the need for scaling solutions has increased in demand as well. This is where layer 2 networks come in.",
+  "layer-2-why-do-we-need-layer-2-2": "Ethereum currently processes <a href=\"https://etherscan.io/chart/tx\" target=\"_blank\">1+ million transactions per day</a>. The demand for transactions to be included in blocks can cause gas prices be high. This is where layer 2 networks come in.",
```
konopkja, line 30:
```suggestion
  "layer-2-why-do-we-need-layer-2-2": "Ethereum currently processes <a href=\"https://etherscan.io/chart/tx\" target=\"_blank\">1+ million transactions per day</a>. The demand to use Ethereum can cause transaction fee prices be high. This is where layer 2 networks come in.",
```
@jmcook1186 i think we need some framework how to tackle jargon.. when to be specific and use terms like gas or blocks and when to use normal language like transaction fee

on pages like L2 i suspect we should attempt to remove this jargon as its audience probably doesnt need to know the precise details or terms.  This is a discussion tho and curious about other's take. 

cc @minimalsm

src/intl/en/page-layer-2.json
```diff
@@ -40,19 +40,19 @@
...
+  "layer-2-rollups-1": "Rollups bundle (or ’roll up’) hundreds of transactions into a single transaction on layer 1. This distributes the L1 transaction fees across everyone in the rollup, making it cheaper for each user.",
+  "layer-2-rollups-2": "Rollup transactions get executed off layer 1 but the transaction data gets posted to layer 1. By posting transaction data onto layer 1, rollups inherit the security of Ethereum. This is because once the data is posted to layer 1, reverting a rollup transaction requires reverting Ethereum. There are two different approaches to rollups: optimistic and zero-knowledge - they differ primarily on how this transaction data is posted to L1.",
```
konopkja, line 47:
```suggestion
  "layer-2-rollups-2": "Rollup transactions get executed off layer 1 but the transaction data is submitted to layer 1. By submitting transaction data onto layer 1, rollups inherit the security of Ethereum. This is because once the data is uploaded to layer 1, reverting a rollup transaction requires reverting Ethereum. There are two different approaches to rollups: optimistic and zero-knowledge - they differ primarily on how this transaction data is submitted to L1.",
```

"to post" is dev jargon, language like "submit" contains same message, but is much more user friendly.

src/intl/en/page-layer-2.json
```diff
@@ -40,19 +40,19 @@
...
   "layer-2-zk-rollups-title": "Zero-knowledge rollups",
   "layer-2-zk-rollups-description": "Zero-knowledge rollups use validity proofs where transactions are computed off-chain, and then compressed data is supplied to Ethereum Mainnet as a proof of their validity.",
   "layer-2-zk-rollups-childSentance": "More on ZK-rollups",
   "layer-2-dyor-title": "Do your own research: risks of layer 2",
-  "layer-2-dyor-1": "Since layer 2 chains inherit security from Ethereum, in an ideal world, they are as safe as L1 Ethereum. However, many of the <strong>projects are still young and somewhat experimental</strong>. After years of research and development, many of the L2 technologies that will scale Ethereum launched in 2021. Many projects still have additional trust assumptions as they work to decentralize their networks. Always do your own research to decide if you're comfortable with any risks involved.",
+  "layer-2-dyor-1": "Many layer 2 projects are relatively young. Many projects still have additional trust assumptions as they work to decentralize their networks. Always do your own research to decide if you're comfortable with any risks involved.",
```
konopkja, line 55: i dont love that this phrase "trust assumptions" is used so often here. 

It is a difficult jargon that our readers may not really understand what that means, instead I would welcome an explanation. 

Especially because this urges users to do research, but if they dont understand why, then they are less likely to do so.

konopkja, line 55:
```suggestion
  "layer-2-dyor-1": "Many layer 2 projects are relatively young and still have additional trust assumptions as they work to decentralize their networks. Always do your own research to decide if you're comfortable with any risks involved.",
```
2x "many" used in a row

konopkja:
Screenshot 2023-08-09 at 22 32 28

This note should be rewritten to something like this:  If you are using smart contract wallet such as Safe or Argent, you will not have control over this address on a layer 2 until you redeploy your contract account to that address on the layer 2. Classic accounts with recovery phrase will automatically own the same account on all layer 2 networks.

src/intl/en/page-layer-2.json
```diff
@@ -67,8 +67,7 @@
   "layer-2-sidechains-2": "Both scale similarly to layer 2s - they offer lower transaction fees and higher transaction throughput - but have different trust assumptions.",
   "layer-2-more-on-sidechains": "More info on sidechains",
   "layer-2-more-on-validiums": "More info on validiums",
-  "layer-2-sidechains-4": "Some <strong>layer 1 blockchains</strong> have higher throughput and lower transaction fees than Ethereum. These alternative layer 1s have had to <b>sacrifice on security or decentralization</b> in order to achieve higher transactions per second and lower transaction fees.",
-  "layer-2-sidechains-5": "The Ethereum ecosystem is firmly aligned that <strong>layer 2 scaling is the only way to solve the scalability trilemma</strong> while remaining decentralized and secure.",
+  "layer-2-sidechains-4": "Some layer 1 blockchains report higher throughput and lower transaction fees than Ethereum, but generally with trade-offs elsewhere, for example greater hardware requirements for running nodes.",
```
konopkja, line 70:
```suggestion
  "layer-2-sidechains-4": "Some layer 1 blockchains report higher throughput and lower transaction fees than Ethereum, but generally with trade-offs elsewhere, for example greater hardware requirements for running nodes.",
```
can we add here a comment that that results in significant centralization and control over the network?  or something like that

jmcook1186: The arguments around centralizing forces are very nuanced. I think it's better to stay very high level here as different alt-L1's have different trade-offs and so does Ethereum's L2-centric model of scaling.

src/intl/en/page-layer-2.json
```diff
@@ -40,24 +40,24 @@
...
+  "layer-2-how-does-layer-2-work-1": "As we mentioned above, layer 2 is a collective term for Ethereum scaling solutions that handle transactions off Ethereum layer 1 while still taking advantage of the robust decentralized security of Ethereum layer 1. <strong>A layer 2 is a separate blockchain that extends Ethereum</strong>. How does that work?",
+  "layer-2-how-does-layer-2-work-2": "There are several different types of layer 2, each having their own trade-offs and trust assumptions. Layer 2s take the transactional burden away from the layer 1 allowing it to become less congested, and everything becomes more scalable.",
```
jmcook1186, line 44:
```suggestion
  "layer-2-how-does-layer-2-work-2": "There are several different types of layer 2, each having their own trade-offs and security models. Layer 2s take the transactional burden away from the layer 1 allowing it to become less congested, and everything becomes more scalable.",
```

src/intl/en/page-layer-2.json
```diff
@@ -40,24 +40,24 @@
...
   "layer-2-zk-rollups-title": "Zero-knowledge rollups",
   "layer-2-zk-rollups-description": "Zero-knowledge rollups use validity proofs where transactions are computed off-chain, and then compressed data is supplied to Ethereum Mainnet as a proof of their validity.",
   "layer-2-zk-rollups-childSentance": "More on ZK-rollups",
   "layer-2-dyor-title": "Do your own research: risks of layer 2",
-  "layer-2-dyor-1": "Since layer 2 chains inherit security from Ethereum, in an ideal world, they are as safe as L1 Ethereum. However, many of the <strong>projects are still young and somewhat experimental</strong>. After years of research and development, many of the L2 technologies that will scale Ethereum launched in 2021. Many projects still have additional trust assumptions as they work to decentralize their networks. Always do your own research to decide if you're comfortable with any risks involved.",
+  "layer-2-dyor-1": "Many layer 2 projects are relatively young and still have additional trust assumptions as they work to decentralize their networks. Always do your own research to decide if you're comfortable with any risks involved.",
```
jmcook1186, line 55:
```suggestion
  "layer-2-dyor-1": "Many layer 2 projects are relatively young and still require users to trust some operators to be honest as they work to decentralize their networks. Always do your own research to decide if you're comfortable with any risks involved.",
```

minimalsm (approved):

Issue #10905, opened 2023-08-09
Bug report: Error and success states in quiz component showing white bars in dark mode


corwintines:
### Describe the bug

![Screen Shot 2023-08-09 at 11 29 44 AM](https://github.com/ethereum/ethereum-org-website/assets/15589226/9e80f2aa-c802-4146-ba0d-eb3f0c32b9b7)

Whether answering correct or incorrect in the /quizzes/ page, there is a bug in dark mode where the we see white bars from the modal. 

We fixed this for the neutral state, but need to fix this in all cases: https://github.com/ethereum/ethereum-org-website/pull/10860

### To reproduce

1. Go to https://ethereum.org/en/quizzes
2. Do a quiz
3. Answer a question right or wrong

### Expected behavior

The background of the modal in the success and error states should be the same as the background of the quiz widget

### Screenshots

![Screen Shot 2023-08-09 at 11 29 44 AM](https://github.com/ethereum/ethereum-org-website/assets/15589226/c87bffbd-6410-42f6-b1e4-6bae4fd617cb)

### Would you like to work on this issue?

- [ ] Yes
- [ ] No

pettinarip:
Issue being fixed in #11423

F3n4nd0gm:
http://192.168.1.254
Passwords>
https://coinmarketcap.com/

PR #10907, opened 2023-08-09
Fix for issue #10905

checkomkar:
## Description

Inconsistencies in background color of quiz Modal

Before:
image

After:
image

## Related Issue

#10905

### DIFF
--- a/src/@chakra-ui/gatsby-plugin/semanticTokens.ts
+++ b/src/@chakra-ui/gatsby-plugin/semanticTokens.ts
@@ -73,14 +73,14 @@ const semanticTokens = {
     // ? Keep "error" or rename to "fail" ?
     error: {
       base: "red.500",
-      light: "red.100",
+      light: { _light: "red.100", _dark: "red.900" },
       outline: { _light: "error.base", _dark: "error.light" },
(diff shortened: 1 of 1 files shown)


src/@chakra-ui/gatsby-plugin/semanticTokens.ts
```diff
@@ -73,14 +73,14 @@ const semanticTokens = {
...
-      light: "red.100",
+      light: { _light: "red.100", _dark: "red.900" },
       outline: { _light: "error.base", _dark: "error.light" },
       // ! Deprecating error.neutral
       neutral: { _light: "red.100", _dark: "red.900" },
     },
     success: {
       base: "green.500",
-      light: "green.100",
+      light: { _light: "green.100", _dark: "green.900" },
```
pettinarip, line 83: @nloureiro `red.900` and `green.900` were marked as deprecated in the code because we removed them from the DS. Since we are still using that color in the Quiz Hub modals, should we add it back to the DS?

Take a look at the original issue https://github.com/ethereum/ethereum-org-website/issues/10905 to have more context on the current problem.

PR #10912, opened 2023-08-11
fix image and link on /refi/ page

corwintines:
## Description

- Reduce size of image
- Change link to https://capitalinstitute.org/8-principles-regenerative-economy/ in intro paragraph


nloureiro:
Besides the fine-tuning of the image, should we add a caption to the image with credits? Where does this image come from?

@jmcook1186 probably, you can help on this one. 😀

nloureiro:
@corwintines I change the image to 450x450px, but on the dev preview it's not showing. Can you take a look?

src/assets/use-cases/refi-regenerative-economy-diagram.png
wackerow, line 1: image

image


This has lost image quality when previewing, which may be related to pixel ratios. Do we need to downsize the image file or could we just put a max size on it when rendered?

wackerow (changes requested):

src/assets/use-cases/refi-regenerative-economy-diagram.png
corwintines: I think since this is done through markdown, we might need to have it be 450px if that is our goal dimensions. Ill see if I can do this without losing quality.

corwintines:
I pushed a better res version of the 450x450px image. I think this is ready.

wackerow (approved):

PR #10928, opened 2023-08-14
Glossary tooltip

corwintines:
## Description

- Implements GlossaryDefinition, GlossaryTooltip, their stories
- Move glossary terms and definitions into a glossary.json translatable file
- Implements GlossaryDefinition on /glossary/ page

NOTE: Blocked by https://github.com/ethereum/ethereum-org-website/pull/10706

TODO:
- [x] Pablo changes
- [x] Helper cursor type
- [x] Text hover effect color
- [x] Tooltip arrow color needs to be fixed
- [x] Add tooltip animation for when it pops up
- [x] Finish breakout glossary
- [x] Bring GlossaryTooltip into all templates
- [x] Story with multiple link types to have all links in a single story
- [x] Nice to have: test load time regressions
- [ ] Implement glossary tooltip on current `/glossary/#` uses in the codebase


src/components/Glossary/GlossaryDefinition/index.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react"
+import { Box, Heading, Text } from "@chakra-ui/react"
+
+import Translation from "../../Translation"
+
+interface IProps {
+  term: string
+  tooltip?: boolean
```
pettinarip, line 8: nitpick: for DX purposes, would rename this prop to something decoupled from implementation references and that express styling purposes. We could call this something like `size` which I think it is basically what we are doing here

```suggestion
  size?: "md" | "sm"
```

And then default it to `md`
```tsx
const GlossaryDefinition: React.FC<IProps> = ({ term, size = "md" }) => {
```

From the consumer's side, you would do
```tsx
<Tooltip content={<GlossaryDefinition term={hash!} size="sm" />}>
```

Now, other devs could use GlossaryDefinition in a more natural and easy to understand way.

pettinarip (commented):
Great start @corwintines

I've tested it and its working perfectly.

Left a couple of comments but I also wanted to add a few more thoughts:

- We should run some heavy loading tests to see that this solution scales to a reasonable level (to support at least the amount of terms we have in mind for this epic)
- Check that client loading times are not negatively affected by this change
- I'd argue against having `cursor="pointer"` in our `Tooltip` component. Makes me think that it is a link, which is incorrect. cc @konopkja @nloureiro

I can help with the first two.

nloureiro:
The hover color is not changing. Is it just me?

https://github.com/ethereum/ethereum-org-website/assets/1120748/1cdc2385-0d47-4524-8df0-5611cf1f64bc

nloureiro:
I think I found another tiny bug; the bubble's arrow is white and should be a background.highlight. same as the tooltip background

Screen Shot 2023-08-16 02 23 54 PM

pettinarip (commented):
Noted one minor potential visual issue.

This is how I see the underline in my laptop (zoomed)
image

This is how it looks inside a paragraph
image

From the designs and screens that you have shared, it seems like in other browsers or OS this looks a bit different. It seems like it has more space between the text and the dotted line.

We could tweak that space by adding [`text-underline-offset`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-offset)

cc @nloureiro

src/intl/en/glossary.json
```diff
@@ -0,0 +1,322 @@
+{
+  "51%-attack-term": "51% attack",
+  "51%-attack-definition": "A type of attack on a decentralized <a href=\"#network\">network</a> where a group gains control of the majority of <a href=\"#node\">nodes</a>. This would allow them to defraud the blockchain by reversing <a href=\"#transactions\">transactions</a> and double spending <a href=\"ether\">ether</a> and other tokens.",
```
pettinarip, line 3: Noting that a lot of definitions have links inside. Most of those links are pointing to other definitions like `/#transactions` but those only exist on the glossary page, so they will not work when used inside the tooltips.

Maybe we could modify those links to always point to the glossary page.
For example, `#transactions` => should be transformed to `/glossary/#transactions`

src/components/Glossary/GlossaryTooltip/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+          textDecorationColor: "primary.hover",
+          color: "primary.hover",
+        }}
+        cursor="help"
+      >
+        {children}
+      </Text>
+    </Tooltip>
+  ) : (
+    <Text>{children}</Text>
```
pettinarip, line 30: On small resolutions, this will create a `p` tag which is a block element and will create some problems like:
image
(the text `dilemma` is the one wrapped with the tooltip)

Proposal: use an inline element
```suggestion
    <Text as="span">{children}</Text>
```

src/intl/en/glossary.json
corwintines: Good catch, yeah will update this

corwintines:
@pettinarip @nloureiro will bring the same underline style that is in the Link component here

nloureiro:
Can we make the affordable to be all the shadow?

as it is, we need to go over the arrow to make the cursor go to the tooltip. We should make it all the length of the work

https://github.com/ethereum/ethereum-org-website/assets/1120748/1f6bce25-dab2-4cc4-952b-190f5a330b03


tested here in safari
https://ethereumorgwebsitedev01-glossarytooltip.gatsbyjs.io/en/defi/

nloureiro:
one very small change to the design.

it need to stand out a little bit
it's just adding this: `border: 1px solid var(--eth-colors-primary-lowContrast);`

Screen Shot 2023-08-23 09 18 52 AM


Screen Shot 2023-08-23 09 27 16 AM


the arrow do not need it, makes it more complex to implement with no need
Screen Shot 2023-08-23 09 28 19 AM

nloureiro:
> Noted one minor potential visual issue.
> 
> This is how I see the underline in my laptop (zoomed) image
> 
> This is how it looks inside a paragraph image
> 
> From the designs and screens that you have shared, it seems like in other browsers or OS this looks a bit different. It seems like it has more space between the text and the dotted line.
> 
> We could tweak that space by adding [`text-underline-offset`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-offset)
> 
> cc @nloureiro

interesting. I don´t see it like this.

but, yes we can force it to have     `text-underline-offset: 4px;` just to make sure.

@pettinarip can you test this on the dev tools to check if fixes it on your side?

Screen Shot 2023-08-23 09 39 14 AM

konopkja:
how can i view this? dont see it when i look at the preview deploy

nloureiro:
> how can i view this? dont see it when i look at the preview deploy

on the defi page
https://ethereumorgwebsitedev01-glossarytooltip.gatsbyjs.io/en/defi/

konopkja:
not ready on phone yet? dont see it

pettinarip:
@nloureiro looks good now. It is set to 3px, not sure if you'd prefer to change it to 4px. LMK

image

nloureiro:
> @nloureiro looks good now. It is set to 3px, not sure if you'd prefer to change it to 4px. LMK
> 
> image

in this images is 3px ou 4px?

pettinarip:
> in this images is 3px ou 4px?

3px

nloureiro:
> > in this images is 3px ou 4px?
> 
> 3px

ok, let's go with 3px :)

minimalsm (commented):
What is the expected behaviour on non-English pages?

**English:**
Shows glossary tooltip:

Screenshot 2023-08-24 at 11 56 03


**German:**
Shows glossary link:

Screenshot 2023-08-24 at 11 55 55

minimalsm (commented):
One annoying thing (that we can probably fix separately)

The tooltip disappears when you exit the hover area into a 'gap' between tooltip and link

tooltip

nloureiro:
> One annoying thing (that we can probably fix separately)
> 
> The tooltip disappears when you exit the hover area into a 'gap' between tooltip and link
> 
> tooltip

Yes, this, for me, is not a minor thing. It's not a show-stopper, but I would like it fixed immediately after merging.

nloureiro:
> What is the expected behaviour on non-English pages?
> 
> **English:** Shows glossary tooltip:
> 
> Screenshot 2023-08-24 at 11 56 03
> **German:** Shows glossary link:
> 
> Screenshot 2023-08-24 at 11 55 55

Something that I did not test, but I'm ok with this behavior.

- Tooltip in English > link to glossary page as a fallback for other languages.

Can we assume we will not have both (tooltip and fallback link) in the same language?

 Is it possible? 

If yes, this might lead to having the tooltip only in English unless the X language has all the terms translated. 
Are we ok with this?

corwintines:
@konopkja there wasn't going to be mobile support for this feature, only larger screens. This was something that @pettinarip had pushed on when designing this. So at the moment, in mobile views it will only show the text and not the tooltip. Maybe the fallback here should be a link to the glossary instead? Either way, the decided experience was no tooltip in mobile when designing.

corwintines:
@nloureiro @minimalsm will look into the gap issue now

@nloureiro will also add that border.

corwintines:
@nloureiro @minimalsm the german version will be fixed in this PR where it migrates the codebase to use this https://github.com/ethereum/ethereum-org-website/pull/10980/files#diff-d53b94d5fd28ef81dedc92e0db688e723ad6f7d689980243c23fdee8b37de3af

The example of the defi page in this PR is just one implementation to test. Wanted to separate it out to make it a bit more clean in this PR.

corwintines:
@minimalsm @nloureiro I have added that border in and a transparent box to keep the open context when going between the term and tooltip. 

@minimalsm @nloureiro @konopkja @pettinarip I think all concerns have been addressed now either in this PR, or in the migration PR to implement the GlossaryTooltip here https://github.com/ethereum/ethereum-org-website/pull/10980

src/components/Tooltip.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { ReactNode, useState } from "react"
-import { Box, useColorModeValue } from "@chakra-ui/react"
+import { Box, useColorModeValue, useToken } from "@chakra-ui/react"
+import { motion, AnimatePresence } from "framer-motion"
 import * as utils from "../utils/isMobile"
+import { border } from "styled-system"
```
pettinarip, line 5: ✂️

pettinarip (approved):
Nicely done @corwintines lgtm

minimalsm (approved):
Ship it 🚀

src/content/defi/index.md
```diff
@@ -317,7 +317,7 @@ You can think of DeFi in layers:
 
 1. The blockchain – Ethereum contains the transaction history and state of accounts.
 2. The assets – [ETH](/eth/) and the other tokens (currencies).
-3. The protocols – [smart contracts](/glossary/#smart-contract) that provide the functionality, for example, a service that allows for decentralized lending of assets.
+3. The protocols – <GlossaryTooltip termKey="smart-contract">smart contracts</GlossaryTooltip> that provide the functionality, for example, a service that allows for decentralized lending of assets.
```
minimalsm, line 320: This

PR #10932, opened 2023-08-14
fix missed how-to-register to how-to-create in translations

corwintines:
## Description

- Fix `how-to-register-` to `how-to-create-` in translation files
- This above also fixes breadcrumb issues on https://ethereum.org/en/guides/how-to-create-an-ethereum-account/


TylerAPfledderer:
@corwintines Related or an extension of #10427 ?

corwintines:
@TylerAPfledderer I think an extension. This isn't adding a new one in the sense of that PR I think, but fixing a bug after changing this guides path. I think this is fine to have in isolation since it just touches the bug for the `How to "create" and ethereum account` guide

TylerAPfledderer:
@corwintines cool. Thank you for the clarification!

wackerow (approved):
Separate to this PR, would suggest we ditch the quotes around "create"... we added those for "register" as that's a term that people use but isn't exactly accurate, but I think we can do without for create.

Issue #10748, opened 2023-07-21
Accordion


pettinarip:
Create a new component called `Accordion` with [the DS styles](https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?node-id=1220%3A19589).

![image](https://github.com/ethereum/ethereum-org-website/assets/468158/9cc9917f-2e3b-498c-934e-257757d1f5d2)

amit-ksh:
Hey @pettinarip, can you assign this to me?

justynleung:
![ezgif com-video-to-gif](https://github.com/ethereum/ethereum-org-website/assets/114822505/676e3596-8b22-461f-b789-fa6b411d37eb)

@pettinarip Does this look good? I made a prototype according to the requirements.

pettinarip:
Hey @justynleung looks good! soon we will review the PR and implementation. Thanks for the help

justynleung:
@pettinarip No worries. I am glad that you like it. Please let me know if any change is needed.

Issue #10744, opened 2023-07-21
SideNav


pettinarip:

Nagapadmini7:
I can work on this issue

ameeetgaikwad:
@pettinarip would like to work on this issue

pettinarip:
@ameeetgaikwad pls consider using the [Accordion](https://chakra-ui.com/docs/components/accordion/usage) for this implementation

ameeetgaikwad:
@pettinarip Not able to see much difference in the current implementation and the DS one. anyways here I have used the Accordion from chakra, let me know if you need any changes.

![image](https://github.com/ethereum/ethereum-org-website/assets/76063569/038a8389-7452-4569-9c2c-7a5f86a8d55b)

ameeetgaikwad:
@TylerAPfledderer so sorry, I'll delete that comment.

pettinarip:
Correct. Good point @TylerAPfledderer. This PR will be on hold until we merge the Accordion implementation.

F3n4nd0gm:
https://github.com/ethereum/ethereum-org-website/assets/147960298/494a6cd2-cf19-4eee-9f02-5fb041e35ed4

F3n4nd0gm:
![pettinarip.github.io-develop.zip](https://github.com/ethereum/ethereum-org-website/files/13061848/pettinarip.github.io-develop.zip)

wackerow:
Circling back, this hasn't been forgotten... #10936 has had a couple updates, trying to get that in then will come back to this one.

corwintines:
Removing stale tag since this has a draft PR in progress

minimalsm:
Reassigning this to @pettinarip as he has an open (draft) PR here.

PR #10936, opened 2023-08-15
Accordion made according to DS style in figma [issue #10748]

justynleung:
## Description
The accordion component is made with Chakra UI and strictly followed figma description.

The component is located at src/components/Accordions. 
It supports light/dark mode, correct hover and focus effect with desired color.
The rotation of the accordion arrow icon is same as the design in Figma
It is mobile responsive according to the requirements in Figma.

## Related Issue
The only criticism I would have is that in dark mode the bg color on hover is with low contrast with text color. 
Besides, I am not sure about what is "active: top level" "active: last level" about in the Figma description. 

[issue #10748]


TylerAPfledderer:
Hey @justynleung! Thanks for your assistance and additional thoughts. 😄 

In order for this to be properly reviewed and approved, any component new or existing that is directly affected by the new DS needs to have at least one storybook story (unless otherwise), and following a proposed [Storybook structure outlined in this figma file](https://www.figma.com/file/Ne3iAassyfAcJ0AlgqioAP/DS-to-storybook-structure?type=design&node-id=42%3A92&mode=design&t=2Eazmr5Dep7ZhwKN-1). This component is under `Molecules / Disclosure Content / Accordions`.

Follow the [Applying Storybook](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/applying-storybook.md) documentation and let us know if you have any further questions. 😄

I will still help review what you currently have. 👍🏼

TylerAPfledderer (changes requested):
Left some additional thoughts.

With this project, import/export of components are not handled in the same way as you find in a component library.

Since a storybook file should also be added, this `Accordions` folder can stay, and the component file itself would be called `index.tsx` and not export twice with a secondary file, as you currently have it.

So the structure would be:
```
src/
└── components/
    └── Accordion/
        ├── index.tsx
        ├── Accordion.stories.tsx
```

NOTE: it should be `Accordion` and not `Accordions`, and give the Chakra import the alias `ChakraAccordion`

---

With your question regarding the active level reference in the figma file, Nuno is referring to the active style that would be applied to a nested accordion versus the root level accordion. There is a pending discussion in the file with this issue, as the Chakra component doesn't handle styling like this... at least in a straight-forward manner. Honestly, I haven't found a clean way around it!

If you can see a solution here, go for it! 💪🏼  Maybe I missed something.

src/components/Accordions/Accordions.tsx
```diff
@@ -0,0 +1,59 @@
...
+  AccordionItem,
+  AccordionButton,
+  AccordionPanel,
+  AccordionIcon,
+  Box,
+} from "@chakra-ui/react"
+
+const Accordions = () => {
+  const { toggleColorMode } = useColorMode()
+  const primary = useColorModeValue("#8282FF", "#FFE3D3")
```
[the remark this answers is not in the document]
justynleung: @TylerAPfledderer 

No worries. 

This is my first time contributing so it may take some time to understand all the logics. Hope you don't mind.

I will read the extra information you provided and let you know when I am done.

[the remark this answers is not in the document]
TylerAPfledderer: @justynleung no problem here! 😄  Me and @pettinarip are certainly aware of some documentation that can be created to further support contributing while maintaining consistency in the codebase and the desired workflow. 

In the meantime, you are always welcome to reach out for any clarification!

src/components/Accordion/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+  AccordionItem,
+  AccordionButton,
+  AccordionPanel,
+  AccordionIcon,
+  Box,
+} from "@chakra-ui/react"
+
+const Accordion = () => {
+  const { toggleColorMode } = useColorMode()
+  const primary = useColorModeValue("#8282FF", "#FFE3D3")
```
pettinarip, line 14: Adding to what Tyler said and replying to your original question, the color was fixed in figma file, we should use the semantic token [`primary.lowContrast`](https://github.com/ethereum/ethereum-org-website/blob/dev/src/@chakra-ui/gatsby-plugin/semanticTokens.ts#L42) in this case.

pettinarip (commented):
Thanks @justynleung! great start!

I agree with @TylerAPfledderer that we should create a new Accordion theme file to define and override the styles we want from the DS.

Probably there is no need to create a new component for this as we want users to keep using all the exported Accordion related components, to give them more flexibility.

Having said that, probably we would want to:
1. create a new Accordion theme file
2. remove the custom Accordion component
3. move the generated story to `/src/components/BaseStories`

========

Regarding the "active top level" and "active last level", I think we could add those styles as variants and let the consumers decide when to use/apply them.

cc @TylerAPfledderer

src/components/Accordion/index.tsx
justynleung: Sure, I will fix this as soon as possible. Thanks for letting me know.

TylerAPfledderer:
Hey @justynleung! Any updates with this?

justynleung:
Hi Tyler, I have been working on this. Please expect an update around mid
Nov. I am sorry about the delay.

justynleung:
> Hey @justynleung! Any updates with this?

Hi @TylerAPfledderer , I have made some changes according to the suggestion by you and @pettinarip. Please let me know if there is any issue. Again, sorry for the delay and I am exciting to update more frequently here.

wackerow:
Thanks @justynleung! I just pushed a commit with a couple updates.. mainly just cleaning up unused/unnecessary elements and switching to use Chakra-UI tokens wherever possible.
Updated the props as well to accept a label and children; moved the story into the component's own folder and added some default meta `arg` values for the props.

cc: @TylerAPfledderer @pettinarip for final review

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export default meta
+
+type Story = StoryObj<AccordionType>
+
+export const Basic: Story = {
+  render: (args) => (
+    <VStack>
+      <Accordion {...args} />
+    </VStack>
```
TylerAPfledderer, line 24: I would lean towards matching what we have in the figma file. Starting here. `VStack` is not the best as a wrapper because it wants to center align its content, when we want the accordion to stretch the width of the container to get the best visual representation of the expected design. Use `Box` instead.

Also have a defined width of the container.

```suggestion
    <Box width="full" maxWidth="268px">
      <Accordion {...args} />
    </Box>
```

src/components/Accordion/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  AccordionIcon,
+  Box,
+} from "@chakra-ui/react"
+
+type AccordionProps = {
+  label: string
+  children: React.ReactNode
+}
+
+const Accordion = ({ label, children }: AccordionProps) => (
```
TylerAPfledderer, line 15: @nloureiro do we have a defined minimum width for the accordion? Or should we define a minimum gap between the button text and the arrow icon?

src/components/Accordion/Accordion.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Meta, StoryObj } from "@storybook/react"
+import { VStack } from "@chakra-ui/react"
+
+type AccordionType = typeof Accordion
+
+const meta: Meta<AccordionType> = {
+  title: "Molecules / Disclosure Content / Accordions",
+  component: Accordion,
+  args: {
+    label: "Accordion label",
```
TylerAPfledderer, line 11: This is another item to match the figma file, where we should be using the same text.

```suggestion
    label: "Label text of the accordion",
```

src/components/Accordion/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+            _hover={{ bg: "primary.lowContrast" }}
+            _expanded={{ bg: "primary.lowContrast" }}
+            _focus={{
+              border: "4px",
+              borderColor: "primary.lowContrast",
+              borderRadius: "md",
+            }}
+            py="2"
+            px={{ base: 2, md: 4 }}
+          >
```
TylerAPfledderer, line 33: All of this styling should still move to a component theme file, as there is an [imported default theme]() that could cause specificity issues.

Also, with the focus styling, it should be applied to the `outline` and not `border`, else you will get a layering when tabbing to the button, which will not comply with the DS.

TylerAPfledderer (changes requested):
@wackerow netlify build is failing due to eslint errors. Need to run `yarn lint:fix` and push those updates.

The only other thing missing here that I notice is addressing nested accordion design. 

And I've added additional thoughts. :)

src/components/Accordion/index.tsx
nloureiro: it should be the minimum gap, in this case 0.5rem

something like this?
Screen Shot 2024-03-27 12 33 25 AM

TylerAPfledderer: @nloureiro ok! Or `2` as the Chakra token value here.

I did not see this defined gap in figma; it was only being marked as "auto"

nloureiro: true.... figma limitation here. It assumes a "flex box" set to auto OR a gap fixed spacing. (without making it too complex on Figma. I could have made another layer on top with the fixed gap built-in)
 on code, we can have both.
I should have done a better spec. Sorry about this...

TylerAPfledderer: Oh interesting! Good to know, no problem 👍

wackerow:
Pushed a couple commits, extracting the styles to a theme file and updating the story label. Not sure how we want to do the "active" or "disabled" styling I see in Figma... I don't see a way to "disable" these, and by "active" do we mean "expanded"?

pettinarip:
> On the stablecoins: can it have the same animation open/close?

Fixed.

> Did you change the accordions on the dev docs left bar?

No, that one is not using the chakra accordion for now. Is going to be part of #10744

nloureiro:
Screen Shot 2024-04-30 02 29 05 PM

@pettinarip spotted a bug > The chevron should be pointing down on the find wallet accordion

nloureiro:
@pettinarip I have another question: 

I'm curious about the focus that appears when opening and closing with the mouse. It should be visible only when navigating with the keyboard.

Screen Shot 2024-04-30 02 35 51 PM

Is it on purpose for accessibility? Is it a chakra thing?

We don't have this in production, and it's annoying when using the mouse to open and close. Not sure what can be done. (this happens on Brave and Firefox, but not Safari)

nloureiro:
> @pettinarip I have another question:
> 
> I'm curious about the focus that appears when opening and closing with the mouse. It should be visible only when navigating with the keyboard.
> 
> Screen Shot 2024-04-30 02 35 51 PM
> 
> Is it on purpose for accessibility? Is it a chakra thing?
> 
> We don't have this in production, and it's annoying when using the mouse to open and close. Not sure what can be done. (this happens on Brave and Firefox, but not Safari)

Same here.... the blue outline is persistent on the FAQ, too
Screen Shot 2024-04-30 02 42 19 PM

TylerAPfledderer:
> > @pettinarip I have another question:
> > 
> > I'm curious about the focus that appears when opening and closing with the mouse. It should be visible only when navigating with the keyboard.
> > 
> > Screen Shot 2024-04-30 02 35 51 PM
> > 
> > Is it on purpose for accessibility? Is it a chakra thing?
> > 
> > We don't have this in production, and it's annoying when using the mouse to open and close. Not sure what can be done. (this happens on Brave and Firefox, but not Safari)
> 
> Same here.... the blue outline is persistent on the FAQ, too
> Screen Shot 2024-04-30 02 42 19 PM
> 

@nloureiro the accordion triggers (these visible headers with the Chevron) are button elements. It is expected behavior in fundamental HTML to have a click event trigger focus state.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus

pettinarip:
@nloureiro @TylerAPfledderer pushed a commit that should fix the visual outline on mouse clicks and align with the [chakra examples](https://github.com/chakra-ui/chakra-ui/blob/v2/packages/theme/src/components/accordion.ts#L22), basically by using `_focusVisible` instead of `_focus`.

nloureiro:
> > > @pettinarip I have another question:
> > > I'm curious about the focus that appears when opening and closing with the mouse. It should be visible only when navigating with the keyboard.
> > > Screen Shot 2024-04-30 02 35 51 PM
> > > Is it on purpose for accessibility? Is it a chakra thing?
> > > We don't have this in production, and it's annoying when using the mouse to open and close. Not sure what can be done. (this happens on Brave and Firefox, but not Safari)
> > 
> > 
> > Same here.... the blue outline is persistent on the FAQ, too
> > Screen Shot 2024-04-30 02 42 19 PM
> 
> @nloureiro the accordion triggers (these visible headers with the Chevron) are button elements. It is expected behavior in fundamental HTML to have a click event trigger focus state with buttons.
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#clicking_and_focus

yes, but the focus should stay focused, does it? on the spec...

nloureiro (approved):

nloureiro:
> @nloureiro @TylerAPfledderer pushed a commit that should fix the visual outline on mouse clicks and align with the [chakra examples](https://github.com/chakra-ui/chakra-ui/blob/v2/packages/theme/src/components/accordion.ts#L22), basically by using `_focusVisible` instead of `_focus`.

approved! 🤘🤘🤘🤘🤘🤘🤘🤘🤘

pettinarip (approved):

PR #10949, opened 2023-08-16
Fixed typo on staking withdrawals page

woseK:
## Description

Fixed typo on staking withdrawals page, but dont know if the "prepared" was necessary so I just deleted it. If it was necessary, then sorry 😅


corwintines (approved):
Thanks @woseK

PR #10962, opened 2023-08-18
Add ChainIDE to Ethereum support-tooling

LyingZz:
## Description

[ChainIDE](https://www.chainide.com) is a multi-chain cloud-powered development platform that provides one-stop development services for blockchain developers. 
We want to add ChainIDE to Ethereum support-tooling.


minimalsm:
Hi! Thanks for the PR. I don't think we want to add ChainIDE to this page. To be clear, the page's traffic primarily comes from people searching for "Ethereum support" on search engines, and we decided to add a list of popular products to illustrate that although Ethereum itself is decentralised and has no 'official' support, individual products do. 

We have some re-thinking to do of this page, but I don't think this addition helps our aims here.

Thanks again for the PR, and your patience in awaiting a response here :).

PR #10964, opened 2023-08-18
Fix typos

jimgreen2013:


src/content/developers/docs/nodes-and-clients/index.md
```diff
@@ -56,7 +56,7 @@ If you want to [run your own node](/developers/docs/nodes-and-clients/run-a-node
...
+Full nodes do a block-by-block validation of the blockchain, including downloading and verifying the block body and state data for each block. There are different classes of full node - some start from the genesis block and verify every single block in the entire history of the blockchain. Others start their verification at a more recent block that they trust to be valid (e.g. Geth's 'snap sync'). Regardless of where the verification starts, full nodes only keep a local copy of relatively recent data (typically the msot recent 128 blocks), allowing older data to be deleted to save disk space. Older data can be regenerated when it is needed.
```
minimalsm, line 59:
```suggestion
Full nodes do a block-by-block validation of the blockchain, including downloading and verifying the block body and state data for each block. There are different classes of full node - some start from the genesis block and verify every single block in the entie history of the blockchain. Others start their verification at a more recent block that they trust to be valid (e.g. Geth's 'snap sync'). Regardless of where the verification starts, full nodes only keep a local copy of relatively recent data (typically the msot recent 128 blocks), allowing older data to be deleted to save disk space. Older data can be regenerated when it is needed.
```

minimalsm (approved):
Thanks for the PR @jimgreen2013.

Really sorry for the delay on getting to this. Looks like the change has been solved already, but I'll merge this PR anyway to give you credit ❤️ 

Thanks again!

minimalsm:
@all-contributors please add @jimgreen2013 for docs

PR #10973, opened 2023-08-20
Fix portal network link in light client index.md

arikg:
* Fix the link to the portal network in the light client documentation to point to the relevant article on the docs - currently broken link


corwintines:
Thanks @arikg 

It seems like the portal networks website is working. Not sure if it was down when you opened this PR, but since its working I will keep it as is. Going to close this out.

PR #10974, opened 2023-08-21
update tooltip component based on new DS

### LINKED ISSUE #10747: Tooltip
Implement the [new designs](https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?node-id=1069%3A17908) for the [Tooltip component](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Tooltip.tsx).

image

amit-ksh:
## Description

Updated the **tooltip** component according to the new design system using the Chakra **Tooltip** component.

## Related Issue

#9548 
Closes #10747 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `Popover` component for Chakra UI with customizable styles and configurations.
	- Added a new `Tooltip` component that displays information on hover, utilizing the Chakra UI `Popover`.

- **Enhancements**
	- Updated various components (`DencunBanner`, `EthPriceCard`, `WalletMoreInfoCategory`, `Simulator`, `StakingStatsBox`, `GridItem`) for improved layout and styling, including better use of Chakra UI components like `Box`, `Icon`, and tooltips.
	- Enhanced the `Tooltip` component's content rendering and configuration options.

- **Code Quality**
	- Simplified popover content and styling in `Simulator` components for cleaner code and better user experience.
	- Adopted `forwardRef` in `Address` component for improved component reference handling.

- **Documentation and Examples**
	- Added storybook stories for the new `Tooltip` component to demonstrate usage and configuration.


TylerAPfledderer:
Hey @amit-ksh!

In order for this to be properly reviewed and approved, any component new or existing that is directly affected by the new DS needs to have at least one storybook story (unless otherwise), and following a proposed [Storybook structure outline in this figma file](https://www.figma.com/file/Ne3iAassyfAcJ0AlgqioAP/DS-to-storybook-structure?type=design&node-id=42%3A92&mode=design&t=2Eazmr5Dep7ZhwKN-1). This component is under `Molecules / Overlay Content / Tooltips`.

Follow the [Applying Storybook](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/applying-storybook.md) documentation and let us know if you have any further questions. 😄 

I will still help review what you currently have. 👍🏼

src/components/Tooltip.tsx
```diff
@@ -1,85 +1,112 @@
-import React, { ReactNode, useState } from "react"
-import { Box, useColorModeValue } from "@chakra-ui/react"
+import React, { ReactNode } from "react"
```
TylerAPfledderer, line 1: Unfortunately, this is not going to be the desired result in the creation of the tooltip.

The primary focus with creating or recreating these components is to utilize what Chakra UI provides to simplify the need for building custom logic. There is a [Tooltip](https://chakra-ui.com/docs/components/tooltip/usage) component available, which means at least one of two things will need to occur:

---
First and foremost is creating a tooltip theme config in `src / @chakra-ui / gatsby-plugin / components` to handle the custom styling set by the Design System. Check out the `component.utils.ts` file in this sub-directory and the other component theme files to see how the default component theming from Chakra is applied.

Secondly, if the tooltip needs additional reusable logic that is not already built into Chakra's `Tooltip`, then the component file used is simply a wrapper of the same name with Chakra's component imported as `ChakraTooltip` and rendered accordingly.

```tsx
import { Tooltip as ChakraTooltip } from '@chakra-ui/react'

const Tooltip = (props) => {
  // Any custom logic here

  /* 
   * Can be a simple return like below, 
   * unless further structure needs to be created for the custom logic
   */ 
  return <ChakraTooltip />
}

export default Tooltip
```

---
If no additional reusable logic is needed, then no component file is needed. This means we only import the Tooltip straight from the Chakra package any time it's needed and have the styling isolated in the theming config. In this case, a storybook story would still need to be made, but this new file will reside in `src / components / BaseStories`.

Have you tried any of the above? Let me know if you are running into trouble and I can try to assist. 🙂

TylerAPfledderer (changes requested):
Only one additional thought.

src/components/Tooltip.tsx
amit-ksh: `Tooltip` also don't allow to focus on Tooltip content using keyboard

amit-ksh: Thank you for review @TylerAPfledderer

I tried the `Tooltip` component from Chakra but in DS, the tooltip is shown as not opened on focus that's why I used it because if I want to add some custom logic to the `Tooltip` component then all the Chakra logic for tooltip is overridden by it. If `Tooltip` is going to opened on focus then I think I can use Chakra Tooltip without custom logic.

Another catch with Chakra `Tooltip` is that it uses `Portal` which makes the hard to navigate using Keyboard.

amit-ksh: Hi @TylerAPfledderer, I have created the **Tooltip** theme config as per your request and will be working on the story shortly. However, I'm facing an issue with setting the `z-index` of the tooltip content despite using `PortalManager`. Do you have any suggestions on how to resolve this?

amit-ksh: IMG_20230830_194956

The z-index of tooltip content is higher than the navbar. in the main site, tooltip is hidden from the navbar.

TylerAPfledderer: @amit-ksh this is actually "expected" behavior of this component... and by expected, I mean that if you checked out the [Placement section of the Chakra Docs](https://chakra-ui.com/docs/components/tooltip/usage#placement) and checked this with the "top" button in the example, you will see this exact same thing!

For now, I would ignore this behavior and I will make a note to check in with the Chakra core team to check if this is intended.

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export default meta
+
+type Story = StoryObj<TooltipType>
+
+export const Tooltip: Story = {
+  render: () => (
+    <TooltipComponent content="Text here">
+      <Icon fill="'text" name="info" />
```
TylerAPfledderer, line 20: This should render a specific component, and just use the `Icon` component. Probably...
```suggestion
      <Icon as={ RiInformationLine } />
```

No need to provide a `name` here. And because Chakra's `Icon` component automatically renders `fill="currentColor"` then any color should be established from the parent only via the `color` prop. (Especially if any text or icon rendered were to use the same color styles anyway!)

src/@chakra-ui/gatsby-plugin/components/Tooltip.ts
```diff
@@ -0,0 +1,31 @@
...
+  cssVar,
+  defineStyle,
+  defineStyleConfig,
+  getToken,
+} from "@chakra-ui/react"
+
+const $arrowBg = cssVar("popper-arrow-bg")
+
+const baseStyle = defineStyle((props) => {
+  const tableBoxShadow = getToken("colors", "tableBoxShadow")(props.theme)
```
TylerAPfledderer, line 11: This will not work. It looks like you are trying to grab a token that resides in the old `theme.ts` file, but Chakra will never see this file.

Because this shadow might only be used for the tooltip, I would suggest simply hardcoding the value here. Per the Figma file, only one `boxShadow` value is going to be used in both light and dark mode.

TylerAPfledderer (changes requested):
Here are my thoughts so far! 😄 

Primarily, I believe the feature branch here should be updated to the latest version of `dev` and looked over.

There is a new `GlossaryTooltip` component that is in prod. It is now clear compared to the Figma file that the tooltip component's usage is for specific text in body copy as well as with the info icon that is rendered with content like short statistical information.

I wonder then, if this should be renamed `InfoTooltip` and reworked so that it only renders the info icon?

If this can be better established, then can comment further on styling issues involving the inner wrapper (the `Center` component).

cc @pettinarip what do you think?

pettinarip (changes requested):
Thanks @amit-ksh 💪🏼 

While I was reviewing this I noticed that the Tooltip from Chakra might not be the ideal component to use for our tooltips since we use them a bit differently.

1. we want to put rich content inside the tooltip (like links), not just plain text as it is expected in the Chakra tooltip
2. we want to keep it open when the user hovers over the tooltip content

That behavior sounds like the `Popover` is a better fit for this use case. Should be used like this:

```tsx
<Popover trigger="hover">
  ...
</Popover>
```

@TylerAPfledderer what do you think of that? or is it possible to control the Tooltip and achieve something like that?

============

Another thing to mention is that we probably don't need to create a new component for this. We probably just need to create a theme file to change the default styles to the ones we want from the DS.

pettinarip:
> Here are my thoughts so far! 😄
> 
> Primarily, I believe the feature branch here should be updated to the latest version of `dev` and looked over.
> 
> There is a new `GlossaryTooltip` component that is in prod. It is now clear compared to the Figma file that the tooltip component's usage is for specific text in body copy as well as with the info icon that is rendered with content like short statistical information.
> 
> I wonder then, if this should be renamed `InfoTooltip` and reworked so that it only renders the info icon?
> 
> If this can be better established, then can comment further on styling issues involving the inner wrapper (the `Center` component).
> 
> cc @pettinarip what do you think?

We can keep the `GlossaryTooltip` as it is for now.

My idea is to create this "base" tooltip with the base styles and then refactor the `GlossaryTooltip` accordingly.

Also, this Tooltip should be able to be used anywhere, in any context. I mean, not only on body copy or icons, we should wrap a Card with the tooltip if we would like.

Note: perhaps we should start calling this Popover rather than Tooltip :P

TylerAPfledderer:
@pettinarip ah, right. Yea the popover is needed to be able to navigate to that rich content via keyboard.

src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <PopoverTrigger>
+          <Center
+            as="span"
+            tabIndex={0}
+            p={1}
+            cursor="pointer"
+            color="body.base"
+            _hover={{
+              color: isMobile ? "primary.hover" : "primary.base",
+            }}
```
TylerAPfledderer, line 33: Per the figma file, the `primary.hover` is marked for the `:active` state, which makes sense in mobile because you have to touch the icon in order to get the popover to appear.

Instead of trying to spend a check, I would use `_active` pseudo prop. This would be semantically correct and follows the [W3C Specifications](https://drafts.csswg.org/selectors/#the-active-pseudo).

```suggestion
            _hover={{
              color: "primary.base",
            }}
            _active={{
              color: "primary.hover",
            }}
```

@pettinarip I'm interpreting this as the `:active` color is showing onPress, and then when releasing the press it switches to the `:hover/open` color per the figma file labelling.

TylerAPfledderer (changes requested):
One quick item from me at the moment.

src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Tooltip: Story = {
+  args: {
+    content: "text here",
+    children: <Icon as={RiInformationLine} />,
+  },
+}
```
pettinarip, line 23: Could you add a new story that shows how to use rich content (e.g. a text with links)?

pettinarip (changes requested):
Looking good @amit-ksh. I've added a few minor comments.

I think that overall, the new component looks good but is not ready to be used in all the pages we have. We need to tweak a few things from the consumer's side before we can use the new Tooltip.

- [ ] For that reason, lets revert the `src/components/Tooltip.tsx` removal from this PR.

In a different issue/PR, we will do the transition from the old tooltip to the new tooltip, which is going to involve some style adjustments + more tests across the different pages that are involved.

src/components/Tooltip/Tooltip.stories.tsx
amit-ksh: Done!

amit-ksh:
> Looking good @amit-ksh. I've added a few minor comments.
> 
> I think that overall, the new component looks good but is not ready to be used in all the pages we have. We need to tweak a few things from the consumer's side before we can use the new Tooltip.
> 
> * [ ]  For that reason, lets revert the `src/components/Tooltip.tsx` removal from this PR.
> 
> In a different issue/PR, we will do the transition from the old tooltip to the new tooltip, which is going to involve some style adjustments + more tests across the different pages that are involved.

Hi @pettinarip, I reverted the removal of the old `Tooltip.tsx`. It's conflicting with the new one. Do you want me to rename the component and story file or should I do something else?

pettinarip:
> Hi @pettinarip, I reverted the removal of the old `Tooltip.tsx` locally. Now, it's conflicting with the new one. Do you want me to rename the component and story file or should I do something else?

Got it. I've solved the conflicts and reverted it.

I'll wait the build to finish, test again if nothing breaks, and then, we should be good to go 👍🏼

src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+} from "@chakra-ui/react"
+
+export interface IProps {
+  content: ReactNode
+  children?: ReactNode
+}
+
+const Tooltip: React.FC<IProps> = ({ content, children }) => {
+  return (
+    <Box title="More Info!" display="inline-block" ml={2}>
```
pettinarip, line 20: We've removed this title from the tooltip
```suggestion
    <Box display="inline-block" ml={2}>
```

pettinarip (changes requested):
@amit-ksh I've added one new story and fixed the imported Tooltip to use the newly added version and not the old one (I think that was the conflict you were talking about).

======
Also, noting that the arrow looks wider in the DS

image

DS version is 22px wide
image

amit-ksh:
@pettinarip any update?

pettinarip:
> @pettinarip any update?

I've pushed some fixes on the styles and ended up implementing the new Tooltip throughout the codebase.

pettinarip:
Reviving this PR.

I've fixed the tooltip theme config to be consistent with all the tooltips we have across the site.

@konopkja this PR fixes the positioning & zindex issues we have with the current custom component.

pettinarip (approved):
🚀

konopkja:
@pettinarip i noticed in the PR that on find-wallet page some tooltips adopt wrong styles. 
Screenshot 2024-03-11 at 15 12 28

pettinarip:
> @pettinarip i noticed in the PR that on find-wallet page some tooltips adopt wrong styles. What should we do with this? 

I see. I'll create a fix for this soon 👍🏼

PR #10978, opened 2023-08-22
Interactive simulator widget + "How to use a wallet" series

wackerow:
## Description
Implements a new interactive tutorial feature.

- `/components/Simulator/index.tsx` contains the root of the component
- The parent component must pass an object with the simulation data, allowing reusability of the component in the future.
- To start, a three-part series of tutorials are being added related to **wallet onboarding**, including:
  - Creating an account
  - Sending and receiving tokens
  - and Connecting to Web3
- Data for this series is stored in `/data/WalletSimulatorData.tsx`
- This series is being rendered on the `/wallets/` page.
  - Currently, feature only available in English. This component replaced the "Compare wallets based on features" section currently in place, but has a conditional to continue showing that section on non-English pages until translations for this component are rolled out (external to this PR).
- The `/Simulator/index.tsx` component renders a starting-point, which provides buttons to begin each lesson/path.
- Clicking a path will open a modal containing the guided interactive tutorial.
- Inside this modal, "Explanation" information is displayed, alongside "Screens" that are shown inside a phone-shaped figure.
  - The data for these explanations and screens are passed from the data provided by original parent, and render based on which "step" the user is on.
  - Step navigation tracking the step number is handled by the `/Simulator/index.tsx` component. This data and associated helper functions are passed to children as needed.

### In-page element that acts as a starting point:
image

### Modal
image
image

Mobile:
image

## Preview URL
- https://ethereumorgwebsitedev01-simulator.gatsbyjs.io/en/wallets/


src/components/Simulator/StartingPoint.tsx
```diff
@@ -0,0 +1,194 @@
...
+    clearUrlParams()
+    onClose()
+  }
+
+  const totalSteps: number = pathId
+    ? simulatorData[pathId].explanations.length
+    : 0
+
+  const PATH_ID_QUERY_PARAM = "sim" as const
+  const STEP_QUERY_PARAM = "step" as const
```
pettinarip, line 28: These constants can be defined outside of the function component.

src/components/Simulator/StartingPoint.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  const clearUrlParams = (): void => {
```
pettinarip, line 30: This function can be defined outside of the component.

src/components/Simulator/StartingPoint.tsx
```diff
@@ -0,0 +1,194 @@
...
+                handleClick={() => openPath(pathId as PathId)}
+              />
+            )
+          })}
+        </Flex>
+      </Flex>
+      <SimulatorModal isOpen={isOpen} onClose={handleClose}>
+        {state && simulator && (
+          <Template
+            state={state!}
```
pettinarip, line 184:
```suggestion
            state={state}
```

src/components/Simulator/StartingPoint.tsx
```diff
@@ -0,0 +1,194 @@
...
+
+  const openPath = (pathId: PathId): void => {
+    resetStepper()
+    setPathId(pathId)
+    onOpen()
+  }
+
+  const state: SimulatorState | null = pathId
+    ? {
+        pathId,
```
pettinarip, line 96: Already saw your TODO comments about cleaning this but just to make it clearer, `pathId` shouldn't be an interest to any of the children.

src/components/Simulator/ProgressCta.tsx
```diff
@@ -0,0 +1,108 @@
...
+  createIcon,
+  Flex,
+  type FlexProps,
+  Text,
+  ButtonProps,
+} from "@chakra-ui/react"
+import { motion } from "framer-motion"
+import type { SimulatorStateProps } from "./interfaces"
+import { simulatorData } from "./data"
+// TODO: Remove simulatorData import; pass data needed as props
```
pettinarip, line 14: 👍🏼

pettinarip (commented):
Great job @wackerow. This looks amazing.

Left a comment and some thoughts mainly on the entry point of the simulator. The controller. I think there are a few things that we could improve but I wouldn't consider them as blockers.

The only thing I see as a blocker is the direct access to the `simulatorData` file from `ProgressCta` but I've already seen that you have a TODO there 👍🏼

src/components/Simulator/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const CREATE_ACCOUNT = "create-account" as const
+export const SEND_RECEIVE = "send-receive" as const
+export const CONNECT_WEB3 = "connect-web3" as const
```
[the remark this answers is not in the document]
wackerow: hm, I think you're right that they aren't actually needed here... Got in the habit of using it to make sure we can use those explicit strings as the type here (for use in Simulator/types.ts), but that seems to be default in this case

src/components/Simulator/utils.ts
```diff
@@ -15,3 +17,8 @@ export const shareOnTwitter = (): void => {
 export const clearUrlParams = (location: Location): void => {
   navigate(location.pathname, { replace: true })
 }
+
+export const getValidPathId = (pathIdString: PathId | null): PathId | null => {
```
pettinarip, line 21: This could be a good case to use [type narrowing](https://www.typescriptlang.org/docs/handbook/2/narrowing.html) where you receive any string and you return a more specific type, in this case `PathId`

```ts
export const getValidPathId = (pathIdString: string | null): PathId | null => {
  if (!isValidPathId(pathIdString)) return null
  return pathIdString
}

const isValidPathId = (pathIdString: string | null): pathIdString is PathId => {
  return PATH_IDS.includes(pathIdString as PathId)
}
```

src/components/Simulator/StartingPoint.tsx
```diff
@@ -0,0 +1,202 @@
...
+}
+export const StartingPoint: React.FC<IProps> = ({
+  children,
+  data,
+  location,
+}) => {
+  // Track pathID
+  const params = new URLSearchParams(location.search)
+  const pathIdString = params.get(PATH_ID_QUERY_PARAM)
+  const pathId: PathId | null = getValidPathId(pathIdString as PathId | null)
```
pettinarip, line 27: We could simply do
```ts
const pathId = getValidPathId(pathIdString)
```
with the `getValidPathId` refactor

src/components/Simulator/StartingPoint.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  // If pathId present, modal is open, else closed
+  const isOpen = !!pathId
```
pettinarip, line 30: 🚀

src/pages/simulator.tsx
wackerow, line 1: Note: This page is to be removed before PR merged (was for dev only)... See `pages-conditional/wallets/tsx` for prod placement

src/components/Simulator/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+                  eventAction: `${pathId}_click`,
+                  eventName: `next-lesson-${id}`,
+                })
+                openPath(id)
+              }}
+              logFinalCta={logFinalCta}
+            />
+            <Phone>
+              <Screen nav={nav!} ctaLabel={ctaLabel!} />
+            </Phone>
```
pettinarip, line 199: A question for future iterations about re-usability: how could we refactor this part since another simulator might not want to use a `Phone` layout?

It could be part of the given `Screen`.

src/components/Simulator/Explanation.tsx
```diff
@@ -0,0 +1,148 @@
...
+  nextPathId,
+  finalCtaLink,
+  openPath,
+  logFinalCta,
+}) => {
+  const { regressStepper, step, totalSteps } = nav
+  const { header, description } = explanation
+
+  const Description: React.FC<BoxProps> = (props) => (
+    <Box {...props}>{description}</Box>
```
pettinarip, line 40: I don't see too much utility in creating this abstraction instead of just using `<Box>{description}</Box>` in the places you need.

Also, a component should receive the variables through the props and should be declared outside of the `Explanation` component.

src/components/Simulator/Explanation.tsx
```diff
@@ -0,0 +1,148 @@
...
+          </Text>
+        </Grid>
+        {/* Header and description */}
+        <Box>
+          <Text
+            fontSize={{ base: "xl", sm: "2xl", md: "3xl", lg: "4xl" }}
+            lineHeight={{ base: 8, md: 10 }}
+            fontWeight="bold"
+            mb={{ base: 4, md: 8 }}
+          >
```
pettinarip, line 91: Shouldn't this be a `Heading` component + using an `h` tag? currently this is a paragraph.

src/components/Simulator/screens/ConnectWeb3/Web3App.tsx
```diff
@@ -0,0 +1,27 @@
...
+  useColorModeValue,
+} from "@chakra-ui/react"
+import React from "react"
+import { GrMenu } from "react-icons/gr"
+import { EthGlyphIcon } from "../../icons"
+
+interface IProps extends BoxProps {}
+export const Web3App: React.FC<IProps> = ({ children, ...boxProps }) => (
+  <Box h="full" w="full" bg="background.highlight" {...boxProps}>
+    <Box p={1} bg={useColorModeValue("#e8e8e8", "#171717")}>
```
pettinarip, line 16: This is a hook. Should be defined at the top of the function.

```tsx
export const Web3App: React.FC<IProps> = ({ children, ...boxProps }) => {
  const bgColor = useColorModeValue("#e8e8e8", "#171717")

  return(
  <Box h="full" w="full" bg="background.highlight" {...boxProps}>
    <Box p={1} bg={bgColor}>
```

pettinarip (commented):
Looks great @wackerow 🎉  gj

The main concern I've left is about not using heading tags. Wondering if this was on purpose or just something missing.

src/pages/simulator.tsx
```diff
@@ -0,0 +1,29 @@
...
+      >
+        Interactive explainer
+      </Text>
+      <Text
+        fontSize={{ base: "3xl", md: "4xl", lg: "5xl" }}
+        lineHeight="115%"
+        fontWeight="bold"
+        m={0}
+      >
+        How to use a wallet
```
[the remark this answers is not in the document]
pettinarip: Note: we have merged the new Headings from the DS. So, you could use `<Heading as="h1" size="2xl">` for example. You'd need to merge `dev` first.

src/components/Simulator/index.tsx
wackerow: Yeah, or perhaps we create a prop for this `Simulator` component called `Display`, and pass `Phone` through from the parent, which could be swapped out for say `BrowserWindow`...
Adding it to the Screen would be somewhat redundant, but would offer the flexibility of switching displays potentially within one sim.

src/pages/simulator.tsx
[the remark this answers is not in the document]
wackerow: I think originally I wasn't viewing this as an article/document, so I avoided the `h` tags, but think it's probably better to leave them in. I'll use an `h2` for "How to use a wallet" (since it'll already be nested inside a page with an `h1`... in the modal I'll then use `h3`s for the explanation header text.

src/components/Simulator/screens/ConnectWeb3/Web3App.tsx
wackerow: oh good catch, thanks

src/pages/simulator.tsx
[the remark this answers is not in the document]
pettinarip: Cool. In that case you wouldn't need to specify the `as` prop since the default is already `h2`.

corwintines (approved):
❤️

PR #11005, opened 2023-08-25
Dapp page changes

minimalsm:
## Description

- Changes editors' choice to beginner-friendly dapps
- Changes hero primary CTA to point to beginner-friendly dapps
- Temporarily removes add dapp section

## Related Issue

[Discussion here](https://www.notion.so/efdn/Dapps-page-sync-a039dcfc3a30443abe825731781188d6?pvs=4#743ee3fc2f0b4535a58c7c9224668c4d)


wackerow (approved):
Maybe we should remove Dark Forest, there hasn't been a round in 18 months... can do this separately though.

corwintines:
Overall I think the changes here are good, but do think the apps that are on this beginner-friendly list should change as well in this PR. 

I would suggest:
1 - Uniswap
2 - Opensea
3 - Gods unchained
4 - Lens/Lenster

I think this would give a bit more variety (one finance, NFT, game, and social) of applications that are generally pretty easy to use.

konopkja:
there are other action points from the discussion i dont see in this PR, @minimalsm who when will followup? should it be in this or separate PR? 
Screenshot 2023-09-26 at 19 59 35


For example based on tab switching risk level and tailored message in this box instead of generic DYOR

PR #11021, opened 2023-08-28
Markdown links component

pettinarip:
Recently, we added the glossary tooltip component in #10928.

To use that component inside markdown files and since all the glossary terms in md files are links, I've created a new component called `MdLink` that decides which component to use in a given link, either the InlineLink or the GlossaryTooltip component.

The main reason behind it is to keep md files untouched, using the same syntax we had, since it works better with Crowdin.

## Description

Adds a new Link component for md templates.

## Related Issue

Implementation of #10928

### DIFF
--- a/src/components/MdLink.tsx
+++ b/src/components/MdLink.tsx
@@ -1,0 +1,36 @@
+import React, { ReactNode } from "react"
+
+import GlossaryTooltip from "./Glossary/GlossaryTooltip"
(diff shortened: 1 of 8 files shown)


minimalsm (approved):
Nice, LGTM 😀

I put up [a PR to revert the change to the DeFi page](https://github.com/ethereum/ethereum-org-website/pull/11027) introduced in #10928

PR #11036, opened 2023-08-30
refactor: combine directories of the button components

TylerAPfledderer:
## Description

Moves the `Button`, `ButtonLink`, `IconButton`, and `ButtonTwoLines` to a central subdirectory `Buttons`.

- Consolidate similar components under a single location
- With the current changes supplied, this breaks convention with component exports, as there is a single `index` file all the components export from as named exports.
  - Alternatively, have no `index` file entry point and import directly from the individual component files (negligible on readability difference?) 

So
```tsx
import { Button, type IButtonProps } from './components/Buttons'

// OR

import Button, { type IProps as IButtonProps } from './components/Buttons/Button'
```

## Related Issue
N/A


docs/ds-implementation.md
```diff
@@ -17,10 +17,10 @@ Follow the new component directory structure:
 ```markdown
 src/
 └── components/
-    └── ComponentA/
-        ├── index.tsx
-        ├── ComponentA.stories.tsx
-        └── // Any other files as applicable (utils, child components, useHook, etc.)
```
pettinarip, line 23: Probably triggered by the autoformat hook but this is incorrect. Revert this formatting.

pettinarip (changes requested):
Nice! looks good. Adding a small fix to convert a few remaining files.

pettinarip (approved):

PR #11047, opened 2023-09-02
feat(components): add `ButtonTwoLines` component

TylerAPfledderer:
## Description

Creates a button component with the following features:

- dedicated two distinct lines of text content with a main label and a helper label
- required icon
- ability to show the text in reverse stacking order
- ability to have the content left- or right-aligned (i.e. for pagination)
- renders the custom `Button` or `ButtonLink` component.

Note: part of the new DS as an intermediary step to creating a pagination button or the interactive simulator button that uses the design.

## Related Issue
N/A

### DIFF
--- a/src/components/ButtonTwoLines/index.tsx
+++ b/src/components/ButtonTwoLines/index.tsx
@@ -1,0 +1,105 @@
+import * as React from "react"
+import { Icon, Stack, Text } from "@chakra-ui/react"
+import type { IconType } from "react-icons/lib"
+import Button, { type IProps as ButtonProps } from "../Button"
+import ButtonLink, { type IProps as ButtonLinkProps } from "../ButtonLink"
+
+type CommonProps = {
+  icon: IconType | typeof Icon
+  iconAlignment?: "left" | "right"
+  /**
+   * Reduced choices of the button variant.
+   *
+   * This component only accepts the `solid` or `outline` variant
+   */
+  variant?: "solid" | "outline"
+  /**
+   * Reduced choices of the button size
+   *
+   * This component only accepts the `md` or `sm` sizes
+   */
+  size?: "md" | "sm"
+  mainText: string
+  helperText: string
+  /**
+   * Should the main text be below the helper text instead of ab?
+   */
+  reverseTextOrder?: boolean
+}
+
+type OmittedTypes = "variant" | "size"
+
+type ButtonTypeProps = CommonProps &
+  Omit<ButtonProps, OmittedTypes> & {
+    href?: never
+  }
+
+type ButtonLinkTypeProps = CommonProps &
+  Omit<ButtonLinkProps, OmittedTypes> & {
+    toId?: never
+  }
+
+type ButtonTwoLinesProps = ButtonTypeProps | ButtonLinkTypeProps
+
+const hasHref = (props: ButtonTwoLinesProps): props is ButtonLinkTypeProps => {
+  return "href" in props
+}
+
+const ButtonTwoLines = (props: ButtonTwoLinesProps) => {
+  const {
+    icon: Icon,
+    iconAlignment = "left",
+    mainText,
+    helperText,
+    reverseTextOrder = false,
+    size = "md",
+    ...rest
+  } = props
+
+  const isIconLeft = iconAlignment === "left"
+
+  const vertPadding: ButtonTwoLinesProps["py"] = size === "md" ? "4" : "2"
+
+  const buttonStyles: ButtonProps = {
+    [isIconLeft ? "leftIcon" : "rightIcon"]: <Icon />,
+    textAlign: isIconLeft ? "start" : "end",
+    justifyContent: isIconLeft ? "flex-start" : "flex-end",
+    w: "278px",
+  }
+
+  const Component = hasHref(props) ? ButtonLink : Button
+
+  return (
+    <Component
+      {...buttonStyles}
+      size={size}
+      py={vertPadding}
+      {...rest}
+      sx={{
+        ".chakra-button__icon svg": {
+          // Force icon to be the same size for both button sizes
+          fontSize: "2xl",
+        },
+      }}
+    >
+      <Stack
+        spacing={0}
+        flexDir={reverseTextOrder ? "column-reverse" : "column"}
+      >
+        <Text
+          as="span"
+          fontSize="md"
+          fontWeight={size === "md" ? "bold" : "normal"}
+        >
+          {mainText}
+        </Text>
+        {/* TODO: use `size='xs'` instead when the Text theme is added  */}
+        <Text as="span" fontSize="xs">
+          {helperText}
+        </Text>
+      </Stack>
+    </Component>
+  )
+}
+
+export default ButtonTwoLines
--- a/src/components/ButtonTwoLines/index.tsx
+++ b/src/components/ButtonTwoLines/index.tsx
@@ -64,7 +64,6 @@ const ButtonTwoLines = (props: ButtonTwoLinesProps) => {
     [isIconLeft ? "leftIcon" : "rightIcon"]: <Icon />,
     textAlign: isIconLeft ? "start" : "end",
     justifyContent: isIconLeft ? "flex-start" : "flex-end",
-    w: "278px",
   }
 
   const Component = hasHref(props) ? ButtonLink : Button
--- a/src/components/ButtonTwoLines/ButtonTwoLines.stories.tsx
+++ b/src/components/ButtonTwoLines/ButtonTwoLines.stories.tsx
@@ -20,6 +20,7 @@ export const ButtonTwoLines: Story = {
     icon: BiCircle,
     mainText: "Main Text",
     helperText: "Helper Text",
+    w: "300px",
   },
   render: (args) => (
     <Stack spacing="8">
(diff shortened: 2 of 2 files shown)


TylerAPfledderer:
@pettinarip should I go ahead and add the `PaginationButton` and the `SimulatorButton` components here, or leave them for separate PRs?

src/components/ButtonTwoLines/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  const isIconLeft = iconAlignment === "left"
+
+  const vertPadding: ButtonTwoLinesProps["py"] = size === "md" ? "4" : "2"
+
+  const buttonStyles: ButtonProps = {
+    [isIconLeft ? "leftIcon" : "rightIcon"]: <Icon />,
+    textAlign: isIconLeft ? "start" : "end",
+    justifyContent: isIconLeft ? "flex-start" : "flex-end",
+    w: "278px",
```
pettinarip, line 67: Probably we've talked about this value but could you remind me why we want this fixed width?

pettinarip:
> @pettinarip should I go ahead and add the `PaginationButton` and the `SimulatorButton` components here, or leave them for separate PRs?

Good question.

I thought that we wouldn't need to create those components as they wouldn't have any reusable use case atm. We will simply use them in just one place.
- The "PaginationButton" would only be used in the dev docs template
- The "SimulatorButton" only on the simulator

And they don't hold any additional logic that would justify creating a new component.

LMK if you think differently.

src/components/ButtonTwoLines/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+   *
+   * This component only accepts the `md` or `sm` sizes
+   */
+  size?: "md" | "sm"
+  mainText: string
+  helperText: string
+  /**
+   * Should the main text be below the helper text instead of ab?
+   */
+  reverseTextOrder?: boolean
```
pettinarip, line 27: Nice

TylerAPfledderer:
> > @pettinarip should I go ahead and add the `PaginationButton` and the `SimulatorButton` components here, or leave them for separate PRs?
> 
> Good question.
> 
> I thought that we wouldn't need to create those components as they wouldn't have any reusable use case atm. We will simply use them in just one place.
> - The "PaginationButton" would only be used in the dev docs template
> - The "SimulatorButton" only on the simulator
> 
> And they don't hold any additional logic that would justify creating a new component.
> 
> LMK if you think differently.

Ah understood. That would be it then. I will mark as "Ready for Review" 😀

src/components/ButtonTwoLines/index.tsx
TylerAPfledderer: I believe we have yet to discuss this. 😁

I had noticed at one point in the figma file some playing around with the text length and having it wrap and keep the same button length.

With the expected usage, I would think having a fixed width to maintain uniformity for every instance of this component.

Alternatively, could make it `min-width` instead and allow text to stretch the button and not wrap.

pettinarip: Interesting. `min-width` would make more sense to me here.

@nloureiro is this something we want to keep in code? or was just to display it nicely in figma?

nloureiro:
> 


I'm not sure, I think this is the parent component that should do and not the button component

example: the pagination will be inside a <Pagination> component that can force the buttons to be X width and other logic like disable, height, etc.

I just did a simulation on Figma and in the pagination, they should have the same height to keep the balance and grid.

Screen Shot 2023-09-05 09 31 38 AM

pettinarip:
> > 
> 
> I'm not sure, I think this is the parent component that should do and not the button component
> 
> example: the pagination will be inside a component that can force the buttons to be X width and other logic like disable, height, etc.

@nloureiro correct.

I will be 99% of the time aligned with letting the consumer decide and control as much as they can with these components. That generates simpler and more flexible components. That is why I think we shouldn't have anything set by default in terms of sizes.

So, I think you agree with that. We can remove the fixed width for now and get the PR merged.

src/components/ButtonTwoLines/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const vertPadding: ButtonTwoLinesProps["py"] = size === "md" ? "4" : "2"
+
+  const buttonStyles: ButtonProps = {
+    [isIconLeft ? "leftIcon" : "rightIcon"]: <Icon />,
+    textAlign: isIconLeft ? "start" : "end",
+    justifyContent: isIconLeft ? "flex-start" : "flex-end",
+    w: "278px",
+  }
+
+  const Component = hasHref(props) ? ButtonLink : Button
```
pettinarip, line 70: @TylerAPfledderer was wondering about this decision since this kind of divert from the convention we have been taking with other similar components.

What I would expect from a dev's standpoint is to follow Chakra conventions instead of relying on the component's logic (you wouldn't know about this logic unless you look into the code)

```tsx
// what I mean we should do
<ButtonTwoLines as={Link} href="/" ... >

// instead of assuming the component is going to know how to work with an href prop
<ButtonTwoLines href="/" ... >
```

TylerAPfledderer: The first thing that comes to mind is the re-usability. There are I think a few parts to this:
- Are we expecting more than just the custom `Button` and `ButtonLink` component to be used here?
- By creating a different render through `as` you would have added an additional abstraction layer. If we are to choosing the custom `Button` component as the main render in `ButtonTwoLines` and then choose to do `as={Link}`, this `Link` is then applied to `Button` and then you lose the `activeStyle={{}}` prop needed in `ButtonLink`.
- If the purpose of this button is only for interactivity in the site, then the user might think of using `as={ BaseLink }`. If that is the case, then why not just `as={ ButtonLink }`? And then becomes the abstraction issue that could be present when passing `as={ ButtonLink }` to the render of `<Button />`.  (That's at least three layers of abstraction here)
- How would you define the prop types of `ButtonTwoLines` if we are manually introducing a `Link` component or similar without causing a type-error like "`toId` is not defined" because `Link` does not have this prop, but `Button` (as the main render) does?

Does this make sense?

Since the custom `BaseLink` component (which is used in `ButtonLink`) needs the `href` prop (per the [outstanding Todo comment in it's file](https://github.com/ethereum/ethereum-org-website/blob/cfd311d75f6257ddf473c4a3d22a4c56e15508f2/src/components/Link.tsx#L64-L66)), then I would think to determine the render based on this prop requirement. 

Then with `hasHref()` determining which component to render between `Button` and `ButtonLink` you have a type guard to prevent any prop usage that would be tied to `Button` that is not present in the props of `ButtonLink`. This and the overall typing of `ButtonTwoLines` props is based on a [recommended TypeScript approach to type narrowing](https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase#type-narrowing-based-on-props). Again, we are assuming that `ButtonTwoLines` should only be rendering either `Button` or `ButtonLink` and nothing else. If we want it to render a link, we would have to pass the `href` prop anyway.

Also, for future-proofing, the `as` prop is expected to be deprecated in Chakra v3 in favor of `asChild` which will require a child element rendering. See Sage's [latest post on the plans for Chakra](https://www.adebayosegun.com/blog/chakra-panda-ark-whats-the-plan#moving-from-as-prop-to-aschild). Might be excessive here for readability and re-usability.

I see `ButtonTwoLines` as an extension of either `Button` or `ButtonLink`, constrained only to those components and it morphing based on the content it is given.

pettinarip: Thanks for the detailed reply.

My concern is more related to what a dev would expect from this component and making it easier to use (followin conventions).

As of now, we have:
- Link => renders `a`
- Button => renders `button`
- ButtonLink => renders `a`

They are pure in what they render.

- `ButtonTwoLines` => renders `a` or `button`

IMO this breaks the convention and creates a confusion in dev's mind. Even the name ButtonTwoLines clearly indicates to the dev that this is going to render a button, not a link.

That is why I thought on giving this decision to the dev. Do you want a link to be rendered? then follow the same convention we did in `ButtonLink` => `<Button as={BaseLink} ...>`

So, following our previous conventions, I would expect having the following:
- `ButtonTwoLines` => renders `button`
- `ButtonLinkTwoLines` => renders `a` (I don't think this necessary, just to explain my idea)

======

Having that, the types are going to be simpler, basically the same we have in our custom `Button`. We would extend them.

TylerAPfledderer: I'm sorry. I understand wanting improved DX, but I'm confused on what you are proposing. 😅 

So if a user wants to have `ButtonTwoLines` as a link, you are thinking of `<ButtonTwoLines as={ BaseLink }>` or `<ButtonTwoLines as={ ButtonLink ] />`?

pettinarip: I'm probably overthinking this too much for just one component haha. I understand your points and think this will work well as it is 👍🏼 

----

Just for clarification: yea, the basic idea was to simplify this component to just render a button as its name tells (that would simplify the logic, types, and api). Then, if the user wants a link to be rendered, they could use the `as` prop. This would also align with how we use in the current Button and ButtonLink tandem, making it easier to grasp.

pettinarip (approved):
🚀

PR #11050, opened 2023-09-02
oneliner, add revm to list of standalone EVMs

rakita:
Small change, add revm to the list: https://github.com/bluealloy/revm


corwintines (approved):
Thanks @rakita

PR #11075, opened 2023-09-05
Added hover animations for nav icons

rohan9024:
For context, refer this PR--> https://github.com/ethereum/ethereum-org-website/pull/10811

These were the changes requested- 
1. Icon sizes should go back to original
2. The gap between icons increased, this should be the same as before
3. Add the .env.example file back
4. The highlight of the language icon and text should be in sync

Changes done by me-
1. Icon sizes changed to original
2. Fixed gap between icons
3. Added back the .env.example
4. When hovered over text the language icon, transition applies on icon.

Here is the working,

https://github.com/ethereum/ethereum-org-website/assets/64919642/58b8addd-40bc-418b-8689-6a8ca8ca8c50


src/components/Nav/index.tsx
```diff
@@ -92,16 +104,44 @@ const Nav: FC<IProps> = ({ path }) => {
...
                   variant="ghost"
                   isSecondary
                   px={1.5}
+                  _hover={{
+                    color: "blue", // Change color to blue on hover
+                  }}
+                  style={{
+                    display: "inline-flex",
+                    alignItems: "center",
+                  }}
```
TylerAPfledderer, line 128: This is excessive, as the Chakra `Button` being rendered already supplies these styles.
```suggestion
```

src/components/Nav/index.tsx
```diff
@@ -92,16 +104,44 @@ const Nav: FC<IProps> = ({ path }) => {
...
+                    transform: "rotate(30deg)", // Rotate the icon to 30 degrees on hover
+                    color: "blue", // Change color to blue on hover
+                  }}
                   onClick={toggleColorMode}
                 ></IconButton>
+
                 <ButtonLink
+                  onMouseOver={() => setLanguagesHover(true)}
+                  onMouseOut={() => setLanguagesHover(false)}
+                  as={Link}
```
TylerAPfledderer, line 117: It already is a link.
```suggestion
```

src/components/Nav/index.tsx
```diff
@@ -92,16 +104,44 @@ const Nav: FC<IProps> = ({ path }) => {
...
                   to={`/languages/${fromPageParameter}`}
-                  leftIcon={<Icon as={MdLanguage} />}
                   variant="ghost"
                   isSecondary
                   px={1.5}
+                  _hover={{
+                    color: "blue", // Change color to blue on hover
+                  }}
```
TylerAPfledderer, line 124: @pettinarip this is actually a bug, where the `isSecondary` prop is causing an override to any pseudo class. So adding `_hover` is necessary here for now.

@rohan9024 please keep to semantic tokens and refer to the [Ethereum DS Figma file](https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?type=design&node-id=2283-45220&mode=design&t=5c6BwHguyB4nMFv1-0) when determining which semantic token to use. Here, the token should be `primary.hover`

```suggestion
                  _hover={{
                    color: "primary.hover", // Change color to blue on hover
                  }}
```

src/components/Nav/index.tsx
```diff
@@ -92,16 +104,44 @@ const Nav: FC<IProps> = ({ path }) => {
...
+                      style={{
+                        transition: "transform 0.3s ease-in-out, color 0.2s", // Apply the transition to the icon
+                        transform: languagesHover
+                          ? "rotate(30deg)"
+                          : "rotate(0deg)", // Rotate the icon to 30 degrees on hover
+                        color: languagesHover ? "blue" : "inherit", // Change color to blue on hover
+                      }}
+                    />
+                  </Box>
+                  <span>Languages {i18n.language.toUpperCase()}</span>
```
TylerAPfledderer, line 144: This is excessive. The Chakra `Button` which `ButtonLink` uses has the props `leftIcon` and `rightIcon` which handle alignment of an icon against text. The spacing of token `2` is already the default.
```tsx
<ButtonLink
  // -- Other props -- //
  leftIcon={
    <Icon {...iconProps} />
  }
>
  Languages {i18n.language.toUpperCase()}
</ButtonLink>
```

src/components/Nav/index.tsx
```diff
@@ -92,16 +104,44 @@ const Nav: FC<IProps> = ({ path }) => {
...
+                    {" "}
+                    {/* Add spacing between the text and the icon */}
+                    <Icon
+                      as={MdLanguage}
+                      style={{
+                        transition: "transform 0.3s ease-in-out, color 0.2s", // Apply the transition to the icon
+                        transform: languagesHover
+                          ? "rotate(30deg)"
+                          : "rotate(0deg)", // Rotate the icon to 30 degrees on hover
+                        color: languagesHover ? "blue" : "inherit", // Change color to blue on hover
```
TylerAPfledderer, line 140: This needs to be removed, and will be inherited by the hover state supplied directly to `ButtonLink`
```suggestion
```

src/components/Nav/index.tsx
```diff
@@ -10,6 +10,7 @@ import Search from "../Search"
 import IconButton from "../IconButton"
 import { EthHomeIcon } from "../icons"
 import { useNav } from "./useNav"
+import { FaLastfmSquare } from "react-icons/fa"
```
TylerAPfledderer, line 13:
```suggestion
```

src/components/Nav/index.tsx
```diff
@@ -92,16 +104,44 @@ const Nav: FC<IProps> = ({ path }) => {
...
+                  <Box mr={2} mt={2}>
+                    {" "}
+                    {/* Add spacing between the text and the icon */}
+                    <Icon
+                      as={MdLanguage}
+                      style={{
+                        transition: "transform 0.3s ease-in-out, color 0.2s", // Apply the transition to the icon
+                        transform: languagesHover
+                          ? "rotate(30deg)"
+                          : "rotate(0deg)", // Rotate the icon to 30 degrees on hover
```
TylerAPfledderer, line 139: There are a couple of pretty cool things to do here. Both `transition` and `transform` are style props, so they should not be under the `style` prop.

Secondly, we can put `transform` under the hover pseudo and not waste energy on `useState`. This is a two step process. 
1. Add `data-group` to `ButtonLink`
2. In the `Icon` use the `_groupHover` prop which will target `data-group` with a descendant selector. so when hovering `ButtonLink` the styles under this prop will trigger. Then no need for this ternary!

So in short...
```tsx
<ButtonLink
  // -- Other props -- //
  data-group
  leftIcon={
    <Icon
      as={MdLanguage}
      transition="transform 0.3s ease-in-out, color 0.2s"
      _groupHover={{
        transform: "rotate(30deg)"
      }}
    />
  }
/>
```

src/components/Nav/index.tsx
```diff
@@ -83,7 +85,17 @@ const Nav: FC<IProps> = ({ path }) => {
...
+                    isDarkTheme ? (
+                      <MdWbSunny
+                        style={{ transition: "transform 0.5s, color 0.2s" }}
+                      />
+                    ) : (
+                      <MdBrightness2
+                        style={{ transition: "transform 0.5s, color 0.2s" }}
+                      />
+                    )
+                  }
```
TylerAPfledderer, line 98: Can simply supply `transition` as a prop to `IconButton`
```suggestion                  
                  transition="transform 0.5s, color 0.2s"
                  icon={isDarkTheme ? <MdWbSunny /> : <MdBrightness2 />}
```

TylerAPfledderer (changes requested):
Hey @rohan9024! I have some suggestions below for now.

src/components/Nav/index.tsx
```diff
@@ -92,16 +94,38 @@ const Nav: FC<IProps> = ({ path }) => {
                   variant="ghost"
                   isSecondary
                   px={1.5}
+                  _hover={{
+                    transform: "rotate(30deg)", // Rotate the icon to 30 degrees on hover
+                    color: "blue", // Change color to blue on hover
```
TylerAPfledderer, line 99: Same as with the other button icon
```suggestion
                    color: "primary.hover", // Change color to blue on hover
```

TylerAPfledderer (commented):
I missed one! :D

rohan9024:
@TylerAPfledderer Thanks for providing your valuable suggestions😊. I have made the changes you've requested. Incase if I have missed out something please guide me✌

src/components/Nav/index.tsx
TylerAPfledderer: @rohan9024 this suggested change is still outstanding :)

rohan9024: @TylerAPfledderer done✌

src/components/Nav/index.tsx
```diff
@@ -92,16 +94,42 @@ const Nav: FC<IProps> = ({ path }) => {
...
+                    color: "primary.hover", // Change color on hover
+                  }}
                 >
-                  {t("languages")} {i18n.language.toUpperCase()}
+                  <Box mr={2} mt={2}>
+                    {" "}
+                    {/* Add spacing between the text and the icon */}
+                    <ButtonLink
+                      data-group
+                      leftIcon={
```
pettinarip, line 120: Not sure if I follow what we are trying to do here but currently, you have a nested `ButtonLink` inside another `ButtonLink`, resulting in a nested `a` tag.

I guess that what we would like to do here is to render just the icon, right? and add the `data-group` tag on the parent `ButtonLink`.

====
Note aside, the build is failing because this `ButtonLink` is missing the `href` prop.

pettinarip (changes requested):
Hey @rohan9024 I've noticed that the PR is failing to build.

The error is that the `ButtonLink` doesn't have an `href` prop which is a required prop.

So, I ran this PR locally and fixed that to see the result and this is how I see the button now:
image

You can see that the icon is displayed as a button and that is a bit vertically misaligned.

As I said in the comment below, I think that by removing the nested `ButtonLink`, that should fix these visual issues.

src/components/Nav/index.tsx
TylerAPfledderer: Yea, the example I gave was not suggesting nested, but a simplified version of what the component should look like with that `data-*` prop and the `leftIcon` prop

rohan9024:
> Hey @rohan9024 I've noticed that the PR is failing to build.
> 
> The error is that the `ButtonLink` doesn't have an `href` prop which is a required prop.
> 
> So, I ran this PR locally and fixed that to see the result and this is how I see the button now: image
> 
> You can see that the icon is displayed as a button and that is a bit vertically misaligned.
> 
> As I said in the comment below, I think that by removing the nested `ButtonLink`, that should fix these visual issues.

@pettinarip Alright got it! Updating the code.....

rohan9024:
@pettinarip Done 👍

src/components/Nav/index.tsx
pettinarip: Hey @rohan9024 I don't see that this was implemented. I think this change is important to simplify the overall code and to use the already provided Chakra api.

Marking it as unresolved for now.

src/components/Nav/index.tsx
pettinarip: @rohan9024 marking this as unresolved.

Currently we have the `_groupHover` but not the `data-group` set in the parent ButtonLink.

src/components/Nav/index.tsx
```diff
@@ -31,6 +31,7 @@ const Nav: FC<IProps> = ({ path }) => {
   const searchModalDisclosure = useDisclosure()
 
   const navWrapperRef = useRef(null)
+  const [languagesHover, setLanguagesHover] = useState(false)
```
pettinarip, line 34: Not needed. Lets remove this state.

src/components/Nav/index.tsx
```diff
@@ -92,16 +94,37 @@ const Nav: FC<IProps> = ({ path }) => {
                   variant="ghost"
                   isSecondary
                   px={1.5}
+                  _hover={{
+                    transform: "rotate(30deg)", // Rotate the icon to 30 degrees on hover
```
pettinarip, line 98: Re-reading the conversations from the issue, I think we ended up agreeing on just [rotating this 10 degrees](https://github.com/ethereum/ethereum-org-website/pull/10811#issuecomment-1665601093).
```suggestion
                    transform: "rotate(10deg)",
```

src/components/Nav/index.tsx
```diff
@@ -92,16 +94,37 @@ const Nav: FC<IProps> = ({ path }) => {
...
+                    color: "primary.hover", // Change color to blue on hover
```
pettinarip, line 99: Unnecessary comment. The code is self-explanatory.

```suggestion
                    color: "primary.hover",
```

src/components/Nav/index.tsx
```diff
@@ -92,16 +94,37 @@ const Nav: FC<IProps> = ({ path }) => {
...
                 <ButtonLink
+                  onMouseOver={() => setLanguagesHover(true)}
+                  onMouseOut={() => setLanguagesHover(false)}
                   to={`/languages/${fromPageParameter}`}
-                  leftIcon={<Icon as={MdLanguage} />}
                   variant="ghost"
                   isSecondary
                   px={1.5}
+                  _hover={{
+                    color: "primary.hover", // Change color on hover
```
pettinarip, line 112: Unnecessary comment. The code is self-explanatory.
```suggestion
                    color: "primary.hover",
```

src/components/Nav/index.tsx
```diff
@@ -92,16 +94,37 @@ const Nav: FC<IProps> = ({ path }) => {
...
+                  }}
                 >
-                  {t("languages")} {i18n.language.toUpperCase()}
+                  <Box mr={2} mt={2}>
+                    {" "}
+                    {/* Add spacing between the text and the icon */}
```
pettinarip, line 117: No need to this extra space

src/components/Nav/index.tsx
```diff
@@ -92,16 +94,37 @@ const Nav: FC<IProps> = ({ path }) => {
...
+                    <Icon
+                      as={MdLanguage}
+                      transition="transform 0.3s ease-in-out, color 0.2s"
+                      // _groupHover renders "[data-group]:hover &"
```
pettinarip, line 121: ✂️

src/components/Nav/index.tsx
```diff
@@ -92,16 +94,37 @@ const Nav: FC<IProps> = ({ path }) => {
...
+                      _groupHover={{
+                        transform: "rotate(30deg)",
```
pettinarip, line 123:
```suggestion
                        transform: "rotate(10deg)",
```

yarn.lock
pettinarip, line 1: We shouldn't have any changes added in the `yarn.lock` file. Lets revert this changes

pettinarip (changes requested):
@rohan9024 I've added a few comments to make this cleaner. There are also 2 change requests added by Tyler that should be tackled as well.

Thanks for the time you have put in this. Let me know if you need help doing any of these changes or if I wasn't clear enough.

src/components/Nav/index.tsx
rohan9024: I didn't get this😅

pettinarip: Sorry. I meant to cut it, to remove it xD

Its a comment that doesn't add any new information to the developer.

pettinarip (approved):

corwintines (approved):
Thanks @rohan9024 

I committed a few changes to get this over the finish line:
- Took text out of a span, and added the translated template string back here (just reverted your change really around this)
- Added the leftIcon back in for the ButtonLink and removed the Icon you had added (reverted to how to use the ButtonLink component)
- Added animation to the icon for languages, which you can check out in the hover styles if interested.

PR #11101, opened 2023-09-07
Fix broken link: Learn more about MACI

Mirror-Tang:
The original link is invalid and has been updated to the correct link.

https://ethereum.org/zh/zero-knowledge-proofs/#how-maci-works-with-zk-proofs


nhsz (approved):
LGTM!

PR #11106, opened 2023-09-07
Update index.md

ilanolkies:
## Description

Fix errata

In the yellow paper:
>  If the concatenated serialisations of each contained item is less than 56 bytes in length, then the output is equal
to that concatenation prefixed by the byte equal to the length of this byte array plus 192


corwintines (approved):

corwintines:
Thanks @ilanolkies!

@all-contributors can you add @ilanolkies for content

PR #11114, opened 2023-09-08
minor change to strong statelessness info

jmcook1186:
## Description

Adds small detail about strong statelessness making contract interactions more difficult


wackerow (approved):
👍

PR #11133, opened 2023-09-12
feat: create Chakra `GatsbyImage` component

### LINKED ISSUE #11109: Apply `GatsbyImage` with Chakra via `chakra()` Factory Function
## Current Issue

Currently, any instance of the `GatsbyImage` is consumed by Chakra's `Image` or `Img` component.
```tsx
<Image
  as={ GatsbyImage }
/>
```

This loses `GatsbyImage`'s usage of some common props that are handled differently. Chakra primarily filters style props from reaching the DOM. But common style props used in `GatsbyImage` are applied differently than in Chakra, and Chakra will override the usage. For example, the `objectFit` prop in `GatsbyImage` has to reach the `img` element deep in its DOM structure in order to be handled correctly. Chakra will prevent this, because it thinks the prop is being used at a root level.

## Proposed Change(s)

Instead, use the `chakra()` factory with the `shouldForwardProp` option to ensure the props used by `GatsbyImage` retain their use in the logic and not allow Chakra to override the usage.

```tsx
import { chakra, type ChakraComponent } from '@chakra-ui/react'
import { GatsbyImage as Image, type GatsbyImageProps } from 'gatsby-plugin-image'

// Proper typing of the props when using the component
const GatsbyImage: ChakraComponent<'img', GatsbyImageProps> = chakra( Image, {
  shouldForwardProp: (prop) => {
    // Have to add all used GatsbyImage props, else console warnings will throw
    return ["image", "loading", "objectFit", "alt"].includes(prop)
  }
})

<GatsbyImage
  alt=""
  image={ imgSrc }
  loading="eager"
  objectFit="contain"
  boxSize="full"
/>
```
Would create a custom `GatsbyImage` component file to house this for re-usability.

## Additional Information
Related to the move to NextJS: The `@chakra-ui/nextjs` integration package includes an [Image component](https://github.com/chakra-ui/chakra-ui/blob/63431c1ef3b7e8ac9b97c94ca025dd54b9aa3338/packages/integrations/next-js/src/image.tsx) that does this for the `NextImage` component to optimize it with Chakra.

TylerAPfledderer:
## Description

Creates a custom `GatsbyImage` component using the `chakra()` factory func so the Gatsby component can be properly optimized with Chakra.

Updates all pages and components that use this.

Also adds a section to the `best-practices.md` doc about this component.

## Related Issue
Closes #11109


TylerAPfledderer:
@pettinarip needing assistance in looking over the changes, since this update is wide-sweeping.

I also just noticed that Netlify is being used, but I have trouble checking when it keeps failing. 🤔

src/components/GatsbyImage.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export type GatsbyImageType = ChakraComponent<"img", GatsbyImageProps>
+
+/**
+ * Custom version of `GatsbyImage` optimized for Chakra
+ */
+const GatsbyImage: GatsbyImageType = chakra(Image, {
+  shouldForwardProp: (prop) => {
+    // Forward props that only GatsbyImage should be applying, else warnings may throw
+    return ["image", "loading", "objectFit", "alt", "objectPosition"].includes(
```
TylerAPfledderer, line 15: I'm afk at the time of this comment, but this list is still missing props that need to be forwarded.

pettinarip: Full list of props https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-plugin-image/

src/templates/upgrade.tsx
```diff
@@ -101,7 +101,7 @@ const InfoColumn = (props: ChildOnlyProp) => (
   />
 )
 
-// Apply styles for classes within markdown here
+// Apply styles for classes within markdown here[]
```
pettinarip, line 104:
```suggestion
// Apply styles for classes within markdown here
```

src/content/developers/tutorials/scam-token-tricks/index.md
pettinarip, line 1: This file shouldn't be affected by this PR

docs/best-practices.md
```diff
@@ -229,3 +229,11 @@ export const query = graphql`
...
 // These query results get passed as an object `props.data` to your component
 ```
+
+### Using custom `GatsbyImage`
+
+[GatsbyImage](https://www.gatsbyjs.com/plugins/gatsby-plugin-image/) is the component of choice to handle responsive images processed through graphql. However, we use a custom version of this component that is properly optimized with Chakra. This way we can use style props from Chakra but still be able to forward common or GatsbyImage-specific props to the Gatsby component for correct usage and rendering.
+
+```tsx
+import GatsbyImage from "./components/GatsbyImage"
+```
```
pettinarip, line 239: 💯

pettinarip (commented):
Awesome @TylerAPfledderer, thanks for the big refactor 💪🏼 

I'm adapting 1 more file that was added when you updated the branch with dev.

src/content/developers/tutorials/scam-token-tricks/index.md
TylerAPfledderer: Looks like it bled through the diff with small formatting changes.

pettinarip (approved):
🚀

PR #11148, opened 2023-09-13
Interactive simulator patches

wackerow:
## Description
- Updates some copy around ether/ETH
- Updates `NotificationPopover` (previously named DemoOnlyPopover) and further rolls out its usage with more helper text
- Switched final "get wallet" link to use relative path, and thus keep it as an "internal" link that will not open in a new tab

## Preview URL
- https://ethereumorgwebsitedev01-simpatches.gatsbyjs.io/en/wallets/

### DIFF
--- a/src/components/Simulator/NotificationPopover.tsx
+++ b/src/components/Simulator/NotificationPopover.tsx
@@ -5,6 +5,9 @@ import {
   PopoverBody,
   PopoverContentProps,
   PopoverProps,
+  PopoverHeader,
+  PopoverCloseButton,
+  Flex,
 } from "@chakra-ui/react"
 import React, { type ReactNode } from "react"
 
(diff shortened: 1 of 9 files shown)


konopkja:
@wackerow can we please in the first flow second step change headline from "They are free apps you can download" to "Wallets are free apps you can download..."
image

src/components/Simulator/NotificationPopover.tsx
```diff
@@ -0,0 +1,52 @@
...
+          bg="background.highlight"
+          px={4}
+          py={2}
+          maxW="15rem"
+          borderRadius="base"
+          boxShadow="tooltip"
+          fontSize="xs"
+          {...restProps}
+        >
+          <Flex gap={2} zIndex={100000000}>
```
pettinarip, line 41: Is this needed for a special case?

The Popover component already has the correct `zIndex` value https://chakra-ui.com/docs/styled-system/theme#z-index-values

```suggestion
          <Flex gap={2}>
```

pettinarip (approved):
LGTM!

src/components/Simulator/NotificationPopover.tsx
corwintines: I think we can get rid of this as well now that the Portal is added @wackerow

wackerow: Yes... I meant to switch that to something more specific, but the Portal should fix this. Removing

corwintines (approved):

PR #11149, opened 2023-09-13
Updated Pragma New York and added Pragma Istanbul

arredr2:
Updated Pragma New York and added Pragma Istanbul


corwintines (approved):
LGTM, thanks @arredr2!

PR #11161, opened 2023-09-15
Add ETHDubai 2024 conference and hackathon dates

patcito:
## Description

Add ETHDubai 2024 conference and hackathon dates (the page already has a few speakers and link to our call for paper).

### DIFF
--- a/src/data/community-events.json
+++ b/src/data/community-events.json
@@ -9,6 +9,15 @@
     "endDate": "2023-10-21"
   },
   {
+    "title": "ETHDubai",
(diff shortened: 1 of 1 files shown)


corwintines (approved):
Thanks @patcito 

I still have hesitation given there is no way to get tickets yet, just submit info to get an email about this. But, given we have had this event in the past will accept this time. 

Side note for us to consider in the future when accepting events is someone's ability to get what they need to attend the event, but that's a separate conversation for the future.

PR #11196, opened 2023-09-19
fix: add forward ref to `BaseLink`

TylerAPfledderer:
## Description

Currently the following React warning throws in development:

```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```

This points to the `BaseLink` component in the warning stack.

This PR wraps the component in Chakra's `forwardRef`.

In addition, a minor refactor with the `commonProps` object as the `ref` prop is used in all render conditions.

## Related Issue
N/A

### DIFF
--- a/src/components/Link.tsx
+++ b/src/components/Link.tsx
@@ -1,6 +1,5 @@
 import React from "react"
 import {
-  Box,
   forwardRef,
   Icon,
   Link as ChakraLink,
@@ -89,7 +88,9 @@ export const BaseLink = forwardRef<IProps, "a">(
       eventName: to,
     }
 
-    const commonProps = {
+    const commonProps: LinkProps & { ref: React.ForwardedRef<any> } = {
+      ref,
+      href: to,
       dir,
       ...restProps,
     }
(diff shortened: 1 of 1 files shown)


src/components/Link.tsx
```diff
@@ -47,125 +47,131 @@ export interface IProps extends IBaseProps, LinkProps {
...
-          if (!isHash) {
-            return
-          }
-
-          e.stopPropagation()
-          trackCustomEvent(
-            customEventOptions ? customEventOptions : hashLinkEvent
-          )
-        }}
+        as={IntlLink}
```
pettinarip, line 161: The Gatsby Link expects a `to` prop instead of the `href`.

pettinarip (changes requested):
Nice! thanks for fixing this warning.

Left a comment regarding the `commonProps` as the Gatsby Link ones are a bit different.

src/components/Link.tsx
TylerAPfledderer: Ah shoot. I'll revert the addition of `href` to `commonProps` for now.

TylerAPfledderer:
@pettinarip should be good to go now!

pettinarip (approved):
🚀

PR #11214, opened 2023-09-20
Add a new resource to help professionals looking for a job in Ethereum related projects

mpaskalev:
## Description

Added a new resource to help professionals looking for a job in Ethereum related projects. 
https://workwithweb3.com/ethereum


wackerow:
Hey @mpaskalev, sorry for delays here but appreciate the request. We've actually been looking at ways to potentially clean this list (and other lists) up given the already-large list of choices that are here. As a policy we try to not let these lists get too long as it starts to hurt the user experience by overwhelming with options.

After a look, I'm not seeing a whole lot different here that isn't already achieved with the listings already available. Going to close this out, but if you disagree please feel free to re-open and make the case for how this resource would add to what is already there, and how it would uniquely benefit users. Thanks!

cc: @minimalsm

PR #11236, opened 2023-09-21
Retouched the primary Colours

nloureiro:
This PR changes the color primates for the blue and orange colors and updates some design tokens.

## Description

As we move on to integrate the design system, I felt that the color scheme was good for some details, and this PR wants to fix them, keeping the same number of design tokens (if this does not work on some corners cases, we might need to add more design tokens in the future)

###goals

- Fix the darker colors in the dark theme; they were too brown and need to warm up with red to make them more orange.
- Finish the scale of both colors > from blue/orange 50 up to blue/orange 900. We needed some middle colors.
- The hover color was too intense in the primary buttons. The fix was to make it a little darker.

This is a screen shot form Figma. Here, I compare both color scales-
Screen Shot 2023-09-21 02 16 06 PM

This change was made on the Design system Figma file, where you can see more differences in components that still need to be coded.

Link to Figma > https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?type=design&node-id=9024%3A27036&mode=design&t=qPkpiEKut6DnlNaJ-1


pettinarip (approved):
Nice! looks better 💪🏼

PR #11243, opened 2023-09-22
Update smart contract development frameworks

miohtama:
- Deprecate Truffle, Brownie, OpenZeppelin SDK as they are currently unmaintained
- Update links and description of Foundry, as Foundry has now new home
- Move unmaintained frameworks to the bottom of the list, as developers likely do not want to use something that no longer works

## Description

As above.

## Related Issue

-


corwintines (approved):
LGTM, thanks @miohtama

PR #11249, opened 2023-09-24
Connect to Web3 sim updates

wackerow:
## Description
- Adds a new starting screen/step to the Connect to Web3 sim
- Introduced a new `Browser` component for that step that starts with the user needing to go to the website for the dapp, after receiving an NFT as part of the story line
- Adjusts some copy to fit the new screen flow
- Typo correction, and added "Mint new NFT" to dapp account screen

## Preview link
https://ethereumorgwebsitedev01-simupdates.gatsbyjs.io/en/wallets/?sim=connect-web3#sim

## Related Issue
None filed

### DIFF
--- a/src/components/Simulator/PulseAnimation.tsx
+++ b/src/components/Simulator/PulseAnimation.tsx
@@ -1,7 +1,12 @@
 import React from "react"
 import { Box } from "@chakra-ui/react"
 import { motion } from "framer-motion"
-import { CIRCLE, FULL_BUTTON, NARROW_BUTTON } from "./constants"
+import {
+  BASE_ANIMATION_DELAY_SEC,
+  CIRCLE,
+  FULL_BUTTON,
+  NARROW_BUTTON,
+} from "./constants"
 import type { PulseOption } from "./types"
 
 const MotionBox = motion(Box)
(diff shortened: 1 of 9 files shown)


konopkja:
Screenshot 2023-09-28 at 21 33 49

on second review 
@wackerow i think this naming is unclear. What does it mean web app?  Maybe we rename it to "Go to NFT market"? or something like that since the icon name is the same

konopkja:
> @konopkja Was just looking for a way to make it clean that this is a website (since _most_ dapps are effectively just websites, aka web apps, as their front-end)
> 
> Not picky though... "Go to NFT market" would work for me as well

problem is that "web app" is tech jargon that normal people dont use.

Either we just say "Visit website" or "Visit NFT market" i guess or something in those lines

wackerow:
Okay... I had something close to "Visit website" originally... I think that works nicely

wackerow:
Per offline discussion:
- Switched button label to "Visit NFT market"
- Extended the delay before the URL typing animation starts
- Added matching delay before enabling the button (also matches the pulse animation timing)
- Added "www globe" icon to make it more obvious that user is "in a browser"

Screen grab preview of mobile:

https://github.com/ethereum/ethereum-org-website/assets/54227730/4a0b02f3-6ea7-4e2b-8e86-985c7f42d19f

corwintines (approved):

PR #11250, opened 2023-09-24
fix(.storybook): import inter font css to preview

TylerAPfledderer:
## Description
The `Inter` font-face set in the local was not exposed in the Storybook server.

Here, importing the css file `inter-font-face.css` from the `static` folder to `preview.ts` should provide the necessary exposure.

(Not quite sure how that was missed 🤔 )

## Related Issue
N/A


corwintines (approved):
Nice catch

PR #11253, opened 2023-09-25
Update index.md

hipiyushhere:
Updating Spheron Network for Ethereum community to not just get cheaper solution but a great performant system.


minimalsm:
Hey @hipiyushhere, thanks for the PR. I don't think we want to add generic storage/compute solutions to this page. The other solutions (LMK if I'm wrong here) are blockchain specific solutions, optimised for that use case. If we add every storage solution we'll end up with an unmanagable list.

Going to close this PR out for that reason, but open to discussion here if I've got something wrong.

PR #11256, opened 2023-09-25
Typo - Missing word

vdusart:
I think the word `client` is missing


corwintines (approved):
Thanks @vdusart 

I think this makes sense. @wackerow could  you confirm this as well

wackerow:
Yeah, I feel like it's fine either way... it's typically bundled with the consensus client, but it is its own piece of software. Thanks @vdusart!

PR #11270, opened 2023-09-26
Chinese Simplified content buckets (Foundation docs, Foundational docs - Nodes and clients)

corwintines:
## Description

- Add Chinese Simplified content buckets (Foundation docs, Foundational docs - Nodes and clients)


wackerow (approved):
lgtm

Issue #11279, opened 2023-09-27
Add p2p.org as Staking as a Service Provider to the website


rgb2hsl:
### Project name

P2P.org

### Product type

Staking as a service

### Logo

![32x32](https://github.com/ethereum/ethereum-org-website/assets/3310192/2b19188e-8612-4337-b984-1efcfae1ea27)

### Description

P2P.org provides secure non-custodial staking services for professional investors.

### Website

https://p2p.org/networks/ethereum

### If software is involved, is everything open source?

Portions of code are open source:
https://github.com/p2p-org/p2p-eth2-depositor
https://github.com/p2p-org/eth-staking-fee-distributor-contracts

### Is the project a fork? If yes, which project was forked?

No

### Is the product out of beta development?

The product is out of beta and in production since January 2023 and actively maintained by p2p.org development team.

### What wallets support the product or service?

MetaMask, Safe, Ledger, Trust, MetaMak Mobile, WalletConnect

### If the product or service enables staking with <32 ETH, what is the minimum ETH required to stake?

Minimum amount is 32 ETH.

### If a service, what are the fees associated with using the service?

The fee is 5%.

### What date did the project or service go live?

Since January 10th 2023, two weeks after our first depositor contract have passed it's security audit

### Has the project undergone an external security audit?

First security audit of depositor smart-contract (Dec 2022) https://github.com/mixbytes/audits_public/blob/79cd8153bfb6098227b43dfba8d061881a94a248/P2P.org/ETH2%20Depositor%20%26%20ETH%20Staking%20Fee%20Distributor%20(v.1)/P2P.org_ETH2%20Depositor%20%26%20ETH%20Staking%20Fee%20Distributor%20Smart%20Contracts%20Audit%20Report.pdf

Second security audit of renewed depositor and fee-divider smart-contracts (April 2023)
https://github.com/mixbytes/audits_public/blob/79cd8153bfb6098227b43dfba8d061881a94a248/P2P.org/ETH2%20Depositor%20%26%20ETH%20Staking%20Fee%20Distributor%20(v.2)/ETH2%20Depositor%20%26%20ETH%20Staking%20Fee%20Distributor%20Security%20Audit%20Report%20(v.2).pdf

### Has the project undergone any security bug bounties?

No, we don't have bug bounty.

### Is the project being actively maintained?

Yes

### Is the product or service free of trusted/human intermediaries?

Yes, the product is free of trusted/human intermediaries. No KYC.

[...]

PR #11281, opened 2023-09-27
Added p2p.org as Staking as a Service Provider to the website

rgb2hsl:
1. Added P2P.org icon
2. P2P.org added to staking services as SaaS

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/11279


src/data/staking-products.json
```diff
@@ -403,6 +403,49 @@
...
+      "monthlyFee": 5,
+      "monthlyFeeUnit": "%",
+      "isFoss": false,
+      "hasBugBounty": false,
+      "isTrustless": true,
+      "isPermissionless": true,
+      "pctMajorityExecutionClient": 50,
+      "pctMajorityConsensusClient": 30,
+      "isSelfCustody": true,
+      "platforms": ["Browser", "Linux", "macOS", "Windows", "iOS", "Android"],
```
wackerow, line 434: If everything is through the browser, I would just use Browser here... the others are meant for native, non-web applications

```suggestion
      "platforms": ["Browser"],
```

src/data/staking-products.json
```diff
@@ -403,6 +403,49 @@
...
+      "additionalStakeUnit": null,
+      "monthlyFee": 5,
+      "monthlyFeeUnit": "%",
+      "isFoss": false,
+      "hasBugBounty": false,
+      "isTrustless": true,
+      "isPermissionless": true,
+      "pctMajorityExecutionClient": 50,
+      "pctMajorityConsensusClient": 30,
+      "isSelfCustody": true,
```
wackerow, line 433: From current "Self Custody" definition: 

> User maintains custody of any validator credentials, including signing and withdrawal keys

and your note:

> Client have no access to the signing keys

This wouldn't qualify currently, though I did raise this as an issue to potentially alter the definition here to account for the security of having withdrawal credentials set to the users own account: #11405

```suggestion
      "isSelfCustody": false,
```

I would encourage you to comment about this on that issue in the meantime 🙏

src/components/icons/staking/P2PGlyphIcon.tsx
```diff
@@ -0,0 +1,28 @@
...
+  viewBox: "0 0 32 32",
+  defaultProps: {
+    width: "32px",
+    height: "32px",
+    fill: "currentColor",
+  },
+  path: [
+    <path
+      fill-rule="evenodd"
+      clip-rule="evenodd"
```
wackerow, line 15: Small syntax note... these attributes should be switched to camelCasing instead of kebab-casing to conform with React:
```suggestion
      fillRule="evenodd"
      clipRule="evenodd"
```

src/components/icons/staking/P2PGlyphIcon.tsx
```diff
@@ -0,0 +1,28 @@
...
+    />,
+    <path d="M1.71212 18.5918C1.71212 18.5276 1.65785 18.4756 1.59091 18.4756H0.621212C0.554269 18.4756 0.5 18.5276 0.5 18.5918V19.5214C0.5 19.5855 0.554269 19.6376 0.621212 19.6376H1.59091C1.65785 19.6376 1.71212 19.5855 1.71212 19.5214V18.5918Z" />,
+    <path d="M24.0152 18.5918C24.0152 18.5276 23.9609 18.4756 23.8939 18.4756H22.9242C22.8573 18.4756 22.803 18.5276 22.803 18.5918V19.5214C22.803 19.5855 22.8573 19.6376 22.9242 19.6376H23.8939C23.9609 19.6376 24.0152 19.5855 24.0152 19.5214V18.5918Z" />,
+    <path d="M11.6515 18.5918C11.6515 18.5276 11.7058 18.4756 11.7727 18.4756H21.2273C21.2942 18.4756 21.3485 18.5276 21.3485 18.5918V19.5214C21.3485 19.5855 21.2942 19.6376 21.2273 19.6376H11.7727C11.7058 19.6376 11.6515 19.5855 11.6515 19.5214V18.5918Z" />,
+    <path
+      fill-rule="evenodd"
+      clip-rule="evenodd"
```
wackerow, line 23: same as above
```suggestion
      fillRule="evenodd"
      clipRule="evenodd"
```

wackerow (changes requested):
@rgb2hsl Hey! Thanks for the PR here... just gave it a review and things are overall looking good...
I left a couple change requests for ya, and if you're okay with these I think we can pull this in!

src/data/staking-products.json
rgb2hsl: True, regarding your definition we are not self custody at the moment. We just put different value in the "custody" 
 ownership of ETH. So we position ourselves self custody because we don't have access to client's ETH's. We still have plans for the future to make convenient yet self custody in your definition.
 
 So if the definition is so, I guess we have no options.

rgb2hsl:
@wackerow hey, I've accepted all the changes you've suggested. Let's move this on :)

src/data/staking-products.json
wackerow: Fully understand... this has been getting raised a bit more lately which was why I raised that issue. Again, if you wouldn't mind expressing some thoughts on that PR it would certainly help reassure the direction and help it gain some traction.

wackerow (approved):
Looks good! Thanks @rgb2hsl!

wackerow:
@all-contributors please add @rgb2hsl for content

PR #11285, opened 2023-09-27
Hindi content buckets (Essentials, Exploring, Use Ethereum, Use cases)

corwintines:
## Description

- Hindi content buckets (Essentials, Exploring, Use Ethereum, Use cases)


wackerow (approved):

minimalsm (approved):
Building locally for me too... bringing in as there is a big build backlog and we can deal with it if it breaks dev 😬

PR #11286, opened 2023-09-27
Hungarian content buckets (Homepage, Essentials, Exploring, Use Ethereum)

corwintines:
## Description

- Hungarian content buckets (Homepage, Essentials, Exploring, Use Ethereum)


wackerow (approved):

PR #11312, opened 2023-09-29
style(ButtonTwoLines): alter `size` and `color` prop usage for text

TylerAPfledderer:
## Description

- Primary: use `body.medium` color for helper text with `outline` variant or `isSecondary` (per DS)
- Secondary: use `size` theme prop instead of `fontSize` to include `lineHeight` for both `Text` components

## Related Issue
N/A


pettinarip (approved):
🎉 Approved and merging now! 🙏🏼

PR #11324, opened 2023-10-01
refactor: update Storybook docs and usage

TylerAPfledderer:
## Description

This PR updates the `applying-storybook` docs to match the latest usage of Storybook, and updates existing stories to match said usage.

This PR recognizes that these docs and approach to using storybook for the project might be considered fluid as implementation of the new DS still continues, and further contributions from the community can alter the approach in the future for improved DX.

## Related Issue
N/A


src/components/BannerNotification/BannerNotification.stories.tsx
```diff
@@ -1,28 +1,34 @@
...
-      <Text>{t("page-upgrades-post-merge-banner-tutorial-ood")}</Text>
-    </BannerNotification>
-  )
+export const PostMergeBanner: StoryObj<typeof meta> = {
+  args: {
+    shouldShow: true,
+  },
+  render: (args) => {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const { t } = useTranslation("page-upgrades")
```
wackerow, line 26: Oh interesting, I wasn't sure we could use hooks in here, good to know 🙏

src/components/BaseStories/Link.stories.tsx
```diff
@@ -45,7 +43,7 @@ export const ExternalLink: Story = {
   render: (args) => <MockParagraph {...args} />,
 }
 
-export const LinkList: Story = {
+export const LinkList = {
```
wackerow, line 46: @TylerAPfledderer So if I understand correctly, if we're not supplying `args` (props) then we don't need to provide a type here?... but if we are, we should provide `StoryObj<typeof meta>` here, correct?

src/components/Select/Select.stories.tsx
```diff
@@ -10,16 +10,14 @@ const selectData = [
...
-}
+} satisfies Meta<typeof SelectComponent>
 
 export default meta
 
-type Story = StoryObj<SelectType>
+type Story = StoryObj<typeof meta>
 
 export const Select: Story = {
   args: {
```
wackerow, line 23: @TylerAPfledderer Getting some merge conflicts on this file (`Select.stories.tsx`)... I think we probably want to overwrite with what we have here, but I'm getting a type error with `minW` in the args when I try to do that, since we seem to be attempting to spread that onto ReactSelect (not a Chakra-UI component).

I think we need to remove this `minW` arg, but wanted to check with you.

src/components/BannerNotification/BannerNotification.stories.tsx
TylerAPfledderer: Storybook does not seem to like this, where better would be to make a wrapper component to then apply to this render. 

But this is quicker 😅

src/components/BaseStories/Link.stories.tsx
TylerAPfledderer: @wackerow This probably needs clarity (including for myself!)

Because the `StoryObj` as a generic takes an optional prop type, it should be used on every story even if it's just 
```tsx
export const ExampleStory: StoryObj = {}
```
Because we still want to type the available options in the object.

I initially chose this removal because it didn't make sense to have `StoryObj<typeof meta>` when this is a complex story render that didn't need the args option.

So with the `LinkList` story, its type should be define as just `StoryObj` to maintain the base consistency.

---

Now with the choices in supplying a prop type to `StoryObj`, it can either be:
1. `StoryObj<typeof meta>`
2. `StoryObj<typeof ComponentName>`

With No. 1, [per the SB docs](https://storybook.js.org/docs/writing-stories/typescript#using-satisfies-for-better-type-safety) in conjunction with using `satisfies` to the `meta` object, you would apply this in the event you supply `args` to meta to be distributed to all the stories. This prevents TypeScript from throwing errors about missing required props at the story level if they were defined at the meta level.

No. 2 is for just the story level and not the meta level.

I have overlooked this specific detail, but I think `StoryObj<typeof meta>` also stays with consistency and the story file receives the most typing coverage.

> As an aside, not the details about custom args in the next section on that referenced page.

What do you think?

src/components/Select/Select.stories.tsx
TylerAPfledderer: This version of the select stories file is outdated and that arg no longer exists. Fixed the conflict to ensure the file currently in prod stays intact. (It's now removed from the diffs in this PR, btw)

TylerAPfledderer:
> For my own understanding, is there any need/benefit to match the text in these stories (ie a label, or children text) to the text found in the Figma DS components?

@wackerow I would argue that doing this would make it helpful for Nuno with consistency in verifying the design has been implemented to code as intended. Applying some other content may bring out an edge case or an underlying bug with widths or spacing, but also using the content from Figma would provide a true baseline in the conversion to the Storybook visual testing environment.

Side note: "consistency" seems to be the unfortunate buzzword for me 🤣

wackerow:
Merged... thanks @TylerAPfledderer! I agree that we should strive to match the text to Figma where possible, especially on the base components. Can address that separate from this.

Issue #10924, opened 2023-08-14
improve the left side navigation


nloureiro:
### Is your feature request related to a problem? Please describe.

The left-side navigation on content pages raised some UX problems with readability and off-screen content.

These issues might be improved in a re-design that should be done under the Design system changes, but meanwhile we can improve the UX of the page with some quick CSS changes.
![Screen Shot 2023-08-14 02 51 03 PM](https://github.com/ethereum/ethereum-org-website/assets/1120748/82d0def4-1b73-476c-ad1b-2f61c639f78c)

### Describe the solution you'd like

![Screen Shot 2023-08-14 02 51 59 PM](https://github.com/ethereum/ethereum-org-website/assets/1120748/7e62dea4-2376-4947-a28c-4e2c51a1dd50)

This is a screenshot of this Figma file where you can find the proposed changes.
https://www.figma.com/file/ut0RrfN81MLW28Do4X36kQ/sidebar-ux-quick-fix?type=design&node-id=37%3A235&mode=design&t=4H9l9AD1xYCSrosy-1

this would be the finished outcome
![Screen Shot 2023-08-14 02 53 16 PM](https://github.com/ethereum/ethereum-org-website/assets/1120748/a34f459d-8161-4e8b-9f60-111a91083fc3)

These changes are targeted at desktop resolutions. The mobile version should not change.

### Describe alternatives you've considered

I discussed this proposed change with the team, and moving forward is OK.

### Would you like to work on this issue?

- [ ] Yes
- [ ] No

corwintines:
@rohan9024 would you like to work on this?

rohan9024:
@corwintines yeah sure. Can you please assign me ?

priyadarshss:
@minimalsm Would it be possible to merge my PR if I finish it earlier although it has been assigned to someone else? I have already begun working on it, which is why I am asking.

rohan9024:
> @minimalsm Would it be possible to merge my PR if I finish it earlier although it has been assigned to someone else? I have already begun working on it, which is why I am asking.

You can work on other issues which haven't been assigned to anyone yet😊

priyadarshss:
@rohan9024 Works for me :). Thank you! 
I've pretty much fixed this part: 
![image](https://github.com/ethereum/ethereum-org-website/assets/58765597/e1b6232d-d1a6-480b-8a4b-6a01aca56674)
[...]

PR #11330, opened 2023-10-02
Improve the left side navigation

corwintines:
## Description

- Abstract out repeated LeftNavBar logic from all templates into one component
- Implement new styles in LeftNavBar
- Refactor templates to use new LeftNavBar component

## Related Issue

Fixes https://github.com/ethereum/ethereum-org-website/issues/10924


pettinarip (commented):
Nice!

Looks good. The [learn page](https://ethereumorgwebsitedev01-improveleftsidebar.gatsbyjs.io/en/learn/) and the event template are missing this migration.

Wondering also if we could remove the deprecated components like `InfoColumn`  and `InfoTitle`.

corwintines:
@pettinarip @konopkja thanks!

@konopkja ive fixed the title

@pettinarip ive added the missing migrations and removed unused components now from use-cases template.

src/templates/roadmap.tsx
pettinarip, line 1: Noting that this is the only template where the sidebar is not sticky. 

Is this a deliberate decision or should we keep it consistent with the rest?

e.g.
in https://ethereumorgwebsitedev01-improveleftsidebar.gatsbyjs.io/en/roadmap/
if you scroll, the sidebar is static and not sticky as in the other templates

src/components/LeftNavBar/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+} from "../ButtonDropdown"
+import OldHeading from "../OldHeading"
+import Translation from "../Translation"
+import UpgradeTableOfContents from "../UpgradeTableOfContents"
+
+export const H2 = (props: HeadingProps) => (
+  <OldHeading
+    fontWeight={700}
+    lineHeight={1.4}
+    fontSize="2xl"
```
pettinarip, line 16: Using `fontSize="2xl"` will produce a fontSize of `24px` which is a size not defined in the DS for headings.

@konopkja @nloureiro since this is a new design, I'm wondering if we should go ahead and implement one of the predefined heading sizes from the DS. The closest sizes we have for this case is whether the `md` (20px) or `lg` (30px) size.

image

@corwintines if that is the case, then we could do something like this
```tsx
<Heading size="md"> // or `lg` whatever is decided
```

nloureiro: hum.... I think I found out the problem. 

on the design system, we have 2 scales > 1 for heading and another for text

In the design proposal I used the node for text > 2xl and I probably should have used the headings scale > In this case you are right it's LG for the equivalent (Heading LG=30px on desktop and 24 on mobile)

I think we can change to heading LG

src/templates/roadmap.tsx
corwintines: I don't think this was a deliberate decision, but something overlooked in the implementation.

Ill look into whats going on here

corwintines: Im really not sure whats going on here...

pettinarip: There was a `overflow: hidden` style in the roadmap container. Having an overflow in the container makes the sticky positioned children to now work as expected.

pettinarip (approved):

PR #11331, opened 2023-10-02
Update date Ethereum Costa Rica Day

Coderoostr:
gm


minimalsm (approved):
LGTM thanks @Coderoostr 🎉 

Could you update the website to point to the event (I assume you're involved in Ethereum Costa Rica)? The website has no mention of it.

Will merge this change as the event is already live, but I personally wouldn't have added this until the website had a clear CTA for the event.

PR #11333, opened 2023-10-02
style: declare secondary styling for outline buttons in "primary/secondary" button pairs

### LINKED ISSUE #10739: Secondary buttons styling does not correctly reflect design system
### Describe the bug

Across the website, we are using the primary outline variant of the button instead of the secondary variant of the button.

E.g., in the design system, we use a secondary button (black outline) for the secondary button.

Screenshot 2023-07-20 at 10 00 54

Here is the difference:

Screenshot 2023-07-20 at 09 57 44 1

However, on production, we use the primary 'outline' variant ([see Layer 2 page](https://ethereum.org/en/layer-2/)) also everywhere.

Screenshot 2023-07-20 at 10 02 49

We accidentally do the same on the homepage:

Screenshot 2023-07-20 at 10 08 33

And on the docs:

Screenshot 2023-07-20 at 10 08 40

### To reproduce

1. Go to 'https://ethereum.org/en/layer-2/'
1. See error

### Expected behavior

I'm currently unsure about the correct button designs. It appears that around 90% of our buttons should feature black outlines (secondary outline).

To avoid inconsistencies, I suggest we standardize all outline button variants to secondary outline buttons. We can revise this as needed once we establish clearer design rules or exceptions.

TylerAPfledderer:
## Description
Declare `isSecondary` prop on buttons with `outline` variant that are in pairs with a button that is considered the primary action (has `solid` variant)

## Related Issue
Fixes #10739


src/components/PageHero.tsx
```diff
@@ -99,6 +99,7 @@ const PageHero = ({
         {buttons && (
           <Wrap spacing={2} overflow="visible" sx={{ ul: { m: 0 } }}>
             {buttons.map((button, idx) => {
+              const isSecondary = idx !== 0
```
nhsz, line 102: Does this work ok with RTL layouts?

nhsz (approved):
LGTM, thanks!

nhsz (approved):

PR #11341, opened 2023-10-03
Copy/content polishes on `/roadmap/account-abstraction/`

m4sterbunny:
- Applies US English
- Improves punctuation
- Deprecates whitelist: Consider retiring whitelist in favor of allowlist as per the [Google Developer's style guide](https://developers.google.com/style/word-list#whitelist). While whitelist and blacklist are embedded terms in tech, here we are presenting cutting-edge technology and arguably should do so in modern terms. Allowlist and blocklist are explicit, translate well, and aren't tainted with cultural implications. I believe it is time to evolve. WDYT?


wackerow (approved):
Looks good! Thanks @m4sterbunny for the improvements, and apologies for delays in reviewing. Agree with your suggestions and changes; approving 🙂

wackerow:
@all-contributors please add @m4sterbunny for maintenance

PR #11352, opened 2023-10-04
Update block-explorers list by adding ethstakers.club

anumJ12:
## Description

Added a new entry to the block-explorers list with a link to "ethstakers.club."

### DIFF
--- a/src/content/developers/docs/data-and-analytics/block-explorers/index.md
+++ b/src/content/developers/docs/data-and-analytics/block-explorers/index.md
@@ -15,6 +15,7 @@ You should understand the basic concepts of Ethereum so you can make sense of th
 
 - [Etherscan](https://etherscan.io/) -_Also available in Chinese, Korean, Russian, and Japanese_
 - [Beaconcha.in](https://beaconcha.in/)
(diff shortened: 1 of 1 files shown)


wackerow:
Hey @anumJ12, apologies for the delays on this. Just took a look; is this product still active? 

- I see 1.5 months of activity in the GitHub repo then nothing for the past 5 or so. 
- The website is showing "Epoch 0, Slot 0"
- Querying for a validator shows me the correct pub key, but clicking on it sends me to a `/validator/` path ([for example](https://ethstakers.club/validator/420)) which gives a `500` server error
- Client breakdown lists out-of-date clients
- Slots/Epochs pages seem to show some data in the chart, but it stopped in January.
- I did not sign up; not sure if I need to for the basic features to work, but we should not be listing block explorers on this page that require an account and login in order to have basic access (not saying that's the case here, just noting)

Let me know if I'm missing anything... I do appreciate that this is targeting a free, ad-free, no-tracking experience, but currently it doesn't appear to be very usable from my end. I'm going to close this out for now, but **please** re-open and **@ me** if you disagree here or have more information and can take another look. Thanks!

PR #11380, opened 2023-10-05
Remove past events from community-events.json

Shiva-Sai-ssb:
## Description

I've excluded the events that have already been completed from the community-events.json file.


corwintines (approved):
Thanks @Shiva-Sai-ssb!

PR #11383, opened 2023-10-06
Fix the hover color on dark theme

nloureiro:
we found the need to change the color of the visited links on dark theme


pettinarip (approved):
Looks good! nice @nloureiro merging

PR #11437, opened 2023-10-14
Remove note about Erigon binaries

ardislu:
Erigon has provided pre-built binaries on their [GitHub releases page](https://github.com/ledgerwatch/erigon/releases) since release [v2.39.0](https://github.com/ledgerwatch/erigon/releases/tag/v2.39.0) (released February 19, 2023), so the note is no longer accurate.


corwintines (approved):
Thanks @ardislu!

PR #11443, opened 2023-10-15
fix: fixed the date format

### LINKED ISSUE #11335: Incorrect intl number formatting
### Describe the bug

Dates on event cards are not being formatted using intl locale.

### To reproduce

1. Go to https://ethereum.org/zh/community/events/
2. Notice date format in event cards compared to date at the top of the page.

### Expected behavior

Despite the contents of these cards not being translated into other languages, I would expect these dates to be rendered according to the locale.

The [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) should be used by passing the current locale and the date to render accordingly.

### Screenshots

image

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes
- [X] No

ameeetgaikwad:
## Related Issue
fixes #11335

### DIFF
--- a/src/components/UpcomingEventsList.tsx
+++ b/src/components/UpcomingEventsList.tsx
@@ -50,6 +50,12 @@ const UpcomingEventsList: React.FC<IProps> = () => {
     )
   }
 
+  const formatDate = (date: Date): string => {
+    return (
+      date.getFullYear() + "-" + (date.getMonth() + 1) + "-" + date.getDate()
+    )
+  }
+
   useEffect(() => {
     const eventsList: Array<ICommunityEventData> = [...events]
     const yesterday = new Date()
@@ -70,10 +76,10 @@ const UpcomingEventsList: React.FC<IProps> = () => {
     const formattedEvents = orderedEvents.map((event) => {
       const dateRange =
         event.startDate === event.endDate
-          ? dateParse(event.startDate).toLocaleDateString()
-          : `${dateParse(event.startDate).toLocaleDateString()} - ${dateParse(
-              event.endDate
-            ).toLocaleDateString()}`
+          ? formatDate(dateParse(event.startDate))
+          : `${formatDate(dateParse(event.startDate))} - ${formatDate(
+              dateParse(event.endDate)
+            )}`
 
       const details = `${event.sponsor ? "(" + event.sponsor + ")" : ""} ${
         event.description


src/components/UpcomingEventsList.tsx
```diff
@@ -50,6 +50,12 @@ const UpcomingEventsList: React.FC<IProps> = () => {
     )
   }
 
+  const formatDate = (date: Date): string => {
+    return (
+      date.getFullYear() + "-" + (date.getMonth() + 1) + "-" + date.getDate()
+    )
+  }
+
```
wackerow, line 58: Hey @ameeetgaikwad, thanks for the PR and sorry for delays... JavaScript has a library called `Intl` which should be used here. It allows us to pass it the users locale (language/region) to format the date appropriately for their region.

Examples:
image

For this repo, we're right in the middle of migrating from using Gatsby to using Next.js... SO, I wouldn't update this _right now_, but very soon when the [Next.js repo fork](https://github.com/ethereum/ethereum-org-fork) has been merged into this repo (within 1-2 weeks), we can access the users "locale" using the `useRouter` hook:

```tsx
// example use
// ...then inside a component function:
// ...then use this locale with the built-in Intl.DateTimeFormat function to format `date` accordingly
  Intl.DateTimeFormat(locale).format(date)
```

To avoid redundancy, you could makes a helper function inside the component so it has access to `locale`, something like this:

```tsx
  const formatDate = (date: Date): string => new Intl.DateTimeFormat(locale).format(date)
```

Then use `formatDate` below the same way you are currently doing.

wackerow (changes requested):

src/components/UpcomingEventsList.tsx
ameeetgaikwad: @wackerow  Sure, I'll wait for the merge. 

Also rather than using `formatDate` can we like this?: 
image

wackerow: I would suggest using the `Intl` package here (although admittedly results would be similar, we've just been trying to standardize around using the `Intl` package for these types of things)...

Ideally we pull this out as a global helper function since we could use it in multiple places on the site. Again, _after_ the Next.js migration is complete (moving along, should be ready very soon), I would look to add this to `src/lib/utils/date.ts` as a new exported function where it takes in a date and a locale, and returns the formatted date string for that locale. Then we can import that wherever we need it to reduce redundancy.

ameeetgaikwad: @wackerow sure then I'll add the function in `src/lib/utils/date.ts` . And use that function here.

wackerow: Hey @ameeetgaikwad! Just checking in, were you still interested in making these updates? If not no worries, can assign someone else, just checking

nhsz:
Closing as its duplicated (https://github.com/ethereum/ethereum-org-website/pull/12287)

PR #11463, opened 2023-10-18
Added new audit company

Andrewsnobre:
Added new audit company.

## Description

Added new audit company - Smartzation.com


wackerow:
Hey @Andrewsnobre, sorry for delay here... Couple things on this, I see you added it to the Portuguese Brazilian version of the page... we would want to add this to the English page first, and then it would go through a process of being translated and updated in all other languages. 

Also, would use the `https://smartzation.webflow.io` link instead of the one you listed (getting SSL cert issues with the one you posted, and it seems to simply forward to the one I just mentioned)

If I may, I'm going to close this one out in the meantime. If you're still interested in listing this I'd request that we [open an issue](https://github.com/ethereum/ethereum-org-website/issues/new?assignees=&labels=feature+%3Asparkles%3A%2Ccontent+%3Afountain_pen%3A&projects=&template=suggest_dev_tool.yaml&title=Suggest+a+developer+tool) for this first with planned details, and if it looks good and once assigned, a follow-up PR could be opened against that issue. Again, initial content updates should always be in English, and the translation process will be handled separately. Thanks!

PR #11477, opened 2023-10-19
refactor(StakingProductsCardGrid): update types, logic, and UI composition

TylerAPfledderer:
## Description

This PR overhauls the structure of `StakingProductsCardGrid`.

- Update types to favor of improved constraints
  - Created type signatures from the `stakingProducts` json data
  - Alter existing logic and types to match the new constraints
- Code-split UI and logic for improved readability

## Related Issue
N/A


F3n4nd0gm:
Ok no tengo pc

src/components/Staking/StakingProductsCardGrid/StakingProductCard.tsx
wackerow, line 1: Is it worth breaking this component apart as well?

wackerow (approved):
This looks great @TylerAPfledderer, thanks! Much cleaner. Working well for me and think we can pull this in.

Left a question on `StakingProductCard.tsx`, but that doesn't have to block this.

src/components/Staking/StakingProductsCardGrid/StakingProductCard.tsx
TylerAPfledderer: @wackerow Hmm I'm not sure. If I continue to break it down, I guess it would be making it's own useHook to abstract away the data and logic. I don't see the usefulness in splitting up the component structure; to me it's not big enough.

I don't think there is enough there right now to justify it unless there becomes an issue in readability.

PR #11496, opened 2023-10-24
Update index.md

Swiftevo:
Update DeSci Asia in DeSci project

## Description

Update DeSci Asia in DeSci project for connecting DeSci friends in Asia.


minimalsm (commented):
Personally I don't think a small telegram group is mature enough to be considered as something we should link out to on this page.

wackerow:
A user experience issue in my opinion... the link should ideally work for everyone not just those who have Telegram. Going to close this one out

PR #11514, opened 2023-10-26
Added references to Web3 sveltekit bundle

bumi001:


minimalsm (commented):
Hey @bumi001, thanks for the PR and creating this bundle.

To stop users from getting cognitively overloaded, we avoid listing products that do not have significant traction already (e.g. this has 0 stars, 0 forks, and no activity since it was created)

Feel free to open a PR once this is a bit more stable. Thanks again!

PR #11530, opened 2023-10-28
Fix layer 2 button on gas page

s-crypt:
## Description

Make "Use Layer 2" button go to layer 2 page instead of the get eth page on the gas learning page


corwintines (approved):
Thanks @s-crypt!

corwintines:
@all-contributors can you add @s-crypt for content

PR #11536, opened 2023-10-29
Add .replace(/\\./g, "/") to gatsby.node 

corwintines:
## Description

- add .replace(/\\./g, "/") to `const dirname = path.resolve('./)`

### DIFF
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -38,7 +38,7 @@ const commonRedirectProps = {
  * @returns boolean for if file is outdated or not
  */
 const checkIsMdxOutdated = (filePath: string): boolean => {
-  const dirname = path.resolve("./")
+  const dirname = path.resolve("./").replace(/\\./g, "/")
   const splitPath = filePath.split(dirname)
   const tempSplitPath = splitPath[1]
   const tempSplit = tempSplitPath.split("/")
--- a/gatsby-node.ts
+++ b/gatsby-node.ts
@@ -38,6 +38,7 @@ const commonRedirectProps = {
  * @returns boolean for if file is outdated or not
  */
 const checkIsMdxOutdated = (filePath: string): boolean => {
+  // .replace(/\\./g, "/") to replace \ in windows paths ex: C:\\folder\\myfile.txt becomes C:/folder/myfile.txt
   const dirname = path.resolve("./").replace(/\\./g, "/")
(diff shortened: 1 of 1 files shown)


corwintines:
@TylerAPfledderer do you mind confirming this fixes the tempSplit issue for you 🙏

gatsby-node.ts
```diff
@@ -38,7 +38,8 @@ const commonRedirectProps = {
  * @returns boolean for if file is outdated or not
  */
 const checkIsMdxOutdated = (filePath: string): boolean => {
-  const dirname = path.resolve("./")
+  // .replace(/\\./g, "/") to replace \ in windows paths ex: C:\\folder\\myfile.txt becomes C:/folder/myfile.txt
+  const dirname = path.resolve("./").replace(/\\./g, "/")
```
TylerAPfledderer, line 42: It actually should not have the dot.  It is not used for the problem we are trying to solve, and it's deleting the first letter after each slash.
```suggestion
  // .replace(/\\/g, "/") to replace \ in windows paths ex: C:\\folder\\myfile.txt becomes C:/folder/myfile.txt
  const dirname = path.resolve("./").replace(/\\/g, "/")
```

TylerAPfledderer (changes requested):
@corwintines I gave you an error in this regex I suggested.

gatsby-node.ts
corwintines: Woops, mistyped it. Thanks!

nhsz (approved):
LGTM 🪟

gatsby-node.ts
```diff
@@ -38,7 +38,8 @@ const commonRedirectProps = {
  * @returns boolean for if file is outdated or not
  */
 const checkIsMdxOutdated = (filePath: string): boolean => {
-  const dirname = path.resolve("./")
+  // .replace(/\\/g, "/") to replace \ in windows paths ex: C:\\folder\\myfile.txt becomes C:/folder/myfile.txt
+  const dirname = path.resolve("./").replace(/\\/g, "/")
```
pettinarip, line 42: There are a few other places were we use `path.resolve`. Wondering why we wouldn't have this problem in those cases.

Also, `path.resolve` takes care of this problem https://nodejs.org/api/path.html#windows-vs-posix

I'm trying to understand why we need to do this and why this is not breaking in other places.

PR #11552, opened 2023-11-01
Update index.md

jeremyfritzen:
Updated info about zkSync and Added Starknet.


corwintines (approved):
Thanks @jeremyfritzen 

I updated the copy here a bit. Starknet is a zk-rollup, not a Validium. StarkWare does have a validium chain, but that is StarkEx not Starknet.

corwintines:
@all-contributors can you add @jeremyfritzen for content

PR #11561, opened 2023-11-02
Fix punctuation

bocarw121:
## Description

Removed extra comma that was added after the period


darigovresearch (approved):
Believe that either option works as `etc` is an acronym but the latter looks less weird as it doesn't have a comma after a full stop

https://en.wikipedia.org/wiki/Et_cetera

bocarw121:
> Believe that either option works as `etc` is an acronym but the latter looks less weird as it doesn't have a comma after a full stop
> 
> https://en.wikipedia.org/wiki/Et_cetera

Yea that totally makes sense, I missed the fact that it was after an etc. So whatever
works best for you is totally fine with me. Thanks for the link.

corwintines (approved):

corwintines:
Thanks @bocarw121 

I think both are correct. I approved, but I do think I prefer having the comma there as a reader given this isn't a hard stop in the sentence. I'll close this out if no one else approves, but im not strongly against it if others agree.

PR #11576, opened 2023-11-05
Content revisions for dev docs /scaling/

mcmoodoo:
## Description

Merely syntax and punctuation revisions


wackerow (changes requested):
Thanks @mcmoodoo! Sorry again for delay; left a few requests/suggestions then think we can pull this in.

corwintines (approved):
Committed @wackerow suggestions, but this looks good to me now. Thanks @mcmoodoo!

PR #11585, opened 2023-11-05
fix: Corrects transaction links in glossary

darigovresearch:
## Description
Two terms in the glossary have a dead link to another term in the glossary, see diff for more details.

## Related Issue
Thought this issue was small enough to not need to make an issue but happy to do so if needed. Internal link is #transaction & not #transactions

Issue found while making educational flashcards based off of the glossary https://github.com/darigovresearch/Ethereum-dot-Org-Glossary


corwintines (approved):
Thanks @darigovresearch

PR #11595, opened 2023-11-06
Fix flashing color mode issue

### LINKED ISSUE #7727: Flashing color mode on hard refresh
**Describe the bug**

When the user is using dark mode and a hard refresh is done, the user will see the site in light mode for a few seconds until the page loads. Once it finishes, dark mode will be applied as expected.

This behavior produces a flashing effect that is not desired.

**To Reproduce**

1. Go to https://ethereum.org/en/
2. Select dark mode
3. Do a hard refresh
4. For a second you will see the site in light mode and quickly after that it turns back to dark mode

**Expected behavior**

If the user is using dark mode, when the user hard refreshes the site, the page loads immediately in dark mode.

**Additional context**

Chakra already provides a [solution for this problem](https://chakra-ui.com/docs/styled-system/color-mode#add-colormodemanager-optional-for-ssr). The problem is that that solution is not available in the [Gatsby plugin that we are using](https://www.gatsbyjs.com/plugins/@chakra-ui/gatsby-plugin/). Additionally, I have opened [an issue on Chakra's repo](https://github.com/chakra-ui/chakra-ui/issues/6621) to see if someone can help us with it.

Digging more into the issue, I tried to set up Chakra without using the plugin.

My idea was on using `cookieStorageManagerSSR` inside the `gatsby-ssr.ts` file BUT, as the manager name indicates, it requires you to pass the cookies to it. For some reason, I can not access the cookies inside `gatsby-ssr.ts` which blocks me to solve this issue (and I guess it is the same issue the plugin authors faced).

**Want to contribute?**

pettinarip:
## Description

Uses the defined background semantic token for the body bg color. This will fix the flashing color issue.

More context about this fix in [this comment](https://github.com/ethereum/ethereum-org-fork/pull/105#issuecomment-1793285311) from @TylerAPfledderer.

Fixes #7727

### DIFF
--- a/src/@chakra-ui/gatsby-plugin/styles.ts
+++ b/src/@chakra-ui/gatsby-plugin/styles.ts
@@ -1,7 +1,5 @@
-import { mode } from "@chakra-ui/theme-tools"
-
 const styles = {
-  global: (props) => ({
+  global: {
     /**
      * THESE ARE OLD GLOBAL STYLES - carried over from old css files
      *
@@ -12,9 +10,7 @@ const styles = {
      * don't need the global styles anymore
      */
     body: {
-      // TODO: when we have Chakra v2, this should be done by overriding the
-      // native Chakra semantic tokens
-      bg: mode("white", "gray.800")(props),
+      bg: "background.base",
       lineHeight: "base",
       fontSize: "md",
     },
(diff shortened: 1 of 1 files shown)


TylerAPfledderer (approved):
@pettinarip it appears to be working splendidly! 😄

nnnnnnsssss (approved):

corwintines (approved):
Amazing, thank you @TylerAPfledderer !

PR #11599, opened 2023-11-06
Nov 6 Farsi and Japanese

corwintines:
## Description

- Nov 6 Farsi (Learn pages, Learn hub & guides, Community, Foundational docs)
- Nov 6 Japanese (Upgrades)

## Related Issue

Prereq https://github.com/ethereum/ethereum-org-website/pull/11598 as this was base


src/content/translations/ja/roadmap/index.md
```diff
@@ -0,0 +1,117 @@
...
+
+将来のプロトコルの具体的な改善案については、イーサリアムのロードマップに概要が描かれています。 このロードマップは、イーサリアムのユーザーに次のメリットをもたらします。
+
+<CardGrid>
+  <RoadmapActionCard
+    to="/roadmap/scaling"
+    title="より安価なトランザクション"
+    image="scaling"
+    description="Rollups are too expensive and rely on centralized components, causing users to place too much trust in their operators. The roadmap includes fixes for both of these problems."
+    buttonText="More on reducing fees"
```
wackerow, line 31: Realizing these were not translated... @lukassim it looks like the `title` field above it being translated correctly... is there a way in Crowdin we can flag `description` and `buttonText` as fields that also need to be translated?

wackerow (approved):
Left a comment about some strings that still need to be translated, but don't see that as a blocker to get the rest of this in. Merging

PR #11615, opened 2023-11-07
Update quizzes home hero, use new illustration

wackerow:
## Description
- Switch quizzes home hero to use `HubHero` component
- Import new illustration and use with new hero


src/pages/quizzes.tsx
```diff
@@ -223,7 +219,7 @@ export const query = graphql`
         }
       }
     }
-    doge: file(relativePath: { eq: "doge-computer.png" }) {
+    heroImage: file(relativePath: { eq: "heroes/quizzes-hub-hero.png" }) {
```
corwintines, line 222: The image looks a little grainy with these image settings. In the Learn page we use the following:

```
childImageSharp {
        gatsbyImageData(
          width: 1504
          layout: CONSTRAINED
          placeholder: BLURRED
          quality: 100
        )
      }
      ```
      
  I think we just need to increase the width here and should be good to go

wackerow: Hm... I don't think we need to increase the width since that is the max width the container will allow... cc: @nloureiro we may need to re-export this and make sure it doesn't get compressed too much

wackerow:
I agree that it looks degraded

src/pages/quizzes.tsx
wackerow: I agree that it looks degraded though

pettinarip: Given that this image is going to be much bigger (~1500px) than the previous one (~600px), I think we should increase the size of the image query param. Otherwise, the container will stretch the image.

Another way to handle it could be to use the `layout: FULL_WIDTH`. That will fetch the biggest possible size of the image.

nloureiro: Can we use .jpg here?

There is no benefit in using PNG because there is no transparency, and using JPG will do a much better job compressing the image.

I scaled down the image to 1600px width and exported it as a JPG

Play

wackerow: Will let it build and take a peak first

wackerow: Just eyeing the jpg file though, it looks fine visually, but the file size is even larger. If the `FULL_LAYOUT` update improves the appearance I think we can just keep the png

nloureiro: > Just eyeing the jpg file though, it looks fine visually, but the file size is even larger. If the `FULL_LAYOUT` update improves the appearance I think we can just keep the png

are we doing any compression on our side? I know we do it for other images, not sure if we do it for hero images

corwintines: I think the hero images were left out. Can you confirm @pettinarip?

pettinarip: We are setting `quality: 100` in the query, so there shouldn't be compression at all but, the images are being processed anyway by Gatsby (`sharp` lib). The process is cropping the image in multiple sizes, so I'd expect not to be the exact same image as the source one.

pettinarip: In summary, no, we are not compressing the hero images.

corwintines (approved):

PR #11666, opened 2023-11-14
fix(HubHero): create semantic token for content background color

TylerAPfledderer:
## Description

In prod, the content background of the `HubHero` component suffers from color flash in dark mode. This PR creates a semantic token in place of `useColorModeValue` to render the color.

## Related Issue
N/A


corwintines (approved):
🔥 nice

PR #11672, opened 2023-11-16
Use Merkle tree root be scpecific

howjmay:
## Description

Using `Merkle tree` is not clear, so replace it with  `Merkle tree root`


corwintines (approved):
Thanks @howjmay

PR #11714, opened 2023-11-24
Update description of codeHash regarding CREATE2

gsalzer:
The `codeHash` field was described as not changing. Since CREATE2, and also before with self-destruct, this field can change.


lightclient:
Now that cancun is live on mainnet, the original text is actually correct. This is because `SELFDESTRUCT` no longer allows metamorphic contracts*.

* The caveat is that metamorphic contracts are still possible intra-transaction, but I don't think it is worth noting.

wackerow:
Thanks @lightclient, appreciate your input on the latest details. 🙏 

@gsalzer Appreciate the post! I do apologize that we didn't get to this sooner as our team has been working through a backlog that mounted in Q4 last year.

Given the new updates to the network I'm going to close this out, but please feel free to open a fresh PR if you'd like to adjust anything else!

wackerow:
@all-contributors please add @gsalzer for idea

wackerow:
@all-contributors please add @lightclient for maintenance

PR #11739, opened 2023-11-29
Roadmap hero

corwintines:
## Description

- Adds new hub hero image
- Removes some unused hero code from /developers I found
- Changes image prop in roadmap markdown
- Implement HubHero in roadmap template


src/assets/heroes/roadmap-hub-hero.jpg
pettinarip, line 1: The resolution of the image on the page is not good because the query is cropping the image at 600px.
image

We might need to query this image separately and fetch the full-size image since this is now used as a hero image.

nloureiro (approved):

pettinarip (approved):
🎉

PR #11748, opened 2023-12-01
Removing /discord/ from redirects

fredriksvantes:
## Description

/discord/ redirects to https://discord.com/invite/rZz26QWfCg which is invalid.


fredriksvantes:
Making another PR changing to correct URL instead

PR #11751, opened 2023-12-01
Update community-meetups.json

duckdegen:
Adding CryptoCanal events for Amsterdam, Rotterdam and Zurich

## Description

Added three cities with CryptoCanal meetups: Amsterdam, Rotterdam and Zurich


src/data/community-meetups.json
```diff
@@ -1,4 +1,22 @@
...
+    "location": "Amsterdam",
+    "link": "https://www.cryptocanal.org/agenda"
+  },
+  {
+    "title": "CryptoCanal",
+    "emoji": ":ch:",
+    "location": "Zurich",
+    "link": "https://www.cryptocanal.org/agenda"
+  },
+  {
```
wslyvh, line 14: the Netherlands is too small to add the same meetup for multiple cities imo 
Maybe just update location to 'the Netherlands' instead?

duckdegen: Sure, can do!

wslyvh: I think its fine to add CryptoCanal as 2 location for the Netherlands and Zurich, Swiss.
That makes sense. Someone browsing through the list is looking for meetups in his city/country.

But no need to differentiate between Amsterdam/Rotterdam imo.

corwintines (approved):
Thanks @duckdegen!

PR #11772, opened 2023-12-07
fix H4 sizing on /staking page: [Fixes #11409]

### LINKED ISSUE #11409: H4 sizing on /staking page
### Describe the bug

H4's on /staking pages appear to have grown in size to 36px from 20px.

### To reproduce

1. Go to https://ethereum.org/en/staking/#comparison-of-options
2. See size of H4's

### Expected behavior

This is what is was previously like (and would expect):

image

### Screenshots

Currently:

image

_No response_

_No response_

### Additional context

@nloureiro or @konopkja if you could confirm this

- [ ] Yes
- [X] No

shyam-patel-kira:
fixes the bug causing H4 sizing issue on /staking page

## Description

added `size` field to `<OldHeading>` component to circumvent the issue

## Related Issue

fixes #11409


nloureiro (approved):

shyam-patel-kira:
fixed the merge conflicts

nhsz:
I think this should be solved at the theme level and not individually per heading cc @pettinarip

TylerAPfledderer:
> I think this should be solved at the theme level and not individually per heading cc @pettinarip @TylerAPfledderer 

There is currently not a setup matching a heading `size` to a specific heading level by default. Can look into setting up these defaults in a separate PR.

At the very least, might it be better to switch to the current `Heading` component here instead of sticking with the old version?

nhsz:
> > I think this should be solved at the theme level and not individually per heading cc @pettinarip @TylerAPfledderer
> 
> There is currently not a setup matching a heading `size` to a specific heading level by default. Can look into setting up these defaults in a separate PR.
> 
> At the very least, might it be better to switch to the current `Heading` component here instead of sticking with the old version?

Switching to the current `Heading` makes sense to me (cc @shyam-patel-kira), any reason you're aware why we're still using the `OldHeading` in some components?

TylerAPfledderer:
> > > I think this should be solved at the theme level and not individually per heading cc @pettinarip @TylerAPfledderer
> > 
> > There is currently not a setup matching a heading `size` to a specific heading level by default. Can look into setting up these defaults in a separate PR.
> > 
> > At the very least, might it be better to switch to the current `Heading` component here instead of sticking with the old version?
> 
> Switching to the current `Heading` makes sense to me (cc @shyam-patel-kira), any reason you're aware why we're still using the `OldHeading` in some components?

It references styles from the old theme that is not really consistent with the new DS. 

So if there is a switch to the current `Heading`, then you need to double-check to make sure it doesn't break the content layout. (I doubt it with an `h4` heading level)

shyam-patel-kira:
> > > I think this should be solved at the theme level and not individually per heading cc @pettinarip @TylerAPfledderer
> > 
> > 
> > There is currently not a setup matching a heading `size` to a specific heading level by default. Can look into setting up these defaults in a separate PR.
> > At the very least, might it be better to switch to the current `Heading` component here instead of sticking with the old version?
> 
> Switching to the current `Heading` makes sense to me (cc @shyam-patel-kira), any reason you're aware why we're still using the `OldHeading` in some components?

It was reintroduced in this commit https://github.com/ethereum/ethereum-org-website/commit/7f26b77083b73a34a5d5d28fa8df5ce3f3342aa2, since some components didn't work correctly with the new migrations

nhsz (approved):

Issue #11782, opened 2023-12-10
Bug: [ja/MEV] Incorrect Calculation Formula for Gas Fees


nakmr:
### Describe the bug

The calculation formula for "Gas Fee" mentioned in the Japanese translated article "MAXIMAL EXTRACTABLE VALUE (MEV)" is incorrect as follows:

Incorrect
gas fees = gas price **÷** gas used
Correct
gas fees = gas price **×** gas used

### To reproduce

1. Go to https://ethereum.org/ja/developers/docs/mev/
2. Scroll down to [ガス・ゴルフ](https://ethereum.org/ja/developers/docs/mev/#mev-extraction-gas-golfing)
4. See error as follows:
> ガス・ゴルフにより、サーチャーはより高価なガス代を設定しつつ、ガス代の合計額を一定に抑えられるためです（ガス代＝ガス価格 ÷ 使用ガス量）。

### Expected behavior

The calculation formula for gas fees should be 'ガス価格 × 使用ガス量' (which means 'gas price × gas used').

### Screenshots

![ja-MEV-gas-fees-calculation](https://github.com/ethereum/ethereum-org-website/assets/145886319/41a65ad7-6f79-4bb0-aece-70e6457ce6dc)

### Would you like to work on this issue?

- [X] Yes
- [ ] No

HiroyukiNaito:
It is a mistranslation.
I raised issue in the Crowdin.
https://crowdin.com/editor/ethereum-org/5525/en-ja?view=comfortable&filter=basic&value=0#q=%E3%82%AC%E3%82%B9%E3%83%BB%E3%82%B4%E3%83%AB%E3%83%95

HiroyukiNaito:
I corrected the mistranslation in the Crowdin, so it will be fixed in the next release of the content.

PR #11783, opened 2023-12-10
Fix calculation formula for gas fees in ja/MEV [#11782]

nakmr:
Corrected the incorrect gas fees calculation formula in the Japanese-translated "MAXIMAL EXTRACTABLE VALUE (MEV)" article, changing it from 'gas price **÷** gas used' to 'gas price **×** gas used'.

## Description

Corrected the calculation formula for "gas fees" mentioned in the article "MAXIMAL EXTRACTABLE VALUE (MEV)" that was translated into Japanese, as it was incorrect as follows:

```
Incorrect
gas fees = gas price ÷ gas used
Correct
gas fees = gas price × gas used
```


corwintines (approved):
Thanks @nakmr

corwintines:
@all-contributors can you add @nakmr for content

PR #11792, opened 2023-12-13
Adding Cyfrin offerings

juliettech13:
## Description

Added Cyfrin to the article's list.

Cyfrin is a smart contract security firm, offering services like private audits, competitive audits, education platform, among others.


wackerow (approved):
Thanks again @juliettech13!

PR #11795, opened 2023-12-13
Fix typo

bskrksyp9:
## Description

Corrected spelling


minimalsm (commented):
Both forms are correct, but "rebroadcasted" is more commonly used, especially in American English. 

The hyphenated form "re-broadcasted" could be used, but it is not necessary...

I think we should just keep it as is to avoid bike-shedding on it

PR #11805, opened 2023-12-14
Add examples of blockchain based persistence

cryptoryda:


minimalsm:
Thanks for the PR @cryptoryda. I think we should keep examples focused on the Ethereum ecosystem here, as if we go too wide the list will be endless. Closing this PR out as a result of that 👌

PR #11862, opened 2024-01-03
Fix country flag placement in meetup list [Fixes #11861]

### LINKED ISSUE #11861: Fix country flag placement in Ethereum meetups list
### Describe the bug

While browsing the website, I noticed the placement of the country flags in the meetup list was a little off. 
The country flags are showing higher than the country names (not sure if this was done intentionally, but assuming it wasn't). (see first screenshot)

I did a quick test in devtools and think it looks better when the flags and country names are aligned (see second screenshot). I did this by removing the line-height from the country flag's (emoji) component. 

### To reproduce

1. Go to http://localhost:8000/en/community/events/.
2. Scroll down till you see the Ethereum meetups list.
3. See 'strange' country flag placement. 

### Expected behavior

Country flag and country name should be aligned. 

## Screenshot 1: 

Screenshot 2024-01-04 at 00 11 57

## Screenshot 2: 

Screenshot 2024-01-04 at 00 12 33

_No response_

_No response_

### Additional context

I will open a PR for this. 

- [X] Yes
- [ ] No

hursittarcan:
## Description
- Fixes the strange placement of the country flags -> described in issue #11861. 

- I fixed this by removing the line-height from the emoji component (used to display the country flags). I wasn't sure about removing the line-height entirely because the component is being used on many different places too, soo I added an extra prop to check for flags.  

## Related Issue
- Related to #11861


nloureiro:
@corwintines this fix looks good, it fixes the [issue](https://github.com/ethereum/ethereum-org-website/issues/11861) in questions and make the list look better.

src/components/MeetupList.tsx
```diff
@@ -128,7 +128,7 @@ const MeetupList: React.FC<IProps> = () => {
               me={4}
               flexWrap="wrap"
             >
-              <Emoji text={meetup.emoji} boxSize={4} me={2} />
+              <Emoji text={meetup.emoji} boxSize={4} me={2} isFlag />
```
wackerow, line 131:
```suggestion
              <Emoji text={meetup.emoji} boxSize={4} me={2} lineHeight="unset" />
```

wackerow (changes requested):
Hey @hursittarcan! Thanks for dropping this fix... This does fix the issue, but I would push back on altering the Emoji component here for this one-off use case.

The Emoji component spreads `{...props}` into the component, so instead of making a whole new boolean prop here, we can just pass `lineHeight` when declaring an Emoji to override this prop specifically.

I would revert everything changed in `Emoji.tsx`, and instead just change `isFlag` to `lineHeight="unset"` inside it's usage within `MeetupList`. `lineHeight` will be passed to the `Emoji` component accomplishing the same thing you're doing here, but without altering the component itself.

hursittarcan:
Hey @wackerow 👋
Thanks for the review! 
Did the requested changes 👍 
And wow, your solution was way cleaner than mine, didn't realise it could be this simple. 😅

wackerow (approved):
> And wow, your solution was way cleaner than mine, didn't realise it could be this simple. 😅

Hah, glad I could help =) Thanks again!

corwintines (approved):

PR #11870, opened 2024-01-04
Update Block Explorer List to include Ethernow .md

aseoconnor:
Adding Ethernow.xyz to the list of explorers available to the ecosystem.

## Description

Adding in Ethernow.xyz as a block explorer


corwintines (commented):
Thanks @aseoconnor

PR #11872, opened 2024-01-05
Specify that you can get one GitPOAP for each year

sandakersmann:
Specify that you can get one GitPOAP for each year in README.md


corwintines (approved):
Thanks @sandakersmann

PR #11892, opened 2024-01-09
Add 2x multiplier banner for Dencun on bug bounty page

fredriksvantes:
## Description

Add 2x multiplier banner for Dencun on bug bounty page

(should be active from now up until two weeks prior to when the mainnet hardfork is scheduled to be at (currently unknown))


fredriksvantes:
Appears that the build failed so perhaps something is amiss.. but I would like to add the following to a banner at the top of the bug bounty page:

All Dencun-related bounties currently receive a 2x bonus multiplier
(up to 500,000 USD) until two weeks prior to the scheduled mainnet hardfork.

corwintines:
@fredriksvantes ill take a look whats going on here. We just finished a migration from Gatsby to Next so might have been an artifact of Netlify being setup for NextJS, but the was based off of Gatsby version.

corwintines (approved):
LGTM

PR #11903, opened 2024-01-10
Refactor Show/Hide component usage in layouts

wackerow:
**⚠️ Note: Builds on Next.js -> Do not merge until Next.js fork has been fully merged**

## Description
Usage of Chakra-UI `Show` and `Hide` components causing layout shifting when a page loads, since these components are built on the server where the width of the display is not yet known.

- Updates usage within all "layout" components to use `hideBelow` and/or `hideFrom`, removing usage of `Hide` and `Show`
- Added these directly to the shared components as their usage for mobile vs desktop does not change


corwintines (approved):
Awesome 🚀

PR #11905, opened 2024-01-11
LTR/RTL fixes

wackerow:
## Description
- Use LTR direction for content portion of pages without translated content available
  - (Not implemented on all pages yet. Started with applying to _layouts_, and _tutorials landing_ page)
- Force child components that use "common" namespace (always available for a locale) to use native text direction, overriding any upstream LTR declarations if page content is untranslated
- Fix search modal direction-responsive css styling
- Fix direction-responsive absolute positioning for some components

## Preview URL
- https://deploy-preview-11905--ethereumorg.netlify.app/

## Screenshots:
https://deploy-preview-11905--ethereumorg.netlify.app/ar/developers/docs
ex. RTL locale, but content not translated:
image

https://deploy-preview-11905--ethereumorg.netlify.app/fa/developers/docs
ex. RTL locale where content is translated (unchanged, but for comparison))
image

RTL search modal:
image


wackerow:
(Initial Netlify build failed from unrelated reasons which caused the check to fail... re-ran and it passed. Preview link from passing build is above, but the check never reset)

src/components/Hero/HubHero/index.tsx
```diff
@@ -37,7 +53,7 @@ const HubHero = (props: CommonHeroProps) => {
         borderRadius={{ xl: "base" }}
         bg={{ xl: "hubHeroContentBg" }}
         position={{ xl: "absolute" }}
-        insetStart={{ xl: "8" }}
+        style={{ insetInlineStart }}
```
wackerow, line 56: This change was made throughout as the style props `insetStart/End` or `insetInlineStart/End` have not been working as expected with Chakra components. Using `style` fixes this, but then we lose access to the Chakra tokens. As an alternative to using css `var(--eth-type-token)` syntax, the `useToken` hook was used.

src/components/Search/utils.ts
```diff
@@ -149,7 +151,7 @@ export const getSearchModalStyles = (): SystemStyleObject => ({
   },
 
   ".DocSearch-Hit-Select-Icon:focus, .DocSearch-Hit-Select-Icon:hover": {
-    color: "switchBackground",
+    color: "switchBackground", // TODO: Remove? Causing low contrast in dark mode
```
wackerow, line 154: image

@nloureiro When hovering directly over the arrow we get this low-contrast fill on dark mode:

image

It's not as bad in light mode image but is there a color token you would recommend that looks good in both?

Or maybe we just consider removing this hover effect since it only triggers when hovering over the arrow itself, but the action doesn't change compared to the rest of the box. This would be my preference.

src/components/Search/utils.ts
wackerow, line 1: These file changes affect the search modal

image

src/components/Search/utils.ts
```diff
@@ -91,6 +91,8 @@ export const getSearchModalStyles = (): SystemStyleObject => ({
     },
     input: {
       fontSize: { base: "md", md: "xl" },
+      p: 0,
+      ps: 2,
```
wackerow, line 95: `p: 0` overrides existing `padding-left` from `@docsearch/css`, then reassigns with direction-responsive `ps`

src/components/Search/utils.ts
```diff
@@ -184,17 +186,24 @@ export const getSearchModalStyles = (): SystemStyleObject => ({
...
     placeItems: "center",
     borderRadius: "base",
+    mr: "unset",
+    me: 1.5
   },
 
   ".DocSearch-Logo": {
     ".DocSearch-Label": {
-      color: "body.light",
+      color: "body.medium",
```
wackerow, line 195: Sorry to make design adjustments in this PR, but this was producing an unreadable contrast ratio:
before:
image

with change:
image

cc: @nloureiro if you have any other preference for this, but figured it was worth patching

src/components/TranslationBanner.tsx
```diff
@@ -34,14 +48,17 @@ const TranslationBanner: React.FC<IProps> = ({
...
   return (
     <Box
       as="aside"
       display={isOpen ? "block" : "none"}
       bottom={{ base: 0, md: 8 }}
-      insetInlineEnd={{ base: 0, md: 8 }}
+      style={{ insetInlineEnd }}
       position="fixed"
-      zIndex="99"
+      zIndex="banner"
```
wackerow, line 60: Fixes bug where menu contents (ie. UseCases pages) were overlaying this banner.

src/pages/_app.tsx
wackerow, line 1: These changes enable use of `inset*` style props for Chakra-UI components.

h/t @TylerAPfledderer

Went through and reverted the switch to using `style={{}}` for these

src/components/Breadcrumbs/index.tsx
```diff
@@ -65,7 +70,7 @@ const Breadcrumbs = ({
     .slice(startDepth)
 
   return (
-    <Breadcrumb {...props}>
+    <Breadcrumb {...props} dir={dir}>
```
TylerAPfledderer, line 73: Most instances of the `dir` prop should now be removed as this is redundant and is taken care of under the hood in the Chakra component structure/theme or by the use of an `inset*` prop.

An exception is where flexbox is used for content positioning, where CSS is not smart enough to handle RTL automatically with flex props. (I will point out the exceptions)

src/components/CodeModal.tsx
```diff
@@ -25,12 +25,7 @@ const CodeModal = ({ children, isOpen, setIsOpen, title }: CodeModalProps) => (
     <ModalOverlay />
     <ModalContent>
       <ModalHeader>{title}</ModalHeader>
-      <ModalCloseButton
-        style={{
-          right: "unset",
-          insetInlineEnd: "var(--eth-sizes-4)",
-        }}
-      />
+      <ModalCloseButton right="unset" insetInlineEnd="4" />
```
TylerAPfledderer, line 28: Hmm this button's positioning looks to be working as expected on my end without these to props. And `insetInlineEnd` is already set the the Modal component theme.

src/components/FeedbackCard.tsx
```diff
@@ -67,6 +71,7 @@ const FeedbackCard = ({ prompt, isArticle, ...props }: FeedbackCardProps) => {
       mt="8"
       w="full"
       {...props}
+      dir={dir}
```
TylerAPfledderer, line 74: This is an exception where the `dir` prop should be kept, as flexbox is controlling the content positioning here.

src/components/Trilemma/index.tsx
```diff
@@ -81,7 +81,7 @@ const Trilemma = () => {
               justifyContent="flex-start"
               my={8}
             />
-            <DrawerCloseButton top={6} insetInlineEnd={6} />
+            <DrawerCloseButton top={6} right="unset" insetInlineEnd="6" />
```
TylerAPfledderer, line 84: `right="unset"` does nothing here

```suggestion
            <DrawerCloseButton top={6} insetInlineEnd="6" />
```

src/layouts/Docs.tsx
```diff
@@ -241,7 +249,7 @@ export const DocsLayout = ({
           <Translation id="page-developers-docs:banner-page-incomplete" />
         </BannerNotification>
       )}
-      <ContentContainer>
+      <ContentContainer dir={contentNotTranslated ? "ltr" : "unset"}>
```
TylerAPfledderer, line 252: This is a primary situation where `dir` should be used to override existing logic. 👍🏼

src/lib/types.ts
```diff
@@ -231,6 +231,7 @@ export type CommonHeroProps = {
   title: string
   description: string
   buttons?: [CallToActionProps, CallToActionProps?]
+  dir?: "rtl" | "ltr"
```
TylerAPfledderer, line 234: I think this is also no longer needed

TylerAPfledderer (changes requested):
My thoughts regarding RTL support here 😄

src/hooks/useLocaleDirection.ts
```diff
@@ -9,11 +9,15 @@ import { isLangRightToLeft } from "@/lib/utils/translations"
...
  * responding to changes in the locale without requiring refresh.
  */
-export const useLocaleDirection = (): void => {
+export const useLocaleDirection = (): string => {
+  const [direction, setDirection] = useState<"ltr" | "rtl">("ltr")
   const { locale } = useRouter()
 
   useEffect(() => {
     const dir = isLangRightToLeft(locale as Lang) ? "rtl" : "ltr"
     document.documentElement.setAttribute("dir", dir)
```
wackerow, line 18: @pettinarip Fyi, despite passing the `direction` to the Chakra theme provider, I'm still requiring `document.documentElement.setAttribute("dir", dir)` to automatically flip the direction when the locale is changed. Without this, switching from English to Farsi would require a refresh to update the DOM direction:

https://github.com/ethereum/ethereum-org-website/assets/54227730/88e1e662-b8d4-4ab2-8ded-9242da513a31

I too would prefer to remove the use of `document` here, but going to keep it for now until we find a solution that automatically updates this on locale change without a refresh.

src/components/FeedbackCard.tsx
wackerow: Honestly I'm not having any problems with flex boxes here...

I'll double check all of your comments, but the reason for these was not _just_ to use the `dir` for the chosen locale, but to specifically do so at the component level for components that:

1. Exclusively use the "common" namespace for translations (which _every_ language has before even being listed
2. Are (or potentially) nested inside _page content_ which may end up NOT being translated.

In this case (ie. [/ar/developers/docs/blocks](https://ethereum.org/ar/developers/docs/blocks)) `html` could be set to `dir="rtl"`, the _main content div_ could be set with `dir="ltr"` (overriden because `contentNotTranslated=true`), but then some components (ie. `FeedbackCard`) are fully translated, which should again run **rtl**. 

The change here addressed this by independently setting the `dir` based on locale, without relying on `html[dir]` since the content it's nested in may flip it back.

src/components/CodeModal.tsx
wackerow: Okay, this was set with `style` in a previous PR, so I had updated to use style props instead, but I think changes since have corrected things so it's not necessary anymore. Good catch, will remove

src/components/Breadcrumbs/index.tsx
wackerow: See other comment about specifically why I'm overriding these at the component level... same applied to this one.

src/components/Trilemma/index.tsx
wackerow: Yep, good catch.. this was added before we fixed the `inset*` issue which seems to have made these close buttons work more as-expected in RTL

src/lib/types.ts
wackerow: Hm... I had added this to HubHero props, but I'm not actually using it yet... I believe this was in preparation for updating the other page directions using `contentNotTranslated` logic, but that will be in a separate PR. Will revert this and the HubHero prop changes for now.

src/hooks/useLocaleDirection.ts
pettinarip: Got it! no worries, thanks for investigating.

src/components/FeedbackCard.tsx
pettinarip: Got it. Agreed, I think we need this for these "global" components.

pettinarip (approved):
@wackerow amazing job!!!! thanks for the work done here. Looks really nice the RTL version!

==========
Few minor things that **don't block** this PR.

Small detail in the slider arrows, we shouldn't use rtl props here.
https://deploy-preview-11905--ethereumorg.netlify.app/fa/what-is-ethereum
image

Not aligned
image

The results from the search are not translated, right? if so, I think we shouldn't change the direction
image

wackerow:
@pettinarip 

> Small detail in the slider arrows, we shouldn't use rtl props here.

We should, but we need to pass `{ direction: "rtl" | "ltr" }` to the `useEmblaCarousel({ direction })`, and flip the icons with `flipForRtl` (`scaleX(-1)`)

https://github.com/ethereum/ethereum-org-website/assets/54227730/bccf3bf0-d0b9-4a1c-b570-e12fdfdbef5d

Will put a patch up for this in a separate PR.

---

> The results from the search are not translated, right? if so, I think we shouldn't change the direction

Yeah, they should be translated (if available):

image

image

...but, these languages aren't translated 100%, so we're still going to end up with English results for that content (ie. Algolia parses /ar/whitepaper into the `ar` lang facet... if it has English content, this is included in the Arabic search results)

Can discuss how we'd like to approach this.

---

The Trophy icon I'll look into; also separate to this PR. Thanks for spotting!

pettinarip:
Nice, thanks @wackerow

PR #10997, opened 2023-08-24
refactor(FeedbackWidget): Improve structure with Chakra UI

TylerAPfledderer:
## Description
This PR updates the `FeedbackWidget` to improve its usage with Chakra UI.
- Codesplits the "Dot" button into a separate component file.
- Abstracts away the logic to a useHook.
- Uses Chakra's `Modal` component for the structure
- Create a story for visual testing

## Related Issue
- Fixes #12184

## Summary by CodeRabbit

- **New Features**
	- Introduced a `FeedbackWidget` for collecting user feedback directly within the app.
	- Added a `FixedDot` component for easy access to the feedback widget, with dynamic styling and hover effects.
	- Implemented a custom hook `useFeedbackWidget` to manage the feedback widget's state and interactions.
	- Set up a Storybook story for the `FeedbackWidget`, showcasing its features and styling.

PR #11906, opened 2024-01-11
Refactor FeedbackWidget with Chakra-UI

wackerow:
## Description
Updates FeedbackWidget component, refactored using Chakra-UI AlertDialog
- Removes need for `focus-trap-react` package previously being used
- Removes need for `useKeyPress` hook previously being used
- Removed need for `useOnClickOutside` hook previously being used


src/components/FeedbackCard.tsx
corwintines, line 1: I think we should keep what was here before. We have a translated link in here, and would need to go through all the translated files to fix this if we want to make this change. But, the link changes where it is in the sentance depending on the language so might be hard to keep this change and not have downstream effects with translated languages. 

Screenshot 2024-01-16 at 1 39 08 PM

wackerow: @corwintines Oh good catch, thanks... just pushed a patch to use `<Translation id="" />` component for that string instead to use its `htmr` support... should be good now

TylerAPfledderer:
@wackerow should close #10997 in favor of this, correct?

wackerow:
@TylerAPfledderer Curious your thoughts if you see anything from that PR we should incorporate here, but I think its a bit cleaner utilizing the AlertDialog here

corwintines (approved):

TylerAPfledderer:
> @TylerAPfledderer Curious your thoughts if you see anything from that PR we should incorporate here, but I think its a bit cleaner utilizing the AlertDialog here

@wackerow I intended to go ahead and add a story for it.

I would note that `AlertDialog` adds `role='alertdialog'` to the modal. By MDN definition for screen readers: 

_The alertdialog role is used to notify users of urgent information that demands the user's immediate attention. Including role="alertdialog" on the element containing the dialog helps assistive technology identify the content as being grouped and separated from the rest of the page content. Examples include error messages that require confirmation and other action confirmation prompts._

When testing with NVDA, it doesn't seem to have that kind of effect here. However, unlike the current version, a screen reader might readout everything in the modal immediately.

Here are the following set of prompts NVDA gives as soon as you click the button to open the modal:
```
Is this page helpful? dialog

clickable Close button

Is this page helpful?

button Close

Is this page helpful?

button Yes
button No
```

Would be better to get proper user feedback, but this looks to be an improvement.

wackerow:
@konopkja Btw, this PR fixes the bug you noted about "Closed feedback widget" events being triggered inappropriately.

PR #11933, opened 2024-01-16
feat(Quiz): create QuizManager component

TylerAPfledderer:
_Originally in the NextJS forked repo as https://github.com/ethereum/ethereum-org-fork/pull/257_

## Description

Creates a reusable `QuizManager` component, containing `QuizModal`, `QuizWidget`, `QuizzesList`, and `QuizzesStats`.

Used for providing separate quiz sets for multiple pages outside of the Quiz Hub.


src/components/Quiz/QuizManager/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+import { useLocalQuizData } from "../useLocalQuizData"
+
+const handleGHAdd = () =>
+  trackCustomEvent({
+    eventCategory: "quiz_hub_events",
+    eventAction: "Secondary button clicks",
+    eventName: "GH_add",
+  })
+
+type QuizPageKeys = "quizzes-stats"
```
TylerAPfledderer, line 28: Proposal: Have a single source of truth (as a string union type) for storage key names for each page the uses `QuizManager` as a way to help prevent unexpected changes that result in typos of the keys. (Avoid negative impact on users with existing stats)

Definitely a matter of preference here if this is going to help.

Also need to consider a way to replace `quizzes-stats` with a key name more unique to the hub page so it stands out from key names supplied for other pages using `QuizManager`.

TylerAPfledderer:
Putting back to draft to manually bring in and review logic changes from prod

wackerow:
Hey @TylerAPfledderer, just circling through here... do we know the plan with this one?

TylerAPfledderer:
@wackerow if I recall, there was a discussion I had with you bringing a point of contention in the actual usefulness of this reusability. I don't remember if there was supposed to be further discussion before I kept going with this.

TylerAPfledderer:
@wackerow 

This update was suggested by @pettinarip . I believe at the time the intention was reusability in the event of another quiz hub-like page was made with quiz results to store separately.

I'm not sure about its usefulness at this point; I think I was running into a couple of issues or roadblocks and needed more information about the approach around them.

pettinarip:
Correct. We wanted to refactor the code to make it reusable because we had a small idea to create another hub back then.

I still think it's useful but we're focused on other priorities right now, so we probably won't move forward with this anytime soon.

Let's close this PR for now and reopen it when we really need it.

Thanks for your work on this @TylerAPfledderer and sorry for the delay in feedback.

PR #11934, opened 2024-01-16
refactor: extend eslint

### LINKED ISSUE #11939: [Proposal] Add additional ESLint configurations
To further improve the linting experience and to ensure there are no missing fixes, consider improving upon the ESLint configuration

## 1. Set `no-unused-vars` rule
This is the most noticeable and simplest addition to the config. With the current set-up (and when `eslint-config-prettier` gets applied), this rule gets missed.

Would need to add the following rule:
```json
"no-unused-vars": [
  "warn",
  /**
    * Per a preferred convention used, we use an underscore
    * or variable name prefixed with an underscore to denote an
    * argument not being used in a callback function.
    * This rule ignores all args that begin with an underscore.
    *
    * i.e. `arrObj.map((_, index) => index + 1)`
    * The first argument of `value` is not used.
    */
  { "argsIgnorePattern": "^_" }
],
```

This is not the most ideal solution, as this will throw warnings for function args defined in a type signature. They would not have a name with an underscore.

## 2. Include linting plugins for TypeScript

This would be a more encompassing solution to No. 1 and be able to handle other things specific to typescript. The most common option is the [typescript-eslint packages](https://typescript-eslint.io) to ensure proper configuration with TypeScript, and this package can handle the unused vars issue with types.

## Other thoughts

The overall inquiry from this is how far should this project need to go to ensure code consistency. NextJS configs should handle most use cases for best practice, and the above are the additions not included in it's [default recommended configuration](https://nextjs.org/docs/pages/building-your-application/configuring/eslint#eslint-plugin). The default is currently not being applied because of initial setup of the fork using only `next/core-web-vitals`, but should be flipped to extending `next` in the future when ready.

TylerAPfledderer:
_Originally in the NextJS forked repo as https://github.com/ethereum/ethereum-org-fork/pull/267_

## Description

This PR introduces the following packages for the Eslint configuration:
- `@typescript-eslint/eslint-plugin`: extends a recommended set of rules for Typescript and supersedes some existing eslint rules for improved checks.
- `@typescript-eslint/parser`: required parser for all sourced files
- `eslint-plugin-unused-imports`: plugin to allow for auto fixing unused imports

See [typescript-eslint docs](https://typescript-eslint.io/linting/configs#recommended) to view the recommended rules configuration and other possible configurations available.

See the [eslint-plugin-unused-imports repo](https://github.com/sweepline/eslint-plugin-unused-imports) for an overview of this plugin.

In addition, add the `next` configuration along with `next/core-web-vitals` to ensure the framework's recommended rules are being accounted for.

### Additional Information
In the eslint config, extending `@typescript-eslint:recommended-type-checked` instead of `*:recommended` catches unsafe types, unsafe type overrides, unnecessary type assertions, etc. that can bleed into the project. i.e. `JSON.parse()` will always return `any`, and adding a type assertion to it would be beneficial if using the resulting object later.

> This will add a lot of errors and warnings about `any` in this project!

## Related Issue
- Closes #11939


TylerAPfledderer:
Throwing back into draft to address type errors caught after running `yarn tsc`.

src/pages/contributing/translation-program/contributors.tsx
```diff
@@ -66,6 +62,7 @@ const Contributors = () => {
 
   // TODO: Remove specific user checks once Acolad has updated their usernames
   const translatorData =
+    // @ts-expect-error Not able to manually type `allTimeData` (too big). Need to generate a type signature.
```
TylerAPfledderer, line 65: @pettinarip This is failing Netlify deploy, where Typescript is assuming there is no error when there actually is.

To get rid of this and to not error on an `any` type, I would try and create a type for the `allTimeData` json because an inference is not available. There is a generator available to copy/paste the json to spit out a type set. Then can apply it here.

TylerAPfledderer: See #12037

pettinarip: Merged!

.eslintrc.json
```diff
@@ -17,30 +26,67 @@
...
+    "simple-import-sort/exports": "error",
+    "no-unused-vars": "off",
+    "@typescript-eslint/no-unused-vars": "off",
+    "@typescript-eslint/no-explicit-any": [
+      "error",
+      {
+        "fixToUnknown": true
+      }
+    ],
+    "unused-imports/no-unused-imports-ts": 2
```
pettinarip, line 90: Same here, what does this one mean?

pettinarip (commented):
Looks good @TylerAPfledderer! I like the addition of `@typescript-eslint`.

.github/workflows/quality.yml
```diff
@@ -0,0 +1,32 @@
...
+
+      - name: Install deps
+        run: yarn
+
+      - name: Run ESLint Fix
+        run: yarn lint:fix
+
+      - uses: stefanzweifel/git-auto-commit-action@v5
+        with:
+          commit_message: "chore: run lint:fix"
```
pettinarip, line 32: I wasn't expecting a GH job for this but I think it is fine and covers the goal pretty well. How do you see adding prettier to it? since my idea was to create a precommit/prepush hook to deal with these two.

.eslintrc.json
```diff
@@ -17,30 +26,67 @@
...
+          [
+            "^\\u0000"
+          ]
         ]
       }
     ],
-    "simple-import-sort/exports": "error"
+    "simple-import-sort/exports": "error",
+    "no-unused-vars": "off",
+    "@typescript-eslint/no-unused-vars": "off",
```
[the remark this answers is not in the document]
TylerAPfledderer: `@typescript-eslint/no-unused-vars` rule overrides `unused-imports/no-unused-vars` where the former will throw an error for an unused param, usually denoted by an underscore.

So for the former, `Array.from({length: 5}).map((_, idx) => {})` will throw an error for the underscore, whereas `unused-imports/no-unused-vars` will know that this is a param ignored intentionally, but still throw a warning or error for declared variables not being used.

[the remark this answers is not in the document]
TylerAPfledderer: This is recommended in the [unused-imports eslint docs](https://github.com/sweepline/eslint-plugin-unused-imports?tab=readme-ov-file#usage)

.github/workflows/quality.yml
TylerAPfledderer: We can certainly do a pre-commit instead. You could add a format script that should essentially do the same type of job run as `lint:fix`.

But if you prefer a pre-commit, I can remove this. I figure I would at least make sure I presented this as an option. :D

What was suggested at this point in doing the pre-commit/push hook?

.eslintrc.json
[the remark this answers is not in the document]
pettinarip: Ah got it! thanks

.eslintrc.json
pettinarip: Yeah, a warning is ok. As long as it is fixed by the `lint:fix` it is the same to me.

.eslintrc.json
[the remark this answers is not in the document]
pettinarip: Understood your explanation Tyler but I don't see that we are declaring the rule for `unused-imports/no-unused-vars`. I think we need to add that since I don't get any error for unused vars currently.

.github/workflows/quality.yml
pettinarip: Ok cool, that is fine. Lets keep this job and see how it work for us.

I think this might have an advantage over the precommit hook: it will always catch & fix linting errors, whether the user checks them or not.

.eslintrc.json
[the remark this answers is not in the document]
TylerAPfledderer: @pettinarip looking at this again, not sure what is up with `unused-imports/no-unused-vars` but can do this instead:
```json
"@typescript-eslint/no-unused-vars": [
  "error", {
    "varsIgnorePattern": "^_"
  }
],
```
To still make sure instance of underscore params are not flagged. This will also not flag any variable that begins with `_`, which will include components.

Should this be an `error` or `warn`?

.github/workflows/quality.yml
TylerAPfledderer: Sounds good!

src/components/Staking/StakingProductsCardGrid/StakingProductCard.tsx
```diff
@@ -29,6 +29,7 @@ const getIconFromName = (
 ): ComponentType<SVGProps<SVGElement>> => {
   const {
     [imageName + "GlyphIcon"]: Icon,
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
```
TylerAPfledderer, line 32: As far as eslint is concerned, [using `require` in  pretty much banned](https://typescript-eslint.io/rules/no-var-requires/).

Possible workaround here is to run a call in `useEffect` and storing the import in state. This would use the more modern `import()` to handle dynamic importing, which returns a promise.

```tsx
const [Svg, setSvg] = useState<ChakraIconType | null>(null)

useEffect(() => {
  const getIconFromName = async (imageName: string) => {
    const Icon: ChakraIconType = await import(
      "@/components/icons/staking"
    ).then((icons) => icons[imageName + "GlyphIcon"])
    setSvg(Icon)
  }

  getIconFromName(imageName)
})
```

Caveat here is now that there is essentially a fetch occurring, there is a delay in the render of the icons.

.github/workflows/quality.yml
```diff
@@ -0,0 +1,32 @@
...
+on:
+  push:
+    branches:
+      - dev
+  pull_request:
+    branches:
+      - dev
+
+jobs:
+  eslint:
```
TylerAPfledderer, line 12: @pettinarip realizing that there might be some trouble with this job when triggered on a pull request.

Because of Github security limitations with forks, it might not be possible to run a commit from a file change done by this job.

See the `git-auto-commit-action` [README section regarding public repos with forks](https://github.com/stefanzweifel/git-auto-commit-action?tab=readme-ov-file#use-in-forks-from-public-repositories)

There should be no issue running this job on push to the base branch (`dev`).

So with that said, I'm thinking of two parts here:

1. Job that just runs on pull request to simply execute `yarn lint` and returns any errors/warnings
2. Job that runs on push to base branch and executes `yarn lint:fix` to auto fix any warnings or errors that can be done and then make a commit to push separately.

Though it's starting to sound like running a pre-commit is easier here! 😅 

Is there something I am missing that you might know about?

.eslintrc.json
```diff
@@ -1,11 +1,20 @@
 {
   "extends": [
+    "eslint:recommended",
```
TylerAPfledderer, line 3: Updated this configuration. Each line does the following:
- `eslint:recommended`: follows the recommended rules [as defined in the eslint docs with the ✅ ](https://eslint.org/docs/latest/rules/)
- `plugin:@typescript-eslint/recommended`: follows the recommended rules [as defined in this rule config](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/recommended.ts) and defined in the [general rules table](https://typescript-eslint.io/rules/?=recommended)
- `next/core-web-vitals`: includes NextJS base lint config plus two rules [as defined within this rules config](https://github.com/vercel/next.js/blob/3f907d38cbb3a97d5201d4add5757d64453756c7/packages/eslint-plugin-next/src/index.ts#L57)
- `prettier`: have prettier rules override conflicting eslint rules
- `plugin:storybook/recommended`: lint rules specific to storybook files

src/scripts/markdownChecker.ts
```diff
@@ -2,12 +2,13 @@ import fs from "fs"
...
 
 import matter from "gray-matter"
+import minimist from "minimist"
 
 import type { Lang } from "../lib/types"
 
 type Summary = Record<string, string[]>
 
-const argv = require("minimist")(process.argv.slice(2))
+const argv = minimist(process.argv.slice(2))
```
TylerAPfledderer, line 11: If the `require` statement was left, there would be the following error:
```bash
 Error: Require statement not part of import statement.  @typescript-eslint/no-var-requires
```

src/components/UpcomingEventsList.tsx
```diff
@@ -75,6 +75,7 @@ const UpcomingEventsList = () => {
     })
 
     setOrderedUpcomingEvents(formattedEvents)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
```
TylerAPfledderer, line 79: Should the `locale` variable be added as a dep to this `useEffect`?

.github/workflows/quality.yml
pettinarip: Ah good catch! yea, I think that at this point, a pre commit hook would be just easier. We could use something like https://typicode.github.io/husky/ that is a well-known lib for this.

TylerAPfledderer: @pettinarip, I've used Husky. So I will look at converting to the pre-commit! 😁

pettinarip: Yeah, that's what I used to use for this sort of thing too. I don't know if there is a better new thing now, but Husky should work fine.

TylerAPfledderer: @pettinarip not that I know of. Just combo Husky with `lint-staged`

pettinarip: Exactly. Ok.

src/lib/rehype/remarkInferToc.ts
wackerow, line 1: Are the "falls through" comments needed here?

src/components/UpcomingEventsList.tsx
wackerow: I don't believe we need it here

.eslintrc.json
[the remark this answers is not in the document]
wackerow: Can we just modify to use `"^_$"` so the `_` must be the first/only/last character for it to match? Then we can make that an error, without it being triggered by the `_` placeholder variables

wackerow (commented):
Looking good overall! 
cc @pettinarip if you have thoughts, but this looks almost ready

src/lib/rehype/remarkInferToc.ts
TylerAPfledderer: @wackerow yes in this case. See [no-fallthrough rule](https://eslint.org/docs/latest/rules/no-fallthrough).

.eslintrc.json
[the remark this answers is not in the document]
TylerAPfledderer: @wackerow @pettinarip I had to review this again and think this through...

So the primary deal here is the use of `unused-imports/no-unused-imports-ts` from `eslint-plugin-unused-imports`, if this did not exist, an unused import would be incorrectly linted under `@typescript-eslint/no-unused-vars`.

With that, `eslint-plugin-unused-imports` recommends this basic rules setup:
```json
rules: {
    "no-unused-vars": "off", // or "@typescript-eslint/no-unused-vars": "off",
    "unused-imports/no-unused-imports": "error",
    "unused-imports/no-unused-vars": [
        "warn",
        {
            "vars": "all",
            "varsIgnorePattern": "^_",
            "args": "after-used",
            "argsIgnorePattern": "^_",
        },
    ]
}
```
And add `"unused-imports/no-unused-imports-ts": "warn"` so a warning shows for the unused import in typescript files.

> Can we just modify to use "^_$" so the _ must be the first/only/last character for it to match? Then we can make that an error, without it being triggered by the _ placeholder variables

You have to turn off `no-unused-vars` and `@typescript-eslint/no-unused-vars` so there are no unexpected overrides.

If you want to be strict to `^_$` and not allow a variable or arg to be named `_variable`, then would need to modify the recommended setup of the `unused-imports/no-unused-vars` rule above:
```json
"unused-imports/no-unused-vars": [
  "error",
    {
    "args": "all",
    "argsIgnorePattern": "^_$",
    "varsIgnorePattern": "^_$"
  }
],
```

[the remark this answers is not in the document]
wackerow: I don't think we need to not allow an arg like `_variable`, just want to make sure we can still use `_` as an unused placeholder var. Sounds like this is good where it's at.

wackerow (approved):

PR #11957, opened 2024-01-17
style: align toc item indicator with start border

wackerow:
## Description
- Adjust placement of ToC active indicator to align with the starting border, regardless of depth.

## Preview links
- https://deploy-preview-11957--ethereumorg.netlify.app/whitepaper
- https://deploy-preview-11957--ethereumorg.netlify.app/fa/security

## Screenshot
image

## Related Issue
https://www.notion.so/efdn/BUG-ToC-circle-marker-out-of-position-54104c6d5395479a90ac3e2b6a737fb3?pvs=4


wackerow:
image

Spotted this on deeper levels, will investigate

wackerow:
Looks like we have some weird logic here and old usage of `css` styles. We start the `depth` counter at zero (which I disagree with here), and have n `isNested` boolean if `depth === 2`. We then apply css classes for `.nested` and `.active`... we should be using Chakra style props in this case.

We've been displaying items more deeply nested than we did prior to the Next.js migration, which per our discussions has increased the cluttered appearance of the ToC.

I can patch this:
- Set the _starting_ depth to be 1 (representing the `h2` elements, as the "first" depth)
- Update to use Chakra over css classes
- Make sure our `sidebarDepth` override is working appropriately as the max depth (where `sidebarDepth: 3` would result in showing from h2 to h4)
- Ensure default maxDepth of 2 is working properly, and declared only once (removing any redundant fallbacks for maxDepth of 2)
- Ensure the circle indicator (active/hover) is positioned appropriately for both LTR/RTL directions at all depth levels.

wackerow:
Hey @TylerAPfledderer, I'm curious if you can spot any reason here as the cause of this layout direction bug... In the above ar/security example, the _page_ should be RTL, the _content_ is not translated though, so that flips back to LTR... _but_, the ToC active/circle indicator didn't seem to get the memo.

TylerAPfledderer:
> Hey @TylerAPfledderer, I'm curious if you can spot any reason here as the cause of this layout direction bug... In the above ar/security example, the _page_ should be RTL, the _content_ is not translated though, so that flips back to LTR... _but_, the ToC active/circle indicator didn't seem to get the memo.

Should the circle indicator still be reliant only on the `useRTLFlip()` hook? Seems like an additional check is missing.

wackerow:
@TylerAPfledderer Hm, well the `{ flipForRtl }` value specifically that we pull off of `useRtlFlip` is only intended for visual items that need to be flipped horizontally (`scaleX(-1)`)...  So for this layout, this is really only needed for the angle indicator (`⌞`).

This shouldn't be used for positioning, but it does raise a point that this hook should also be made aware if `contentNotTranslated` since we would no longer want those items to be visually flipped.

The layout and positioning _should_ be controlled through direction-agnostic styling, ie. `ms` instead of `ml`, or `insetInlineStart` instead of `left`, and then inherit the correct direction from higher-up DOM elements

TylerAPfledderer:
That's right.

I think you need to set the `dir` prop at a higher up level if not the `ButtonLink` so that RTL-aware props on the indicator can respond to that prop instead. Is that what you are thinking of?

TylerAPfledderer:
@wackerow pinging back here if there is still a blocker here with your previous question

wackerow:
Closing this our in lieu of #13156, can re-open or cherry-pick as needed

PR #11992, opened 2024-01-19
Strongly type all Quiz data types [refactor]

wackerow:
## Description
- Refactor all Quiz-related `interface`s to `type`s; migrate from `interfaces.ts` to `types.ts`
- Creates `QuizKey`, `QuestionKey`, and `AnswerKey` types
- Rolls these new types out throughout Quiz components
- Some related changes to get these components up-to-date with latest site conventions

`QuizKey`, `QuestionKey`, and `AnswerKey` types: Implemented using `satisfies` and `keyof type` pattern to both quiz and question banks. Allows us to create these types based on the strings chosen in each data mapping (`keyof typeof` pattern), while maintaining type-checking for child object structure (using `satisfies`)


src/lib/types.ts
```diff
@@ -132,17 +135,60 @@ export type LoadingState<T> =
...
+
+export type RawQuestion = {
+  prompt: TranslationKey
+  answers: Answer[]
+  correctAnswerId: string
+}
+
+export type QuestionBank = Record<string, RawQuestion>
+export type QuestionKey = keyof typeof allQuestionData
+export type AnswerKey = typeof allQuestionData[QuestionKey]["answers"][number]["id"]
```
pettinarip, line 154: Pretty cool

src/data/quizzes/questionBank.ts
```diff
@@ -887,6 +887,6 @@ const questionBank: QuestionBank = {
     ],
     correctAnswerId: "h005-b",
   },
-}
+} as const satisfies QuestionBank
```
pettinarip, line 890: Is there a benefit doing `as const satisfies` instead of just `satisfies`?

src/components/Quiz/QuizWidget/useQuizWidget.tsx
```diff
@@ -46,18 +45,9 @@ export const useQuizWidget = ({
     setUserQuizProgress([])
     setShowAnswer(false)
 
-    const currentQuizKey =
-      quizKey ||
-      Object.keys(allQuizzesData).filter((quizUri) =>
-        window?.location.href.includes(quizUri)
-      )[0] ||
```
pettinarip, line 53: I did a search and it seems that we always pass a `quizKey`, this default is never used. I'm ok with removing it if not used, but just curious about its original purpose, do you know it?

src/components/Quiz/QuizWidget/useQuizWidget.tsx
```diff
@@ -99,10 +89,7 @@ export const useQuizWidget = ({
   const quizScore = Math.floor(ratioCorrect * 100)
   const isPassingScore = quizScore > PASSING_QUIZ_SCORE
 
-  const showConfetti = useMemo<boolean>(
-    () => !!quizData && showResults && isPassingScore,
-    [quizData, showResults, isPassingScore]
-  )
```
pettinarip, line 105: Makes sense 👍🏼

pettinarip (approved):
Looks good @wackerow good refactor. Left a couple of questions, thanks!

src/data/quizzes/questionBank.ts
wackerow: I tried without, but had to add back `as const` ([in this commit](https://github.com/ethereum/ethereum-org-website/pull/11992/commits/e5d5f1729624de9a4301910b66745a7af3d94715)) after it wasn't casting the `typeof` as literals to perform `keyof` on... instead it was just simplified to `string` (IIRC)...

The other place `satisfies` is used didn't give me that problem, so I kept `as const` off of that one.

src/components/Quiz/QuizWidget/useQuizWidget.tsx
wackerow: I'm honestly not sure... but that was my same reasoning... there was no circumstance where the fallback would be used since a valid key is required.

wackerow: Thinking a little here, it may have simply been a product of not having proper typing, so any `string` could be used which may-or-may-not be available. This shouldn't be an issue now.

PR #11997, opened 2024-01-22
better described how gas is used when interacting with a smart contra…

### LINKED ISSUE #11989: [Documentation discussion] Gas cost for smart contract interaction
When reading the Developers documentation about [Transaction](https://ethereum.org/en/developers/docs/transactions), in the [On Gas](https://ethereum.org/en/developers/docs/transactions#on-gas) session it says

> Gas is required for any smart contract interaction too.

But to my understanding, it is not true. A view function as described in [solidity](https://docs.soliditylang.org/en/v0.8.23/contracts.html#view-functions), does not need any gas to run (as long as the function call is done from the outside).

This can easily be checked by deploying this simple contract:
```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract SimpleContract {
    address private _owner;

    constructor() {
        _owner = msg.sender;
    }

    function getOwner() public view onlyOwner returns(address){
        return _owner;
    }
}
```
If you call the `getOwner` function (from remix or from the tool of your choice), you are not paying anything to get the result.

Did I misunderstood something or is this sentence a bit confusing ?

stolab:
## Description
Suggestion to better describe how the gas is used when interacting with a smart contract. The documentation say
> Gas is required for any smart contract interaction too.

But there is a use case when you don't have to pay any gas when calling a smart contract function.

## Related Issue
[Fixes #11989 ]


public/content/developers/docs/transactions/index.md
```diff
@@ -160,7 +160,9 @@ The base fee will be burned **-0.00399 ETH**
 
 Validator keeps the tip **+0.000210 ETH**
 
-Gas is required for any smart contract interaction too.
+In case of an interaction with a smart contract there is 2 cases possible:
+1. A `view` or `pure` function (as described in [solidity documentation](https://docs.soliditylang.org/en/v0.8.23/contracts.html#view-functions)) is called from an EOA. In that specific case, no gas will be paid by the caller and the result will be returned. The underlying RPC call for this scenario is [eth_call](https://ethereum.org/en/developers/docs/apis/json-rpc#eth_call)
```
wackerow, line 164: I see where you're going with this but it doesn't seem appropriate for where it is on the page. I might break this into it's own h3 header, ie `### Smart contract interactions {#smart-contract-interactions}`

I would probably leave something similar to the comment you deleted as well though, just updated. Maybe "Gas is require for any transaction that alters the state of a smart contract" instead... weaving in the rest of the content. 

Also, the link to `eth_call` should be relative: `/developers/docs/apis/json-rpc#eth_call`, and consider wrapping the term itself in single-backticks to style it as code.

wackerow (changes requested):
Hey @stolab, thanks for this! Sorry for delay. I like the direction of the content you're suggesting, just had a few thoughts on how we could clean it up a bit more. Let me know your thoughts of if you'd like someone else to resume from here. Thanks!

public/content/developers/docs/transactions/index.md
stolab: Hey @wackerow, thanks for your feedback, I made changes accordingly.

I agree for the h3 header, but in that case I would suggest to rename the `### On gas {#on-gas}` to `### Regular transactions {#regular-transactions}`. Let me know what's your opinion is on this

public/content/developers/docs/transactions/index.md
```diff
@@ -160,13 +160,20 @@ The base fee will be burned **-0.00399 ETH**
...
-Gas is required for any smart contract interaction too.
 
 ![Diagram showing how unused gas is refunded](./gas-tx.png)
 _Diagram adapted from [Ethereum EVM illustrated](https://takenobu-hs.github.io/downloads/ethereum_evm_illustrated.pdf)_
 
 Any gas not used in a transaction is refunded to the user account.
 
+### Smart contract interactions {#smart-contract-interactions}
+
+Gas is required for any transaction that alters the state of a smart contract.
```
wackerow, line 171: This wording is a little redundant, as a "transaction" does commit alterations to the chain: https://ethereum.org/en/glossary/#transaction

By changing "interaction" to "transaction" we could probably just drop the rest:

```suggestion
Gas is required for any transaction that involves a smart contract.
```

public/content/developers/docs/transactions/index.md
```diff
@@ -160,13 +160,20 @@ The base fee will be burned **-0.00399 ETH**
...
+
+Therefore, calling a [`view`](https://docs.soliditylang.org/en/latest/contracts.html#view-functions) or [`pure`](https://docs.soliditylang.org/en/latest/contracts.html#pure-functions) function from an EOA will not require any gas. The underlying RPC call for this scenario is [`eth_call`](/developers/docs/apis/json-rpc#eth_call)
```
wackerow, line 173:
```suggestion
Smart contracts can also contain functions known as [`view`](https://docs.soliditylang.org/en/latest/contracts.html#view-functions) or [`pure`](https://docs.soliditylang.org/en/latest/contracts.html#pure-functions) functions, which do not alter the state of the contract. As such, calling these functions from an EOA will not require any gas. The underlying RPC call for this scenario is [`eth_call`](/developers/docs/apis/json-rpc#eth_call)
```

public/content/developers/docs/transactions/index.md
```diff
@@ -160,13 +160,20 @@ The base fee will be burned **-0.00399 ETH**
...
+
+However calling a `view` or `pure` function internally (ie: from the contract itself or from another contract) will cost gas.
```
wackerow, line 175:
```suggestion
Unlike when accessed using `eth_call`, these `view` or `pure` functions are also commonly called internally (ie. from the contract itself or from another contract) which does cost gas.
```

wackerow (changes requested):
@stolab Like the updates! Left a few thoughts on some final touches, let me know what you think 😄

stolab:
@wackerow Yes indeed your suggestion sounds much better ! thanks for the suggestion I merged them

public/content/developers/docs/transactions/index.md
```diff
@@ -160,13 +160,20 @@ The base fee will be burned **-0.00399 ETH**
...
 ![Diagram showing how unused gas is refunded](./gas-tx.png)
 _Diagram adapted from [Ethereum EVM illustrated](https://takenobu-hs.github.io/downloads/ethereum_evm_illustrated.pdf)_
 
 Any gas not used in a transaction is refunded to the user account.
 
+### Smart contract interactions {#smart-contract-interactions}
+
+Gas is required for any transaction that involves a smart contract.
+
+Smart contracts can also contain functions known as [`view`](https://docs.soliditylang.org/en/latest/contracts.html#view-functions) or [`pure`](https://docs.soliditylang.org/en/latest/contracts.html#pure-functions) functions, which do not alter the state of the contract. As such, calling these functions from an EOA will not require any gas. The underlying RPC call for this scenario is [`eth_call`](/developers/docs/apis/json-rpc#eth_call)
```
minimalsm, line 173: I feel like we could simplify this whole thing and say that any attempt at state change requires gas, but reading Ethereum does not.

stolab: Yes agree it could be simplified but as this is developer doc I thought it make sense to be a bit more explanatory. And as calling `view` and `pure` function internally cost gas I think that make sense to keep the part explaining the those two things
However I agree that the underlying RPC call is probably not the most interesting/relevant thing

wackerow (approved):
Gonna lean on the side of keeping a bit more detail here given dev docs location; pulling in

stolab:
@wackerow, can we merge it then ? not sure why exactly one check is not successful

wackerow:
Oops, apologies @stolab, didn't mean to let this sit. No worries about that check; not required (and a false negative).

wackerow:
@all-contributors please add @stolab for content

stolab:
@wackerow thanks for  your time and the discussion we had on this PR!

PR #11998, opened 2024-01-22
Local fonts

pettinarip:
Currently, our fonts are consumed from Google by using `next/font/google`. As these fonts are fetched at build time to then be stored and selfhosted, sometimes those fetches fail making the build to fail.

## Description

Add the fonts to the repo to fetch them locally at build time instead of relying on third-party services.


corwintines (approved):
🚀

corwintines (changes requested):
Sorry, approved this but noticing some differences now. 

The font spacing seems off only for the Inter font, but IBMPlexMono (used in code) is fine. 

Screenshot 2024-01-23 at 1 54 20 PM
Screenshot 2024-01-23 at 1 54 34 PM

wackerow:
+1, noticing a slight difference in how Inter appears... `mono` fonts look unchanged

wackerow (approved):
Nice, perfect match now... pulling in. Thanks @pettinarip!

PR #12000, opened 2024-01-23
Add 2024 community events and refactor to typescript

corwintines:
## Description

- Refactor community-events.json to be typescript
- Clean old 2023 events
- Add 2024 events
- Fix load more bug where it would show when there were no more events to load


src/data/community-events.ts
```diff
@@ -0,0 +1,330 @@
+export interface CommunityEvent {
+  title: string
+  to: string
+  location: string
+  description: string
+  startDate: string
+  endDate: string
+}
```
wackerow, line 8: Per our conventions, would refactor to `type`, move to/export from `@/lib/types`, and then import it here and in `UpcomingEventsList.tsx`

src/components/UpcomingEventsList.tsx
```diff
@@ -5,24 +5,14 @@ import Translation from "@/components/Translation"
...
-  to: string
-  sponsor: string | null
-  location: string
-  description: string
-  startDate: string
-  endDate: string
-}
-
-interface IOrderedUpcomingEventType extends ICommunityEventData {
+interface IOrderedUpcomingEventType extends CommunityEvent {
```
wackerow, line 15: Would refactor to `type`, ditch the `I` prefix and `Type` suffix in the name, and consider moving to `@/lib/types` (though for this one it doesn't seem as imperative, if you want to just keep it here).

src/components/UpcomingEventsList.tsx
```diff
@@ -33,7 +23,6 @@ const UpcomingEventsList: React.FC = () => {
     Array<IOrderedUpcomingEventType>
```
wackerow, line 23: (While we're cleaning up... would switch to `T[]` syntax here over `Array<T>`... Also, we can remove the `: React.FC` typing per conventions)

src/components/UpcomingEventsList.tsx
```diff
@@ -70,9 +59,7 @@ const UpcomingEventsList: React.FC = () => {
               event.endDate
             ).toLocaleDateString()}`
 
-      const details = `${event.sponsor ? "(" + event.sponsor + ")" : ""} ${
-        event.description
-      }`
+      const details = `${event.description}`
```
wackerow, line 62: 👍

wackerow (changes requested):
Nice! Overall looks good, just left a few suggestions (while we're in this file cleaning up TypeScript) to tidy up a couple more things per our code conventions

src/data/community-events.ts
```diff
@@ -0,0 +1,395 @@
+import { CommunityConference } from "@/lib/types"
+
+const CommunityConferences: CommunityConference[] = [
```
wackerow, line 3: @corwintines Given this isn't a component, we should lowercase this variable name

src/components/UpcomingEventsList.tsx
```diff
@@ -1,39 +1,30 @@
...
 import { Box } from "@chakra-ui/react"
 
+import type { CommunityConference } from "@/lib/types"
+
 import Translation from "@/components/Translation"
 
 import { trackCustomEvent } from "@/lib/utils/matomo"
 
-import events from "../data/community-events.json"
+import CommunityConferences from "../data/community-events"
```
wackerow, line 10: (Similar to where it's declared, this isn't a component so I'd lowercase it by convention)

src/components/UpcomingEventsList.tsx
```diff
@@ -1,39 +1,30 @@
...
-  to: string
-  sponsor: string | null
-  location: string
-  description: string
-  startDate: string
-  endDate: string
-}
-
-interface IOrderedUpcomingEventType extends ICommunityEventData {
+type OrderedUpcomingEventType = CommunityConference & {
```
wackerow, line 17: small nitpick that having "Type" in the name is redundant

wackerow (changes requested):

src/components/UpcomingEventsList.tsx
```diff
@@ -1,39 +1,30 @@
...
 
-interface IOrderedUpcomingEventType extends ICommunityEventData {
+type OrderedUpcomingEvent = CommunityConference & {
   date: string
   formattedDetails: string
 }
 
-const UpcomingEventsList: React.FC = () => {
+const UpcomingEventsList = () => {
+  const { t } = useTranslation("page-community")
```
wackerow, line 22: Screenshot 2024-01-24 at 17 15 42

Found a string for an empty set that was missing its namespace prefix... went ahead and swapped this to use `useTranslation` instead of the `Translation` component.

wackerow (approved):
Nice job!

PR #12009, opened 2024-01-24
fix: Internal transaction term visibility in glossary list

Shiva-Sai-ssb:
## Description

I've fixed the issue that was making the definition and term for internal transactions not show up in the glossary list. Look at the screenshots for a clearer understanding.

### Before
Screenshot (115)

### After
Screenshot (114)


corwintines (approved):
Ah great, thank you @Shiva-Sai-ssb!

PR #12014, opened 2024-01-25
refactor: replace `React.FC` and `interface` with direct declaration and `type`

TylerAPfledderer:
## Description
To update the project to be inline with the [latest code convention docs](https://github.com/ethereum/ethereum-org-website/blob/dev/docs/next-docs/code-conventions.md#component-props), this PR does the following:

- Replace all instances of the `React.FC` type with direct component prop declaration
```tsx
// From
const Component: React.FC<Props> = (props) => JSX.Element

// To
const Component = (props: Props) => JSX.Element
```
- Replace all instances of `interface` with `type` for component prop shapes only in order to provide typing constraint.
- Provide meaningful naming to type aliases of all component props that matches the name of the component .
- Remove redundant type signatures. i.e. Redundant instance of type signatures similar to `ChildOnlyProp`
- Other cleanup and adjustments as necessary for the above changes

> Side note: This helps in addressing other possible type constraint additions, such as better typing for custom MDX components, where type errors might arise from conflicts with `React.FC`

## Related Issue
N/A


corwintines (approved):
Awesome! Thanks @TylerAPfledderer!

PR #12015, opened 2024-01-25
fix(CallToAction): update and constrain prop object type shape

TylerAPfledderer:
## Description

This PR primarily addresses a React console warning from the `CallToAction` hero component:
```bash
Warning: CallToAction: `key` is not a prop. Trying to access it will result in `undefined` being returned. If you need to access the same value within the child component, you should pass it as a different prop.
```

Trying to use the `key` prop will only result in receiving `undefined` and create styling bugs when using it to render the button variant. Instead create unique prop `index` to access an index value for use in the props logic.

Also modifies typing related `CallToAction` so that:
- Usage of the `Button` component props `variant` and `isSecondary` are constrained to only the logic within `CallToAction` to determine their values per instance
- When creating data for the buttons in a hero component the user is not required to explicitly supply an `index` value when that is provided via render mapping.

## Related Issue
N/A


wackerow (approved):
Lgtm, thanks @TylerAPfledderer!

PR #12033, opened 2024-01-26
feat: Header updates

nhsz:
This PR adds some requested header changes (see [Figma](https://www.figma.com/file/rv850lIICwOwRnyqSdtZED/Redesign-find-wallet?type=design&node-id=1079-2783&mode=design&t=aQStkUJB7U44Jm3j-0)) to `/find-wallets` page

## Description

- adds link to wallets page in the header
- updates header illustration 
- updates text in the header
- rename Main headline to “Choose your wallet”
- align to the left end of list “Showing all wallets (53)” text

**Design:** https://www.figma.com/file/rv850lIICwOwRnyqSdtZED/Redesign-find-wallet

**Preview:** https://deploy-preview-12033--ethereumorg.netlify.app/wallets/find-wallet


src/pages/wallets/find-wallet.tsx
```diff
@@ -138,7 +139,13 @@ const FindWalletPage = () => {
   }
 
   return (
-    <Flex as={MainArticle} direction="column" position="relative" w="full">
+    <Flex
+      as={MainArticle}
+      direction="column"
+      position="relative"
+      w="full"
+      suppressHydrationWarning
```
pettinarip, line 147: I don't think it is a good idea to suppress this warning. At some point we should make the effort to fix the root problem.
```suggestion
```

public/wallets/find-wallet-hero.png
pettinarip, line 1: We are already using this image in the `/wallets` page. We should reuse the same one.

pettinarip (changes requested):

pettinarip:
@nhsz fyi the preview deploy in the description is incorrect. Should be https://deploy-preview-12033--ethereumorg.netlify.app/wallets/find-wallet

src/pages/wallets/find-wallet.tsx
corwintines: Agree here, should this be fixed in the PR open for wallet filtering? Pretty sure the hydration error is caused by the shuffle being done client side.

public/wallets/find-wallet-hero.png
corwintines, line 1: Since this is used on the /wallets page, we should either check we want both changed, or add this new image and change only the find wallets page. @nloureiro @konopkja do you have any preferences around the hero image?

corwintines (changes requested):
Looks great overall, just one thing with the image that we should check

public/wallets/find-wallet-hero.png
konopkja: i intended to have the same image there because they are directly related pages, but if we want to have a variation in the future, thats possible

nhsz:
> Looks great overall, just one thing with the image that we should check

@pettinarip @corwintines Updated, re-using the same hero from `/wallets` as defined by design

PR #12034, opened 2024-01-27
build: add Chakra CLI

TylerAPfledderer:
## Description

This PR adds back in the Chakra CLI to generate typing based on the custom theme configuration. This is helpful for IDE auto-completion to be able to recognize the correct tokens.

- Adds scripts to both run the command once, and to have a watcher to update changes on the go.
- Runs the postinstall script to generate the theme typing once `yarn install` has run

## Related Issue

N/A


package.json
```diff
@@ -3,7 +3,8 @@
   "version": "8.0.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "npx concurrently \"yarn next dev\" \"yarn theme:watch\"",
```
TylerAPfledderer, line 6: This is very much a personal preference. concurrently is not added to the project as a dev dependency, but run remotely. The only requirement that would be prompted in the console is for the user to install it to their local machine. (This package is used in Chakra projects)

This command ensures that the dev environment and the Chakra CLI run in parallel without the user needing to remember an extra step.

Can certainly revert this if this is too much.

package.json
```diff
@@ -16,7 +17,9 @@
     "build-storybook": "storybook build",
     "crowdin-clean": "rm -rf .crowdin && mkdir .crowdin",
     "crowdin-import": "ts-node src/scripts/crowdin-import.ts",
-    "markdown-checker": "ts-node -O '{ \"module\": \"commonjs\" }' src/scripts/markdownChecker.ts"
+    "markdown-checker": "ts-node -O '{ \"module\": \"commonjs\" }' src/scripts/markdownChecker.ts",
+    "theme": "chakra-cli tokens ./src/@chakra-ui/theme.ts --strict-token-types --strict-component-types",
+    "theme:watch": "chakra-cli tokens ./src/@chakra-ui/theme.ts --strict-token-types --strict-component-types --watch"
```
TylerAPfledderer, line 22: The `strict-token-types` and `strict-component-types` flags are used to constrain the typing, so that a combination of the `string` and `{}` primitives are not part of the type unions.

At the moment, this can aid in auto-completion. When talking with a member of the Chakra team, we recognized two issues (one kinda related):
- If we have a component that is taking is props that are typed with a shape from Chakra (i.e. StackProps) and you spread them into a component (i.e. Stack), you lose typing if you also directly add props to the component
```tsx
<Stack
  color="blue.500" // Typing lost here
  {...props}
/>
```
- I would expect that with strict typing of the themes that TypeScript would throw an error if a typo or invalid token exists. This unfortunately does not seem to be the case.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ESNext",
```
TylerAPfledderer, line 3: Otherwise an error throws with `es5`:
`Transforming object literal extensions to the configured target environment ("es5") is not supported yet`

TylerAPfledderer (commented):
A couple of thoughts...

TylerAPfledderer:
On initial failed build through Netlify, I believe this is exposing an inconsistency.

When `yarn lint` this only does checks with eslint. This does not account for TypeScript. Waiting for a build to run via GH CI and Netlify seems slow for Dev Experience.

I would consider a script to run `tsc` at the very least, maybe include it as part of `lint-staged` to ensure it's checked locally.

TylerAPfledderer:
@pettinarip see the above comment. I'm going to stick this PR back into draft until I look at it further.

Basically, the errors in the build were actually what I was hoping for, but was not expecting it to stay in a preview deploy check. I'm going to take some time to play around with this. If it gets too complicated, I'll remove the strict typing flags from the theme script and look at those later.

I think restricting the typing if possible would help make sure typos or invalid token values are not overlooked and not bleed through.

package.json
TylerAPfledderer: These flags have now been removed.

TylerAPfledderer:
After further triage, I have removed those strict typing flags from the scripts and propose the following:

- Open up GH issue for tracking to explore these flags separately. They are exposing typing issues that are hidden. Plus...
- Set up a package script to run typescript checks locally, since this isn't done otherwise unless running the build script. This way, we are able to do a complete check of lints and type safety before pushing commits.

package.json
pettinarip: Agreed, this is more of a personal preference. In that sense, I'd prefer to keep the `dev` script as clean as possible.

However, I do see the benefit of running the Chakra script in parallel. We could add a comment in the readme or docs to let devs know how this works.

cc @wackerow @nhsz @corwintines

tsconfig.json
pettinarip: Won't this affect our browser support? if not sure, I'd recommend doing this in a separate PR to run tests accordingly.

pettinarip (changes requested):

tsconfig.json
TylerAPfledderer: According to the CanIUse site, ES6 shows 98% support.

https://caniuse.com/?search=ECMASCRIPT

package.json
nhsz: Agree that this should be opt-in with another script, not default `npm dev` behavior, which should follow the standard.

TylerAPfledderer: I have reverted the `dev` script.

Would you want some comment added in this PR?

tsconfig.json
TylerAPfledderer: I failed to clarify that this errors throws on instances of using `cssVar()` in component themes and attempting to compile dynamic keys... say:
```ts
[$colorBg.variable]: 'some value'
```
Which is an ES6 only feature. Not sure if this is a Chakra CLI issue (doing testing in the Chakra package is not recreating the error), and really a typescript issue within this project.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ES6",
```
TylerAPfledderer, line 3: @pettinarip your question here and my response was removed due to a minor change I did.

So as to if this effects browser support, ES6 is supported by roughly 98% from [caniuse.com](https://caniuse.com/es6).

[The TypeScript docs of course also recommends ES6](https://www.typescriptlang.org/tsconfig#target)

Because there are dynamic keys in some component themes (i.e. `[$border.variable]`), the CLI will fail with the `ES5` target because dynamic keys are part of the`ES6` spec.

wackerow:
> However, I do see the benefit of running the Chakra script in parallel. We could add a comment in the readme or docs to let devs know how this works.

@pettinarip Personally would just add a `dev:watch` script, but not a strong preference here.

pettinarip:
> > However, I do see the benefit of running the Chakra script in parallel. We could add a comment in the readme or docs to let devs know how this works.
> 
> @pettinarip Personally would just add a `dev:watch` script, but not a strong preference here.

I'd prefer to minimize the number of scripts we have to make it simpler and more manageable. But if you think that adding that script will add value and will be used by a majority of devs, then I'd totally support that.

What I wouldn't want to happen is to have a lot of scripts that nobody uses.

pettinarip (approved):
LGTM! thanks @TylerAPfledderer

Issue #12035, opened 2024-01-27
Feedback Widget Causes Horizontal Scrolling


ArianHamdi:
### Describe the bug

Hovering over the feedback widget element results in an unsightly horizontal scrollbar. This only occurs when the window size is smaller than the max-width of the body.

### To reproduce

1. Visit https://ethereum.com/.
2. Hover over the feedback widget element.
3. Notice the jump due to the horizontal scrollbar.

### Expected behavior

The widget should not create a horizontal scrollbar, and hovering over it should be a smooth experience.

### Screenshots

https://github.com/ethereum/ethereum-org-website/assets/59548833/3a948529-7010-4e6d-bfad-d7d4bb1f6cf7

### Desktop (please complete the following information)

- OS : macOS 14.3
- Browser : Edge 120

### Would you like to work on this issue?

- [X] Yes
- [ ] No

ArianHamdi:
Hi @pettinarip,

Thank you for taking the time to check this issue.

I agree with you, and I think resolving the root problem is better.

I tried your idea for fixing this, but it didn't work the way I expected. When applying the transform to children instead of sticky parent, it still causes a scrollbar.

This was a bit tricky, and I found two solutions for this bug.

**Solution 1:**
Move the widget to the left side of the area and bring it back to the right side using the left property. In this scenario, hovering over it won't cause a scrollbar to appear.

You can see the changes here: [solution 1](https://github.com/ArianHamdi/ethereum-org-website/pull/2/files).

**Solution 2:**
Apply `transform-origin: right bottom`. In this scenario, when hover occurs, the widget element gets scaled up from the left and the upper side and doesn't cause a scrollbar. In this solution, the hovering animation is a bit different.

You can see the changes here: [solution 2](https://github.com/ArianHamdi/ethereum-org-website/pull/1/files).

Whichever solution you think is more acceptable, I can create a PR with that implementation.

pettinarip:
@ArianHamdi awesome! thanks for sharing this. Both solutions work great, I think the 1st solution works well while keeping the same animation. So, I'd go with that approach.

ArianHamdi:
> @ArianHamdi awesome! thanks for sharing this. Both solutions work great, I think the 1st solution works well while keeping the same animation. So, I'd go with that approach.

@pettinarip Thanks!
I opened a PR with first solution implementation: #12052

Issue #7163, opened 2022-07-22
Opcode table horizontal scrolling


jmcook1186:
**Describe the bug**

The Opcode table on this page https://ethereum.org/en/developers/docs/evm/opcodes is hard to use because of the horizontal scroll functionality. The table has a large number of rows that do not fit into a single page view - to view all the rows vertical scrolling is required. There are also more columns than can fit into a single page view. 

To access out-of-view columns it is necessary to scroll right, which currently requires scrolling to the bottom of the table. However, this puts the row/opcode number out of view, so that with the desired column in view, the row identifier is now obscured. 

This makes it very awkward to access information for specific opcodes using the table.

**To Reproduce**

1.Go to https://ethereum.org/en/developers/docs/evm/opcodes
2. Make an arbitrary choice of row, e.g. 44 - the DIFFICULTY Opcode
3. Try to view content in columns currently out of view - this requires scrolling rightwards at the bottom of the page
4. Try to find the 44 DIFFICULTY Opcode with the desired column still in view.

**Expected behavior**

Either enable arrow-key/click-and-drag horizontal scrolling or put all columns in view by default.

**Screenshots**

Screencast of the problem here: https://discordapp.com/channels/714888181740339261/999252773457633290/999634096839983275

**Desktop (please complete the following information):**

- OS: Linux (Ubunto 20.04)
- Browser: Brave
- Version: 1.40.107

**Additional context**

Issue seems to be desktop specific - on Android the issue is fixed by right-swiping without needing to scroll vertically to the scroll-bar.

vdusart:
Hey @jmcook1186,
you can do `[Shift] + scroll` to do an horizontal scroll.

We should perhaps add a small note on the top of the table to explain it.
Something like :

`Note: In order to see all the content, you can do an horizontal scroll by using [Shift] + scroll`.

I'm not sure everything can be in the view by default it's too large.

Tell me what you think about it :smile:

SebastianSupreme:
@vdusart Great idea if you ask me 👍

ArianHamdi:
<img width="1608" alt="image" src="https://github.com/ethereum/ethereum-org-website/assets/59548833/df98cc21-cf3e-4a75-953d-6f790677df14">

The horizontal scrollbar breaks the layout.

PR #12036, opened 2024-01-27
fix: apply overflow-x hidden to body

ArianHamdi:
## Fix for Unwanted Horizontal Scrollbar

This PR applies `overflow-x: hidden` to the body to prevent the appearance of a horizontal scrollbar in certain scenarios.

## Related Issue
Fixes: #7163 
Fixes: #12035

### DIFF
--- a/src/@chakra-ui/styles.ts
+++ b/src/@chakra-ui/styles.ts
@@ -13,6 +13,7 @@ const styles = {
       bg: "background.base",
       lineHeight: "base",
       fontSize: "md",
+      overflowX: "hidden",
     },
     a: {
       color: "primary.base",


pettinarip (commented):
Hey @ArianHamdi thanks for the PR!

Adding overflow hidden does fix the visual problems but it doesn't fix the root problems we have in those issues. Ideally, we shouldn't have overflow:hidden because it could interfere with future implementations and become more of a problem than a solution.

Having said that, I think we will need different solutions for those issues.

I'll close this PR for now and keep the conversation about possible solutions in the issues. Feel free to open a new PR later on.

PR #12037, opened 2024-01-28
fix(contributors): create shape for allTimeData json type

TylerAPfledderer:
## Description

Currently in the `Contributors` page,  there is a flatmap using the `allTimeData` json to create a simplified array of translator info `translatorData`. 

A type error currently throws: `Property 'data' does not exist on type '{}'`

This is because a type is not being inferred for the json file and is only providing an empty object.

This PR creates the shape for this data structure so there is type safety in the flatmap.

The shape was generated using [Transform Web Converter App](https://transform.tools/json-to-typescript) and double-checked for any undefined fields with the data used.

According to the shape generated, there should be no fields in the JSON structure that would be undefined or null.

## Related Issue
N/A

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -19,6 +19,9 @@ import type { CallToActionProps } from "@/components/Hero/CallToAction"
 
 import { layoutMapping } from "@/pages/[...slug]"
 
+// Credit: https://stackoverflow.com/a/52331580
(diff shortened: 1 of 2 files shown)


pettinarip (approved):
LGTM, thanks @TylerAPfledderer

PR #12042, opened 2024-01-29
feat: supported locale filter

nhsz:
This PR updates the wallets data display order. If a wallet supports user's locale, they appear at the top of the table, shuffled.

**Preview:** https://deploy-preview-12042--ethereumorg.netlify.app/wallets/find-wallet


src/pages/wallets/find-wallet.tsx
```diff
@@ -107,15 +108,21 @@ export const getStaticProps = (async ({ locale }) => {
...
   const { isOpen: showMobileSidebar, onOpen, onClose } = useDisclosure()
   const [filters, setFilters] = useState(filterDefault)
   const [selectedPersona, setSelectedPersona] = useState(NaN)
 
+  // If any wallet supports user's locale, show them (shuffled) at the top and then the remaining ones
+  const supportedLocaleWallets = getSupportedLocaleWallets(locale!)
+  const noSupportedLocaleWallets = getNonSupportedLocaleWallets(locale!)
+  const randomizedWalletData = supportedLocaleWallets.concat(
+    noSupportedLocaleWallets
+  )
```
pettinarip, line 124: Since we are doing a `shuffle`, we will get a different result on each render, causing a hydration warning. We should only do this on client side (maybe inside a `useEffect`).

corwintines:
I think this looks great overall, agree with @pettinarip comment above if we want to address this in this PR though. Great job though!

nhsz:
> I think this looks great overall, agree with @pettinarip comment above if we want to address this in this PR though. Great job though!

@pettinarip @corwintines I've added the `useEffect` in the last commit but still getting the hydration error

nhsz:
> I think this looks great overall, agree with @pettinarip comment above if we want to address this in this PR though. Great job though!

@pettinarip @corwintines I've added the `useEffect` in the last commit to set the random wallet data but still getting the hydration error

pettinarip (approved):
LGTM @nhsz!

I think we have several hydration issues. The one you are getting may be related to the one you mentioned in the other PR. It might be coming from an upper component in the tree (?).

PR #12050, opened 2024-01-29
Fix i18n links not having the locale in hrefs

pettinarip:
Currently, when the user loads a page, all the internal navigation links display the `href` without the locale that the user is using.
For example, when you load https://ethereum.org/de/layer-2, all the internal links ignore the `de` locale and build the href from the root level.

## Description

The problem seems to come from the `Link` component imported from [`@chakra-ui/next-js`](https://github.com/chakra-ui/chakra-ui/tree/main/packages/next-js). I'll try to reproduce this bug on a codesandbox later and open an issue in that repo.

For the scope of this PR, it:
- Removes the dependency
- Implements a custom integration between Chakra & Next Link component

### DIFF
--- a/src/lib/utils/cx.ts
+++ b/src/lib/utils/cx.ts
@@ -1,0 +1,1 @@
+export const cx = (...classNames: any[]) => classNames.filter(Boolean).join(" ")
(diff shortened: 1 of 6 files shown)


src/lib/utils/cx.ts
corwintines, line 1: I don't see this used anywhere, is this necessary?

pettinarip: Ah, good catch. That was left over from a previous implementation. Let me remove it

corwintines (approved):
LGTM

PR #12052, opened 2024-01-30
fix: feedback widget by modifying left property

ArianHamdi:
## Description
This PR prevent the appearance of a horizontal scrollbar wen hovering over the feedback widget.

## Related Issue

Fixes: #12035


pettinarip (commented):
@ArianHamdi thanks again for the help on this.

The button is perfectly placed now and it doesn't create any horizontal overflow. However, when you click on the button and open the feedback popover, I see this:
image

It is not aligned with the button. Perhaps we need to apply the same `left` prop in the popover container.

ArianHamdi:
> @ArianHamdi thanks again for the help on this.
> 
> The button is perfectly placed now and it doesn't create any horizontal overflow. However, when you click on the button and open the feedback popover, I see this: image
> 
> It is not aligned with the button. Perhaps we need to apply the same `left` prop in the popover container.

Yes, you are right, @pettinarip.

This is the behavior currently observed on the website, and changing the `left` property didn't produce it.

If you want the dialog to render in the same position as the button, I will try to fix it too.

ArianHamdi:
Hi @pettinarip,

Could you please check it again? I tried a workaround to align the popup position. Also, there was a bug that caused a horizontal jump when clicking on the button, but that has been fixed too.

pettinarip (approved):
Thanks @ArianHamdi!

This is working as expected now 💪🏼 

FYI I'm creating a new PR to migrate the underlying component (AlertDialog) to a Popover since it is a better component to use in this case. Happy to have your eyes and feedback on that one as well if you are interested!

PR #12058, opened 2024-01-30
Change operator `||` to `??` where reading environmental valuable  [Fixes #12057]

### LINKED ISSUE #12057: Bug report: Refactoring by using The Nullish Coalescing Operator (??) in settings files
### Describe the bug

Using Nullish Coalescing Operator in the settings files is safer coding.
But, it is a minimal improvement.

### To reproduce

none

### Expected behavior

none

### Screenshots

_No response_

_No response_

_No response_

### Additional context

_No response_

- [X] Yes
- [ ] No

HiroyukiNaito:
Change operator `||` to `??` where reading environmental valuable  [Fixes #12057]

## Description

When reading environmental valuables, it is safer to use '??' operator.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

## Related Issue

#12057

### DIFF
--- a/next-i18next.config.js
+++ b/next-i18next.config.js
@@ -1,6 +1,6 @@
 const i18nConfig = require("./i18n.config.json")
 
-const BUILD_LOCALES = process.env.BUILD_LOCALES
+const BUILD_LOCALES = process.env.BUILD_LOCALES ?? false
 // Supported locales defined in `i18n.config.json`
 const locales = BUILD_LOCALES
   ? BUILD_LOCALES.split(",")
--- a/next-i18next.config.js
+++ b/next-i18next.config.js
@@ -1,6 +1,6 @@
 const i18nConfig = require("./i18n.config.json")
 
-const BUILD_LOCALES = process.env.BUILD_LOCALES ?? false
+const BUILD_LOCALES = process.env.BUILD_LOCALES
 // Supported locales defined in `i18n.config.json`
 const locales = BUILD_LOCALES
   ? BUILD_LOCALES.split(",")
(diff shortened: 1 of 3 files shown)


wackerow (changes requested):
Hey, thanks for the PR @HiroyukiNaito

I agree with the `const LIMIT_CPUS = Number(process.env.LIMIT_CPUS ?? 2)` change, since notably the nullish coalescing operator `??` will allow `0` to be passed without it being treated as falsy (which we want).

Oddly, I have my `LIMIT_CPUS` env var set to `0` locally and this is still working (using `0` instead of `2`)... not totally sure why, but I'm in favor of updating this to the more appropriate operand. 

---

The other two I don't think we need/want. 

`?? false` is redundant... In either case, this value will either be `undefined` (if env var not declared) or an empty string `""` (if env var declared but left blank), both of which will be treated as "falsy" in the ternary expression beneath it. This is the desired outcome, as both of these should fall back to using all locales.

In the case of the `siteUrl`, we would NOT want to accept an empty string here...

image

Similar to how `0` is accepted with `??`, an empty string is also accepted, and will not fall back to the right-hand of the expression. Using `||` will treat `""` or `0` as falsy and move to the right side or the operand to use the hard-coded fallback.

---

Let me know if that makes sense, or if I'm missing something. cc @pettinarip curious if you have thoughts

HiroyukiNaito:
@wackerow Thank you for the deep insight.
I would change the code according to your description.

HiroyukiNaito:
@wackerow 
I changed codes according to your advice.

wackerow (approved):
Nice, thanks @HiroyukiNaito! Lgtm =)

PR #12062, opened 2024-01-30
Fixes ERC4626 broken link

mario-eth:
## Description
The old ERC4626 link to solmate was broken.


mario-eth:
Worth thinking if the OpenZeppelin implementation is also worth adding as reference

corwintines (approved):
Thanks @mario-eth!

Issue #12080, opened 2024-01-31
Previous & Next Button in Developer Documentation shows wrong link


yigittopm:
### Describe the bug

The previous and next button links in this [ethereum.org/developers/docs/intro-to-ethereum](https://ethereum.org/developers/docs/intro-to-ethereum) address do not look correct.

The following code fragment,
```javascript
if (asPath.indexOf(docsArray[i].to) > -1) { 
    currentIndex = i 
}
 ```

**if you are on this page:**
- /developers/docs/intro-to-ethereum

**accepts both of the following values:**

- /developers/docs/intro-to-ethereum
- /developers/docs/intro-to-ether

**and _currentIndex_ must be x, but is now y.**

Example:
```javascript
"/developers/docs/intro-to-ethereum".indexOf("/developers/docs/intro-to-ethereum") > -1 // true
"/developers/docs/intro-to-ethereum".indexOf("/developers/docs/intro-to-ether") > -1    // true
 ```

Therefore, currentIndex is updated to a different value. And previous & next show the wrong link.

### To reproduce

1. Go to [ethereum.org/developers/docs/intro-to-ethereum](https://ethereum.org/developers/docs/intro-to-ethereum)
2. Scroll down to the previous & next buttons.
3. Look at the previous and next pages that should be relative to your current page.
4. See error.

### Expected behavior

Each page should show the previous (if any) and next page link.

### Screenshots

<img width="1438" alt="Screenshot 2024-01-31 at 19 46 30" src="https://github.com/ethereum/ethereum-org-website/assets/62249955/4e07abbc-a277-44c7-90b3-b6857571311a">

### Desktop (please complete the following information)

- OS: macOS 14.2.1
- Browser: Brave
- Version:  1.62.153

### Would you like to work on this issue?

- [X] Yes
- [ ] No

yigittopm:
🚀 I have fixed this issue and issued a pull request. You can find the details in the PR.

PR #12083, opened 2024-02-01
Fix/dev docs prev and next link 

yigittopm:
## Description

I rechecked the 2 values captured with the `indexOf` method according to their lengths and the error was solved.

I changed the `> 0` query that causes skipping the first page when showing the previous page.

## Related Issue

- #12080

### DIFF
--- a/src/components/DocsNav.tsx
+++ b/src/components/DocsNav.tsx
@@ -137,7 +137,7 @@ const DocsNav = ({ contentNotTranslated }: DocsNavProps) => {
   }
 
   // Extract previous and next doc based on current index +/- 1
-  const previousDoc = currentIndex - 1 > 0 ? docsArray[currentIndex - 1] : null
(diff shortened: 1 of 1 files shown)


src/components/DocsNav.tsx
```diff
@@ -128,13 +128,16 @@ const DocsNav = ({ contentNotTranslated }: DocsNavProps) => {
   // Find index that matches current page
   let currentIndex = 0
   for (let i = 0; i < docsArray.length; i++) {
-    if (asPath.indexOf(docsArray[i].to) > -1) {
+    if (
+      asPath.indexOf(docsArray[i].to) > -1 &&
```
nhsz, line 132:
```suggestion
      asPath.indexOf(docsArray[i].to) >= 0 &&
```

nhsz (changes requested):
Nice, thanks for the fix! Small suggestion to keep consistency in how we're handling the indexes

nhsz (approved):
LGTM!

Issue #12115, opened 2024-02-05
Suggest lazy-etherscan


woxjro:
### Dev tool name

lazy-etherscan

### Dev tool description

`lazy-etherscan` is a simple Terminal UI for the Ethereum Blockchain Explorer.
The main features are the following:
- **No Browser Required** - Use it effortlessly even in environments where browsers are unavailable, such as within servers.
- **Developer Friendly** - Operate efficiently using keyboard shortcuts.
- **Easily Switch Endpoints** - Switch between endpoints, including Mainnet, Testnets, custom node connections, and even BSC endpoints.
- **Rich Search Functionality** - You can search by the following words.
    - Address
    - Block Number
    - ENS ID
    - Transaction Hash
    - Ticker Name (`USDT`, `BNB`,`UNI`, ...)

### When did the dev tool go live?

2 weeks ago

### Where do you think this page should be shown?

Here: https://ethereum.org/developers/docs/data-and-analytics/block-explorers#services

### Is the product open source?

Yes

### GitHub URL

https://github.com/woxjro/lazy-etherscan

### Documentation URL

https://woxjro.github.io/lazy-etherscan/

### Additional context

This software has been developed with a grant from the [Ethereum Foundation](https://blog.ethereum.org/2023/12/19/esp-allocation-q323) and [Gitcoin](https://www.gitcoin.co/).

### Would you like to work on this issue?

- [X] Yes
- [ ] No

corwintines:
Going to close with above PR merged

PR #12116, opened 2024-02-05
Add lazy-etherscan to the list of block explorers [Issue #12115]

woxjro:
## Description
Add [`lazy-etherscan`](https://github.com/woxjro/lazy-etherscan) to the list of block explorers.

## Related Issue
#12115

### DIFF
--- a/public/content/developers/docs/data-and-analytics/block-explorers/index.md
+++ b/public/content/developers/docs/data-and-analytics/block-explorers/index.md
@@ -28,6 +28,7 @@ You should understand the basic concepts of Ethereum so you can make sense of th
 - [Rantom](https://rantom.app/)
 
 ## Open source tools {#open-source-tools}
+
 - [Otterscan](https://otterscan.io/)
 - [lazy-etherscan](https://github.com/woxjro/lazy-etherscan)
 
--- a/public/content/developers/docs/data-and-analytics/block-explorers/index.md
+++ b/public/content/developers/docs/data-and-analytics/block-explorers/index.md
@@ -25,8 +25,11 @@ You should understand the basic concepts of Ethereum so you can make sense of th
 - [Ethplorer](https://ethplorer.io/) -_Also available in Chinese, Spanish, French, Turkish, Russian, Korean and Vietnamese_
 - [EthVM](https://www.ethvm.com/)
 - [OKLink](https://www.oklink.com/eth)
-- [Otterscan](https://otterscan.io/)
 - [Rantom](https://rantom.app/)
+
+## Open source tools {#open-source-tools}
+- [Otterscan](https://otterscan.io/)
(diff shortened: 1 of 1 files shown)


wackerow:
Hey @woxjro, very cool tool! Spun up super fast and love that I could point to my own node.

My main concern here would be it's location. This is very different from the others. I'm not against keeping it here, but I would suggest maybe a short subtext here related to be being a command-line tool. Thoughts?

Also wondering if this could be better fit in one of the pages on nodes... I see this as a nice add-on to anyone running their own RPC.

woxjro:
@wackerow 
Thank you for your suggestion!

> I would suggest maybe a short subtext here related to be being a command-line tool.

It sounds totally good!

> wondering if this could be better fit in one of the pages on nodes...

I also agree with your idea!
Is it the following link you are talking about? 
https://ethereum.org/en/developers/docs/nodes-and-clients/

I'm unsure where I have to put lazy-etherscan's link and the subtext, so if you tell me about that, I recreate PR!

public/content/developers/docs/data-and-analytics/block-explorers/index.md
```diff
@@ -25,6 +25,7 @@ You should understand the basic concepts of Ethereum so you can make sense of th
 - [Sirato](https://www.web3labs.com/sirato)
 - [EthVM](https://www.ethvm.com/)
 - [DexGuru Block Explorer](https://ethereum.dex.guru/)
+- [lazy-etherscan](https://github.com/woxjro/lazy-etherscan)
```
wackerow, line 28: @woxjro Can just add something here similar to language callouts.

```suggestion
- [lazy-etherscan](https://github.com/woxjro/lazy-etherscan) -_Command-line/terminal UI_
```

As I look at this again, it still feels awkward having this under "Services" since it's more of a tool the user can self-host, and tap into any RPC they'd like. 

I'd suggest we create a new heading here for "Open source tools" perhaps, and place lazy-etherscan and Otterscan in there (and any others I may have missed that work similarly)

```suggestion

## Open source tools {#open-source-tools}

- [Otterscan](https://otterscan.io/)
- [lazy-etherscan](https://github.com/woxjro/lazy-etherscan)
```

(Then remove Otterscan from above list)

wackerow:
Also, 

> Is it the following link you are talking about?
> https://ethereum.org/en/developers/docs/nodes-and-clients/

Yeah, loosely around that page though that one in particular is getting quite lengthy. Not sure we have a great place for "node add-ons" if-you-will.

wackerow:
...wouldn't include that in this PR though just to keep this one concise.

woxjro:
@wackerow 
Thank you for your code suggestion!
I forcefully pushed a new commit, adding a heading about otterscan and lazy-etherscan.

Issue #11955, opened 2024-01-17
Epic: Implement navigation redesign


wackerow:
## Summary
Last quarter (Q4 2023) designs were created for a revamp of the ethereum.org navigation bar. Feedback on the existing navigation structure left many users unsure of where to start, or how to access the information they are looking for quickly.

The new designs offer a tiered approach, revamping the old top-level navigation, and re-prioritizing and re-organizing different pages to improve user experience.

This quarter (Q1 2024) these design changes are planned to be implemented on the site. This includes implementing the new nav bar layout, and improving accessibility of these items for keyboard users and screen readers.

## Design
https://www.figma.com/file/VFKy8nGjdd4PuSdogBUJGA/Redesign-menu-navigation?type=design&node-id=1207-122492&mode=design&t=9ELyGK0OXhohPIGm-0

## Notion
https://www.notion.so/efdn/Implement-navigation-changes-aa93cef4f2244acb96bb12e9f54e5271?pvs=4

PR #12125, opened 2024-02-06
Nav menu - redesign implementation

wackerow:
## Description
Implements the newly redesigned navigation menu for both desktop and mobile, including changes to the information architecture

- [Figma design](https://www.figma.com/file/VFKy8nGjdd4PuSdogBUJGA/Redesign-menu-navigation?type=design&node-id=1377%3A11591&mode=design&t=IttxtLUVJ7QgAUe9-1)
- Updates the information architecture inside the `useNav` hook with new menu item structure, updated labels and addition of item descriptions/subtext.

### Desktop
- Uses Radix-UI library for `NavigationMenu` component
- Includes keyboard navigation/a11y

### Mobile
- Uses Chakra-UI `Accordion` component

## Preview link
- https://deploy-preview-12125--ethereumorg.netlify.app/

## Related issue
- #11955

## TODO
- [x] Import translated intl strings
- [x] Improve animations


src/components/Nav/Menu/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+          </NavigationMenu.List>
+          <Portal.Root container={refs.lvl1.current}>
+            <NavigationMenu.Viewport />
+          </Portal.Root>
+        </NavigationMenu.Root>
+      </Box>
+
+      <Grid
+        position="absolute"
+        visibility="hidden"
```
pettinarip, line 73: Not blocker. Wondering if we should add some `aria-hidden` attribute here for sr, and switch it when the menu is shown.

pettinarip (commented):
This version looks wonderful @wackerow!!! great job 🔥 

Noticed a small problem with keyboard navigation that I mentioned to you on another channel, but is not a blocker. This is related to nested menus, the user can't close the menu by pressing the `Escape' key. The problem seems to occur when using a Portal.

nloureiro:
@wackerow, we removed the second navigation, but I found that the sidebar should be right after the header without that grey spacing.
Screen Shot 2024-02-09 02 54 36 PM

nloureiro:
@wackerow, something broke here without the top bar, it’s working in production

Screen Shot 2024-02-09 02 59 42 PM

nloureiro:
@wackerow If I click on an option >>> then hover over another one, the first one stays active


Screen Shot 2024-02-09 03 10 24 PM

src/@chakra-ui/semanticTokens.ts
```diff
@@ -96,6 +96,37 @@ const semanticTokens = {
       _light: "rgba(255, 255, 255, 0.80)",
       _dark: "rgba(34, 34, 34, 0.80)",
     },
+
+    // Menu
+    menu: {
```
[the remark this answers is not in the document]
wackerow: Refactored this into `src/hooks/useNavMenuColors.ts`

wackerow:
@nloureiro 

> we removed the second navigation, but I found that the sidebar should be right after the header without that grey spacing.

Done!

> something broke here without the top bar, it’s working in production

As we chatted about, this appears to be an issue unrelated to this PR (currently in production); can address separately

> If I click on an option >>> then hover over another one, the first one stays active

Yeah, I think this is still a bug. Will see if I can resolve; it's stemming from how these menus are handled under-the-hood. While I investigate, I'm curious if you see this as a blocker or not.

nloureiro:
> @nloureiro
> 
> > we removed the second navigation, but I found that the sidebar should be right after the header without that grey spacing.
> 
> Done!
> 
> > something broke here without the top bar, it’s working in production
> 
> As we chatted about, this appears to be an issue unrelated to this PR (currently in production); can address separately
> 
> > If I click on an option >>> then hover over another one, the first one stays active
> 
> Yeah, I think this is still a bug. Will see if I can resolve; it's stemming from how these menus are handled under-the-hood. While I investigate, I'm curious if you see this as a blocker or not.


Quick one: I got this on Firefox, only on Firefox.

Screen Shot 2024-02-14 08 45 07 AM

konopkja:
Screenshot 2024-02-15 at 19 18 45

hey @wackerow  there is a typo in the menu father -> gather :P

konopkja:
Screenshot 2024-02-15 at 19 35 50

@wackerow @nloureiro  because how to guides isnt clickable, we should add How to guides home to the selection as first item, because there are more guides than just these in the menu, but there is no way of getting to the hub

konopkja:
> Screenshot 2024-02-15 at 19 35 50
> 
> @wackerow @nloureiro because how to guides isnt clickable, we should add How to guides home to the selection as first item, because there are more guides than just these in the menu, but there is no way of getting to the hub

I think similar situation with roadmap

nloureiro:
On Figma, I proposed the active to be the same background as the hover but with the base text color.

Now it's the hight contrast.

Screen Shot 2024-02-15 06 34 35 PM
Screen Shot 2024-02-15 06 33 57 PM

konopkja:
Screenshot 2024-02-19 at 16 57 27

we have a little bit of inconsistency with use cases, can we change Desci and Refi to be same format as with main use cases? (acronym first, full name second)

nloureiro:
[bug] @wackerow, I think this spacing is too big. There is an extra margin that leaves the navigation too far out for the glyph

Screen Shot 2024-02-19 12 07 43 AM

nloureiro:
@wackerow, there are two bugs in the menu's interactivity. We can chat about it if it needs more details.

1. I think it would be great if we could open the menu by clicking and by hovering with delay, but it creates an issue: 
 - Accidental closing of the menu when clicking >>> I'm reading the menu and hover one option > it opens with the hover delay > and because it delays a little bit, I click too > accidentally closes the menu.
 
 solution: add the condition: if open && hover { disable the close option }

2. The gap between the nav button and the menu causes it to close without the intention to, but visually, everything looks great, the button size included.

- if the user takes a little bit, more than the delay, to go from the header to the menu, it closes the menu

Solution????: make the button go all the way to the bottom part of the header, with an inside <span> tag that is the size of the actual button to create balance visually, but the interaction keeps the menu open (not sure it makes sense,  open to discuss other solutions )

Screen Shot 2024-02-19 01 52 54 PM

src/components/Nav/Menu/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  return (
+    <Box {...props}>
+      <NavigationMenu.Root
+        dir={direction}
+        orientation="horizontal"
+        delayDuration={500}
+        onValueChange={handleSectionChange}
+      >
+        <NavigationMenu.List asChild>
+          <Flex as="ul" listStyleType="none">
```
pettinarip, line 38: Nice to have: use `UnorderedList` component from chakra. Similar as in the SubMenu

src/components/Nav/Menu/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+    handleSectionChange,
+    isOpen,
+    menuColors,
+    onClose,
+  } = useNavMenu(sections)
+
+  return (
+    <Box {...props}>
+      <NavigationMenu.Root
+        dir={direction}
```
pettinarip, line 32: Nice. So it was mostly what you needed to do in terms of rtl?

src/components/Nav/Menu/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+        delayDuration={500}
+        onValueChange={handleSectionChange}
+      >
+        <NavigationMenu.List asChild>
+          <Flex as="ul" listStyleType="none">
+            {SECTION_LABELS.map((sectionKey) => {
+              const { label, items } = sections[sectionKey]
+              const isActive = activeSection === sectionKey
+              return (
+                <NavigationMenu.Item key={sectionKey} value={label}>
```
pettinarip, line 43: Out of curiosity: could you explain the purpose of the `value`? it is controlling what?

wackerow: I don't think we actually need that here actually... it gets assigned as part of the element ID, so it can be selected easier, but also can be assigned as "defaultValue" for the Root or Sub, so it knows which one should be open at first... we are not assigning this though, and the reason I had assigned a value here is no longer needed. Going to remove.

src/components/Nav/Menu/index.tsx
wackerow: Yea, this seems to have been enough for it to propagate down the tree to the sub-menus.

src/components/Nav/Menu/SubMenu.tsx
```diff
@@ -0,0 +1,162 @@
...
+                          roundedStart: "md",
+                          roundedEnd: "none",
+                          bg: menuColors.lvl[lvl].activeBackground,
+                          me: -PADDING,
+                          pe: PADDING,
+                        },
+                      }}
+                    >
+                      {isLink ? (
+                        <NextLink href={action.href!} passHref legacyBehavior>
```
wackerow, line 109: @pettinarip fyi, in case you were wondering why I took this approach: https://www.radix-ui.com/primitives/docs/components/navigation-menu#with-client-side-routing

pettinarip (approved):
🚀 🚀 🚀 🚀

src/intl/en/common.json
```diff
@@ -145,17 +148,126 @@
...
+  "nav-basics-label": "Basics",
+  "nav-bridges-description": "Web3 has evolved into an ecosystem of primary L1 blockchains and L2 scaling solutions",
+  "nav-builders-home-description": "A builder's manual for Ethereum—by builders, for builders",
+  "nav-builders-home-label": "Builder's home",
+  "nav-contribute-description": "If you want to help, this will guide you",
+  "nav-contribute-label": "Contributing to ethereum.org",
+  "nav-dao-description": "Member-owned communities without centralized authority",
+  "nav-dapps-description": "Explore a rich ecosystem of apps using Ethereum",
+  "nav-defi-description": "A global, open alternative to the traditional financial market",
+  "nav-desci-description": "A global, open alternative to th current scientific system",
```
corwintines, line 164:
```suggestion
  "nav-desci-description": "A global, open alternative to the current scientific system",
```

src/intl/en/common.json
```diff
@@ -145,17 +148,126 @@
...
+  "nav-open-research-label": "Open research",
+  "nav-overview-description": "All things Ethereum education",
+  "nav-overview-label": "Overview",
+  "nav-participate-overview-description": "Overview on how to participate",
   "nav-primary": "Primary",
-  "nft-page": "Non-fungible tokens (NFTs)",
+  "nav-private-description": "Developer resources for private enterprise Ethereum",
+  "nav-quizzes-description": "Find out how well you understand Ethereum and cryptocurrencies",
+  "nav-quizzes-label": "Test your knowledge",
+  "nav-refi-description": "An alternative economic system build on regenerative principles",
```
corwintines, line 222:
```suggestion
  "nav-refi-description": "An alternative economic system built on regenerative principles",
```

corwintines (approved):
Just left a couple comments on some spelling, but this looks good otherwise for me

src/intl/en/common.json
wackerow: Good catch thanks!

PR #12129, opened 2024-02-07
Remove broken links on get-involved page

MiConnell:
## Description

`useweb3` leads to a 404 and I can't find that page anymore on the site and `crypto.jobs` leads to a 500. not sure if that's just a temporary problem though


minimalsm (approved):
Thanks for the PR @MiConnell 💪

PR #12133, opened 2024-02-08
[Windows] fix(rehypeImg): set posix path for filename in `setImagePlaceholders`

TylerAPfledderer:
## Description

In development, posix path is not set in Windows OS for the `rehypeImg` plugin when initializing `FILENAME` in `setImagePlaceholders`. This PR supplies this declaration with `toPosixPath` to ensure forward slashes are applied when running the dev server.

## Related Issue
N/A


TylerAPfledderer:
You know what? This could probably be done at a higher level so that plugins, lower-level methods, etc. aren't concerned about what slashes are rendered in the initial file-path string.

wackerow:
@TylerAPfledderer How we feeling about this one? Ready for a review? Should we switch to draft while finding another solution / have we found one already? Or should we abandon?

TylerAPfledderer:
@wackerow this can be reviewed. My previous comment needed feedback if doing something at a higher level would make more sense or is it fine as is for the effort.

wackerow (approved):
I think this works for the problem at hand for now; can always address a global solution separately if more issues are encountered.

PR #12134, opened 2024-02-08
Default locale prefix in url

### LINKED ISSUE #12207: bug: staking page breaks when adding the /en prefix
### Describe the bug

If we use the `/en` prefix, the page `https://ethereum.org/en/staking` throws a 404. Seems to be working ok with the rest of the languages

### To reproduce

1. Go to `https://ethereum.org/en/staking`
2. Page shoudl throw a 404

### Expected behavior

`https://ethereum.org/en/staking` working ok

### Screenshots

_No response_

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes
- [ ] No

pettinarip:
## Description

Forces the display of the default locale prefix in the url.

Code taken from https://nextjs.org/docs/pages/building-your-application/routing/internationalization#prefixing-the-default-locale

## Related issue
- Fixes #12207


src/lib/utils/translations.ts
```diff
@@ -12,12 +12,17 @@ export const languages: Languages = i18nConfig.reduce((result, config) => {
 export const isLangRightToLeft = (lang: Lang): boolean => {
   const langConfig = i18nConfig.filter((language) => language.code === lang)
 
-  if (!langConfig.length)
-    throw new Error("Language code not found in isLangRightToLeft")
+  if (!langConfig.length) {
+    return false
```
pettinarip, line 16: cc @wackerow I'm removing this since now we have a case where a locale doesn't exist and this will make the builds to fail. I guess we could add another condition to check these faked locales but didn't want to add more logic to this function tbh. LMK what you think and if you prefer to keep the error.

wackerow: Sounds good... realizing we have a handful of locales that are live but not yet included in the `Lang` type ('am', 'be', 'bs', 'hy-am', 'kn', 'ne-np', 'fil', 'tk')... we could probably clean that up separately to enforce this through typing; looks good for this PR though

src/lib/utils/translations.ts
```diff
@@ -12,12 +12,17 @@ export const languages: Languages = i18nConfig.reduce((result, config) => {
...
+  }
 
   return langConfig[0].langDir === "rtl"
 }
 
+export const filterFakedLocales = (locales: string[] | undefined) => {
```
wackerow, line 22: This feels like the output would be a list of faked locales from the name... we're filtering _for_ the real locales here so maybe `filterRealLocales`, `filterForRealLocales` or `filterOutFakedLocales`?

src/middleware.ts
```diff
@@ -0,0 +1,21 @@
...
+    PUBLIC_FILE.test(req.nextUrl.pathname)
+  ) {
+    return
+  }
+
+  if (req.nextUrl.locale === FAKE_LOCALE) {
+    return NextResponse.redirect(
+      new URL(`/${DEFAULT_LOCALE}${req.nextUrl.pathname}`, req.url)
+    )
+  }
```
wackerow, line 20: We seem to have lost the automatic redirect to the users preferred browser language, possibly with these lines here.

This was new with Next.js, but a user with a Spanish preference going to the homepage would get redirected to `/es` but now it's going to `/en`

Should we add logic here to get the users preferred language instead of just using `DEFAULT_LOCALE`?

wackerow: Also, seeing some `307 Temporary Redirect` when redirecting from "default"
```
/about 307 > 200
/about/ 307 > 301 > 200
```

Is there a way to use a 301 here instead?

wackerow (changes requested):
@pettinarip Nice! This is working really well when testing it out, just spotted a few things, in particular the 307 temporary redirects I'm getting.
Otherwise just a few suggestions, and patched what was needed in PageMetadata.

src/components/PageMetadata.tsx
```diff
@@ -44,17 +44,8 @@ const PageMetadata = ({
...
-   * Remove trailing slash
-   * @example ethereum.org/about/ -> ethereum.org/about
-   * @example ethereum.org/pt-br/web3/ -> ethereum.org/pt-br/web3
-   */
-  const url = new URL(
-    join(locale === DEFAULT_LOCALE ? "" : locale!, path),
-    SITE_URL
-  ).href.replace(/\/$/, "")
+  // Set canonical URL w/ language path to avoid duplicate content
+  const url = new URL(join(locale!, path), SITE_URL).href
```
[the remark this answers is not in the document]
pettinarip: Ok, thanks for that.

src/middleware.ts
pettinarip: > Is there a way to use a 301 here instead?

Hmm as far as I can see there is no option to do a permanent redirect https://nextjs.org/docs/pages/api-reference/functions/next-response#redirect

pettinarip: > We seem to have lost the automatic redirect to the users preferred browser language, possibly with these lines here.
> 
> This was new with Next.js, but a user with a Spanish preference going to the homepage would get redirected to `/es` but now it's going to `/en`
> 
> Should we add logic here to get the users preferred language instead of just using `DEFAULT_LOCALE`?

Oh good catch 😕 maybe that is why they were setting `localeDetection: false` in their example https://nextjs.org/docs/pages/building-your-application/routing/internationalization#prefixing-the-default-locale

That sucks...will play a bit to see if there is a way to handle this case.

pettinarip: Ok, I tried several things like forcing the `localeDetection: true` but it didn't work. So, as a workaround, I'm implementing a manual detection of the user's locale inside the middleware. It works well in dev mode, will double check it here once it finish to build.

wackerow: @pettinarip 🤔 Hm... will test more to be 100% sure, but the latest commit df66026 seems to be working the way we want. And yeah, I'm not sure why, but testing the commit with `localeDetection:true` worked locally with `netlify dev` but not in the Netlify preview. I think this looks good if we don't discover any issues with it.

public/_redirects
wackerow, line 1: @pettinarip Do you think we could/should attempt to place a 301 permanent redirect in here somehow to avoid the 307 temporary redirect code?

pettinarip: I'd like to but not sure how to do it. Do you know?

src/middleware.ts
pettinarip: Yeah, not feeling happy about that. Would prefer to use the built in solution 😞

public/_redirects
wackerow: As we discussed, not seeing a clean/obvious way to do this. I don't think this needs to be a blocker, but we should just keep an eye on SEO metrics to make sure this doesn't cause issues.

wackerow (approved):
Lgtm! Working well when testing. Only remaining issue is the 307 temporary redirect, but as mentioned I don't think that should block this. 

Approving but will leave open for the time being to let others test; ideally would like to see this in for next weeks deploy.

wackerow:
@pettinarip Nice! Looks like that `{ status: 301 }` actually got rid of the temporary redirect... getting `301` now.

Side note, the output of `asPath` will change with this PR... it will now include the trailing slash, so anywhere that we do checks against this string we need to double-check that we haven't broken anything.

For example, I'm thinking about #12125 which has all the nav links in `useNav` which currently don't have trailing slashes... the `===` check on this will no longer work when determining if the link is the "active" page or not once these two branches are combined.

@pettinarip Thoughts on if that stuff should block this PR, or if we should update these as a separate task?

wackerow:
For #12125 specifically, I will update these when this is merged... I'll take a look now as well to see if I can find other cases where this will cause an impact.

wackerow:
@pettinarip Opened #12210 to tackle these paths... PR is against this branch.

public/_redirects
```diff
@@ -179,5 +177,3 @@
 /studio /en/deprecated-software/#ethereum-studio 301!
 
 /*/studio /:splat/deprecated-software/#ethereum-studio 301!
-
-/en/stablecoins /stablecoins 301!
```
pettinarip, line 183: Similar issue we have with `/staking` https://github.com/ethereum/ethereum-org-website/issues/12207. Those issues are now fixed with this PR.

pettinarip:
> @pettinarip Opened #12210 to tackle these paths... PR is against this branch.

Great @wackerow. I'll get a look and merge that first, thanks!

wackerow (approved):

Issue #11551, opened 2023-11-01
The link of a voting platform website used during the DAO fork does not work


aslikaya:
### Describe the bug

The link of a voting platform website used during the DAO fork does not work

### To reproduce

1. Go to https://ethereum.org/en/governance/#dao-fork
2. Click on [a voting platform(opens in a new tab)](http://v1.carbonvote.com/)
3. See error

### Expected behavior

If it is a clickable item, it should be directed to a non-dead website. Otherwise, it shouldn't be clickable.

### Would you like to work on this issue?

- [ ] Yes
- [ ] No

PR #12160, opened 2024-02-12
Change the voting platform link

aslikaya:
The previous link was dead. Changed it with the one from the archive which works fine. 

## Description

Change the link for the voting platform
## Related Issue

#11551


minimalsm (approved):
LGTM, thanks for the PR @aslikaya 💪

PR #12177, opened 2024-02-13
Web3 Antivirus. Add information about Web3 security tools. [Fixes #12176]

### LINKED ISSUE #12176: Web3 Antivirus. Add information about Web3 security tools to the Wallet Security block.

There is no information about security tools such as Web3 Antivirus on the page https://ethereum.org/en/security#wallet-security, which would be very useful.

Add information about security tools to the Wallet Security block.

-

### Additional context

_No response_

- [X] Yes
- [ ] No

web3-antivirus:
## Description

This pull request adds information about security tools such as Web3 Antivirus to the Wallet Security block.

## Related Issue

#12176

### DIFF
--- a/public/content/security/index.md
+++ b/public/content/security/index.md
@@ -161,6 +161,19 @@ Many Ethereum wallets offer limits protection to safeguard against accounts bein
 
 [How to revoke smart contract access to your crypto funds](/guides/how-to-revoke-token-access/)
 
+### Use reliable Web3 security tools {#use-security-tools}
+
+Risks associated with transactions and signature requests are often unidentifiable by the naked eye and by signing them, you risk compromising the safety of your tokens. For example, allow contract owners to transfer or burn tokens as well as grant approvals without your consent, potentially draining your wallet. Specialized security solutions can offer significant protection.
+
+Among such tools, Web3 Antivirus (W3A) Snap is an official plugin for MetaMask designed to simulate transactions and warn you if it finds signs of scams. It detects over 20 risks, including phishing, poisoning attacks, wallet drainers, scam tokens, fake addresses, and more.
+
+It is open-source and passed an audit by Veridise.
+
+#### Try a security tool: {#try-security-tool}
+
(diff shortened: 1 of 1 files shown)


minimalsm (commented):
Thanks for the PR @web3-antivirus!

Not opposed to including this on this page, but right now this reads like a product specific ad. We should rework this to a slightly higher level of abstraction so that it covers other tools (like Rabby's inbuilt features that perform a similar function and others)

Any thoughts @corwintines @konopkja @wackerow

web3-antivirus:
Thank you for your feedback, @minimalsm!

Indeed, basic security features, which can detect common risks like scam tokens and ensure transaction address legitimacy, are essential. These are integrated within some of the popular wallets such as Rabby, MetaMask, and Phantom, providing users with a fundamental layer of security.

However, as threats evolve, there is an increasing need for specialized security solutions that offer advanced protection. Tools like Web3 Antivirus complement these foundational features by providing deeper security analyses and uncovering more complex risks, such as smart contract vulnerabilities, scam tokens, poisoning attacks, and wallet drainers.

Agreed, it’s crucial for users to choose the level of protection that best suits their needs, balancing convenience with the degree of security required. A more balanced overview, highlighting the variety of tools and approaches in the ecosystem, encourages users to explore beyond built-in wallet features for enhanced protection when necessary.

nhsz:
@minimalsm Seems a bit weird to me to include this tool here as we're not listing other products in this page, but more general security tips

wackerow:
I like the idea of expanding options for people to learn about and access security tools to help them stay safe... Agree though that this page may be a bit tricky. I see why you mentioned it @web3-antivirus, but as was noted, we don't really list any products on this page.

In general and where possible it's preferred to offer options to users instead of just one product (striking a balance, since too many products becomes overwhelming). 

Just floating the idea, but do we think there are enough "wallet security add-ons" out there to add information about this to the wallets page or find-wallet page?

wackerow:
Closing this out for now; if we want to approach this I'd suggest a new issue regarding how we'd like to handle the software category of "wallet security add-ons" first, and we can take it from there.

PR #12201, opened 2024-02-16
[WIP] Custom useRouter hook - fix hydration warnings

wackerow:
use English as initial locale until client loaded to prevent hydration errors
see: https://traviswimer.com/blog/easily-fix-react-hydration-errors

Adds two new hooks
- `useHydrated.ts`
- `useRouter.ts`

`useHydrated` simply returns `false` until the page loads on the client, when it flips to `true`.

`useRouter` effectively loads and returns the router object from `next/router` `useRouter()`, but overrides `locale` to `DEFAULT_LOCALE` (English  / `en`) until hydrated, then flips to the users locale.

All instances of `import { useRouter } from "next/router"` replaced with `import { useRouter } from "@/hooks/useRouter"` (linter run to auto-sort)


wackerow:
Proof-of-concept only; closing

PR #12204, opened 2024-02-16
Fix fallback for locale display names

wackerow:
## Description
- Removes all empty strings from the `intl` folder (will fallback to English string)
- Improve check inside `useLanguagePicker` to check if we have an internal translation available or not
  If the output of `t()` matches the English name _or_ the i18n key, then this string isn't available internally and Intl.DisplayNames used as fallback to display the locale name.

### DIFF
--- a/src/components/LanguagePicker/useLanguagePicker.tsx
+++ b/src/components/LanguagePicker/useLanguagePicker.tsx
@@ -71,8 +71,14 @@ export const useLanguagePicker = (
       const fallbackSource =
         intlSource !== localeOption ? intlSource : englishName
       const i18nKey = "language-" + localeOption.toLowerCase()
-      const i18nSource = t(i18nKey)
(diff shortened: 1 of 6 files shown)


pettinarip (approved):
Lgtm! it seems to fix the issue.

However, I'd strongly suggest removing the `throw` errors we have inside the `useEffect`, since all of them are going to break the client experience. Instead, we can `console.error` them.

wackerow:
image

Okay... @pettinarip note that without the guard for `(!sourceName || !targetName)`, we end up with the possibility of `undefined` here for these two items... If you're okay with it, let's get this in and I can look into that separately.

pettinarip:
@wackerow we should assign something to those vars
```ts
if (!sourceName) {
  console.error(...)
  sourceName = ""
}

if (!targetName) {
  console.error(...)
  targetName = ""
}
```

PR #12208, opened 2024-02-16
refactor: useLanguagePicker warning and fallback

wackerow:
## Description
- Update `useLanguagePicker.tsx` to remove error throwing when a source or target locale display name is not found
- Adds `console.warn` instead, and adds an empty string fallback to ensure values are never `undefined`

## Related Issue
- https://github.com/ethereum/ethereum-org-website/pull/12204#pullrequestreview-1885619513

### DIFF
--- a/src/components/LanguagePicker/useLanguagePicker.tsx
+++ b/src/components/LanguagePicker/useLanguagePicker.tsx
@@ -108,8 +108,23 @@ export const useLanguagePicker = (
               (dataItem!.words.approved / dataItem!.words.total) * 100
             ) || 0
 
-      if (progressData.length === 0) {
-        console.warn("Missing translation progress data; check GitHub action")
+      const isBrowserDefault = browserLocales.includes(localeOption)
+
+      const returnData: Partial<LocaleDisplayInfo> = {
+        localeOption,
+        sourceName: sourceName ?? localeOption,
+        targetName: targetName ?? localeOption,
+        englishName,
+        isBrowserDefault,
+      }
+
+      if (progressData.length < 1) {
+        console.warn(`Missing translation progress data; check GitHub action`)
+        return {
+          ...returnData,
+          approvalProgress: 0,
+          wordsApproved: 0,
+        } as LocaleDisplayInfo
       }
 
       const totalWords = progressData[0].words.total
@@ -119,17 +134,11 @@ export const useLanguagePicker = (
           ? totalWords || 0
           : dataItem?.words.approved || 0
 
-      const isBrowserDefault = browserLocales.includes(localeOption)
-
       return {
-        localeOption,
+        ...returnData,
         approvalProgress,
-        sourceName: sourceName ?? "",
-        targetName: targetName ?? "",
-        englishName,
         wordsApproved,
-        isBrowserDefault,
-      }
+      } as LocaleDisplayInfo
     }
 
     const displayNames: LocaleDisplayInfo[] =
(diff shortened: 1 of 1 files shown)


src/components/LanguagePicker/useLanguagePicker.tsx
```diff
@@ -106,10 +108,9 @@ export const useLanguagePicker = (
               (dataItem!.words.approved / dataItem!.words.total) * 100
             ) || 0
 
-      if (progressData.length === 0)
-        throw new Error(
-          "Missing translation progress data; check GitHub action"
-        )
+      if (progressData.length === 0) {
+        console.warn("Missing translation progress data; check GitHub action")
+      }
```
wackerow, line 113: Looking at this one again, I think we can probably keep this error. It should never throw as long as there is translation progress data available... if not, we should let the build break. cc: @pettinarip Curious your thoughts

src/components/LanguagePicker/useLanguagePicker.tsx
```diff
@@ -123,8 +124,8 @@ export const useLanguagePicker = (
       return {
         localeOption,
         approvalProgress,
-        sourceName,
-        targetName,
+        sourceName: sourceName ?? "",
+        targetName: targetName ?? "",
```
wackerow, line 128: @pettinarip Seeing your comment [here](https://github.com/ethereum/ethereum-org-website/pull/12204#issuecomment-1948963646)... I went with this approach to avoid using mutable variables, sticking with `const`... lemme know if you disagree

src/components/LanguagePicker/useLanguagePicker.tsx
pettinarip: I've to disagree on that. This condition is inside a `useEffect`. It will never be checked at build time/server side. If it is ever checked, then it will be on client side, creating bad ux since it will make the app crash (unless we are catching that error and failing gracefully).

However, I hear you, it would be nice to have that type of check at build time.

And the change to use `console.warn` is ok I guess but its not doing anything. Shouldn't  we return something else here? given that in the next line we are accessing it `progressData[0].words.total`.

src/components/LanguagePicker/useLanguagePicker.tsx
pettinarip: Wondering if we should fallback to default locale or something else. Not sure if `"'` isn't going to break something else in the chain tbh. If not, it ok.

pettinarip (commented):
@wackerow left some questions. Maybe there are potential scenarios we are not considering.

src/components/LanguagePicker/useLanguagePicker.tsx
wackerow: Okay yeah, that makes sense, thanks

> Shouldn't we return something else here? given that in the next line we are accessing it progressData[0].words.total.

I think we should... I can update this to return early in this case, and pass everything but the `approvalProgress` and `wordsApproved`. We can also use `localeOption` (`Lang` type) as the fallback for the source or target names... this means a language that failed this for whatever reason would just show the language code instead of an empty string.

Will push a patch

src/components/LanguagePicker/useLanguagePicker.tsx
wackerow: It wouldn't break anything, but it would just render without a name... I switched this to fallback to its language code.

pettinarip (approved):
GJ!

Issue #12019, opened 2024-01-25
Investigate & replace `luxon` with native solutions


pettinarip:
We’re importing [luxon](https://bundlephobia.com/package/luxon@3.4.3)[ package](https://bundlephobia.com/package/luxon@3.4.3) for very few operations (date manipulation) that could potentially be replaced with a native implementation using [JS native Date API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) or [DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat). Investigate the feasibility of this.

PR #12209, opened 2024-02-16
refactor: remove luxon package, replace with JS Intl library

wackerow:
## Description
- Updates `getLocaleTimestamp.ts` inside `src/lib/utils/time.ts` to use `Intl.DateTimeFormat` instead of luxon tooling. 
- Adds an optional `options` prop here to override display options as-needed
- Remove usage of luxon inside `CommunityEvents/index.tsx`, replacing with updated usage of `getLocaleTimestamp`
- `yarn remove luxon @types/luxon`

## Related Issue
https://www.notion.so/efdn/Review-if-we-can-replace-some-packages-with-native-implementation-1875f79b99f04555aa60bb5f7b14e8c0?pvs=4

Fixes https://github.com/ethereum/ethereum-org-website/issues/12019

## Summary by CodeRabbit

- **Refactor**
	- Improved date handling across the app for better localization and efficiency.
- **New Features**
	- Updated display logic for last updated and deploy dates to enhance user experience.
- **Chores**
	- Removed unnecessary dependencies and streamlined date formatting.


nhsz:
@wackerow Excellent! Seems we also need to remove `@types/luxon`

wackerow:
Good catch; removed the types package... thanks @nhsz!

wackerow:
Getting a strange result where the lastDeployDate in the Footer is causing the build to fail with:

```
Export encountered errors on following paths:
        /_error: /500
       ...
```

Added a temporary debugging patch to check for paths starting with `/500` and skipping the rendering of the deploy date for those cases. Want to see if this builds okay on Netlify (working locally), but I don't really see this as a final solution though; open to thoughts.

pettinarip (changes requested):
Nice job! All looks fine @wackerow except this one in the tutorials page
image

pettinarip (approved):
Fixed! merging @wackerow 💪🏼

PR #12210, opened 2024-02-16
Update paths for trailingSlash rule

wackerow:
**Target branch: default-locale-prefix (#12134)**

## Description
- Update paths for new `trailingSlash=true` rule
- Adds `addSlashes` util to `url.ts` which forces a leading and trailing slash.
- Updates `translations.ts` to have leading and trailing slashes on all paths. Runs the path through `addSlashes` first to ensure all paths match this pattern\*.
  - \*Alternatively, we can update these in each `getStaticProps` to follow that pattern

## Related Issue
- https://github.com/ethereum/ethereum-org-website/pull/12134#issuecomment-1948881490


wackerow:
^ cc: @pettinarip

pettinarip (approved):
LGTM, thanks @wackerow

PR #12226, opened 2024-02-18
Keyboard nav: nav menu number shortcuts

wackerow:
## Description
- Adds keyboard shortcuts for numbers 1-5 to move focus to corresponding nav menu section button; enter to open
- Guards against cases where cursor is in a `textarea`, `input` or `select`

## Summary by CodeRabbit

- **New Features**
	- Enhanced keyboard navigation in the main menu by focusing on sections when number keys are pressed.
- **Refactor**
	- Improved the structure and organization of constants related to navigation and styling in the codebase.


src/components/Nav/Menu/useNavMenu.ts
```diff
@@ -11,6 +14,25 @@ export const useNavMenu = (sections: NavSections) => {
...
+  useEventListener("keydown", (event) => {
+    if (!document || !event.key.match(/[1-9]/)) return
+    if (event.target instanceof HTMLInputElement) return
+    if (event.target instanceof HTMLTextAreaElement) return
+    if (event.target instanceof HTMLSelectElement) return
+
+    const sectionIdx = parseInt(event.key) - 1
+    if (sectionIdx >= SECTION_LABELS.length) return
+
+    const button = document.querySelector(
```
pettinarip, line 27: One observation: this is not a very react way to do this. We already know all the buttons that are in the menu and we could easily have the reference of each one. This would make all the checking you have to do with this approach a lot easier.

Not a blocker as that works fine too. Just to see what you think, maybe I'm missing something important.

pettinarip (approved):
Handy shortcut, works really well 👍🏼

src/components/Nav/Menu/useNavMenu.ts
wackerow: Yes, I agree with the overall sentiment... I can look at this again, but I did try this is the initial approach... I forget exactly why at the moment, but I hit some issues passing the refs around and getting the correct ones to trigger correctly... I recall it ended up having a lot more code than this, since this approach just takes advantage of the `nth-of-type` selector. Happy to check this out again, but wanted to just drop a reply.

PR #12234, opened 2024-02-19
Fix OG page images

wackerow:
## Description
**refactor: extract `getOgImage` util**
-  Added to `@/lib/utils/metadata.ts`, including refactor of the default OG images into a data list `imageForSlug` for easy updating

**feat: add/update page OG images**
- Fixes pages passing an imported Image component to PageMetadata using `.src` to pass public path string directly
- Adds `image` prop to `PageMetadata` for pages that have hero images available to be used as OG image

**fix: metadata string namespaces**
- Few namespace issues spotted that were causing some metadata strings to not render corretly


src/pages/community.tsx
```diff
@@ -226,6 +226,7 @@ const CommunityPage = () => {
       <PageMetadata
         title={t("page-community-meta-title")}
         description={t("page-community-meta-description")}
+        image="/heroes/community-hero.png"
```
nhsz, line 229: seems we already have a default image defined in `getOgImage` for `/community`

src/pages/dapps.tsx
```diff
@@ -1334,8 +1333,8 @@ const DappsPage = () => {
     <Page>
       <PageMetadata
         title={t("common:decentralized-applications-dapps")}
-        description={t("common:page-dapps-desc")}
-        image={ogImage.src}
+        description={t("page-dapps-desc")}
+        image="/doge-computer.png"
```
nhsz, line 1337: why not define this as default in `imageForSlug`?

src/pages/developers/index.tsx
```diff
@@ -217,6 +217,7 @@ const DevelopersPage = () => {
       <PageMetadata
         title={t("page-developers-index:page-developer-meta-title")}
         description={t("page-developers-index:page-developers-meta-desc")}
+        image="/heroes/developers-hub-hero.jpg"
```
nhsz, line 220: this seems to be redundant as it's already defined in `getOgImage`

src/pages/quizzes.tsx
```diff
@@ -77,11 +76,12 @@ const QuizzesHubPage: NextPage<
       <PageMetadata
         title={t("quizzes-title")}
         description={t("quizzes-subtitle")}
+        image="/heroes/quizzes-hub-hero.png"
```
nhsz, line 79: can be defined in `getOgImage` as the others?

src/pages/community.tsx
wackerow: Yeah, we could get rid of these where we already have the correct image defined in `imageForSlug`. Similar to other comment, tried to place these here for explicitness for each page, but I'm not married to that approach, both should work

nhsz (changes requested):
@wackerow Thanks. I think one asset can be safely removed. I also left some other minor comments

src/pages/developers/index.tsx
nhsz: I'm ok with this approach

corwintines (approved):

nhsz (approved):
LGTM

PR #12236, opened 2024-02-20
[WIP] Add new glossary tooltips

pettinarip:
## Description

List of pages with new tooltips (in progress):
- /get-eth
- /dapps


konopkja:
Screenshot 2024-02-22 at 20 42 11

hi Pablo, noticed few issues with the tooltip:
1. the background seems transparent (should not be)
2. it sticks right at the start of the page, the positioning is offf

konopkja:
Screenshot 2024-02-22 at 20 46 47
 
positioning and z-index issues

pettinarip:
@konopkja nice catches. I'll create separate issues for those in order to keep this PR in the scope.

src/components/Glossary/GlossaryTooltip/index.tsx
```diff
@@ -13,7 +13,15 @@ const GlossaryTooltip = ({ children, termKey }: GlossaryTooltipProps) => {
   const isLargeScreen = useBreakpointValue({ base: false, lg: true })
 
   return isLargeScreen ? (
```
minimalsm, line 15: We use desktop menu at 768px wide, but tooltip is only available at 992px wide. Is this intentional @pettinarip @konopkja?

src/components/Glossary/GlossaryTooltip/index.tsx
```diff
@@ -13,7 +13,15 @@ const GlossaryTooltip = ({ children, termKey }: GlossaryTooltipProps) => {
...
-    <Tooltip content={<GlossaryDefinition term={termKey} size="sm" />}>
+    <Tooltip
```
minimalsm, line 16: Probably for the backlog (solved by Chakra?), but collision handling isn't great. Don't think it's blocking as it's pretty standard behaviour for this type of feature.

Screenshot 2024-03-05 at 08 48 23

src/components/Glossary/GlossaryTooltip/index.tsx
konopkja: no, i think the original component was just built for desktop, but we want it make it available on on all screen resolutions.

I believe there is a PR to migrate the tooltip to chakra component first tho. Maybe pablo can shed more light on it.

src/components/Glossary/GlossaryTooltip/index.tsx
konopkja: we have discussed this one with pablo already, chakra tooltip should be fixing htis

src/components/Glossary/GlossaryTooltip/index.tsx
pettinarip: Correct. We will enable the tooltip for all the resolutions once we migrate our current component to Chakra (#10974).

src/components/Glossary/GlossaryTooltip/index.tsx
pettinarip: Yes, this is going to be fixed in the tooltip chakra migration #10974

src/components/Glossary/GlossaryDefinition/index.tsx
```diff
@@ -17,12 +32,20 @@ const GlossaryDefinition = ({ term, size = "md" }: GlossaryDefinitionProps) => {
...
 
   return (
-    <Box>
+    <Box textAlign="left">
       <OldHeading as="h3" lineHeight={1.4} id={term} {...headingStyles}>
-        <Translation id={"glossary:" + term + "-term"} />
+        <Translation
+          id={term + "-term"}
+          options={options}
+          transform={components}
```
wackerow, line 40: 🔥

src/components/Translation.tsx
```diff
@@ -1,35 +1,38 @@
...
 
   const { t } = useTranslation(requiredNamespaces)
   const translatedText = t(id, options)
 
   // Use `htmr` to parse html content in the translation text
-  // @ts-ignore
-  return htmr(translatedText, { transform })
+  return htmr(translatedText, {
+    // @ts-ignore
+    transform: { ...defaultTransform, ...transform },
```
wackerow, line 34:
```suggestion
    transform: { ...defaultTransform, ...transform } as HtmrOptions["transform"],
```

Would this work to remove the ts-ignore?

wackerow (commented):
image

Not sure if this is in scope here, but noting the opacity makes this a little hard to read (from /get-eth page).

image

I don't think I'm seeing this same opacity issue on all cases though... this is the /dapps page and the text behind it doesn't bleed through like it does above.

---

Overall working well.. just left a few comments but otherwise think this is close to pulling in.

src/components/Glossary/GlossaryDefinition/index.tsx
```diff
@@ -1,14 +1,29 @@
...
 import Translation from "@/components/Translation"
 
 interface GlossaryDefinitionProps {
   term: string
   size?: "md" | "sm"
+  options?: ComponentProps<typeof Translation>["options"]
 }
 
-const GlossaryDefinition = ({ term, size = "md" }: GlossaryDefinitionProps) => {
+const DEFAULT_NS = "glossary"
```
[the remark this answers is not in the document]
pettinarip: Sure, can do that!

src/components/Glossary/GlossaryDefinition/index.tsx
```diff
@@ -17,12 +32,20 @@ const GlossaryDefinition = ({ term, size = "md" }: GlossaryDefinitionProps) => {
   const textStyles = size === "sm" ? { mb: 0 } : {}
 
   return (
-    <Box>
+    <Box textAlign="left">
```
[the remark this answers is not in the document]
pettinarip: Ah! nice catch, ty

src/components/Translation.tsx
pettinarip: I revisited this and found the root problem (check [eb02aa2](https://github.com/ethereum/ethereum-org-website/pull/12236/commits/eb02aa2eaa260df75aaa1199a28b1a5235bf665c)). Basically, `htmr` was expecting some props from the Link component to be optional.

pettinarip:
> image
> Not sure if this is in scope here, but noting the opacity makes this a little hard to read (from /get-eth page).
> 
> image
> I don't think I'm seeing this same opacity issue on all cases though... this is the /dapps page and the text behind it doesn't bleed through like it does above.
> 
> Overall working well.. just left a few comments but otherwise think this is close to pulling in.

Good observation @wackerow. I'll take note of those styles differences since I'm going to tackle them once we finish the migration of the Tooltip component in https://github.com/ethereum/ethereum-org-website/pull/10974. That migration should fix most of the style issues but will double-check these ones as well.

wackerow (approved):
lgtm! Nice job @pettinarip

PR #12251, opened 2024-02-21
Nextjs configs

AkimBW:
Add various nextjs configuration

## Description

This changes are related to some configuration that can be highlighted for:
- Nextjs Image optimizer (usage of sharp) → Edit: package seems to have some issue at build time. need to add the ignore engine for yarn.
- Preventing iframes
- Preventing clickjackings or XSS attack

## Related Issue

No issue created for this PR

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -25,6 +25,7 @@ module.exports = (phase, { defaultConfig }) => {
   let nextConfig = {
     ...defaultConfig,
     reactStrictMode: true,
+    poweredByHeader: false,
     webpack: (config) => {
       config.module.rules.push({
         test: /\.ya?ml$/,
--- a/next.config.js
+++ b/next.config.js
@@ -44,6 +44,7 @@ module.exports = (phase, { defaultConfig }) => {
           source: "/(.*?)",
           headers: [
             {
+              // See https://nextjs.org/docs/pages/api-reference/next-config-js/headers#x-frame-options
               key: "X-Frame-Options",
               value: "deny",
             },
--- a/next.config.js
+++ b/next.config.js
@@ -44,6 +44,10 @@ module.exports = (phase, { defaultConfig }) => {
           source: "/(.*?)",
           headers: [
             {
+              key: "X-Frame-Options",
+              value: "deny",
+            },
+            {
               key: "Content-Security-Policy",
               value: ContentSecurityPolicy.replace(/\s{2,}/g, " ").trim(),
             },
--- a/next.config.js
+++ b/next.config.js
@@ -4,6 +4,7 @@ const { i18n } = require("./next-i18next.config")
 
 const LIMIT_CPUS = Number(process.env.LIMIT_CPUS || 2)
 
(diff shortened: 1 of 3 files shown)


next.config.js
```diff
@@ -33,6 +38,23 @@ module.exports = (phase, { defaultConfig }) => {
 
       return config
     },
+    async headers() {
+      return [
+        {
+          source: "/(.*?)",
+          headers: [
+            {
+              key: "X-Frame-Options",
```
nhsz, line 47:
```suggestion
              // See https://nextjs.org/docs/pages/api-reference/next-config-js/headers#x-frame-options
              key: "X-Frame-Options",
```

next.config.js
```diff
@@ -4,6 +4,10 @@ const { i18n } = require("./next-i18next.config")
 
 const LIMIT_CPUS = Number(process.env.LIMIT_CPUS || 2)
 
+const ContentSecurityPolicy = `
```
nhsz, line 7:
```suggestion
// See https://nextjs.org/docs/app/building-your-application/configuring/content-security-policy
const ContentSecurityPolicy = `
```

next.config.js
```diff
@@ -33,6 +39,24 @@ module.exports = (phase, { defaultConfig }) => {
 
       return config
     },
+    async headers() {
+      return [
+        {
+          source: "/(.*?)",
```
nhsz, line 45: not sure this would work with i18n (https://nextjs.org/docs/pages/api-reference/next-config-js/headers#headers-with-i18n-support)

next.config.js
```diff
@@ -33,6 +39,24 @@ module.exports = (phase, { defaultConfig }) => {
 
       return config
     },
+    async headers() {
+      return [
```
nhsz, line 43: can we move this config to a `headersConfig` variable at the top?

nhsz (commented):
Thanks, I left a couple of comments

next.config.js
AkimBW: It should be fine.

From Nextjs documentation
>  // this gets converted to /(en|fr|de)/(.*) so will not match the top-level
> // `/` or `/fr` routes like /:path* would

There is no need to match / or /:lang because it gets automatically redirected to index (/:lang/) who will match the pattern.

next.config.js
AkimBW: Sure. I'll do it for the end of the day (UTC+2)

wackerow:
Circling through... Is this ready for another review @AkimBW?

pettinarip (changes requested):
@AkimBW nice proposal, thanks for this PR!

While testing, it didn't work with the following paths:
- root path https://deploy-preview-12251--ethereumorg.netlify.app/
- locale prefix path https://deploy-preview-12251--ethereumorg.netlify.app/en/

It seems that the regex only matches all paths except root paths. For example, https://deploy-preview-12251--ethereumorg.netlify.app/en/assets works fine, its blocked when used in an iframe.

wackerow:
Closing out from inactivity

Issue #12252, opened 2024-02-21
Empty and missing H1 headings


minimalsm:
### Describe the bug

There are 349 instances of empty or missing H1 tags on the website, which negatively effects SEO

[See report on SEO Review Google Sheet](https://docs.google.com/spreadsheets/d/1SH7fWd0ggcP37g-k8F1duvkn99PbKmf4JRvu2IA_9yo/edit#gid=1459175556)

### To reproduce

.

### Expected behavior

An 

### Would you like to work on this issue?

- [X] Yes
- [ ] No

corwintines:
Nice one @minimalsm 

I have brought in all the open PR's related to this. I don't have access to that report at the moment, so I will let you close this if you think it is done.

PR #12258, opened 2024-02-21
Add missing h1s to markdown

minimalsm:
## Description

- Changes first H2 or adds new H1s to markdown files missing first level headings
- Adds h1 heading to tutorials page

## Related Issue

#12252


wackerow (approved):
Nice, thanks @minimalsm! Great to see these patched...

Couple things I'll note... for now I think this works, but we should really consider just updating the templates (that don't already do this) to use the `title` front matter property to render an h1. This would remove the inconsistency between different templates, and make it a standardized approach to our markdown files (no h1's inside markdown across-the-board)

Only other things to note, we don't have auto-linking on the h1's since they're always at the top, so the custom header ID on these is not really useful. (See [preview of code-of-conduct page](https://deploy-preview-12258--ethereumorg.netlify.app/community/code-of-conduct)... notice hovering the h1 has no link icon, unlike the others). We also do not include the h1 in the table of contents. 

Doesn't necessarily hurt leaving them in, but we may want to consider removing to trim out what isn't used. That doesn't need to block this PR though imo, approving.

minimalsm:
Thanks @wackerow 

> _Couple things I'll note... for now I think this works, but we should really consider just updating the templates (that don't already do this) to use the title front matter property to render an h1. This would remove the inconsistency between different templates, and make it a standardized approach to our markdown files (no h1's inside markdown across-the-board)_

Yep, I think we discussed this before. While I agree in principle, but I'd rather not tweak anything SEO related (e.g. web3 page uses a different meta-title than its h1) until we clean up some of the current issues we're having and can measure its effect.

> _Only other things to note, we don't have auto-linking on the h1's since they're always at the top, so the custom header ID on these is not really useful. (See [preview of code-of-conduct page](https://deploy-preview-12258--ethereumorg.netlify.app/community/code-of-conduct)... notice hovering the h1 has no link icon, unlike the others). We also do not include the h1 in the table of contents._

Oh, aren't these required? I think we get warnings in the build if we don't include them 🤔

corwintines (approved):

PR #11979, opened 2024-01-18
Italian import

corwintines:
## Description

- Italian import (Homepage, Essential pages, Explore, Use Ethereum, Use cases, Staking, Learn pages, Learn hub & guides, Roadmap, Community, Foundational docs, Foundational docs - nodes and clients, Foundational docs - proof-of-stake, Foundational docs - proof-of-work, Docs - tech stack pages, Learn pages 2, Advanced docs, Advanced docs - scaling, Research documentation, Misc, Contributing, Tutorials I, Tutorials II)

## Summary by CodeRabbit

- **New Features**
	- Added new project and reading recommendations in decentralized identity.
	- Updated framework information with new tools and improvements.
	- Introduced detailed descriptions for Ethereum accounts and transactions in Italian.
	- Enhanced clarity on NFTs, their uniqueness, ownership, and use cases.
	- Refined the foundational principles of regenerative economics in the ReFi context.
	- Added details about validator withdrawals on the Launchpad page for staking.
- **Bug Fixes**
	- Corrected hyperlinks and refined descriptions across various sections.
	- Fixed formatting errors and improved table presentations.
	- Adjusted text for clarity and accuracy in guides and tutorials.
	- Updated Discord invite links and external URLs for better accessibility.
- **Documentation**
	- Updated grant programs, Ethereum Improvement Proposals (EIPs) links, and community support resources.
	- Enhanced descriptions for proof-of-stake and proof-of-work mechanisms.
	- Provided updated information on Ethereum's energy consumption.
	- Clarified the roadmap and the collaborative process behind Ethereum's development.
	- Emphasized community involvement in maintaining project lists and contributions.

PR #12262, opened 2024-02-21
fix: hydration warnings on glossary

wackerow:
## Description
While cleaning up #11979, encountered a hydration warning that was breaking the `it/glossary` page, which traces back to `GlossaryDefinition`.

This refactors the `GlossaryDefinition` component to use `useTranslation("glossary")` instead of `<Translation />`, which locally appears to resolve the hydration bug.


wackerow:
My bad, just built locally and realized we have html usage in these strings, which requires `htmr` and the `Translation` component... going to close this back out and investigate separately. 

image

PR #12263, opened 2024-02-21
fix: hydration warnings on glossary

wackerow:
wrap definition in span, preventing errors when attempting to nest certain elements such as `ul`, which are currently contained in some definitions.


pettinarip (approved):
LGTM, i've fixed some code conflicts but should be good to go when the build is done.

PR #12266, opened 2024-02-22
build(deps): bump Next to v14 with Netlify build Workaround

TylerAPfledderer:
## Description

Upgrades NextJs to v14.

Two main changes added to avoid problems with the build:
- Wrapped `process.cwd()` in a function, otherwise the Netlify runtime plugin would _for some reason_ bundle the entire project folder, causing the deploy to fail due to size limits (250mb max).
- Ignored deps and `public` folder from the deployed functions by using more specific syntax: `[functions.{function-name}]` instead of just `[functions]`.


pettinarip (approved):
🚀

netlify.toml
```diff
@@ -36,9 +36,21 @@
...
+    ]
+
+  [functions.___netlify-handler]
+    included_files = [
+      "./src/intl/**/*",
+      "!./public/**/*",
+      "node_modules/next/dist/server/future/route-modules/pages/vendored/contexts/router-context*",
+      "node_modules/next/dist/server/future/route-modules/pages/vendored/contexts/amp-context*",
+      "node_modules/next/dist/server/future/route-modules/pages/vendored/contexts/head-manager-context*",
+    ]
```
pettinarip, line 56: https://github.com/netlify/next-runtime/issues/2385#issuecomment-1911785301

TylerAPfledderer:
@pettinarip Great! This is kinda insane to have to do all of that configuring. 😅

pettinarip:
> @pettinarip Great! This is kinda insane to have to do all of that configuring. 😅

Yeah, right? all the nuances you have to deal with 🤷🏼 the cost of living on the edge :P

PR #12272, opened 2024-02-22
Hide contributor's profile link if the url doesn't exist

### LINKED ISSUE #12269: 404 links on developer docs
### Describe the bug

On every developer docs page (I believe), including all languages, we have a random 404 link where we're pointing to a target url of `https://ethereum.org/[...slug]` with the anchor text for this url being `@`

This is not visible on the website, but flags up using SEO tools

[CSV of all 404s caused by this issue](https://github.com/ethereum/ethereum-org-website/files/14376131/ethereum_20-feb-2024_page-has-links-to-broken_2024-02-22_12-55-01.csv)

### Expected behavior

We should have no links that 404 except the 404 page :) 

### Screenshots

Screenshot 2024-02-22 at 12 57 24

### Additional context

@pettinarip thinks this is probably related to the use of `pathname` somewhere, since the dynamic params are included in there and not the final path

- [ ] Yes
- [X] No

pettinarip:
## Description

Hides the contributor's profile link if the url doesn't exist.

## Related Issue

Fixes #12269

### DIFF
--- a/src/components/FileContributors.tsx
+++ b/src/components/FileContributors.tsx
@@ -146,7 +146,7 @@ const FileContributors = ({
           <Skeleton isLoaded={!loading}>
             <Text m={0} color="text200">
               <Translation id="last-edit" />:{" "}
-              {lastContributor.user && (
+              {lastContributor.user?.url && (
                 <InlineLink to={lastContributor.user.url}>
                   @{lastContributor.user.login}
                 </InlineLink>


wackerow (approved):
Nice, this lgtm... 

Tried testing on a few pages including 
- https://deploy-preview-12272--ethereumorg.netlify.app/ka/developers/docs/consensus-mechanisms/pos/keys/
- (Compare to) https://ethereum.org/ka/developers/docs/consensus-mechanisms/pos/keys

Only question I have here... should we consider any case where there is a `user.login` but no `user.url`? In that case we could potentially just render the name without a link... don't think that needs to block this PR though.

nhsz (approved):

minimalsm (approved):

PR #12276, opened 2024-02-22
feat: add matomo event to lang picker input

wackerow:
## Description
- Add matomo event when user focuses in language picker filter input
- De-duped to only send one event per session, per language (resets if user follows a language result and switches languages)

cc: @konopkja


nhsz (approved):
LGTM

PR #12278, opened 2024-02-23
chore: use `href` instead of deprecated to `prop` in InlineLink usage

enjoyooor:
Dupe of #12218 (accidentally closed), prefer to contribute using my personal account

Noticed that the `to` prop in `InlineLink` component is deprecated, so updating all usages to reflect this.

## Summary by CodeRabbit

- **New Features**
	- Enhanced navigation experience by updating internal link handling to standard web link behavior across various components and pages.

- **Bug Fixes**
	- Corrected link destinations by switching from custom `to` props to standard `href` attributes, ensuring proper redirection and improving user navigation on the site.

- **Refactor**
	- Streamlined linking mechanism throughout the application for consistency and improved maintainability.

- **Chores**
	- Removed unnecessary imports and updated the codebase to align with best practices for link handling.


enjoyooor:
cc @nhsz would love your 👀

minimalsm (approved):
LGTM, thanks @enjoyooor!

Just waiting on second review as it touches a lot.

minimalsm:
@all-contributors please add @enjoyooor for code

src/pages/dapps.tsx
```diff
@@ -107,8 +107,10 @@ import dai from "@/public/dapps/stabledai.png"
 import status from "@/public/dapps/status.png"
 import superrare from "@/public/dapps/superrare.png"
 import synthetix from "@/public/dapps/synthetix.png"
-import uniswapec from "@/public/dapps/uni.png"
-import uniswap from "@/public/dapps/uni.png"
+import {
+  default as uniswap,
+  default as uniswapec,
+} from "@/public/dapps/uni.png"
```
pettinarip, line 113: Why are we doing this?

pettinarip (changes requested):
Thanks for the refactor @enjoyooor. Everything looks good but there is one import condition that I don't think is necessary and shouldn't be included in the scope of this PR. LMK what do you think.

src/pages/dapps.tsx
TylerAPfledderer: That is indeed an auto-fix from eslint. Else the following error throws:
`Module '"*.png"' has no exported member 'uniswapec'. Did you mean to use 'import uniswapec from "*.png"' instead?`

I'm unfamiliar with how this is possible to get multiple named imports from a single image file anyway 🤔

TylerAPfledderer: @pettinarip that is indeed an auto-fix from eslint, as it is the same image file imported twice. I guess eslint is unaware that it's an image file, so it consolidated the alias naming.

I'll look at throwing up a quick PR to address this duplication.

pettinarip: Aha! didn't notice that, good call, thanks. Ok, so I guess we are ok then here 👍🏼

pettinarip (approved):
Thanks @enjoyooor! 💯

PR #12287, opened 2024-02-25
fix: fixed date format

### LINKED ISSUE #11335: Incorrect intl number formatting

ameeetgaikwad:
## Description
Created a function in src/lib/utils/date.ts which can be used to format dates in a certain format according to the language.

## Related Issue
fixes #11335
closes #11443


nhsz:
@wackerow I think PR https://github.com/ethereum/ethereum-org-website/pull/12209 overlaps a bit with this one, not sure if this one is still needed

src/lib/utils/date.ts
```diff
@@ -1,2 +1,5 @@
 export const dateToString = (published: Date | string) =>
   new Date(published).toISOString().split("T")[0]
+export const formatDate = (date: Date, locale: string): string => {
+  return Intl.DateTimeFormat(locale).format(date)
+}
```
wackerow, line 5: We have `getLocaleTimestamp` inside `src/lib/utils/time.ts` which does this same thing with a bit more flexibility and type safety (allowing display options to be passed for example).

I think we should remove this addition and use `getLocaleTimestamp` instead

src/components/UpcomingEventsList.tsx
```diff
@@ -56,10 +59,11 @@ const UpcomingEventsList = () => {
...
         event.startDate === event.endDate
-          ? dateParse(event.startDate).toLocaleDateString()
-          : `${dateParse(event.startDate).toLocaleDateString()} - ${dateParse(
-              event.endDate
-            ).toLocaleDateString()}`
+          ? formatDate(dateParse(event.startDate), locale)
+          : `${formatDate(dateParse(event.startDate), locale)} - ${formatDate(
+              dateParse(event.endDate),
+              locale
+            )}`
```
wackerow, line 66: Would update this to use `getLocaleTimestamp` from `time.ts`... it accepts the `locale`, (of `Lang` type) then `timestamp` as a string (then optional `options` override argument which may be able to be left out here)

ie. `getLocaleTimestamp(locale! as Lang, dateParse(event.startDate).getTime())`

`locale!` force unwraps it, since we know we'll get a locale back from `useRouter`... and `as Lang` casts it to the expected type as defined inside `getLocaleTimestamp`.

^ With this, we need to fix the `locale` instantiation (see above), and also import `Lang` type above:

`import type { CommunityConference, Lang } from "@/lib/types"`

src/components/UpcomingEventsList.tsx
```diff
@@ -19,6 +21,7 @@ type OrderedUpcomingEvent = CommunityConference & {
 }
 
 const UpcomingEventsList = () => {
+  const { locale = "" } = useRouter()
```
wackerow, line 24:
```suggestion
  const { locale } = useRouter()
```

src/components/UpcomingEventsList.tsx
```diff
@@ -9,6 +10,7 @@ import EventCard from "@/components/EventCard"
 import InfoBanner from "@/components/InfoBanner"
 import InlineLink from "@/components/Link"
 
+import { formatDate } from "@/lib/utils/date"
```
wackerow, line 13:
```suggestion
```

src/lib/utils/date.ts
wackerow:
```suggestion
```

wackerow (changes requested):
Hey @ameeetgaikwad! Thanks for the PR... as @nhsz mentioned, we do have some other related helper functions already in the codebase for this.. Left some suggestions as to how we could reuse these and get this cleaned up. Let us know if you hit issues!

wackerow:
Just pushed a commit to update this with the latest from `dev` branch

nhsz:
@wackerow do you think this PR is still relevant?

wackerow (approved):
Went ahead and made a few updates to just use the getLocaleTimestamp function that already exists. Made a little helper function in-component to avoid repeating the options we want for this specific component. Pulling in!

wackerow:
> @wackerow do you think this PR is still relevant?

@nhsz And yes, this patches the number formatting on the community events page

PR #12299, opened 2024-02-26
feat: Filters UI (desktop & mobile)

nhsz:
[WIP] This PR implements the UI updates and changes for the wallet filters

**Preview:** https://deploy-preview-12299--ethereumorg.netlify.app/en/wallets/find-wallet/

## Description

- components
  - updates `WalletFilterProfile` component
  - updates `WalletFilterSidebar` component
  - updates `WalletsTable` styles
    - change background color to `background.highlight` when expanded
  - updates `WalletMoreInfo` styles
  - updates `WalletMoreInfoCategory` styles
  - adds `WalletFilterPersona` component
  - adds `SupportedLanguagesTooltip` component
  - adds `MobileFiltersMenu` component (see [new mobile menu design](https://www.figma.com/file/rv850lIICwOwRnyqSdtZED/Redesign-find-wallet?type=design&node-id=1961-25576&mode=design&t=KQuFmLvwA7z4ULYR-0))
  - adds `ResetFiltersButton` component
  - adds `LanguageSupportFilter` component
  - adds `MobileFiltersButton` component
- filters
  - updates `featureDropdownItems`
  - updates `filterOptions`
  - adds persona filter title
  - adds `Advanced` label/category to filters
  - removes `eip_1559_support` feature filter
  - features filter
    - remove description for devices 
- utils
  - adds `getPersonaBorderColor` util
  - adds `getLocaleFormattedDate` util
  - adds `getLanguageCodeName`util
  - adds `capitalize` util
  - adds `getAllWalletsLanguages` util
  - adds `walletsListingCount` util
  - adds `getWalletsTopLanguages` util
- hooks
  - moves `useWalletTable` inside `/hooks` dir
  - moves `useWalletFilterFeature` inside `/hooks` dir
- constants
  - adds `WALLETS_FILTERS_DEFAULT` constant
- types
  - adds `WalletFilter` type
  - adds `WalletSupportedLanguageContextType` type
- icons
  - removes unused `EIP1559Icon` 
- wallets data
  - fixes inconsistent `wallet.last_updated` format 
- translations
  - adds `page-find-wallet-personas-title` key
  - adds `page-find-wallet-see-wallets` key
  - updates `page-find-wallet-non-custodial` value 
  - updates `page-find-wallet-connect-to-dapps-desc` value
  - removes `page-find-choose-to-compare` unused key
  - removes `page-find-wallet-choose-to-compare` unused key
  - removes `page-find-wallet-persona-desc` unused key
  - removes `page-find-wallet-profile-filters` unused key
  - removes `page-find-wallet-feature-filters` unused key
  - removes `page-find-wallet-fee-optimization` unused key
  - removes `page-find-wallet-fee-optimization-desc` unused key
  - removes `page-find-wallet-mobile-desc` unused key
  - removes `page-find-wallet-desktop-desc` unused key
  - removes `page-find-wallet-browser-desc` unused key
  - removes `page-find-wallet-hardware-desc` unused key
  - removes `page-find-wallet-choose-features` unused key
  - removes `page-find-wallet-walletconnect` unused key
  - removes `page-find-wallet-walletconnect-desc` unused key
  - removes `page-find-wallet-token-support` unused key


src/components/FindWallet/WalletFilterSidebar/WalletFilterFeature/index.tsx
```diff
@@ -104,7 +104,7 @@ const WalletFilterFeature = ({
                     _hover={{ background: "transparent" }}
                   >
                     <Box as="span" flex={1}>
-                      {filterOption.title}
+                      <Text color="text">{filterOption.title}</Text>
```
pettinarip, line 107: `text` is not a token from the DS. Would be nice if we can use one of the semantic tokens. Not a blocker.

src/components/FindWallet/WalletFilterPersona.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { useWalletPersonas } from "../../hooks/useWalletPersonas"
+
+const WalletFilterPersona = ({
+  resetFilters,
+  setFilters,
+  selectedPersona,
+  setSelectedPersona,
+}) => {
+  const personas = useWalletPersonas()
+  const filterBg = useColorModeValue("chakra-subtle-bg", "black400")
```
pettinarip, line 23: Similar to the previous comment, I wonder where these colors come from since they are not defined in the DS. I'd like to see more reusability of the colors we've defined.

For reference, these are the [semantic tokens](https://github.com/ethereum/ethereum-org-website/blob/dev/src/@chakra-ui/semanticTokens.ts) we have.

cc @konopkja @nloureiro

src/lib/utils/wallets.ts
```diff
@@ -62,10 +62,18 @@ export const getWalletPersonas = (wallet: WalletData) => {
...
 export const getSupportedLanguages = (
   walletSupportedLanguages: string[],
-  languages: {}[]
+  languages: {}[],
+  locale: string
 ) => {
   const supportedLanguages = [] as string[]
 
+  // current locale should appear first on the list, this manipulates the array to move it to the top if needed
+  const localeIndex = walletSupportedLanguages.findIndex(
```
pettinarip, line 71: Won't this affect any other implementations we have that use this function?

pettinarip (commented):
@nhsz good job! this is looking great.

One small issue I noticed while testing:
https://github.com/ethereum/ethereum-org-website/assets/468158/0283ea10-b345-4137-b299-15b110f4baec

You can see that when you click on the description, it works fine, but when you click on the checkbox icon or the label next to it, it doesn't work. It seems that this retriggers the same filter condition.

konopkja:
Screenshot 2024-02-29 at 17 08 52

@nhsz till we add the functionality of selecting multiple personas, the checkbox should be changed back for radio button as checkbox indicates option for multiple select, which isnt going to be part of MVP (correct?)

konopkja:
Screenshot 2024-02-29 at 17 14 07
during testing i noticed that persona description can have bigger max length than the headline?

can we unify that? On some languages we have the issue that "new to crypto" has too little space and is divided into too many rows

konopkja:
Screenshot 2024-02-29 at 17 26 20

headlines bold to improve contrast/distinction with the list

src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -467,41 +335,46 @@ const WalletTable = ({
...
+
+                            {/* Supported languages */}
+                            <Flex alignItems="center" gap={3}>
+                              <Icon as={LanguagesIcon} fontSize="2xl" />
+
+                              <Text
+                                fontSize="1rem !important"
+                                fontWeight="normal !important"
+                              >
+                                {formatSupportedLanguages(supportedLanguages)}
```
minimalsm, line 365: Currently we're incorrectly showing that a wallet only supports the current language, when they infact only support English.

Screenshot 2024-02-29 at 16 00 26

Screenshot 2024-02-29 at 16 00 32

src/lib/utils/wallets.ts
nhsz: This util is only used to get the supported languages per wallet

src/components/FindWallet/WalletTable/index.tsx
nhsz: Fixed

konopkja:
@nhsz sry didnt know which PR to put this feedback, can we make the line go from the icon down when wallet expanded?
Group 896

hidden by default, but once a wallet is opened it goes all the way instead of how its now, because the gap seems a bit strange on desktop i think

nhsz:
> Screenshot 2024-02-29 at 17 14 07 during testing i noticed that persona description can have bigger max length than the headline?
> 
> can we unify that? On some languages we have the issue that "new to crypto" has too little space and is divided into too many rows

@konopkja could you check this again with the latest preview? thx

src/components/FindWallet/WalletFilterSidebar/WalletFilterFeature/index.tsx
nhsz: Fixed

konopkja:
> > Screenshot 2024-02-29 at 17 14 07 during testing i noticed that persona description can have bigger max length than the headline?
> > can we unify that? On some languages we have the issue that "new to crypto" has too little space and is divided into too many rows
> 
> @konopkja could you check this again with the latest preview? thx

looks good now!

src/components/FindWallet/WalletFilterPersona.tsx
nhsz: @pettinarip fixed

konopkja:
Screenshot 2024-03-04 at 15 09 46

this language name is missing

nhsz:
> Screenshot 2024-03-04 at 15 09 46
> 
> this language name is missing



> Hi @nhsz, noticed there are few new UI bugs on deploy. Persona boxes have some border that they should not have Screenshot 2024-03-04 at 15 23 11

I'm aware. Related to this, seems it won't be possible to display languages that are not supported on the site as we don't have the name and translation for them. I will only show the languages we support on this label and the filter for now.

konopkja:
> e



> > Screenshot 2024-03-04 at 15 09 46
> > this language name is missing
> 
> @konopkja I'm aware. Related to this, seems it won't be possible to display languages that are not supported on the site as we don't have the name and translation for them. I will only show the languages we support on this label and the filter for now.

hmm, this should be followed up on later as it is rather weird codependence and will limit users.

nhsz:
> > e
> 
> > > Screenshot 2024-03-04 at 15 09 46
> > > this language name is missing
> > 
> > 
> > @konopkja I'm aware. Related to this, seems it won't be possible to display languages that are not supported on the site as we don't have the name and translation for them. I will only show the languages we support on this label and the filter for now.
> 
> hmm, this should be followed up on later as it is rather weird codependence and will limit users.

It's not possible to display languages we don't support as we don't have the name and translation for them (doesn't make sense to add a new language to the site every time we add a new wallet that supports other languages we don't)

konopkja:
> > > e
> > 
> > 
> > > > Screenshot 2024-03-04 at 15 09 46
> > > > this language name is missing
> > > 
> > > 
> > > @konopkja I'm aware. Related to this, seems it won't be possible to display languages that are not supported on the site as we don't have the name and translation for them. I will only show the languages we support on this label and the filter for now.
> > 
> > 
> > hmm, this should be followed up on later as it is rather weird codependence and will limit users.
> 
> It's not possible to display languages we don't support as we don't have the name and translation for them (doesn't make sense to add a new language to the site every time we add a new wallet that supports other languages we don't)

I agree, of course these lists (wallet languages, website languages) should be separate, independent of each other in the future.

konopkja:
Screenshot 2024-03-04 at 22 40 39

When I mouse over the tooltip I expect to see only the amount of options that the number states. Currently it lists all languages, even those that are already shown. In this screenshot I expected to see 2 languages in the tooltip.

konopkja:
IMG_8061
if i tap on the number, it also opens wallet detail. This is unexpected. Not a biggie but if possible keep the wallet closed in this area.

nhsz:
> IMG_8061 if i tap on the number, it also opens wallet detail. This is unexpected. Not a biggie but if possible keep the wallet closed in this area.

@konopkja not sure how to handle this as the whole wallet area is clickable and is expected to open when clicked or tapped

wackerow:
> > > > this language name is missing
> > > 
> > > 
> > > @konopkja I'm aware. Related to this, seems it won't be possible to display languages that are not supported on the site as we don't have the name and translation for them. I will only show the languages we support on this label and the filter for now.
> > 
> > 
> > hmm, this should be followed up on later as it is rather weird codependence and will limit users.
> 
> It's not possible to display languages we don't support as we don't have the name and translation for them (doesn't make sense to add a new language to the site every time we add a new wallet that supports other languages we don't)

[`Intl.DisplayNames`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames) can be used for this.

For example, to render the display name for Amharic (`am`) in Korean (`ko`), you would use:

```js
new Intl.DisplayNames(["ko"], {type: "language"}).of("am") // => '암하라어'
```

nhsz:
> Intl.DisplayNames



> > > > > this language name is missing
> > > > 
> > > > 
> > > > @konopkja I'm aware. Related to this, seems it won't be possible to display languages that are not supported on the site as we don't have the name and translation for them. I will only show the languages we support on this label and the filter for now.
> > > 
> > > 
> > > hmm, this should be followed up on later as it is rather weird codependence and will limit users.
> > 
> > 
> > It's not possible to display languages we don't support as we don't have the name and translation for them (doesn't make sense to add a new language to the site every time we add a new wallet that supports other languages we don't)
> 
> [`Intl.DisplayNames`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames) can be used for this.
> 
> For example, to render the display name for Amharic (`am`) in Korean (`ko`), you would use:
> 
> ```js
> new Intl.DisplayNames(["ko"], {type: "language"}).of("am") // => '암하라어'
> ```

Nice! I'm gonna check this

src/components/FindWallet/WalletTable/WalletMoreInfoCategory.tsx
```diff
@@ -29,35 +36,41 @@ export const WalletMoreInfoCategory = ({
...
           const featureColor = wallet[feature.filterKey!] ? "text" : "secondary"
 
+          const FeatureIcon = () => (
+            <Icon
+              as={
+                wallet[feature.filterKey!]
+                  ? GreenCheckProductGlyphIcon
+                  : WarningProductGlyphIcon
+              }
+              fontSize="1rem"
```
wackerow, line 57:
```suggestion
              fontSize="md"
```

or, we may not even need this since I believe that would be the default size.

src/components/FindWallet/WalletTable/WalletMoreInfo.tsx
```diff
@@ -68,52 +75,77 @@ export const WalletMoreInfo = ({
...
             {/* Social icons */}
             <Heading
               as="h4"
               lineHeight={1.4}
               fontSize="md"
-              color="text"
-              fontWeight={500}
-              mx="0.2rem"
+              fontWeight="bold"
+              ms={{ lg: "0.5rem" }}
```
wackerow, line 142:
```suggestion
              ms={{ lg: 2 }}
```

wackerow (commented):
Not sure these are in scope for this PR, but just noting the following:

image

Noticing the tag background color is kinda hard to see on dark mode and blends into the background on hover for both light/dark modes.

---

image

I'm not sure we want the explainer text to also be low contrast, do we? Also looks like the popover arrow is misaligned; would expect it to be centered above the circle icon.

---

Otherwise this is looking really good... just left some comments related to trying to use our existing design tokens for consistency.

src/lib/utils/wallets.ts
```diff
@@ -62,25 +62,40 @@ export const getWalletPersonas = (wallet: WalletData) => {
...
   const supportedLanguages = [] as string[]
 
+  // current locale should appear first on the list, this manipulates the array to move it to the top if needed
+  const supportsCurrentLocale = (current) => current === locale
+  const localeIndex = walletSupportedLanguages.findIndex(supportsCurrentLocale)
+
+  if (localeIndex >= 0) {
+    walletSupportedLanguages.splice(localeIndex, 1)
+    walletSupportedLanguages.unshift(locale)
+  }
```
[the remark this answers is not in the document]
nhsz: No need to sort the array

src/pages/wallets/find-wallet.tsx
```diff
@@ -279,54 +272,56 @@ const FindWalletPage = () => {
...
-          />
-        </Show>
-        <Box
-          w="full"
-          sx={{
-            scrollbarWidth: "thin",
-            scrollbarColor: `${theme.colors.lightBorder} ${theme.colors.background}`,
+      <Box px={{ md: 4, "2xl": 0 }}>
+        <Flex pt={4} pb={6} gap={6}>
+          <Show above="lg">
```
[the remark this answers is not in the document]
nhsz: Fixed, thanks!

src/components/FindWallet/WalletFilterPersona.tsx
```diff
@@ -0,0 +1,111 @@
...
+                setFilters(persona.presetFilters)
+                trackCustomEvent({
+                  eventCategory: "UserPersona",
+                  eventAction: `${persona.title}`,
+                  eventName: `${persona.title} true`,
+                })
+              }
+            }}
+          >
+            <Flex alignItems="center" gap={2} mb="0.2rem" px={1.5}>
```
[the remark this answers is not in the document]
nhsz: Updated

src/components/FindWallet/WalletFilterPersona.tsx
```diff
@@ -0,0 +1,111 @@
...
+                  outlineOffset="3px"
+                  fontSize={8}
+                  _groupHover={{ outlineColor: "primary.hover" }}
+                />
+              </Box>
+
+              <Heading
+                as="h3"
+                ms={0}
+                fontSize="1.2rem"
```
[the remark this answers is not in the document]
nhsz: Fixed

src/components/FindWallet/WalletFilterPersona.tsx
```diff
@@ -0,0 +1,111 @@
...
+                color="primary.base"
+              >
+                {persona.title}
+              </Heading>
+            </Flex>
+
+            <Text
+              p="0.4rem"
+              color={selectedPersona === idx ? "text" : "text200"}
+              fontSize="0.9rem"
```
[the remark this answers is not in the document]
nhsz: Fixed, thanks

src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -421,9 +270,19 @@ const WalletTable = ({
...
           const supportedLanguages = getSupportedLanguages(
             wallet.languages_supported,
-            languagesList
+            languagesList,
+            locale!
+          )
+          const numberOfSupportedLanguages = supportedLanguages.length
+          const rest = numberOfSupportedLanguages - 5
+          const restText = `${rest > 0 ? "+" : ""} ${rest > 0 ? rest : ""}`
+          const sliceSize = 5
```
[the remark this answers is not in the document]
nhsz: Fixed, thanks

src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -441,24 +300,42 @@ const WalletTable = ({
...
+                      <Box
+                        w={{ base: "24px", md: "56px" }}
+                        mb={deviceLabels.length > 0 ? 0 : 36}
+                      >
                         <Image
                           src={wallet.image}
                           alt=""
                           objectFit="contain"
-                          boxSize="56px"
+                          boxSize={{ base: "24px", md: "56px" }}
```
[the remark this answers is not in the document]
nhsz: Fixed

src/components/FindWallet/WalletTable/WalletMoreInfoCategory.tsx
```diff
@@ -72,16 +85,12 @@ export const WalletMoreInfoCategory = ({
...
-                      ? GreenCheckProductGlyphIcon
-                      : WarningProductGlyphIcon
-                  }
-                  fontSize="1rem"
-                  color={featureColor}
-                />
-                <p>{feature.label}</p>
+                <ListIcon as={FeatureIcon}></ListIcon>
+
+                <Text px="0.2rem" lineHeight={1}>
```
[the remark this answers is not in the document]
nhsz: Fixed

src/components/FindWallet/WalletTable/WalletMoreInfo.tsx
nhsz: Fixed

nhsz:
> Not sure these are in scope for this PR, but just noting the following:
> 
> image
> Noticing the tag background color is kinda hard to see on dark mode and blends into the background on hover for both light/dark modes.
> 
> image
> I'm not sure we want the explainer text to also be low contrast, do we? Also looks like the popover arrow is misaligned; would expect it to be centered above the circle icon.
> 
> Otherwise this is looking really good... just left some comments related to trying to use our existing design tokens for consistency.

@wackerow I've fixed the tooltip text color issue for disabled items. Tags contrast is out of scope and should be reviewed in the DS.

konopkja:
Screenshot 2024-03-06 at 17 39 45
When selecting "new to crypto" filter I see Mew wallet being the only one that doesnt have the tag actually listed there.

konopkja:
> Screenshot 2024-02-29 at 17 12 31
> 
> there is a too much empty space between tags and wallet name

unsure if it has been tackled yet, but still an issue

konopkja:
Screenshot 2024-03-06 at 17 50 16
 cant find the comment so just in case adding it again - when I click on a persona and then select another filter, the persona gets unclicked even though the filter is still active.
 
 The persona should remain clicked.

nhsz:
> > Screenshot 2024-02-29 at 17 12 31
> > there is a too much empty space between tags and wallet name
> 
> unsure if it has been tackled yet, but still an issue

I'm using the same padding as the 'Visit website' button there

nhsz:
> Screenshot 2024-03-06 at 17 50 16 cant find the comment so just in case adding it again - when I click on a persona and then select another filter, the persona gets unclicked even though the filter is still active.
> 
> The persona should remain clicked.

This is a pre-existent issue (check behavior on https://ethereum.org/en/wallets/find-wallet/) and not easy to fix, so I'll give it a low priority for now

minimalsm:
@CodiumAI-Agent /review

konopkja:
Screenshot 2024-03-06 at 20 04 42
unsure if this the right PR but why is the tooltip on mobile so limited, it could definitely be wider. @nloureiro was this DS decision?

nloureiro:
> Screenshot 2024-03-06 at 20 04 42 unsure if this the right PR but why is the tooltip on mobile so limited, it could definitely be wider. @nloureiro was this DS decision?

I don't know why; maybe it's a chakra default.

In Production, they have 300px on desktop and 140px on mobile. We can make them wider on mobile, I don't oppose to that.

src/pages/wallets/find-wallet.tsx
wackerow, line 1: > > Screenshot 2024-03-06 at 17 50 16 cant find the comment so just in case adding it again - when I click on a persona and then select another filter, the persona gets unclicked even though the filter is still active.
> > The persona should remain clicked.
> 
> This is a pre-existent issue (check behavior on https://ethereum.org/en/wallets/find-wallet/) and not easy to fix, so I'll give it a low priority for now

On line 108, I tried commenting out `setSelectedPersona(NaN)` and it seems to be fixing this issue by not resetting the persona when a filter is toggled. Haven't tested thoroughly, but @nhsz just curious if you've tried this yet or not.

src/pages/wallets/find-wallet.tsx
```diff
@@ -203,86 +163,50 @@ const FindWalletPage = () => {
...
+          resetFilters={resetFilters}
+          setFilters={setFilters}
+          selectedPersona={selectedPersona}
+          setSelectedPersona={setSelectedPersona}
+          showMobileSidebar={showMobileSidebar}
+        />
+      </Box>
+
+      {/* Mobile filters menu */}
+      <Hide above="lg">
```
wackerow, line 189: Similar to a previous comment, I'm getting a glitch where this displays on Desktop on load, then quickly disappears... recurring issue when using the `Show` or `Hide` components which are rendered server-side before it knows what size the device is. Switching to a Box (or other Chakra component) with `hideFrom="lg"` should fix this.

src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -441,24 +299,42 @@ const WalletTable = ({
...
 
                       <Box w={{ base: "100%", md: "auto" }}>
-                        <Stack gap={5}>
-                          <Text>{wallet.name}</Text>
+                        <Stack>
+                          <Box mb={3}>
+                            <Text lineHeight={1.2} fontSize="xl !important">
+                              {wallet.name}
+                            </Text>
+                          </Box>
```
wackerow, line 330: > > > Screenshot 2024-02-29 at 17 12 31
> > > there is a too much empty space between tags and wallet name
> > 
> > 
> > unsure if it has been tackled yet, but still an issue
> 
> I'm using the same padding as the 'Visit website' button there

Looks like there is a `gap` being used in this flex box, but then we also have a margin-bottom of 12px (chakra `3`)... perhaps we can just remove this margin?

image

```suggestion
                          <Text lineHeight={1.2} fontSize="xl !important">
                            {wallet.name}
                          </Text>
```

Removing the container Box/margin displays as:

image

cc: @konopkja

konopkja: lgtm

konopkja:
Screenshot 2024-03-07 at 14 34 30


1.  seems like the column width his inconsistent? buy/sell crypto should break to 2 rows sooner
2. Some tooltips seem to be very distanced from the term

src/pages/wallets/find-wallet.tsx
nhsz: Solves part of the issue, but notice tha when you click on a filter and then a Persona, the filters checkboxes get unchecked

nhsz:
> Screenshot 2024-03-07 at 14 34 30
> 1. seems like the column width is inconsistent? buy/sell crypto should break to 2 rows sooner
> 2. Some tooltips seem to be very distanced from the term

I've fixed 1. Number 2 is becase if long text wrapping, you can see better below

Screen Shot 2024-03-07 at 11 58 32

nhsz:
> > > Screenshot 2024-03-07 at 14 34 30
> > > 1. seems like the column width is inconsistent? buy/sell crypto should break to 2 rows sooner
> > > 2. Some tooltips seem to be very distanced from the term
> > 
> > 
> > I've fixed 1 (not published yet). Number 2 is becase if long text wrapping, you can see better below
> > Screen Shot 2024-03-07 at 11 58 32
> 
> ok that makes sense, hmm is there anything that can be done on this?

I'll investigate it, maybe @wackerow you have a hint here? (reducing gap between text and tooltip when text is wrapped)

konopkja:
Screenshot 2024-03-07 at 21 21 48

noticed a UI bug in latest deploy around the personas there is  some extra background

konopkja:
Screenshot 2024-03-07 at 21 23 23

z-index issues with the language dropdown and sticky bar

konopkja:
Screenshot 2024-03-07 at 21 24 52

if user doesnt select any filter, the button should not show (0) instead it should indicate the total amount of wallets. Showing '0" might be interpreted as there will be no wallets visible.

konopkja:
Screenshot 2024-03-07 at 21 30 19

selected developer persona, but first wallet doesnt have the tag?

src/pages/wallets/find-wallet.tsx
wackerow: Okay, interesting... I tried only removing this from `updateFilterOptions` (plural) and I thought it was working, but I see the filter _toggle_ remains activated, while the sub-item checkboxes all get unchecked.

https://github.com/ethereum/ethereum-org-website/assets/54227730/67c3bda9-3978-433f-a337-c1941480077c

Yeah, looks like this may require a little digging... don't see it as a blocker for this PR.

wackerow:
> > > > Screenshot 2024-03-07 at 14 34 30
> > > > 1. seems like the column width is inconsistent? buy/sell crypto should break to 2 rows sooner
> > > > 2. Some tooltips seem to be very distanced from the term
> > > 
> > > 
> > > I've fixed 1 (not published yet). Number 2 is becase if long text wrapping, you can see better below
> > > Screen Shot 2024-03-07 at 11 58 32
> > 
> > 
> > ok that makes sense, hmm is there anything that can be done on this?
> 
> I'll investigate it, maybe @wackerow you have a hint here? (reducing gap between text and tooltip when text is wrapped)

image

Ah.. yeah this is an annoying bug with wrapping text, since the browser calculates the space the text should take up, and then wraps if it extends beyond, but it does _not_ collapse the existing space it's been allocated (hence why the "i" icon is pushed all the way over as if the text was taking all that other space. I'll think on this... we've encountered this before, but I'd need to dig to find where, and how we fixed it.

konopkja:
Screenshot 2024-03-08 at 22 04 23

Sorry @nhsz for putting the comment here on friday :D the blue languages are not clickable yet - is that WiP?

konopkja:
Screenshot 2024-03-08 at 22 06 04
dont see this filter in the sidebar

konopkja:
Screenshot 2024-03-08 at 22 14 29
Screenshot 2024-03-08 at 22 15 05
Screenshot 2024-03-08 at 22 15 17


noticed some wallets have the line in detail correct and some go through the logo

nhsz:
> Screenshot 2024-03-08 at 22 04 23
> Sorry @nhsz for putting the comment here on friday :D the blue languages are not clickable yet - is that WiP?

Yes, as I mentioned I've only pushed the UI of the filter. I'm usually updating the PR once a day only, no changes from yesterday

konopkja:
Screenshot 2024-03-08 at 22 22 24

when a wallet detail is expanded change background color to color highlight

konopkja:
Screenshot 2024-03-08 at 22 28 03

noticed that when the list is smaller than the sidebar, the main body sticky header disappears resulting in a strange look, should probably remain there

wackerow:
> > > > > Screenshot 2024-03-07 at 14 34 30
> > > > > 1. seems like the column width is inconsistent? buy/sell crypto should break to 2 rows sooner
> > > > > 2. Some tooltips seem to be very distanced from the term
> > > > 
> > > > 
> > > > I've fixed 1 (not published yet). Number 2 is becase if long text wrapping, you can see better below
> > > > Screen Shot 2024-03-07 at 11 58 32
> > > 
> > > 
> > > ok that makes sense, hmm is there anything that can be done on this?
> > 
> > 
> > I'll investigate it, maybe @wackerow you have a hint here? (reducing gap between text and tooltip when text is wrapped)
> 
> image
> Ah.. yeah this is an annoying bug with wrapping text, since the browser calculates the space the text should take up, and then wraps if it extends beyond, but it does _not_ collapse the existing space it's been allocated (hence why the "i" icon is pushed all the way over as if the text was taking all that other space. I'll think on this... we've encountered this before, but I'd need to dig to find where, and how we fixed it.

Played around with this a little... it's hacky, but found a _potential_ solution:

The _icon (check/X) + label + icon (`i`)_ are contained in a Flex, and each are `block` display... 

image

If we move the `i` icon _inside_ the `label` element as `display: inline-block` we can force that `i` icon to be inline with the label text...

image

_but_, as shown, this could then land the icon by itself on the next line which looks bad. This is where it gets a little more hacky, but we could split off the _last word_ as it's own string, and wrap the `i` icon with only the last word using `whiteSpace: nowrap`, which will force the `i` icon to stick to the last word and only wrap together:

image

The working code I used for this was inside `WalletMoreInfoCategory.tsx`:

```tsx
          if (feature.category === sectionName) {
            const split = feature.label.split(" ")
            const firstWords = split.slice(0, -1).join(" ")
            const lastWord = split.slice(-1)
            return (
              <ListItem
                key={feature.label}
                fontSize="0.9rem"
                listStyleType="none"
                display="flex"
                mb={2}
                width={{ base: "auto", xl: "full" }}
                sx={{
                  p: {
                    color: featureColor,
                    mb: 0,
                  },
                  "span + p": {
                    textDecor: "none",
                  },
                  "p + div, div + div": {
                    svg: {
                      width: 6,
                      fill: "secondary",
                      pe: 2,
                    },
                  },
                }}
              >
                <ListIcon as={FeatureIcon}></ListIcon>

                <Text px={1} lineHeight={1} display="inline-block">
                  {firstWords}{" "}
                  <Text as="span" whiteSpace="nowrap">
                    {lastWord}
                    <Tooltip
                      content={
                        <Text color="body.base !important">
                          {t(walletFilterData[feature.filterKey].description)}
                        </Text>
                      }
                    >
                      <Icon as={MdInfoOutline} color={featureColor} ms="1" />
                    </Tooltip>
                  </Text>
                </Text>
              </ListItem>
            )
          }
```

New above are the constants to break the string apart, and the `<Text></Text>` block that contains them. 

I'm not sure we can solve this without using inline styling (ie a solution that keeps the `i` icon to the right of all the text instead of just to the right of the last word)... @konopkja curious your thoughts on how this looks. @nhsz Feel free to use any or none of the above.

nhsz:
@wackerow Thanks! As you mentioned this is hacky and generates an extra embedded p tag which is not ideal in terms of accessibility and semantics, which IMO should be preserved. I'll leave as it is for now, not a big issue

wackerow:
Yeah, not a huge fan of it hah, but a proof-of-concept at least. One note, it would embed a `span` not a `p`, but in either case I'm not married to the solution. If I come across a better solution I can post here

src/components/FindWallet/LanguageSupportFilter.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+              {getWalletsTopLanguages(5, locale!)
+                .map<ReactNode>((language) => {
+                  return (
+                    <Text
+                      key={language.code}
+                      as="span"
+                      color="primary.base"
+                      fontSize="sm"
+                      lineHeight={1.6}
```
pettinarip, line 122: Same in this case `size="sm"`

src/components/FindWallet/LanguageSupportFilter.tsx
```diff
@@ -0,0 +1,149 @@
...
+                          ),
+                        })
+                      }}
+                    >
+                      {language.langName}
+                    </Text>
+                  )
+                })
+                // `.reduce()` used to generate the comma separator between languages
+                .reduce((prev, curr) => [prev, ", ", curr])}
```
pettinarip, line 142: nitpick
```suggestion
                .join(", ")
```

src/components/FindWallet/WalletFilterSidebar.tsx
```diff
@@ -0,0 +1,122 @@
...
+  updateFilterOptions,
+  top,
+  showMobileSidebar,
+  ...tabsProps
+}: WalletFilterSidebarProps) => {
+  const theme = useTheme()
+  const { t } = useTranslation("page-wallets-find-wallet")
+
+  return (
+    <Tabs
```
pettinarip, line 49: We are not using Tabs anymore, right?

src/components/FindWallet/WalletTable/WalletMoreInfo.tsx
```diff
@@ -68,52 +75,85 @@ export const WalletMoreInfo = ({
...
+    locale as Lang,
+    wallet.last_updated
+  )
+
+  // autoColumns={{ base: "200px", lg: "minmax(0, 1fr)" }}
+  //     templateColumns={
+  //       showMobileSidebar
+  //         ? "repeat(2, 1fr)"
+  //         : { base: "200px", lg: "minmax(0, 1fr)" }
+  //   }
```
pettinarip, line 95: ✂️

src/components/FindWallet/WalletTable/WalletMoreInfoCategory.tsx
```diff
@@ -29,35 +36,40 @@ export const WalletMoreInfoCategory = ({
...
-        as="h4"
-        lineHeight={1.4}
-        fontSize="md"
-        fontWeight={500}
-        mx="0.2rem"
-        mb={2}
-      >
+    <Box mx={{ base: 0, md: 0, lg: 2 }} w="100%">
+      {/* Category label */}
+      <Heading as="h4" lineHeight={1.4} fontSize="md" fontWeight="bold" mb={2}>
```
pettinarip, line 41:
```suggestion
      <Heading as="h4" size="md" mb={2}>
```

src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -308,27 +323,30 @@ const WalletTable = ({
...
+                        <Flex alignItems="start">
+                          {/* Wallet image */}
+                          <Stack
+                            w={{ base: "24px", md: "56px" }}
+                            h={{ base: "24px", md: "56px" }}
+                            me={4}
+                          >
+                            <Image
+                              src={wallet.image}
+                              alt=""
```
pettinarip, line 336: Would recommend limiting the size of these images. With the current config, we would be fetching an unnecessarily larger image, like the "block wallet" image that is 500pxs, and then shrinking it with css.

```suggestion
                            <Image
                              src={wallet.image}
                              height={56}
                              alt=""
```

src/components/FindWallet/MobileFiltersMenu.tsx
```diff
@@ -0,0 +1,182 @@
...
+  onOpen,
+  onClose,
+}: MobileFiltersMenuProps) => {
+  const { t } = useTranslation("page-wallets-find-wallet")
+  const { filteredWallets } = useWalletTable({ filters, t, walletData })
+
+  return (
+    <>
+      <Box
+        display={{ base: "block", lg: "none" }}
```
pettinarip, line 59: This seems unnecessary. Seems that the parent (find-wallet page) is controlling this already.

pettinarip (commented):
On the functional testing side, I think the only thing I would question is the language selector. It doesn’t work in the same way as the other filters. When you select a language:

- I would expect to see the selected language displayed instead of the “Search language” placeholder
- The filter is working fine but it is not shown in the “Filters (0)” counter

These issues can cause confusion for the user when trying to understand what filters are being applied.

image

I see a visual bug in mobile. The modal footer sometimes gets behind the content
image


=======
On the code side, I've left a couple of suggestions/questions, nothing blocking. Looks good overall.


Great job @nhsz 💯

src/components/FindWallet/LanguageSupportFilter.tsx
```diff
@@ -0,0 +1,149 @@
...
+              aria-label={t("page-find-wallet-languages-supported")}
+              options={allWalletsLanguagesOptions}
+              onChange={handleLanguageFilterSelectChange}
+              placeholder={t("page-find-wallet-languages-search-language")}
+              isSearchable
+              variant="outline"
+            />
+
+            <Box mt={2}>
+              <Text color="body.medium" fontSize="sm" lineHeight={1.6}>
```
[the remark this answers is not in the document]
wackerow: Wasn't aware of this shorthand, thanks @pettinarip

src/components/FindWallet/WalletFilterPersona.tsx
```diff
@@ -0,0 +1,149 @@
...
+                  <Icon
+                    as={MdCircle}
+                    borderRadius="full"
+                    boxSize={2.5}
+                    my={0}
+                    mx={1}
+                    fill={
+                      selectedPersona === idx
+                        ? "primary.base"
+                        : "rgba(0, 0, 0, 0)"
```
wackerow, line 99:
```suggestion
                        : "transparent"
```

wackerow: Could also consider extracting since this logic is reused immediately below

src/components/FindWallet/WalletFilterPersona.tsx
```diff
@@ -0,0 +1,149 @@
...
+                display={showMobileSidebar ? "none" : "block"}
+                p="0.4rem"
+                color={selectedPersona === idx ? "text" : "text200"}
+                fontSize="sm"
+                fontWeight="normal"
+                transition="0.5s all"
+                lineHeight={1.3}
+              >
+                {persona.description}
+              </Text>
```
wackerow, line 140: Just noting, I feel like we use the `{someBoolean && <ConditionalComponent {...} />}` pattern much more than `display{someBoolean ? "none" : "block"}` pattern. Not sure if we have this documented anywhere... not a blocker but just noting

Also, maybe a question for @nloureiro, but is it possible to move away from custom padding sizes here? (ie. `0.4rem`, or `6.4px`)... could we not just use `"2"` here for half a rem?

src/components/FindWallet/WalletFilterSidebar.tsx
```diff
@@ -0,0 +1,122 @@
...
+        "::-webkit-scrollbar-thumb": {
+          bgColor: "lightBorder",
+          borderRadius: "base",
+          border: "2px solid",
+          borderColor: "background.base",
+        },
+      }}
+      {...tabsProps}
+    >
+      <Box position="sticky" top={top ?? 0} zIndex={2} w={{ md: "330px" }}>
```
wackerow, line 70: Similar to above, @nloureiro could we shift towards the presets we have available? A width of `"xs"` would be 20rem (320px)?

image

src/components/FindWallet/WalletFilterSidebar.tsx
```diff
@@ -0,0 +1,122 @@
...
+              filters
+            )})`}
+          </Text>
+
+          <Box display={showMobileSidebar ? "none" : "block"}>
+            <ResetFiltersButton
+              resetFilters={resetFilters}
+              resetWalletFilter={resetWalletFilter}
+            />
+          </Box>
```
wackerow, line 95: Similar to above comment, could we just use the following?

```suggestion
          {!showMobileSidebar && (
            <ResetFiltersButton
              resetFilters={resetFilters}
              resetWalletFilter={resetWalletFilter}
            />
          )}
```

src/components/FindWallet/WalletTable/WalletMoreInfoCategory.tsx
```diff
@@ -29,35 +36,40 @@ export const WalletMoreInfoCategory = ({
...
+            />
+          )
 
           if (feature.category === sectionName)
             return (
-              <HStack
+              <ListItem
                 key={feature.label}
-                spacing="0.2rem"
                 fontSize="0.9rem"
```
wackerow, line 65: Again not a blocker, but cc: @nloureiro if there is a way we can try to work our way towards standard font sizes

src/components/FindWallet/ResetFiltersButton.tsx
```diff
@@ -0,0 +1,41 @@
...
+        resetWalletFilter.current()
+        trackCustomEvent({
+          eventCategory: "WalletFilterSidebar",
+          eventAction: `Reset button`,
+          eventName: `reset_click`,
+        })
+      }}
+    >
+      <Icon as={BsArrowCounterclockwise} aria-hidden="true" fontSize="sm" />
+      {t("page-find-wallet-reset-filters").toUpperCase()}
```
wackerow, line 38: Small non-blocking nit pick, to avoid JS for styling, might suggest textTransform here:

```suggestion
      textTransform="uppercase"
    >
      <Icon as={BsArrowCounterclockwise} aria-hidden="true" fontSize="sm" />
      {t("page-find-wallet-reset-filters")}
```

src/components/FindWallet/WalletFilterFeature.tsx
```diff
@@ -160,42 +180,46 @@ const WalletFilterFeature = ({
...
+                                    }
+                                  />
+                                )}
+                                {item.showOptions !== undefined && (
+                                  <FilterToggle
+                                    ariaLabel={item.title}
+                                    conditionItem={item.showOptions}
+                                  />
+                                )}
+                              </>
```
wackerow, line 208: Are these fragments doing anything? I think they can be removed

src/components/FindWallet/WalletFilterPersona.tsx
```diff
@@ -0,0 +1,149 @@
...
+            <Flex
+              h="100%"
+              direction="column"
+              alignItems="flex-start"
+              padding={2}
+              border="solid"
+              borderColor={getPersonaBorderColor(selectedPersona, idx)}
+              bg="background.highlight"
+              borderRadius="base"
+              cursor="pointer"
```
wackerow, line 63: For a11y, would make this a button to make them tab-stoppable

src/components/FindWallet/WalletFilterPersona.tsx
```diff
@@ -0,0 +1,149 @@
...
+                        ? "primary.base"
+                        : "rgba(0, 0, 0, 0)"
+                    }
+                    background={
+                      selectedPersona === idx ? "primary.base" : "transparent"
+                    }
+                    outline="1.5px solid"
+                    outlineColor="primary.base"
+                    outlineOffset="3px"
+                    fontSize={8}
```
wackerow, line 107: fontSize={8} is 2rem correct? @nloureiro Do you know how often we're trying to use this font size? Not blocking for this, but again noting we have some standard font sizes that would be nice to be able to stick to:

image

src/components/FindWallet/LanguageSupportFilter.tsx
nhsz: @pettinarip its an array of objects, doesnt work with `join()`

src/components/FindWallet/WalletTable/WalletMoreInfo.tsx
nhsz: Oops, thx

src/components/FindWallet/WalletTable/WalletMoreInfoCategory.tsx
nhsz: `size="md"` is not the same font size defined [in the design](https://www.figma.com/file/rv850lIICwOwRnyqSdtZED/Redesign-find-wallet?type=design&node-id=2289-7095&mode=design&t=Wlxi1fSerY50efXv-0) @konopkja

src/components/FindWallet/WalletTable/index.tsx
nhsz: @pettinarip height is already defined below, `h={{ base: "24px", md: "56px" }}`

wackerow (commented):
Small note, if I select a filter, then select a persona, the filter _toggle_ stays on, but the checkboxes get unchecked.

https://github.com/ethereum/ethereum-org-website/assets/54227730/4776ae06-1e9c-42a6-a78e-65461456517d

---

Otherwise this is looking great.. functionally it's working as expected aside from the couple things noted. Just left some clean up notes, and a few things we could discuss to address separately (ie using standard sizes for a handful of things)

Nice job @nhsz! Excited to move this forward

src/components/FindWallet/MobileFiltersMenu.tsx
nhsz: Fixed

src/components/FindWallet/MobileFiltersMenu.tsx
```diff
@@ -0,0 +1,182 @@
...
+  WalletFilterSidebarProps,
+} from "./WalletFilterSidebar"
+
+import { useWalletTable } from "@/hooks/useWalletTable"
+
+interface MobileFiltersMenuProps extends WalletFilterSidebarProps {
+  showMobileSidebar: boolean
+  onOpen: () => void
+  onClose: () => void
+}
```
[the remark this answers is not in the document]
nhsz: @wackerow `showMobileSidebar` is the same as `isOpen` but renamed for clarity, comes from `find-wallet`

src/components/FindWallet/LanguageSupportFilter.tsx
pettinarip: Ah right! good then

src/components/FindWallet/WalletTable/index.tsx
pettinarip: `h` is not the same as `height`. `h` is a Chakra shorthand for the css prop. `height` is a prop that is used by the [NextJS image component](https://nextjs.org/docs/app/api-reference/components/image#height) to determine the size of the image served.

pettinarip: Not a blocker though but it is recommended.

pettinarip (approved):
Awesome @nhsz, lgtm! gj

The only thing left for me (which I don't see as a blocker) is this issue on mobile:
image

I keep seeing this on a physical phone when the top frame of the browser (where the url input is) is shown.

src/components/FindWallet/WalletTable/index.tsx
nhsz: Right, I've already fixed it, thanks

src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -228,24 +225,42 @@ const WalletTable = ({
...
-          ) : (
-            <Text ps={{ base: 2, md: 0 }} as="span">
-              {t("page-find-wallet-showing")}{" "}
-              <strong>
-                {filteredWallets.length} / {walletCardData.length}
-              </strong>{" "}
-              {t("page-find-wallet-wallets")}
+          <Flex justifyContent="space-between" px={{ base: 2.5, md: 0 }}>
+            <Text
+              display={{ base: "block", lg: "none" }}
```
wackerow, line 237: Not a requirement, just noting we can use this syntax as well:
```suggestion
              hideFrom="lg"
```

wackerow (approved):
Looks good! Thanks @nhsz! Just left one comment about the `hideFrom` syntax, but we don't need to deal with that here, just noting. Pulling in when build finishes! 🎉

src/components/FindWallet/WalletTable/index.tsx
nhsz: Thanks!, fixed in `find-wallet-revamp` branch

PR #12303, opened 2024-02-27
New L2 rollup diagram

nloureiro:
## Description
A new diagram for the L2 page to increase comprehension of the topic

## Preview link
- https://deploy-preview-12303--ethereumorg.netlify.app/en/layer-2


minimalsm (changes requested):
I think we should change the text to 'bundle many transactions into one transaction' for clearer language.

Also 'finalized' instead of 'finalised' as we use American English per style guide

wackerow (commented):
@nloureiro Left a few comments directly on the Figma here related to it's flow with time

image

wackerow (approved):
Looks good! Thanks @nloureiro!

Issue #12150, opened 2024-02-12
Hard to see the writings on a diagram


aslikaya:
### Describe the bug

When in the dark mode, reading the writings on a diagram is very hard.

### To reproduce

1. Go to https://ethereum.org/developers/docs/consensus-mechanisms/pos/attack-and-defense#reorgs
2. Scroll down to the [diagram](https://ethereum.org/content/developers/docs/consensus-mechanisms/pos/attack-and-defense/reorg-schematic.png)
3. See error

### Expected behavior

It should be easily readable

### Screenshots

<img width="636" alt="Screenshot 2024-02-12 at 13 26 57" src="https://github.com/ethereum/ethereum-org-website/assets/9151261/59afa163-ded8-459b-b602-2e1d8108271c">

### Desktop (please complete the following information)

MacOS, Chrome

### Would you like to work on this issue?

- [ ] Yes
- [ ] No

iepn:
> ### 描述错误
> 在黑暗模式下，阅读图表上的文字非常困难。
> 
> ### 重现
> 1. 转到https://ethereum.org/developers/docs/consensus-mechanisms/pos/attack-and-defense#reorgs
> 2. 向下滚动到[图表](https://ethereum.org/content/developers/docs/consensus-mechanisms/pos/attack-and-defense/reorg-schematic.png)
> 3. 查看错误
> 
> ### 预期行为
> 它应该易于阅读
> 
> ### 截图
[...]

PR #12344, opened 2024-03-04
fix: Enable night mode support for images #12150

iepn:
## Description

image

image

## Related Issue

Hard to see the writings on a diagram #12150


iepn:
Not liking this color scheme? We can also follow the design style of ETH!

image

image

PR #12369, opened 2024-03-05
Delete the word "whether"

AaronChen0:
Nodes can vary in their lightness, depending upon ~~whether~~ the combinations of light and full client software they run.


wackerow (approved):
Agree, thanks @AaronChen0!

wackerow:
@all-contributors please add @AaronChen0 for bug fix

PR #12370, opened 2024-03-05
ERC4526 withdraw event documentation wrong field name for shares

melpomene:
In the ERC4526 the Withdraw event parameter is described as `shares` not `share`

## Description

Changing documentation to align with the [EIP for ERC4626 ](https://eips.ethereum.org/EIPS/eip-4626#events).

This also aligns with how it is implemented in e.g. sDAI.


wackerow (approved):
Good catch, looks good! Thanks @melpomene!

wackerow:
@all-contributors please add @melpomene for bug fix

PR #12374, opened 2024-03-05
Fix CardList external links not opening in new tabs

pettinarip:
## Description

Fixed this by passing the `isExternal` prop to the `LinkOverlay` component.


nhsz (approved):
LGTM

PR #12380, opened 2024-03-05
fix: nav item hover zone, opening delay duration

wackerow:
## Description
- Removes delay duration when hovering over a nav menu item (menu now opens immediately upon hovering)
- Extends the click/hover zone for each button to the bottom of the nav bar, eliminating dead-space between the buttons and the menu that opens which could close the menu if cursor is paused in this space

## Related Issue
Bug surfaced from #12125 review

cc: @konopkja


pettinarip (approved):
Nice! interesting solution, LGTM!

konopkja:
LGTM! i think this will improve the current experience! nice job

PR #12390, opened 2024-03-06
Update security guidelines for Ethereum and cryptocurrency users page

setgree:
stytlistic edits to bring article more in line with Ethereum guidelines. 

I notice a few passive voice instances, e.g. "They are considered the most secure wallet option for storing your private keys: your private key never touches the internet and stays completely local on your device," that could use attribution -- who considers this the most secure wallet option? is it an empirical fact that people who employ this technique get hacked the least? 

There's also a part telling a story about a scammer who "stayed in touch" that could use a hyperlink, but I don't know what story the OP is talking about

## Summary by CodeRabbit

- **Documentation**
	- Updated security advice related to Ethereum and cryptocurrency, including password security, Two-Factor Authentication (2FA), and common crypto scams.


public/content/security/index.md
```diff
@@ -125,15 +125,15 @@ One of the biggest reasons people get scammed in crypto generally is a lack of u
 
 **Never, for any reason, share your private keys!**
 
-The private key to your wallet acts as a password to your Ethereum wallet. It is the only thing stopping someone who knows your wallet address from draining your account of all of its assets!
+The private key to your wallet is a password to your Ethereum wallet. It is the only thing stopping someone who knows your wallet address from draining your account of all of its assets!
```
wackerow, line 128: My side-concern with this statement is it can give people the false impression that they need to enter this "password" when they use their account... vs a password used to encrypt/decrypt your wallet, which works differently and is more commonly the "password" users will need to actually remember or have easy access to (unlike their private key which they may never see)

I see that as out-of-scope to the current changes though, just noting for the future

wackerow (approved):
Changes look good to me! Thanks @setgree.. cc: @minimalsm Mind taking another peek here?

wackerow:
@all-contributors please add @setgree for content

Issue #8698, opened 2022-11-21
Add ethereumjs client as an execution client?


samajammin:
### Is your feature request related to a problem? Please describe.

AFAIK the @ethereumjs/client is a complete Ethereum client implementation:
https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/client

### Describe the solution you'd like

So far we list it as an EVM implementation:
https://ethereum.org/en/developers/docs/evm/#evm-implementations

But perhaps we should also add it as an execution client implementation:
https://ethereum.org/en/developers/docs/nodes-and-clients/#execution-clients

I suspect we'd want to add some caveats e.g. that it hasn't been audited & may not be a "production-ready" client.

### Describe alternatives you've considered

Not sure how to best represent it on this page / table of execution clients:
- should we have a "production-ready" column on this table (how would we define that)?
- should we have an "audited" column on this table (which links to any audits)?
- other considerations?

g11tech:
would love to see ethereumjs up there, as we believe that we are upto date with the EIP implementations in production and run any mainnet block provided pre-state is  available 

(we can do this - running mainnet blocks- in limited capacity currently by hooking up ethereumjs with an ethers provider to provide us state data to run any block). And this works in browser too.

Overall:

i) We have to achieve few milestones like snap sync etc to sync mainnet but 
  - smaller testnets we can sync and 
  - even produce blocks and 
  - process transactions (case in point: shandong testnet and previously kiln testnet), 
  - serve other client's full sync requests, broadcast/receive peer txpool announcements 
 
ii) Our client is experimental and not battle tested, but 
  - we are merge enabled via engine API (and just recently withdrawals enabled too :) )
  - can beacon sync/optimistic sync
  - serve RPC api with good coverage for eth methods
 
iii) EVM performance optimizations (and may be across all of the client)

Team has been working on all these fronts

cc @holgerd77, @acolytec3

holgerd77:
Will try to come back to this soon, just didn't have the time yet.

corwintines:
@samajammin do you think we should deviate from launchpad clients?

PR #12402, opened 2024-03-08
Add EthereumJS client as an execution client

corwintines:
## Description

- Add information for EthereumJS to nodes-and-clients page
- Add `beta` tag

Fixes: https://github.com/ethereum/ethereum-org-website/issues/8698


wackerow (approved):
Looks good... if we get more information on the state pruning for EthereumJS we can update those separately.

Otherwise agree we should add Holesky and deprecate the Goerli references.

PR #12410, opened 2024-03-09
typo: change "asset" to "assert" in glossary

ryanpwaldon:
## Description

Glossary entry for "assert" is incorrectly titled "asset".

Typo was found in 4 translations. Fixed for all.


wackerow (approved):
looks good! Thanks @ryanpwaldon 😄 

cc: @lukassim Looks like this would affect some translations, though this is a bit of a unique term since `assert` is a specific code command as well, so we may want to instruct translators to keep this English term in parentheses.

wackerow:
@all-contributors please add @ryanpwaldon for content typo fix

PR #12422, opened 2024-03-11
Fix translation error in zk rollups docs (Chinese Simplified)

akijoey:
Fix a translation error.

## Summary by CodeRabbit

- **Documentation**
	- Corrected a typo in the zk-rollups component description for improved clarity and consistency in the Chinese translation.


UNOFFICIALbgd:
Hello, as I know, all issues related with translations are handled in Crowdin project, so you should join Crowdin project (https://crowdin.com/project/ethereum-org) and suggest better translation there

wackerow:
Hey @akijoey, thank you for the suggestion here... as @UNOFFICIALbgd mentioned, all of our translation updates are handled through Crowdin. I'm unable to verify this change myself, and it will still need to be updated in Crowdin if we patch it here.

If you can stop over there to fix the problem that would help, but will also cc @lukassim here to help facilitate... Going to close for now but Luka if you think we can pull this in just say so and we can open this back up.

PR #12423, opened 2024-03-11
Fix tooltip positioning and closing in mobile

pettinarip:
## Summary by CodeRabbit

- **Refactor**
	- Improved the `Tooltip` component's behavior on mobile devices by automatically closing it upon scrolling.


src/components/Tooltip/index.tsx
```diff
@@ -14,11 +15,35 @@ export interface IProps extends PopoverProps {
...
+  const { isOpen, onOpen, onClose } = useDisclosure()
+
+  // Close the popover when the user scrolls.
+  // This is useful for mobile devices where the popover is open by clicking the
+  // trigger, not hovering.
+  useEffect(() => {
+    const handleScroll = () => {
+      if (isOpen) {
+        onClose()
+      }
```
wackerow, line 27: Consider a guard clause

```suggestion
      if (!isOpen) return
      onClose()
```

...or we could consider removing the check all together, but I think this is good

src/components/Tooltip/index.tsx
```diff
@@ -1,24 +1,57 @@
...
+  useEffect(() => {
+    let originalPosition = 0
+
+    const handleScroll = () => {
+      const delta = window.scrollY - originalPosition
+
+      // Close the popover if the user scrolls more than 80px
+      if (isOpen && Math.abs(delta) > 80) {
+        onClose()
+      }
```
wackerow, line 34: This works great! I like that it doesn't close from the slightest nudge... it feels natural

wackerow (approved):
Looks great! Thanks @pettinarip!

PR #12426, opened 2024-03-11
feat: add LocaleDateTime for markdown time

### LINKED ISSUE #12425: Markdown dates/times are not timezone responsive
> maybe date (13-Mar-2024) should be formatted according to locale

_Originally posted by @nhsz in https://github.com/ethereum/ethereum-org-website/pull/12413#discussion_r1519960486_

---

Dates and times used within markdown have always been static. A React component could be used to pass a datetime string to and parse it according to the users locale, to take into account different timezones. 

PR incoming

wackerow:
## Description
- Adds a component that parses a date/time string for use within static Markdown pages
- Component mapped to `time` html element in `MdComponents` for use within markdown pages
- Accepts `utcDateTime` prop which will be parsed as a `Date` object
- Accepts a `hideDate` OR `hideTime` prop to optionally hide one half of the resulting string
- Accepts optional `options` override to customize styling if needed

## Example usage
This PR only adds the component but does not implement it anywhere.

Example of how to use would be `<time utcDateTime="2024-04-20T01:00:00Z" />` inside a markdown file, which would render as `April 19, 2024 at 9:00:00 PM` for Eastern US timezone. Adding `hideTime` would result in `April 19, 2024`, and adding `hideDate` would render as `6:00:00 PM`. A different `locale` would adjust accordingly.

## Related Issue
- Fixes #12425

## Summary by CodeRabbit

- **New Features**
	- Introduced a `LocaleDateTime` component for displaying date and time in a localized format with customization options.

### DIFF
--- a/src/components/LocaleDateTime.tsx
+++ b/src/components/LocaleDateTime.tsx
@@ -1,0 +1,53 @@
+import { useRouter } from "next/router"
+
+type LocaleDateTimeProps = {
+  utcDateTime: string
+  hideDate?: boolean
+  hideTime?: boolean
+  options?: Intl.DateTimeFormatOptions
+}
+
+/**
+ * Renders a localized date and time component
+ * @param utcDateTime - The UTC date and time string, ie "2022-04-20T16:20:00Z"
+ * @param hideDate - Whether to hide the date portion
+ * @param hideTime - Whether to hide the time portion
+ * @param options - Options override for formatting the date and time
+ * @returns The rendered LocaleDateTime component
+ */
+const LocaleDateTime = ({
+  utcDateTime,
+  hideDate,
+  hideTime,
+  options,
+}: LocaleDateTimeProps) => {
+  if (hideDate && hideTime)
+    throw new Error(
+      "LocaleDateTime hideDate and hideTime props cannot both be true"
+    )
+
+  const { locale } = useRouter()
+  const date = new Date(utcDateTime)
+  const defaultDateOptions: Intl.DateTimeFormatOptions = {
+    month: "long",
+    day: "numeric",
+    year: "numeric",
+  }
+  const defaultTimeOptions: Intl.DateTimeFormatOptions = {
+    hour: "numeric",
+    minute: "numeric",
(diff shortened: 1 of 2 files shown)


src/components/LocaleDateTime.tsx
```diff
@@ -0,0 +1,53 @@
...
+const LocaleDateTime = ({
+  utcDateTime,
+  hideDate,
+  hideTime,
+  options,
+}: LocaleDateTimeProps) => {
+  if (hideDate && hideTime)
+    throw new Error(
+      "LocaleDateTime hideDate and hideTime props cannot both be true"
+    )
```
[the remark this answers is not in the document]
nhsz: Agree with the rabbit @wackerow, this should log but not throw

src/components/LocaleDateTime.tsx
```diff
@@ -0,0 +1,53 @@
...
+  utcDateTime: string
+  hideDate?: boolean
+  hideTime?: boolean
+  options?: Intl.DateTimeFormatOptions
+}
+
+/**
+ * Renders a localized date and time component
+ * @param utcDateTime - The UTC date and time string, ie "2022-04-20T16:20:00Z"
+ * @param hideDate - Whether to hide the date portion
```
nhsz, line 13: Do we have cases where we just want to display date and not time or viceversa?

nhsz (changes requested):
@wackerow thanks, asking to remove the throw and left a question

src/components/LocaleDateTime.tsx
wackerow: Throwing an error will occur at build, not for the user. This prevents accidentally hiding the entire date and time.

Hiding one or the other can be desired, ie the Dencun date... we don't need to show the exact time, but depending which timezone you live in this will either occur on the 13th or 14th. That was why I placed "UTC" after the date to remove that ambiguity, but this component would just render it accordingly in the users local timezone (which doesn't just affect the time, but the date as well)

nhsz (approved):

PR #12444, opened 2024-03-14
feat: setup Storybook Interaction Testing

TylerAPfledderer:
Sets up testing for storybook and applies testing for `FeedbackWidget` and `QuizWidget`

- Adds `@storybook/test`
- Adds interaction test for `FeedbackWidget`, which creates a snapshot capture for the modal.
- Adds mock user interaction of the `QuizWidget`, which runs a quiz with all correct answers.


TylerAPfledderer:
> Hey @TylerAPfledderer, checking this out... 
> 
> image
> 
> The layer-2 quiz seems to work great, but the others are not yet...
> 
> image
> 
> Were we looking to add those here as well or in a separate task?
> 
> image
> 
> FeedbackWidget modal passing on my end 👍 

To clarify, how you are checking the other quizzes? Are you just replacing the name of the key in the story?

I was only considering just one of the quiz banks to be able to verify that the object shape is valid. If one works, then they should all work. 🤔

I didn't want to run all of them because that would mean creating a story per quiz, and that would bloat the amount of snapshots created. 

I would add one to verify when some questions are answered incorrectly, though!

These interaction tests should be about how the user is navigating the quizzes and what they should be seeing, nothing more.

If you are seeing an inconsistency, then I wonder if my implementation of the testing is wrong.

wackerow:
> To clarify, how you are checking the other quizzes? Are you just replacing the name of the key in the story?

Under "Controls" > `quizKey*` downdown:

image

> If one works, then they should all work. 🤔

Okay yeah, hah sounds reasonable

> I would add one to verify when some questions are answered incorrectly, though!

Agree, that would be a nice case to test

Thanks for clarifying!

wackerow:
@TylerAPfledderer Did you want to include a test case for a wrong answer in this PR, or keep that separate?

TylerAPfledderer:
> @TylerAPfledderer Did you want to include a test case for a wrong answer in this PR, or keep that separate?

I probably should add one here.

I would simply call on the layer-2 question bank again and randomly select two questions to have an incorrect answer

TylerAPfledderer:
> > To clarify, how you are checking the other quizzes? Are you just replacing the name of the key in the story?
> 
> Under "Controls" > `quizKey*` downdown:
> 
> image
> > If one works, then they should all work. 🤔
> 
> Okay yeah, hah sounds reasonable
> 
> > I would add one to verify when some questions are answered incorrectly, though!
> 
> Agree, that would be a nice case to test
> 
> Thanks for clarifying!

Cool!

Now that I finally had a chance to properly check this. I have this setup constrained to just pulling content data for the layer 2 quiz in the story file.

wackerow:
Great, looking good... 

image

I noticed this string is showing up by it's key... would we fix that here? I don't see that bug in production

TylerAPfledderer:
> Great, looking good... 
> 
> image
> 
> I noticed this string is showing up by it's key... would we fix that here? I don't see that bug in production

I haven't quite figured out why it's an issue here, because the intl file this key is a part of is exposed in storybook (`common.json`). So I wouldn't know if it's a quick fix.

If it's not a quick fix, then it should be addressed separately; I don't think it's a blocker for this PR.

TylerAPfledderer:
Outstanding tasks:

- Update the PR with `dev` branch to make sure fix from #12781 is included
- Add interaction test for `Tooltip` to ensure the rendered popover is viewable in a snapshot

TylerAPfledderer:
Sticking back into draft for a brief time. Still pesky issues with the tooltip snapshot. Will put to "Ready for review" when either the issue is resolved, or that test is scrapped and to be addressed in a future PR.

Currently working with a developer from Chromatic to help figure out the issue.

.storybook/preview.ts
```diff
@@ -72,6 +79,7 @@ const preview: Preview = {
           },
         }
       }, {}),
+      defaultViewport: "2xl",
```
TylerAPfledderer, line 82: Revert this. Not needed at this time.

```suggestion
```

TylerAPfledderer:
Ready to go here, but seeing unexpected build errors unrelated to the PR

wackerow (approved):

wackerow (approved):

nloureiro:
@wackerow @TylerAPfledderer, what kind of approval can I do here? What should I look for?

TylerAPfledderer:
> @wackerow @TylerAPfledderer, what kind of approval can I do here? What should I look for?

For a PR like this, I don't think we need your explicit approval. No component affected should have their design altered.

PR #12455, opened 2024-03-14
Add auto-linked headers on glossary page [Fixes #12393]

### LINKED ISSUE #12393: Missing auto-linked headers on glossary page
### Describe the bug

The glossary page was previously entirely markdown, which had custom header IDs for each glossary entry header, ie `### Account {#account}`. This has been converted to using a custom component for each term entry, ie `<GlossaryDefinition term="account" />`.

This component renders a header that _does_ contain an **id** (ie. `<h3 ... id="account">`), but does _not_ include the auto-linking (link to self) like the section headers do.

### To reproduce

1. Go to https://ethereum.org/en/glossary/
2. Hover over "A", and notice the link icon that appears, which contains a link to itself
3. Hover over a glossary term such as "Account", and notice this _does not_ contain a link to itself

### Expected behavior

All headers on this page should have an auto-link to the ID of that section, including a hover effect to quickly access this link.

### Screenshots

image

When hovering over "A"... but this does not occur when hovering over "Account"

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes
- [X] No

checkomkar:
Fixes [#12393]

## Summary by CodeRabbit

- **New Features**
	- Enhanced the Glossary Definition component to display a link icon on hover, improving user interactivity and visual feedback.
	- Added a link icon to the Glossary Definition component for better user experience.


wackerow (changes requested):
Hey @checkomkar, thanks for this!

Couple things... we seem to be recreating a lot of logic and styling that has already been built elsewhere...

- We should use the same icon as the others (`CiLink` from `react-icons/ci`), and the size should match
- The word itself should not be clickable... the link icon should appear while hovering the word, but only the icon should contain the link itself
- Needs a `scrollMarginTop` so the header isn't hidden by the nav bar when navigating to it

This is all implemented inside `MdComponents` for the standard markdown headers, ie `Heading3`. These utilize a reusable `IdAnchor` component (also in that file) that handles the link and icon:

image

This component is then placed inside a header, along with `{...commonHeadingProps(id)}` to get the props needed for styling, and to assign the appropriate ID.

I'd ask that we take another look here and try to simplify, reusing components where possible and removing the redundant logic. Let us know if you're having issues!

(Last note, the files need to be linted/formatted before they will build correctly... in particular, the imports at top need to follow a certain order... VS Code formatter can do this automatically, but if you're not sure how to set this up we can handle fixing that up when ready)

checkomkar:
> Hey @checkomkar, thanks for this!
> 
> Couple things... we seem to be recreating a lot of logic and styling that has already been built elsewhere...
> 
> * We should use the same icon as the others (`CiLink` from `react-icons/ci`), and the size should match
> * The word itself should not be clickable... the link icon should appear while hovering the word, but only the icon should contain the link itself
> * Needs a `scrollMarginTop` so the header isn't hidden by the nav bar when navigating to it
> 
> This is all implemented inside `MdComponents` for the standard markdown headers, ie `Heading3`. These utilize a reusable `IdAnchor` component (also in that file) that handles the link and icon:
> 
> image
> This component is then placed inside a header, along with `{...commonHeadingProps(id)}` to get the props needed for styling, and to assign the appropriate ID.
> 
> I'd ask that we take another look here and try to simplify, reusing components where possible and removing the redundant logic. Let us know if you're having issues!
> 
> (Last note, the files need to be linted/formatted before they will build correctly... in particular, the imports at top need to follow a certain order... VS Code formatter can do this automatically, but if you're not sure how to set this up we can handle fixing that up when ready)

Thank you @wackerow for the detailed review. Will make the changes as suggested. Also with regards to linting, I tried through VS code somehow it's not working, the import order doesn't seem to sort. Is there any extension that I am supposed to use or any lint command?

wackerow:
> Thank you @wackerow for the detailed review. Will make the changes as suggested. Also with regards to linting, I tried through VS code somehow it's not working, the import order doesn't seem to sort. Is there any extension that I am supposed to use or any lint command?

@checkomkar Try `yarn lint:fix` to handle this... If that doesn't work, try Cmd/Ctrl+Shift+P to open the command palette and type/select "ESLint: Fix all auto-fixable Problems"... this should apply the same rules.

checkomkar:
> Hey @checkomkar, thanks for this!
> 
> Couple things... we seem to be recreating a lot of logic and styling that has already been built elsewhere...
> 
> * We should use the same icon as the others (`CiLink` from `react-icons/ci`), and the size should match
> * The word itself should not be clickable... the link icon should appear while hovering the word, but only the icon should contain the link itself
> * Needs a `scrollMarginTop` so the header isn't hidden by the nav bar when navigating to it
> 
> This is all implemented inside `MdComponents` for the standard markdown headers, ie `Heading3`. These utilize a reusable `IdAnchor` component (also in that file) that handles the link and icon:
> 
> image
> This component is then placed inside a header, along with `{...commonHeadingProps(id)}` to get the props needed for styling, and to assign the appropriate ID.
> 
> I'd ask that we take another look here and try to simplify, reusing components where possible and removing the redundant logic. Let us know if you're having issues!
> 
> (Last note, the files need to be linted/formatted before they will build correctly... in particular, the imports at top need to follow a certain order... VS Code formatter can do this automatically, but if you're not sure how to set this up we can handle fixing that up when ready)

Hey @wackerow I tried importing and reusing the components as you suggested, but it was causing cyclic import error so I just copied the `IdAnchor` into the `GlossaryDefinition` index file and this is working. Please let me know if I can better this. Thanks.

wackerow:
> Hey @wackerow I tried importing and reusing the components as you suggested, but it was causing cyclic import error so I just copied the `IdAnchor` into the `GlossaryDefinition` index file and this is working. Please let me know if I can better this. Thanks.

@checkomkar To use the existing `IdAnchor` component, we just need to export it where it's defined by adding `export` before `const IdAnchor ...` inside `MdComponents.tsx`, ie:

```tsx
export const IdAnchor = ({ id }: { id?: string }) => {
```

Then we can import this inside `GlossaryDefinition/index.tsx` with:

```tsx
import { IdAnchor } from "@/components/MdComponents"
```

We can then remove the entire `IdAnchor` function from this file, and remove the unused imports (`CiLink`, `Icon` and `Link`). Would run `yarn lint:fix` again after this.

wackerow:
Just checking in, @checkomkar let me know if you want a hand getting this over the line

checkomkar:
> > Hey @wackerow I tried importing and reusing the components as you suggested, but it was causing cyclic import error so I just copied the `IdAnchor` into the `GlossaryDefinition` index file and this is working. Please let me know if I can better this. Thanks.
> 
> @checkomkar To use the existing `IdAnchor` component, we just need to export it where it's defined by adding `export` before `const IdAnchor ...` inside `MdComponents.tsx`, ie:
> 
> ```tsx
> export const IdAnchor = ({ id }: { id?: string }) => {
> ```
> 
> Then we can import this inside `GlossaryDefinition/index.tsx` with:
> 
> ```tsx
> import { IdAnchor } from "@/components/MdComponents"
> ```
> 
> We can then remove the entire `IdAnchor` function from this file, and remove the unused imports (`CiLink`, `Icon` and `Link`). Would run `yarn lint:fix` again after this.

I get this cyclic import error whenever I am trying to import `IdAnchor` in  `GlossaryDefinition/index.tsx`, not sure how to resolve this.

image

wackerow:
Okay I think I know what's going on, two things...
1- Make sure you added `export` before `const IdAnchor` on line 66 inside `MdComponents.tsx`
2- Make sure you're using a _named import_ and not a default:

```tsx
// This:
import { IdAnchor } from "@/components/MdComponents"
// Not this:
import IdAnchor from "@/components/MdComponents"
```

That file had a `default` export as well at the bottom, which is an object with all the markdown components, and would be assigned to `IdAnchor` with the "Not this" approach above. You'd get a circular error in that case because that object also has other components that do form a circular reference with that `TooltipLink` link being imported into `GlossaryDefinition`.

You should be able to use the `{ IdAnchor }` approach though after adding the keyword `export` before that function declaration.

wackerow:
Checking in... Any luck @checkomkar? Happy to try and get this over the line if you're not available

wackerow (approved):
Just extracted the IdAnchor component as it's own file component to avoid any circular reference issues... looks good! Pulling in, thanks a lot @checkomkar!

wackerow:
@all-contributors please add @checkomkar for code

PR #12456, opened 2024-03-14
Max blobs per block should be 6 not 1

blmalone:
Blobs per block is not 1 but 6.

## Description

[eip-4844](https://eips.ethereum.org/EIPS/eip-4844#throughput): states that the max size is 0.75 MB which is equivalent to 6, [125kb blobs](https://x.com/VitalikButerin/status/1764139239233749047).

## Summary by CodeRabbit

- **Documentation**
	- Updated the roadmap section on Danksharding to reflect the increase in the number of blobs attached to blocks from 6 in Proto-Danksharding to 64 in full Danksharding, along with the necessary adjustments for consensus client operations.


wackerow (approved):
https://github.com/ethereum/consensus-specs/blob/5d80b1954a4b7a121aa36143d50b366727b66cbc/presets/mainnet/deneb.yaml#L10C1-L10C20
LGTM! Thanks @blmalone

wackerow:
@all-contributors please add @blmalone for content correction

PR #12479, opened 2024-03-18
Performance: lazy load the `Emoji` & `SearchModal` components

pettinarip:
## Description

Lazy load these components to separate these libs into their own chunks of js bundles.

Currently, they are part of the `app` (main) bundle that is loaded by all the pages and blocking the main thread. We can load these libs async and minimize the work done in it.

`app.js` bundle before (~1.01Mb parsed):
image

`app.js` bundle after (~788Kb parsed):
image

## Summary by CodeRabbit

- **Performance Improvements**
	- Improved loading times by implementing dynamic imports for the Emoji and Search Modal components, enhancing code splitting and lazy loading.

### DIFF
--- a/src/components/Emoji.tsx
+++ b/src/components/Emoji.tsx
@@ -4,8 +4,9 @@ import { Box, type BoxProps } from "@chakra-ui/react"
 
 import { IS_DEV } from "@/lib/utils/env"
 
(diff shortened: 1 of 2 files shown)


wackerow (approved):
Lgtm! Thanks @pettinarip!

PR #12482, opened 2024-03-18
Performance: avoid mono font preload

pettinarip:
## Description

As the `mono` font is not critical for the UI, lets load it once the page is ready.

## Summary by CodeRabbit

- **Performance Improvements**
	- Improved font loading speed by preloading `inter` and `mono` fonts.


wackerow (approved):
Love these little performance improvements, thanks @pettinarip!

PR #12484, opened 2024-03-18
fix: return home link on 404 page and netlify redirect [FIXES #12471]

### LINKED ISSUE #12471: Unknown paths (404s) redirect user to English
### Describe the bug

In the 404 page with text "We couldn't find that page" the link `return home` is expected to redirect to  "https://ethereum.org/en/" or "https://ethereum.org/" but it redirects you to "https://ethereum.org/en/enindex/" which again is a non-existent page 

### To reproduce

1. Visit a 404 page , i used "https://ethereum.org/en/1"
2. Click on `return home` link or ethereum logo(top-left)
3. Now observe you've been redirected to "https://ethereum.org/en/enindex" instead of home page 

### Expected behavior

When clicking on `return home` link or ethereum logo from a 404 page it should redirect to home page instead of "https://ethereum.org/en/enindex" page 

### Screenshots

404Page
404enIndex

- Browser: Brave

-Browser: Brave

### Additional context

_No response_

- [ ] Yes
- [X] No

tomasgrusz:
- add redirect from faulty `/enindex` page to home page
- change link component in `404.tsx` to redirect browser directly to `/en`

## Description

The issue describing a redirect to home page turning into a 404 instead is working fine within local environment.
Therefore it seems that it triggers somewhere within prod/Netlify setup.
These changes modify the link to hopefully redirect to homepage without triggering the error.
_(there should be no stylistic differences introduced with this change)_

Also added a Netlify redirect, which should push all false 404s to the home page:
```
/enindex/ /en/ 301!
```

[This article](https://answers.netlify.com/t/how-to-get-netlify-to-point-to-an-index-html-file-that-is-in-a-folder-that-is-not-the-main-folder/14146) may also be worth taking a look at. The support guide gives great insights into how this issue can happen with integrations between Netlify and JS-based navigation such as Next.js.

## Related Issue
-Fixes #12471

## Summary by CodeRabbit

- **New Features**
	- Implemented a new redirect rule to enhance user navigation.
- **Bug Fixes**
	- Improved the "return home" link on the 404 page for better usability and consistency with the app's design system.

### DIFF
--- a/public/_redirects
+++ b/public/_redirects
@@ -173,3 +173,5 @@
 /*/guides/how-to-register-an-ethereum-account /:splat/guides/how-to-create-an-ethereum-account/ 301!
 
 /*/deprecated-software /:splat/dapps/ 301!
+
+/enindex/ /en/ 301!
--- a/src/pages/404.tsx
+++ b/src/pages/404.tsx
@@ -1,10 +1,9 @@
 import type { GetStaticProps } from "next"
 import { serverSideTranslations } from "next-i18next/serverSideTranslations"
-import { Box, Flex, Heading, Text } from "@chakra-ui/react"
+import { Box, Flex, Heading, Link, Text } from "@chakra-ui/react"
 
 import { BasePageProps } from "@/lib/types"
 
-import InlineLink from "@/components/Link"
 import MainArticle from "@/components/MainArticle"
 import Translation from "@/components/Translation"
 
@@ -37,9 +36,9 @@ const NotFoundPage = () => (
       </Heading>
       <Text mb={8}>
         <Translation id="try-using-search" />{" "}
-        <InlineLink href="/">
+        <Link href="/" isExternal={false}>
           <Translation id="return-home" />
-        </InlineLink>
+        </Link>
         .
       </Text>
     </Box>


tomasgrusz:
> ### ✅ Deploy Preview for _ethereumorg_ ready!
> Name	Link
> 🔨 Latest commit	[eed3f33](https://github.com/ethereum/ethereum-org-website/commit/eed3f330da54e40d9e99945607a16d93d299f3af)
> 🔍 Latest deploy log	https://app.netlify.com/sites/ethereumorg/deploys/65f8754e8040070008b50f5f
> 😎 Deploy Preview	https://deploy-preview-12484--ethereumorg.netlify.app
> 📱 Preview on mobile	
> Toggle QR Code...
> _To edit notification comments on pull requests, go to your [Netlify site configuration](https://app.netlify.com/sites/ethereumorg/configuration/deploys#deploy-notifications)._

seems to be working fine on deploy preview

wackerow:
Hey @tomasgrusz, thanks again for jumping in... I'm not sure this is the approach we want to be taking... this would be a bandaid, but we need to fix the underlying issue. We need to investigate and fix why our 404 page is behaving this way.

I suspect it has something to do with a relatively-recent update related to our "default locale/language"... We previously used English as the default locale, but this move all of our English paths from `/en/some/path/` to `/some/path` which we don't want for a variety of reasons... We now use a fake locale called `default` as the default, which I suspect is causing issues here:

image

@pettinarip Mind weighing in here? Curious if you have any thoughts as to how this could be impacting our 404 page, since it seems to be using `default` for it's locale/language, regardless if a language path exists

wackerow (changes requested):
Hey @tomasgrusz, thanks again for jumping in... I'm not sure this is the approach we want to be taking... this would be a bandaid, but we need to fix the underlying issue. We need to investigate and fix why our 404 page is behaving this way.

I suspect it has something to do with a relatively-recent update related to our "default locale/language"... We previously used English as the default locale, but this move all of our English paths from `/en/some/path/` to `/some/path` which we don't want for a variety of reasons... We now use a fake locale called `default` as the default, which I suspect is causing issues here:

image

@pettinarip Mind weighing in here? Curious if you have any thoughts as to how this could be impacting our 404 page, since it seems to be using `default` for it's locale/language, regardless if a language path exists

pettinarip:
> Hey @tomasgrusz, thanks again for jumping in... I'm not sure this is the approach we want to be taking... this would be a bandaid, but we need to fix the underlying issue. We need to investigate and fix why our 404 page is behaving this way.
> 
> I suspect it has something to do with a relatively-recent update related to our "default locale/language"... We previously used English as the default locale, but this move all of our English paths from `/en/some/path/` to `/some/path` which we don't want for a variety of reasons... We now use a fake locale called `default` as the default, which I suspect is causing issues here:
> 
> image
> @pettinarip Mind weighing in here? Curious if you have any thoughts as to how this could be impacting our 404 page, since it seems to be using `default` for it's locale/language, regardless if a language path exists

Hmm not sure what is really happening here yet but [this](https://github.com/i18next/next-i18next/issues/2166) could be related. Haven't tested it yet. Will take a look later this week.

wackerow:
Just flipped this to "draft" mode in the meantime... if/when ready just hit the "Ready for review" button

wackerow:
Closing this out; can address in new PR.

PR #12436, opened 2024-03-13
Discord GFI webhook

pettinarip:
## Summary by CodeRabbit

- **New Features**
	- Added a GitHub Actions workflow to send "good first issues" from GitHub to Discord via webhook.
	- Introduced a script for fetching and formatting "good first issues" for Discord notifications.
- **Chores**
	- Updated `.env.example` with configurations for Discord webhook and GitHub token.

PR #12500, opened 2024-03-19
Discord GFI webhook

pettinarip:
## Description

Simpler alternative to #12436 using the GH webhooks service. Whenever an issue changed, the new api route `/api/gfi-issues-webhook` is going to be called.

## Summary by CodeRabbit

- **New Features**
	- Added a feature to convert GitHub issue labels to human-readable text.
	- Introduced a webhook handler for notifying a Discord channel when a new "good first issue" is labeled on GitHub.

### DIFF
--- a/src/lib/utils/gh.ts
+++ b/src/lib/utils/gh.ts
@@ -45,13 +45,21 @@ export const getLastModifiedDateByPath = (path: string): string => {
   return extractDateFromGitLogInfo(logInfo)
 }
 
-const LABELS_TO_SEARCH = ["content", "design", "dev", "doc", "translation"]
+const LABELS_TO_SEARCH = [
+  "content",
+  "design",
+  "dev",
+  "doc",
+  "translation",
+  "event",
+]
 const LABELS_TO_TEXT = {
   content: "content",
   design: "design",
   dev: "dev",
   doc: "docs",
   translation: "translation",
+  event: "event",
 }
 
 // Given a list of labels, it returns a new array with the labels that match the
--- a/src/lib/utils/gh.ts
+++ b/src/lib/utils/gh.ts
@@ -52,8 +52,9 @@ const LABELS_TO_SEARCH = [
   "doc",
   "translation",
   "event",
-]
-const LABELS_TO_TEXT = {
+] as const
+
+const LABELS_TO_TEXT: Record<(typeof LABELS_TO_SEARCH)[number], string> = {
   content: "content",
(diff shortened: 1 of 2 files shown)


src/lib/utils/gh.ts
```diff
@@ -44,3 +44,35 @@ export const getLastModifiedDateByPath = (path: string): string => {
...
+      if (labelIndex === -1) {
+        return
+      }
+
+      const labelMatched = LABELS_TO_SEARCH[labelIndex]
+      return LABELS_TO_TEXT[labelMatched]
+    })
+    .filter(Boolean)
+    .join(", ")
+}
```
wackerow, line 78:
```suggestion
export const rawLabelsToText = (labels: string[]) => {
  const list = labels
    .map((label) => {
      const labelIndex = LABELS_TO_SEARCH.findIndex((l) =>
        label.toLocaleLowerCase().includes(l)
      )

      if (labelIndex === -1) {
        return
      }

      const labelMatched = LABELS_TO_SEARCH[labelIndex]
      return LABELS_TO_TEXT[labelMatched]
    })
    .filter(Boolean)

  return new Intl.ListFormat(
    DEFAULT_LOCALE,
    { style: "long", type: "conjunction" }
  ).format(list),
}
```

Just a small suggestion if we want to make the list a little more "readable" with natural language. (An array of ["dev", "content", "design"] would turn into "dev, content and design"... [Documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat))

src/pages/api/gfi-issues-webhook.ts
wackerow, line 1: @pettinarip Am I correct that this isn't implemented anywhere yet?

pettinarip: Hmm not sure if I follow you on this. Inside our project? this api doesn't exist nor the logic itself does.

pettinarip: Maybe you are confused with the similar logic in #12436. I'm sorry about that, we have decided to go with this implementation for now. I have closed the other one.

src/lib/utils/gh.ts
pettinarip: Won't implement this for now since we are going to just show one label in the new implementation. Great idea though, didn't know about it, thanks for sharing.

wackerow (approved):

PR #12501, opened 2024-03-19
Performance: use inter static fonts

pettinarip:
## Description

Replaces the current variable inter font with static ones to improve loading times.

The variable font we are using is too heavy ~800kb. Loading the necessary static subset will improve the overall loading times.

Since `next/local` doesn't support `unicode-range` atm, we are loading and preloading the fonts manually.

## Summary by CodeRabbit

- **Refactor**
	- Updated the font system across the app for improved readability and performance.
- **New Features**
	- Added support for preloading the Inter font for faster text rendering.


public/fonts/inter/UcC73FwrK3iLTeHuS_fvQtMwCp50KnMa0ZL7SUc.woff2
wackerow, line 1: Hm... is there any way we could clean up the typeface file names here so they're descriptive?

wackerow (changes requested):
Bummer that the variable font is so bulky, but agree that performance takes precedent here. My only request would be to use descriptive filenames instead of hashes so they're easier to reference/trace/understand. Otherwise looks good!

nhsz:
@pettinarip This change causes to execute more requests (and some fonts are failing), we would need to measure it to be sure its improving the performance, example of fonts loading on the homepage below

Screen Shot 2024-03-20 at 18 23 37

public/fonts/inter/UcC73FwrK3iLTeHuS_fvQtMwCp50KnMa0ZL7SUc.woff2
pettinarip: Ah good call. Yeah, I can definitely do that.

pettinarip:
> @pettinarip This change causes to execute more requests (and some fonts are failing), we would need to measure it to be sure its improving the performance, example of fonts loading on the homepage below
> 
> Screen Shot 2024-03-20 at 18 23 37

You might have some of them cached or coming from a different source (extenstions perhaps?) because the ones that are failing are not being declared in this PR. There are others there in the screenshot that are also not part of this PR, they are probably coming from Netlify, for the Netlify UI elements of the preview deploy. Try to test this in incognito mode and lmk the results you get 🙏🏼 

The other fonts, are ok, expected. The first 2 fonts are the most relevant fonts loaded here. The latin one is preloaded, the rest are loaded after the website is fully loaded and ready, so that is not going to affect the overall performance of the site.

wackerow (approved):
image

This is working great for me locally with fresh cache, and no Netlify additions.

nhsz (approved):
LGTM, thanks!

PR #12541, opened 2024-03-22
chore: make Filters clickable on mobile to open filters menu

nhsz:
This PR makes the "Filters" label clickable to allow opening filters menu on mobile

### DIFF
--- a/src/components/FindWallet/WalletTable/index.tsx
+++ b/src/components/FindWallet/WalletTable/index.tsx
@@ -244,6 +244,7 @@ const WalletTable = ({
               textTransform="uppercase"
               cursor="pointer"
               onClick={onOpen}
+              as="button"
             >
               {`${t("page-find-wallet-filters")} (${
                 walletsListingCount(filters) +
--- a/src/components/FindWallet/WalletTable/index.tsx
+++ b/src/components/FindWallet/WalletTable/index.tsx
@@ -208,6 +208,7 @@ export type WalletTableProps = {
   resetFilters: () => void
   resetWalletFilter: React.MutableRefObject<() => void>
   walletData: WalletData[]
+  onOpen: () => void
 }
 
 const WalletTable = ({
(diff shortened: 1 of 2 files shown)


src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -233,12 +235,15 @@ const WalletTable = ({
...
+            {/* Displayed on mobile only */}
             <Text
               hideFrom="lg"
               lineHeight={1.6}
               fontSize="md"
               color="primary.base"
               textTransform="uppercase"
+              cursor="pointer"
+              onClick={onOpen}
             >
```
wackerow, line 247: Just curious, is there any reason we don't want to use a button here? Seems more appropriate from an a11y standpoint

wackerow (approved):
Looks good! Just left one question that I don't see as a blocker, but wanted to raise... can address separately.

src/components/FindWallet/WalletTable/index.tsx
pettinarip: Agreed. Now, we are basically turning this component into a button.

nhsz: @wackerow It was defined in the design that should behave like a button but without the button styles. I think I can add the `as={button}` prop to improve a11y

wackerow: Okay cool.. yeah that raising a small design system question... we have `variant="link"` for our buttons, but this will bold and underline the text (inconsistent with what we want here)... @nloureiro Should we have another variant that is even more simplified for buttons like this? Or should we convert this one to use the `link` variant styling?

nhsz: @wackerow do you think that could be a blocker for this PR or could be address later?

wackerow: Nope, not at all! We can address design system stuff separately, this works

wackerow (approved):

PR #12555, opened 2024-03-25
IconButton `isSecondary` warning

pettinarip:
Warning seen in the console
```
Warning: React does not recognize the isSecondary prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase issecondary instead. If you accidentally passed it from a parent component, remove it from the DOM element. at button
```

## Description

Fixes this warning by using our custom `IconButton` component instead of chakra native one that knows how to implement the `isSecondary` prop.


nhsz (approved):
LGTM, thanks! 🙌

PR #12568, opened 2024-03-26
Update oasis_std link to reference latest version on `/rust` page

elec7reify:
Reference to the latest OASIS API version.


wackerow (approved):
Looks good! Thanks @elec7reify, I just added a commit making the same change to the non-English versions of this page. Pulling in

Issue #12590, opened 2024-03-27
Release blocking - Find wallet on Safari not working when clicking a wallet or more information


corwintines:
### Describe the bug

When I use the find wallet page on Safari, I go to get more information on a wallet and the site crashes. 

### To reproduce

1. Go to https://deploy-preview-12586--ethereumorg.netlify.app/en/wallets/find-wallet/ in Safari
2. Scroll down and click on Rainbow wallet
3. See the crash

https://github.com/ethereum/ethereum-org-website/assets/15589226/c1e50fd5-a32b-4c0b-a00e-f6a394a61240

### Expected behavior

I should be able to get more info on all wallets in Safari

### Would you like to work on this issue?

- [x] Yes
- [ ] No

wackerow:
Closed with #12593

PR #12593, opened 2024-03-27
fix: wallets data dates consistency causing crashes in Safari

nhsz:
This PR fixes wallets data inconsistencies related to dates formatting that are causing crashes in Safari when opening the `MoreInfo` section

## Description

- fixes `last_updated` format inconsistency where needed
- removes unused `wallet_live_date` wallet data prop that also has inconsistent formatting

---

Fixes https://github.com/ethereum/ethereum-org-website/issues/12590


.github/ISSUE_TEMPLATE/suggest_wallet.yaml
```diff
@@ -48,13 +48,6 @@ body:
       description: Please provide a URL (e.g. to the website of the wallet).
     validations:
       required: true
-  - type: input
-    id: wallet_live_date
-    attributes:
-      label: When did the wallet go live to users?
-      description: Please provide a date when the wallet was usable by the public
-    validations:
-      required: true
```
wackerow, line 57: @nhsz I think we still need this as part of the initial issue request... Our [listing policy](https://ethereum.org/en/contributing/adding-wallets/#the-must-haves) requires the product be live for 6 months, and this would be how we check that.

Shouldn't have any impact on the other changes

wackerow (changes requested):

.github/ISSUE_TEMPLATE/suggest_wallet.yaml
corwintines, line 57: I would add this back in.

I think we need it given listing policy requires 6 months. While we might not need it in the actual wallet data file, its good information to have in the issues being filed.

corwintines (changes requested):
Only change request I have. Im fine with getting rid of the data in the wallet-data file, but this is useful when reviewing issues being submitted.

wackerow (approved):
Looks good, thanks @nhsz!

corwintines (approved):

PR #12626, opened 2024-04-01
Typo: Corrected Ethereum.org -> ethereum.org on Design contribution page

HiroyukiNaito:
Corrected Ethereum.org -> ethereum.org on Design contribution page

## Description

Corrected Ethereum.org -> ethereum.org on Design contribution page

## Related Issue

None

### DIFF
--- a/public/content/contributing/design/index.md
+++ b/public/content/contributing/design/index.md
@@ -6,7 +6,7 @@ lang: en
 
 # Design contribution to ethereum.org {#design-contributions}
 
-Design is a critical component of any project, and by devoting your time and design skills to Ethereum.org, you can help to improve the user experience for our visitors. Contributing to an open-source project provides an opportunity to gain relevant experience and develop your skills in a collaborative environment. You will have the chance to work with other designers, developers, and community members, all of whom will have their own unique perspectives and insights.
+Design is a critical component of any project, and by devoting your time and design skills to ethereum.org, you can help to improve the user experience for our visitors. Contributing to an open-source project provides an opportunity to gain relevant experience and develop your skills in a collaborative environment. You will have the chance to work with other designers, developers, and community members, all of whom will have their own unique perspectives and insights.
 
 Ultimately, this is a great way to build a diverse and impressive portfolio that showcases your design skills.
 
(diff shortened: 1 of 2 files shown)


public/content/contributing/design/index.md
```diff
@@ -30,7 +30,7 @@ Provide feedback on our website by:
 
 ### <Emoji text=":three:" size={1} /> &nbsp;Find design related issues on the website and report them {#report-design-issues}
 
-Ethereum.org is a fast growing website with many features and content. Some of the UI can easily become obsolete or could be improved. If you encounter any such instance, please report it so that it gets our attention.
+ethereum.org is a fast growing website with many features and content. Some of the UI can easily become obsolete or could be improved. If you encounter any such instance, please report it so that it gets our attention.
```
wackerow, line 33: Hm... Looking at this, we use a split throughout the repo when we start a sentence with the domain/name... about 47 use lowercase, and about 99 capitalize.

Personally think starting a sentence with a lowercase letter always looks awkward and would advocate for using "Etheruem.org" in these cases.

@minimalsm Thoughts here?

```suggestion
Ethereum.org is a fast growing website with many features and content. Some of the UI can easily become obsolete or could be improved. If you encounter any such instance, please report it so that it gets our attention.
```

HiroyukiNaito: I thought `ethererm.org` is started with always lowercase.
Look like this page https://ethereum.org/en/about/.
Similar to iPad and iPhone. 
If we use `IPhone is very cool` looks awkward.

HiroyukiNaito: Or maybe it needs to be aligned with Ethereum.org as a proper noun for preventing this situation.

HiroyukiNaito: I opened the discussion about this.

https://github.com/ethereum/ethereum-org-website/discussions/12669

wackerow: Great thanks! We don't usually use GitHub "Discussions" so I'm honestly not sure how much traction it'll get over there.

This isn't something I personally have a strongly held opinion on, was just noting above that we more commonly use "Ethereum.org" over "ethereum.org" when starting a sentence. I tend to prefer consistency above all.

I agree with your point about how "iPhone" is used, not "IPhone", but still feels a little strange I guess, but not a big deal.

cc: @minimalsm Did you have any thoughts on this? We have always tended to use the lowercase domain as the official name, so I wouldn't be against just formalizing that and keeping that pattern even when starting a sentence.

Issue #12552, opened 2024-03-24
Add security policy to repo


darigovresearch:
### Is your feature request related to a problem? Please describe.

There appears to be no way to understand what to do if there is a security bug

### Describe the solution you'd like

Add a policy to the security tab - https://github.com/ethereum/ethereum-org-website/security

### Describe alternatives you've considered

Can be in the readme but the security tab is probably best

### Would you like to work on this issue?

- [X] Yes
- [ ] No

wackerow:
Hey @darigovresearch, what security bugs are you referring to? Security bugs with ethereum.org?  Or Ethereum?

- For "Ethereum" as a protocol, there is the bug bounty page here: https://ethereum.org/en/bug-bounty
- For "ethereum.org" the website, all bugs are handled through GitHub

The `/security` page is aimed at providing security tips for users in this space, but wasn't really intended to be a place for handling bug reports. Would you mind elaborating?

wackerow:
Oops, sorry the link you posted for the GitHub repo security page, not ethereum.org/security, my bad... Can ignore my statement about the `/security` page, but same question persists which bugs you're referring to here.

darigovresearch:
Hi @wackerow, thanks for the quick response. The issue was mainly about Security bugs with the ethereum.org website. The generic GitHub issues may not be ideal to report security bugs with the site as the issue if published may allow for vulnerabilities to be exploited before a maintainer can assess and patch it correctly.

wackerow:
Got it. You're talking about the root level [`SECURITY.md`](https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) file it looks like.

GitHub's initial template is this:

```markdown
# Security Policy

## Supported Versions

Use this section to tell people about which versions of your project are
currently being supported with security updates.

| Version | Supported          |
| ------- | ------------------ |
| 5.1.x   | :white_check_mark: |
| 5.0.x   | :x:                |
| 4.0.x   | :white_check_mark: |
| < 4.0   | :x:                |

## Reporting a Vulnerability

Use this section to tell people how to report a vulnerability.

Tell them where to go, how often they can expect to get an update on a
reported vulnerability, what to expect if the vulnerability is accepted or
declined, etc.
```

I'm not sure the supported versions table would be very useful here, but the later could be helpful.

[...]

PR #12630, opened 2024-04-02
Create SECURITY.md

Sintayew4:


Sintayew4:
#12630

wackerow:
Currently being worked on via issue #12552. PR appears to be the default template from GitHub which is not what we're looking for. Closing out.

PR #12613, opened 2024-04-01
French import

corwintines:
## Description

Buckets: Use Ethereum, Use cases, Staking pages, Learn pages, Roadmap, Community, Foundational docs

PR #12610, opened 2024-04-01
Chinese Simplified import

corwintines:
## Description

Buckets: Essentials, Use Ethereum, Use Cases, Staking pages, Learn pages, Roadmap, Community, Foundational docs, Foundational docs - nodes and clients, Foundational docs - proof-of-stake, Docs - tech stack pages, Smart contracts - Basics, Advanced docs, Advanced docs - Scaling, Research documentation, Misc, Contributing

PR #12428, opened 2024-03-11
Deprecate /languages page

wackerow:

PR #12636, opened 2024-04-02
revert: removal of old page-languages ns strings

wackerow:
## Description
The `page-languages.json` namespace was recently merged into `common.json` in PR:

- #12428

The Crowdin imports in 

- #12611
- #12610
- #12613

are not yet updated with these changes, which accidentally removed these strings.

Affects strings with keys beginning with `language-` or `page-languages-` -- see [diff](https://github.com/ethereum/ethereum-org-website/pull/12428/files#diff-c79b5a32828cff7e979c627dcbbce26ee696bf39fb5ccaaba35772712eba08b9), file `src/intl/en/page-languages.json`

(Sorting applied)

## Related Issue
- Bug created in #12610, #12611, #12613


pettinarip (commented):
@wackerow shouldn't we also remove the `src/intl/cs/page-languages.json` file that was created?

wackerow:
Yes... I also updated this to do the same for `fr` and `zh`.

pettinarip:
@wackerow I see some conflicts with the deleted json files. Let me see if I can resolve them.

pettinarip (approved):
@wackerow fixed the conflicts. Once it finishes building I think we should be good to go.

PR #12651, opened 2024-04-03
Updated security_at_ethereum.org.asc file

fredriksvantes:
## Description
Updated security_at_ethereum.org.asc file


wackerow (approved):
Thanks @fredriksvantes!

PR #12658, opened 2024-04-04
Fix the unintended behaviour of clicking visit website on find wallet #12644

### LINKED ISSUE #12644: Bug report: find wallet - clicking "Visit website" unintended behavior
### Describe the bug

clicking on main "visit website" button in wallet listings unintentionally also opens wallet detail which triggers additional event and distorts our usage reports.

### To reproduce

2. Click on "visit website" of any wallet
3. wallet detail opens (should not)

### Expected behavior

wallet detail does not open, this is unintended behavior and is namely problematic because it also triggers "More info wallet" event which distorts the data.

In event reports it now looks that clicking on wallet detail is more popular than in reality it is because of this dynamic

### Screenshots

_No response_

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes
- [X] No

Dharmik79:
## Description

When the Visit website is clicked on the wallet, the event is compared with the associated class and allows to show the detailed view accordingly. [Fixes #12644 ]

### DIFF
--- a/src/components/FindWallet/WalletTable/index.tsx
+++ b/src/components/FindWallet/WalletTable/index.tsx
@@ -310,8 +310,6 @@ const WalletTable = ({
           )
 
           const showMoreInfo = (wallet) => {
-            // Prevent expanding the wallet more info section when clicking on the "Visit website" button
-
             updateMoreInfo(wallet.key)
             // Log "more info" event only on expanding
             wallet.moreInfo &&
@@ -333,6 +331,7 @@ const WalletTable = ({
                   if (e.key === "Enter") showMoreInfo(wallet)
                 }}
                 onClick={(e) => {
+                  // Prevent expanding the wallet more info section when clicking on the "Visit website" button
                   if (
                     (e.target as HTMLElement).matches("a, a svg")
                   )
--- a/src/components/FindWallet/WalletTable/index.tsx
+++ b/src/components/FindWallet/WalletTable/index.tsx
@@ -309,7 +309,15 @@ const WalletTable = ({
             sliceSize
           )
 
-          const showMoreInfo = (wallet) => {
+          const showMoreInfo = (wallet, event) => {
+            // Prevent expanding the wallet more info section when clicking on the "Visit website" button
+            let classList = event?.target?.classList
+            if (
+              Object.values(classList)?.includes("chakra-button") ||
+              Object.values(classList)?.includes("css-1bbj57d")
+            ) {
+              return
+            }
             updateMoreInfo(wallet.key)
             // Log "more info" event only on expanding
             wallet.moreInfo &&
@@ -328,9 +336,11 @@ const WalletTable = ({
               <Wallet
                 // Make wallets more info section open on 'Enter'
                 onKeyUp={(e) => {
-                  if (e.key === "Enter") showMoreInfo(wallet)
+                  if (e.key === "Enter") showMoreInfo(wallet, e)
                 }}
-                onClick={() => showMoreInfo(wallet)}
+                onClick={(e) => {
+                  showMoreInfo(wallet, e)
+                }}
               >
                 <Td lineHeight="revert">
                   <Flex
--- a/src/components/FindWallet/WalletTable/index.tsx
+++ b/src/components/FindWallet/WalletTable/index.tsx
@@ -334,8 +334,7 @@ const WalletTable = ({
                 }}
                 onClick={(e) => {
                   if (
-                    (e.target as HTMLTableRowElement).tagName === "A" ||
-                    (e.target as HTMLTableRowElement).tagName === "svg"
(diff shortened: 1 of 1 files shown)


src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -328,9 +336,11 @@ const WalletTable = ({
               <Wallet
                 // Make wallets more info section open on 'Enter'
                 onKeyUp={(e) => {
-                  if (e.key === "Enter") showMoreInfo(wallet)
+                  if (e.key === "Enter") showMoreInfo(wallet, e)
+                }}
+                onClick={(e) => {
+                  showMoreInfo(wallet, e)
```
wackerow, line 342: Hey @Dharmik79, thanks for taking a stab at this! I don't think we need to get into looking at class names here... I think this could be solved with a one-liner to just check if the event target is an "A" tag, and if so, return early to avoid toggling "showMoreInfo"

```suggestion
                onClick={(e) => {
                  if ((e.target as HTMLTableRowElement).tagName === "A") return
                  showMoreInfo(wallet)
```

wackerow (changes requested):
Thanks @Dharmik79, left my thoughts!

src/components/FindWallet/WalletTable/index.tsx
Dharmik79: Hey @wackerow, The solution will work only when the button is pressed, but by clicking the SVG icon within the button unintentionally triggers both the 'showMoreInfo' function and opens a new page simultaneously. To resolve this, I have used 'classList' to specifically address and prevent the issue from occurring when the button or the SVG inside the button is clicked.

Dharmik79: I have added the below code along with your suggestion to avoid the showMoreInfo function call.

 `   (e.target as HTMLTableRowElement).tagName === "A" ||
                    (e.target as HTMLTableRowElement).tagName === "svg"`

Dharmik79:
@wackerow I have added the code according to the suggestion provided.

src/components/FindWallet/WalletTable/index.tsx
Dharmik79: @wackerow  I have added the below code to resolve the issue, but if click on the arrow icon for the right side, then it will not open. So, I have used classList to resolve the issue.

`  if (
              Object.values(classList)?.includes("chakra-button") ||
              Object.values(classList)?.includes("css-1bbj57d")
            ) {
              return
            }`
Should I revert the code, with the classList or not?

src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -330,7 +332,14 @@ const WalletTable = ({
                 onKeyUp={(e) => {
                   if (e.key === "Enter") showMoreInfo(wallet)
                 }}
-                onClick={() => showMoreInfo(wallet)}
+                onClick={(e) => {
+                  if (
+                    (e.target as HTMLTableRowElement).tagName === "A" ||
+                    (e.target as HTMLTableRowElement).tagName === "svg"
```
wackerow, line 338: @Dharmik79 Ahh, good catch with the SVG, but careful because this line will now prevent it from opening for any SVG in this entire box, which there are a few and we don't want that.

I'd suggest we switch to a selector approach, since we only want to get the svg's that are inside an anchor link:

```suggestion
                    (e.target as HTMLElement).matches("a, a svg")
```

konopkja:
Hi @Dharmik79 and @wackerow - what do we need to do here to get this fix over finish line? :)

Dharmik79:
@konopkja I have fixed the changes requested by @wackerow, and waiting for the review and merge.

src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -310,6 +310,8 @@ const WalletTable = ({
           )
 
           const showMoreInfo = (wallet) => {
+            // Prevent expanding the wallet more info section when clicking on the "Visit website" button
+
```
wackerow, line 314: Last thing, just going to move this comment next to the onClick handler where it's referencing

```suggestion
```

src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -330,7 +332,13 @@ const WalletTable = ({
                 onKeyUp={(e) => {
                   if (e.key === "Enter") showMoreInfo(wallet)
                 }}
-                onClick={() => showMoreInfo(wallet)}
+                onClick={(e) => {
+                  if (
```
wackerow, line 336:
```suggestion
                  // Prevent expanding the wallet more info section when clicking on the "Visit website" button
                  if (
```

wackerow (approved):
Looks good! Thanks @Dharmik79

wackerow:
@all-contributors please add @Dharmik79 for bug fix

PR #12660, opened 2024-04-04
Fix ToC items space

pettinarip:
## Description

Fixes an incorrect value used for the `mb` styleprop.


wackerow (approved):
image

Thank you @pettinarip! This has been bugging me =)

PR #12662, opened 2024-04-04
Add Remix-IDE link to smart contract languages page

Hitisha-G:
Include Remix link in the Important Links section for easy access.

## Description

Updating documentation by incorporating a direct link to the Remix IDE, facilitating easy access for reviewing and experimenting with code.


wackerow (commented):
@Hitisha-G Worth noting that Remix-IDE also supports Vyper... Few options if we want to list this:

A) Leave it where it is, for Solidity only
B) We could consider a place on the page that could highlight it for both Solidity and Vyper, but only mention it once
C) We could add this to both the links list for Solidity and Vyper

Thoughts?

minimalsm:
> B) We could consider a place on the page that could highlight it for both Solidity and Vyper, but only mention it once

B) seems preferable to me.

wackerow:
@Hitisha-G I think we want to keep this page focused on the smart contract languages and not get too much into IDE's... I support you're suggestion to add a mention in the opening section:

Would also recommended adding a link to Remix, but question is which one... we could 
1. direct users to their homepage at https://remix-project.org where they can learn about the product first, or
2. direct straight to the browser IDE at https://remix.ethereum.org to.

I'd lean towards just directing to the IDE, but open to either. Something like this:

```
The two most active and maintained languages are:
- Solidity
- Vyper

Remix IDE provides a comprehensive development environment for creating and testing contracts in both Solidity and Vyper. [Try the in-browser Remix IDE](https://remix.ethereum.org) to start coding.
```

wackerow:
Hey @Hitisha-G, circling back.. did you want to make these changes, or would you like a hand finishing this up?

wackerow (approved):
Pushed the sudiscussed changes... thanks @Hitisha-G!

Issue #12648, opened 2024-04-03
Bug report: update the table styles based on design system /stablecoins page


konopkja:
### Describe the bug

There is a text contrast issue on https://ethereum.org/en/stablecoins/ when using dark mode.

In a table on this page when user mouse overs a stablecoin in the table, the text becomes greyed out and hard to read.

### To reproduce

1. Go to https://ethereum.org/en/stablecoins/
2. Scroll down to "Top stablecoins by market capitalisation" section and see table
3. Use dark mode
4. Mouse over item in the table
5. See error

### Expected behavior

- on mouse over keep the text white instead of greyed out (only background changes)
- The problem is it is not following the DS because it inherited properties from the page that overrides the DS ones

Design system: https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?type=design&node-id=1961%3A33521&mode=design&t=cm7YrGNCoOVawgRV-1

### Screenshots

<img width="1474" alt="Screenshot 2024-04-03 at 12 43 13" src="https://github.com/ethereum/ethereum-org-website/assets/100724231/c9493a0e-9e90-487b-879f-d477d6e97167">

### Would you like to work on this issue?

- [ ] Yes
- [ ] No

iepn:
@konopkja  I would like to work on this issue.

konopkja:
> @konopkja I would like to work on this issue.

awesome! i am assigning you the issue. Pinging @nloureiro  for visibility as DS owner

iepn:
> > @konopkja I would like to work on this issue.
> 
> awesome, thank you! i am assigning you the issue. Pinging @nloureiro for visibility as DS owner

Thank you, and wish you a wonderful day!

#12666

iepn:
@konopkja Hello, I think this PR can be closed now since it's already merged. https://github.com/ethereum/ethereum-org-website/pull/12666

PR #12666, opened 2024-04-04
fix: StablecoinsTable components dark mode text color not see#12648

iepn:
## Describe the bug
There is a text contrast issue on https://ethereum.org/en/stablecoins/ when using dark mode.

In a table on this page when user mouse overs a stablecoin in the table, the text becomes greyed out and hard to read.

CleanShot 2024-04-05 at 02 05 50

## Description

https://github.com/ethereum/ethereum-org-website/assets/57232813/32e48329-2b77-4f21-a68f-3d7f4198c6d2

https://deploy-preview-12666--ethereumorg.netlify.app/en/stablecoins/

## Related Issue

#12648


nhsz (changes requested):
Thanks! Background color needs to be updated too as it's not using the new DS defined colors cc @nloureiro

iepn:
> Thanks! Background color needs to be updated too as it's not using the new DS defined colors cc @nloureiro

okay

iepn:
I forgot about the new design, sorry! I'll update it shortly.






image

wackerow (commented):
Hey @iepn, are all of the changes here necessary for the scope of this issue?

iepn:
> Hey @iepn, are all of the changes here necessary for the scope of this issue?


Hello, if the issue is solely with the background color not displaying correctly in dark mode, then the original submission would suffice. However, upon the reminder from @nhsz, I reviewed the Figma design system of ethereum.org and found differences between the design system and the current implementation. Therefore, I applied the styles of the new table design system.

Please let me know if this is not appropriate, and I will revert it.

wackerow:
Ah okay, sounds good... apologies, had missed that message. Will review again from that scope

src/components/StablecoinsTable.tsx
```diff
@@ -50,20 +50,18 @@ const StablecoinsTable = ({
...
       my={8}
-      borderRadius="sm"
-      border={`1px solid ${textColor}`}
       bg="background.base"
       mb={8}
       minW="720px"
     >
-      <Thead bg="ednBackground" color="text200">
-        <Tr borderBottom={`1px solid ${textColor}`} mb="1px">
+      <Thead bg="background.highlight" color="text200">
```
wackerow, line 57: cc: @nloureiro okay switching `text200` (light: #666666, dark: #b2b2b2) to `body.medium` (light: gray.500 or #616161, dark: gray.400 or #8C8C8C)?

```suggestion
      <Thead bg="background.highlight" color="body.medium">
```

src/components/StablecoinsTable.tsx
```diff
@@ -50,20 +50,18 @@ const StablecoinsTable = ({
...
+        <Tr mb="1px">
```
wackerow, line 58: Could someone drop the link to the Figma for this? Feel like we shouldn't have 1px adjustments

src/components/StablecoinsTable.tsx
```diff
@@ -90,30 +88,26 @@ const StablecoinsTable = ({
...
                 <>{name}</>
               </Flex>
             </Td>
             <Td>
-              <Flex align="center">{marketCap}</Flex>
+              <Flex align="center" pt={2}>{marketCap}</Flex>
             </Td>
             <Td>
-              <Flex align="center">{stablecoinsType[type]}</Flex>
+              <Flex align="center" pt={2}>{stablecoinsType[type]}</Flex>
```
wackerow, line 110: Would need to double check with the actual Figma, but would advocate that we not rely on padding-top for positioning these. @nloureiro is this small adjustment desired?

If we're adding `pt="2"` to these, we should likely ditch that and add 1 to the y-padding of the parent `Td` which currently has `p="4"`... adding `py="5"` would do this, but again, using 5 is a bit odd, and we may want to consider simplifying.

wackerow (commented):
cc: @nloureiro

src/components/StablecoinsTable.tsx
iepn: > Could someone drop the link to the Figma for this? Feel like we shouldn't have 1px adjustments

https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?type=design&node-id=1961%3A33521&mode=design&t=cm7YrGNCoOVawgRV-1

src/components/StablecoinsTable.tsx
iepn: I used pt 2 here to try to horizontally center the content as much as possible, but please give me some time, I will come up with a better solution in the next submission.

iepn: > Would need to double check with the actual Figma, but would advocate that we not rely on padding-top for positioning these. @nloureiro is this small adjustment desired?
> 
> If we're adding `pt="2"` to these, we should likely ditch that and add 1 to the y-padding of the parent `Td` which currently has `p="4"`... adding `py="5"` would do this, but again, using 5 is a bit odd, and we may want to consider simplifying.

I used pt 2 here to try to horizontally center the content as much as possible, but please give me some time, I will come up with a better solution in the next submission.

src/components/StablecoinsTable.tsx
iepn: > cc: @nloureiro okay switching `text200` (light: #666666, dark: #b2b2b2) to `body.medium` (light: gray.500 or #616161, dark: gray.400 or #8C8C8C)?

I'm very sorry, I overlooked `text200`. Thank you for pointing it out, I appreciate your help.

src/components/StablecoinsTable.tsx
iepn: use `verticalAlign="inherit"` ！

iepn:
You will absolutely love the code in this commit! https://github.com/ethereum/ethereum-org-website/pull/12666/commits/85d573cfe8cf2cb511f17f4d3baeebb70fd98795

https://deploy-preview-12666--ethereumorg.netlify.app/en/stablecoins/

src/components/StablecoinsTable.tsx
```diff
@@ -47,23 +47,15 @@ const StablecoinsTable = ({
...
+      <Thead bg="background.highlight" color="body.medium">
+        <Tr>
           {columns.map((column, idx) => (
             <Th
               key={idx}
-              py={5}
+              fontWeight="bold"
               fontSize="md"
-              fontWeight="normal"
+              verticalAlign="inherit"
```
wackerow, line 58: Do we need this?
```suggestion
```

wackerow (changes requested):
Nice! Yeah, this looks more like our DS table... I see the middle vertical positioning which makes sense here give the icons, but left a note about the inherited property... I think we can cut that.

Otherwise lgtm!

src/components/StablecoinsTable.tsx
iepn: Trust me, it's absolutely necessary. Without it, the text won't be centered anymore.

iepn: image

I've made a diagram where you can clearly see the difference between having "verticalAlign="inherit"" and not having it.

cc @wackerow

nloureiro (approved):

wackerow:
@all-contributors please add @iepn for code

PR #12668, opened 2024-04-04
feat(storybook): build story modes

### LINKED ISSUE #12656: [RFC] Improve Variability of Storybook Snapshots with Modes
As part of an ongoing effort to improve the usage of Storybook with the new DS, apply the feature of [Story Modes](https://www.chromatic.com/docs/modes/) to Chromatic snapshots.

In short, with allow the generation of various snapshots to each story to show components in various viewports and color modes, among other possible combinations.

This can help cut down review time significantly by not forcing a reviewer to go into the Storybook canvas to check color mode and viewports widths for any design regressions.

An example of it's usefulness would for snapshots of the hero components, as the design for these components have defined alterations in the DS between various viewport widths.

TylerAPfledderer:
## Description

This PR adds Story mode configurations to Storybook, which generate in Chromatic various snapshots for different view settings. These can include a snapshot per viewport width, dark mode, language, etc.

- Adjusts the generated viewports used where they should supply width values in `px` and `em` (coming from the Chakra default theme)
- Adds modes combining viewports and chosen languages (`en`, `ru`, `uk`, `zh`)
- Applies the modes at the story to the following components:
  - `HubHero`
  - `HomeHero`
  - `ContentHero`
  - `MdxHero`
 

## Related Issue
Closes #12656


TylerAPfledderer:
@nloureiro would like your review on this. Visit the below link to the Chromatic changeset for this build
https://www.chromatic.com/review?appId=63b7ea99632763723c7f4d6b&number=12668&type=linked

A couple of things to note:
- The intention with your review of this PR is for thoughts and feedback on what snapshots we can do at this time. 
  - I think this PR should only include snapshots for existing stories that should show different modes as described above
  - If you notice and design inconsistencies between this PR and in the DS, they should also be handled in a separate PR (I've noticed a couple of things already)
  - I still need to triage being able to render dark mode snapshot based on how the Chakra addon currently behaves in accessing the toggled modes.
- You will notice that the heros rendering two button show both with the outline variant. There is still an open PR that addresses this: https://github.com/ethereum/ethereum-org-website/pull/12015
- You will notice that with the heroes rendering two buttons, the second one just says "button". This is a non-translated "dumby" button to mock what two buttons looks like on the hero. We can certainly consider using specific content already present in the translation files instead.

cc @pettinarip

TylerAPfledderer:
- [ ] @TylerAPfledderer add Farsi language to the Storybook preview option, and check the new snapshots 
  - [ ] Add comment somewhere that there is an expected bug on initial SB load, where the RTL isn't triggering based on the language set.

wackerow:
Thanks for the update @TylerAPfledderer! Should we mark as draft until these are checked off?

TylerAPfledderer:
> Thanks for the update @TylerAPfledderer! Should we mark as draft until these are checked off?

Ah yes, probably should. Will do that now!

TylerAPfledderer:
@pettinarip I have removed the Farsi language option and the directional config. Should be good for a review from you!

.storybook/modes.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  return {
+    ...arr,
+    ...currLangViewObj,
+  }
+}, {})
+
+export const allModes = {
+  ...langViewportModes,
+}
```
wackerow, line 52: @TylerAPfledderer Do we need this? Looks like we only import `langViewportModes` elsewhere... Otherwise this lgtm!

TylerAPfledderer: Good point. Examples in the docs show compiling different mode variations into one object like this to extract. But definitely not necessary here. I can remove it.

TylerAPfledderer: @wackerow this is now removed.

pettinarip (approved):
LGTM @TylerAPfledderer, thanks!

PR #12673, opened 2024-04-08
fix: update Scaffold-ETH link

futantan:


wackerow (approved):
Thanks @futantan! Thought we got all these already but I guess not... added a commit that applies this same change to the rest of the repo.

Only exception, there are a couple external tutorials that use Austin's original repo, which are not available in the new one so I left those links. 

Thanks again!

PR #12684, opened 2024-04-09
Update translator acknowledgements

lukassim:
## Description

Update the Translation Program and translation acknowledgement pages to replace the mentions of POAPs with OATs and update the process to claim them

## Related Issue

No related issue

### DIFF
--- a/src/pages/contributing/translation-program/acknowledgements.tsx
+++ b/src/pages/contributing/translation-program/acknowledgements.tsx
@@ -249,14 +249,14 @@ const TranslatorAcknowledgements = () => {
         </Flex>
       </Content>
 
-      <Content id="poap">
+      <Content id="oats">
         <ContentHeading
           as="h2"
           fontSize={{ base: "2xl", md: "2rem" }}
           fontWeight={600}
         >
           {t(
-            "page-contributing-translation-program-acknowledgements-poaps-title"
+            "page-contributing-translation-program-acknowledgements-oats-title"
           )}
         </ContentHeading>
         <Text>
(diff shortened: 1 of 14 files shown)


minimalsm (approved):

src/pages/contributing/translation-program/acknowledgements.tsx
```diff
@@ -249,14 +249,14 @@ const TranslatorAcknowledgements = () => {
...
-      <Content id="poap">
+      <Content id="oats">
         <ContentHeading
           as="h2"
           fontSize={{ base: "2xl", md: "2rem" }}
           fontWeight={600}
         >
           {t(
-            "page-contributing-translation-program-acknowledgements-poaps-title"
+            "page-contributing-translation-program-acknowledgements-oats-title"
```
wackerow, line 259: @lukassim 

image

We end up with a weird case here... The above .json string changes are in the English files only... we changed this string name, but this doesn't exist in other languages yet, so it falls back to the English "OATs"... but the other strings do exists in those langs and have the old copy.

I'd suggest we update this string key in all languages as well, which will display POAP along with everything else in there until these get translated.

lukassim: Good point, though I'd argue OAT shouldn't be translated anyway, same as with POAP
But yeah, there's still mention of POAP in all the subsequent text...

lukassim: One thing to note is that this page doesn't get that many views even in English, and even less in other languages, so I'd also be fine with keeping it this way until the new copy gets translated

wackerow: Hah okay... I just went ahead and fixed that key so the non-English pages will just be consistently not yet updated... the updates will then trickle down in cohesive chunks at least... and like you said, not a lot of eyes.

wackerow (approved):

PR #12696, opened 2024-04-09
feat: Crowdin build automation

wackerow:
## Description
- `build-crowdin.yml` - New GitHub action workflow, which runs a new `triggerBuild.ts` script on the 1st of every month.
- `triggerBuild.ts` - Simple script to call `crowdin.translationsApi.buildProject()`, triggering Crowdin to build our project

## Related Issue
Crowdin CI/CD efforts

cc: @lukassim

### DIFF
--- a/.github/workflows/build-crowdin.yml
+++ b/.github/workflows/build-crowdin.yml
@@ -2,7 +2,7 @@ name: Build Crowdin project
 
 on:
   schedule:
-    - cron: "20 04 1 * *"
+    - cron: "20 04 1 * *" # Runs at 4:20 AM on the first day of every month
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/build-crowdin.yml
+++ b/.github/workflows/build-crowdin.yml
@@ -2,7 +2,7 @@ name: Build Crowdin project
 
 on:
   schedule:
-    - cron: "20 16 * * Mon"
+    - cron: "20 04 1 * *"
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/build-crowdin.yml
+++ b/.github/workflows/build-crowdin.yml
@@ -1,0 +1,30 @@
+name: Build Crowdin project
+
+on:
+  schedule:
+    - cron: "20 16 * * Mon"
+  workflow_dispatch:
+
+jobs:
+  create_pr:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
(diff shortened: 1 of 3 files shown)


minimalsm:
@wackerow converted this to draft for now after our discussion on a more granular approach

in short: we agreed we could use [build project directory translation](https://developer.crowdin.com/api/v2/#operation/api.projects.translations.builds.directories.post) to build a directory (content bucket) as needed in a specific language, instead of relying on a cron job that could allow build content to get out of sync

wackerow:
With the latest updates, I think just building the entire project with approved strings only once a month is enough. This is set to trigger 12 hours before the import workflow is triggered. I think it's worth keeping these separate to avoid having potentially large amounts of waiting time during one workflow which could lead to timeouts while waiting for the project build to be completed.

wackerow:
With the logic we have, all we really need is this to build the entire project; we can pull the approved easily from there.

### Current workflow

At least two workflows run in succession at a regular interval (currently on the 1st of the month)

1. The workflow in this PR will build the entire project.
1. Shortly after, a second workflow (#12698) will download this build, check which buckets are fully approved, import those buckets and place PRs for each language
1. A third workflow (not yet built) is planned to follow these actions by bundling all the new English content changes and upload to Crowdin for translation. At this point, the "progress" for any updated pages will be set back, awaiting further translation and review by Acolad.
1. The report generation enabled by PR #12738 can be used as-needed by @lukassim to generate a report for Acolad of any fully translated buckets that have not been fully approved yet. 

The frequency can be changed in the future; for example to every two weeks, in cadence with our deploy schedule.

lukassim (approved):
Amazing, LGTM

In the future, I would advocate for us to try and trigger this more frequently, e.g. every two weeks, but let's run it with actual reviewed content first and see how that goes
We'll have completed reviews ready to be pulled in by the end of April

.github/workflows/build-crowdin.yml
```diff
@@ -0,0 +1,30 @@
+name: Build Crowdin project
+
+on:
+  schedule:
+    - cron: "20 04 1 * *"
```
pettinarip, line 5: Add comment to make this human readable
```suggestion
    - cron: "20 04 1 * *" # Runs at 4:20 AM on the first day of every month
```

.github/workflows/build-crowdin.yml
```diff
@@ -0,0 +1,30 @@
...
+  workflow_dispatch:
```
pettinarip, line 6: Was reading a bit about this, pretty cool. This would allow us to trigger the workflow manually if necessary, right?

pettinarip (approved):
@wackerow LGTM!

PR #12697, opened 2024-04-09
fix: gh auth step in workflows

wackerow:
## Description
Due to a change in the GitHub CLI gh version 2.0.0 the --with-token flag no longer accepts an argument and instead it reads from standard input

- Updates workflows to pipe standard input into `gh auth` command

## Related Issue
Related workflows have been failing for the past few weeks as a result of this.


pettinarip (approved):
👍🏼 lgtm

PR #12698, opened 2024-04-09
feat: Crowdin import automation

wackerow:
## Description
Adds a GitHub action cron job that runs a series of actions on the first of every month:

- Downloads latest build for our project from Crowdin
- Unzips contents temporarily into `.crowdin/`
- Check which directories from Crowdin (buckets) for each locale have an approval of 100%, and form bucket import list
- An upgraded Crowdin Import script accepts the desired list of bucket imports, and imports them similar to how `crowdin-import.ts` worked in the past.
- The markdown checker is automatically run on completion, and has been modified to not just console log the results, but to save them temporarily to be posted in a PR body
- Git changes are staged to unique branches by language, then committed, and a PR is opened
- Language-specific markdown checker results are added to the body of the PR for review.

### Tasks
- [x] Confirm logic for which buckets are in-fact approved and ready to be imported
- [x] Enable pushing these new branches to GitHub with `gh auth` and posting one PR per language

## Related Issue
Crowdin CI/CD efforts


wackerow:
Latest commits add in a script that fetches the approval status of each directory on Crowdin and generates the "buckets list". We need to test this, and likely debug some details, but otherwise the main thing left would be enabling the actually pushing of these branches to GitHub and opening PRs for each.

The full "action" will not run until merged into dev. May be worth orchestrating this to only work for a couple languages, then bring it in and see how it does with those before removing that restriction.

src/scripts/markdownChecker.ts
```diff
@@ -325,17 +332,29 @@ function getLineNumber(file: string, index: number): string {
...
-      processFrontmatter(path, lang)
-      processMarkdown(path)
+      processFrontmatter(path, lang, summary[lang])
+      processMarkdown(path, summary[lang])
     }
+
+    if (!summary[lang].length) delete summary[lang]
   }
+
+  console.table(summary)
```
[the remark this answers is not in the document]
wackerow: I've semi-reverted the changes here... made a `log` function that will log each event to the console with the previous levels (`log`, `warn`, or `error`), and also tracks these in a list to be used for the PR body.

This should log them the same way as before (hence removal of the final `console.table(summary)`, and thus should still be able to be run locally to get these clickable paths.

The PR summary will output a copy/paste-able code block with `yarn markdown-checker` there as a convenience/reminder.

wackerow:
Testing working great! The PRs for the test were posted appropriately and made a few adjustments based on that. Those branches had changes from _this_ branch included in it, which won't happen when this is executed from a fresh `dev` branch from the GitHub action. 

Using the "approved progress" for each language for each directory (representing a bucket) to filter for only those 100% ready to be pulled in.

lukassim (approved):
Looks great, thanks @wackerow!

I think this is basically ready to run as is, we can test it out at the start of May for all of the April reviews that are currently in progress

src/scripts/crowdin/import/utils.ts
```diff
@@ -0,0 +1,171 @@
...
+ * to their corresponding destinations in the repo. Function is called
+ * again recursively for subdirectories.
+ *
+ * @param _path An absolute path to the directory being scraped.
+ * @param contentSubpath The subpath deep to the lang-code directory,
+ * used to construct destination for markdown content files
+ * @param repoLangCode Language code used within the repo
+ * @returns void
+ */
+export const scrapeDirectory = (
```
pettinarip, line 27: This function do more than just scrape the dir. We could break out the logic with something like this and simplify the function
```ts
function scrapeDirectory(dir, cb) {
  // ...
  ls.forEach(cb)
}

scrapeDirectory(dir, (path) => {
  // do our logic
})
```

Having done this, we could move `scrapeDirectory` to somewhere in `@/libs/utils` since it is a reusable function.

src/scripts/crowdin/import/utils.ts
```diff
@@ -0,0 +1,171 @@
...
+      item.endsWith(".md") ||
+      item.endsWith(".svg") ||
+      item.endsWith(".xlsx")
+    ) {
+      const mdDestDirPath: string = join(
+        "public",
+        "content",
+        "translations",
+        repoLangCode,
+        contentSubpath
```
pettinarip, line 57: Use shared constants for these paths

src/scripts/crowdin/import/utils.ts
```diff
@@ -0,0 +1,171 @@
...
+import i18Config from "../../../../i18n.config.json"
+import { DOT_CROWDIN } from "../translations/constants"
+
+import { BucketsList, SelectionItem, SummaryItem, TrackerObject } from "./types"
+
+/**
+ * Some language codes used in the repo differ from those used by Crowdin.
+ * This is used to convert any codes that may differ when performing folder lookup.
+ */
+export const getCrowdinCode = (code: string): string =>
```
pettinarip, line 13: Shouldn't we move some of these (if not all) functions to the `@/lib/utils/crowdin` file?

I'd probably only keep the summary logic here since it is tightly related to this script.

src/scripts/crowdin/translations/getApprovedBuckets.ts
```diff
@@ -0,0 +1,45 @@
...
+
+      const n = parseInt(bucketDir.name.substring(0, 2))
+      bucketsList[code].push(n)
+    }
+  }
+
+  return bucketsList
+}
+
+getApprovedBuckets()
```
pettinarip, line 43: Why is this function exec here?

src/scripts/crowdin/translations/getTranslations.ts
pettinarip, line 1: Following the pattern, I'd call this `main.ts` since it is the main script file that is going to be executed.

src/scripts/crowdin/translations/postLangPRs.ts
```diff
@@ -0,0 +1,25 @@
+import fs from "fs"
+
+import i18n from "../../../../i18n.config.json"
+import { BucketsList } from "../import/types"
+
+import { BUCKETS_PATH } from "./constants"
+import { processLocale } from "./utils"
+
+const locales = i18n.map((lang) => lang.code)
```
pettinarip, line 9: We could use `LOCALES_CODES` from our constants file.

src/scripts/crowdin/translations/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+\`\`\`
+
+<details><summary>Unfold for ${summaryJson[locale].length} result(s)</summary>
+
+${qaResults}
+</details>
+`
+}
+
+export const processLocale = (locale: string, buckets: number[]) => {
```
pettinarip, line 62:
```suggestion
export const createLocaleTranslationPR = (locale: string, buckets: number[]) => {
```

pettinarip (changes requested):
@wackerow nice work, overall it looks good to me! I've left, mostly, refactor ideas that I'd like to see your opinions on.

=====
Been looking at the official [GH/Crowdin action integration](https://github.com/crowdin/github-action), it looks like it does most of functionality we need. Do you know what's stopping us from using it? couldn't we build our custom logic on top of it?

src/scripts/crowdin/translations/getApprovedBuckets.ts
wackerow: Good catch; a remnant from when I was calling this from a yarn script. Removing

src/scripts/crowdin/import/utils.ts
wackerow: @pettinarip If we do this we just need to make sure we don't import dependencies using `@` syntax anywhere in the chain... In the `src/scripts` folder we are typically running these with `ts-node` which fails when it encounters that shorthand with `code: 'MODULE_NOT_FOUND'` (at least with our current setup). 

In `src/lib` we use the `@` import syntax, so if we want to mix them we may need to reverse that syntax in some cases.

src/scripts/crowdin/import/utils.ts
```diff
@@ -0,0 +1,171 @@
...
+): void => {
+  // Initialize tracker for language
+  trackers.langs[repoLangCode] = {
+    buckets: [],
+    jsonCopyCount: 0,
+    mdCopyCount: 0,
+    error: "",
+  }
+  // Initialize working directory and check for existence
+  const _path: string = join(DOT_CROWDIN, crowdinLangCode)
```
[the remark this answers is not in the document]
wackerow: I believe I was previously doing `import path from 'path'` so I used this as an alternate, but I'm destructuring `{ join }` now so I can just remove it.

src/scripts/crowdin/import/utils.ts
wackerow: @pettinarip Curious if you see this as blocking? This logic was ported forward from the previous `crowdin-import.ts` file. Any chance we could handle this refactor separately? Similar with above comment about migrating things to `@/lib/utils/crowdin`

wackerow:
> Been looking at the official [GH/Crowdin action integration](https://github.com/crowdin/github-action), it looks like it does most of functionality we need. Do you know what's stopping us from using it? couldn't we build our custom logic on top of it?

Interesting! I hadn't looked into that, but I've already got a handful of ideas for how we could iterate a v2 of this new approach. If you're okay with it I'd like to note this one for later.

src/scripts/crowdin/import/utils.ts
```diff
@@ -0,0 +1,170 @@
...
+      if (!existsSync(jsonDestDirPath))
+        mkdirSync(jsonDestDirPath, { recursive: true })
+      const jsonDestinationPath: string = join(jsonDestDirPath, item)
+      copyFileSync(source, jsonDestinationPath)
+      // Update .json tracker
+      trackers.langs[repoLangCode].jsonCopyCount++
+    } else if (
+      item.endsWith(".md") ||
+      item.endsWith(".svg") ||
+      item.endsWith(".xlsx")
```
corwintines, line 51:
```suggestion
      item.endsWith(".md") 
```

I don't think we want to commit .svg and .xlsx file formats. There is a different process for  importing images that @nloureiro and @lukassim have sorted out I believe.

corwintines (changes requested):
Just have one comment above on the files being committed. Other than that, this seems good to go for me. I agree with @pettinarip about maybe checking out the integration, but I do think this can wait for later. This is working now and we should test this out before testing out that integration IMO.

src/scripts/crowdin/import/utils.ts
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Some language codes used in the repo differ from those used by Crowdin.
+ * This is used to convert any codes that may differ when performing folder lookup.
+ */
+export const getCrowdinCode = (code: string): string =>
+  i18Config.filter((lang) => lang.code === code)?.[0]?.crowdinCode || code
+
+/**
+ * Reads `ls` file contents of `_path`, moving .md and .json files
```
corwintines, line 18:
```suggestion
 * Reads `ls` file contents of `path`, moving .md and .json files
```

src/scripts/crowdin/import/utils.ts
pettinarip: No, not a blocker, was more an refactor idea to make it clearer. Got it, makes sense what you are saying! we need to investigate a bit about those aliases and ways to use them in scripts.

src/scripts/crowdin/import/utils.ts
pettinarip: Got it! didn't know about that problem.

pettinarip:
> > Been looking at the official [GH/Crowdin action integration](https://github.com/crowdin/github-action), it looks like it does most of functionality we need. Do you know what's stopping us from using it? couldn't we build our custom logic on top of it?
> 
> Interesting! I hadn't looked into that, but I've already got a handful of ideas for how we could iterate a v2 of this new approach. If you're okay with it I'd like to note this one for later.

Of course! just wanted to share that with you because it looked relevant.

pettinarip (approved):
LGTM!

Issue #7743, opened 2022-09-08
Help needed badge renders poorly, giving broken and low quality impression


miohtama:
**Describe the bug**

"Help needed" badge on certain pages renders poorly, giving an impression of low quality on Ethereum.org website.

[This badge is found on Python page](https://ethereum.org/en/developers/docs/programming-languages/python/) but might be on other pages as well.

**Expected behavior**

The badge looks good regardless of a web browser.

**Screenshots**

<img width="656" alt="image" src="https://user-images.githubusercontent.com/49922/189080026-1e95ab35-d11a-49ac-bf4d-eda5a018ac17.png">

**Desktop (please complete the following information):**

- OS: macOS
- Browser: Firefox

**Additional context**

- I suggest the badge is hidden until a replacement image is found
- The replacement image should be SVG or high-quality PNG

corwintines:
Agree @miohtama, this is pretty ugly.

@nloureiro I think we should still have a callout component, but this should be redesigned to something a little more appealing 😅 Do you mind helping with a design here so that we could change this up.

PR #12715, opened 2024-04-11
Remove doge from CallToContribute component

corwintines:
## Description

- Remove doge from CallToContribute component
- Update styling on H2 component in CallToContribute

Fixes: https://github.com/ethereum/ethereum-org-website/issues/7743


wackerow (approved):
@nloureiro Perhaps separately we replace this with some new image, just a thought

konopkja:
we should unify these callouts and use DS, but for the purpose of this task its ok

konopkja (approved):

PR #12717, opened 2024-04-11
Fix hydration issue with tooltip and glossary tooltip

pettinarip:
## Description

Fixes invalid html around the GlossaryTooltip + use a Portal to display the tooltip content to avoid having nested divs inside inline text.


wackerow (approved):
Nice! Looks good @pettinarip, thanks!

PR #12639, opened 2024-04-03
refactor(ContentHero): update and apply component usage

TylerAPfledderer:
## Description

This PR add modifications to the `ContentHero` component and applies it to the `Upgrade` layout.

- The `description` section of the hero can accept either a string or components
  - In prod for the `Upgrade` layout, a list of descriptions render with a "last updated" timestamp
- The `Image` component has new styling with predefined dimensions
- Update types to allow for the `heroImg` prop to accept either a string or the static image data object, depending on the hero component to be used.
- Affects the [/merge](https://ethereum.org/en/roadmap/merge/) and [beacon-chain](https://ethereum.org/en/roadmap/beacon-chain/) pages

## Additional Information

This relieves a bug in the [roadmap merge page](https://ethereum.org/en/roadmap/merge/) where the alignment of the image and content in not centered and the image is clipped.

PR #12718, opened 2024-04-11
Fix list item color in layouts

pettinarip:
Currently we have this
image
a different color than `body.base` in our list items.

## Description

This PR fixes the color by removing the custom `ListItem` components used in the layouts to default to the used in the `mdComponents` list.

## Related Issue

#9548

### DIFF
--- a/src/layouts/Upgrade.tsx
+++ b/src/layouts/Upgrade.tsx
@@ -9,7 +9,6 @@ import {
   Icon,
   List,
   ListItem,
-  Skeleton,
   Text,
   useToken,
 } from "@chakra-ui/react"
@@ -51,10 +50,6 @@ const Title = (props: ChildOnlyProp) => (
     mt={0}
     {...props}
   />
-)
-
-const SummaryPoint = (props: ChildOnlyProp) => (
-  <ListItem color="text300" mb={0} {...props} />
 )
 
 type ContainerProps = Pick<BoxProps, "children" | "dir">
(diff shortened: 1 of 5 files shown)


minimalsm (commented):
Should we also remove this from [the Tutorial layout](https://github.com/ethereum/ethereum-org-website/blob/cfaadb1a9add22aa9b1432c77eed7643258b7ae8/src/layouts/Tutorial.tsx#L128) @pettinarip?

src/layouts/Upgrade.tsx
```diff
@@ -53,10 +52,6 @@ const Title = (props: ChildOnlyProp) => (
   />
 )
 
-const SummaryPoint = (props: ChildOnlyProp) => (
-  <ListItem color="text300" mb={0} {...props} />
```
minimalsm, line 57: I think, at very least, we want to keep the margin-bottom here. I'd also advocate for not using pure white here. 

Before:

Screenshot 2024-04-12 at 16 39 00


After:

Screenshot 2024-04-12 at 16 38 46


Maybe we can allow these changes to be handled in this discussion: https://github.com/ethereum/ethereum-org-website/pull/12639#issuecomment-2045619701

cc: @nloureiro

wackerow (approved):

wackerow:
Removed the design tag as this aligns with the DS; but comment with any disagreements and we can patch

PR #12728, opened 2024-04-13
fix: cryptography tooltip

Vanilladelphia:


pettinarip (approved):
Thanks @Vanilladelphia! good catch

PR #12761, opened 2024-04-17
Added 2023 to Translatathon page

shravanandoria:
I've added the missing year after the dates under the page https://ethereum.org/en/contributing/translation-program/

### DIFF
--- a/public/content/contributing/translation-program/translatathon/index.md
+++ b/public/content/contributing/translation-program/translatathon/index.md
@@ -18,10 +18,10 @@ We invite you to join us in breaking down language barriers and making ethereum.
 
 ### When {#when}
 
-- Application period: August 1st - August 15th
-- Translation period: August 16th - August 23rd
-- Evaluation & QA period: August 23rd - August 30th
-- Results announcement: August 31st
+- Application period: August 1st - August 15th, 2023
+- Translation period: August 16th - August 23rd, 2023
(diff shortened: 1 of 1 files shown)


wackerow:
Hey @shravanandoria, appreciate this suggestion. Not sure exactly how we want to handle this since this page has been archived at this point (there are no more links to this inside the site)... We should probably place a banner at the top indicating that this is over and maintained for archival purposes only, or consider removing the page entirely.

@lukassim Curious your thoughts on how we want to handle this page

lukassim:
> Hey @shravanandoria, appreciate this suggestion. Not sure exactly how we want to handle this since this page has been archived at this point (there are no more links to this inside the site)... We should probably place a banner at the top indicating that this is over and maintained for archival purposes only, or consider removing the page entirely.
> 
> @lukassim Curious your thoughts on how we want to handle this page

Well, we are reusing this page and redoing it with relevant information for the 2024 edition in the next couple of months, so I don't really see the point in changing it now... but given that this was a good first issue opened by Isabella, we should probably just merge the changes until we get to rethinking this page

wackerow (approved):
Thanks @lukassim, makes sense

And looks good @shravanandoria! I just added one last similar update below where the application date range is mentioned again. Pulling in

wackerow:
@all-contributors please add @shravanandoria for maintenance

shravanandoria:
Thanks @wackerow for your feedback!

PR #12779, opened 2024-04-19
update the position of the tooltip and set it near the label text

### LINKED ISSUE #12643: Bug report: poor tooltip position - find wallet
### Describe the bug

in some instances the tooltip is too far from the text
Screenshot 2024-04-03 at 11 55 09

### To reproduce

2. Click on wallet detail

### Expected behavior

the tooltip should be right after the text without blank gap

### Screenshots

_No response_

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes
- [X] No

Dharmik79:
## Description

Added the tooltip inside the feature label text and made the list item aligned horizontally

## Preview URL

https://deploy-preview-12779--ethereumorg.netlify.app/en/wallets/find-wallet

## Related issue
- Fixes #12643


src/components/FindWallet/WalletTable/WalletMoreInfoCategory.tsx
```diff
@@ -86,21 +87,20 @@ export const WalletMoreInfoCategory = ({
               >
                 <ListIcon as={FeatureIcon}></ListIcon>
 
-                <Text px={1} lineHeight={1}>
+                <Text as="span" whiteSpace="nowrap" px={1} lineHeight={1}>
```
wackerow, line 90: I don't think we want to use `nowrap` here...

image

It causes overflow.

This was why I had suggested breaking out the last word and attaching the icon to just that word... so the icon will never wrap by itself, but the words still will. Unfortunately there is no CSS pseudo-selector for ":last-word", which would make this easier. 

Hm.. thinking more here though, perhaps we could use a non-breaking-space `&nbsp;` between the string and the Emoji instead? Haven't tried yet, but could be worth a shot... should prevent the emoji from wrapping alone while allowing the spaces between the words to break a line.

wackerow (changes requested):

src/components/FindWallet/WalletTable/WalletMoreInfoCategory.tsx
wackerow: I think using `&nbsp;` ("non-breaking space") would work actually, @Dharmik79 are you down to give this a try?

Dharmik79: @wackerow The solution using &nbsp worked.

src/components/FindWallet/WalletTable/WalletMoreInfoCategory.tsx
```diff
@@ -87,20 +88,19 @@ export const WalletMoreInfoCategory = ({
                 <ListIcon as={FeatureIcon}></ListIcon>
 
                 <Text px={1} lineHeight={1}>
-                  {feature.label}
+                  {feature.label}&nbsp;
```
wackerow, line 91: ☹️ This is pesky... I'm still seeing this wrap. Made some adjustments locally but only getting it to work on Safari but not Brave.
In either case, I would stick with either the non-breaking space, or the margin (`ms="2"` below) but not both... I'm still not sure of a clean way to do this for all browsers.

wackerow: @Dharmik79 Just checking in... you still willing/able to help with this, or would you prefer someone to try and take it from here?

Dharmik79: @wackerow I think removing the ms="2" will be a better option.

wackerow:
Circling through, still seeing buggy positioning... going to switch this to draft mode.

corwintines:
Just to confirm what you mean by buggy positioning @wackerow, is it because the Tooltip icon is aligning with the top of the text, and not center?

wackerow:
@corwintines I'm seeing this

image

..with the info icon on it's own line

corwintines (approved):
LGTM, thanks @Dharmik79!

src/components/FindWallet/WalletTable/WalletMoreInfoCategory.tsx
```diff
@@ -66,6 +66,7 @@ export const WalletMoreInfoCategory = ({
                 listStyleType="none"
                 display="flex"
                 mb={2}
+                alignItems="center"
```
wackerow, line 69: image

We should revert this as it's out of scope with issue #12643, which was only referring to the positioning of the info "i" icon.

```suggestion
```

If this is a change you'd like to suggest please open a new issue for it.

wackerow (changes requested):
Going to revert the alignment adjustment for the containing flex box, then pull this in

wackerow (approved):
Okay, LGTM! Thanks @Dharmik79

PR #12787, opened 2024-04-21
fix(storybook): expose `Inter` local fonts to storybook

TylerAPfledderer:
## Description

Somehow in a past build (probably during the project migration to NextJS), the `Inter` font lost exposure in Storybook; It is no longer available in the existing configuration to being available to the Chakra theme in the SB environment.

And with the Inter font now being served locally via PR #12501, there needed to be  different approach to applying it back into the environment

- Creates `preview-head.html` to redefine the CSS variable `--font-inter` to it is available to the Chakra theme in the SB environment
  - This variable in prod is currently declared in `_app.tsx` so it is unavailable to the SB environment from that location
- Exposes the `global.css` to make the inter font faces available

## Additional Information
This PR does not consider the mono font, which is also not exposed to Storybook. Because of the current approach to making the font available for Chakra in NextJS, it makes it difficult to apply in SB. Recommend also adding this font locally.

## Related Issue
N/A

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,0 +1,7 @@
+<style type="text/css">
+
+    :root {
+        --font-inter: 'Inter', sans-serif;
+    }
+
+</style>
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -5,3 +5,5 @@
     }
 
 </style>
+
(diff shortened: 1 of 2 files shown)


pettinarip (approved):
Good catch @TylerAPfledderer. Never realized that we weren't using the Inter font in sb lol!

PR #12795, opened 2024-04-22
fix(Stat): wrap `Tooltip` child in button

TylerAPfledderer:
## Description

This is a follow-up to an issue noticed in the [Stat component creation PR](https://github.com/ethereum/ethereum-org-website/pull/11574#pullrequestreview-2015599608)

Fixes placement of the tooltip popover for the `Stat` component, by wrapping the child content in a button.

Per [Chakra docs for the Popover](https://v2.chakra-ui.com/docs/components/popover/usage#basic-usage) the child element should be able to take have a forwarded ref and be focusable (the later currently not considered). This means that at the very least the element should be a `button` or an `a`.

Wrapping in a button also allows for keyboard focus which triggers the popover.

## Related Issue

N/A

### DIFF
--- a/src/components/Stat/index.tsx
+++ b/src/components/Stat/index.tsx
@@ -1,7 +1,7 @@
 import { useEffect, useState } from "react"
 import type { IconType } from "react-icons/lib"
 import { MdInfoOutline, MdWarning } from "react-icons/md"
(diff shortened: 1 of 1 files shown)


wackerow:
Okay cool, something maybe we can chat about separately. Thanks!

PR #12837, opened 2024-04-25
docs(applying-storybook): update documentation and usage

TylerAPfledderer:
As essentially task #0 in rolling out new initiatives in including more component stories, this PR updates and adds clarification to the `Applying Storybook` documentation for contributors.

This also updates existing stories to ensure they are inline with this updated documentation.

### DIFF
--- a/docs/applying-storybook.md
+++ b/docs/applying-storybook.md
@@ -22,7 +22,7 @@ It's as easy as running `yarn storybook` to boot up a dedicated localhost to see
 
 A Storybook "story" is an instance of a component in a certain state or with certain parameters applied to show an alternative version of the component.
 
-Each component will only need one file containing all the stories.
+There may be some exceptions, but generally each component should have only on story file.
 
 The stories file will reside with each component. So the base folder structure in `src` will look like this:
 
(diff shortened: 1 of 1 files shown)


docs/applying-storybook.md
```diff
@@ -157,3 +158,56 @@ Outlined below are each area going from left to right in the selections.
...
+// In the `meta` object
+parameters: {
+  chromatic: {
+    modes: {
+      base: viewportModes['base']
+      'ru-xl': langViewportModes['ru-xl']
+    }
+  }
+}
+```
```
pettinarip, line 213: Great job @TylerAPfledderer, I think you mentioned all the important parts. Maybe a nice to have would be a mention of the `disableSnapshot` option since we will probably use it in cases where we want the story but not the visual test.

TylerAPfledderer: @pettinarip I will review and apply a section covering this. Thank you!

TylerAPfledderer: @pettinarip this has now been added.

pettinarip (approved):
Great @TylerAPfledderer, lgtm!

Will wait for @corwintines review in case we are missing something.

docs/applying-storybook.md
```diff
@@ -116,7 +117,7 @@ export const Variants = {
 
 ### Story file containing a single story
 
-If only one story is provided for a component, the name of the exported object should match the name in the `title` meta option. (If the title is `Atoms / Form / Button` then the object should be named `Button`) This will hoist the display name up to the parent level in the Storybook dashboard's sidebar. This will also mean you have to rename the import of the component. Call it `ButtonComponent`, say.
+If only one story is provided for a component, the name of the exported object should match the name in the `title` meta option. (If the title is `Atoms / Form / Button` then the story should be named `Button`) This will hoist the display name up to the parent level in the Storybook dashboard's sidebar. This will also mean you have to rename the import of the component. Call it `ButtonComponent`, say.
```
corwintines, line 120:
```suggestion
If only one story is provided for a component, the name of the exported object should match the name in the `title` meta option. For example, If the title is `Atoms / Form / Button` then the story should be named `Button`. This will hoist the display name up to the parent level in the Storybook dashboard's sidebar. This will also mean you have to rename the import of the component. Call it `ButtonComponent`, say.
```

Just cleaning up one section. Sentence syntax confused me a bit.

corwintines (approved):
Left one suggestion, but don't think it needs to block merging. Nice one @TylerAPfledderer

docs/applying-storybook.md
TylerAPfledderer: Fixing the missed punctuation. 😅 

```suggestion
If only one story is provided for a component, the name of the exported object should match the name in the `title` meta option. For example, if the title is `Atoms / Form / Button` then the story should be named `Button`. This will hoist the display name up to the parent level in the Storybook dashboard's sidebar. This will also mean you have to rename the import of the component. Call it `ButtonComponent`, say.
```

PR #12842, opened 2024-04-26
fix: strip query params/hashes form matomo events

wackerow:
## Description
- Customizes the page URL logged for Matomo events using the `setCustomUrl` function
- Query params and hash links are removed from the current URL before being logged

See:
- https://developer.matomo.org/guides/tracking-javascript-guide#manually-trigger-a-page-view
- https://matomo.org/faq/how-to/how-do-i-set-a-custom-url-using-the-matomo-javascript-tracker/

## Related Issue
cc: @konopkja


corwintines (approved):

pettinarip (approved):
LGTM!

PR #12858, opened 2024-04-29
Netlify lighthouse plugin config

pettinarip:
## Description

Adds pages to be audited by the Netlify lighthouse plugin.


pettinarip:
Confirming that the contributors component is impacting the LCP score: they gave us a ~2/3 seconds LCP while in prod we have > ~9s.

minimalsm (approved):

PR #12860, opened 2024-04-29
[TEST] Force chunk generation for `framer-motion`

pettinarip:


pettinarip:
Closing, as it doesn't add any real benefit on loading times. It will create a chunk but it will be loaded at the same time as the `app.js` file.

PR #12863, opened 2024-04-29
Performance - contributors component refactor

pettinarip:
## Description

Immediately display the contributors modal button to reduce LCP.

Other changes:
- Used the MdChevronRight icon to match the initial close state and avoid the initial animation causing the LCP to be higher sometimes
- Removed the Skeletons and add one inside the modal

====
### Tests after changes
Webpagetest https://www.webpagetest.org/result/240430_BiDc8E_92E/

All the LCPs are the same now, ~2/3s.

### DIFF
--- a/src/components/SideNav.tsx
+++ b/src/components/SideNav.tsx
@@ -13,13 +13,13 @@ import docLinks from "../data/developer-docs-links.yaml"
 
 export const dropdownIconContainerVariant = {
   open: {
-    rotate: 0,
+    rotate: 90,
     y: 3,
     transition: {
       duration: 0.4,
     },
   },
-  closed: { rotate: -90, y: 0 },
+  closed: { rotate: 0, y: 0 },
 }
 
 const innerLinksVariants = {
@@ -105,8 +105,14 @@ const NavLink = ({ item, path, isTopLevel }: NavLinkProps) => {
             variants={dropdownIconContainerVariant}
             animate={isOpen ? "open" : "closed"}
             cursor="pointer"
+            display="flex"
           >
-            <Icon as={MdExpandMore} boxSize={6} color="secondary" />
+            <Icon
+              as={MdExpandMore}
+              transform="rotate(-90deg)"
+              boxSize={6}
+              color="secondary"
+            />
           </Box>
         </LinkContainer>
         <Box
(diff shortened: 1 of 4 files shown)


src/components/SideNavMobile.tsx
```diff
@@ -108,11 +108,19 @@ const NavLink = ({ item, path, toggle }: NavLinkProps) => {
...
             animate={isOpen ? "open" : "closed"}
           >
-            <Icon as={MdExpandMore} boxSize={6} color="secondary" />
+            <Icon
+              as={MdExpandMore}
+              // display by default the icon in this position `>` (close state)
+              // this will prevent the icon from rotating when the page loads
+              transform="rotate(-90deg)"
+              boxSize={6}
+              color="secondary"
```
minimalsm, line 122: Should we just use `MdChevronRight` where rightward facing arrows are default state?

src/components/SideNav.tsx
```diff
@@ -13,13 +13,13 @@ import docLinks from "../data/developer-docs-links.yaml"
 
 export const dropdownIconContainerVariant = {
   open: {
-    rotate: 0,
+    rotate: 90,
     y: 3,
```
minimalsm, line 17:
```suggestion
```

The framer positioning can be removed due to the flex container being added.

src/components/SideNav.tsx
```diff
@@ -13,13 +13,13 @@ import docLinks from "../data/developer-docs-links.yaml"
...
     transition: {
       duration: 0.4,
     },
   },
-  closed: { rotate: -90, y: 0 },
+  closed: { rotate: 0, y: 0 },
```
minimalsm, line 22:
```suggestion
  closed: { rotate: 0, },
```

src/components/SideNavMobile.tsx
pettinarip: Aha! makes total sense, yup. Didn't know we had that icon.

src/components/SideNav.tsx
pettinarip: Ah right! nice catch. Lets remove that as well.

minimalsm (approved):

nloureiro:
yes, this is the temporary change we agreed on the other day.

Can I ask just to add the same spacing above and below the title?
I think it's `var(--eth-space-8)` between the title and the button

Screen Shot 2024-04-30 03 28 54 PM

pettinarip:
@nloureiro fixed

nloureiro (approved):

nhsz (approved):

PR #12866, opened 2024-04-29
Add Blocknative Blob Archive to Dencun Historical Access

laurencedelisle:
## Description
Added the Blocknative Blob Archive to the Dencun historical access page for researchers and public good projects to find. The Blocknative Blob Archive stores all blobs that land in the mempool (confirmed on-chain or not) for anyone to access. This data is available via API.


wackerow:
Hey @laurencedelisle, appreciate the post! This page currently is not listing products. There are many sources for this data, but starting to list block explorers or indexing services by name on this page I would push back against. 

As an alternative, I'd suggest perhaps we link out to one of the other pages that lists products of this variety, such as the [block explorers page](https://ethereum.org/en/developers/docs/data-and-analytics/block-explorers/). Going to close this out for now, but would be open to a separate issue/PR for a link-out to the existing lists. Thanks!

PR #12872, opened 2024-04-30
Upgrade NextJs to v13.5 & netlify runtime plugin

pettinarip:
## Description

Bump
- nextjs to latest v13
- netlify/next runtime plugin to latest


wackerow:
House-keeping question, with this PR do we still want to keep #12266 open, which is bumping to v14?

pettinarip:
> House-keeping question, with this PR do we still want to keep #12266 open, which is bumping to v14?

We will merge v14 after the RC is merged and close this PR after that if all goes well.

PR #12901, opened 2024-05-04
feat(Heading): create a wrapper for default sizes

TylerAPfledderer:
Recreates a `Heading` wrapper component to set default `size` token values based on the heading element supplied through the `as` prop.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `Heading` wrapper component that automatically assigns size theme tokens based on the heading tag used.
  
- **Refactor**
  - Streamlined import statements in `Heading.stories.tsx` and enhanced usage by directly passing required props.
  
- **Documentation**
  - Added comments to clarify the usage of explicit size values in the `HeadingComponent`.


TylerAPfledderer:
@pettinarip curious here: it looks like all the instances of the Chakra Heading component used in prod are explicitly defining a `fontSize` instead of using the `size` token (sometimes an explicit `lineHeight`). 

I take it this in intentional until the components they are a part of are update or replaced for the DS, correct?

I have not touched them here.

TylerAPfledderer:
@coderabbitai review

PR #12905, opened 2024-05-06
[TEST] performance: update size of home hero image

TylerAPfledderer:
This PR shrinks the image used for the home hero to the largest expected width of 1536px.

Currently checking this against lighthouse to see how much improvement this makes on LCP for the home page.


pettinarip (commented):
Thanks for the PR @TylerAPfledderer 

I believe that just having the `fill` prop will not change the requested size for this image. That should be controlled by the `sizes' prop. In this case the image will still be `100vw`.

Regarding the image format, this is now fixed in #12861, we shouldn't need to change the source image.

TylerAPfledderer:
> Thanks for the PR @TylerAPfledderer
> 
> I believe that just having the `fill` prop will not change the requested size for this image. That should be controlled by the `sizes' prop. In this case the image will still be `100vw`.
> 
> Regarding the image format, this is now fixed in #12861, we shouldn't need to change the source image.

Thanks @pettinarip !

I didn't even consider that issue with the source image, but I will revert the two modifications you mentioned, and just focus on have a smaller image size.

TylerAPfledderer:
@pettinarip after those revisions, there doesn't seem to be any change in Lighthouse to performance with a smaller initial image size

pettinarip:
> @pettinarip after those revisions, there doesn't seem to be any change in Lighthouse to performance with a smaller initial image size

There is. So, this change will affect the high resolutions, desktop devices. In those, you will see that the largest image served is the 1500px, ~458kb. While in prod, we serve a 1920px image, ~774kb. This is indeed an improvement in weight.

1. The first question would be whether we want to serve a lower resolution image. This depends on whether or not we want to support 2x or 3x dpr. These devices would require even larger images. cc @nloureiro we have discussed this recently, and I think we want to support them.

2. If we don't want to support those devices, then this change makes sense, but I would implement it in a different way. I'd control the image sizes from the `next.config` file by adjusting the `deviceSizes` option. Basically removing the `1920` size and leaving the `1500` size as the largest.

TylerAPfledderer:
> > @pettinarip after those revisions, there doesn't seem to be any change in Lighthouse to performance with a smaller initial image size
> 
> There is. So, this change will affect the high resolutions, desktop devices. In those, you will see that the largest image served is the 1500px, ~458kb. While in prod, we serve a 1920px image, ~774kb. This is indeed an improvement in weight.
> 
> 1. The first question would be whether we want to serve a lower resolution image. This depends on whether or not we want to support 2x or 3x dpr. These devices would require even larger images. cc @nloureiro we have discussed this recently, and I think we want to support them.
> 2. If we don't want to support those devices, then this change makes sense, but I would implement it in a different way. I'd control the image sizes from the `next.config` file by adjusting the `deviceSizes` option. Basically removing the `1920` size and leaving the `1500` size as the largest.

Oh, good!

As to No. 1, it looks like NextJS `Image` will only render 1x and 2x if no `sizes` value is specified. There would not be a 3x rendered. If that's OK, I can use Chakra's `useBreakpointValue` to check viewport and set `sizes='100vw` or leave it undefined. (This hook should work fine for this purpose, unless there becomes an SSR issue)

For No. 2, Don't you mean `1536`? (That's the largest width present in the DS)
So...
```js
deviceSizes: [640, 750, 828, 1080, 1200, 1536]
```

TylerAPfledderer:
@pettinarip circling back to this to see if you have any thoughts to my response above 😄

pettinarip:
Hey @TylerAPfledderer I've added an improvement on the hero image in this other PR recently https://github.com/ethereum/ethereum-org-website/pull/13041

I've set up a media query to best handle image sizes on larger screens. I think we want to keep the 1900px image version size to support devices with more dpr.

TylerAPfledderer:
Ok! I will bail on this then 😁

PR #12935, opened 2024-05-08
fix: ButtonLink onClick handler

nhsz:
This PR updates the `ButtonLink` component to handle custom `onClick` events and override its behavior when needed. Fixes the issue of `/wallets/find-wallet/` `Visit website` CTA not being tracking the event when clicked.

## Testing

- Go to `/wallets/find-wallet/` now the `Visit website` button should be triggering the custom `handleWalletWebsiteClick` event handler
- Go to any other page like `/community/` to check that the `ButtonLink` component is still working as expected for other cases

Screen Shot 2024-05-08 at 17 06 47

## Preview

https://deploy-preview-12935--ethereumorg.netlify.app/wallets/find-wallet

### DIFF
--- a/src/components/Buttons/ButtonLink.tsx
+++ b/src/components/Buttons/ButtonLink.tsx
@@ -4,12 +4,18 @@ import { BaseLink, type LinkProps } from "@/components/Link"
 import { type MatomoEventOptions, trackCustomEvent } from "@/lib/utils/matomo"
 
 export type ButtonLinkProps = LinkProps &
-  Omit<ButtonProps, "toId"> & {
+  Omit<ButtonProps, "toId" | "onClick"> & {
     customEventOptions?: MatomoEventOptions
   }
 
-const ButtonLink = ({ ...props }: ButtonLinkProps) => {
-  return <Button as={BaseLink} activeStyle={{}} {...props} />
+const ButtonLink = ({ customEventOptions, ...props }: ButtonLinkProps) => {
+  const handleClick = () => {
+    customEventOptions && trackCustomEvent(customEventOptions)
+  }
+
+  return (
+    <Button as={BaseLink} activeStyle={{}} {...props} onClick={handleClick} />
+  )
 }
 
 export default ButtonLink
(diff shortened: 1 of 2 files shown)


src/components/Buttons/ButtonLink.tsx
```diff
@@ -4,16 +4,26 @@ import { BaseLink, type LinkProps } from "@/components/Link"
...
     customEventOptions && trackCustomEvent(customEventOptions)
   }
+
   return (
-    <Button as={BaseLink} activeStyle={{}} {...props} onClick={handleClick} />
+    <Button
+      as={BaseLink}
+      activeStyle={{}}
+      {...props}
+      onClick={onClick ? onClick : handleClick}
```
wackerow, line 25:
```suggestion
      onClick={onClick || handleClick}
      {...props}
```

@nhsz

src/components/Buttons/ButtonLink.tsx
```diff
@@ -4,16 +4,26 @@ import { BaseLink, type LinkProps } from "@/components/Link"
...
-  Omit<ButtonProps, "toId" | "onClick"> & {
+  Omit<ButtonProps, "toId"> & {
     customEventOptions?: MatomoEventOptions
   }
 
-const ButtonLink = ({ customEventOptions, ...props }: ButtonLinkProps) => {
+const ButtonLink = ({
+  customEventOptions,
+  onClick,
+  ...props
```
wackerow, line 14: Hm.. this is a little confusing since we're accepting an `onClick`, which could override the click handler per below, but we also have `customEventOptions` which wouldn't be used if `onClick` is provided... 

@nhsz If we need to accept `onClick` here, should we merge it into the `handleClick()` function below? ie

```tsx
const handleClick = () => {
  customEventOptions && trackCustomEvent(customEventOptions)
  onClick()
}
```

pettinarip: @nhsz @wackerow the underlying [`Button` component already handles this logic](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Buttons/Button.tsx#L43). So I'd say, let's get rid of the `onClick` handler from this component completely and rely on the logic set in the button component. Otherwise, we are (and we were) duplicating this logic.

pettinarip (changes requested):

wackerow (approved):

PR #12939, opened 2024-05-09
Update programming-langauges index.md

setgree:
Looks good! small change: the ledge sentence should say what developers *can* do rather than attempting to correct a misunderstanding that people may or may not have.

### DIFF
--- a/public/content/developers/docs/programming-languages/index.md
+++ b/public/content/developers/docs/programming-languages/index.md
@@ -4,8 +4,7 @@ description:
 lang: en
 ---
 
-A common misconception is that developers must write [smart contracts](/developers/docs/smart-contracts/) in order to build on Ethereum. This is false.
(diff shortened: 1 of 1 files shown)


minimalsm (changes requested):
Thanks for the PR! I believe you may have unintentionally altered the meaning of this content.

Developers can interact with smart contracts using any programming language (for example, if you're building a website, you can access Ethereum using JavaScript). But, as far as I'm aware, smart contracts must be written in a dedicated smart contract language (e.g., Solidity, Vyper, or Fe)

cc: @wackerow @corwintines

corwintines:
Thanks @setgree!

I agree with what @minimalsm has written here. I think the original captures the concept of being able to do more for development in Ethereum than just smart contracts. Im going to close this out in favor of that, but if you disagree please do reopen and let us know!

PR #12942, opened 2024-05-10
Get GitHub file contributors during getStaticProps, cache to avoid rate limit

wackerow:
## Description
Updates fetching for GitHub contributors for each page, removing the client-side fetch, replacing with cached solution during getStaticProps (server-side), removing all error/loading conditionals allowing the component the render immediately.

- Deprecates `useClientSideGitHub*` hooks
- Adds new `fetchGitHistory.ts` script, designed to be run server-side during getStaticProps
- Implements caching solution to avoid repeating API calls for non-English paths, using in-memory object from the `[...slug].tsx]` page component that stores already fetched data.
- This data is now fetched in `getStaticProps`, and passed forward. Updated consuming components to remove all conditionals, as this data should never be in a loading or error state anymore.
- Patch: Removed the `isDesktop` boolean inside FileContributors and replaced with `hideBelow="md"` to fix the initial patch render glitch where this component was empty to start.

## Preview URL
- https://deploy-preview-12942--ethereumorg.netlify.app/en/developers/docs/transactions/


src/layouts/Docs.tsx
```diff
@@ -206,6 +204,7 @@ type DocsLayoutProps = Pick<
   | "lastUpdatedDate"
   | "crowdinContributors"
   | "contentNotTranslated"
+  | "gitContributors"
```
wackerow, line 207: Note that `gitContributors` and `creowdinContributors` will be deprecated in #12969 in favor of a single list of `contributors`

src/components/FileContributors.tsx
```diff
@@ -8,13 +8,11 @@ import {
   ListItem,
   ModalBody,
   ModalHeader,
-  SkeletonText,
```
pettinarip, line 11: bye bye 💀 oh much cleaner now

src/lib/api/fetchGitHistory.ts
```diff
@@ -0,0 +1,81 @@
...
+  while (true) {
+    const response = await fetch(url.href, {
+      headers: { Authorization: `token ${gitHubToken}` },
+    })
+
+    if (
+      response.status === 403 &&
+      response.headers.get("X-RateLimit-Remaining") === "0"
+    ) {
+      const resetTime = response.headers.get("X-RateLimit-Reset") as string
```
pettinarip, line 30: Nice use of headers for this, TIL

pettinarip (approved):
LGMT! good job @wackerow

PR #12943, opened 2024-05-10
fix(storybook): fix alias paths with tsconfig plugin

TylerAPfledderer:
## Description

Adds and configures the [tsconfig-paths-webpack-plugin](https://github.com/dividab/tsconfig-paths-webpack-plugin#tsconfig-paths-webpack-plugin) for webpack to consistently resolve path aliases in TypeScript modules.

See [Storybook Docs](https://storybook.js.org/docs/builders/webpack#typescript-modules-are-not-resolved-within-storybook) regarding this webpack recommendation.

Also updates import paths in existing story files that use imports from the `public` folder.

### Current Behavior

Errors would throw specifically for the path alias to the `public` folder (`@/public`) saying it could not resolve the aliased path. But the alias `@` for the `src` directory would always resolve successfully.

For example, attempt to render a story involving an image being import from `@/public/**`, either in the component or in the story file.

### New Behavior

`tsconfig-paths-webpack-plugin` utilizes the project's tsconfig to handle resolving of path aliases, without the need to explicitly specify them again.

cc @pettinarip, @corwintines

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,7 +1,7 @@
-import path from "path"
-
 import type { StorybookConfig } from "@storybook/nextjs"
 import { propNames } from "@chakra-ui/react"
+
+import TsconfigPathsPlugin from "tsconfig-paths-webpack-plugin"
 
 /**
  * Note regarding package.json settings related to Storybook:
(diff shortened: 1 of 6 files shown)


pettinarip (approved):
Nice, reducing redundancy and fixing issues 👍🏼 Thanks @TylerAPfledderer

corwintines (approved):

PR #12944, opened 2024-05-10
refactor: replace `@chakra-ui/addon` with local setup

TylerAPfledderer:
## Description

In response to attempting to add a RTL language as a locale option for Storybook, this PR removes `@chakra-ui/storybook-addon` in favor of a local decorator setup.

It is expected that the addon will not be updated due to ongoing building of Chakra v3 (which will deprecate the addon). In it's current state, the addon is not set up to properly sync with the changing of locale direction via `i18next`, affecting DX.

Creation of a local decorator similar to the addon expects a resolution to this, along with other modifications.

The `.storybook/types` brings in the `getThemingArgTypes` from the addon to generate SB controls based on the given component theme. (Currently in use in `Button.stories`)

The Farsi language is added to SB as a locale option to generate RTL snapshots in Chromatic.

### DIFF
--- a/.storybook/ChakraDecorator.tsx
+++ b/.storybook/ChakraDecorator.tsx
@@ -1,0 +1,63 @@
+import {
+  ChakraBaseProvider,
+  extendBaseTheme,
+  useColorMode,
+} from "@chakra-ui/react"
+import type { Decorator } from "@storybook/react"
+
+import theme from "../src/@chakra-ui/theme"
+import { useEffect, useMemo, useState } from "react"
+import i18n from "./i18next"
+
+type DecoratorProps = Parameters<Decorator>
+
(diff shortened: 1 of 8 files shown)


pettinarip (approved):
Thanks for digging into this issue and adding the context of what is going on @TylerAPfledderer.

LGTM and rtl is working as expected after this change 👍🏼

TylerAPfledderer:
@pettinarip should I add a story mode for Farsi here to be absolutely sure?

pettinarip:
> @pettinarip should I add a story mode for Farsi here to be absolutely sure?

Aren't you covering that on the other PR related to story modes? Same for me but yes, a specific test for that locale to test RTL would be really helpful.

PR #12952, opened 2024-05-13
performance: replace recharts charting lib with Chart.js

nhsz:
This PR replaces the [`recharts`](https://recharts.org/en-US/) charting lib with [`Chart.js`](https://www.chartjs.org/docs/latest/), with the goal of providing a lighter alternative and reduce the bundle size.

We currently have 2 types of charts on the site, the [`/StatsBoxGrid/GridItem`](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/StatsBoxGrid/GridItem.tsx) on the homepage and [`EnergyConsumptionChart`](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/EnergyConsumptionChart.tsx)

## Description

- components
  - updates `src/components/StatsBoxGrid/GridItem.tsx` with Chart.js implementation
  - updates `src/components/EnergyConsumptionChart.tsx` with Chart.js implementation
- hooks
  - adds `useIsClient` custom hook 
- misc
  - updates `constants.ts`
  - adds `charts.ts` to `/lib/utils` 
- updates dependencies
  - removes [`recharts`](https://recharts.org/en-US/)
  - adds [`chart.js`](https://www.chartjs.org/docs/latest/)
  - adds [`chartjs-plugin-datalabels`](https://chartjs-plugin-datalabels.netlify.app/) (required plugin for custom datalabels)
  - adds [`react-chartjs-2`](https://react-chartjs-2.js.org/) (React Chart.js wrapper)

## Charts location

- `StatsBoxGrid`
  - https://ethereum.org
- `EnergyConsumptionChart`
  - https://ethereum.org/en/energy-consumption/ 
  - https://ethereum.org/en/what-is-ethereum/ 

## Previews

- Homepage charts: https://deploy-preview-12952--ethereumorg.netlify.app/en/
- Energy consumption charts:
  - https://deploy-preview-12952--ethereumorg.netlify.app/en/energy-consumption/
  - https://deploy-preview-12952--ethereumorg.netlify.app/en/what-is-ethereum/

### Chart.js (152.54kb)

Screen Shot 2024-05-21 at 10 51 38

### recharts (271.81kb)

Screen Shot 2024-05-21 at 10 52 00

**Reduction of 43.88%**

## Note

Found a small glitch on the StatsBox chart (don't think it's a blocker) that happens from time to time on some browsers like Chrome/Brave. Works ok on Firefox & Safari. From the comments on github, seems to be a browser issue and can't be fixed on the lib

Screen Shot 2024-06-03 at 14 17 06

Check the comments below:

- https://github.com/chartjs/Chart.js/issues/10727
- https://github.com/chartjs/Chart.js/issues/11310#issuecomment-1597012157


wackerow:
https://github.com/ethereum/ethereum-org-website/assets/54227730/7d318e63-6f09-4b23-8405-3c7c51613899

@nhsz Noticing this when playing around. It's a fairly unnatural approach for someone to resize a window like that, but also noticing it starts getting cramped on mobile:

image
(From a larger iPhone)

nhsz:
> Screen.Recording.2024-05-15.at.09.47.34.mov 
> @nhsz Noticing this when playing around. It's a fairly unnatural approach for someone to resize a window like that, but also noticing it starts getting cramped on mobile:
> 
> image (From a larger iPhone)

@wackerow Yes, still working on the responsive fixes, thanks

src/components/StatsBoxGrid/GridItem.tsx
wackerow, line 1: Having a couple issues here on desktop... In the Netlify preview, the homepage loads correctly showing the 30-day charts, but clicking on the 90-day button for any of them gives me the dreaded `Application error: a client-side exception has occurred (see the browser console for more information).`

image

---

Locally I do not get this, and I am able to toggle between them. I then notice a difference in our transition between 30 and 90 days, where it seems to only be showing 30 days worth of data during the transition (in both directions) leaving a dead space in the chart during the transition.

https://github.com/ethereum/ethereum-org-website/assets/54227730/e1d831f8-1389-4dec-8ca6-f593ead1c1b8

Is there a way to maintain all 90 days in the state while this transitions?

src/components/EnergyConsumptionChart.tsx
```diff
@@ -1,241 +1,264 @@
...
+    "Google",
+    t("energy-consumption-chart-netflix-label"),
+    t("energy-consumption-chart-paypal-label"),
+    t("energy-consumption-chart-airbnb-label"),
+    t("energy-consumption-chart-eth-pos-label"),
+  ]
+
+  const chartOptions = {
+    // chart styles
+    barThickness: 38,
```
wackerow, line 60: Wondering if we could pass a responsive Chakra token here to make the width responsive... or perhaps we could use a maxBarThickness here to cap it on Mobile.

src/components/EnergyConsumptionChart.tsx
```diff
@@ -1,241 +1,264 @@
...
+      // required plugin to display labels on top
+      datalabels: {
+        // https://chartjs-plugin-datalabels.netlify.app/guide/positioning.html#alignment-and-offset
+        anchor: "end" as any, // position of the labels (start, end, center), casted to avoid TS issue
+        align: "end" as any, // alignment of the labels (start, end, center), casted to avoid TS issue
+        offset: -0.5, // distance (in pixels) to pull the label away from the anchor point
+        font: {
+          size: "14px",
+        } as any, // casted to avoid TS issue,
+        color: useColorModeValue("#333333", "#F2F2F2"),
```
wackerow, line 77: Could we use theme tokens here? (and throughout for font sizes / colors, etc)

src/components/EnergyConsumptionChart.tsx
```diff
@@ -1,241 +1,264 @@
...
+    t("energy-consumption-chart-airbnb-label"),
+    t("energy-consumption-chart-eth-pos-label"),
+  ]
+
+  const chartOptions = {
+    // chart styles
+    barThickness: 38,
+    borderRadius: 4,
+    aspectRatio: 1.1,
+    responsive: true,
```
wackerow, line 63: Odd that the chart doesn't re-expand if you shrink then widen the device window... Would think this would take care of it, and the Box it's contained within should be responsive... Will keep an eye out for a possible patch

src/components/EnergyConsumptionChart.tsx
```diff
@@ -1,241 +1,264 @@
...
+  //       amount: 0.02,
+  //       color: "#E50914",
+  //     },
+  //     {
+  //       name: t("energy-consumption-chart-eth-pos-label"),
+  //       amount: 0.0026,
+  //       color: "#C1B6F5",
+  //     },
+  //   ],
+  // })
```
wackerow, line 255: (✂️ eventually)

src/lib/constants.ts
```diff
@@ -149,3 +149,21 @@ export const DEFAULT_GLOSSARY_NS = "glossary"
...
+  "#D7B14A",
+  "#D7B14A",
+  "#71BB8A",
+  "#C1B6F5",
+  "#E50914",
+  "#E50914",
+  "#C1B6F5",
+  "#E50914",
+  "#C1B6F5",
+]
```
wackerow, line 169: ❤️

src/lib/utils/charts.ts
```diff
@@ -0,0 +1,18 @@
...
+  const result = str.split(" ").reduce((res, word, idx) => {
+    if (!idx || word.length > res) {
+      res = word.length
+    }
+
+    return res
+  }, 0)
+  const regex = new RegExp(
+    "(?=\\S).{0," + (result - 1) + "}\\S(?!\\S)|\\S{" + result + "}",
+    "gm"
```
wackerow, line 14: Gm

src/components/Buttons/ButtonLink.tsx
```diff
@@ -4,7 +4,7 @@ import { BaseLink, type LinkProps } from "@/components/Link"
 import { type MatomoEventOptions, trackCustomEvent } from "@/lib/utils/matomo"
 
 export type ButtonLinkProps = LinkProps &
-  Omit<ButtonProps, "toId" | "onClick"> & {
+  Omit<ButtonProps, "toId"> & {
```
wackerow, line 7: Is this from #12935?

src/components/EnergyConsumptionChart.tsx
nhsz: Doesn't work with the tokens

src/components/Buttons/ButtonLink.tsx
nhsz: Fixed

pettinarip (changes requested):
@nhsz good job! Left a couple of observations.

src/components/EnergyConsumptionChart.tsx
```diff
@@ -194,48 +152,112 @@ const EnergyConsumptionChart = () => {
...
+          padding: 0, // removes default padding betwen x-labels and chart
+          maxRotation: 0, // turns off rotation
+          beginAtZero: true,
+        },
+      },
+    },
+  }
+
+  // chart custom legend
+  const chartLegend = wrapLabel(
```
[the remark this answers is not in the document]
nhsz: `label` prop from `cahrtData` expects a `string` or `string[]`, cannot use components with it

src/components/StatsBoxGrid/GridItem.tsx
```diff
@@ -79,48 +88,70 @@ export const GridItem = ({ metric }: GridItemProps) => {
...
+        },
+      },
+    },
+  }
+
+  const chartData = {
+    labels: hasData ? filteredData(state.data) : [],
+    datasets: [
+      {
+        data: hasData ? state.data.map((item) => item.value) : [],
```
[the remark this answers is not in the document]
nhsz: Should be fixed now. `filteredData` values refer to the filtered `timestamp` (x range), while `state.data` contains the values (y range)

src/components/StatsBoxGrid/GridItem.tsx
pettinarip: Noting the same @nhsz. Locally (dev) its working fine but on the preview deploy it is crashing.

wackerow:
https://deploy-preview-12952--ethereumorg.netlify.app/en/energy-consumption

From Safari:
image

From Brave:

https://github.com/ethereum/ethereum-org-website/assets/54227730/7dd1da74-ce64-41ae-8691-44afcc199257

https://github.com/ethereum/ethereum-org-website/assets/54227730/a6595c03-fb11-4f21-ac5e-cff30872b82b

---

Cleared cache and tried this out... seeing some strange behavior on the engergy consumption chart where it glitches by loading/animating the chart twice, and seems to not be placing them in the correct sorting order. These adjust a little when the window size is adjusted.

src/components/EnergyConsumptionChart.tsx
```diff
@@ -1,86 +1,56 @@
...
+)
 
 const EnergyConsumptionChart = () => {
   const { t } = useTranslation("page-what-is-ethereum")
-  const textColor = useToken("colors", "text")
   const { locale } = useRouter()
   const isRtl = isLangRightToLeft(locale as Lang)
 
-  const data = useBreakpointValue<Data>({
+  const useIsClient = () => {
```
pettinarip, line 42: This is a custom hook. It should be defined outside the component (ideally in a separate file in the hooks folder) and used like this
```tsx
const isClient = useIsClient()
```

pettinarip (changes requested):
@nhsz the energy consumption chart is looking better now even though it breaks when resized and the columns start to overlap a bit. I wouldn't consider this last issue as a blocker, as I doubt most users play with the window size.

One thing that could be discussed with the designers (@nloureiro @konopkja) is the height of the chart on mobile. I see that it is ~550px on the preview vs ~350px in prod. Not sure how easy it is to set it to something similar to prod.
image

====
Regarding the home page charts, I still get the error crashing the app in the preview deploy.  I know that we saw that you didn't have this error yesterday but I have double checked this in different browsers (FF, chrome, brave, chrome android) and I still get the error. I'll try to debug a bit this today.

src/components/EnergyConsumptionChart.tsx
nhsz: Fixed

wackerow:
Great, the height looks good on the energy chart now. Responding well between screen sizes, I'm just still getting the double-load animation bug. I thought this was remedied with #13212 but maybe not? @pettinarip @nhsz Do you see this as a blocker?

The homepage charts are working as expected again aside from the slight difference in how it animates. No more crashes or errors when switching between 30/90-day ranges. 

---

Small nice-to-have regarding the chart animation, right now it's anchored to the left which is a little strange since it's the right edge of the chart data that remains the same "day" (today)... would be nice if we could flip the animation anchor point to be the right edge instead.

pettinarip:
@wackerow 

> I'm just still getting the double-load animation bug

There is a hydration issue that is not related to this PR (its a pre-existing bug). We are going to tackle that on a separate PR.

pettinarip (approved):
LGTM @nhsz!

@wackerow the nice to have that you mentioned could be explored and added on a separate PR (to not block this long waiting PR), or even we could deactivate it if it doesn't convince us. Open for that as well. LMK what do you think or if you already have an idea about how we could achieve this.

wackerow (approved):
Looks good! Can tackle anything remaining in separate issue(s)

PR #12954, opened 2024-05-13
Performance: fix compression issues by adjusting middleware matcher

pettinarip:
Our current implementation has odd behavior when serving static files that are compressed. Depending on the browser, it will serve files using `gzip` or `br`, since the same browsers support both algorithms (same req header in all these browsers: `Accept-Encoding: gzip, deflate, br, zstd`).

We would like to always serve the `br` version as it is much lighter than the `gzip` version. For example, the `_app.js` bundle size drops from ~330kb to ~215kb when using Brotli.

The issue is related to the custom middleware we use to redirect requests to different locales. Currently, we process all incoming requests (which is not necessary) and for some reason this seems to cause a conflict with the netlify server which ends up serving `gzip` files instead.

## Description

This PR adds a custom `matcher` to the exported middleware config to only process the matched paths and ignore the rest.

## Summary by CodeRabbit

- **New Features**
  - Enhanced middleware to handle locales and redirects more effectively.
  - Introduced configuration to exclude specific paths from middleware processing.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -41,12 +41,14 @@
 [functions]
 
   [functions.___netlify-odb-handler]
+    external_node_modules = ["sharp"]
     included_files = [
       "./src/intl/**/*",
       "!./public/**/*",
       "node_modules/next/dist/server/future/route-modules/pages/vendored/contexts/router-context*",
       "node_modules/next/dist/server/future/route-modules/pages/vendored/contexts/amp-context*",
       "node_modules/next/dist/server/future/route-modules/pages/vendored/contexts/head-manager-context*",
+      "node_modules/sharp/**/*",
     ]
 
   [functions.___netlify-handler]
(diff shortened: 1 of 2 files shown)


minimalsm (changes requested):
Images aren't loading for me on iOS.

corwintines (changes requested):
Screenshot 2024-05-13 at 8 12 22 PM

Using MacOS, Brave browser, not able to get images here as well.

pettinarip:
@minimalsm @corwintines good catch, there seem to be a few weird things going on with redirects/middleware/trailingslashes with Next 14. Investigating...

minimalsm:
@coderabbitai review

corwintines (approved):
Working for me now, LGTM

minimalsm (approved):
LGTM now :) 

Images fixes, and seeing br encoding on js and css files.

PR #12969, opened 2024-05-15
Deprecate CrowdinContributors, GitHubContributors

wackerow:
Sub-branch PR: `get-static-contributors <- get-static-contributors-consolidation`

## Description
- Deprecates CrowdinContributors and GitHubContributors components in lieu of just using the FileContributors component
- Builds on server-side logic from #12942 to compile a single list of `contributors` that will be passed to the layout and components from `[...slug].tsx`.
- For non-English pages, this will be a list of Crowdin contributors; else it will be GitHub contributors
- Extracts logic into `contributors.ts` util file


src/lib/utils/contributors.ts
```diff
@@ -0,0 +1,47 @@
...
+import { getLastModifiedDate } from "./gh"
+
+import { fetchAndCacheGitContributors } from "@/lib/api/fetchGitHistory"
+
+export const getFileContributorInfo = async (
+  mdDir: string,
+  mdPath: string,
+  slug: string,
+  locale: string,
+  fileLang: string,
```
pettinarip, line 20: Is there a scenario where these two are different?

src/lib/utils/contributors.ts
```diff
@@ -0,0 +1,47 @@
...
+  getCrowdinContributors,
+} from "./crowdin"
+import { getLastModifiedDate } from "./gh"
+
+import { fetchAndCacheGitContributors } from "@/lib/api/fetchGitHistory"
+
+export const getFileContributorInfo = async (
+  mdDir: string,
+  mdPath: string,
+  slug: string,
```
pettinarip, line 18: Minor: I see too many args for this function. For example, `mdDir` and `mdPath` are derived from `slug`. Perhaps we could abstract that logic (how to calc mdDir and mdPath) somewhere in the utils and reduce the noise form this function.

IMO, ideally, this function should only require 2 args: slug and locale (the cache is arguable but ok).

src/lib/utils/contributors.ts
```diff
@@ -0,0 +1,47 @@
...
+  locale: string,
+  fileLang: string,
+  layout: Layout,
```
pettinarip, line 21: I see we use this arg to determine whether or not we want to fetch the crowdin contributors. Couldn't we deduce this from the slug? `slug.includes("/developers/")`

pettinarip (approved):
Great job @wackerow! a very useful refactor.

Left a few minor comments related to the `getFileContributorInfo` function but nothing blocking.

corwintines (approved):

PR #12972, opened 2024-05-16
Adjust wording on dev docs /consensus-mechanisms/pos

CodeMongerrr:
The word node( alone ) , could potentially confuse users to figure out that non vaidators could also be block proposers. As Validator was not used, only the term Validator Client mentioned later could confirm it.

## Description
Added a simple clarification word, because it confused me a little, could confuse others too.
## Related Issue

Its a minor change so I did not create a seperate Issue for it.


public/content/developers/docs/consensus-mechanisms/pos/index.md
```diff
@@ -27,7 +27,7 @@ The following provides an end-to-end explanation of how a transaction gets execu
...
+4. One of the nodes( A Validator ) on the network is the block proposer for the current slot, having previously been selected pseudo-randomly using RANDAO. This node is responsible for building and broadcasting the next block to be added to the Ethereum blockchain and updating the global state. The node is made up of three parts: an execution client, a consensus client and a validator client. The execution client bundles transactions from the local mempool into an "execution payload" and executes them locally to generate a state change. This information is passed to the consensus client where the execution payload is wrapped as part of a "beacon block" that also contains information about rewards, penalties, slashings, attestations etc. that enable the network to agree on the sequence of blocks at the head of the chain. The communication between the execution and consensus clients is described in more detail in [Connecting the Consensus and Execution Clients](/developers/docs/networking-layer/#connecting-clients).
```
wackerow, line 30: Hey, thanks for the suggestion @CodeMongerrr.. I see what you're going at here... let's adjust to make it flow easier with just "One of the validator nodes"

```suggestion
4. One of the validator nodes on the network is the block proposer for the current slot, having previously been selected pseudo-randomly using RANDAO. This node is responsible for building and broadcasting the next block to be added to the Ethereum blockchain and updating the global state. The node is made up of three parts: an execution client, a consensus client and a validator client. The execution client bundles transactions from the local mempool into an "execution payload" and executes them locally to generate a state change. This information is passed to the consensus client where the execution payload is wrapped as part of a "beacon block" that also contains information about rewards, penalties, slashings, attestations etc. that enable the network to agree on the sequence of blocks at the head of the chain. The communication between the execution and consensus clients is described in more detail in [Connecting the Consensus and Execution Clients](/developers/docs/networking-layer/#connecting-clients).
```

Will commit this then bring in

wackerow (approved):
Thanks @CodeMongerrr!

wackerow:
@all-contributors please add @CodeMongerrr for idea

PR #13005, opened 2024-05-21
Comment out TranslationBanner components until logic is fixed

corwintines:
## Description

Commenting out TranslationBanner components until logic for rendering right banner is resolved


minimalsm (approved):
LGTM :)!

PR #13012, opened 2024-05-21
[Draft] - add query parameter support to find wallets page

### LINKED ISSUE #9209: Support URL parameters for pre-applied filters on find wallets page

It would be helpful for the find wallet page to support URL parameters and apply filters based on these parameters. This would allow us to link out to the find wallets page with a set of query parameters and see wallets with that filter applied.

For example, the NFT page links out to find wallets with a query parameter and this gets applied filtering down the wallets to ones that support NFTs.

Support URL parameters being passed and applying a filter on page load. 

Ex url: /wallets/find-wallet?device=mobile&buy_crypto=true

This would then apply filters for mobile and buy_crypto on page load.

No other alternative at the moment.

### Additional context

_No response_

### LINKED ISSUE #12646: Feature request: list of wallets filters saved in URL

Our list of wallets offers a lot of filters to narrow down the huge list of products. Some users during user testing requested ability to send others a tailored filtered list view. This request has also appeared in user surveys further confirming potential need.

**Currently we do not save switched on filters in URL.** 

**Feature request:**
Add filter parameters to URL so that the website and its users can link to a specific view of wallets. This can be used also in how to guides and other content on ethereum.org to improve UX.

- Every triggered filter should be added as a parameter to URL so that we can link to the filtered view.

Screenshot 2024-04-03 at 12 06 39

If too difficult, consider at least adding personas to URL

### Additional context

_No response_

- [ ] Yes
- [X] No

corwintines:
## Description

- Adds query support to the find wallet page. 
    - Support for personas as query parameters
    - Support for activeFilters as query parameters

## Related Issue

- Fixes #9209
- Fixes #12646

### DIFF
--- a/src/components/FindWallet/MobileFiltersMenu.tsx
+++ b/src/components/FindWallet/MobileFiltersMenu.tsx
@@ -38,8 +38,6 @@ export const MobileFiltersMenu = ({
   updateFilterOption,
   updateFilterOptions,
   selectedPersona,
-  setFilters,
-  setSelectedPersona,
   showMobileSidebar,
   onOpen,
   onClose,
@@ -83,9 +81,7 @@ export const MobileFiltersMenu = ({
 
               <WalletFilterPersona
                 resetFilters={resetFilters}
-                setFilters={setFilters}
                 selectedPersona={selectedPersona}
-                setSelectedPersona={setSelectedPersona}
                 showMobileSidebar={showMobileSidebar}
               />
             </Box>
@@ -99,8 +95,6 @@ export const MobileFiltersMenu = ({
                 updateFilterOptions,
                 resetFilters,
                 selectedPersona,
-                setFilters,
-                setSelectedPersona,
                 showMobileSidebar,
               }}
             />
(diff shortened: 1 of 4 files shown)


minimalsm (changes requested):
When navigating to a link with query parameters, there is a significant delay (2-3 seconds) from page load until the correct wallets are displayed. This is quite janky IMO.

This feature also creates a strange user experience... when you navigate to the wallet page, select a filter, move to a different page, and then go back to the wallet page, you initially see one list of wallets, which is then replaced by a different list due to the query parameters. This happens without a clear indication of why, leading to confusion.

From the issue we're trying to solve here:

> It would be helpful for the find wallet page to support URL parameters and apply filters based on these parameters. This would allow us to link out to the find wallets page with a set of query parameters and see wallets with that filter applied.
> 
> For example, the NFT page links out to find wallets with a query parameter and this gets applied filtering down the wallets to ones that support NFTs.

If we really want this feature (which, to be honest, I'm not entirely convinced of its usefulness), I think we should focus on allowing us to link to a page with query parameters, but not adding them on user selection. 

Even still, I think we need to reduce the wait time or add a loading state. If either the loading state or any other changes disrupt the experience for the majority of users who will use this page 'normally', I'd still advocate against this feature.

corwintines:
Going to close this in favor of https://github.com/ethereum/ethereum-org-website/pull/13377

PR #13018, opened 2024-05-22
Fix: add explicit matcher for the root path to work on dev mode

pettinarip:
## Description

Adds missing root route to the middleware matcher.

### DIFF
--- a/src/middleware.ts
+++ b/src/middleware.ts
@@ -23,13 +23,11 @@ function detectLocale(acceptLanguage: string | null) {
 
 export const config = {
   matcher: [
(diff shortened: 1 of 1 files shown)


wackerow (approved):

wackerow:
@pettinarip Any reason this is in draft still? This seems to be fixing the localhost issues we were having.

pettinarip:
> @pettinarip Any reason this is in draft still? This seems to be fixing the localhost issues we were having.

Just wanted to avoid an early merge bc I didn't know if this was going to work on Netlify. It looks like it is also working fine on the Netlify servers.

PR #13024, opened 2024-05-23
Performance: optimize hero image for use cases layout

pettinarip:
## Description

Adds the missing `sizes` prop for the hero image in the use cases layout. This will create lighter images for smaller resolutions.

Example https://deploy-preview-13024--ethereumorg.netlify.app/en/nft/

### DIFF
--- a/src/layouts/UseCases.tsx
+++ b/src/layouts/UseCases.tsx
@@ -216,7 +216,7 @@ export const UseCasesLayout = ({
           width={1200}
           height={610}
           style={{ objectFit: "cover" }}
-          sizes="(max-width: 992px) 100vw, 50vw"
+          sizes="(max-width: 992px) 100vw, 750px"
           priority
           alignSelf={{
             base: "center",
--- a/src/layouts/UseCases.tsx
+++ b/src/layouts/UseCases.tsx
@@ -216,6 +216,7 @@ export const UseCasesLayout = ({
           width={1200}
           height={610}
           style={{ objectFit: "cover" }}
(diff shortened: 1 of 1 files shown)


nhsz:
@pettinarip seems to be ignoring the max-width and generating more images compared to current `dev`, or am I misunderstanding the `srcset`

PR version (https://deploy-preview-13024--ethereumorg.netlify.app/en/nft/)

Screen Shot 2024-05-29 at 19 51 19

---

`dev` version (https://dev--ethereumorg.netlify.app/en/nft/)

Screen Shot 2024-05-29 at 19 51 41

pettinarip:
> @pettinarip the img seems to be ignoring the max-width and generating more images compared to current `dev`, or am I misunderstanding the `srcset`

Correct, from the [nextjs `sizes` docs](https://nextjs.org/docs/app/api-reference/components/image#sizes):
>  If no sizes value is present, a small srcset is generated, suitable for a fixed-size image (1x/2x/etc). If sizes is defined, a large srcset is generated, suitable for a responsive image (640w/750w/etc).

So, when you define a `sizes` you will get a more granular srcset that depends on width rather than just dpr.

Example of optimizations: the current `dev`s version serves **always** a 1200px image, no matter what resolution you are using. If you are on a mobile phone with 1x dpr, you will get a 1200px image. With the `sizes` improvement you will get a ~640px image depending on the res of the phone.

I've pushed a new commit that adjust the expected size on big screen resolutions from `50vw` to just `750px` since this is going to be the max width an image will be in this case (> 992px).

nhsz (approved):
LGTM

PR #13037, opened 2024-05-26
Added context to how the block max limit can be changed

JacobSharples:
Added two sentences on how the block gas limit can be changed by consensus


wackerow:
Hey @JacobSharples, guess I wasn't familiar with this exact detail... do you have a reference for this by chance?

corwintines:
I've been trying to find information on this. But, I haven't seen this in EIP-1559. 

I know that block size can go up to 30 million gas, but that is the limit and there isn't an in protocol way to increase this. Do you have a source this information so I can better understand this change @JacobSharples

JacobSharples:
Apologies for not attaching more information before. I used the [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) as a reference
image

after seeing some blocks with a gas limit > 30M. I started digging into this after I saw the website [PumpTheGas](https://pumpthegas.org/ ) and learned that stakers could change their max limit with a flag. However, to ensure that all nodes can execute all transactions, the change in the gas limit is capped (this is my understanding, there definitely seems to be a lack of info on this topic).


The gas limit change is restricted on how or low it change in a single block:
image



For example, [block 20170216](https://etherscan.io/block/20170216) has a max limit 30,029,295 (the parent and child block is 30M).

Doing the math, the max gas increase from a parent block of 30M is 30,000,000//1024 = 29,295

wackerow (approved):
Very cool, thanks for this @JacobSharples!

wackerow:
@all-contributors please add @JacobSharples for content

PR #13040, opened 2024-05-27
Bump netlify/next runtime and fix api routes in middleware 

pettinarip:
## Description
Things done in this PR:
- bump `@netlify/plugin-nextjs` to the latest, v5
- since in this latest version, the `included_files` from `netlify.toml` doesn't seem to work, we excluded the heavy files from Netlify function bundles in the `next.config.js` file
- ignore `/api` routes from middleware

Bonus:
This also fix the issue we currently have when navigating to the home page by using the eth logo on the navbar where it was doing a full page nav instead of a client side one.


pettinarip (commented):
~When navigating to a non-existing nested page like `/en/dapps/test` we enter in a redirect loop.~

src/middleware.ts
```diff
@@ -48,7 +51,7 @@ export async function middleware(req: NextRequest) {
     const localeDetected = detectLocale(req.headers.get("accept-language"))
     const locale = localeDetected || DEFAULT_LOCALE
 
-    const redirectUrl = new URL(`/${locale}${pathname}`, req.url)
+    const redirectUrl = new URL(`/${locale}${pathname}${search}`, req.url)
```
pettinarip, line 54: Adding the `search` params to the redirect, thing that was missing.

wackerow (approved):

PR #13065, opened 2024-05-29
performance: lazy load nav menu content

nhsz:
## Description

- extracts `MenuContent` component
- `MenuContent` lazy loading
- updates `Nav/Menu` component
- updates `@radix-ui` imports to bring only neccesary components


wackerow (approved):
Interface seems to be working well, nice job @nhsz! Looks like the Lighthouse performance metric is scoring about 9 points higher.

pettinarip (approved):
LGTM!

PR #13066, opened 2024-05-29
Performance: fix hydration issues on find wallets page

pettinarip:
## Description

Fixes hydration issues by:
- sorting the wallets list at build time
- replacing the table/th/tr/td components with divs because it was throwing invalid html errors
- format the languages support for each wallet at build time

## Tests results
Main improvement in Task 2 (first render), nearly 1 second faster using CPU 4x slowdown + slow 4g connection.

Current dev
- Task 1: 500ms
- **Task 2: 3s**
- Task 3: 1.5s

Fixing hydration issue
- Task 1: 500ms
- **Task 2: 2s**
- Task 3: 1.2s

image

## Preview URL
- https://deploy-preview-13066--ethereumorg.netlify.app/en/wallets/find-wallet

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -581,7 +581,7 @@ export interface WalletData {
   new_to_crypto?: boolean
 }
 
-export interface Wallet extends WalletData {
+export type Wallet = WalletData & {
   supportedLanguages: string[]
 }
 
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -581,6 +581,10 @@ export interface WalletData {
   new_to_crypto?: boolean
 }
 
+export interface Wallet extends WalletData {
+  supportedLanguages: string[]
+}
+
 export type WalletFilter = typeof WALLETS_FILTERS_DEFAULT
 
 export interface WalletFilterData {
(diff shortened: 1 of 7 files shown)


src/lib/types.ts
```diff
@@ -581,6 +581,10 @@ export interface WalletData {
   new_to_crypto?: boolean
 }
 
+export interface Wallet extends WalletData {
+  supportedLanguages: string[]
+}
+
```
wackerow, line 587: Could we use a type here? (it's inside `types.ts`)

```suggestion
export type Wallet = WalletData & {
  supportedLanguages: string[]
}

```

pettinarip: Sure. I think we used more intersections than extensions, right? I did that because `WalletData` is already an interface.

wackerow (approved):
@pettinarip Am I correct that this would update with a new sort order every time the site is built/deployed?

This would put it around a 2-week cadence; have we considered (or tried) the `revalidate` flag to tell the server to fetch this again periodically (maybe once a day)? Don't see that as a blocker, but would be nice to increase the frequency imo.

pettinarip:
> @pettinarip Am I correct that this would update with a new sort order every time the site is built/deployed?
> 
> This would put it around a 2-week cadence; have we considered (or tried) the `revalidate` flag to tell the server to fetch this again periodically (maybe once a day)? Don't see that as a blocker, but would be nice to increase the frequency imo.

Great point. Completely forgot to add that....https://github.com/ethereum/ethereum-org-website/pull/13138

PR #13097, opened 2024-06-01
Correct 2 things on page-developers

woseK:
## Description

Changed the Polish translation in page-developers-docs.json file and added a missing dot in the page-developers-index.json so that all the 4 texts end with a dot


minimalsm (approved):
LGTM, thanks @woseK! :)

PR #13099, opened 2024-06-02
feat: Create stories for DS color scheme

TylerAPfledderer:
## Description

Adds a set of stories to Storybook, showcasing the Color scheme.

Only shows the base color tokens and the semantic colors.

Does not create Chromatic snapshots


TylerAPfledderer:
@pettinarip I am not a fan of placing this in `components/BaseStories` as we are not featuring a component here. Maybe have a `stories` folder in `@chakra-ui`?

pettinarip:
Wohoo! this is great man. cc @nloureiro look at this 💪🏼 

> @pettinarip I am not a fan of placing this in `components/BaseStories` as we are not featuring a component here. Maybe have a `stories` folder in `@chakra-ui`?

Good point and good idea. Lets do that. It will be more organized.

nloureiro:
Awesome!!! 
Thank you @TylerAPfledderer 

Just a heads up, we probably will do a big update on the colors alongside the homepage EPIC.
This is a great base to bootstrap this change later next quarter.

TylerAPfledderer:
@pettinarip @nloureiro as an FYI, I am hoping to flip this over to creating the [stories via MDX](https://storybook.js.org/docs/writing-docs/mdx).

For Example, check out the [Narmi Design System Color Stories](https://narmi.github.io/design_system/?path=/docs/design-tokens-color--docs) which have the MDX doc of different groups of colors in their DS, which in turn generates the individual stories for each group.

This would probably be better discussed with overall consideration in adding docs to the other stories from the DS.

src/@chakra-ui/stories/Colors.stories.tsx
```diff
@@ -0,0 +1,184 @@
...
+            <HStack>
+              {primitiveColorsMap[color].map(([tokenKey, value]) => (
+                <Stack key={`${color}${tokenKey}`} direction="column">
+                  <Box>
+                    <Square size="20" bg={`${color}.${tokenKey}`}></Square>
+                  </Box>
+                  <Stack direction="column">
+                    <Text size="sm">{value}</Text>
+                    <Text size="sm">
+                      {color}.{tokenKey}
```
pettinarip, line 52: Good job @TylerAPfledderer. I like how we are displaying the colors in the `Primitives` case.

Two things I think would be useful for devs to display on the semantic tokens story:
1. code example or semantic token usage
2. primitive color reference

For example, for the primary "base" token we could add:
1. how it would be used in code: `primary.base`
2. the primitive color used by this token: `blue.500` or `orange.500`

pettinarip (commented):
LGTM @TylerAPfledderer, left one potential improvement.

src/@chakra-ui/stories/Colors.stories.tsx
TylerAPfledderer: @pettinarip will do! I think I have an approach in mind.

TylerAPfledderer: @pettinarip I updated the labelling of the semantic tokens. It got a little tricky trying to come up with a good way to extract the foundation tokens for each. I don't want to try and spend too much time trying to come up with some generator that would get messy in this PR.

src/@chakra-ui/stories/Colors.stories.tsx
```diff
@@ -0,0 +1,185 @@
...
+}
+
+const ColorGroupWrapper = ({
+  color,
+  children,
+}: {
+  color: string
+  children: ReactNode
+}) => {
+  return (
```
wackerow, line 72: image

@TylerAPfledderer Getting this when I look at the primitives in Dark Mode. Otherwise this all looks great!

TylerAPfledderer: @wackerow yea... they should not be view in dark mode. Only light mode. I did not create this story with that in mind.

Need to figure out how I can update the SB config to force a color mode! That's been a back-burner task with the Chakra provider

src/@chakra-ui/stories/Colors.stories.tsx
pettinarip: Sure. Cool, it looks more useful now with the actual tokens that we use on the code, thanks!

src/@chakra-ui/stories/Colors.stories.tsx
pettinarip: Maybe we could just remove the `color` and `bg` to leave it under control of the parent.

Would look like this:
image

TylerAPfledderer: I didn't think that would make sense because we don't see the orange shades in light mode.

TylerAPfledderer: Maybe here I could more easily show the blues or the oranges based on color mode. I'll take a look at that.

corwintines (approved):
Overall I think this is ready to pull in as is, and can make any improvements in other PRs. Any objections to pulling this in @pettinarip @TylerAPfledderer @wackerow?

src/@chakra-ui/stories/Colors.stories.tsx
pettinarip: Not sure I followed you there @TylerAPfledderer.

IMO we should not do anything here in terms of color mode, since the purpose of this story is to list the color palette we use. The color mode should only be reflected within the semantic tokens story.

TylerAPfledderer: Then I would try and disable the color mode toggle on the toolbar. We don't want to be switching modes because we just want to view everything at the same time (mimicking what it looks like in the figma file)

TylerAPfledderer: Then I would try and disable the color mode toggle on the Storybook toolbar when this is previewed. We don't want to be switching modes because we just want to view everything at the same time (mimicking what it looks like in the figma file)

PR #13102, opened 2024-06-03
Performance: bundle optimization, ui libs chunk

pettinarip:


pettinarip:
Closing as we don't see a clear improvement on loading times.

PR #13109, opened 2024-06-05
Clarify account in the context of a wallet

### LINKED ISSUE #12061: Contradiction between smart contract accounts and wallets
In the following section, it describes both "contract accounts", and then goes on to say, "An account is the keypair for a user-owned Ethereum account." So therefore a contract account is not an account?

Happy to suggest a revision through a PR!

https://github.com/ethereum/ethereum-org-website/blob/9c48ba8c9305a351c815aa494974923f9dbb0b97/public/content/developers/docs/accounts/index.md?plain=1#L94-L112

shoenseiwaso:
## Description
Minor update to clarify the meaning of an account in the context of a wallet, considering both EOAs as well as contract accounts.

## Related Issue

- Closes #12061

### DIFF
--- a/public/content/developers/docs/accounts/index.md
+++ b/public/content/developers/docs/accounts/index.md
@@ -109,7 +109,7 @@ There is also another type of key in Ethereum, introduced when Ethereum switched
 
 ## A note on wallets {#a-note-on-wallets}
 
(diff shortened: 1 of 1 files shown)


wackerow (approved):
Thanks @shoenseiwaso, looks good!

PR #13121, opened 2024-06-07
performance: lazy load Codeblock component

nhsz:


wackerow:
https://github.com/ethereum/ethereum-org-website/assets/54227730/16f3cfc1-5696-4ff8-b6db-d686bccc8f0e

@nhsz Noticing a small delay when clicking on these the first time... After scrolling down, clicking on one of the smart contract examples there is first a short delay where nothing happens, then the Title Bar for the popup appears at the bottom (but no content), then finally the content loads.

Is there a way we could populate something that occupies space here while the component loads?

pettinarip:
Agree with @wackerow, I think we need a holistic solution that can be reused in other modalities that have similar async behavior. We might need some designer eyes/ideas on this cc @nloureiro @konopkja 

That said, on the technical side, I think we could:
- Keep loading the `CodeModal` in a sync way to have an immediate interactive action to the user
- Async load just the `Codeblock` which is the real heavy dep in this case (it has to load `prism`)
- Display a loading/skeleton component while we are loading the content by using the `loading` prop from `dynamic`
```tsx
const Codeblock = dynamic(() => import("@/components/Codeblock"), {
  ssr: false,
  loading: () => <p>Loading...</p>,
})
```

nhsz:
@pettinarip @wackerow I think to unblock this I can update the PR to lazy load the `Codeblock` only for now, until we have a better solution for the modal

nhsz:
@pettinarip @wackerow I've updated the PR to lazy-load the `Codeblock` component only, adding a skeleton during loading

wackerow:
https://github.com/ethereum/ethereum-org-website/assets/54227730/185ee459-a860-4557-9745-de4b51e25660

image

@nloureiro Curious how this looks to you? Seems to fill the space... the skeleton lines kinda like like they glitch out a bit(not super natural looking I guess)... works good enough though I suppose.

pettinarip (commented):
Nice! I now see that these two chunks (prism dep & code example content) are loaded just when they are needed 👍🏼 
image

Waiting for designers approves. From the code side looks good.

nhsz:
> > Screen.Recording.2024-06-19.at.07.54.10.mov
> > image
> > @nloureiro Curious how this looks to you? Seems to fill the space... the skeleton lines kinda like like they glitch out a bit(not super natural looking I guess)... works good enough though I suppose.
> 
> err... no, this seems to be a visual glitch
> 
> if we want to keep the skeleton, we need to make the lines as high as the text size with the same spacing between them and make it a fake loading for a quick load. Like never less than Xxms because ti can't be too fast; otherwise feels like a glitch. Normally, it is animated, but I think we can try without animation.
> 
> Maybe something like this? ske

- I've updated the skeleton lines height to match the code font size
- **same spacing between them:** don't think this would be possible as we have different code samples underneath, not plain text with uniform spacing, and the skeleton is always the same. Also applies to the number of lines
- **make it a fake loading for a quick load:** not sure about this one, this would also add extra loading time for slow connections and unnecesary logic IMO, but I'd leave that to what @pettinarip and @wackerow think

pettinarip:
@nloureiro I understand your point about having a minimum time to avoid flickering on fast connections, but on the other hand I feel this is an unnecessary wait and logic. However, I can be convinced to add it if you feel strongly about it. Most of the time, most users won't feel this as a flicker or glitch.

Another important thing to note is that the skeleton is only displayed the first time the modal is loaded or opened. Once the code is loaded, you won't see the skeleton the next time you open the modal. On the @wackerow demo, you can see that he is reloading the page to see the skeleton but you can test it yourself on the preview deploy.

nloureiro:
> @nloureiro I understand your point about having a minimum time to avoid flickering on fast connections, but on the other hand I feel this is an unnecessary wait and logic. However, I can be convinced to add it if you feel strongly about it. Most of the time, most users won't feel this as a flicker or glitch.
> 
> Another important thing to note is that the skeleton is only displayed the first time the modal is loaded or opened. Once the code is loaded, you won't see the skeleton the next time you open the modal. On the @wackerow demo, you can see that he is reloading the page to see the skeleton but you can test it yourself on the preview deploy.

I don't want to make a big deal out of this, but it is weird as it is.

This is how I see it, and it feels like a glitch. skeletons are harder to do than we imagine with all the details and animations that they need to be well-implemented
https://github.com/ethereum/ethereum-org-website/assets/1120748/de083169-6cdc-46fd-93dd-f96ef75a212f

Maybe play with a loading gif, we can use a retro style because it's the command line... dunno

pettinarip:
@nloureiro ok, yeah makes sense. What minimum time do you have in mind?

In terms of using a skeleton or a gif, either is fine for me. Up to you I would say, if you have time to create something.

@nhsz perhaps we need to switch it to use `lazy/Suspense` instead of dynamic. Found this code that could help us on this case:
```ts
const Codeblock = lazy(() =>
  Promise.all([
    import("@/components/Codeblock"),
    new Promise((resolve) => setTimeout(resolve, 500)),
  ]).then(([module]) => module)
)
```

nloureiro:
> @nloureiro ok, yeah makes sense. What minimum time do you have in mind?
> 
> In terms of using a skeleton or a gif, either is fine for me. Up to you I would say, if you have time to create something.
> 
> @nhsz perhaps we need to switch it to use `lazy/Suspense` instead of dynamic. Found this code that could help us on this case:
> 
> ```ts
> const Codeblock = lazy(() =>
>   Promise.all([
>     import("@/components/Codeblock"),
>     new Promise((resolve) => setTimeout(resolve, 500)),
>   ]).then(([module]) => module)
> )
> ```


From the top of my mind, I'm not sure of time, 1s?

we can keep the skeleton with this minimum time on screen

pettinarip (approved):
The styles changes on the skeleton looks good @nloureiro 

Merging now!

PR #13139, opened 2024-06-11
Performance: find wallets first render optimizations

pettinarip:
## Description

Refactor the find wallet logic to optimize the first render.

**Current dev**
Task 1: 550ms
Task 2: 2.2s - 2.7s
Task 3: 1.4s

**With optimizations**
Task 1: 500ms
Task 2: 1.8s
Task 3: 1.2s


src/components/FindWallet/MobileFiltersMenu.tsx
```diff
@@ -47,7 +43,6 @@ export const MobileFiltersMenu = ({
   onClose,
 }: MobileFiltersMenuProps) => {
   const { t } = useTranslation("page-wallets-find-wallet")
-  const { filteredWallets } = useWalletTable({ filters, t, walletData })
```
pettinarip, line 50: Moved this hook to the parent (page component) to share the results and minimize the calls we do for it since its doing the main filtering.

src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -196,28 +196,30 @@ const FlexInfoCenter = (props: { children: ReactNode; className?: string }) => (
...
+  featureDropdownItems,
   onOpen,
 }: WalletTableProps) => {
   const { t } = useTranslation("page-wallets-find-wallet")
-  const {
-    featureDropdownItems,
-    filteredWallets,
-    updateMoreInfo,
-    walletCardData,
-  } = useWalletTable({ filters, t, walletData })
```
pettinarip, line 220: As mentioned before, moved this to the parent.

src/components/FindWallet/WalletTable/index.tsx
```diff
@@ -232,6 +234,17 @@ const WalletTable = ({
     })
   }
 
+  const showMoreInfo = (wallet) => {
```
pettinarip, line 237: Moved this function creation outside of the `.map` loop.

src/hooks/useWalletTable.tsx
```diff
@@ -120,20 +120,12 @@ export const useWalletTable = ({
     },
   ]
 
-  const [walletCardData, setWalletData] = useState(
+  const [walletCardData, setWalletData] = useState(() =>
```
pettinarip, line 123: Use the initialized function to run this map only once.

src/pages/wallets/find-wallet.tsx
```diff
@@ -216,28 +232,33 @@ const FindWalletPage = ({
...
-                updateFilterOption,
-                updateFilterOptions,
-                resetFilters,
-                selectedPersona,
-                setFilters,
-                setSelectedPersona,
-              }}
-            />
+            {/* Use `Show` instead of `hideBelow` prop to avoid rendering the sidebar on mobile */}
+            <Show above="lg">
```
pettinarip, line 236: This will add a bit of CLS on the table section on desktop versions. I think it is worth it considering that we are removing this from the render of slower mobile devices.
image

wackerow: Okay.. yeah, if we're preventing a large unnecessary chunk of data from loading then I think it's worth it.

wackerow (approved):
Looks good, nice job @pettinarip!

PR #13141, opened 2024-06-11
Performance: lazy load menus

pettinarip:
This PR loads the mobile menu async using `lazy` and `Suspense` for better UX and TTI metrics. The idea is to display these interactive elements only when they are ready to be used, which mostly affects the users with a slow connection.

Details:
- Refactor the desktop menu code: moved it to its own separate component for readability purposes
- Lazy load mobile menu component
- Wrapped the mobile menu (lazy loaded) + the search (sync) with `Suspense` to display both at the same time


nloureiro (approved):

netlify.toml
```diff
@@ -11,6 +11,7 @@
 
 [build.environment]
   NEXT_FORCE_EDGE_IMAGES="true"
+  BUILD_LOCALES="en"
```
pettinarip, line 14:
```suggestion
```

src/components/Nav/index.tsx
```diff
@@ -102,83 +63,26 @@ const Nav = () => {
...
-              <MobileNavMenu
-                {...mobileNavProps}
-                linkSections={linkSections}
-                hideFrom="md"
-                toggleSearch={searchModalDisclosure.onOpen}
-                drawerContainerRef={navWrapperRef}
-              />
+              <Show below="md">
+                {/* Mobile */}
+                {/* use Suspense to display the Search & the Meny at the same time */}
```
pettinarip, line 75:
```suggestion
                {/* use Suspense to display the Search & the Menu at the same time */}
```

src/components/Nav/Desktop/index.tsx
```diff
@@ -0,0 +1,113 @@
...
+  useDisclosure,
+  useEventListener,
+} from "@chakra-ui/react"
+
+import { IconButton } from "@/components/Buttons"
+import LanguagePicker from "@/components/LanguagePicker"
+
+import { DESKTOP_LANGUAGE_BUTTON_NAME, NAV_PY } from "@/lib/constants"
+
+function DesktopNavMenu({ toggleColorMode }: { toggleColorMode: () => void }) {
```
wackerow, line 22: @pettinarip Arrow function with extracted props for consistency?

```suggestion
type DesktopNavMenuProps = {
  toggleColorMode: () => void
}

const DesktopNavMenu = ({ toggleColorMode }: DesktopNavMenuProps) => {
```

src/components/Nav/index.tsx
```diff
@@ -102,83 +63,26 @@ const Nav = () => {
...
+              </Show>
 
-              <MobileNavMenu
-                {...mobileNavProps}
-                linkSections={linkSections}
-                hideFrom="md"
-                toggleSearch={searchModalDisclosure.onOpen}
-                drawerContainerRef={navWrapperRef}
-              />
+              <Show below="md">
```
wackerow, line 73: image

Edge case at the exact `md` breakpoint width (768px) where both are showing up... seems `above` and `below` are both inclusive of the breakpoint.

Maybe this instead?

```suggestion
              <Hide sbove="md">
```

(If we do this, need to import `Hide` above)

src/components/Nav/index.tsx
```diff
@@ -102,83 +63,26 @@ const Nav = () => {
...
+                {/* Mobile */}
+                {/* use Suspense to display the Search & the Menu at the same time */}
```
wackerow, line 75: > use Suspense to display the Search & the Menu at the same time

I see we have `<Search {...searchModalDisclosure} />` declared both in the Desktop suspense and Mobile suspense... I take it this is why? So items render together at the same time in each environment?

src/components/Nav/index.tsx
```diff
@@ -1,67 +1,28 @@
...
-  Flex,
-  HStack,
-  Icon,
-  MenuButton,
-  Text,
-  useColorModeValue,
-  useDisclosure,
-  useEventListener,
-} from "@chakra-ui/react"
+import { Box, Flex, Show, useDisclosure } from "@chakra-ui/react"
```
wackerow, line 3: (Only if below suggestion implemented)

```suggestion
import { Box, Flex, Hide, Show, useDisclosure } from "@chakra-ui/react"
```

wackerow (changes requested):
Looks good @pettinarip! Just one bug at the screen breakpoint I left a suggestion for, then can pull in

I like the use of `lazy` and `Suspect`. Found [this web.dev article](https://web.dev/articles/code-splitting-suspense) a handy refresher on this for anyone curious.

Remind me, did we settle on not showing any type of loading fallback here?

src/components/Nav/index.tsx
pettinarip: Ah good catch! thanks, will test the `Hide` component and see if we are not missing or adding anything new.

src/components/Nav/index.tsx
pettinarip: Exactly. We could move it out of the suspense wrapper to make it simpler, but we will have this delayed experience, showing the search before the hamburger icon, which is what we want to avoid.

src/components/Nav/Desktop/index.tsx
pettinarip: Nice, adding. I forgot to refactor this one.

wackerow (approved):

PR #13142, opened 2024-06-11
chore: update zksync name + add ecosystem portal + update docs links

zk-Lumi:
## Description

- Updated "zkSync" -> "ZKsync" according to new branding
- Added ZKsync Ecosystem Portal Page
- Updated all broken links to docs site for DA

## Related Issue

@wackerow is it a problem that I updated docs URLs and spelling of ZKsync for non-english pages (outside of Crowdin)?

I know normally wouldn't be allowed but ->
- Links were all different to zkPorter DA across translations, and broken in many cases.
- I didn't do any translation work (only updated zkSync -> ZKsync)


zk-Lumi:
@wackerow is it a problem that I updated URLs and spelling outside of Crowdin? 

- Links were all different to zkPorter DA, and broken in many cases. 
- Also I didn't do any translation work (only updated zkSync -> ZKsync)

wackerow:
> @wackerow is it a problem that I updated docs URLs and spelling of ZKsync for non-english pages (outside of Crowdin)?

Hey @zk-Lumi, it's okay, but I'll explain...

Part of why we try to avoid changing non-English files directly through GitHub is because these won't sync to Crowdin where all of our translations are managed. This is the rough flow:

**English content changes in GitHub repo -> English content pushed to Crowdin -> Translated / approved -> Approved non-English translations pulled into GitHub (1st of each month)**

If we alter non-English content in GitHub, Crowdin doesn't know about these. On the 1st of the month, automated PRs will attempt to revert these changes when the Crowdin translations are pulled.

For this reason, typically we try to update the English file only, upload that to Crowdin where the string(s) can be corrected, approved, and then those will be imported back into the GitHub repo on their own schedule.

---

@lukassim In the meantime, what is the quickest way we could apply a name/capitalization change like this to every language on the Crowdin side?

wackerow (approved):
Pulling in, thanks @zk-Lumi

zk-Lumi:
Thanks mate, really appreciate it ❤️

PR #13156, opened 2024-06-12
DS implementation - Table of Contents

### LINKED ISSUE #10745: TableOfContents
Implement the new `TableOfContents` [styles from the DS](https://www.figma.com/file/NrNxGjBL0Yl1PrNrOT8G2B/ethereum.org-Design-System?node-id=2415%3A45636) in the [current `TableOfContents` component](https://github.com/ethereum/ethereum-org-website/tree/dev/src/components/TableOfContents).

image

pettinarip:
## Related issue
- Closes #10745

## Preview URLs
- https://deploy-preview-13156--ethereumorg.netlify.app/en/developers/docs/transactions/ (LTR)
- https://deploy-preview-13156--ethereumorg.netlify.app/fa/developers/docs/transactions/ (RTL)


wackerow:
This can likely replace #11957, as it fixes the indicator alignment issue... This PR also incorporates the new styling, so it should be preferred.

That PR got blocked when trying to polish for both LTR and RTL... this seems to also currently be a bug here: https://deploy-preview-13156--ethereumorg.netlify.app/fa/developers/docs/transactions/

image

pettinarip:
> This can likely replace #11957, as it fixes the indicator alignment issue... This PR also incorporates the new styling, so it should be preferred.
> 
> That PR got blocked when trying to polish for both LTR and RTL... this seems to also currently be a bug here: https://deploy-preview-13156--ethereumorg.netlify.app/fa/developers/docs/transactions/

Hmm is weird because on [Storybook is working fine](https://63b7ea99632763723c7f4d6b-nczeyeglju.chromatic.com/?path=/story/molecules-navigation-tableofcontents--default&globals=locale:fa). For some reason, `insetInlineStart` is being converted to `right` or `left` by Chakra or emotion, but on the site, it is not doing that...is always `left` 🤷🏼 

@TylerAPfledderer do you have any ideas on what could be going on here?

src/components/TableOfContents/TableOfContentsLink.tsx
```diff
@@ -29,16 +26,20 @@ const Link = ({
...
+      borderColor: $dotBg.reference,
       borderRadius: "50%",
       boxSize: 2,
       position: "absolute",
-      insetInlineStart: "-1.29rem",
+      // 16px is the initial list padding
+      // 8px is the padding for each nested list
+      // 4px is half of the width of the dot
+      // 1px for the border
+      insetInlineStart: `calc(-16px - 8px * ${depth} - 4px - 1px)`,
```
wackerow, line 42: @TylerAPfledderer Do you happen to know why this is creeping up again? We import `SystemStyleObject` from `@chakra-ui/react`, and use `insetInlineStart` here, but it ends up compiling to `left`:

image

Causing this:

image

TylerAPfledderer: @pettinarip @wackerow Chakra compiles `insetInlineStart` to either right or left depending on the global set `dir`.

However, here I'm not sure why it won't compile correctly. Digging into the source code, my guess it that Chakra does not like compiling a logical prop like this when it's in a pseudo. (I can't quite confirm that because the compiling flow is quite dense for me). I do see it working as expected in Storybook. Would need some time to see if there is something coming from an ancestor component affecting this.

A workaround here would be to force the original CSS prop with the kebab case.

```tsx
Before
insetInlineStart: `calc(-16px - 8px * ${depth} - 4px - 1px)`

After
"inset-inline-start": `calc(-16px - 8px * ${depth} - 4px - 1px)`
```

Now it should work in both Storybook and the dev environment.

> I'm not expecting this kind of logic for Chakra v3. Seems like overkill here, and would be able to go back to camelCase.

pettinarip: Cool, that worked fine. Pushing this change. Thanks

wackerow: Okay cool, thanks @TylerAPfledderer! If this is indeed related to pseudo-elements, we could also potentially just pull these into their own regular element (this doesn't need to be a pseudo-element)... at least in the future.. I think using `inline-inset-start` works fine for now in my opinion.

wackerow:
@pettinarip This is looking good now! Still in draft mode though... is this ready for a formal review yet? Flip to "Ready for review" and @ me when ready and will try to get this in.

wackerow:
Only thing I'm noticing now (_not a blocker imo_) is the direction of the chevron's for the accordion, RTL vs LTR:

image

image

Notice they point RIGHT no matter what... in LTR (ie English) these should point Right, but in RTL (ie. Farsi, Arabic, Hebrew, Urdu) these should point _Left_.

We can address this separately though to prevent scope creep... I suspect there may be a couple annoyances fixing that when it comes to the animation.

wackerow:
Oops, never mind! Wrong PR... my previous comments are fixed in #13063, please disregard

src/components/TableOfContents/index.tsx
```diff
@@ -80,33 +83,25 @@ const TableOfContents = ({
...
+      {!hideEditButton && editPath && (
+        <ButtonLink
+          leftIcon={<Icon as={FaGithub} />}
+          href={editPath}
+          variant="outline"
+        >
+          {t("edit-page")}
+        </ButtonLink>
+      )}
+      <Box textTransform="uppercase">{t("on-this-page")}</Box>
```
wackerow, line 95: Figma:
image

Preview:
image
image

```suggestion
      <Box textTransform="uppercase" color="body.medium">
        {t("on-this-page")}
      </Box>
```

wackerow (changes requested):
@pettinarip Looks great, just spotting one color mis-match.

wackerow (approved):

src/components/TableOfContents/TableOfContents.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+    ),
+  ],
+} satisfies Meta<typeof TableOfContents>
+
+export default meta
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  render: () => <TableOfContents slug="#web3" items={tocItems} maxDepth={2} />,
```
wackerow, line 94:
```log
9:11:34 AM: Failed to compile.
9:11:34 AM: 
9:11:34 AM: ./src/components/TableOfContents/TableOfContents.stories.tsx:94:34
9:11:34 AM: Type error: Type '{ slug: string; items: ToCItem[]; maxDepth: number; }' is not assignable to type 'IntrinsicAttributes & BoxProps & { items: ToCItem[]; maxDepth?: number | undefined; editPath?: string | undefined; hideEditButton?: boolean | undefined; isMobile?: boolean | undefined; }'.
9:11:34 AM:   Property 'slug' does not exist on type 'IntrinsicAttributes & BoxProps & { items: ToCItem[]; maxDepth?: number | undefined; editPath?: string | undefined; hideEditButton?: boolean | undefined; isMobile?: boolean | undefined; }'.
9:11:34 AM:   92 |
9:11:34 AM:   93 | export const Default: Story = {
9:11:34 AM: > 94 |   render: () => <TableOfContents slug="#web3" items={tocItems} maxDepth={2} />,
9:11:34 AM:      |                                  ^
9:11:34 AM:   95 | }
9:11:34 AM:   96 |
9:11:34 AM: error Command failed with exit code 1. (https://ntl.fyi/exit-code-1)
```

cc: @pettinarip Getting an error here... `slug` is not a prop on `TableOfContents`, do we need this here?

```tsx
// src/components/TableOfContents/index.tsx
export type TableOfContentsProps = BoxProps & {
  items: Array<ToCItem>
  maxDepth?: number
  editPath?: string
  hideEditButton?: boolean
  isMobile?: boolean
}
```

wackerow (changes requested):

wackerow (approved):

PR #13157, opened 2024-06-13
fix icons alignment in wallets page

### LINKED ISSUE #11370: Icons alignment issue on /wallets/#sim page
### Describe the bug

Icons used in some lists inside `/wallets/#sim` page are not aligned properly with text

### To reproduce

1. Go to https://ethereum.org/en/wallets/#sim
2. Scroll down to **Types of wallets** and **How to stay safe** sections
3. See icons not aligned properly

### Expected behavior

Icons aligned properly with text

### Screenshots

Screen Shot 2023-10-04 at 13 20 08

Screen Shot 2023-10-04 at 13 20 41

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes
- [ ] No

MahendraBishnoi29:
## Description
fixed icons alignment in [this](https://ethereum.org/en/wallets/#sim) page

## Related Issue
- Fixes #11370


.env.example
```diff
@@ -1,33 +0,0 @@
...
-IS_PREVIEW_DEPLOY=false
-
-# Build pages only for the specified langs. Leave it empty to build all the langs
-# e.g. `en,fr` will only build English and French pages
-# Note: always include `en` as it is the default lang of the site
-BUILD_LOCALES=
-
-# If resource constraints are being hit during builds, change LIMIT_CPUS to a
-# fixed number of CPUs (e.g. 2) to limit the demand during build time
-LIMIT_CPUS=
```
minimalsm, line 33: We shouldn't remove this file :)

minimalsm (changes requested):
Thanks for the PR @MahendraBishnoi29!

Assigning @wackerow to check this out.

src/pages/wallets/index.tsx
```diff
@@ -117,7 +117,12 @@ const ChecklistItem = (props: HorizontalCardProps) => (
   <HorizontalCard
     border={0}
     display="flex"
-    alignItems="flex-start"
+    sx={{
+      "& > span": {
+        marginTop: "6.20px !important",
```
wackerow, line 122: Thanks @MahendraBishnoi29! Appreciate the attempt at precision 😆 but don't think we want to work with sub-pixel values here... 

I was going to suggest perhaps updating HorizontalCard to use a Grid so we could center align the first row with emoji and title, but this component is used in other places where this would not be appropriately. This component is designed to primarily adjust the emoji size, for example:

image

I think this particular usage with the green check looks strange because the emoji (16px) is smaller than the text next to it (18px).

A simply solution here could be to just increase the size of the check emoji to about 1.5rem by adding `emojiSize={1.5}` to the `ChecklistItem` declaration (as a prop to `HorizontalCard`) inside `pages/wallets/index.tsx`: 

image

@nloureiro or @konopkja Thoughts on this?

MahendraBishnoi29: I've made your suggested changes @wackerow @minimalsm

wackerow (approved):
Thanks @MahendraBishnoi29! Going to bring this in as it's a minimal adjustment which addresses the issue.

cc: @nloureiro or @konopkja Feel free to comment/create an issue if you disagree with a slightly larger icon here (or if you think maybe it could be even bigger)... can always adjust further.

PR #13162, opened 2024-06-13
content: add Shisui to list of Portal clients

pauljickling:
## Description

Updated content on the doc about the Portal Network to include the client [Shisui](https://github.com/GrapeBaBa/shisui). 

## Related Issue

N/A


wackerow (approved):
Thanks @pauljickling!

Issue #12097, opened 2024-02-02
Setup texts for translation (2)


Satglow:
### Is your feature request related to a problem? Please describe.

These items are not available for translation in Crowdin, this also means that it can not be corrected if it shows a strange text.

Menu: Learn: Gas fees

![Gas fees](https://github.com/ethereum/ethereum-org-website/assets/151942349/527ba5a8-b8bd-429a-9267-fa8f7e293a8c)

https://ethereum.org/what-is-ethereum
criminal activity: the title of the linked resources are not translatable

![criminal activity - reports](https://github.com/ethereum/ethereum-org-website/assets/151942349/1387eb1e-cf7f-4563-a84f-6a78393d98af)

https://ethereum.org/hu/eth
Popular types of token: Stablecoins (this show a strange translation, but not in Crowdin)

![What is ether - stablecoin string](https://github.com/ethereum/ethereum-org-website/assets/151942349/6449d886-84dd-43ac-86d7-0f48a4b443c4)

https://ethereum.org/hu/developers
title is not translatable (DEVELOPERS)

![Developers main page - title](https://github.com/ethereum/ethereum-org-website/assets/151942349/ee9b0a52-b05d-4999-bd5d-be15a5b39a27)

https://ethereum.org/hu/developers/learning-tools
https://ethereum.org/hu/developers/local-environment
The first part of the button ’Open’ is not translatable.

![Learn by coding - button](https://github.com/ethereum/ethereum-org-website/assets/151942349/0f9bb103-1731-420f-9661-863878d80b68)

https://ethereum.org/hu/roadmap/beacon-chain
title is missing for translated page, it says: page-upgrades-beacon-date, instead of ’Shipped!’

![Beacon chain - title](https://github.com/ethereum/ethereum-org-website/assets/151942349/051494d2-4091-4647-b8de-fd2922b056a3)

https://ethereum.org/hu/staking/saas
https://ethereum.org/hu/staking/solo
https://ethereum.org/hu/staking/pools

Same type of issues on all the 3 pages:
- Why stake solo? /  Why stake with a service? / Why stake with a pool? - there are buckets which text is not available for translation
- Comparison with other options – this title is not available for translation; under this subtitles also stay in English: Staking as a service (SaaS), Pooled staking
- Explore node and client setup tools / Explore staking pools/ Explore staking service providers – there is a text like ’from 10,4 ETH’ or ’any amount’ – this could be translatable too

[...]

PR #13167, opened 2024-06-14
Fix: Developer title string

wackerow:
## Description
- Adds `common:` namespace to properly fetch `developer` string value

## Related Issue
- #12097


wackerow:
image

instead of 

image

corwintines (approved):

Issue #13199, opened 2024-06-19
Update: Edit/proof of node clients: 


m4sterbunny:
docs(proof) needed to fix consistency of abrevs, reduce verboseness, improve clarity

I will take this one

wackerow:
@m4sterbunny Appreciate the posts.. Housekeeping feedback, I would request in the future you use [issue templates](https://github.com/ethereum/ethereum-org-website/issues/new/choose) to help guide the creation of an issue with more details, before posting a PR. Mixed opinions of #13200 and it can help to discuss before hand.

wackerow:
Left a note on the PR, but didn't agree with those changes proposed. Will leave this open though, would encourage you to provide more details or argue your case more and we can figure out a game plan.

m4sterbunny:
The validator client, while typically bundled with the Consensus client, does not need to be run >> this is an active decision that also requires stake >> IMO it helps to clarify this early as the term Full node could otherwise be misleading (Full implies you run everything from a low-knowledge persepective).

Unidirectional arrow is to clear up the implication that EL can "drive" CL. CL makes calls to EL via the Engine API not vv. Yes, the EL responds to these requests >> This was from Besu team when I was attempting to understand further, so I am trusting my source. Alt would be dotted arrow back from EL.

WRT issue templates -- I did shop, found nothing that matched.

PR #13111, opened 2024-06-05
docs(proof): fixes consistency of abrevs, reduces verboseness, 

m4sterbunny:
fixes consistency of abrevs, reduces verboseness, clarifies 3 clients

## Description

Proofing rounds fixes p2p mixed with P2P
Applies more common variant Proof of Stake
Reduces verboseness
Clarifies 3rd client required for block building in intro

DOES NOT fix line 25: It seems unlikely to me that "Consensus clients include the execution payload in every block", do they need validator client to assist them to do this? I don't have the domain knowledge to add value here, but the position of this information is under "What does the execution client do?", therefore at minimum, relocation probably advised.

PR #13200, opened 2024-06-19
13199(docs):proofs & edit

m4sterbunny:
fixes  for https://github.com/ethereum/ethereum-org-website/issues/13199
consistency of abrevs, reduces verboseness, clarifies 3 clients

Description
Proofing rounds fixes p2p mixed with P2P
Applies more common variant Proof of Stake
Reduces verboseness
Clarifies 3rd client required for block building in intro

DOES NOT fix line 25: It seems unlikely to me that "Consensus clients include the execution payload in every block", do they need validator client to assist them to do this? I don't have the domain knowledge to add value here, but the position of this information is under "What does the execution client do?", therefore at minimum, relocation probably advised.

replaces #13111 (my bad fork hygiene)


wackerow:
Hey @m4sterbunny!

The "Validator client" is part of the consensus client and not typically considered a third client. Personally not sure about most of these changes, and some of them ("Proof of Work") go directly against [our established style guide](https://ethereum.org/en/contributing/style-guide/content-standardization/#proof-of). 

Curious why we made the arrow in the image unidirectional? This should be a bidirectional arrow to show communication between them. Appears that this is the only change to that image so I would revert. ([Engine JSON-RPC API](https://github.com/ethereum/execution-apis/tree/main/src/engine), [Engine API: A Visual Guide](https://hackmd.io/@danielrachi/engine_api))

The `P2P` changes I'd support, but would open as an independent change; not sure about the rest. Going to close this out, but can discuss more in the issue #13199 if you want to elaborate on these changes.

m4sterbunny:
The validator client, while typically bundled with the Consensus client, does not need to be run >> this is an active decision that also requires stake >> IMO it helps to clarify this early as the term Full node could otherwise be misleading (Full implies you run everything from a low-knowledge persepective).

Unidirectional arrow is to clear up the implication that EL can "drive" CL. CL makes calls to EL via the Engine API not vv. Yes, the EL responds to these requests >> This was from Besu team when I was attempting to understand further, so I am trusting my source. Alt would be dotted arrow back from EL.

WRT issue templates -- I did shop, found nothing that matched.

PR #13207, opened 2024-06-19
fix: wallet filter device boolean tracking

wackerow:
## Description
- Fixes reversal of true/false for "Toggle Desktop/Browser/Mobile" Matomo events
- Adds the word "to" to differentiate between old logs (where T/F have flipped meanings) and the new corrected entries.

Example of new usage: If "Mobile" is switched from off to on, it will log `Toggle Mobile to true`.

## Related Issue
None filed; from UX research sync

### DIFF
--- a/src/components/FindWallet/WalletFilterFeature.tsx
+++ b/src/components/FindWallet/WalletFilterFeature.tsx
@@ -176,7 +176,7 @@ const WalletFilterFeature = ({
                                     eventAction: `${filterOption.title}`,
                                     eventName: `Toggle ${
                                       item.title
-                                    } ${!item.showOptions}`,
+                                    } to ${item.showOptions}`,
                                   })
                                 }
                           }


src/components/FindWallet/WalletFilterFeature.tsx
```diff
@@ -176,7 +176,7 @@ const WalletFilterFeature = ({
                                     eventAction: `${filterOption.title}`,
                                     eventName: `Toggle ${
                                       item.title
-                                    } ${!item.showOptions}`,
+                                    } to ${item.showOptions}`,
```
wackerow, line 179: image

Tested all the different options we have... this event is only fired for the `Browser` `Desktop` or `Mobile` switches shown here

corwintines (approved):
LGTM. @konopkja can you look and sign off on this too?

konopkja:
lgtm

nhsz (approved):

Issue #13209, opened 2024-06-19
Unaccessible link on the "Intro to Ethereum" page


AlexanderLukin:
### Describe the bug

The link [How does Ethereum work, anyway?](https://www.preethikasireddy.com/post/how-does-ethereum-work-anyway) in the [FURTHER READING](https://ethereum.org/en/developers/docs/intro-to-ethereum/#further-reading) section of the "Intro to Ethereum" page is no longer accessible. It should probably be removed from this section or replaced with its up-to-date version.

### To reproduce

Go to this page
https://ethereum.org/en/developers/docs/intro-to-ethereum/#further-reading
The second link in the section is no longer accessible.

### Expected behavior

All links on the site should be correct.

### Would you like to work on this issue?

- [X] Yes
- [ ] No

wackerow:
Thanks for surfacing @AlexanderLukin! PR welcome to remove this link.

AlexanderLukin:
Great! Here it is https://github.com/ethereum/ethereum-org-website/pull/13219

I decided to update the link instead of removing it, because the article is excellent.

AlexanderLukin:
Looks like the domain is fixed, so, we may not change the original link. I'm closing the issue together with my PR, as the existing link is working fine.

PR #13219, opened 2024-06-21
fix: update incorrect link

AlexanderLukin:
The old link "How does Ethereum work, anyway?" in the [FURTHER READING](https://ethereum.org/en/developers/docs/intro-to-ethereum/#further-reading) section of the "Intro to Ethereum" page is no longer accessible, because the domain "preethikasireddy.com" expired a long time.

Now this link is replaced with the correct link to this article hosted on medium.com.

This PR has been discussed in this issue https://github.com/ethereum/ethereum-org-website/issues/13209


AlexanderLukin:
Looks like the domain is fixed, so, we may not change the original link. I'm closing my PR, as the existing link is working fine.

PR #13223, opened 2024-06-21
remove reference to unmaintained eEVM

Otto-AA:
## Description

[eEVM](https://github.com/microsoft/eEVM) is archived and hasn't been updated in a year. I did not find any link to a new code location, so I think this project is not maintained in any more.

This PR removes the link from this page: https://ethereum.org/en/developers/docs/evm/


nhsz (changes requested):
Thanks @Otto-AA. We would also need to remove the link from the translations (check the `/public/content/translations/` path or just look or the url)

Otto-AA:
Thanks for the remark, forgot about this.

I've removed eEVM from the translations now and also added revm to a few translations where it was missing.

nhsz (approved):
LGTM, thanks!

PR #13224, opened 2024-06-21
Upgrade TS to v5.5

pettinarip:
## Description

Upgrades TS.

### DIFF
--- a/src/components/Codeblock.tsx
+++ b/src/components/Codeblock.tsx
@@ -236,8 +236,8 @@ const Codeblock = ({
     className = codeLanguage || ""
   }
 
-  const matches = className?.match(/language-(?<lang>.*)/)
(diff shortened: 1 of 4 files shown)


src/components/Codeblock.tsx
```diff
@@ -236,8 +236,8 @@ const Codeblock = ({
     className = codeLanguage || ""
   }
 
-  const matches = className?.match(/language-(?<lang>.*)/)
```
pettinarip, line 239: Now ts can check regexs! got an error here: `Named capturing groups are only available when targeting 'ES2018' or later`

wackerow: Nice!

wackerow (approved):
Lgtm!

Issue #11591, opened 2023-11-06
Bug: Click zone on color toggle rotating with icon effect


wackerow:
### Describe the bug

Recently an animation was added to the color-mode icon in the nav bar, to rotate slightly on hover. Noticing that we're rotating the whole container, including the click zone, instead of just the icon.

We should also make sure to apply this same effect for `focus` state, not only `hover`

### To reproduce

1. Go to https://ethereum.org
2. Use keyboard to tab over to the moon/sun color-mode toggle icon
3. Hover over this element and notice that the whole click box rotate

### Expected behavior

Would expect the icon _only_ to rotate, not the containing box. Also, would expect same effect on keyboard focus. 

### Screenshots

<img width="434" alt="image" src="https://github.com/ethereum/ethereum-org-website/assets/54227730/3158a83f-cc58-475d-b5e9-aef1037713f2">

### Would you like to work on this issue?

- [ ] Yes
- [ ] No

ATREAY:
Hey @wackerow, can I work on this issue if it's available?

wackerow:
@ATREAY Hey, thanks for offering! We had a volunteer for this task via Discord already... let me check on that before assigning you

ATREAY:
Sure, @wackerow, if no one takes this issue, let me know. I'll look for another one in the meantime.

rohan9024:
> @ATREAY Hey, thanks for offering! We had a volunteer for this task via Discord already... let me check on that before assigning you

Hey @wackerow Assign me!

wackerow:
Thanks @rohan9024, just assigned you, let us know if you hit issues... apologies @ATREAY, can try to get you on another one!

rohan9024:
Thanks @wackerow

tomasgrusz:
Seems the issue is stale @wackerow . I'm happy to look into it if it's ready to be picked up again.

wackerow:
Thanks @tomasgrusz! Went ahead and assigned

corwintines:
Closing with the above PR

PR #13229, opened 2024-06-24
style: shift hover transition to button children icon

tomasgrusz:
> [!note]
> TODO: add `:focus` state

- remove hover animation from `IconButton`
- add hover animation to `IconButton`'s child svg (the icon)

## Related Issue
[Bug: Click zone on color toggle rotating with icon effect](https://github.com/ethereum/ethereum-org-website/issues/11591)


src/components/Nav/index.tsx
wackerow, line 1: This looks good @tomasgrusz! I would just assign these `_hover` styles to a constant above, so you can reuse them for both `_hover` and `_focus`

ie:

```tsx
const desktopHoverFocusStyles = {
    '& > svg': {
      transform: "rotate(10deg)",
      color: "primary.hover",
      transition: "transform 0.5s, color 0.2s"
    }
  }
```

```tsx
_hover={desktopHoverFocusStyles}
_focus={desktopHoverFocusStyles}
```

Then that should do it!

tomasgrusz: Perfect @wackerow thanks for the help. It's been on my todo list as I had other responsibilities, but here it goes. 👍🏻

PR #13237, opened 2024-06-25
style(theme/Tag): set background style on close button hover

### LINKED ISSUE #10266: [Tag] Update close button styles

TylerAPfledderer:
## Description

Updates the close button style on hover to sync with the DS.

This is not a complete sync — see the "Normal" and "Tag" Solid variants. This provided additional complication that already exists in the theming.

## Related Issue

- Closes #10266


TylerAPfledderer:
@pettinarip and @nloureiro, I would like to try and review the tag styling with you, and present an alternative that I have made (Pablo has seen this).

What makes creating the styling in the code with all the variants is inconsistency in how all the different colors are being used. The generates a bloated set of styles that have to be hard-coded. I would like to present a solution that reduces the colors used to just the two different shades in the palette for each "status" type which would also improve a11y in the color contrast, especially with this hover styles of the close button.

yarn.lock
pettinarip, line 1: we shouldn't have this file modified

pettinarip (changes requested):
@TylerAPfledderer nice, lgtm. Once we revert the `yarn.lock` we could merge this.


=====
Not related to this PR but noted a bug on the Tags as links
image

When hovered, it is displaying the text with the Link styles. Not a big deal because we are not currently using this case on the site.

pettinarip:
> @pettinarip and @nloureiro, I would like to try and review the tag styling with you, and present an alternative that I have made (Pablo has seen this).
> 
> What makes creating the styling in the code with all the variants is inconsistency in how all the different colors are being used. The generates a bloated set of styles that have to be hard-coded. I would like to present a solution that reduces the colors used to just the two different shades in the palette for each "status" type which would also improve a11y in the color contrast, especially with this hover styles of the close button.

Sounds good to me!

FYI new colors are coming to the DS. Just saying for context but I don't think it will affect what you're suggesting here.

yarn.lock
TylerAPfledderer: @pettinarip The file modified when I brought in the latest changes from `dev` and ran `yarn install`

TylerAPfledderer: @pettinarip merging the latest `dev` still provided a diff for the lockfile

pettinarip (approved):
Fixed. The difference was added in your first commit, I added a revert for that file.

The rest looks good, approving 👍🏼

PR #13263, opened 2024-06-27
Fix hydration issue on mobile

### LINKED ISSUE #12749: Error: Text content does not match server-rendered HTML when developing on Windows
### Describe the bug

I follow the instruction to run the project locally  on my Windows 11 PC (node v20.12.2)
But the following bug appear each time I visit any page and the label strings is not translated.
```
Unhandled Runtime Error
Error: Text content does not match server-rendered HTML.

See more info here: https://nextjs.org/docs/messages/react-hydration-error

Call Stack
checkForUnmatchedText
node_modules\react-dom\cjs\react-dom.development.js (9647:0)
diffHydratedProperties
node_modules\react-dom\cjs\react-dom.development.js (10310:0)
hydrateInstance
node_modules\react-dom\cjs\react-dom.development.js (11306:0)
prepareToHydrateHostInstance
node_modules\react-dom\cjs\react-dom.development.js (12564:0)
completeWork
node_modules\react-dom\cjs\react-dom.development.js (22181:0)
...
````

### To reproduce

1. `yarn`
2. `yarn dev`

### Expected behavior

The page should render correctly

### Screenshots

image

- Windows 11 PC (node v20.12.2)

_No response_

### Additional context

_No response_

- [X] Yes
- [ ] No

### PR DESCRIPTION
**Pull request by pettinarip**
## Description

The desktop menu was still rendered on ssr because `isMobile` is always `false` on the server. Changed the logic to only render the menu on the client side as originally intended.

## Related issue

Fixes #12749

### DIFF
--- a/src/components/Nav/index.tsx
+++ b/src/components/Nav/index.tsx
@@ -6,7 +6,7 @@ import { EthHomeIcon } from "@/components/icons"
 import { BaseLink } from "@/components/Link"
 import Search from "@/components/Search"
 
-import { isMobile } from "@/lib/utils/isMobile"
+import { isDesktop } from "@/lib/utils/isDesktop"
 
 import { NAV_PY } from "@/lib/constants"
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/components/Nav/index.tsx [hunk 1]
```diff
@@ -60,7 +63,12 @@ const Nav = () => {
             ms={{ base: 3, xl: 8 }}
           >
             {/* avoid rendering desktop Menu version on mobile */}
-            {isDesktop && <Menu hideBelow="md" sections={linkSections} />}
+
+            {isClient && isDesktopFlag ? (
```
wackerow, line 67: I think just adding `isClient` might have done the trick, since it would be `false && _` and the rest wouldn't even be evaluated. Only when on the client.

Then again, I think it's probably best to have an appropriate `isDesktop` that checks for `window` itself as you've added

wackerow (approved):
Nice, looks good! Thanks @pettinarip

Issue #10269, opened 2023-05-19
Fix typo on JavaScript API Libraries page


johnnygee19:
There's a hyphen missing. Spelling should be consistent across the entire the web page which is "JSON-RPC". 

![eth1](https://github.com/ethereum/ethereum-org-website/assets/132227312/d942b21c-ee0b-4b81-8b76-1d2b37e36dfe)

https://ethereum.org/en/developers/docs/apis/javascript/

I can work on this.

Re1da:
#9608

wackerow:
<img width="519" alt="image" src="https://github.com/ethereum/ethereum-org-website/assets/54227730/2092ff1f-905a-4c4e-9100-a7abec05b461">

Looks good! Agree, we should standardize around "JSON-RPC"... I see 11 instances of "JSON RPC" in 7 English files when doing a quick search... open for someone to patch these up!

PR #13269, opened 2024-06-28
JSON RPC -> JSON-RPC

corwintines:
## Description

- JSON RPC -> JSON-RPC

## Related Issue

- Closes: https://github.com/ethereum/ethereum-org-website/issues/10269


wackerow (approved):
Hey @corwintines, per our discussion I went ahead and reverted the non-English updates here, as this is a low-priority adjustment, and it's in a string segment that would not automatically propagate through the translations on Crowdin side (ie unlike a link `href`).

Pulling in the English file updates; will let the Crowdin flow handle the rest.

Issue #13247, opened 2024-06-26
An answer and its explanation are contradictory


iridiumcao:
### Page URL

https://ethereum.org/en/quizzes/

### Question content

Using Ethereum -> 3. Layer 2
![image](https://github.com/ethereum/ethereum-org-website/assets/754000/3bd78bb3-e325-4df1-93b9-960ee95bdeb9)

Question:
![image](https://github.com/ethereum/ethereum-org-website/assets/754000/d4a151c6-75f0-4442-bd4e-0d9f19703787)

The answer and its explanation are contradictory:
![image](https://github.com/ethereum/ethereum-org-website/assets/754000/be226f4b-60af-49a6-8ceb-f60ecf5aa411)

### Would you like to work on this issue?

- [ ] Yes
- [X] No

iridiumcao:
The pics I updated are not as expected, I'm going to edit...

wackerow:
Hey @iridiumcao, thanks for reporting... Just noting that this is the incorrect answer because it's "All of the above", so perhaps we can add a simple note (as we do in other similar questions) that "This is not the only correct answer."

**Will open this up for a community contributor with the scope of adding `"This is not the only correct answer."` to the explanation for answer choices A, B an C here.**

raspberri05:
Hi I can work on this

raspberri05:
I've opened a pull request for this: https://github.com/ethereum/ethereum-org-website/pull/13272
Please let me know if I need to make any changes

wackerow:
Great! I've assigned you @raspberri05.. commented on your PR, any reason this got closed?

raspberri05:
Sorry, was a mistake. Ive reopened it here https://github.com/ethereum/ethereum-org-website/pull/13380

corwintines:
Closing with above PR

PR #13272, opened 2024-06-30
Update learn-quizzes.json to fix vague explanation (issue #13247)

raspberri05:
## Description

I added "This is not the only correct answer." to the explanation as to why the other answer choices are wrong and "all of the above" is the correct answer.
## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/13247


wackerow:
Hey @raspberri05! Any reason we closed this out? Looks good to me for the issue it's solving.

raspberri05:
Sorry, was a mistake. Ive reopened it here #13380

PR #13298, opened 2024-07-01
Performance: lazy load stats using intersection obs

pettinarip:


src/pages/index.tsx
```diff
@@ -225,27 +248,6 @@ export const getStaticProps = (async ({ locale }) => {
   }
 }) satisfies GetStaticProps<Props>
 
-const CodeblockSkeleton = () => (
```
pettinarip, line 228: I cleaned up this code a bit. Moved these 2 components to the top and renamed the `CodeblockSkeleton` to just `Skeleton` (to be reused in the Stats section).

src/components/LazyLoadComponent.tsx
wackerow, line 1: ❤️

wackerow (commented):
Looking good @pettinarip! Just flagging the .env changes which I assume we need to revert, else I think this is good to go!

corwintines (approved):

PR #13303, opened 2024-07-02
Adds bottom padding to code block [Fixes #13299]

### LINKED ISSUE #13299: Inappropriate padding in code blocks
### Describe the bug

The `Codeblock` component is used to render various types of code snippets from the triple-backtick markdown syntax (\`\`\`). These blocks currently do not have appropriate padding applied to them.

### To reproduce

1. Go to https://ethereum.org/en/developers/tutorials/all-you-can-cache (or any page with code snippets)
2. Scroll down to some code snippets
3. Notice the lack of padding on the bottom of the code blocks

### Expected behavior

Codeblocks should have 1.5rem padding-bottom applied to improve readability

### Screenshots

image

_No response_

_No response_

### Additional context

Note that some codeblocks have a "copy" button added to it which alters the top padding; this should remain unchanged. The codeblocks _without_ the copy button use `24px` (`1.5rem`, **size token `6`**) for the top padding; I would recommend we use the same for the bottom.

Similarly, some code blocks are longer than others... there is a max-height applied initially, with a "Show all" button for longer blocks. In those cases, we do _not_ need padding if there is more code to scroll to, but once we're at the bottom there _should_ be padding.

- [ ] Yes
- [X] No

sandeepV2:
## Description

The Codeblock component is used to render various types of code snippets from the triple-backtick markdown syntax (```). This fix adds a bottom padding of 1.5rem to code block.
## Before fix:
Screenshot 2024-07-02 at 12 31 47 PM

## After fix:
codeblock-fix

## Related Issue

- Fixes #13299


sandeepV2:
I see that change is replicating as expected in Deployment preview 
https://deploy-preview-13303--ethereumorg.netlify.app/en/developers/tutorials/all-you-can-cache/

src/components/Codeblock.tsx
```diff
@@ -276,6 +276,7 @@ const Codeblock = ({
               style={style}
               className={className}
               pt={hasTopBar ? "2.75rem" : 6}
+              pb={"1.5rem"}
```
wackerow, line 279: @sandeepV2 With our design system, we use spacing tokens where `1` = `0.25rem` = `4px`... let's use the `6` token here instead:

```suggestion
              pb={6}
```

wackerow (changes requested):
Looks great @sandeepV2! Thanks for posting... one small adjustment request and we can pull this in.

src/components/Codeblock.tsx
sandeepV2: @wackerow Yes i see that 6 token in the padding-top styling too.
`pt={hasTopBar ? "2.75rem" : 6}`

I tried to find where we define the spacing tokens where `1 = 0.25rem =4px` , is it in spacing.ts ?

[https://github.com/ethereum/ethereum-org-website/blob/dev/src/%40chakra-ui/foundations/spacing.ts#L2](https://github.com/ethereum/ethereum-org-website/blob/dev/src/%40chakra-ui/foundations/spacing.ts#L2)

> const spacing = {
  7.5: "1.875rem",
  10.5: "2.625rem",
  19: "4.75rem", // Nav height
}

is it hiding in here (`1.875rem / 7.5 = 0.25rem `) ?

wackerow (approved):
Thanks @sandeepV2! Looks good!

wackerow:
@all-contributors please add @sandeepV2 for bug fix

PR #13310, opened 2024-07-03
refactor: update "mining" verbiage to PoS terms

wackerow:
## Description
- Updates use of "mining" terms throughout repo to be consistent with PoS Ethereum

## Related Issue
- None filed


public/content/developers/docs/apis/json-rpc/index.md
```diff
@@ -1722,8 +1722,7 @@ curl --data '{"jsonrpc":"2.0","method": "eth_getTransactionReceipt", "params": [
...
 ```
 
-Our contract was created on `0x4d03d617d700cf81935d7f797f4e2ae719648262`. A null result instead of a receipt means the transaction has
-not been included in a block yet. Wait for a moment and check if your miner is running and retry it.
+Our contract was created on `0x4d03d617d700cf81935d7f797f4e2ae719648262`. A null result instead of a receipt means the transaction has not been included in a block yet. Wait for a moment and check if your consensus client is running and retry it.
```
wackerow, line 1725: (This one I'm not 100% sure on, but pretty sure it works)

PR #13322, opened 2024-07-04
chore: update vitalik blog link

hattizai:
## Description
use `.eth.limo` to link to blog.


corwintines:
Thanks @hattizai 

We have been issues with regressions in our translated files when making these changes, so I am going to close this out. If you notice this issue in the english file, I would suggest changing it there and it will fix through our translation process.

PR #13326, opened 2024-07-04
readme: add tip about how to build in other languages

nhsz:
As local builds are set to include `en` only now by default, this PR adds the instruction on how to include more locales

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ yarn dev
 - Pro Tip:
   - Explore scripts within `package.json` for more build options
   - Get **faster** production builds by building only one language. E.g. in your `.env` file, set `BUILD_LOCALES=en` to build the content only in English
+  - To build the site in other languages too, you need to set them in `BUILD_LOCALES`, eg: `` BUILD_LOCALES=`en,es` `` if you also want to include Spanish
 
 By default the script will build all the languages (complete list in `i18n.config.json`).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -96,6 +96,7 @@ yarn dev
 - Pro Tip:
   - Explore scripts within `package.json` for more build options
   - Get **faster** production builds by building only one language. E.g. in your `.env` file, set `BUILD_LOCALES=en` to build the content only in English
+  - To build the site in other languages too, you need to set them in `BUILD_LOCALES`, eg: `` BUILD_LOCALES=`en,es` `` if you also want to include Spanish
```
wackerow, line 99:
```suggestion
  - To build the site in other selected languages too, you need to set them in `BUILD_LOCALES`, eg: `BUILD_LOCALES=en,es` if you also want to build only English (required) and Spanish.
  - To build all languages, simply comment this line out with a hash mark, eg: `# BUILD_LOCALES=`
```

nhsz: @wackerow `BUILD_LOCALES=en,es` is not the same, backticks needs to be around the locales

wackerow: Oh interesting... so personally I have never used the back-ticks around this list of locales (and it still works just fine). These are not required as everything after the `=` will be treated as a string anyway. For simplicity here I'd recommend we just leave them off.

wackerow (approved):
❤️❤️

PR #13331, opened 2024-07-05
Fixed broken active link in developer docs navbar [Fixes #10813]

### LINKED ISSUE #10813: Bug report: Fix marker in the developer docs sidebar of the current active page

I am not sure 100% if this was a thing in the past but has been removed later from the website design.

The link of the active page is marked with a different color in the sidebar than other links like when it's being hovered for example, this would give users a better experience for end-users when scrolling through the docs.

expected behavior, something like this:
image

ignore it, users can find out where they are when they look closely.

### Additional context

.

- [ ] Yes
- [ ] No

0x0OZ:
## Description
There was an issue checking `isHrefActive` in `Link.tsx` because of the unremoved hashtags from the URL.

I don't know if that's the best way to deal with it or if this fix breaks assumptions on other components that assume hashtags remain in that check.

I am not really a good developer, so it will be great if someone double checks (patch [source](https://github.com/vercel/next.js/issues/25202#issuecomment-966658956))

## Related Issue
Fixes #10813

### DIFF
--- a/src/components/Link.tsx
+++ b/src/components/Link.tsx
@@ -58,11 +58,12 @@ export const BaseLink = forwardRef(function Link(
   ref
 ) {
   const { asPath } = useRouter()
+  const cleanAsPath = asPath.replace(/#.*/, "");
   const { flipForRtl } = useRtlFlip()
 
   let href = (to ?? hrefProp) as string
 
-  const isActive = url.isHrefActive(href, asPath, isPartiallyActive)
+  const isActive = url.isHrefActive(href, cleanAsPath, isPartiallyActive)
   const isDiscordInvite = url.isDiscordInvite(href)
   const isPdf = url.isPdf(href)
   const isExternal = url.isExternal(href)


wackerow (changes requested):
Thanks @0x0OZ! Appreciate you dropping this.

Taking a look here, I think we should probably patch the `isHrefActive` function instead, to do this same logic.

My thought here is, that function should answer "is this href active?", which should always be ignoring the hash links (anything after a `#`) or any query/search params (anything after a `?`). We shouldn't have to pre-sanitize our pathname when passing to this function in my opinion.

> const cleanAsPath = asPath.replace(/#.*/, "");

If you look inside `src/lib/utils/url.ts` (where you're importing `isHrefActive`, you'll see another helper function called `cleanPath`:

```ts
// remove any query params or hashes from the path
export const cleanPath = (path: string): string => path.replace(/[$#].+$/, "")
```

This is very similar to the logic you added, but will also remove any query/search params as well. Let's use this instead.

I would recommend the following:

0. Revert these changes to `Link.tsx`
1. Update the `isHrefActive` function inside `url.ts`: Instead of `pathname === href`, let's wrap `pathname` with `cleanPath`, ie: `cleanPath(pathname) === href`
2. We're using `cleanPath` inside `isHrefActive` so let's move the `cleanPath` function above the other so we're not using functions before they're declared. This isn't strictly necessary due to JS hoisting but is good practice. 
3. Make sure we run the linter on this to keep it tidy: `yarn prettier --write src/lib/utils/url.ts`
4. Run `yarn dev` and test it out in the browser to make sure this is working as expected

Hope that helps!

0x0OZ:
I think this looks good

wackerow (approved):
🎉 Beautiful! Sorry for delay, looks great @0x0OZ! Bringing in

PR #13374, opened 2024-07-10
Feat: redesign events page ✨

### LINKED ISSUE #13367: Events page redesign

The events page (https://ethereum.org/en/community/events/) has grown and the current design does not support showing many events at the same time. This hinders usability of the page.

The team has proposed following design.

light mode

figma designs 
https://www.figma.com/design/BEGvi1RiaiFIK4a7MeCl2A/events-page-redesign?node-id=36-67140&t=P9nvjQJW7xAkHsjK-1

NA

### Additional context

_No response_

- [ ] Yes
- [X] No

saurabhburade:
## Description
- Implementation of #13367 

- Redesign `EventCard`
- Redesign `UpcomingEventsList`
- Redesign `MeetupList`

## Preview URL
https://deploy-preview-13374--ethereumorg.netlify.app/en/community/events

## Related Issue
Fixes #13367


saurabhburade:
Hi @konopkja, please check the preview. Also please let me know if any modifications required

konopkja:
Wow! looking good, there might be few minor adjustments with the interaction but overall this is shaping up pretty well! Thank you.

1. could you please double the padding for Month headline? (right now month headline is too close to the cards from previous month)

2. Can we load by default 2 months instead of one? 

3. The button should also load 2 months more instead of one to prevent excessive need to click on the button

saurabhburade:
Sure

konopkja:
> Thank you! surely, making the required modifications.

Lgtm, thank you!!

src/lib/utils/date.ts
```diff
@@ -1,2 +1,55 @@
...
+  "Apr",
+  "May",
+  "Jun",
+  "Jul",
+  "Aug",
+  "Sep",
+  "Oct",
+  "Nov",
+  "Dec",
+]
```
wackerow, line 18: I appreciate the work put in here @saurabhburade... I'm cautious about taking this approach though because it does not translate. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

We should use the native `Intl.DateTimeFormat` object in JavaScript to generate the correct display names depending on the `locale`. Does that make sense, are you familiar with this object?

wackerow (changes requested):
Hey @saurabhburade! Thanks for the work here, appreciate this. It's overall looking great! I left a note about concerns for non-English pages with the proposed logic for parsing the display date range.

Let me know if what I mentioned makes sense, but ideally these can be displayed properly according to the users locale, and not be forced to just English.

src/components/UpcomingEventsList.tsx
```diff
@@ -100,56 +109,88 @@ const UpcomingEventsList = () => {
...
+                      formattedDetails,
+                      date,
+                      location,
+                      imageUrl,
+                      startDate,
+                      endDate,
+                    },
+                    idx
+                  ) => {
+                    return (
```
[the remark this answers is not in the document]
saurabhburade: Yeah, it makes sense 👍

konopkja:
a small UI change requests after team review:

1. remove year from each event date (month and day should be enough)
2. increase the padding around the button "load more" button as it seems a bit hidden now

saurabhburade:
From the latest comment of  @konopkja

@wackerow, does it make sense to use `locale` Date if we remove the year from events?


Screenshot 2024-07-24 at 4 02 38 PM

saurabhburade:
@konopkja  @wackerow  Thank you for your feedback, here's the modified implementation for dates
Screenshot 2024-07-24 at 5 24 57 PM

konopkja:
> @konopkja @wackerow Thank you for your feedback, here's the modified implementation for dates Screenshot 2024-07-24 at 5 24 57 PM

why is there the "ago" part?

wackerow:
@konopkja I'm assuming this is showing Italian ("August" is "agosto", shortened to "ago")

saurabhburade:
Its locale date-time for `es-ES`

src/components/EventCard.tsx
```diff
@@ -16,87 +19,107 @@ export type EventCardProps = {
...
+            <Box>
+              <Text fontSize={{ base: "sm", md: "sm" }} mb={0} noOfLines={4}>
+                {description}
+              </Text>
+            </Box>
+          </Box>
+        </Box>
+        <Box padding={4} paddingTop={0} width={"100%"}>
+          <ButtonLink href={to} width={"100%"} variant="outline">
+            View Event
```
wackerow, line 117: Noting as a separate task we should extract this string for translation.

wackerow (commented):
Looking good @saurabhburade! Mind clearing the merge conflicts an we'll give this a final review?

src/components/EventCard.tsx
saurabhburade: Should I add it to `en` translations?

src/lib/utils/date.ts
```diff
@@ -1,2 +1,44 @@
 export const dateToString = (published: Date | string) =>
   new Date(published).toISOString().split("T")[0]
+
+export const formatDateRange = (startDate: string, endDate: string) => {
```
wackerow, line 4: Appreciate the work here... few things on this function.

I think we can simply use the `Intl.DateTimeFormat().formatRange` helper built into JavaScript, with the right options. It produces very similar strings to this function, but avoids the custom logic, and will better align with internationalization date-display standards, and offload any need to maintain this.

For example: `16-18 Aug` but also `Aug 23-Sep 13` (custom `formatDateRange`)
would become: `Aug 15 – 17` and `Aug 22 – Sep 12` (native `Intl.DateTimeFormat().formatRange`, with 2-digit day, numeric month, no year or time)

cc: @konopkja I think this may be _slightly_ different than your expressed preference, but I would ask to reconsider to avoid getting into customized ways to handle dates, when dealing with a very internationalized site. 

---

Also, in this custom `formatDateRange` we don't appear to be using `locale` (I see `undefined`) so it's defaulting to english.

In the end we should use `startDate`, `endDate` and `locale` with:

```ts
new Intl.DateTimeFormat(locale, { day: "2-digit", month: "short" }).formatRange(new Date(startDate), new Date(endDate))
```

...to get the result we're going for here.

```js
// Paste me into a node terminal / browser console to try it
var startDate = "2024-08-23"
var endDate = "2024-09-13"
var locale = "es-ES"
var shortDayMonth = new Intl.DateTimeFormat(locale, { day: "2-digit", month: "short" }).formatRange(new Date(startDate), new Date(endDate))
console.log(shortDayMonth) // 22 ago – 12 sept
```

wackerow (changes requested):
Hey @saurabhburade, sorry to block again. Noted a couple things:

1. Clamping lines leads to unreachable content
2. Can use `Intl.DateTimeFormat(locale,options).formatRange(from,to)` to get date range string in a simple, easy to maintain, an intl friendly manner
3. `to` is now deprecated; use `href`

I kinda see 2 and 3 as blocking, curious your take @konopkja

src/lib/utils/date.ts
saurabhburade: This makes complete sense, making the required modifications in date formats

src/components/EventCard.tsx
```diff
@@ -14,89 +18,110 @@ const clearStyles = {
 
 export type EventCardProps = {
   title: string
-  href: string
```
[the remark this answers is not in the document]
saurabhburade: One thing I observed was some events have `to` or `href`

saurabhburade:
@wackerow @konopkja 

Clamping lines are to evenly show the event cards, if not it may lead to uneven heights. As the length of the description of an event is unpredictable

wackerow:
@saurabhburade Gotcha! In that case I'd consider a css `grid` which would allow you to have cards in a row all have the same height, but have that height be flexible depending on the longest card text for that row

nloureiro (approved):
All good from the design side!!!
Great work!

nloureiro:
@saurabhburade, great work, thank you.


one last thing, I got this error on Safari.
Screen Shot 2024-08-05 10 56 47 AM

saurabhburade:
@nloureiro, thank you. The safari issue is fixed now.

saurabhburade:
@wackerow Please do the final review, LGTM for approval.

nloureiro (approved):

wackerow (approved):

PR #13375, opened 2024-07-10
Update enterprise content

tasdienes:
Removed most of the material about private chains, which have largely become obsolete and irrelevant


corwintines (approved):
Thanks @tasdienes!

Issue #13341, opened 2024-07-06
Update Aave logo


omahs:
### Describe the bug

Update Aave logo

### To reproduce

The Aave logo and branding have recently been updated, and our files should be revised in a few places, including this [page](https://ethereum.org/en/stablecoins/).

The new branding materials are available [here](https://github.com/aave-dao/aave-brand-kit/tree/main).

<img width="1455" alt="aave-logo" src="https://github.com/ethereum/ethereum-org-website/assets/73983677/93786a34-4f0a-4f01-a4dc-0f263b307c72">

### Expected behavior

The Aave logo and branding have recently been updated.

### Would you like to work on this issue?

- [X] Yes
- [ ] No

wackerow:
Thanks @omahs! Sounds good, PR welcome, have assigned

PR #13378, opened 2024-07-11
Update Aave logo

omahs:
Resolves https://github.com/ethereum/ethereum-org-website/issues/13341


wackerow (changes requested):
Thanks @omahs! Logos look good, but I think we need to adjust these.

We current display the Aave logo on two pages:
- `/dapps` page -- `public/images/dapps/aave.png`
- `/stablecoins` -- `public/images/stablecoins/aave.png`

On `/stablecoins` we have a gradient background applied, and to match Aave branding I would suggest we use a white monochrome with a transparent background, so it looks similar to <https://aave.com/>:

image

For the `/dapps` page we should try to use a square aspect-ratio, so I would use the file currently being used here for the `/stablecoins` page (colored circle with white logo inside)

omahs:
@wackerow Thanks for the quick review! I updated the files

wackerow (approved):
Looks good! Thanks @omahs, I just pushed a commit to resize the icon back to the previous 160px. Pulling in!

wackerow:
@all-contributors please add @omahs for maintenance

PR #13381, opened 2024-07-12
Update wallet-data.ts

konopkja:
removing 2 wallets from New to crypto

- Shapeshift
- Metamask

### DIFF
--- a/src/data/wallets/wallet-data.ts
+++ b/src/data/wallets/wallet-data.ts
@@ -366,7 +366,7 @@ export const walletsData: WalletData[] = [
     social_recovery: false,
     onboard_documentation: "https://support.metamask.io",
     documentation: "",
-    new_to_crypto: true,
+    new_to_crypto: false,
   },
   {
     last_updated: "2023-01-25",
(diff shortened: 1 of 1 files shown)


src/data/wallets/wallet-data.ts
```diff
@@ -1929,7 +1929,7 @@ export const walletsData: WalletData[] = [
     social_recovery: false,
     onboard_documentation: "https://docs.shapeshift.com/",
     documentation: "https://docs.shapeshift.com/",
-    new_to_crypto: true,
+    new_to_crypto: false,
```
wackerow, line 1932:
```suggestion
```

src/data/wallets/wallet-data.ts
```diff
@@ -366,7 +366,7 @@ export const walletsData: WalletData[] = [
     social_recovery: false,
     onboard_documentation: "https://support.metamask.io",
     documentation: "",
-    new_to_crypto: true,
+    new_to_crypto: false,
```
wackerow, line 369: We can just remove these lines entirely, similar to the others
```suggestion
```

wackerow (approved):

PR #13386, opened 2024-07-12
feat: migrate text and headings to tailwind

TylerAPfledderer:
## Description

Migrates text and heading styles to Tailwind

- Creates a fontSize set of font-size/line-height default pairs in the tailwind config.
- Removes `Text` and `Heading` components.
  - Favors setting the heading level font-sizes in `global.css`
- Adds `Flex` and `Center` stacking components locally
  - These components are in addition to the existing stacking components added in #13347
  - To be reusable throughout the project in place of the Chakra components of the same name.

## Breaking Changes
Due to the approach taken to styling the heading levels, this might create some unexpected results when removing the `Heading` components in prod. 

For example, see instances where the `Accordion` is currently used. Provided an initial fix for the accordion story to revert the heading styling to `initial` (essentially "unsetting" the font-size and font-weight styles). When the accordion migration is addressed, will need to favor this same approach or use a CSS selector to override or ignore the declared heading level styles, then add to the styles for specific renders.

## Related Issue

Continuation of #13382


pettinarip (commented):
Great @TylerAPfledderer 

Looks good from the code side. I'll double check later on the preview deploy if there's no other part affected by the breaking changes you mentioned.

src/styles/global.css
```diff
@@ -141,13 +146,41 @@
 
 @layer base {
   body {
-    @apply bg-background font-body leading-base text-body;
+    @apply bg-background font-body text-body text-sm lg:text-md;
```
pettinarip, line 149: I think we can delete the [global styles](https://github.com/ethereum/ethereum-org-website/blob/dev/src/@chakra-ui/styles.ts#L13) from Chakra now.

pettinarip (approved):
LGTM!

The Accordion case that you mentioned is the only case where we are using a `h2` tag and not a `Heading` component. So, I've tested the app and it looks like everything is working fine. We should be good.

PR #13394, opened 2024-07-14
Bump prettier

pettinarip:
Bump to latest.

Continuation #13397

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,7 @@
     "plaiceholder": "^3.0.0",
     "polished": "^4.2.2",
     "postcss": "^8.4.39",
-    "prettier": "3.3.2",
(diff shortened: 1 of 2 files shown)


TylerAPfledderer:
@pettinarip recommend running `npx prettier ./src --write` here. About 128 changes.

Prettier is also not being accounted for unless the contributor has VSCode configured to format on save.

pettinarip:
> @pettinarip recommend running `npx prettier ./src --write` here. About 128 changes.
> 
> Prettier is also might not being accounted for unless the contributor has VSCode configured to format on save.

Yeah, was planning on doing that on a different PR because I wanted to add prettier to lintstaged to avoid IDE config problems.

minimalsm (approved):

PR #13395, opened 2024-07-14
Dune API to fetch total eth staked

pettinarip:
The current implementation fetches ~20 times to get all the data needed for the total eth staked chart on the home page.

This causes a few problems:
- bad DX because you have to wait a long time for these fetches to run
- the granularity of the chart is not very good since we are omitting days to avoid calling more times the API
- (most importantly) hitting rate limits and causing the build to fail

## Description

This PR uses a different API (Dune API) to do just one fetch call and simplify the whole process.

We shouldn't have any issues with rate limits now since the Dune API free plan supports [40 req per min](https://docs.dune.com/api-reference/overview/rate-limits) for this type of operation.

It uses the following [query](https://dune.com/queries/3915587) forked and customized for our case (limited to the last 90 days).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,6 +16,9 @@
 # GOOGLE_API_KEY=
 # GOOGLE_CALENDAR_ID=
 
+# Dune Analytics API key (required for total eth staked)
+# DUNE_API_KEY=
+
 # Matomo environment (URL and site ID required for analytics)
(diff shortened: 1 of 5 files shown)


wackerow:
@pettinarip Looks great! Higher resolution chart in the end as well... Still in draft for any reason?

wackerow (approved):

PR #13411, opened 2024-07-16
Shadcn migration - footer, skiplink, list & link components

pettinarip:
## Description

Migrates the Footer, SkipLink, List and Link to tailwind/shadcn.

Notes:
- For simplicity, I have created a new Link component under `/tailwind` to keep the PR small. When we migrate the rest, we can adopt this new component and ignore the old one.
- I disabled `resetCSS` from the Chakra provider to avoid global style conflicts cc @TylerAPfledderer
- Migrated the List component (composed by List, ListItem, UnorderedList and OrderedList). Currently, the Footer is the only component using it.

## Blockers
- ~I've added a `button.tsx` file as a temp component to migrate/test the Footer. @TylerAPfledderer is working on the real buttons implementation, so once that is ready, we can merge this.~


src/components/ui/list.tsx
pettinarip, line 1: Need to polish more this component. I think I could import the old global styles (see `global.css`) we had on the lists here.

TylerAPfledderer: I agree with that approach, since it looks like that was favored anyway with the Chakra components

src/components/Footer.tsx
```diff
@@ -306,132 +298,69 @@ const Footer = ({ lastDeployLocaleTimestamp }: FooterProps) => {
...
-        flexWrap="wrap"
-        gap={8}
-        pt={4}
-        pb={4}
-        borderTop={"1px solid"}
-        borderColor={"body.light"}
-      >
-        <Text fontSize={"sm"} fontStyle={"italic"} color={"body.medium"}>
+    <footer className="px-8 py-4">
+      <div className="flex flex-wrap items-center justify-center gap-8 border-t border-body-light pb-4 pt-4 md:justify-between">
```
wackerow, line 303:
```suggestion
      <div className="flex flex-wrap items-center justify-center gap-8 border-t border-body-light py-4 md:justify-between">
```

src/components/Footer.tsx
```diff
@@ -306,132 +298,69 @@ const Footer = ({ lastDeployLocaleTimestamp }: FooterProps) => {
...
         {linkSections.map((section: FooterLinkSection, idx) => (
-          <Box key={idx}>
-            <Heading as="h3" fontSize="sm" lineHeight="base" my="1.14em">
+          <div key={idx}>
+            <h3 className="my-5 text-sm font-bold">
               <Translation id={section.title} />
-            </Heading>
-            <List fontSize="sm" lineHeight="base" fontWeight="normal" m="0">
+            </h3>
+            <List className="m-0 mb-4 list-none text-sm">
```
wackerow, line 319: `mb-4`?

image

> _Production_

I don't see this in production; are we desiring this change?

src/components/Footer.tsx
```diff
@@ -306,132 +298,69 @@ const Footer = ({ lastDeployLocaleTimestamp }: FooterProps) => {
...
               {section.links.map((link, linkIdx) => (
-                <ListItem key={linkIdx} mb={4}>
-                  <BaseLink href={link.to} {...linkProps}>
+                <ListItem key={linkIdx} className="mb-4">
+                  <BaseLink
+                    href={link.to}
+                    className="text-body-medium no-underline hover:text-primary hover:after:text-primary"
```
wackerow, line 324: Should we have all this styling here? Or higher up in the component structure to avoid duplicating in the future?

src/components/Footer.tsx
```diff
@@ -306,132 +298,69 @@ const Footer = ({ lastDeployLocaleTimestamp }: FooterProps) => {
...
+        <div className="flex gap-4">
+          {socialLinks.map(({ to, ariaLabel, icon: Icon }) => (
             <BaseLink
               key={to}
               href={to}
               hideArrow
-              color="body.base"
               aria-label={ariaLabel}
-              _focus={{ color: "primary.base" }}
+              className="text-body hover:text-primary"
```
wackerow, line 343: Will this cover the `focus` state as well, or are we removing that?

src/layouts/BaseLayout.tsx
```diff
@@ -1,9 +1,8 @@
 // import { join } from "path"
 
 import dynamic from "next/dynamic"
-// import { useRouter } from "next/router"
-import { Container } from "@chakra-ui/react"
 
+// import { useRouter } from "next/router"
```
wackerow, line 5: ✂️ 
```suggestion
```

tailwind/Link.tsx
```diff
@@ -0,0 +1,193 @@
...
+import { getRelativePath } from "@/lib/utils/relativePath"
+import * as url from "@/lib/utils/url"
+
+import { DISCORD_PATH, SITE_URL } from "@/lib/constants"
+
+import { useRtlFlip } from "@/hooks/useRtlFlip"
+
+type BaseProps = {
+  /** @deprecated Use `href` prop instead */
+  to?: string
```
wackerow, line 19: At this point, could we officially remove the `to` prop here? We've been migrating these over time, haven't checked if there are any usages left

tailwind/Link.tsx
```diff
@@ -0,0 +1,193 @@
...
+              eventValue: href,
+            }
+          )
+        }
+        {...commonProps}
+      >
+        {children}
+        <VisuallyHidden>(opens in a new tab)</VisuallyHidden>
+        {!hideArrow && (
+          <Icon
```
wackerow, line 106: Just confirming we'll migrate this Chakra component later

wackerow (commented):
Thanks @pettinarip, looking great... Left some comments!

src/components/Footer.tsx
pettinarip: We currently have an mb in the `li`, on the ListItems, in production. I replaced it with this mb placed on the parent `ul'. Basically, it is the space between the list section and the next component below it. I moved the place where we declare this space since I think it is more appropriate to place it on the parent than on the children.

src/components/Footer.tsx
pettinarip: You are referring to the two Links that we use just in the Footer, right? not other Links that are outside of the footer.

Was wondering if you have a preferred way to handle this case. I could think of the following options:
- Create a new component called `MutedLink` or something like that that encapsulates these styles, but that seems like overkill in this case where we only have 2 uses
- Create a class for the link called `.link-muted`. Something similar to the previous idea
- Define the classes in a variable => we will use this for now

src/components/Footer.tsx
pettinarip: Ahh, good catch! missed this one. fixing...

tailwind/Link.tsx
pettinarip: Not yet, we still have many places where we are using this prop. However, should be a pretty easy refactor but wouldn't like to include that in this PR just to keep it small.

tailwind/Link.tsx
pettinarip: Ahhh! another good catch, forgot to migrate this part.

src/components/Footer.tsx
pettinarip: Fixed, I added it globally to all the `a` tags.

src/components/Footer.tsx
pettinarip: I've added a variable to store the reusable styles.

tailwind/Link.tsx
pettinarip: Marking this as resolved. I'll create an issue to keep track of this refactor.

pettinarip: #13423

src/components/SkipLink.tsx
```diff
@@ -1,27 +1,19 @@
...
         href={"#" + MAIN_CONTENT_ID}
-        lineHeight="taller"
-        position="absolute"
-        top="-12"
-        ms="2"
-        color="background.base"
-        textDecorationLine="none"
-        _hover={{ textDecoration: "none" }}
-        _focus={{ position: "static" }}
+        className="absolute -top-3 ms-2 leading-8 text-background no-underline hover:no-underline focus:static"
```
wackerow, line 13: Noticing that the `absolute` positioning is enough to hide this (with `focus:static` kicking in when the keyboard tab stops there.

We may be able to just remove `-top-3`, but if we want to keep it, we should change this to `-top-12` to match before, which slides this element 3rem up, and off the top edge of the screen. Right now it's only sliding it up 0.75rem.

wackerow (changes requested):
Looks good @pettinarip! Just a few comments

src/layouts/BaseLayout.tsx
```diff
@@ -54,7 +52,7 @@ export const BaseLayout = ({
        * layout on initial load.
        */}
       <SkipLink />
-      <Container maxW="container.2xl" marginInline="auto">
+      <div className="2xl:container 2xl:mx-auto">
```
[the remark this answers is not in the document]
pettinarip: I understand that this is going to be the same at the end but open to change it if it is clearer.

src/components/SkipLink.tsx
pettinarip: Ah! good catch, my bad. Thanks!

wackerow (approved):

wackerow (changes requested):
@pettinarip Uh oh, noticing the "Skip to content" link isn't actually taking me past the nav bar, and visually seems to get stuck half way. In this video it shows production first, then this PR:

https://github.com/user-attachments/assets/126ee035-58cf-4520-8c73-83b7feb61ba4

pettinarip:
> @pettinarip Uh oh, noticing the "Skip to content" link isn't actually taking me past the nav bar, and visually seems to get stuck half way. In this video it shows production first, then this PR:
> 
>  2024-07-25.12-59-26.mov

Ohh good catch...uhmm okok, I'll take a closer look, thanks!

tailwind/Link.tsx
```diff
@@ -0,0 +1,190 @@
...
+              eventValue: href,
+            }
+          )
+        }}
+        {...commonProps}
+      >
+        {children}
+      </a>
+    )
+  }
```
wackerow, line 159: image

@pettinarip We were not using NextLink before, just Chakra for styling. A simple anchor should suffice here similar to `isExternal`, and this fixes the issue with the SkipNav button not actually moving the focus to the desired `#main-content` element.

wackerow (approved):

tailwind/Link.tsx
pettinarip: Ahh got it, thanks for the fix!

PR #13419, opened 2024-07-18
refactor: migrate buttons to tailwind

TylerAPfledderer:
## Description

Updates the button components to use ShadCN/Tailwind styling.

Also rebuilds `ButtonTwoLines` to be a little more verbose in determining which type of button component it renders via a deterministic prop, to provide better type guarding.

These updates here are isolated to the `tailwind` directory and are not migrated to prod. Update stories are also isolated here to not effect Chromatic snapshots. (To be handled later in conjunction with syncing with updated link components)


src/components/ui/buttons/Button.tsx
```diff
@@ -0,0 +1,149 @@
...
+import { type MatomoEventOptions, trackCustomEvent } from "@/lib/utils/matomo"
+import { scrollIntoView } from "@/lib/utils/scrollIntoView"
+
+const buttonVariants = cva(
+  "inline-flex gap-2 items-center justify-center rounded border-[1px] border-solid border-current text-primary transition focus-visible:outline focus-visible:outline-4 focus-visible:outline-primary-hover focus-visible:-outline-offset-1 disabled:text-disabled disabled:pointer-events-none hover:text-primary-hover [&[data-secondary='true']]:text-body",
+  {
+    variants: {
+      variant: {
+        solid:
+          "!text-background bg-primary border-transparent disabled:bg-disabled disabled:text-background hover:text-background hover:bg-primary-hover hover:shadow-button-hover active:shadow-none",
```
pettinarip, line 17: is it necessary the `!`?
```suggestion
          "text-background bg-primary border-transparent disabled:bg-disabled disabled:text-background hover:text-background hover:bg-primary-hover hover:shadow-button-hover active:shadow-none",
```

src/components/ui/buttons/Button.tsx
```diff
@@ -0,0 +1,149 @@
...
+  ({ linkProps, customEventOptions, children, ...buttonProps }, ref) => {
+    const handleClick = () => {
+      customEventOptions && trackCustomEvent(customEventOptions)
+    }
+    return (
+      <Button asChild {...buttonProps}>
+        <BaseLink
+          ref={ref}
+          activeStyle={{}}
+          // TODO: Redress this override when migrating the link component
```
pettinarip, line 124: Cool, yes. Once we merge this and the link PR, we can fix this.

src/components/ui/buttons/Button.tsx
TylerAPfledderer: @pettinarip Checkout the override story and the `ButtonLink`. Without marking this style as `!important` the text color gets lost.

This still has to do with specificity between Tailwind and Chakra, where I shouldn't have it marked like this once Chakra is not a part of this.

Not going to know until this PR is merged and checked with updates you did to the links.

TylerAPfledderer:
> Looking great @TylerAPfledderer.
> 
> Since that this is a new Button component, we could place it under the `/tailwind` folder to clearly see that this component hasn't been migrated yet on the codebase like the Accordion example or the Link component.
> 
> Another thing, naming, the shadcn cli installs the components using snake-case. Do you know if we can configure it to install using CapitalizedCase? I think we should be consistent with that.

As to moving this under `tailwind` folder, I would be moving the updated stories in there as well so Chromatic doesn't see them. I don't think the existing story files should be replaced yet because of `ButtonLink`, and `Button` and `Link` I think are to coupled to check baselines before properly addressing them together.

With the component naming, I don't see an option in the schema to automatically change them. This seems to be an opinionated part of the tooling.

Also, there are [two different styles of component writing](https://ui.shadcn.com/docs/components-json#style)? I cannot tell what the difference is!

src/components/ui/buttons/Button.tsx
pettinarip: Got it, ok.

pettinarip:
> As to moving this under `tailwind` folder, I would be moving the updated stories in there as well so Chromatic doesn't see them. I don't think the existing story files should be replaced yet because of `ButtonLink`, and `Button` and `Link` I think are to coupled to check baselines before properly addressing them together.

Yeah, sounds good! yes, I don't think we can replace them atm. We can replace them later and check for any regressions that we don't currently see now.

> With the component naming, I don't see an option in the schema to automatically change them. This seems to be an opinionated part of the tooling.

Right, weird that they didn't add that to the CLI. We could go later and rename them to be consistent.

> Also, there are [two different styles of component writing](https://ui.shadcn.com/docs/components-json#style)? I cannot tell what the difference is!

Me neither haha. All I see is that they use different font sizes in some places but they are pretty similar.

TylerAPfledderer:
@pettinarip should be good to look over. Added a comment in the [UI Tests](https://www.chromatic.com/build?appId=63b7ea99632763723c7f4d6b&number=2716) check.

pettinarip:
> @pettinarip should be good to look over. Added a comment in the [UI Tests](https://www.chromatic.com/build?appId=63b7ea99632763723c7f4d6b&number=2716) check.

Hey, I don't see this diff in the last build. Wondering if you fix it.

tailwind/ui/buttons/Button.tsx
```diff
@@ -0,0 +1,149 @@
...
+const buttonVariants = cva(
+  "inline-flex gap-2 items-center justify-center rounded border-[1px] border-solid border-current text-primary transition focus-visible:outline focus-visible:outline-4 focus-visible:outline-primary-hover focus-visible:-outline-offset-1 disabled:text-disabled disabled:pointer-events-none hover:text-primary-hover [&[data-secondary='true']]:text-body",
+  {
+    variants: {
+      variant: {
+        solid:
+          "!text-background bg-primary border-transparent disabled:bg-disabled disabled:text-background hover:text-background hover:bg-primary-hover hover:shadow-button-hover active:shadow-none",
+        outline: "hover:shadow-button-hover active:shadow-none",
+        ghost: "border-transparent",
+        link: "border-transparent font-bold underline py-0 px-1 active:text-primary",
```
pettinarip, line 20: This won't have any effect, I would just remove them to keep this cleaner.

IMO this should be controlled inside the `sizes`, I wouldn't add anything yet though but, in case we need it, I think it would be better to add a `link` or `inline` size.
```suggestion
        link: "border-transparent font-bold underline active:text-primary",
```

tailwind/ui/buttons/Button.tsx
```diff
@@ -0,0 +1,149 @@
...
+      },
+      size: {
+        md: "min-h-10.5 px-4 py-2 [&>svg]:text-2xl",
+        sm: "text-xs min-h-[31px] py-1.5 px-2 [&>svg]:text-md",
```
pettinarip, line 24: Noted that sometimes the icon shrinks and doesn't respect the desired font-size.

image

Perhaps we could add `[&>svg]:flex-shrink-0` to the base styles to avoid this.

tailwind/ui/buttons/ButtonTwoLines.tsx
```diff
@@ -0,0 +1,127 @@
...
+    <>
+      {isIconLeft && <ButtonIcon />}
+      <Stack
+        className={cn(
+          "gap-0",
+          reverseTextOrder ? "flex-col-reverse" : "flex-col"
+        )}
+      >
+        <span
+          className={cn("text-md", size === "md" ? "text-bold" : "text-normal")}
```
pettinarip, line 109:
```suggestion
          className={cn("text-md", size === "md" ? "font-bold" : "font-normal")}
```

tailwind/ui/buttons/ButtonTwoLines.tsx
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+type ButtonLinkTypeProps = CommonProps &
+  Omit<ButtonLinkProps, OmittedTypes> & {
+    componentType: "link"
+  }
+
+type ButtonTwoLinesProps = ButtonTypeProps | ButtonLinkTypeProps
+
+const ButtonTwoLines = ({
```
pettinarip, line 51: One minor observation, in the prod version, we have a bigger `py` when `size=md`.

`size === "md" ? "py-4" : "py-2"`

pettinarip (changes requested):
@TylerAPfledderer left a few minor comments to polish a bit more but overall looks great.

tailwind/ui/buttons/Button.tsx
TylerAPfledderer: @pettinarip Can do a [compound variant](https://cva.style/docs/getting-started/variants#compound-variants) here

tailwind/ui/buttons/Button.tsx
TylerAPfledderer: Makes sense here!

tailwind/ui/buttons/ButtonTwoLines.tsx
TylerAPfledderer: Ahh yes, I missed migrating this over!

tailwind/ui/buttons/Button.tsx
pettinarip: Makes sense! what do you have in mind? something like this?

```ts
// could be weird because we are not composing nothing but could work really well
{
  variant: "link",
  className: "py-0 px-1"
}
```

TylerAPfledderer: @pettinarip Something like that.

Better than 
```tsx
// potential redundancy 
<Button variant="link" size="link" />
```

Also, when are we expected to use this variant when there is a `ButtonLink` component and we just use a normal link anyway inline.

pettinarip: Got it, agree. Ok, I think once we add that we should be good to go. Nice job!

TylerAPfledderer: @pettinarip sorry, that last bit was supposed to be a question! 😅

pettinarip: Ahh got it xD

Good point, this is a really minor use case that we don't use. I just marked it to be consistent with the previous implementation.

We can skip it for now and add it later if we find it more relevant.

pettinarip (approved):
LGTM!

Issue #13423, opened 2024-07-18
Refactor Link component: cleanup deprecated `to` prop


pettinarip:
## Description

The `to` prop in the Link component [is marked as deprecated](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Link.tsx#L23).

TODO:
- Replace all the instances where we are using `to` with `href`.
- Remove the `to` prop from the Link component.

jackabald:
Hey!

I want to take a stab at this issue. Is anyone working on this?

pettinarip:
@jackabald assigned to you, thanks!

jackabald:
@pettinarip Do you want all references to the 'to' prop to be removed? I'm finding instances of 'to' being used for <DocLink> and <ButtonLink>. Should it be replaced here as well?

Thanks!

PR #13433, opened 2024-07-19
fix: removed deprecated 'to' and replaced with 'href' (#13423)

jackabald:
## Description

Removed all deprecated 'to' props with href props 
Kept the 'to' prop for all custom components.

## Related Issue

#13423


pettinarip (commented):
Thanks for the PR @jackabald 

Noticed that the build fails due to a problem with the Breadcrumbs implementation. I'll try to fix this as soon as possible.

We should also remove the deprecated `to` prop from the link component now. I'll
a commit to fix this as well.

pettinarip (approved):
LGTM! thanks @jackabald

corwintines (commented):
LGTM, thanks @jackabald!

corwintines (approved):

PR #13434, opened 2024-07-20
refactor(global.css): add complementary color semantics

TylerAPfledderer:
## Description

Writes the semantic tokens for the complementary colors in the Tailwind config.

## Related Issue
Forgot to be included in #13347

PR #13441, opened 2024-07-21
feat: set Checkbox component for ShadCN

TylerAPfledderer:
## Description

Set up the Checkbox component for ShadCN/Tailwind


tailwind/ui/__stories__/Checkbox.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+    disabled: true,
+  },
+  {
+    id: "disabled-checked",
+    value: "disabled-checked",
+    label: "disabled-checked",
+    disabled: true,
+    checked: true,
+  },
+  // TODO: Waiting for PR #13434 to be merged
```
TylerAPfledderer, line 38: - #13434 

I'd consider this a blocker, as the related styles are not set.

tailwind/ui/Checkbox.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { RxCheck } from "react-icons/rx"
+import * as CheckboxPrimitive from "@radix-ui/react-checkbox"
+
+import { cn } from "@/lib/utils/cn"
+
+export type CheckboxProps = React.ComponentPropsWithoutRef<
+  typeof CheckboxPrimitive.Root
+>
+
+const Checkbox = React.forwardRef<
```
TylerAPfledderer, line 11: @pettinarip this will be different than Chakra, where the preferred convention is to work with [Form Components](https://ui.shadcn.com/docs/components/form) to set the label, controlled values, and layout.

ShadCN like `react-hook-form` for this, and Radix has their own [Form component in preview](https://www.radix-ui.com/primitives/docs/components/form). There is also still an [Open RFC PR](https://github.com/radix-ui/primitives/pull/1977)

Or, modify this to reflect how Chakra sets up rendering the `label`. Would have to do the same with `switch` and `radio`. Personally, I'd rather have more control with the layout of the label and/or utilizing a higher-level form component as mentioned above.

tailwind.config.ts
```diff
@@ -106,7 +107,23 @@ const config = {
       },
     },
   },
-  plugins: [require("tailwindcss-animate")],
+  plugins: [
+    require("tailwindcss-animate"),
+    plugin(function ({ matchVariant }) {
```
TylerAPfledderer, line 112: This plugin creates a utility class for the `:not()` pseudo. It allows for either...

An arbitrary variant
```css
/* "not-[:checked]:bg-primary" */

.not-\[\:checked\]\:bg-primary:not(:checked) {
    background-color: var(--primary);
}
```

Or a specifically defined variant, such as `disabled` here
```css
/* not-disabled:bg-primary */

.not-disabled\:bg-primary:not(:disabled) {
    background-color: var(--primary);
}
```

tailwind/ui/Checkbox.tsx
TylerAPfledderer: Update: A form component library of some kind will need to be used, as the form controls do not render their own `input`.

Checkbox and radio only render a `button` which is the control itself. Using `FormControl` from ShadCN, say, and wrap it around the `Checkbox` will render the checkbox button and a hidden `input`.

pettinarip: Makes sense. I would lean towards using `react-hook-form` since it is the recommended approach and we have used it on other projects.

But it shouldn't be a must to wrap every checkbox with a form container, right? for example the checkbox used in https://ethereum.org/en/wallets/find-wallet/ or do you see a problem using the button tag instead of the input?

As for the scope of this PR, we should be good as it is, right?

TylerAPfledderer: With the checkbox in the find wallet page, it should be possible as it stands, as there is still the `onChange` prop, which is all that is really needed there for functionality.

So in Chakra, the `Checkbox` is already rendering a hidden `input` and the control is a styled `span` tag.


It's essentially built as (in the primitive HTML structure)
```tsx
<label>
  <input hidden />
  <span /> // The visible control
  <span>{ children }</span> // This renders if there is children text
</label>
```

But we can be more flexible here. So leaving the ShadCN checkbox as is will be fine.

ShadCN checkbox is more primitive than Chakra's, because it's just the control as a `button`. The `button` is used so that the control accessible and easily click to toggle `checked` or `unchecked`. You would then do a controlled hook-up to a hidden `input` when used in a form. 

In short, I think we can leave it as is.

pettinarip: Got it. Cool, then we are fine. Thanks for the explanation.

pettinarip (approved):
LGTM

PR #13448, opened 2024-07-22
add further reading and video for /desci

thanedouglass:
added new entries for paige under sections: Further Reading, Get Involved, and Videos


wackerow (changes requested):
Thank @thanedouglass.. we need to remove the opensea link, but the others look good from my end.

corwintines (approved):

PR #13449, opened 2024-07-23
Shadcn migration - mobile menu

pettinarip:
## Description

Mobile menu implementation using Shadcn.

## Related Issue

Depends on #13411 and #13419


src/components/Nav/Mobile/LvlAccordion.tsx
```diff
@@ -36,131 +35,128 @@ const LvlAccordion = ({
...
+
         if (isLink)
           return (
             <AccordionItem
               key={label}
-              borderTop="1px"
-              borderColor="inherit"
-              _last={{ borderBottomWidth: "1px" }}
+              value={label}
+              className={cn("border-t border-body-light last:border-b")}
```
pettinarip, line 61:
```suggestion
              className="border-t border-body-light last:border-b"
```

src/hooks/useDisclosure.ts
```diff
@@ -0,0 +1,18 @@
...
+
+  return {
+    isOpen: value,
+    onOpen: setTrue,
+    onClose: setFalse,
+    onToggle: toggle,
+  }
+}
+
+export default useDisclosure
```
wackerow, line 18: We use named exports for our other hooks. Would suggest we flip this for consistency

```suggestion
export { useDisclosure }
```

src/components/Nav/Mobile/index.tsx
```diff
@@ -14,55 +14,50 @@ import MenuBody from "./MenuBody"
 import MenuFooter from "./MenuFooter"
 import MenuHeader from "./MenuHeader"
 
+import useDisclosure from "@/hooks/useDisclosure"
```
wackerow, line 17: (Only if we switch to a named export)

```suggestion
import { useDisclosure } from "@/hooks/useDisclosure"
```

src/components/Nav/Mobile/ExpandIcon.tsx
```diff
@@ -17,14 +16,9 @@ type ExpandIconProps = {
...
-  <Icon
+  <svg
     viewBox="0 0 24 25"
-    width={6}
-    height={6}
-    position="relative"
-    strokeWidth="2px"
-    display="inline-block"
-    stroke="currentColor"
+    className="relative inline-block h-6 w-6 stroke-current stroke-2 group-hover:text-primary-hover"
```
wackerow, line 21: image

Is `group-hover:text-primary-hover` correct? Production does not use the primary color here on hover/pressing

wackerow (changes requested):
@pettinarip Overall looks good! Left one comment which appears to be a visual regression. If we patch that I think this is ready to be pulled in.

src/components/Nav/Mobile/ExpandIcon.tsx
pettinarip: So, this is prod:
https://github.com/user-attachments/assets/d223a229-0c69-46a9-ad9a-efa5226c2656

Anyway, now that I think, it doesn't really matter, this is a mobile only menu where we are not going to have hover events. So, happy to remove it to make it cleaner.

wackerow (approved):
Looks good @pettinarip! Left a couple comments but going to pull this in

PR #13451, opened 2024-07-23
fix: Recommend Holesky over Goerli on "run a node"

jarrodwatts:
Typo fix to use Holesky over Goerli when setting up a node in testnet


corwintines (approved):
Nice catch, thanks @jarrodwatts!

PR #13456, opened 2024-07-23
Migrate FeedbackWidget to shadcn

wackerow:
## Description
- Migrates FeedbackWidget and associated components to shadcn, deprecating use of Chakra dependencies
- Adds custom `useDisclosure` hook, replacing Chakra import
- Adds shadcn base `popover` component via command line

## Preview URL
https://deploy-preview-13456--ethereumorg.netlify.app/en/
https://deploy-preview-13456--ethereumorg.netlify.app/nft/

## Related Issue
shadcn migration


konopkja:
Screenshot 2024-07-24 at 20 18 23

 think the secondary background color isnt necessary and could be removed

src/components/FeedbackWidget/FixedDot.tsx
```diff
@@ -1,61 +1,58 @@
...
+    return (
+      <Button
+        ref={ref}
+        data-testid="feedback-widget-button"
+        aria-label={t("feedback-widget")}
+        className={cn(
+          "lg:mt-inherit sticky bottom-4 z-20 me-4 ms-auto flex size-12 items-center gap-0 rounded-full text-white shadow-table-item-box",
+          "transition-all duration-200 hover:scale-110 hover:transition-transform hover:duration-200",
+          offsetBottom && "bottom-31 lg:bottom-4",
+          isExpanded ? "lg:w-60 lg:gap-3" : `lg:w-12`,
```
pettinarip, line 28:
```suggestion
          isExpanded ? "lg:w-60 lg:gap-3" : "lg:w-12",
```

src/components/FeedbackWidget/index.tsx
```diff
@@ -31,99 +28,70 @@ const FeedbackWidget = () => {
...
-                    : t("feedback-widget-prompt")}
-                </AlertDialogHeader>
-                <AlertDialogCloseButton alignSelf="start" />
-              </HStack>
+        <PopoverTrigger asChild>
+          <FixedDot offsetBottom={offsetBottom} isExpanded={isExpanded} />
+        </PopoverTrigger>
+
+        <PopoverContent className="mx-2 w-80 max-w-[calc(100vw_-_1rem)] rounded bg-background p-4 sm:p-8">
+          <div className="items-top flex gap-2">
```
pettinarip, line 40: `items-top` doesn't exist.

```suggestion
          <div className="flex items-start gap-2">
```

pettinarip (changes requested):
Looks good @wackerow 

One minor that I see on the preview is that the popover content moves a little bit when you hover the button and it scales.

https://github.com/user-attachments/assets/9989d4f0-7c48-440b-9b82-e987aeb49d50

We could fix this by disabling the scaling effect when the popover is opened. What do you think?

src/components/FeedbackWidget/index.tsx
wackerow: But I want it to! 😉 Good catch

wackerow:
@konopkja Removed that shadow effect seen on hover

@pettinarip Conditionally paused the `scale` effect when open

Ready for re-review.

src/components/FeedbackWidget/index.tsx
pettinarip: haha! is a good name though

pettinarip (approved):
LGTM @wackerow 

I've just pushed a commit to regenerate the yarn.lock file since it was installing an old version of `@netlify/plugin-nextjs`.

PR #13459, opened 2024-07-23
migrate: useDisclosure hook

wackerow:
## Description
- Builds on custom internal `useDisclosure` hook introduced in #13456
- Deprecates all usage of Chakra-UI `useDisclosure` hook, updating imports to use new custom `src/hooks/useDisclosure.ts`

## Related Issue
shadcn migration

### DIFF
--- a/src/hooks/useDisclosure.ts
+++ b/src/hooks/useDisclosure.ts
@@ -8,9 +8,13 @@ type UseDisclosureProps = {
   id?: string
 }
 
-type GetButtonPropsProps = Record<string, unknown> & { onClick?: () => void }
+type GetButtonPropsProps = Record<string, string | undefined> & {
+  onClick?: () => void
+}
 
-type GetDisclosurePropsProps = Record<string, unknown> & { id?: string }
+type GetDisclosurePropsProps = Record<string, string | undefined> & {
+  id?: string
+}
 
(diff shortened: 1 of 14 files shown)


src/hooks/useDisclosure.ts
pettinarip, line 1: I think this hook is too complex for what we need for the new components. I've created a simpler `useDisclosure` [here](https://github.com/ethereum/ethereum-org-website/pull/13449/files#diff-92bd0ffe4a3a1593072af6ae4efac4422a04a07255897cad9ff007dab46ce896) that, I think will be a better fit to our use cases from now on. LMK what do you think, do we need more than a simple boolean flag? all the a11y attr will be handled by the radix components.

pettinarip (changes requested):

src/hooks/useDisclosure.ts
wackerow: This approach was primarily to match previous functionality, so the other places we use `useDisclosure` could be swapped out without any additional refactoring (as in d804a56).

I'd be fine with minimizing to the example you shared since I don't believe we'll need the `getButtonProps` and `getDisclosureProps` helpers anymore with shadcn. In this case, I can close this PR/branch, and I can update #13456 to use your suggested `useDisclosure` from 263698d inside #13449. 👍

PR #13464, opened 2024-07-24
Sentry

pettinarip:
## Description

Adds Sentry.

Its adding 33kb to the main bundle.
image

We will test this on production, see how useful it is, and then decide to keep it or not.


next.config.js
```diff
@@ -67,5 +89,15 @@ module.exports = (phase, { defaultConfig }) => {
     }
   }
 
-  return nextConfig
+  return withBundleAnalyzer(
+    withSentryConfig(nextConfig, {
+      // TODO: temp config, update this to the correct org & project
+      org: "ethereumorg-ow",
+      project: "javascript-nextjs",
```
pettinarip, line 96: Will do this later (in a separate PR) & before the next RC.

next.config.js
```diff
@@ -32,13 +37,30 @@ module.exports = (phase, { defaultConfig }) => {
...
+
       return config
     },
     i18n,
     trailingSlash: true,
     images: {
       deviceSizes: [640, 750, 828, 1080, 1200, 1504, 1920],
     },
+    env: {
+      NEXT_PUBLIC_CONTEXT: process.env.CONTEXT,
```
pettinarip, line 59: Im exposing this env var to segment the reports better. https://docs.netlify.com/configure-builds/environment-variables/#build-metadata

> It can be production, deploy-preview, branch-deploy, or dev.

We could also just trace things on `production` contexts.

netlify.toml
```diff
@@ -11,6 +11,7 @@
 
 [build.environment]
   NEXT_FORCE_EDGE_IMAGES="true"
+  BUILD_LOCALES="en"
```
pettinarip, line 14: TODO

- [ ] remove before merging

instrumentation.ts
```diff
@@ -0,0 +1,28 @@
+import * as Sentry from "@sentry/nextjs"
+
+export async function register() {
+  const dns = process.env.NEXT_PUBLIC_SENTRY_DSN
```
corwintines, line 4:
```suggestion
  const dsn = process.env.NEXT_PUBLIC_SENTRY_DSN
```

I think renaming this variable would be a bit more clear since we are getting NEXT_PUBLIC_SENTRY_DSN and passing as dsn props in commonSentryOptions

sentry.client.config.ts
```diff
@@ -0,0 +1,10 @@
+import * as Sentry from "@sentry/nextjs"
+
+const dns = process.env.NEXT_PUBLIC_SENTRY_DSN
```
corwintines, line 3: Same deal here with `dsn` name

corwintines (approved):
Left a non commenting suggestion, but LGTM

instrumentation.ts
pettinarip: oops, thanks! a typo confused by "domain name server" xD

wackerow (approved):

PR #13472, opened 2024-07-24
term correction

bskrksyp9:
Retained `SELFDESTRUCT` keyword rather than SELF-DESTRUCT


wackerow (approved):
Looks good! Thanks again @bskrksyp9, appreciate the polishes =)

PR #13476, opened 2024-07-24
fix typo

bskrksyp9:
## Description
Corrected spelling (for reference word)


wackerow:
Thanks @bskrksyp9! I see this typo on two pages, both for English and translated versions:

- `/defi/`
- `/developers/docs/networking-layer/portal-network/`

We don't want to be updating the non-English versions in this manner. I'd welcome a PR to adjust this _for English version only_ on each of these two pages, and then the non-English versions can/will be updated through Crowdin in a separate process. Thanks!

bskrksyp9:
> Thanks @bskrksyp9! I see this typo on two pages, both for English and translated versions:
> 
> - `/defi/`
> - `/developers/docs/networking-layer/portal-network/`
> 
> We don't want to be updating the non-English versions in this manner. I'd welcome a PR to adjust this _for English version only_ on each of these two pages, and then the non-English versions can/will be updated through Crowdin in a separate process. Thanks!

Alright, understood 🙂

PR #13485, opened 2024-07-24
Further deprecate `to` for `href`

corwintines:
## Description

- Goes through and further deprecates `to` props


src/components/Link.tsx
```diff
@@ -57,6 +57,7 @@ export const BaseLink = forwardRef(function Link(
   const { flipForRtl } = useRtlFlip()
 
   if (!href) {
+    console.warn(children)
```
corwintines, line 60:
```suggestion
```

wackerow (approved):
cc: @lukassim Ideally these do not all get undone in the next Crowdin import... majority if not all should be content that should not be getting translated anyway, so hopefully there is a way to auto approve these.

PR #13487, opened 2024-07-25
Updated PT-BR language resources

Gwinest:
Added an updated popular portal to the list of Ethereum related sources in brazilian portuguese.


wackerow:
@isabelladebrito or @minimalsm Are we looking to list this type of resource? Little hard for me to tell, but appears to be an investors guide?

Gwinest:
Yes, it is an investors guide, it has some Ethereum related guides,
including EIPs descriptions and explanations in Portuguese, I'm a developer
and I find it very useful

Em sex., 26 de jul. de 2024 18:47, Paul Wackerow ***@***.***>
escreveu:

corwintines:
If there are any specific pages for the EIP's then I think its ok. But, as a whole if there is a lot of investor guides then I don't think there is a need for it.

corwintines:
Going to close this out as is @Gwinest. But, if you would like to suggest some resources on EIPs that would be good here. Wary of including stuff investor guide material.

PR #13490, opened 2024-07-25
Patch color mode sync

pettinarip:
Not an ideal solution but could serve as a workaround.

## Description

Manually syncs the chakra context color mode value in order to keep using the `useColorModeValue` hook from chakra.


src/components/Nav/useNav.ts
pettinarip, line 1: Main change to keep both in sync

corwintines (approved):

PR #13503, opened 2024-07-26
Shadcn migration - desktop menu

pettinarip:
## Description

Desktop menu implementation using Tailwind classes.


wackerow (approved):
What a clean switch here

PR #13531, opened 2024-07-29
Shadcn migration - language picker

pettinarip:
DEPENDS ON:
- #13411
- #13449

## Description

Language picker implementation + refactor using Shadcn's Popover, Command, Progress, and Badge components.

## TODO
- [x] matomo events — focused input event, do we still need this? check with team
- [x] check rtl support
- [x] check keyboard shortcuts
- [x] check `ellipsis` styles on the component item
- [x] implement Shadcn Link component from #13411
- [x] implement Shadcn FooterButton from #13449


konopkja:
Screenshot 2024-08-05 at 16 35 29


based on these data I am assuming the filter input is not being used very frequently and therefore should not be automatically focused as primary action.

This would be especially problematic on mobile where it would also trigger showing keyboard and overlap other options for selection.

konopkja:
cc @minimalsm  @pettinarip

pettinarip:
> based on these data I am assuming the filter input inside language picker is not being used very frequently and therefore should not be automatically focused as primary action.

Does that make any negative impact to the UX? IMO:
- it will improve the UX for those who want to change the language quicker
- it will keep the same UX for those who just open and close the picker

> This would be especially problematic on mobile where it would also trigger showing keyboard and overlap other options for selection.

Agree. However, I tested it on a physical phone and didn't get the focus nor the keyboard popping up. Would be nice to have a second test to confirm that we don't see the keyboard.

konopkja:
> > based on these data I am assuming the filter input inside language picker is not being used very frequently and therefore should not be automatically focused as primary action.
> 
> Does that make any negative impact to the UX? IMO:
> 
> * it will improve the UX for those who want to change the language quicker
> * it will keep the same UX for those who just open and close the picker
> 
> > This would be especially problematic on mobile where it would also trigger showing keyboard and overlap other options for selection.
> 
> Agree. However, I tested it on a physical phone and didn't get the focus nor the keyboard popping up. Would be nice to have a second test to confirm that we don't see the keyboard.



> > based on these data I am assuming the filter input inside language picker is not being used very frequently and therefore should not be automatically focused as primary action.
> 
> Does that make any negative impact to the UX? IMO:
> 
> * it will improve the UX for those who want to change the language quicker
> * it will keep the same UX for those who just open and close the picker
> 
> > This would be especially problematic on mobile where it would also trigger showing keyboard and overlap other options for selection.
> 
> Agree. However, I tested it on a physical phone and didn't get the focus nor the keyboard popping up. Would be nice to have a second test to confirm that we don't see the keyboard.

do you mean this PR (focus on the input) would only affect desktop and mobile?

pettinarip:
> do you mean this PR (focus on the input) would only affect desktop and mobile?

As far as I have tested it, the focus is only happens on desktop. On mobile devices, I have to manually click on the input to filter things.

src/components/LanguagePicker/MenuItem.tsx
```diff
@@ -48,86 +42,58 @@ const MenuItem = forwardRef(({ displayInfo, ...props }: ItemProps, ref) => {
...
       {...props}
+      // TODO double check this
+      // sx={{
+      //   p: {
+      //     textDecoration: "none",
+      //     overflow: "hidden",
+      //     textOverflow: "ellipsis",
+      //     whiteSpace: "nowrap",
+      //   },
+      // }}
```
wackerow, line 62: Haven't encountered any need for this while testing it out.

wackerow (commented):
This looks great @pettinarip! Can we just update status of TODO's and clear any conflicts? Think we can pull in from there

src/components/LanguagePicker/MenuItem.tsx
pettinarip: Thank, yes, I was trying to see where this was coming from but I think we could clean this up since it doesn't seem to add anything atm.

pettinarip:
> This looks great @pettinarip! Can we just update status of TODO's and clear any conflicts? Think we can pull in from there

Done. I think we are ready for another review.

src/components/ui/command.tsx
```diff
@@ -0,0 +1,171 @@
+import * as React from "react"
+import { Command as CommandPrimitive } from "cmdk"
```
wackerow, line 2: Discovering these new packages along the way with the shadcn-ui approach. Loving it.

wackerow (approved):
Looks awesome! Great job @pettinarip! Pulling in

PR #13532, opened 2024-07-29
Update: ERC-20 Token Standard description [Fixes #13218]

### LINKED ISSUE #13218: Highlight the severity of the ERC-20 known issue in the token standard description.

With [8.7.0](https://github.com/ethereum/ethereum-org-website/pull/13204#event-13233428988) a "Known issues" section was added to ERC-20 description https://ethereum.org/en/developers/docs/standards/tokens/erc-20/#erc20-issues

However, the representation looks like it is a known issue so a token developer has nothing to worry about. It mustn't look like this because the amount of damage that this issue is dealing to the ecosystem is still growing over time and if there will be no active countermeasures **people will keep losing money because of this known issue**.

I would propose to add two paragraphs here: 
ERC-20_losses_on_ethereum-org

1. The severity of the issue must be highlighted so that devs wouldn't think that they don't need to worry about it.
2. There are few ways to mitigate a significant part of potential damage that this issue can deal so this options must be proposed.

### 1. Highlight the severity of the issue

I would recommend to add a phrase "As of 06/20/2024 at least $83,656,418 worth of ERC-20 tokens were lost due to this issue."

We have built a script that calculates the amount of lost tokens (https://dexaran.github.io/erc20-losses/) and there are over $230M worth of lost tokens on Ethereum mainnet currently and this amount is growing every day as users keep losing tokens and those tokens that were already lost are never recovered. 

My team did a due diligence on the tokens that were reported as "lost" by this script and we were able to manually verify that at least $83M worth of tokens are clearly unrecoverably lost. All the data is transparently available via Etherscan where you can observe the tokens on the balances of contracts with verified source codes that have no functions to extract these tokens.

### 2. Propose solution options

1. The most common case is when a user is sending tokens to the address of that exact tokens smart-contract. Implementing a simple check within the logic of the `transfer` function would prevent this `require(address _to != address(this), "Sending to token contract");` 
2. When deploying smart-contracts on Ethereum always assume that your contract can receive ERC-20 tokens even if it is not supposed to. It is a good practice to implement an extraction function that would allow tokens to be extracted.
3. Consider using alternative standards. 

It is possible to restrict the logic of `transfer` function so that it wouldn't be possible to deliver tokens to smart-contracts with it but this would break backwards compatibility with some multisig wallets that are supposed to be ERC-20 compatible as well as Uniswap pools.

### Additional context

_No response_

- [X] Yes
- [ ] No

gorbatiukcom:
## Description

I've updated the description of the **ERC-20 token reception issue block** on the **ERC-20 TOKEN STANDARD** page. 

## Related Issue
Closes https://github.com/ethereum/ethereum-org-website/issues/13218

### DIFF
--- a/public/content/developers/docs/standards/tokens/erc-20/index.md
+++ b/public/content/developers/docs/standards/tokens/erc-20/index.md
@@ -151,6 +151,8 @@ print("Addr Balance:", addr_balance)
 
 ### ERC-20 token reception issue {#reception-issue}
 
+**As of 06/20/2024 at least $83,656,418 worth of ERC-20 tokens were lost due to this issue. Note that a pure ERC-20 implementation is prone to this problem unless you implement a set of additional restrictions on top of the standard as listed below.**
+
 When ERC-20 tokens are sent to a smart contract that is not designed to handle ERC-20 tokens, those tokens can be permanently lost. This happens because the receiving contract does not have the functionality to recognize or respond to the incoming tokens, and there’s no mechanism in the ERC-20 standard to notify the receiving contract about the incoming tokens. The main ways this issue takes form is through:
 
 1.	Token transfer mechanism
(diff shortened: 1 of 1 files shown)


wackerow:
@all-contributors please add @gorbatiukcom for content

PR #13534, opened 2024-07-29
[DO NOT MERGE] - Test Chromatic

minimalsm:

### DIFF
--- a/src/components/Card/index.tsx
+++ b/src/components/Card/index.tsx
@@ -16,7 +16,7 @@ const Card = ({ emoji, title, description, children, ...props }: CardProps) => (
     justifyContent="space-between"
     bg="ednBackground"
     borderRadius="sm"
-    border="1px"
+    border="2px"
     borderStyle="solid"
     borderColor="lightBorder"
     p="6"


TylerAPfledderer:
@minimalsm @pettinarip 

Error throwing in Storybook at `BaseLayout` and `FeedbackWidget`:
```bash
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.

Check the render method of `FixedDot`.
```

Pinpointed the failing build down to the feedback widget, and specifically the [rendering of FeedbackGlyphIcon](https://github.com/ethereum/ethereum-org-website/blob/93614536f4d09d3de34408fac827798a300256c5/src/components/FeedbackWidget/FixedDot.tsx#L35-L37)

Simply commenting out this icon successfully renders the component.

TylerAPfledderer:
@pettinarip @minimalsm currently the `FeedbackGlyphIcon` import is a direct import from an SVG file. Possible fix here is to explicitly create a component rendering the SVG contents.

Simply doing it will still cause the error, but after:
```ts
FeedbackGlyphIcon.displayName = "FeedbackGlyphIcon"
``` 

the rendering should work with no error.

pettinarip:
@TylerAPfledderer added a PR to fix this issue https://github.com/ethereum/ethereum-org-website/pull/13570

wackerow:
@minimalsm any updates here?

PR #13560, opened 2024-08-01
refactor: connect ShadCN `Link` to ShadCN `ButtonLink`

TylerAPfledderer:
Syncs the updated `Link` and `ButtonLink` components with ShadCN.

- With Chakra now removed, the `text-background` utility in the "solid" style variant no longer needs the `important` flag
- Override of the text-decoration is still needed due to explicit declaration of the property in the global styles.
- Left `activeClassName=""` as this override might still be needed for active page links.


tailwind/ui/__stories__/Button.stories.tsx
```diff
@@ -118,7 +117,7 @@ export const OverrideStyles: Story = {
           <MdNightlight />
         </Button>
         <ButtonLink
-          className="rounded-full px-0 py-0"
+          className="min-h-0 rounded-full px-0 py-0"
```
TylerAPfledderer, line 120: @pettinarip I wonder if we can do away with the "Override Styles" story. It's now very clear that any incoming utility class from a parent component should override what is existing in the child component.

pettinarip: Yeah, definitely.

pettinarip: Maybe replace it with a simple ButtonLink story? I think that component is an important one and we use it a lot. I would prefer to have at least 1 story for it. What do you think?

TylerAPfledderer: That I will do!

TylerAPfledderer:
@pettinarip replaced the Override story with one for ButtonLink (separate story file)

tailwind/ui/__stories__/ButtonLink.stories.tsx
```diff
@@ -0,0 +1,19 @@
...
+const meta = {
+  title: "Atoms / Form / ShadCn Buttons",
+  component: ButtonLinkComponent,
+} satisfies Meta<typeof ButtonLinkComponent>
+
+export default meta
+
+export const ButtonLink: StoryObj<typeof meta> = {
+  args: {
+    linkProps: {
```
pettinarip, line 14: Hey not sure about this API for the ButtonLink. I think we should flip the props. We should prioritize the link props over the buttons.

```
// this
<ButtonLink href="..." buttonProps={...} />

// rather than
<ButtonLink linkProps={{ href: "..." }} />
```

That would also align with the previous implementation.

TylerAPfledderer: Agree. Will flip it!

tailwind/ui/buttons/ButtonTwoLines.tsx
```diff
@@ -63,9 +63,20 @@ const ButtonTwoLines = ({
   )
 
   if (props.componentType === "link") {
+    const { buttonProps, ...rest } = props
     return (
-      <ButtonLink className={commonClassStyles} size={size} {...props}>
-        <ChildContent {...props} size={size} isIconLeft={isIconLeft} />
+      <ButtonLink
+        className={commonClassStyles}
+        // size={size}
```
pettinarip, line 70: Did comment this out to skip a type error that was breaking the build. I see this as a minor issue for now since this component is not currently used. We can tackle the issue in a separate PR.

I also think that `<ButtonTwoLines componentType="link" />` is not working. Something we can look into later.

pettinarip (approved):
LGTM! good job @TylerAPfledderer  💪🏼

tailwind/ui/buttons/ButtonTwoLines.tsx
TylerAPfledderer: With `componentType`, the type guarding is not working as expected. Need to re-evaluate the approach.

With `size`, it is housed in `buttonProps` when passing props to `ButtonLink`. Not going to try and make the approach any more strict, but can simply run a merge props of `buttonProps` with `size`.

```tsx
if (props.componentType === "link") {
  const { buttonProps, ...rest } = props

  const mergedButtonProps = merge(buttonProps, { size })
  return (
    <ButtonLink
      className={commonClassStyles}
      buttonProps={mergedButtonProps}
      {...rest}
    >
      <ChildContent
        {...rest}
        size={size}
        isSecondary={buttonProps?.isSecondary}
        isIconLeft={isIconLeft}
      />
    </ButtonLink>
  )
}
```
What do you think?

pettinarip: I think we should follow the same approach we took with the original ButtonLink. In this case, I think we should return just a button to simplify everything. Then, in case you need a buttonlink, the user/dev could do the same we are doing with the buttonlink.

```ts
// always a button
<ButtonTwoLines size=".." variant="..." />

// in case you want a buttonlink => compose them with asChild (same as with ButtonLink)
<ButtonTwoLines asChild>
  <BaseLink> ...
</Button>
```

tailwind/ui/Link.tsx
pettinarip, line 1: 👍🏼

pettinarip (approved):

tailwind/ui/buttons/ButtonTwoLines.tsx
TylerAPfledderer: @pettinarip Good with me! Getting too DRY here, and it's showing its cracks

PR #13570, opened 2024-08-03
Fix svg imports in SB

pettinarip:
## Description

Adds the svgr loader in SB to handle svg files as with Nextjs and updates the configs and types.

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -47,6 +47,9 @@ const config: StorybookConfig = {
     },
   },
   webpackFinal: async (config) => {
+    config.module = config.module || {}
+    config.module.rules = config.module.rules || []
+
     if (config.resolve) {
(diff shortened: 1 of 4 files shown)


.storybook/main.ts
pettinarip, line 1: https://storybook.js.org/docs/get-started/frameworks/nextjs#custom-webpack-config

next.config.js
pettinarip, line 1: https://react-svgr.com/docs/next/

This config will now also support `.svg?url` import syntax to be used with the `Image` component.

TylerAPfledderer:
@pettinarip Chromatic failure here is now expected; verified that the feedback widget model is missing its `data-testid` for the failed snapshot (might've been removed by mistake when migrating to ShadCN)

corwintines (approved):

PR #13575, opened 2024-08-04
Text unbold

### LINKED ISSUE #13499: UI Bug report: menu item remove bold
### Describe the bug

Seeing mobile menu navigation, one subsection called "emerging use cases" has subtext in bold, which it should not be. The normal text should be used instead.

### To reproduce

1. Go to https://ethereum.org/en/
2. Click on the menu on **Mobile device**
3. Select "Use"
4. Select "use cases"
5. See error - "Emerging usecases" the subtext has bold and it should not

### Expected behavior

Remove bold on subtext

### Screenshots

Screenshot 2024-07-26 at 12 25 32

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes
- [X] No

sanjanaynvsdl:
This PR resolves #13499


src/components/Nav/Mobile/LvlAccordion.tsx
```diff
@@ -82,7 +82,7 @@ const LvlAccordion = ({
                     {label}
                   </Text>
                   <Text
-                    fontWeight="regular"
+                    fontWeight="normal"
```
TylerAPfledderer, line 85: This did need correction, but the issue in question is coming from the [other return render farther down the component](https://github.com/ethereum/ethereum-org-website/blob/6ecf1d2d53f8e21ac56bb6d29870123ec517acb6/src/components/Nav/Mobile/LvlAccordion.tsx#L140) that also needed the correct value for `fontWeight`.

Currently there is no change to the text weight.

src/pages/get-eth.tsx
```diff
@@ -368,7 +368,7 @@ const GetEthPage = ({
           </Text>
           <Text>{t("page-get-eth-dexs-desc-3")}</Text>
           <Text>{t("page-get-eth-need-wallet")}</Text>
-          <ButtonLink href="/wallets/find-wallet/">
+          <ButtonLink href="/wallets/find-wallet/" mb={8}>
```
TylerAPfledderer, line 371: Revert this. This is not related to this issue in question.

```suggestion
          <ButtonLink href="/wallets/find-wallet/">
```

Nice catch though! Please create a new PR for this hotfix. 😄

TylerAPfledderer (changes requested):
Hi @sanjanaynvsdl! Thank you for taking on this correction. 🎉 

I have only a couple of thoughts here.

src/components/Nav/Mobile/LvlAccordion.tsx
wackerow: @sanjanaynvsdl Agreed, keep in mind there are two returns here, conditioned on whether it's an actual link, or a submenu with deeper items. The one that was problematic in @konopkja's post was the submenu, which is on Line 140. 

All that being said, I'm actually not even able to reproduce this bug in production right now on a few devices. Also, [this file is currently being revamped via migration to tailwind](https://github.com/ethereum/ethereum-org-website/pull/13449/files#diff-5d04d66ce0feb057f40ccd79221664493d4019bb0a7295c144e9059e652aef53) in a simultaneous PR.

I can merge this current fix in, but I don't think it's worth more effort than that since the above PR is likely to be in by next deploy.

wackerow (approved):

PR #13582, opened 2024-08-05
Fix glossary tooltip links

pettinarip:
Currently, our react pages (https://ethereum.org/en/what-is-ethereum/) or more specifically, the places where we use the Translation component, do not use the GlossaryTooltip to display the terms. The reason was that we had a circular import issue.

## Description
Adds back the `TooltipLink` to be used in the `Translation` component and fixes the circular dep issue by overriding the `a` transform used inside the `GlossaryTooltip`.


wackerow (approved):
🔥 Nice job! Thanks @pettinarip!

PR #13597, opened 2024-08-07
Shadcn migration - `/get-eth` page

pettinarip:
Migrates the `/get-eth` page to Tailwind.

## Preview link
https://deploy-preview-13597--ethereumorg.netlify.app/en/get-eth/


netlify.toml
```diff
@@ -11,6 +11,7 @@
 
 [build.environment]
   NEXT_FORCE_EDGE_IMAGES="true"
+  BUILD_LOCALES="en"
```
pettinarip, line 14: - [ ] Remove

src/pages/get-eth.tsx
```diff
@@ -182,309 +150,243 @@ const GetEthPage = ({
...
-            </Text>
-          </Text>
+            </em>
+          </p>
           <InfoBanner emoji=":wave:" shouldCenter mt={8}>
             {t("page-get-eth-new-to-eth")}{" "}
             <InlineLink href="/eth/">
               {t("page-get-eth-whats-eth-link")}
             </InlineLink>
           </InfoBanner>
```
wackerow, line 263: image

Related to the `InfoBanner` component, noting our contrast ratio has taken a hit with links on the default banner variant background.

Not a blocker for this PR

src/pages/get-eth.tsx
```diff
@@ -182,309 +150,243 @@ const GetEthPage = ({
...
+                {t("page-get-eth-dexs-desc-2")}{" "}
+                <InlineLink href="/smart-contracts">
+                  {t("page-get-eth-smart-contract-link")}
+                </InlineLink>
+              </p>
+              <p>{t("page-get-eth-dexs-desc-3")}</p>
+              <p>{t("page-get-eth-need-wallet")}</p>
+              <ButtonLink href="/wallets/find-wallet/">
+                {t("page-get-eth-get-wallet-btn")}
+              </ButtonLink>
```
wackerow, line 293: image

may need a `w-fit` on this if we want it to match production

image

```suggestion
              <ButtonLink href="/wallets/find-wallet/" className="w-fit">
                {t("page-get-eth-get-wallet-btn")}
              </ButtonLink>
```

src/pages/get-eth.tsx
```diff
@@ -182,309 +150,243 @@ const GetEthPage = ({
...
+                  {t("page-get-eth-your-address")}
+                </h3>
+                <p>{t("page-get-eth-your-address-desc")}</p>
+                <div className="mb-6 flex select-none flex-col-reverse justify-between rounded bg-[#191919] p-2 lg:flex-row">
+                  <p className="mb-0 font-monospace text-xs text-white">
+                    0x0125e2478d69eXaMpLe81766fef5c120d30fb53f
+                  </p>
+                  <p className="text-fail300 mx-4 mb-0 text-sm uppercase">
+                    {t("page-get-eth-do-not-copy")}
+                  </p>
```
wackerow, line 356: image

image

I believe we want `text-error` here (`fail300` will not work in our current TW setup)

```suggestion
                  <p className="text-error mx-4 mb-0 text-sm uppercase">
                    {t("page-get-eth-do-not-copy")}
                  </p>
```

src/styles/global.css
```diff
@@ -29,6 +29,12 @@
       rgba(84, 132, 234, 0.2) 51.56%,
       rgba(58, 142, 137, 0.2) 100%
     );
+    --radial-gradient: radial-gradient(
```
wackerow, line 32: We should use a less generic name here

wackerow (changes requested):
Nice job @pettinarip, great start here... left a couple comments highlighting a few changes and potential improvements.

I see some layout shifts compared to production, which look good to me but will let @nloureiro and/or @konopkja chime in as well.

src/pages/get-eth.tsx
wackerow: Have some work done locally trying to update the InfoBanners... will try to get a PR up within the next week

pettinarip: The idea we had with these InfoBanner's was to replace them with [the new Alerts](https://dev--63b7ea99632763723c7f4d6b.chromatic.com/?path=/story/molecules-action-feedback-alerts--variants). Will do that in the scope of this PR but we could replace all of the InfoBanners perhaps.

src/styles/global.css
pettinarip: Good call, do you know if we have a gradient from the DS that we could use in this case? will take a look and try to delete this one.

src/pages/get-eth.tsx
wackerow: > Will do that in the scope of this PR but we could replace all of the InfoBanners perhaps.

@pettinarip _In_ the scope of this PR? There are over 500 instances of this component, touching almost 500 files... 😆 I would strongly advise we not make that part of this PR

wackerow: hah I see your Discord response: out of scope

wackerow (approved):
Looks great!! Nice job @pettinarip... Some discussion about the font size within the old InfoBanner callouts... PR is currently aligned with the DS, so going to approve and pull this in. Can discuss those changes separately.

🎉

PR #13601, opened 2024-08-07
Shadcn migration - image

pettinarip:
## Description

Exports the default image component to be used in the UI transition.

### DIFF
--- a/src/components/Image.tsx
+++ b/src/components/Image.tsx
@@ -41,8 +41,18 @@ const DefaultNextImage = (props: ImageProps) => {
 }
 
 /**
- * TODO: replace this component with import { Image } from "@chakra-ui/next-js"
- * once https://github.com/vercel/next.js/issues/52216 is fixed
+ * TODO: Rename this component to `Image` once all components are using Tailwind
+ */
+export const TwImage = (props: NextImageProps) => (
+  <DefaultNextImage {...props} />
+)
+
+/**
+ * DEPRECATED: Use `TWImage` instead
+ *
+ * TODO: Remove this component once all components are using Tailwind
+ *
+ * @deprecated
  */
 export const Image: ChakraComponent<"img", NextImageProps> = chakra(
   DefaultNextImage,


src/components/Image.tsx
```diff
@@ -41,8 +41,18 @@ const DefaultNextImage = (props: ImageProps) => {
 }
 
 /**
- * TODO: replace this component with import { Image } from "@chakra-ui/next-js"
- * once https://github.com/vercel/next.js/issues/52216 is fixed
+ * TODO: Rename this component to `Image` once all components are using Tailwind
+ */
+export const TwImage = (props: NextImageProps) => (
+  <DefaultNextImage {...props} />
```
wackerow, line 47: @pettinarip How do we want to handle any image sources that are external to the repo? Events, for example, will now be displaying external hosted images. To my knowledge, we have have a few options for rendering these:
- Allowlist the image domain and use Next Image (which we wouldn't necessarily know)
- Download the images in a separate task and import directly
- Or just use the `img` element without the Next wrapper

Any thoughts on this, and do you think we should add a touch of logic here to differentiate the image source, and avoid the Next Image component if needed? Or should we just use `img` directly in those cases, and not bother importing this component? We currently get `eslint @next/next/no-img-element` warnings when doing this.

pettinarip: Good question, I don't have an easy answer. I guess it depends.

_Before I continue, I don't think this needs to be part of the scope of this PR. This PR just exports the default next image component to keep the current implementation and help with the UI transition._

- If we know in advance where these images are coming from, I would lean towards adding these domains to the allowlist. But I would only do that in the case where we only have a few domains and we don't need to maintain too much (I would not expect that to happen).
- Using the `img` seems like the easiest solution, but we lack optimizations & security.
- and downloading images will require some prep work + maintaining the images...

So I would choose one of the following options:
1. I would choose the first option if the domains are predefined and don't change.
2. I would store the images in a separate cloud service and consume them from there (adding that to the allowlist).

wackerow (approved):

PR #13607, opened 2024-08-08
Chromatic workflow update

pettinarip:
## Description

Updates the chromatic workflow using the `tj-actions/changed-files` action approach (https://www.chromatic.com/docs/github-actions/#skip-chromatic-in-a-monorepo-when-no-ui-files-have-changed).

Approach to be tested and see if we can required this as a status check in the branch protection rules and, at the same time, without requiring this check in _all_ the PRs.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,45 +1,41 @@
 # .github/workflows/chromatic.yml
-
-# Workflow name
 name: Chromatic Publish and Testing
 
-# Event for the workflow
-on:
-  pull_request:
-    types:
-      - opened
-      - synchronize
-      - ready_for_review
-    paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/chromatic.yml
```diff
@@ -1,45 +1,41 @@
...
+          cache: npm
+          cache-dependency-path: ./package.json
+          node-version: 20
+      - name: 📥 Install deps
         run: yarn
-      - name: Publish to Chromatic
-        # 👇 Adds Chromatic as a step in the workflow
-        uses: chromaui/action@v1
+      - name: Build Storybook
+        run: npm run build-storybook
```
TylerAPfledderer, line 33: Do you think the `build-storybook:chromatic` script currently in the `package.json` is no longer needed?

PR #13630, opened 2024-08-13
Shadcn migration - tooltip

pettinarip:
## Description

Implement the Tooltip using the shadcn popover to be used in desktop and mobile devices.


src/components/Glossary/GlossaryTooltip/index.tsx
```diff
@@ -1,8 +1,6 @@
 import React, { ReactNode } from "react"
 import { useRouter } from "next/router"
-import { Box, Text, VStack } from "@chakra-ui/react"
 
-import Heading from "@/components/Heading"
```
Wachira3712, line 5: gatsby-node.js

wackerow (changes requested):
https://github.com/user-attachments/assets/ec9c6c3a-0798-4df6-9a35-a3033570a4af

Looking good @pettinarip, but seeing this bug where we're unable to hover over the tooltip now without it closing on us.

pettinarip:
> Looking good @pettinarip, but seeing this bug where we're unable to hover over the tooltip now without it closing on us.

Uhmm good catch @wackerow, yes, that sucks. Trust me, I tested this before haha and it worked as expected. Will look into it soon. Thanks

wackerow (approved):
Looks good! Working just like production now, thanks @pettinarip!

PR #13641, opened 2024-08-15
refactor: consolidate css imports into global.css

wackerow:
## Description
- Consolidates all css imports into the `global.css` file
- Moves other global `css` imports as an `@import` statement within `global.css`

cc: @pettinarip Not sure if we would also like to revert the recent updates to the ESLint rule changes with this PR. Happy to, just drop a note if you'd prefer one way or the other.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -42,7 +42,7 @@
           // Other relative imports. Put same-folder imports and `.` last.
           ["^\\./(?=.*/)(?!/?$)", "^\\.(?!/?$)", "^\\./?$"],
           // Style imports.
-          ["^.+s?css$"],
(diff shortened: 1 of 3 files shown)


pettinarip:
> ## Description
> * Consolidates all css imports into the `global.css` file
> * Moves other global `css` imports as an `@import` statement within `global.css`
> 
> cc: @pettinarip Not sure if we would also like to revert the recent updates to the ESLint rule changes with this PR. Happy to, just drop a note if you'd prefer one way or the other.

That would be ideal @wackerow. This PR looks good though, approving but waiting for that final change if possible.

pettinarip (approved):
🚀

PR #13642, opened 2024-08-15
fix: keyboard shortcut conflict

wackerow:
## Description
- Patches the keyboard shortcut for the Language Picker to exclude usage of the meta (cmd/win) or ctrl keys when using the `\` shortcut
- Avoids <kbd>cmd</kbd>/<kbd>ctrl</kbd>-<kbd>\\</kbd> from opening lang menu; reserved for color mode change


pettinarip (approved):
LGTM, thanks @wackerow

PR #13643, opened 2024-08-15
[Storybook] fix(useQuizWidget): run chromatic check for shuffling question set

TylerAPfledderer:
## Description

Fixes issue where the Chromatic Snapshot for the quiz modal kept updating because the first question rendered was always random. Uses the `isChromatic()` check to determine if `shuffle()` is triggered. The `layer-2` quiz key is used, therefore the first question to appear should always be `Layer 2 blockchain networks are for:`

The shuffle will still run when viewing the story in the daskboard.

The question set should not be shuffled when rendered in the Chromatic snapshot. Snapshots should at least be stable if they are not rendering pure functions with static prop data.


pettinarip (approved):
Nice! would love to not have to add this in the implementation detail but that would require a deeper refactor.

Thanks @TylerAPfledderer

PR #13647, opened 2024-08-15
Shadcn migration - desktop nav menu & search modal fixes

pettinarip:
## Description

Styles the desktop nav menu using Tailwind & refactors the search modal logic.


src/components/Search/SearchButton.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from "react"
-import { useTranslation } from "react-i18next"
```
wackerow, line 2: ??? This was still using react-i18next?

wackerow (commented):
Initial observations:

Animations are a bit more jumpy, and flashy... this could bother people with certain sensitivities. Recording shows production first, then PR:

https://github.com/user-attachments/assets/1ef8bd74-5911-450f-8ab9-4c59970e85cf

---

Prod:
image
PR:
image

- Gap is now present between the selected menu item and the next column, when before were connected
- Search previously collapsed to the magnifying glass icon, while now it collapses to a box with just a keyboard command

pettinarip:
> Animations are a bit more jumpy, and flashy... this could bother people with certain sensitivities. Recording shows production first, then PR:

@wackerow thanks for the early feedback. Good catch, didn't notice that while doing the transition, will take a look and see what Im missing.

> * Gap is now present between the selected menu item and the next column, when before were connected

Correct, I still need to implement this case.

> * Search previously collapsed to the magnifying glass icon, while now it collapses to a box with just a keyboard command

Good observation, I must have missed something here. Thanks

src/components/Search/SearchButton.tsx
pettinarip: uhmm yes! and it created a hydration issue. It may have been an automated import

wackerow:
Awesome @pettinarip !

https://github.com/user-attachments/assets/2d179724-5459-44e0-984e-db6e4946c30d


image

Tiny thing I'm noting, which I'll let @nloureiro or @konopkja weigh in on, where the background of the sub-section fades in, the parent/trigger background overlaps a little which didn't happen before. It's quick and subtle, just noting

---

Second thing, just noting we lost our "hover" styling for the magnifying glass / search component.

src/components/Nav/types.ts
```diff
@@ -9,7 +9,10 @@ type LinkXorItems = LinkOnly | ItemsOnly
 export type NavItem = {
   label: string
   description: string
-  icon?: IconType | ((props: IconProps) => JSX.Element)
+  icon?:
+    | IconType
+    | FC<SVGProps<SVGElement>>
+    | ((props: IconProps) => JSX.Element)
```
wackerow, line 15: Do we still need to account for this Chakra `((props: IconProps) => JSX.Element)` type? Or can we rm this along with the import?

src/components/Nav/useNav.ts
```diff
@@ -21,7 +21,7 @@ import {
 import { PiFlask, PiUsersFourLight } from "react-icons/pi"
 import { useColorMode } from "@chakra-ui/react"
```
wackerow, line 22: I assume we're leaving this for now to allow Chakra to handle the color-mode toggling still?

src/components/Nav/Desktop/index.tsx
```diff
@@ -20,20 +22,12 @@ const DesktopNavMenu = ({ toggleColorMode }: DesktopNavMenuProps) => {
   const { locale } = useRouter()
   const languagePickerRef = useRef<HTMLButtonElement>(null)
 
-  const ThemeIcon = useColorModeValue(<MdBrightness2 />, <MdWbSunny />)
+  const ThemeIcon = useColorModeValue(MdBrightness2, MdWbSunny)
```
wackerow, line 25: Not at all against this, but just noting an alternative way this can be done as we migrate to TW:

```tsx
<MdBrightness2 className="dark:hidden" />
<MdWbSunny className="not-[dark]:hidden" />
```

Can see where both approaches could come in handy, not sure I have a strong preference or desire to standardize it.

wackerow (approved):
This looks great @pettinarip! Left a couple small comments, but don't see those as blockers to this PR (can be addresses separately). 

Pulling in!

src/components/Nav/Desktop/index.tsx
pettinarip: Great idea! I haven't thought about that option, I like the no-js dependency. I'll test it out and let you know.

src/components/Nav/types.ts
pettinarip: Ahh good catch, didn't notice that. Will remove it in a new PR.

PR #13658, opened 2024-08-15
Fix regex to match only numbers

pettinarip:
Currently, if you press the `F1`-`F12` keys, you get the following error message: `Uncaught DOMException: Failed to execute 'querySelector' on 'Document': '#main-navigation li:nth-of-type(NaN) button' is not a valid selector.`

## Description

Fixes the regex used to navigate the nav menu to match only numbers.


minimalsm:
@coderabbit review

minimalsm:
@coderabbitai review

minimalsm (approved):

wackerow (approved):
Interesting edge case, thanks @pettinarip!

PR #13661, opened 2024-08-16
Use `dark` class for color mode selector

wackerow:
## Description
- Refactors to use `.dark` class for color mode management
- Fixes syncing of color theme between in-app toggling and system color preference changes

## Preview URL
- https://deploy-preview-13661--ethereumorg.netlify.app/

## Related Issue
Tailwind migration

### DIFF
--- a/src/components/Nav/useNav.ts
+++ b/src/components/Nav/useNav.ts
@@ -1,3 +1,4 @@
+import { useEffect } from "react"
 import { useTranslation } from "next-i18next"
 import { useTheme } from "next-themes"
 import {
@@ -26,14 +27,10 @@ import { trackCustomEvent } from "@/lib/utils/matomo"
 
 import type { NavSections } from "./types"
 
-import useColorModeValue from "@/hooks/useColorModeValue"
-
 export const useNav = () => {
   const { t } = useTranslation("common")
-  const { theme, setTheme } = useTheme()
+  const { setTheme, resolvedTheme, systemTheme } = useTheme()
   const { setColorMode } = useColorMode()
-
-  const colorToggleEvent = useColorModeValue("dark mode", "light mode") // This will be inverted as the state is changing
 
   const linkSections: NavSections = {
     learn: {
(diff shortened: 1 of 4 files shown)


src/components/Nav/useNav.ts
```diff
@@ -26,15 +27,11 @@ import { trackCustomEvent } from "@/lib/utils/matomo"
 
 import type { NavSections } from "./types"
 
-import useColorModeValue from "@/hooks/useColorModeValue"
-
 export const useNav = () => {
   const { t } = useTranslation("common")
-  const { theme, setTheme } = useTheme()
+  const { setTheme, resolvedTheme, systemTheme } = useTheme()
```
wackerow, line 32: Removed usage of `theme` which was not always responding appropriately to system color preference changes.

In it's place, `resolvedTheme` is a more reliable "light" or "dark" string that the theme will resolve to. Also, `systemTheme` taps into the color mode preference of the system, and updates on change.

src/components/Nav/useNav.ts
```diff
@@ -465,13 +462,27 @@ export const useNav = () => {
...
+
   const toggleColorMode = () => {
-    setTheme(theme === "dark" ? "light" : "dark")
-    setColorMode(theme === "dark" ? "light" : "dark")
+    // resolvedTheme: "light" | "dark" = Current resolved color mode from useTheme
+    const targetTheme = resolvedTheme === "dark" ? "light" : "dark"
+    // If target theme matches the users system pref, set ls theme to "system"
+    const lsTheme = targetTheme === systemTheme ? "system" : targetTheme
+
+    setTheme(lsTheme)
```
wackerow, line 480: When observing this in the Application dev tools, the "system" is automatically parsed (by the embedded `next-theme` I believe*) to "light" or "dark". Using this approach seems to make sure the `html[class]` is updated by the NextThemeProvider.

<details><summary>*embedded `next-theme` script</summary>
<p>

```js
!(function () {
  try {
    var d = document.documentElement,
      c = d.classList;
    c.remove('light', 'dark');
    var e = localStorage.getItem('theme');
    if ('system' === e || (!e && true)) {
      var t = '(prefers-color-scheme: dark)',
        m = window.matchMedia(t);
      if (m.media !== t || m.matches) {
        d.style.colorScheme = 'dark';
        c.add('dark');
      } else {
        d.style.colorScheme = 'light';
        c.add('light');
      }
    } else if (e) {
      c.add(e || '');
    }
    if (e === 'light' || e === 'dark') d.style.colorScheme = e;
  } catch (e) {}
})();
```

</p>
</details>

pettinarip (changes requested):
@wackerow a few things related to this PR and the color theme mechanism

1. I pushed a new PR https://github.com/ethereum/ethereum-org-website/pull/13676 to sync the color themes between chakra and next-themes. Maybe this PR will clarify a bit what is going on between the two color theme syncs.

2. Left a comment/question/opinion here about the manual system color selection.

3. On choosing the `class` method, would you explain again why we would prefer this over  the `data-` attr? question because chakra will automatically detect the latter but not the class and that might save us from manually sync the color themes.

=====
Saw this issue in the preview deploy
1. Load the page in incognito
2. Switch your system color mode
3. Visual diff
image

I guess this issue is related to the usage of  `theme` instead of `resolvedTheme` (fixed in #13663) as this branch is outdated.

wackerow:
Closing out; avoiding `.dark` solution at th time as it doesn't seem to provide benefit over existing selector. We'll also plan to add a "system" option for color modes instead of one-off-solutions.

PR #13672, opened 2024-08-16
[Storybook] fix(.storybook/preview): sync theme from global with storage key

TylerAPfledderer:
## Description

Fixes issue in Storybook where the update of color mode in the global context was not being recognized in the `next-themes` provider, as the storage key the provider uses was not updating.

Creates a custom `withNextthemes` decorator that initializes the global theme and sends it to `next-themes` via `setTheme()`. This should ensure the default theme gets to the provider on first load.

See [Writing a [next-themes] custom decorator](https://github.com/storybookjs/storybook/blob/next/code/addons/themes/docs/api.md#writing-a-custom-decorator)


TylerAPfledderer:
@pettinarip this fixes the issue in Storybook, but does not address the Chromatic snapshot.

I guess we can do:
```tsx
export const ThemeChanger = ({ theme }: { theme: "light" | "dark" }) => {
  const { setTheme } = useTheme()

  useEffect(() => {
    console.log("Syncing color mode for SB and next-themes to: ", theme)
    const updatedTheme = isChromatic() ? "light" : theme
    setTheme(updatedTheme)
  }, [setTheme, theme])

  return null
}
```

Doing `isChromatic()` check and force `light` mode because we currently are not making snapshots specifically for `dark` mode. But if making dark mode snapshots is to be done at some point, then this would be only a hotfix (barring that this even works)

I currently do not know why Chromatic thinks it's in dark mode though.

TylerAPfledderer:
@pettinarip hit a blocker with Chromatic check. Not sure if I can do anything about it

pettinarip:
> @pettinarip hit a blocker with Chromatic check. Not sure if I can do anything about it

A re-run did it. A dep fetch had failed.

pettinarip (approved):
LGTM, thanks @TylerAPfledderer

PR #13686, opened 2024-08-19
fix(Translation): use runtime import for `TooltipLink`

TylerAPfledderer:
In the `Translation` component, use runtime import with `require()` for the `TooltipLink` on transform.

Fixes issue where circular dependency was triggering webpack errors in Storybook, despite no error or warning throwing in NextJS build.


pettinarip:
This may have been fixed by #13812

TylerAPfledderer:
> This may have been fixed by #13812

@pettinarip I believe what you did in that PR is the best approach here.

pettinarip:
Thanks @TylerAPfledderer closing it then

PR #13691, opened 2024-08-20
Shadcn migration - HubHero

saurabhburade:
## Description

 Migrate HubHero to tailwind


saurabhburade:
As of now using Stack from `ui/flex` needs explicit spacings
- Use `space-y` as considering it to be the column
- Use `space-x` as considering it to be the row

Container widths are differently defined than default tw widths
- Used in chakra `--eth-sizes-container-md: 768px`
- Tailwind `max-w-3xl : 768px`

TylerAPfledderer:
> As of now using Stack from `ui/flex` needs explicit spacings
> - Use `space-y` as considering it to be the column
> - Use `space-x` as considering it to be the row
> 
> Container widths are differently defined than default tw widths
> - Used in chakra `--eth-sizes-container-md: 768px`
> - Tailwind `max-w-3xl : 768px`

@saurabhburade can you clarify why we need to explicitly define the spacing in this way? 

Note that those utility classes use `margin` instead of the `gap` properties. And the `Stack` and `VStack` components are for vertically-aligned flex boxes, and the `HStack` for horizontally stacked boxes.

We will need to define how this should be handled for consistency with other contributors. I would think that if we are using flexbox, that we should favor the `gap` property.

src/components/Hero/HubHero/index.tsx
```diff
@@ -38,55 +39,53 @@ const HubHero = ({
...
+          ) : (
+            <h1 className="text-4xl lg:text-5xl">{header}</h1>
+          )}
+
+          <p className="text-lg">{description}</p>
         </Stack>
         <Stack
-          direction={{ base: "column", md: "row" }}
-          justify={{ md: "center", xl: "start" }}
-          spacing="4"
```
saurabhburade, line 81: Hi @TylerAPfledderer, In chakra, in this case if `spacing={4}` mentioned, this is converted to `space-y-4` in column direction and `space-x-4` in row direction.

While converting these responsive cases,  we need to explicitly look for direction for `space` condition while converting it to tailwind class

saurabhburade:
> We will need to define how this should be handled for consistency with other contributors. I would think that if we are using flexbox, that we should favor the `gap` property.


Yeah, this seems like spacings are treated as gap property in Chakra Stack
https://github.com/chakra-ui/chakra-ui/blob/37dae70264921637008bb0780f5559dc5d2ed22f/packages/components/src/stack/stack.tsx#L147

saurabhburade:
@pettinarip @TylerAPfledderer We need to define the container sizes in Tailwind, as of now we have default width sizes from Tailwind.

```
container-sm: 640px;
container-md: 768px;
container-lg: 1024px;
container-xl: 1280px;
container-2xl: 1536px
```

saurabhburade:
@pettinarip @TylerAPfledderer 
- We need to add container size classes, as of now we are using default width sizes from tailwind.
 ```
    --eth-sizes-container-sm: 640px;
    --eth-sizes-container-md: 768px;
    --eth-sizes-container-lg: 1024px;
    --eth-sizes-container-xl: 1280px;
    --eth-sizes-container-2xl: 1536px;
 ```
- Tailwind container class is the default in responsive sizes https://tailwindcss.com/docs/container

src/components/Hero/HubHero/index.tsx
```diff
@@ -38,55 +39,51 @@ const HubHero = ({
...
             {title}
-          </Text>
+          </h1>
         ) : null}
         <Stack
-          alignSelf="center"
-          spacing={{ base: "2", md: "1" }}
-          maxW="container.md"
+          // --eth-sizes-container-md: 768px ==> tailwind max-w-3xl
+          className="max-w-3xl gap-2 self-center md:gap-1"
```
pettinarip, line 61:
```suggestion
          className="max-w-screen-md gap-2 self-center md:gap-1"
```

src/components/Hero/HubHero/index.tsx
```diff
@@ -38,55 +39,51 @@ const HubHero = ({
...
         >
-          <Heading as={title ? "h2" : "h1"} size="2xl">
-            {header}
-          </Heading>
-          <Text size="lg">{description}</Text>
+          {/* Ref: file: src/components/Hero/HubHero/index.tsx:28 */}
+          {/* 2xl font size refers to  ["4xl", null, null, "5xl"] */}
```
pettinarip, line 64: What is this ref indicating?

re: the font size comment, I don't think we need to clarify that given that the classnames are self-explanatory, or am I missing something?

src/components/Hero/HubHero/index.tsx
```diff
@@ -1,9 +1,10 @@
-import { Box, Heading, Stack, Text } from "@chakra-ui/react"
-
 import type { CommonHeroProps } from "@/lib/types"
 
 import { CallToAction } from "@/components/Hero/CallToAction"
```
pettinarip, line 3: Can we also include the migration of this component?

src/components/Hero/HubHero/index.tsx
```diff
@@ -1,9 +1,10 @@
...
 import { Image } from "@/components/Image"
```
pettinarip, line 4: We need to use the `TwImage` export from now on and adapt the styles accordingly.

pettinarip (changes requested):
@saurabhburade this looks great! left a couple of comments to migrate a few more things.

Regarding the container sizes, we have defined them [here](https://github.com/ethereum/ethereum-org-website/blob/dev/tailwind.config.ts#L15-L20) and they match the breakpoints we had in chakra theming.

src/components/Hero/HubHero/index.tsx
saurabhburade: As a workaround it works, but for 
`container-sm` should be `640px` while tailwind `w-screen-sm`  has size of `480px`

src/components/Hero/HubHero/index.tsx
saurabhburade: It is added with reference to the Chakra Heading component at `src/@chakra-ui/components/Heading.ts:21`
The comment is added to avoid confusion between heading with size 2xl from Chakra to ["4xl", "5xl"]  in tailwind classes

src/components/Hero/HubHero/index.tsx
saurabhburade: I think it's doable

TylerAPfledderer:
> > We will need to define how this should be handled for consistency with other contributors. I would think that if we are using flexbox, that we should favor the `gap` property.
> 
> Yeah, this seems like the spacing prop treated as a gap property in Chakra Stack https://github.com/chakra-ui/chakra-ui/blob/%40chakra-ui/react%402.8.0/packages/components/layout/src/stack/stack.tsx#L152

Right, it's an alias prop for `gap`. 

> Chakra v3 no longer has that prop btw, as they didn't find it semantically useful for users. So `gap` is the preferred property to use

src/components/Hero/HubHero/index.tsx
pettinarip: Oh, I get it. Ok, so this is what happens: chakra has two separate values for containers and breakpoints, while tw uses the same value for both.

So from now on, we will no longer have the `640px` value in our theme. This means that in cases where we are using `container.sm` to get the 640px, we will have to rethink why we need that specific size and refactor accordingly.

In this case, since we are setting the container size, `max-w-screen-md` is well used and makes sense because we don't want this box to be bigger than this breakpoint size.

src/components/Hero/HubHero/index.tsx
pettinarip: Right, the thing is that we are moving out of Chakra and these refs and clarifications won't matter when we finish the transition given that we will not have the `size` scale anymore.

JLYK we have migrated these sizes in our globals css file. https://github.com/ethereum/ethereum-org-website/blob/dev/src/styles/global.css#L219

pettinarip (approved):
Nice job @saurabhburade

PR #13695, opened 2024-08-21
feat(hooks): build local `useBreakpointValue` and `useMediaQuery` hooks

TylerAPfledderer:
## Description

Creates the `useBreakpointValue` and `useMediaQuery` hooks locally.

`useBreakpointValue` references the Chakra UI [useBreakpointValue](https://github.com/chakra-ui/chakra-ui/blob/main/packages/react/src/components/use-breakpoint.ts) as is simplified.

`useMediaQuery` is an amalgamation of the hooks from [usehooks-ts](https://usehooks-ts.com/react-hook/use-media-query) and from [Chakra UI](https://github.com/chakra-ui/chakra-ui/blob/main/packages/hooks/src/use-media-query.ts) to account for an array of media queries.

This should be scrutinized in knowing how simple or complex this pair need to be for the project's purposes.


tailwind/screens.ts
pettinarip, line 1: Nice!

src/hooks/useBreakpointValue.ts
```diff
@@ -0,0 +1,51 @@
+import { screens } from "../../tailwind/screens"
+
+import { useMediaQuery } from "./useMediaQuery"
+
+const breakpointMap = {
+  // Essentially to query from no width if smaller than "sm"
+  base: "0px",
```
pettinarip, line 7: 👍🏼

pettinarip (approved):
This is great @TylerAPfledderer. Works perfectly in the current instances.

PR #13696, opened 2024-08-21
chore: remove lucide-react dep

TylerAPfledderer:
## Description

Removes the `lucide-reacts` icon dependency that was added by mistake when adding components from ShadCN.


pettinarip (approved):
Ty @TylerAPfledderer

Issue #13645, opened 2024-08-15
Bug report: remove date from whitepaper page


konopkja:
### Describe the bug

We have a date component that showcases when was a page last edited, however on whitepaper page this may be confusing to users leading them to believe the whitepaper has been updated with new information.

THerefore we should remove the component from this page.

### To reproduce

1. Go to https://ethereum.org/en/whitepaper/
2. See date on top of the page

### Expected behavior

remove date (OR instead hard code static date of when the whitepaper has been published.)

### Screenshots

<img width="1101" alt="Screenshot 2024-08-15 at 11 19 45" src="https://github.com/user-attachments/assets/7c48d12f-6646-4ae1-af5b-5f92c008284f">

### Would you like to work on this issue?

- [ ] Yes
- [X] No

gks2022004:
I solve it, assign it to me please

konopkja:
> I solve it, assign it to me please

assigned, thank you!

gks2022004:
I could not find the 'whitepaper' page, to remove the date, need help😅

konopkja:
> I could not find the 'whitepaper' page, to remove the date, need help😅

https://ethereum.org/en/whitepaper/

gks2022004:
no, in the main repository, where I have to fix the issue, from the repo I get the idea of working, but the whitepaper is fetched on the hero section of WhitePaper.

konopkja:
cc @wackerow  @pettinarip for help

corwintines:
Hey @gks2022004 

The area where this is rendered can be found here https://github.com/ethereum/ethereum-org-website/blob/0e37f74304808aa6612d3e6b8c8a9f0bd186ebb8/src/layouts/Static.tsx#L125

I think all that needs to be done is conditionally render this where if the slug is /whitepaper it isnt rendered, otherwise render.

gks2022004:
Thanks, for the help

gks2022004:
@corwintines @wackerow @konopkja 
I have created the PR
https://github.com/ethereum/ethereum-org-website/pull/13687

gks2022004:
@wackerow @konopkja 
I have created the [PR](https://github.com/ethereum/ethereum-org-website/pull/13710)

PR #13710, opened 2024-08-23
fix: Remove date from whitepaper page #13645

gks2022004:
# Remove date from the whitepaper page

## Description
This pull request modifies the Static Layout component to hide the "Page last updated" text specifically on the Whitepaper page. The change involves adding a conditional check to prevent rendering the timestamp for the /whitepaper route.

The main changes include:
1. Added a condition to check both the `slug` and `asPath` before rendering the "Last Updated" text.
2. The text will now only render if the page is not the Whitepaper page.

##  Issue
 fixes https://github.com/ethereum/ethereum-org-website/issues/13645
Fixes 13645 - remove date from whitepaper


src/layouts/Static.tsx
```diff
@@ -111,13 +111,15 @@ export const StaticLayout = ({
...
               <Breadcrumbs slug={slug} />
 
-              <p
-                className="text-body-medium"
-                dir={isLangRightToLeft(locale as Lang) ? "rtl" : "ltr"}
-              >
-                <Translation id="page-last-updated" />:{" "}
-                {lastEditLocaleTimestamp}
-              </p>
+              {slug !== "/whitepaper" && !asPath.includes("/whitepaper") && (
```
corwintines, line 114:
```suggestion
              {!asPath.includes("/whitepaper") && (
```

The slug check is redundant, so going to remove that one

corwintines (approved):
Thanks @gks2022004!

PR #13723, opened 2024-08-26
Shadcn migration - PageHero

saurabhburade:
## Description

Migrate PageHero to tailwind.


src/components/PageHero.tsx
```diff
@@ -48,113 +49,93 @@ const PageHero = ({
...
                       variant={button.variant}
                       href={button.href}
-                      onClick={() =>
-                        trackCustomEvent({
-                          eventCategory: button.matomo.eventCategory,
-                          eventAction: button.matomo.eventAction,
-                          eventName: button.matomo.eventName,
-                        })
-                      }
+                      customEventOptions={{
```
pettinarip, line 91: Nice!

tailwind.config.ts
```diff
@@ -102,7 +102,10 @@ const config = {
           800: "var(--orange-800)",
           900: "var(--orange-900)",
         },
-
+        text: {
```
pettinarip: Gonna remove these colors because we now default to white in the hero texts.

src/components/PageHero.tsx
```diff
@@ -48,113 +49,93 @@ const PageHero = ({
...
+      )}
     >
-      <Box
-        maxW={{ base: "full", lg: "container.sm" }}
-        pt={{ base: isReverse ? 0 : 8, lg: 32 }}
-        pb={{ base: isReverse ? 8 : 0, lg: 32 }}
-        ps={{ base: 0, lg: 8 }}
-        me={{ base: 0, lg: 4 }}
+      <div
+        // TODO: container.sm width unavailable
```
pettinarip: Removing comment since we will not have this measure (640px) on the new theme. We can treat it now as a custom size.

pettinarip (approved):
@saurabhburade LGTM! thanks

I've pushed a commit with minor style tweaks to align with latest design decisions.

src/components/PageHero.tsx
saurabhburade: Gotcha!! So every `container.sm` is considered as a custom size of `(640px)` .

pettinarip: Yeah, we don't have that breakpoint on the theme anymore.

PR #13737, opened 2024-08-28
Update index.md

Hina-writer:
Just some thoughts ...


corwintines:
I don't understand the change here, if you have any suggestions to changing the copy instead of just comments feel free to open up a PR @Hina-writer :)

PR #13665, opened 2024-08-16
Update Infinity Wallet Metadata

InfinityWallet:
## Description

The Infinity Wallet metadata has been updated, specifically:
- ENS Support
- Brand Color
- Logo image fixed

## Related Issue
https://github.com/ethereum/ethereum-org-website/issues/13564#issuecomment-2292994967

PR #13743, opened 2024-08-29
Remove new_to_crypto flag

wackerow:
## Description
- Removes the `new_to_crypto` flag from the Infinity Wallet entry which was added inadvertently in #13665


Crytogenz:
Whom even decides what is `new_to_crypto`? Since looking at those that have half are some of the worst wallets I have ever tried. Atleast Infinity Wallet is actually made specifically for new users in crypto unlike many that have the flag at random it would seem.

corwintines (approved):

PR #13752, opened 2024-08-30
[DO NOT MERGE] - Test workflow

minimalsm:


wackerow:
@minimalsm Thoughts on this one? I think it'd be cleaner to store the "new to crypto" selection in a separate array. Should be able to just use the `name` value as the unique identifier (or just add your preferred flavor of a UID to each wallet, and use that).

This will be easier to maintain, and significantly improve "readability" as it will be very easy to spot changes to that array.

PR #13755, opened 2024-08-30
fix: hover-off transition on color toggle

wackerow:
## Description
- Adds transition to default state to enable animation when leaving hover state for color toggle nav button

https://github.com/user-attachments/assets/041bb2ad-b2e9-4a77-aafb-677d8871e8de

## Related Issue
None filed


corwintines (approved):
Much better 🎉

PR #13756, opened 2024-08-31
fix: convert js example to ts string export

wackerow:
## Description
- rm comments at top, which were being rendered in code example on homepage
- Export as string literal instead of importing `.js` as raw.

## Related Issue
image
oops =)


corwintines (approved):
Nice catch

PR #13781, opened 2024-09-02
new MEV diagram on the MEV page

nloureiro:
The new MEV diagram on the MEV page


nloureiro:
@aslikaya @LauraSinisterra I've added the image on the page

https://deploy-preview-13781--ethereumorg.netlify.app/en/developers/docs/mev/#maximal-extractable-value

Is it too big? Users need to click on it to read it. I think it's okay, but let me know what you think.

aslikaya:
Yes, it needs to be clicked to read it, I even want to zoom in :) Since the diagram is too big, the writings need to be small to fit in I guess. As far as I remember, most ethereum.org users access via mobile phones, but this is an advanced topic on builders docs so it may not be a big issue. If it is still a concern, we might need to remove some detailed info.

nloureiro:
@aslikaya, at this point, mobile and desktop access to ethereum.org are split 50/50, and the dev docs get a bit more desktop use than the rest of the pages.

This is a complex diagram, but @LauraSinisterra, can we make the text a bit bigger to avoid zooming on the fullscreen image? And if possible, to reduce some space 🙌

@aslikaya, do we still want to make a simpler diagram for the top of the page? might help to compensate the complexity of this one and could be one that the user don´t need zoom to understand

aslikaya:
@nloureiro Yes you are right, a simpler diagram would be nice at the top

nloureiro:
> @nloureiro Yes you are right, a simpler diagram would be nice at the top

Let's merge this one and then focus on the other one. ok?
Thank you for your contribution!

LauraSinisterra:
Hi @nloureiro @aslikaya 
Here's the updated diagram. I can make a simpler version for the top, is there's any information I can remove or reduce?


MEV diagram - updated

aslikaya:
I think we need to decide, 
1. if it is going to be a simpler version of the mev-boost
I actually proposed another one about mev-boost on figma earlier, simpler than this but still I don't think it is a simple diagram :) 

4. Or about the MEV itself. There are different types of MEV such as frontrunning, backrunning and sandwich attacks. 

The upper half of the MEV page is about what MEV is and its examples, the lower other half is about proposer-builder separation and its current off chain implementation (mev-boost). The most complicated part on the page, hard to understand without a diagram is definitely mev-boost.

corwintines (approved):

PR #13803, opened 2024-09-05
Tooltip updates

pettinarip:
## Description

Adjusts the max width of the container to avoid dead space in content and sets the delay duration to 200ms.


wackerow (approved):
Looks great! Fixing the issues I saw on the homepage branch, thanks!

PR #13807, opened 2024-09-05
Product table refactor

corwintines:
## Description

- Refactor out the FindWallets table to a reusable ProductTable that can be used in other use cases (networks coming up, dapps in the future)


src/components/ui/select.tsx
```diff
@@ -0,0 +1,158 @@
+import * as React from "react"
+import { Check, ChevronDown, ChevronUp } from "lucide-react"
```
corwintines, line 2: Refactor lucide out of here and use react icons

src/components/ui/select.tsx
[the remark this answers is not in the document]
corwintines: I think my plan for this will be to bring the PR mentioned in here to avoid duplication as I will be out next week.

src/components/FindWalletProductTable/data/WalletColumns.tsx
```diff
@@ -0,0 +1,47 @@
...
+// You can use a Zod schema here if you want.
+export type WalletColumns = {
+  id: string
+  logo: StaticImageData
+  walletInfo: WalletData
+}
+
+export const WalletColumns: ColumnDef<WalletColumns>[] = [
+  {
+    accessorKey: "information",
```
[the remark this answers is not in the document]
corwintines: will change to id instead of accessorKey

src/components/FindWalletProductTable/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { WalletColumns } from "@/components/FindWalletProductTable/data/WalletColumns"
+import { WalletFilters } from "@/components/FindWalletProductTable/data/WalletFilters"
+import { WalletPersonaPresets } from "@/components/FindWalletProductTable/data/WalletPersonaPresets"
+import ProductTable from "@/components/ProductTable"
+
+import { DEFAULT_LOCALE } from "@/lib/constants"
+
+const FindWalletProductTable = ({ wallets }) => {
+  const walletPersonas = WalletPersonaPresets()
+  const walletFilterOptions = WalletFilters()
```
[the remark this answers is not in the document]
corwintines: Ill change this for the wallet personas, but since the walletfilters have react components in it I think that should be fine as it is. Let me know if you disagree.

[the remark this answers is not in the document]
corwintines: Actually, im getting error warnings for both because they use the useTranslation hook. Let me know what youd prefer here.

konopkja:
Screenshot 2024-09-23 at 17 46 35

@corwintines  these seem to work as multi select, but use radio component leading user to think there is only at a time possible, should switch to checkbox

konopkja:
Screenshot 2024-09-23 at 17 49 41

the subtext should be grey - body medium rather than white by default,  turns white on select

src/components/ui/Table.tsx
```diff
@@ -13,26 +13,46 @@ import { cn } from "@/lib/utils/cn"
 const tableVariants = tv({
```
pettinarip, line 13: I see what you are doing here, but this is losing the point of this utility. We are repeating the same styles many times.

The changes that you've made explain that the product table is very different from the other tables that we have.

Are we sure that we can't override or share some of the base styles that we had?

I don't feel that we need to block this PR for that reason, but I do feel that we should do something a little simpler. I'll take a closer look at it later. Maybe this is also a good opportunity to discuss with the designers and align these styles with the rest.

src/data/wallets/wallet-data.ts
```diff
@@ -44,7 +44,8 @@ export const walletsData: WalletData[] = [
     last_updated: "2022-06-22",
     name: "Keystone",
     image: KeystoneImage,
-    brand_color: "#ffffff",
+    color: "bg-[#FFFFFF]",
+    brand_color: "from-[#ffffff]",
```
pettinarip, line 48: Not blocking: ideally, these files should know nothing about the implementation details to keep them decoupled. I'd have left the raw colors and applied them using `style={{ background-color: ${...} }}`

src/components/FindWalletProductTable/WalletSubComponent.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { MdInfoOutline } from "react-icons/md"
+
+import { FilterOption, Lang, WalletData } from "@/lib/types"
+
+import { WalletFilters } from "@/components/FindWalletProductTable/data/WalletFilters"
+import {
+  GreenCheckProductGlyphIcon,
+  WarningProductGlyphIcon,
+} from "@/components/icons/staking"
+import InlineLink from "@/components/Link"
```
pettinarip, line 13: We should be using the Tailwind version under `@/components/ui/Link`

src/components/FindWalletProductTable/FindWalletProductTable.stories.tsx
```diff
@@ -0,0 +1,32 @@
+import { Meta, StoryObj } from "@storybook/react"
+
+import FindWalletProductTable from "@/components/FindWalletProductTable"
+
+import walletsData from "@/data/wallets/wallet-data"
```
pettinarip, line 5: Since the imported file is quite heavy, it will add to the already heavy bundle we have to upload to Chromatic every time we run these tests.

Having said that, we don't really need to replicate the exact same version as in prod, we could just write two or three mocked rows and we could validate the visual tests with a much lighter story.

pettinarip (commented):
@corwintines left a few observations and small requests, lmk what do you think! 🚀

src/components/FindWalletProductTable/WalletSubComponent.tsx
```diff
@@ -0,0 +1,188 @@
...
+    eventName: `More info ${wallet.name}`,
+  })
+
+  return (
+    <div className="flex flex-row gap-2">
+      <div className="w-1 md:w-14">
+        <div
+          className={cn(
+            `m-auto h-full w-1 bg-gradient-to-b ${wallet.brand_color} to-[rgba(217, 217, 217, 0)] to-97%`
+          )}
```
wackerow, line 75: `[rgba(217, 217, 217, 0)]` = `transparent` = Default for both `from-` and `to-` colors once other tailwind gradient classes have been applied. (`from-0% from-transparent via-50% via-transparent to-100% to-transparent` are the defaults to my knowledge)

And instead of string interpolating here, we should add `wallet.brand_color` as a second argument to `cn()` (similar to `className`)

```suggestion
          className={cn(
            "m-auto h-full w-1 bg-gradient-to-b to-97%",
            wallet.brand_color
          )}
```

src/data/wallets/wallet-data.ts
wackerow: Okay, so @pettinarip while working within the confines of Tailwinds, you're proposing that in this case where we would need to string interpolate to just apply it using `style`...

Just want to note, that while `bg-[#FFFFFF]` is equivilent to `background-color: #FFFFFF`:

```css
.bg-\[\#FFFFFF\] {
  --tw-bg-opacity: 1;
  background-color: rgb(255 255 255 / var(--tw-bg-opacity)) /* #ffffff */;
}
```

...worth noting that `from-[#ffffff]` is actually altering a CSS variable `--tw-gradient-from`:

```css
.from-\[\#ffffff\] {
  --tw-gradient-from: #ffffff var(--tw-gradient-from-position);
  --tw-gradient-to: rgb(0 0 0 / 0) var(--tw-gradient-to-position);
  --tw-gradient-stops: var(--tw-gradient-from), var(--tw-gradient-to);
}
```

When I adjust this inside the `style` attribute I get a type error if I don't add `as React.CSSProperties`:

```
You can't pass property "--tw-gradient-from" to type Properties<string | number, string & {}>.
```

```tsx
  style={{
    "--tw-gradient-from": "#000000",
  } as React.CSSProperties}
```

wackerow: Similar, don't see as a blocker but I do tend to agree that the property name should indicate what it is... so if we're calling it `color`, we should pass a color, not a tw class... alternatively we could rename the property to `twBgClass` for example to more accurately explain what is to be expected. Would patch separately to avoid adding blockers.

src/components/FindWalletProductTable/index.tsx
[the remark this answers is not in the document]
wackerow: Yeah, agree these should be hooks with `useWhatever` naming... As hooks, I would probably also pull them out of the `data` folder and probably just have them sit next to index.tsx

src/data/wallets/wallet-data.ts
pettinarip: Ah I didn't see that `.from-[...]` case. Yeah, I would expect to do something like you shared. The type error I think is fine since it is not a known css prop name, but curious, did that work?

Again, minor thing, we can skip this. I like the renaming idea as well but not a strong stance from my side.

konopkja:
Screenshot 2024-09-24 at 12 26 10

the paragraph text is supposed to be body medium

konopkja:
Screenshot 2024-09-24 at 12 27 35

the text for languages is i think smaller than currently we have in prod?

konopkja:
Screenshot 2024-09-24 at 12 28 35

there is supposed to be sticky header with wallet counter and reset button that is missing in preview

konopkja:
Screenshot 2024-09-24 at 12 29 34

missing mouse over effect in preview (screenshot from prod)

src/data/wallets/wallet-data.ts
wackerow: Hm.. actually, no, not as we'd like it. So, we can declare that variable, but without having a TW class in there to tee-up the other parts of that class, it doesn't work:

```css
.from-\[\#ffffff\] {
  --tw-gradient-from: #ffffff var(--tw-gradient-from-position);
  --tw-gradient-to: rgb(0 0 0 / 0) var(--tw-gradient-to-position);
  --tw-gradient-stops: var(--tw-gradient-from), var(--tw-gradient-to);
}
```

`--tw-gradient-from` and `--tw-gradient-to` are instantiated with the `.from-[<color>]` class, but more importantly they do nothing without the `--tw-gradient-stops: var(--tw-gradient-from), var(--tw-gradient-to);` also found in that TW class.

A workaround is to declare `from-transparent` and then _also_ apply the `style={{"--tw-gradient-from": "#000000"} as React.CSSProperties}` override... this will set up the TW gradient CSS variables, and then we overwrite the `from` token

It's hacky imo... I'm personally leaning towards just updating the variable _names_ to say `textClass` and `bgClass` instead of `color` and `brand_color`.

wackerow: Also just noting for learning purposes that the `to-[]` or `bg-gradient-to-[]` class does _not_ tee up `--tw-gradient-stops`; only `from-[]`

src/components/ProductTable/Table.tsx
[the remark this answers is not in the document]
corwintines: Good idea, changed folder structure for this

src/data/wallets/wallet-data.ts
corwintines: Will update the variable name here in data :)

konopkja:
Screenshot 2024-10-01 at 17 04 24
Screenshot 2024-10-01 at 17 05 00

when selected persona, the description should turn white, there is also small animation in the prod i think that would be nice to keep

konopkja:
when there is subset of wallets bcs of active filter, it should indicate the amount from total that is visible (see prod)
Screenshot 2024-10-01 at 17 07 30

Screenshot 2024-10-01 at 17 07 28

konopkja:
IMG_46911AD66630-1


testing mobile i had issues clicking on persona in the filter modal, several times and it didnt select. (more looked like just mouse over triggered instead of selection)

konopkja:
Screenshot 2024-10-01 at 17 17 34

1. sidebar breaks when scrolling
2. when i scroll down and scroll up again and then try scrolling back down it stops working
3. missing "x" close button on top (see production, there is sticky x)

konopkja:
IMG_853108AA93B2-1

is there a way to fix this issue?

konopkja:
IMG_B43D3C26BC7D-1

somehow on mobile im missing wallet links

konopkja:
Screenshot 2024-10-01 at 17 23 49

above screen shows few issues on mobile in terms of how the design should be structured

bellow screen shows current production - how it should look

Screenshot 2024-10-01 at 17 24 00

nloureiro:
@corwintines the line is broken on mobile (brave on macOS)
Screen Shot 2024-10-18 01 56 33 PM

nloureiro:
I am not sure if it's a bug, but if we aim for feature parity to production the button on tablet size it's not 100% width
just on mobile
Screen Shot 2024-10-18 01 58 31 PM

nloureiro:
we are missing some spacing on the list
Screen Shot 2024-10-18 02 05 40 PM

nloureiro:
When the product doesn´t have a tag, it needs spacing
Screen Shot 2024-10-18 02 09 08 PM

nloureiro:
The head of the table has a different label. On the PR is way simpler. Was it on purpose?
Screen Shot 2024-10-18 02 16 31 PM

nloureiro:
What is this bar on the top of the mobile filters? We don´t have it on production

Screen Shot 2024-10-18 02 21 42 PM

corwintines:
> What is this bar on the top of the mobile filters? We don´t have it on production
> 
> Screen Shot 2024-10-18 02 21 42 PM

This was part of the drawer component, removed it for now but just saying where it came from

corwintines:
> The head of the table has a different label. On the PR is way simpler. Was it on purpose? Screen Shot 2024-10-18 02 16 31 PM

I could update this if needed, but I generalized it in the abstraction of this component

konopkja:
Screenshot 2024-10-21 at 18 43 18

the tooltips position @wackerow  pls check it

wackerow:
> the tooltips position @wackerow pls check it

#14207

wackerow (approved):
🎉🎉🎉 Nice job @corwintines! This feels very robust while testing it out.

---

Found a few tiny nice-to-haves that I do not see as blockers, but will just note here:

- Add `type` to all `"@/lib/types"` imports
- "Filters" button on mobile has weird focus outline... could consider `-ms-1 ps-1` to shift start edge
- Personas on mobile still not totally clear you can scroll to the right for more
- Getting a quick flash/glitch when opening and closing the mobile filter menu (Brave or Safari, iPhone)
- My patch changing the `Icon` attribute for `SwitchFilterInput` may not be the most appropriate way to pass a prop, but working for now
- Separate PR open for the icon alignment

wackerow:
Approved from dev side. cc: @nloureiro and/or @konopkja for approval from design side 🙏, then can pull this in (ideally before this weeks release if no critical blockers)

konopkja:
lgtm! last two things i found:

1.  reset button should be centered with see wallets button
2. this modal window has oval corners on top, should be straihgt instead
Screenshot 2024-10-22 at 11 31 06

konopkja:
Screenshot 2024-10-22 at 16 21 25

when i select device, by default it does not subselect child checkboxes resulting in no filtration 

expected: both ios and android are selected

corwintines:
Going to bring this in with @wackerow review, and approval from @nloureiro and @konopkja in Discord after fixing these last issues.

PR #13814, opened 2024-09-06
Updated CL table

arredr2:
Removed a hanging | that wasn't producing a table.

## Description

Removed a hanging | that wasn't producing a table.


arredr2 (commented):
Added | to the end of Grandine's line!

PR #13817, opened 2024-09-07
ui: round inner meter in ui/progress

wackerow:
## Description
- Updates base `ui/progress` component with `rounded-full` on the inner progress bar element

image

^ Maintains this appearance from previous Chakra-UI `Progress`

cc: @nloureiro I didn't see this component in the DS, but noticed we previously had the rounded approach on the inside bar. Came across it when updating colors on the Quizzes page, so if we don't want to change the base component here, I can just close it out and apply to that usage only.


nloureiro (approved):
I like it, aproved

pettinarip (approved):

PR #13821, opened 2024-09-07
Design system color theming updates

wackerow:
## Description
- Incremental adjustments updating colors to latest design system theming

## Related Issue
Design system


nloureiro (approved):

pettinarip (approved):

pettinarip (changes requested):
@wackerow left a few questions on Chromatic where we have a few visual differences.

wackerow:
Good catch! Was related to the replacement for the "Close" icon in the `DismissableBanner` component... 
- The default variant was taller than the button being used before
- The `ghost` variant was using the same color as the foreground, so it was invisible.

I changed it to use `size="sm"` and removed variant (defaults back to `solid`). The solid button is _close_ here (fixes the text color), but we don't have a proper variant for buttons like this on a `primary` background... so I adjusted the button with `hover:shadow-none ` which should do the trick.

This should reach a nice compromise between keeping the new custom button styling to a minimum, while also matching it close-enough to the existing close button details.

pettinarip (approved):
Thanks @wackerow looks good now on SB.

PR #13827, opened 2024-09-08
Migrate `ExpandableInfo` and `runNodeGradient2`

wackerow:
## Description
`ExpandableInfo` component only used in one place, with significant portions of the component no longer being utilized, nor necessary. Contains the only usage of the `runNodeGradient2` token.

- refactor: Updates the single usage point of `ExpandableInfo` to use tailwind/shadcn instead
- deprecate: ExpandableInfo component
- theme: deprecate runNodeGradient2 token

image

## Related Issue
Design system gradient/color migration


pettinarip (approved):
LGTM

PR #13828, opened 2024-09-08
Migrate EthPriceCard, update associate color tokens

wackerow:
## Extends `colors-1` branch in PR:
- #13821 

(Can be merged in any order)

## Description
- Updates EthPriceCard
- This component utilized success/error colors, so was used as an opportunity to sync theme with Figma DS
- Updates red/yellow/green color palettes; trims orange palette to only what is left being used
- Updates naming of success/error/warning tokens per DS, and updates overall ordering of tokens to match DS

## Related Issue
Design system gradient/color updates


nloureiro:
@wackerow added the same gradient on both themes because on dark there was no gradient.

I know that in production the dark theme does not have a visible gradient, but I think we can fix this.


Screen Shot 2024-09-09 10 49 07 PM
Screen Shot 2024-09-09 10 48 59 PM
Screen Shot 2024-09-09 10 56 35 PM

pettinarip (approved):
GJ! and nice colors cleanup 💪🏼

PR #13840, opened 2024-09-10
Create FAQ Accordion Component

### LINKED ISSUE #11118: FAQ

saurabhburade:
## Description
Create `FaqAccordion` Component

## Related Issue
- Closes #11118


saurabhburade:
Currently, using a tailwind accordion code copy with few modifications.

src/components/FaqAccordian/index.tsx
wackerow, line 1: accordi**O**n not accordi**A**n

Need to fix the file paths here: `FaqAccordion/index.tsx`
(Same with the story: `FaqAccordion/FaqAccordion.stories.tsx`)

src/components/FaqAccordian/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+import { cn } from "@/lib/utils/cn"
+
+import * as RootAccordion from "../../../tailwind/ui/accordion"
+
+const AccordionTrigger = React.forwardRef<
+  React.ElementRef<typeof AccordionPrimitive.Trigger>,
+  React.ComponentPropsWithoutRef<typeof AccordionPrimitive.Trigger>
+>(({ className, children, ...props }, ref) => (
+  // p-4 hover:bg-background-highlight
```
wackerow, line 13: ✂️ 
```suggestion
```

src/components/FaqAccordian/index.tsx
wackerow: But, this also makes me wonder if we should even be calling this "FaqAccordion", since in the DS it's just called `FAQ`... @pettinarip Any thoughts here? Simply calling this `component/FAQ/index.tsx` would work for me, and could just swap "FAQ" for "Accordion" in the exported sub-components in this file.

src/components/FaqAccordian/index.tsx
```diff
@@ -0,0 +1,92 @@
+import React from "react"
+import { MdChevronRight } from "react-icons/md"
+import * as AccordionPrimitive from "@radix-ui/react-accordion"
+
+import { cn } from "@/lib/utils/cn"
+
+import * as RootAccordion from "../../../tailwind/ui/accordion"
+
+const AccordionTrigger = React.forwardRef<
```
wackerow, line 9: We should avoid using the same names as inside the `ui/accordion` component to avoid ambiguity/confusion. Made a note above about potentially renaming the file, but in either case we should include "Faq" or "FAQ" in these component names to make it clear.

src/components/FaqAccordian/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+  React.ElementRef<typeof AccordionPrimitive.Trigger>,
+  React.ComponentPropsWithoutRef<typeof AccordionPrimitive.Trigger>
+>(({ className, children, ...props }, ref) => (
+  // p-4 hover:bg-background-highlight
+  <AccordionPrimitive.Header className="flex w-full">
```
wackerow, line 14: Is there a reason we didn't go with the existing `AccordionTrigger` from `tailwind/ui/accordion.tsx`, instead of rebuilding from the Radix-UI primitive?

src/components/FaqAccordian/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+          "group-hover:border-primary group-hover:text-primary group-hover:shadow-md group-hover:shadow-primary-low-contrast",
+          "transition-transform duration-200"
+        )}
+      />
+    </AccordionPrimitive.Trigger>
+  </AccordionPrimitive.Header>
+))
+AccordionTrigger.displayName = AccordionPrimitive.Trigger.displayName
+
+// TODO: Prop types
```
wackerow, line 43: yes please =)

src/components/FaqAccordian/FaqAccordian.stories.tsx
wackerow, line 1: https://github.com/user-attachments/assets/9343169c-f3f4-45fa-8ecc-d678731eae8c

Functioning pretty well, just noting we should handle the width. In the DS we have `max-width: 52rem;` Ultimately though, the width shouldn't change when open vs closed.

wackerow (changes requested):
Thanks so much for the help here @saurabhburade! Just left some questions, requests, etc =)

src/components/FaqAccordian/index.tsx
saurabhburade: It's used for grouping styles with `MdChevronRight` ==> `"group-hover:border-primary group-hover:text-primary group-hover:shadow-md group-hover:shadow-primary-low-contrast"`

src/components/FaqAccordian/FaqAccordian.stories.tsx
saurabhburade: We can implement `max-width: 52rem;` or near value`max-w-4xl ==> 56rem`. One thing I notice is, its taking parent width if mentioned and fits it to full width (trigger & content both)

saurabhburade: https://github.com/user-attachments/assets/fe55c72a-656d-41a2-be23-750d5e86ba78

wackerow (approved):
https://github.com/user-attachments/assets/26538f96-fafe-4bcb-8c59-f984559cfe6e

Polished this up a bit more; should match DS fully. Bringing in, thanks again for getting this going @saurabhburade!

PR #13845, opened 2024-09-10
Added clarity by specifying the types of node in the trie example

TakeASwing-420:
## Description

I added clarity by specifying the types of node in the trie example; like which one is leaf node which one is extension node etc. in the comments.


TakeASwing-420:
@wackerow, please review this PR.

PR #13854, opened 2024-09-12
Shadcn migration - Breadcrumbs

saurabhburade:
## Description
Migrate Breadcrumbs to tailwind.


pettinarip (changes requested):
@saurabhburade thanks for this PR!

Visually it is correct but I think we are missing some semantic value from the previous version. We are missing the ordered list and items that gave the hierarchical order of the links.

For this reason, and because the shadcn component will also add a few more features, I think it is better to install the [Breadcrumb](https://ui.shadcn.com/docs/components/breadcrumb) component in our `/components/ui` folder using the cli or copying the files yourself. Once we have that we can use these components in our custom `Breadcrumbs` component to have the correct structure and recommended semantics.

```
<Breadcrumb>
  <BreadcrumbList>
    <BreadcrumbItem>
      <BreadcrumbLink href="/">Home</BreadcrumbLink>
    </BreadcrumbItem>
    <BreadcrumbSeparator />
    <BreadcrumbItem>
      <BreadcrumbLink href="/components">Components</BreadcrumbLink>
    </BreadcrumbItem>
    <BreadcrumbSeparator />
    <BreadcrumbItem>
      <BreadcrumbPage>Breadcrumb</BreadcrumbPage>
    </BreadcrumbItem>
  </BreadcrumbList>
</Breadcrumb>
```

pettinarip (approved):
@saurabhburade great! just pushed a commit to use the Next link component instead of the native one to get better i18n support for each link.

Approving, thanks for the PR!

PR #13859, opened 2024-09-13
Shadcn migration - MeetupList

saurabhburade:
## Description
Migrate `MeetupList` to tailwind.

Preview: https://deploy-preview-13859--ethereumorg.netlify.app/en/community/events/#meetups


pettinarip (commented):
@saurabhburade looks good!

You imported the `off-background` and `table-background-hover` colors which is fine but these are old colors from an old theme and we are trying to deprecate them in the process. I'll tweak a bit these colors to use the new Design System colors and that should be all.

Thanks!

pettinarip (approved):

nloureiro (approved):

PR #13882, opened 2024-09-17
Refactor nav's load

pettinarip:
## Description

This PR cleans up the loading exp for the nav a bit.

- remove the `isDesktop` hook and replace the logic behind it with media queries using `useBreakpointValue`. This fixes an edge case where large mobile screens or tablets would not load the desktop menu.
- make most of the nav components client-rendered. This improves the UX by showing the interactive components only when they are ready to be interactive, not before.


wackerow (approved):
Looking better... This fixed the issue I was seeing where landscape mobile devices could end up without any menu available.

Just noting that this still has a very step-wise approach to loading:
1. Logo only, blank bar
2. Right side loads, but menu categories and Cmd/Ctrl-K missing
3. Cmd/Ctrl-K indicator loads
4. Menu items finally load

https://github.com/user-attachments/assets/53ade49d-c8a9-465d-9479-2458032781d0

This is still an improvement so I think we can pull this in now.

PR #13883, opened 2024-09-17
Home page: lazy load images

pettinarip:
## Description

Marks external images used in events and blog posts as lazy loaded.


wackerow (approved):
💪

PR #13906, opened 2024-09-18
Marquee tooltips

pettinarip:
## Description

Sets custom container to tooltips inside marquee to keep animation paused while hovering the tooltip content.


wackerow (approved):
Thanks @pettinarip!

PR #13908, opened 2024-09-19
Fix: "Select" naming collision

wackerow:
## Description
- Moves recently added `/components/ui/Select/*` component up a level to `/components/Select/*`, a component that follows the pattern of exporting the parent element
- Relocated story to same folder
- Leaves the new `/components/ui/select.tsx` in place, which follows that same pattern of exporting it's exploded parts

Fixes build errors on #13807

Can separately address deprecating one of these in favor of the other.

## Related Issue
- #13807


pettinarip (approved):
Approved to unblock the other PR but I don't think we should keep both selects since they both use `react-select` and is not clear when to use one or the other. Ideally, we should keep one of them and adjust the instances accordingly. We could tackle this refactor in a different PR if we agree.

Alkhtri77:
[Uploading 0x51173e6623a29e7ffcc56ad5f0b71025756be1f4caf13a36c7e8a8b4e2b97f4d.pdf…]()

Alkhtri77:
Hello help me

PR #13528, opened 2024-07-29
Homepage redesign

wackerow:
# 🎉 🔜 

## Description

Introduces a revamp to the ethereum.org homepage

### `public/images/**`

Imported new image assets. Up-res'ed the event placeholder image.

### `src/`

- **`@chakra-ui/`**

  - `colors.ts` and `semanticTokens.ts` updated to match new token changes. Stepping-stone for upgrading old components to new theme using Tailwind/css tokens
  - `Button.ts` - Updates solid variant to use new `primary-action` for background, and white for text

- **`components/`**

  - **`BigNumber/index.tsx`** - Replaces old stats box `GridItem`
  - **`Buttons/SvgButton.tsx`** - New unique button type, currently placed as own component
  - **`Chevron/index.tsx`** - RTL-responsive _next_ and _prev_ arrows (instead of using Right and Left chevrons).
  - **`Hero/HomeHero/index.tsx`** - Updated with new theme/design, updated for tailwind
  - **`Homepage/`** - Subdirectory for homepage-specific components and hook logic
    - **`BentoCard.tsx`** - Card specifically designed for the homepage BentoBox and card Swiper
    - **`useBentoBox.ts`** - Extracted logic and setup for homepage BentoBox
    - **`useHome.ts`** - Extracted logic for remainder of homepage tasks
  - **`icons/*.svg`** - `1em` x `1em` SVG icons imported for use throughout the homepage
  - **`Nav/`** - Chevron update; Color mode toggle updates \* See TODOs
  - **`StatsBoxGrid/`**
    - **`useStatsBoxGrid.tsx`**
      - Fetches ETH Price to multiply against total ETH staked, for total USD value securing the network.
      - Replaces node count metric with GrowThePie average tx COST
      - Updates tx COUNT metric to use GrowThePie data
      - Simplifies to new schema where we only have a single value, not a time range array
  - **`Swiper/index.tsx`** - Re-usable swiper component built with [Swiper](https://swiperjs.com/), used on homepage for bento cards and RSS post cards
  - **`ui/`**
    - **`Button.tsx`** - Added `lg` size, and updated with `primary-action` color for `solid` variant
    - **`card.tsx`** - Re-usable card components, initialized via shadcn
    - **`section.tsx`** - Re-usable section components, templated from card.tsx
    - **`skeleton.tsx`** - Re-usable skeleton components, initialized via shadcn, custom added SkeletonLines for variable-line-width multi-line skeletons
[...]

Issue #13851, opened 2024-09-12
Image breaks on events page if src url is not working


saurabhburade:
### Describe the bug

The event image is broken if URL of image is not working.

### To reproduce

1. Go to '/community/events/'
2. Click on 'View More'
3. Check ETHVenice 2024, ETHRome, Permissionless III

### Expected behavior

Add fallback image if src url gives error

### Screenshots

<img width="519" alt="Screenshot 2024-09-12 at 12 35 46 PM" src="https://github.com/user-attachments/assets/91ae90dd-e611-4c8d-a6de-b2dac91c5553">

### Would you like to work on this issue?

- [ ] Yes
- [ ] No

saurabhburade:
cc @wackerow

saurabhburade:
This can be handled with the below but might lose SSR.
https://vercel.com/templates/next.js/image-fallback

PR #13910, opened 2024-09-19
chore: add/fix imageUrl for upcoming events

wackerow:
## Description
- Adds imageUrl for upcoming events

Demo'ing next 30:

image

## Related Issue
- #13528


saurabhburade:
Hi @wackerow, just popping up the issue #13851.

This happens if there is an issue with the image URL. There might be cases where assets are removed from these sites.

corwintines (approved):

PR #13915, opened 2024-09-20
Hotfix color mode

pettinarip:
## Description

1. Drops support for `system` color mode since we have edge cases where the theme providers run out of sync. I would suggest keeping this until we remove chakra from the codebase.
2. Use `data-theme` attr to detect dark mode instead of `class`. This aligns with Chakra's color mode detection method.


minimalsm (approved):
LGTM, tested the errors @konopkja pointed out and they are resolved for me.

konopkja (approved):

Issue #13914, opened 2024-09-20
Bug report: event text too long - introduce character limit


konopkja:
### Describe the bug

The event description has no text limitation and can become too long.  At some point it can break design + looks messy with other events combined with less text next to it.

**We need to introduce a character limit for event description:**

**Max characters:** 230

Then cut and show three dots
... 

### To reproduce

Go to https://ethereum.org/en/community/events/

### Expected behavior

character limit of 230 characters

### Screenshots

<img width="851" alt="Screenshot 2024-09-20 at 10 55 16" src="https://github.com/user-attachments/assets/04650ff9-296d-44f9-a8f2-f2c530e0729a">

### Would you like to work on this issue?

- [ ] Yes
- [X] No

sumitvekariya:
@konopkja Please assign this to me

konopkja:
> @konopkja Please assign this to me

assigned, thank you!

konopkja:
thank you for the fix @sumitvekariya

PR #13919, opened 2024-09-20
truncate long event description

sumitvekariya:
## Description

The event description had no text limitation and could have been too long. So now event description is limited to 230 characters, if longer than that it is truncated to 230 characters followed by `...`

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/13914


src/components/EventCard.tsx
```diff
@@ -40,6 +40,12 @@ const EventCard = ({
     new Date(endDate?.replace(/-/g, "/"))
   )
 
+  const MAX_DESCRIPTION_CHARS = 230
+  const truncatedDescription =
+    description.length > MAX_DESCRIPTION_CHARS
+      ? `${description.slice(0, MAX_DESCRIPTION_CHARS)}...`
+      : description
+
```
wackerow, line 48:
```suggestion
```

src/components/EventCard.tsx
```diff
@@ -108,7 +114,7 @@ const EventCard = ({
               </Text>
             </Box>
             <Box>
-              <Text fontSize="sm">{description}</Text>
+              <Text fontSize="sm">{truncatedDescription}</Text>
```
wackerow, line 117: We'll just migrate this specific element so we can use tailwind:
```suggestion
              <p className="text-sm line-clamp-6">{description}</p>
```

wackerow (changes requested):
Hey @sumitvekariya! Thanks for the PR... Apologies if the issue may have been misleading, but I think we can get away with simply restricting the _line_ count here (instead of _word_ count), which we can use a simply TailwindCSS class for this: `line-clamp-x`... by default this has up to `line-clamp-6`, so I would suggest this )(just chatted with @konopkja as well who is in agreement)

Left a suggestion for how we can simplify to use this approach

Screenshot2024_09_23_090716

corwintines (approved):
Thanks @sumitvekariya!

PR #13925, opened 2024-09-21
fix: Select hover border color contrast ratio

wackerow:
## Description
- Replaces use of low-contrast with high-contrast primary color variant for `Select` component hover border color

## Related Issue
Poor contrast ratio, giving a "disappearing" experience on select hovering


nloureiro (approved):

pettinarip (approved):
👍🏼

PR #13926, opened 2024-09-21
refactor: menu color tokens

wackerow:
## Description
**Proposal**: Migrate our "menu" tokens to tokens that are more generalizable.

Instead of declaring a whole array of specific "menu" tokens, with different level numbers, this PR suggests adding the additional shades of gray that are needed as additional `background` token shades.

We currently only have `background` and `background-highlight` for background tokens. This would add `low`, `medium` and `high` background variants to be used as layers to any stack (ie. nested menus, nested accordions, or any other nested elements).

These tokens were then used directly at the component level, updating and removing the existing usage of "menu-x-(active-)background" with the new corresponding background token.

Similar approach with text, but unsure of potential naming convention here. Went with a nested approach under "body" as "body-menu-DEFAULT/low/medium/high"

**Shade update:**
This also fixes the shade that was `--menu-1-active-background: var(--gray-150);` to now use `--background-low: var(--gray-100);` in light mode:

image

## Related Issue
Design system; theming updates

### DIFF
--- a/src/styles/semantic-tokens.css
+++ b/src/styles/semantic-tokens.css
@@ -14,7 +14,10 @@
 
     --background: var(--white);
     --background-highlight: var(--gray-50);
-    --background-medium: var(--gray-200); /* TODO: VERIFY */
+    /* TODO: Add "low" "medium" and "high" to design system */
+    --background-low: var(--gray-100);
+    --background-medium: var(--gray-200);
+    --background-high: var(--gray-300);
 
     /* TODO: Add border color tokens to match DS */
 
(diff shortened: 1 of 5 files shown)


nloureiro (approved):

src/styles/semantic-tokens.css
```diff
@@ -132,7 +123,11 @@
 
     --background: var(--black);
     --background-highlight: var(--gray-900);
-    --background-medium: var(--gray-950); /* TODO: VERIFY */
+
+    /* TODO: Add "low" "medium" and "high" to design system */
```
pettinarip, line 127: nitpick: would avoid adding todos that are outside of the scope of the codebase as they are hard to maintain and easy to forget.

pettinarip (approved):
@wackerow great job on this. Its much cleaner and easier to reuse these colors now.

Not blocker: I agree with the naming, my only concern is the misunderstanding of usage between the old and new backgrounds (ie `background-highlight` vs `background-high`). I was thinking of something more specific like `background-shade-low` to make them feel like lower level background colors and avoid potential overuse. Not strong opinion either so feel free to ignore xD

src/styles/semantic-tokens.css
wackerow: Yes yes, sounds like we have consensus and approval on this proposal, so I'll go ahead and remove that and add notes in the DS

wackerow:
Pulled in; Removed TODO's and updated design system in Figma... Will chat about and address the naming adjustments separately.

PR #13931, opened 2024-09-23
refactor: explode ui/swiper into compossible parts

wackerow:
## Description
- Explode ui/swiper into compossible parts
- Improves ability to custom style swiper sub-components with tailwind classes
- Fixes navigation/pagination styling to avoid `absolute` positioning, with responsive width


src/components/ui/swiper.tsx
```diff
@@ -0,0 +1,156 @@
...
+        {children.map((child, index) => (
+          <SwiperSlide key={index}>{child}</SwiperSlide>
+        ))}
+        {navigation || (
+          <SwiperNavContainer>
+            <SwiperPrevButton />
+            <SwiperPaginationDots />
+            <SwiperNextButton />
+          </SwiperNavContainer>
+        )}
```
pettinarip, line 141: This is good but I think it is reducing flexibility and customization.

1. To simplify, you could export a `SwiperNavigation` component if wrap this common shape.
2. One goal of exporting all these sub components is to avoid having props like `navigation` because you are reducing flexibility. What if I want the nav to be at the top of the slides and not the bottom? exporting the necessary sub components and rendering just the children gives you all this flexibility for free.

```tsx
// following the convention you could have something like this without any config or custom props
<Swiper>
  <SwiperNavigation />

  <div className="p-4 bg-green-500">
    <SwiperSlide>
      <div className="bg-red-500"></div>
    </SwiperSlide>
  </div>

  <div className="p-4 bg-green-500">
    <SwiperSlide>
      <div className="bg-blue-500"></div>
    </SwiperSlide>
  </div>

  <SwiperNavigation />
</Swiper>
```

pettinarip (approved):
@wackerow this is a great refactor. Thanks for doing it. Liked that we have the styles closer to the components.

Left IMO a comment with an improvement, feel free to disagree and merge.

src/components/ui/swiper.tsx
wackerow: Yeah, couldn't agree more! implementing these suggestions

wackerow: Just pushed those updates. With how SwiperJS works, it's best to have a container to handle it's constraints (the `height` is a must at a minimum)

So we'll have this structure:

```tsx
<SwiperContainer className="tw-styles-as-needed">
  <Swiper {...swiperOptionsAsNeeded}>
    <SwiperSlide>
      <!-- contents, styling -->
    </SwiperSlide>
    <!-- repeat slides to desires -->
    <SwiperNavigation className="tw-styles-as-needed" />
  </Swiper>
</SwiperContainer>
```

for example:

```tsx
<SwiperContainer>
  <Swiper>
    <SwiperSlide>
      <div className="bg-black">
        <p className="text-white">Hello</p>
      </div>
    </SwiperSlide>
    <SwiperSlide>
      <div className="bg-white">
        <p className="text-black">World</p>
      </div>
    </SwiperSlide>
    <SwiperNavigation />
  </Swiper>
</SwiperContainer>
```

This allows option to not add navigation, or to override styling of each of these using TW classNames. Unfortunately there are limitation to Swiper React that limit our ability to manipulate the `.swiper-wrapper` element that is added by this plugin with JS, and then forces the order of the DOM to have the `SwiperNavigation` elements after it.

As a workaround, I've added a `top` and `bottom` placement variant called `navigationPlacement` to allow user to effectively switch this to `flex flex-col-reverse` to place the nav contents above the swiper slides.

pettinarip (approved):
Well done @wackerow, merging 👍🏼

PR #13933, opened 2024-09-23
desktop and mobile 16 px font

nloureiro:
Testing the font size more prominent on mobile for accessibility

### DIFF
--- a/src/styles/global.css
+++ b/src/styles/global.css
@@ -76,7 +76,7 @@
 
 @layer base {
   body {
-    @apply !bg-background font-body text-sm !text-body lg:text-md;
+    @apply !bg-background font-body text-lg !text-body lg:text-md;
   }
 
   a {
--- a/src/styles/global.css
+++ b/src/styles/global.css
@@ -76,7 +76,7 @@
 
 @layer base {
   body {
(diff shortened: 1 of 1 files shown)


wackerow (commented):
Just curious, why would we suddenly make the mobile text larger than desktop?

wackerow:
(And just flipping to draft while "WIP"; flip back when ready)

nloureiro:
> Just curious, why would we suddenly make the mobile text larger than desktop?

The idea was to flip back to 16 on both screen sizes, but then @konopkja "pushed" me to test with 18

What do you think? Could you test it on your phone and tell us how it feels? Is it better to read on the phone?

nloureiro:
@wackerow, let's push this on the next deployment with a 16px on all screens and discuss the 18px later... so, ready for review :)

konopkja:
> @wackerow, let's push this on the next deployment with a 16px on all screens and discuss the 18px later... so, ready for review :)

actually thats next thursday no? id say thats a bit long waiting time

wackerow (approved):
Pulling this into `dev`. Can hot fix separately if we feel it's warranted.

PR #13942, opened 2024-09-23
fix: LanguagePicker max height

wackerow:
## Description
- Moves the max-height property to the `Command` parent for the Language Picker component
- Adjusts height to use `calc` allowing menu to drop to just above the feedback widget, regardless of screen height

## Related Issue
Fixes bug where the "Learn more" section at the bottom of the LanguagePicker was falling off the bottom edge of the screen on shorter displays/windows.

image

With fix:

image

## Preview link
https://deploy-preview-13942--ethereumorg.netlify.app/en/


pettinarip (approved):
lgtm

PR #13947, opened 2024-09-24
Shadcn migration - docs layout

pettinarip:
## Description

Migrates the Docs layout to Tailwind.

### DIFF
--- a/src/layouts/Docs.tsx
+++ b/src/layouts/Docs.tsx
@@ -1,16 +1,6 @@
 import { useRouter } from "next/router"
 import type { HTMLAttributes } from "react"
-import {
-  Badge,
-  Box,
-  type BoxProps,
-  Flex,
-  type FlexProps,
-  type ListProps,
-  OrderedList as ChakraOrderedList,
-  UnorderedList as ChakraUnorderedList,
-  useToken,
-} from "@chakra-ui/react"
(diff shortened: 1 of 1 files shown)


src/layouts/Docs.tsx
```diff
@@ -27,47 +17,33 @@ import FeedbackCard from "@/components/FeedbackCard"
...
-const Page = (props: ChildOnlyProp & Pick<FlexProps, "dir">) => (
-  <Flex
-    direction="column"
-    w="full"
-    borderBottom="1px"
-    borderColor="border"
-    {...props}
-  />
+const Page = (props: HTMLAttributes<HTMLDivElement>) => (
+  <div className="flex w-full flex-col border-b" {...props} />
```
wackerow, line 41: @pettinarip Do we care to clean up some of these one-off named components? `Page`, `ContentContainer`, and `Content` are all only used once bellow in the return, and not exported... Nothing critical, but would be nice imo to clean these up.

wackerow (changes requested):
@pettinarip Looks good! Left a note about just cleaning up a few single-use components.. lemme know if you disagree

src/layouts/Docs.tsx
pettinarip: Good idea! Done

wackerow (approved):

PR #13949, opened 2024-09-24
Shadcn migration - tutorials layout

pettinarip:
## Description

Migrates the Tutorials layout to Tailwind.

### DIFF
--- a/src/layouts/Tutorial.tsx
+++ b/src/layouts/Tutorial.tsx
@@ -1,15 +1,6 @@
 import { useRouter } from "next/router"
 import type { HTMLAttributes } from "react"
-import {
-  Badge,
-  Box,
-  type BoxProps,
-  Flex,
-  Kbd,
-  Text,
-  type TextProps,
-  useToken,
-} from "@chakra-ui/react"
+import { Badge } from "@chakra-ui/react"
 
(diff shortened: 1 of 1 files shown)


src/layouts/Tutorial.tsx
```diff
@@ -34,33 +24,16 @@ import {
...
 import YouTube from "@/components/YouTube"
 
 import { getEditPath } from "@/lib/utils/editPath"
 
-type ContentContainerProps = Pick<BoxProps, "children" | "dir">
-
-const ContentContainer = (props: ContentContainerProps) => {
-  const boxShadow = useToken("colors", "tableBoxShadow")
-
+const ContentContainer = (props: HTMLAttributes<HTMLDivElement>) => {
```
wackerow, line 32: Single use and not exported... Can we move `<MainArticle ...>` to the return statement?

wackerow (changes requested):
@pettinarip Small request similar to #13947, otherwise looks good!

src/layouts/Tutorial.tsx
pettinarip: Definitely

wackerow (approved):

PR #13951, opened 2024-09-24
Shadcn migration - more components

pettinarip:
## Description

Migrate the following components to TW.

- Banners/UpgradeBannerNotification.tsx
- Banners/BugBountyBanner/index.tsx
- CopyToClipboard.tsx
- EnvWarningBanner.tsx
- EthVideo.tsx
- FeaturedText/index.tsx
- IdAnchor.tsx
- LearningToolsCardGrid.tsx
- MarkdownImage.tsx
- MergeArticleList.tsx
- StatErrorMessage.tsx
- TranslationChartImage/index.tsx
- YouTube.tsx

### DIFF
--- a/src/components/IdAnchor.tsx
+++ b/src/components/IdAnchor.tsx
@@ -1,5 +1,4 @@
 import { CiLink } from "react-icons/ci"
-import { Icon } from "@chakra-ui/react"
 
 import Link from "@/components/Link"
 
@@ -16,7 +15,7 @@ const IdAnchor = ({ id }: { id?: string }) => {
       _focus={{ opacity: 1 }}
       transition="opacity 0.1s ease-in-out"
     >
-      <Icon as={CiLink} fontSize="xl" me="1" />
+      <CiLink className="me-1 text-xl" />
     </Link>
   )
 }
--- a/src/components/IdAnchor.tsx
+++ b/src/components/IdAnchor.tsx
@@ -1,22 +1,17 @@
 import { CiLink } from "react-icons/ci"
 
-import Link from "@/components/Link"
+import { BaseLink } from "@/components/ui/Link"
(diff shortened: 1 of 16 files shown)


src/components/IdAnchor.tsx
wackerow, line 1: image

Looks like most of this was missed... Will post a sub-branch PR with a fix real quick

wackerow (changes requested):
Overall looking good, just spotted the IdAnchor needed updating

src/components/IdAnchor.tsx
wackerow: #13980

pettinarip: Ahh nice! good catch @wackerow thanks for the PR!

src/components/MarkdownImage.tsx
```diff
@@ -42,20 +41,25 @@ const MarkdownImage = ({
...
-    <Flex as="span" justify="center">
-      <Link href={transformedSrc} target="_blank" rel="noopener" locale={false}>
-        <Image
+    <span className="flex justify-center">
+      <InlineLink
+        href={transformedSrc}
+        target="_blank"
+        rel="noopener"
+        locale={false}
+      >
```
wackerow, line 50: I'm not sure we need the `target` and `rel` any more.... we're using `ui/Link` now which should parse this with `const isExternal = url.isExternal(href)`

```suggestion
      <InlineLink href={transformedSrc} locale={false}>
```

wackerow: Never mind! I see we're feeding local images to this so this override is needed. Approving =)

wackerow (approved):
Looks good! Thanks @pettinarip!

PR #13953, opened 2024-09-24
theme: use 16px font for alert descriptions

wackerow:
## Description
Proposal: Use 16px font size as default for `ui/alert` component

## Example screenshot
image

## Preview link
https://deploy-preview-13953--ethereumorg.netlify.app/en/get-eth
https://ethereum.org/en/get-eth

### DIFF
--- a/src/components/ui/alert.tsx
+++ b/src/components/ui/alert.tsx
@@ -67,7 +67,7 @@ const AlertDescription = React.forwardRef<
 >(({ className, ...props }, ref) => (
   <div
     ref={ref}
-    className={cn("text-sm [&_p]:leading-relaxed", className)}
+    className={cn("[&_p]:leading-relaxed", className)}
     {...props}
   />
 ))


pettinarip (approved):
Approving but this needs Design approval as well.

cc @nloureiro @konopkja 

Here is the Storybook where you can see the new Alerts with the font size change
https://63b7ea99632763723c7f4d6b-aaqognawns.chromatic.com/?path=/story/molecules-action-feedback-alerts--variants

Now both, the Title and Description, have 16px. I think I would prefer to keep the Title a bit bigger in size, maybe 18px?

image

nloureiro:
> Approving but this needs Design approval as well.
> 
> cc @nloureiro @konopkja
> 
> Here is the Storybook where you can see the new Alerts with the font size change https://63b7ea99632763723c7f4d6b-aaqognawns.chromatic.com/?path=/story/molecules-action-feedback-alerts--variants
> 
> Now both, the Title and Description, have 16px. I think I would prefer to keep the Title a bit bigger in size, maybe 18px?
> 
> image


This is what I have on the DS > The only difference is that the yellow one uses the default (body) text color on all the text. For contrast purposes.

Screen Shot 2024-09-25 11 54 25 AM

nloureiro:
@wackerow one more thing based on your feedback on figma > bump the text to 16px

wackerow:
Okay. That is the only change in this PR (16px font); was not meant to address anything color related. If that looks good to you @nloureiro do me a favor and slap an approval on here 😄

wackerow:
(For the content of our `ui/alert` component)

nloureiro (approved):

pettinarip (approved):

PR #13955, opened 2024-09-25
refactor: components using old theming

wackerow:
## Description
- QuizRadioGroup buttons
- SendFromContacts
- ReceiveEther
- ReceivedEther

Updated to tailwind; colors updated to use latest theming

## Related Issue
Theming updates


src/components/Quiz/QuizWidget/QuizRadioGroup.tsx
```diff
@@ -4,7 +4,6 @@ import {
   Box,
   chakra,
   ChakraProps,
-  Circle,
   HStack,
   Stack,
   Text,
```
pettinarip, line 9: Is it possible to migrate this component completely to Tw? I feel that we only need to change a few components and avoid leaving this in a hybrid situation.

pettinarip (changes requested):
Looks good @wackerow thanks for the migration of these components. Left a small request.

src/components/Quiz/QuizWidget/QuizRadioGroup.tsx
wackerow: @pettinarip I started that and it led to a few complications that I didn't anticipate. For the sake of keeping scope of this PR trimmed to just the color updates, mind if I branch off and patch this up separately?

pettinarip: Got it. Sure, we can tackle that separately.

pettinarip (approved):

PR #13956, opened 2024-09-25
fix: consensus clients table formatting

alphabt:
## Description

The [consensus clients table](https://ethereum.org/en/developers/docs/nodes-and-clients/#consensus-clients) was not rendering correctly due to misplaced `|` character at the end.

The table prior to the fix looks like this:

image


alphabt:
@wackerow, @corwintines, @pettinarip, @minimalsm could you help review this simple PR? Thanks :)

corwintines (approved):

PR #13957, opened 2024-09-25
Shadcn migration - ContributorsQuizBanner 

saurabhburade:
## Description

Migrate `ContributorsQuizBanner` to tailwind

### DIFF
--- a/src/components/Banners/ContributorsQuizBanner.tsx
+++ b/src/components/Banners/ContributorsQuizBanner.tsx
@@ -1,7 +1,7 @@
 import { cn } from "@/lib/utils/cn"
 
-import { ButtonLink } from "../Buttons"
 import { TwImage } from "../Image"
+import { ButtonLink } from "../ui/buttons/Button"
 import { Flex, Stack } from "../ui/flex"
 
 import PeopleLearning from "@/public/images/people-learning.png"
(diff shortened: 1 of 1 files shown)


saurabhburade:
Hi @pettinarip, I need a suggestion about the `PeopleLearning` image. It's visually unused and if we use it where & how to place it.
For reference

Without image ==> Already in prod
Screenshot 2024-09-25 at 12 26 46 PM

With Image
Screenshot 2024-09-25 at 12 27 13 PM

pettinarip:
> Hi @pettinarip, I need a suggestion about the `PeopleLearning` image. It's visually unused and if we use it where & how to place it. For reference

You caught a bug!

Looking at this, the problem is that the `className` is being overridden. We need to destructure it from the props and pass it to the `cn` function to make it work properly.

```tsx
function ContributorsQuizBanner({
  className,
  ...props
}: React.HTMLAttributes<HTMLDivElement>) {
  return (
    <aside
      className={cn(
        "flex flex-col rounded md:flex-row",
        "bg-gradient-to-r from-accent-a/10 to-accent-c/10 dark:from-accent-a/20 dark:to-accent-c-hover/20",
        className
      )}
      {...props}
```

With that you should see the image and continue with the migration (and fix a bug xD).

pettinarip (approved):
Great! @saurabhburade lgtm

PR #13959, opened 2024-09-25
Shadcn migration - heros

pettinarip:
## Description

Migrates the ContentHero and MdxHero component to TW.

### DIFF
--- a/src/components/Hero/ContentHero/index.tsx
+++ b/src/components/Hero/ContentHero/index.tsx
@@ -1,42 +1,22 @@
-import { Box, Heading, SimpleGrid, Stack, Text } from "@chakra-ui/react"
-
 import type { CommonHeroProps } from "@/lib/types"
 
 import Breadcrumbs from "@/components/Breadcrumbs"
-import { Image } from "@/components/Image"
+import { TwImage } from "@/components/Image"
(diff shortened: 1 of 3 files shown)


src/components/Hero/ContentHero/index.tsx
```diff
@@ -1,42 +1,22 @@
...
 
 const ContentHero = (props: ContentHeroProps) => {
-  const {
-    breadcrumbs,
-    heroImg,
-    buttons,
-    title,
-    description,
-    blurDataURL,
-    maxHeight,
```
pettinarip, line 20: Removed this prop as it was not being used and was a bit hacky to set up with tw.

src/components/Hero/MdxHero/index.tsx
wackerow, line 1: @pettinarip Are we using this component anywhere? I don't see any instances of it

wackerow (approved):
Looks good! Did leave a note about MdxHero though... are we using this somewhere or should we remove it?

src/components/Hero/MdxHero/index.tsx
pettinarip: We are not atm. But it is planned to be used once we implement the new DS styles on the docs and other pages I believe.

Issue #13946, opened 2024-09-24
Shadcn/Tailwind migration


pettinarip:
## Description

As part of our [Q3 roadmap](https://github.com/ethereum/ethereum-org-website/issues/13399), this issue tracks the progress of the Shadcn/Tailwind migration.

## How to contribute

If you’d like to help with the component migration, please follow these steps:

1. Read the following “Tips & Guidance” section for basic things to keep in mind while doing the migration.
2. Leave a comment here asking which component you would like to work out.
3. We assign you to the file by adding your handle next to the file name in the list.
    - **If you were assigned to multiple files** then create a separate PR for each component. It will make the testing and review process much simpler and more organized.
4. Once you finish it (PR merged), we’ll mark it as done.
5. Repeat!

## Tips & Guidance

- Migrate only the components defined in the component file.
- Replace all Chakra imports (unless there is an exception mentioned in this file) with native tags or shadcn components installed in `components/ui`
    - `@chakra-ui/react/Icon` ⇒ just use the exported svg component from react-icons instead `<MdMenu className="text-2xl" />`
- Replace other imports
    - OldText ⇒ use the native `<p>` tag
    - OldHeading ⇒ use native `<h1>`, `<h2>`, etc. tags
    - Link ⇒ use the `Link` component defined in `components/ui/Link`
    - Buttons ⇒ use the buttons defined in `components/ui/buttons/Button`
    - images ⇒ use `TwImage` from `components/Image`
    - Chakra Badge => new `Tag` component in `/ui` folder
- For layout and general component structure
    - Use the components in `components/ui/flex` as much as possible
- Colors
    - Define missing colors in the `tailwind.config.ts` file. At the review stage, we will analyze & try to replace old theme colors with new colors from the Design System.

## Done

<details>
  <summary>List</summary>
<ul>
<li>AssetDownload/index.tsx, AssetDownloadArtist.tsx and AssetDownloadImage.tsx @saurabhburade #13960 </li>
<li>Banners/ContributorsQuizBanner.tsx @saurabhburade #13957 </li>
<li>BeaconChainActions.tsx @saurabhburade #13990 </li>
<li>BugBountyCards.tsx @saurabhburade #14000 </li>
<li>Contributors.tsx @Baystef #13989 </li>
<li>DataProductCard.tsx @Baystef #14035 </li>
<li>DevconGrantsBanner.tsx @saurabhburade #14012 </li>
[...]

PR #13961, opened 2024-09-25
Migrate MatomoOptOut to ShadCN

minimalsm:
## Description

Migrates `MatomoOptOut` component to ShadCN from Chakra

## Note

As far as I can tell, the current border color for the box in Chakra theme (#e5e5e5) does not exist in DS/Tailwind theme. I've used `body-light` as the closest, but defer to @nloureiro here :) 

image

## Updated Note

Per @pettinarip's [comment here](https://github.com/ethereum/ethereum-org-website/pull/13961#discussion_r1775316271), I've also stopped using the `OldText` component. This means the red color is slightly, though unrecognisably, different (now using `error` from tailwind config instead of the `fail` color in Chakra theme), and the margin bottom (was 1.45rem) is now 1.25rem per our scale.

## Related Issue

#13946

### DIFF
--- a/src/components/MatomoOptOut.tsx
+++ b/src/components/MatomoOptOut.tsx
@@ -1,9 +1,10 @@
 import { useEffect, useState } from "react"
-import { Checkbox, Flex } from "@chakra-ui/react"
 
 import Text from "@/components/OldText"
 
 import { MATOMO_LS_KEY } from "@/lib/utils/matomo"
+
+import Checkbox from "../../tailwind/ui/Checkbox"
 
 const MatomoOptOut = () => {
   const [loading, setLoading] = useState<boolean>(true)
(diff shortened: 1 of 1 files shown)


src/components/MatomoOptOut.tsx
```diff
@@ -1,10 +1,11 @@
 import { useEffect, useState } from "react"
-import { Checkbox, Flex } from "@chakra-ui/react"
 
 import Text from "@/components/OldText"
```
pettinarip, line 3: This one needs to be replaced by the native `<p>` tag

pettinarip (changes requested):
@minimalsm left a small req, the rest is ok 👍🏼

src/components/MatomoOptOut.tsx
minimalsm: Done and added note above.

pettinarip (approved):
🚀 🚀 🚀 🚀

PR #13966, opened 2024-09-25
Migrate EventCard to Shadcn

minimalsm:
## Description

- Updates EventCard to shadcn

## Notes

1. Hardcoded 1.6rem line height for the description text (per OldText) as the defaults we have varied too much from the current design. Our designers should pull this component in line with current DS.

2. Slight differences in spacing compared to the original design (which had kind of unusual spacing). IMO, these are improvements but defer to @nloureiro

3. I've hardcoded the background colors (`bg-[#FCFCFC] dark:bg-[#272627]`) in the CardHeader. Our default theme values didn't provide a good match for this use case. another area where we might want to consult with design

## Related Issue

#13946

### DIFF
--- a/src/components/EventCard.tsx
+++ b/src/components/EventCard.tsx
@@ -3,6 +3,8 @@ import { useRouter } from "next/router"
 import { useTranslation } from "next-i18next"
 import { BsCalendar3 } from "react-icons/bs"
 
+import { cn } from "@/lib/utils/cn"
+
 import type { EventCardProps } from "@/lib/types"
 
 import { ButtonLink } from "@/components/ui/buttons/Button"
--- a/src/components/EventCard.tsx
+++ b/src/components/EventCard.tsx
@@ -35,7 +35,7 @@ const EventCard: React.FC<EventCardProps> = ({
 
   return (
     <Card
-      className={`border-lightBorder flex h-full flex-col rounded-md border ${className}`}
+      className={cn("border-lightBorder flex h-full flex-col rounded-md border", className)}
     >
       <CardHeader className="flex flex-row items-center justify-center rounded-t-md border-b border-primary bg-[#FCFCFC] p-2 dark:bg-[#272627]">
         <BsCalendar3 className="mr-2 h-6 w-6 text-primary" />
--- a/src/components/EventCard.tsx
+++ b/src/components/EventCard.tsx
@@ -59,7 +59,7 @@ const EventCard: React.FC<EventCardProps> = ({
           <h3 className="text-xl font-bold md:text-2xl">{title}</h3>
           <span className="text-sm opacity-60">{location}</span>
         </div>
-        <p className="text-sm leading-[1.6rem]">{description}</p>
+        <p className="md:text-sm md:leading-[1.6rem]">{description}</p>
       </CardContent>
       <CardFooter className="p-4 pt-0">
         <ButtonLink href={href} variant="outline" className="w-full text-sm">
--- a/src/components/EventCard.tsx
+++ b/src/components/EventCard.tsx
@@ -35,7 +35,7 @@ const EventCard: React.FC<EventCardProps> = ({
 
   return (
     <Card
-      className={cn("border-lightBorder flex h-full flex-col rounded-md border", className)}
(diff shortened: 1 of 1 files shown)


nloureiro:
agree. The spacing was off. Thank you for the adjustments 

comapring with Figma, the text size is bigger, but now we are bumping everything to 16 or above so let's keep it like this

Screen Shot 2024-09-25 11 17 39 PM

With this text size, we should do a 2-card line. It's too narrow for an MD page. Would it be on another PR to do this?

Screen Shot 2024-09-25 11 22 51 PM

minimalsm:
Thanks @nloureiro 

> 
> 
> agree. The spacing was off. Thank you for the adjustments 
> 
> 
> 
> comapring with Figma, the text size is bigger, but now we are bumping everything to 16 or above so let's keep it like this
> 

Okay can update to 16px fontsize. With that we can probably normalise line height too?

> 
> With this text size, we should do a 2-card line. It's too narrow for an MD page. Would it be on another PR to do this?
> 
> 

We can handle that in #13967, which is the upcoming events grid wrapper

nloureiro:
> Thanks @nloureiro
> 
> > agree. The spacing was off. Thank you for the adjustments
> > comapring with Figma, the text size is bigger, but now we are bumping everything to 16 or above so let's keep it like this
> 
> Okay can update to 16px fontsize. With that we can probably normalise line height too?

yes, makes sense
> 
> > With this text size, we should do a 2-card line. It's too narrow for an MD page. Would it be on another PR to do this?
> 
> We can handle that in #13967, which is the upcoming events grid wrapper

ok, got it

minimalsm:
Updated font size implementation after discussion with @nloureiro:

Mobile: 16px font size
Desktop: Retaining 14px font size

Rationale:

- 3-column grid on large screens necessitates smaller font due to limited width
- 2-column grid on large screens was tested but felt clunky and reduced event discoverability

Changes implemented && ready for review :)

src/components/EventCard.tsx
```diff
@@ -28,97 +21,52 @@ const EventCard = ({
...
-    <Box
-      className={className}
-      position="relative"
-      mt="0"
-      _before={clearStyles}
-      _after={clearStyles}
-      w="full"
-      h="full"
+    <Card
+      className={`border-lightBorder flex h-full flex-col rounded-md border ${className}`}
```
pettinarip, line 38: We should use the `cn` utility to merge classes correctly.
```suggestion
      className={cn("border-lightBorder flex h-full flex-col rounded-md border", className)}
```

pettinarip (approved):
@minimalsm looks good, will add a small fix to use `cn` function.

src/components/EventCard.tsx
pettinarip, line 38: Ah, what did you mean with `lightBorder`?

minimalsm: No idea, just going to remove lol

src/components/EventCard.tsx
```diff
@@ -28,97 +21,52 @@ const EventCard = ({
...
-    <Box
-      className={className}
-      position="relative"
-      mt="0"
-      _before={clearStyles}
-      _after={clearStyles}
-      w="full"
-      h="full"
+    <Card
+      className={cn("border-lightBorder flex h-full flex-col rounded-md border", className)}
```
minimalsm, line 38:
```suggestion
      className={cn("flex h-full flex-col rounded-md border", className)}
```

src/components/EventCard.tsx
```diff
@@ -2,24 +2,17 @@ import React from "react"
 import { useRouter } from "next/router"
 import { useTranslation } from "next-i18next"
 import { BsCalendar3 } from "react-icons/bs"
-import { Box, Flex, Heading, Icon } from "@chakra-ui/react"
 
 import type { EventCardProps } from "@/lib/types"
```
minimalsm, line 6:
```suggestion
import { cn } from "@/lib/utils/cn"

import type { EventCardProps } from "@/lib/types"
```

pettinarip (approved):

PR #13967, opened 2024-09-25
Migrate UpcomingEventsList to Shadcn

minimalsm:
## Description

Migrates `UpcomingEventList` to Shadcn.

Note, before the H3 had a line-height of 1.4 (not really sure why). This isn't in our system, so I just left it as default line height.

## Related Issue

#13946 #13966

### DIFF
--- a/src/components/UpcomingEventsList.tsx
+++ b/src/components/UpcomingEventsList.tsx
@@ -104,14 +104,14 @@ const UpcomingEventsList = () => {
   }
 
   return (
-    <div className="">
+    <div>
       {Object.keys(monthGroupedEvents)
         ?.slice(0, maxRange)
         ?.map((month) => {
           const events = monthGroupedEvents[month]
           return (
             <div key={`events_in_${month}`}>
-              <h3 className="py-8 text-2xl font-semibold">{month}</h3>
+              <h3 className="py-8">{month}</h3>
               <div className="grid grid-cols-1 gap-5 md:grid-cols-2 lg:grid-cols-3">
                 {events.map(
                   (
--- a/src/components/UpcomingEventsList.tsx
+++ b/src/components/UpcomingEventsList.tsx
@@ -7,8 +7,8 @@ import type { CommunityConference, Lang } from "@/lib/types"
 
 import EventCard from "@/components/EventCard"
 import InfoBanner from "@/components/InfoBanner"
-import InlineLink from "@/components/Link"
 import { Button } from "@/components/ui/buttons/Button"
+import Link from "@/components/ui/Link"
 
 import { trackCustomEvent } from "@/lib/utils/matomo"
 import { getLocaleTimestamp } from "@/lib/utils/time"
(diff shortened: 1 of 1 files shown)


src/components/UpcomingEventsList.tsx
```diff
@@ -2,14 +2,13 @@ import { useEffect, useState } from "react"
 import _ from "lodash"
 import { useRouter } from "next/router"
 import { useTranslation } from "next-i18next"
-import { Box, Grid, Heading } from "@chakra-ui/react"
 
 import type { CommunityConference, Lang } from "@/lib/types"
 
-import { Button } from "@/components/Buttons"
 import EventCard from "@/components/EventCard"
 import InfoBanner from "@/components/InfoBanner"
```
minimalsm, line 9: @pettinarip I see you used [the Alert here](https://github.com/ethereum/ethereum-org-website/pull/13597/commits/bee765c69acbead6caebe339e5502bb915461709) for InfoBanner migration. InfoBanner is used over a thousand of times in our markdown files, so it is probably best we just migrate the component itself. What do you think?

src/components/UpcomingEventsList.tsx
```diff
@@ -2,14 +2,13 @@ import { useEffect, useState } from "react"
...
 import InlineLink from "@/components/Link"
```
minimalsm, line 10: Wasn't sure what to do here, assuming we'll migrate this separately but might be wrong as not on list.

src/components/UpcomingEventsList.tsx
```diff
@@ -164,32 +139,19 @@ const UpcomingEventsList = () => {
...
                   )
                 )}
-              </Grid>
-            </Box>
+              </div>
+            </div>
           )
         })}
 
-      <Box
```
minimalsm, line 172: Did this do something or just a mistake?

src/components/UpcomingEventsList.tsx
pettinarip: Correct, thats the plan for InfoBanner. We are going to tackle that in a differnt PR

src/components/UpcomingEventsList.tsx
pettinarip: We should use the Link that is in `components/ui/Link`. Not a big deal though.

src/components/UpcomingEventsList.tsx
```diff
@@ -109,35 +104,15 @@ const UpcomingEventsList = () => {
   }
 
   return (
-    <>
+    <div className="">
```
pettinarip, line 107:
```suggestion
    <div>
```

src/components/UpcomingEventsList.tsx
```diff
@@ -109,35 +104,15 @@ const UpcomingEventsList = () => {
...
-                  lg: "repeat(3,1fr)",
-                }}
-                justifyContent={"center"}
-                alignItems={"center"}
-                alignSelf={"center"}
-                gap={5}
-                overflow={"hidden"}
-              >
+            <div key={`events_in_${month}`}>
+              <h3 className="py-8 text-2xl font-semibold">{month}</h3>
```
pettinarip, line 114: Based on your comment, I agree, there are a lot of old styles that do not match the DS.

Since you are adapting the line height, we could also just use the standard `h3` size that matches the DS styles.

```suggestion
              <h3 className="py-8">{month}</h3>
```

src/components/UpcomingEventsList.tsx
pettinarip: Hmm good question xD I don't think this is doing anything visually, its adding just a bit of space...I think it is safe to remove it

src/components/UpcomingEventsList.tsx
minimalsm: Okay, replaced it

pettinarip (approved):
Awesome, thanks @minimalsm

PR #13971, opened 2024-09-26
Shadcn migration - button dropdown

pettinarip:
## Description

Migrates the ButtonDropdown using the Shadcn dropdown-menu component.


src/components/MdComponents/index.tsx
wackerow, line 1: 🔥

wackerow (approved):
Looks good, thanks @pettinarip!

PR #13972, opened 2024-09-26
Shadcn migration - tags

pettinarip:
## Description

Migrates the Tag component.


src/styles/colors.css
wackerow, line 1: Good catch! I've updated the labels for these in the Colors page in Figma DS

src/components/ui/tag.tsx
```diff
@@ -0,0 +1,88 @@
...
+        tag: "bg-primary-low-contrast text-primary border-primary",
+        success: "bg-success-light text-success border-success-border",
+        error: "bg-error-light text-error border-error-border",
+        warning: "bg-warning-light text-warning-dark border-warning-border",
+      },
+      variant: {
+        subtle: "border-transparent",
+        solid: "border-transparent text-body-inverse",
+        outline: "bg-transparent",
+      },
```
wackerow, line 21: Any reason we aren't adding the `highContrast` variant shown in DS?

image

src/components/ui/tag.tsx
wackerow, line 1: Looks like we're missing the hover styles?

Also, just checking the focus state (from storybook):

image

Not sure I see this styling in the DS

wackerow (changes requested):

src/components/ui/tag.tsx
pettinarip: Ahh I completely miss this case. ill add it.

src/styles/colors.css
pettinarip: Credits to @nloureiro, he updated these values.

src/components/ui/tag.tsx
pettinarip: Right, I have only implemented the basic info tag version. I've skipped the button version since I have a few doubts about it and since we are not currently using that version on the site, I ignored it.

To clarify, this info version is not a button, so it is not focusable.

pettinarip: What I was thinking was the idea of exporting a `TagCloseButton` since the styling of that button can be a bit challenging.

pettinarip: Created a new PR #13999 to tackle the button version. This will unblock this PR and allow us to keep the Tag migration on the codebase with the basic read-only version.

defihook (commented):
great work
thanks

src/styles/colors.css
```diff
@@ -72,8 +72,8 @@
     --yellow-500: 38, 92%, 50%; /* #F59E0B */
     --yellow-600: 32, 95%, 44%; /* #D97706 */
     --yellow-700: 26, 90%, 37%; /* #B45309 */
-    --yellow-800: 23, 83%, 31%; /* #92400E */
-    --yellow-900: 22, 78%, 26%; /* #78350F */
+    --yellow-800: 23, 88%, 22%; /* #6B2D07 */
+    --yellow-900: 22, 84%, 15%; /* #471E06 */
```
wackerow, line 76:
```suggestion
    --yellow-900: 22, 84%, 15%; /* #471E06 */
```

wackerow (approved):

PR #13979, opened 2024-09-26
chore: import homepage bucket for all locales

wackerow:
## Description
- Update all locales with latest from crowdin, post-homepage updates

## Related Issue
- #13528


pettinarip (approved):
@wackerow noticed that something goes wrong in the marquee section when you select non-english locales. For example with `ar`.
image

Seems like we are missing one repetition of the moving arrays.

Not blocker thought since this is probably not related to this PR.

wackerow:
Yes! Noticed this, likely related to the translateX transform. will get that patched separately

PR #13985, opened 2024-09-27
Theme - add border colors

pettinarip:
## Description

Adds all the new border colors from the DS. cc @nloureiro


minimalsm (approved):
PR looks good.

@nloureiro should lowContrast in dark mode be changed with the background color change we made?

nloureiro (approved):

nloureiro:
yes, it should. good catch.
I will open a PR for that later.

PR #13987, opened 2024-09-27
 Migrate layer 2 page to shadcn, and updates for layer 2 migration 

corwintines:
## Description

Migrates layer 2 page to shadcn
Move layer 2 page to /layer-2/learn
Updates for the layer 2 migration to this page

Link: https://deploy-preview-13987--ethereumorg.netlify.app/layer-2/learn


public/images/layer-2/learn-hero.png
wackerow, line 1: image

Noticing this is an image the chops on the top and bottom edges... looks a little strange having padding in the Y-direction here inside the hero cc: @nloureiro

src/components/Callout.tsx
```diff
@@ -24,48 +25,52 @@ const Callout = ({
...
+      className={cn(
+        "mb-16 mt-32 flex flex-1 flex-col rounded bg-gradient-to-br from-[rgba(127,127,213,0.2)] via-[rgba(134,168,231,0.2)] to-[rgba(145,234,228,0.2)] p-12 lg:mb-4",
+        className
+      )}
     >
       {image && (
-        <Center maxW="263px" minH="200px" mt={-40} alignSelf="center">
-          <Image src={image} alt={alt || ""} height={200} />
-        </Center>
+        <div className="mt-[-10rem] self-center">
```
wackerow, line 43:
```suggestion
        <div className="-mt-40 self-center">
```

src/components/Callout.tsx
```diff
@@ -24,48 +25,52 @@ const Callout = ({
...
+          {descriptionKey && (
+            <p className="mb-6 text-xl leading-[140%] text-body-medium">
+              {t(descriptionKey)}
+            </p>
+          )}
+          {description && (
+            <p className="mb-6 text-xl leading-[140%] text-body-medium">
+              {description}
+            </p>
+          )}
```
wackerow, line 69: I'm a little unclear why we're opening up the props to optionally accept `title` and `description` or the keys for these:

image

Is the plan to change these components to not have to handle the translation keys? As-is we not longer have any of the props as required.

src/lib/utils/translations.ts
```diff
@@ -168,10 +168,15 @@ const getRequiredNamespacesForPath = (relativePath: string) => {
     requiredNamespaces = [...requiredNamespaces, "page-wallets"]
   }
 
+  // TODO: Remove this when the page is translated
   if (path.startsWith("/layer-2/")) {
     primaryNamespace = "page-layer-2"
   }
```
wackerow, line 174: Just noting that as-is the `page-layer-2` namespace will be loaded on `/layer-2/learn/`; is this what we want? If not, I'd change this one to not use `startsWith()`

src/intl/en/page-layer-2-learn.json
```diff
@@ -0,0 +1,55 @@
+{
+  "page-layer-2-learn-meta-title": "What is layer 2",
+  "page-layer-2-learn-title": "What is layer 2",
+  "page-layer-2-learn-description": "Scaling Ethereum for mass adoption",
+  "page-layer-2-learn-button-1-label": "What is layer 2",
+  "page-layer-2-learn-button-2-label": "Use layer 2",
+  "page-layer-2-learn-what-is-layer-2-title": "What is layer 2?",
```
wackerow, line 7: Seeing 4 instances of "What is layer 2"... one of which has a question mark

1. Should we have a question mark at the end of this string? (I think we should)
2. Do we need to repeat this key, or can we simplify to just have one instance of "What is layer 2?"

wackerow (changes requested):
Nice @corwintines! Lookin good, just dropped an initial round of feedback and thoughts.

defihook (commented):
great work
thanks

defihook (commented):
Looks good

PR #13989, opened 2024-09-27
Migrate Contributors component to shadcn

Baystef:
I added a new CSS variable `--table-background-hover` to replace `tableBackgroundHover` as I didn't see the equivalent in the new variables.

~Also, we might need to create a custom LinkBox/LinkOverlay component~
I added custom LinkBox/LinkOverlay component

## Description
On Localhost
Screenshot 2024-09-27 at 5 42 14 PM
Screenshot 2024-09-27 at 5 42 33 PM

On Production
Screenshot 2024-09-27 at 5 40 38 PM
Screenshot 2024-09-27 at 5 40 52 PM

## Related Issue
#13946


defihook (commented):
great work
thanks

src/components/Contributors.tsx
```diff
@@ -31,52 +31,31 @@ const Contributors = () => {
...
-            }}
-            _focus={{
-              textDecoration: "none",
-              borderRadius: "base",
-              boxShadow: "0px 8px 17px rgba(0, 0, 0, 0.15)",
-              background: "tableBackgroundHover",
-              transition: "transform 0.1s",
-              transform: "scale(1.02)",
-            }}
+            className="m-2 max-w-[132px] transform shadow-table transition-transform duration-100 hover:scale-[1.02] hover:rounded hover:bg-background-table-hover hover:no-underline hover:shadow-table-box-hover focus:scale-[1.02] focus:rounded focus:no-underline focus:shadow-table-box-hover"
```
pettinarip, line 38: We can deprecate `tableBackgroundHover` and use `bg-background-highlight` for this. Would help on cleaning these styles a bit.

We can also remove the shadow effects and the underline wouldn't be necessary since we are overriding that on the link itself.

```suggestion
            className="m-2 max-w-[132px] transform shadow transition-transform duration-100 hover:scale-[1.02] hover:rounded hover:bg-background-highlight focus:scale-[1.02] focus:rounded"
```

tailwind.config.ts
```diff
@@ -152,6 +152,9 @@ const config = {
           low: "hsla(var(--background-low))",
           medium: "hsla(var(--background-medium))",
           high: "hsla(var(--background-high))",
+          table: {
+            hover: "var(--table-background-hover)",
+          },
```
pettinarip, line 157: We can replace this with `background-highlight`.

src/components/ui/link-box.tsx
pettinarip, line 1: Nice implementation! this looks good. Thanks for doing that. Will keep this in mind for other components.

pettinarip (changes requested):
@Baystef the PR looks great, the LinkBox implementation is really helpful.

I've left one small request on the colors.

tailwind.config.ts
Baystef: I'll do that shortly

src/components/Contributors.tsx
Baystef: This is noted

pettinarip (approved):
Thanks @Baystef! looks great now.

PR #13991, opened 2024-09-28
Add ETH Pura Vida to Event list

aguzmant103:
## Description

Add ETH Pura Vida to Event list


defihook (commented):
great work
thanks

corwintines (approved):
Thanks @aguzmant103

PR #13993, opened 2024-09-28
Update index.md

AkashMundari:
## Description

Changed "defraud" to "fraud" in line 15


corwintines:
Thanks @AkashMundari, but defraud is the correct word here so going to close this out.

PR #13997, opened 2024-09-29
Removed deprecated eth_coinbase from json-rpc docs[#13996]

### LINKED ISSUE #13996: Deprecated eth_coinbase
### Describe the bug

As per [ethereum/go-ethereum#29747](https://github.com/ethereum/go-ethereum/issues/29747), `eth_coinbase` no longer returns an address in v1.14.0+.
Since this RPC method is deprecated, we should remove it from JSON-RPC API [list](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_coinbase).

### To reproduce

We can see `eth_coinbase` in the RPC list.
https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_coinbase

### Expected behavior

We should not list the unsupported RPC APIs.

### Screenshots

_No response_

_No response_

_No response_

### Additional context

In Deploying a contract [example](https://ethereum.org/en/developers/docs/apis/json-rpc/#deploying-contract), its been mentioned `eth_coinbase` for getting the coinbase address, instead we can call it as developer account and use eth_accounts to retrieve it.

- [X] Yes
- [ ] No

MaryNfs:
## Description

As discussed in [ethereum/go-ethereum#29747](https://github.com/ethereum/go-ethereum/issues/29747), eth_coinbase no longer returns an address in v1.14.0+ and has been deprecated.
This PR removes the deprecated eth_coinbase RPC method from the JSON-RPC API list.

## Related Issue
- Fixes #13996

### DIFF
--- a/public/content/developers/docs/apis/json-rpc/index.md
+++ b/public/content/developers/docs/apis/json-rpc/index.md
@@ -383,31 +383,6 @@ curl -X POST --data '{"jsonrpc":"2.0","method":"eth_syncing","params":[],"id":1}
 }
 ```
 
-### eth_coinbase {#eth_coinbase}
-
-Returns the client coinbase address.
-
-**Parameters**
-
-None
-
-**Returns**
-
-`DATA`, 20 bytes - the current coinbase address.
-
-**Example**
-
-```js
-// Request
-curl -X POST --data '{"jsonrpc":"2.0","method":"eth_coinbase","params":[],"id":64}'
-// Result
-{
-  "id":64,
-  "jsonrpc": "2.0",
-  "result": "0x407d73d8a49eeb85d32cf465507dd71d507100c1"
-}
-```
-
 ### eth_chainId {#eth_chainId}
 
 Returns the chain ID used for signing replay-protected transactions.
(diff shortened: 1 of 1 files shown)


public/content/developers/docs/apis/json-rpc/index.md
```diff
@@ -1671,10 +1646,10 @@ geth --http --dev console 2>>geth.log
 
 This will start the HTTP RPC interface on `http://localhost:8545`.
 
-We can verify that the interface is running by retrieving the Coinbase address and balance using [curl](https://curl.se). Please note that data in these examples will differ on your local node. If you want to try these commands, replace the request params in the second curl request with the result returned from the first.
+We can verify that the interface is running by retrieving the Coinbase address(by obtaining the first address from the array of accounts) and balance using [curl](https://curl.se). Please note that data in these examples will differ on your local node. If you want to try these commands, replace the request params in the second curl request with the result returned from the first.
```
wackerow, line 1649: 1. I know you didn't change this, but we should use lowercase `coinbase` here
2. Need a space

```suggestion
We can verify that the interface is running by retrieving the coinbase address (by obtaining the first address from the array of accounts) and balance using [curl](https://curl.se). Please note that data in these examples will differ on your local node. If you want to try these commands, replace the request params in the second curl request with the result returned from the first.
```

wackerow (changes requested):
Thanks @MaryNfs! Agree with the issue you posted, but am hesitant to remove this entirely for archival/historical purposes.

I might suggest something similar to what the playground we link out to does:

https://ethereum-json-rpc.com/?method=eth_coinbase

If you go here you'll see that the `eth_coinbase` method is still available, but executing shows `"details": "Method not supported"`.

I'd suggest we keep the h3 section and initial comment, and add a note that this method has been deprecated, and note as of when this occured. Something like this:

```
### eth_coinbase {#eth_coinbase}

Returns the client coinbase address.

**No longer supported (deprecated as of _____)**
```

MaryNfs:
> I'd suggest we keep the h3 section and initial comment, and add a note that this method has been deprecated, and note as of when this occured. Something like this:
> 
> ```
> ### eth_coinbase {#eth_coinbase}
> 
> Returns the client coinbase address.
> 
> **No longer supported (deprecated as of _____)**
> ```

Hi @wackerow 
Thanks for the feedback!
I've made the changes as you suggested.
Let me know if there’s anything else you’d like me to adjust!

public/content/developers/docs/apis/json-rpc/index.md
```diff
@@ -387,6 +387,8 @@ curl -X POST --data '{"jsonrpc":"2.0","method":"eth_syncing","params":[],"id":1}
 
 Returns the client coinbase address.
 
+> **Note:** This method has been deprecated as of **v1.14.0** and is no longer supported. Attempting to use this method will result in a "Method not supported" error.
```
wackerow, line 390: > Attempting to use this method will result in a "Method not supported" error.

Should we consider replacing this sentence with a comment similar to below about using `eth_accounts` instead? Don't think it needs to block here, but curious thoughts for further iteration.

wackerow (approved):
Looks good, thanks @MaryNfs!

wackerow:
@all-contributors please add @MaryNfs for content bug patch

PR #14001, opened 2024-09-30
fix: apply flip to parent of svg

wackerow:
## Description
- apply RTL flip to parent of svg to avoid conflicting transitions on the `svg` element
- Fixes issue where hovering on RTL pages would flip the icon back to "LTR" orientation

## Related Issue
None filed


defihook (commented):
great work
thanks

corwintines (approved):

PR #14010, opened 2024-09-30
[Tailwind] add plugin to extend `inset` utility classes

TylerAPfledderer:
Adds plugin to the tailwind config to extend the `inset` utility to include utilities for `inset-inline-start` and `inset-inline-end`.

Currently, there is only `inset-x-*` to establish both the left and right position with a single value. 

Adding `inset-start-*` and `inset-end-*` generates utilities with existing spacing values that define only left or right position, and are `i18n` aware.

### DIFF
--- a/tailwind.config.ts
+++ b/tailwind.config.ts
@@ -309,6 +309,25 @@ const config = {
         }
       )
     }),
+    plugin(function ({ matchUtilities, theme }) {
+      /**
+       * Extend the `inset` utility by including `inset-start` and `inset-end`.
+       *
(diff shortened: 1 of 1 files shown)


defihook (commented):
great work
thanks

wackerow (commented):
Hey @TylerAPfledderer! `start-` and `end-` should handle this, no? https://tailwindcss.com/docs/top-right-bottom-left

Let me know if I'm missing something

TylerAPfledderer:
@wackerow closing. I overlooked the obvious. Thanks! 🤣🤦

PR #14018, opened 2024-10-01
migrate ProductCard component to tailwind

yashtotla:
Related issue – https://github.com/ethereum/ethereum-org-website/issues/13946


src/components/ProductCard.tsx
```diff
@@ -87,71 +82,53 @@ const ProductCard = ({
...
-      border="1px"
-      borderColor="lightBorder"
-      textDecoration="none"
-      _hover={{
-        transition: "transform 0.1s",
-        transform: "scale(1.02)",
-      }}
+      className={cn(
+        "text-text bg-searchBackground flex-col justify-between",
+        "shadow-product-card",
```
pettinarip, line 90: @nloureiro this card has a few custom shadows, Im thinking that perhaps we can simply remove them or replace them by one of the default TW shadows like `shadow-md` or `shadow-lg`. LMK what do you think.
You can see these cards used in https://ethereum.org/en/developers/local-environment/

src/components/ProductCard.tsx
```diff
@@ -87,71 +82,53 @@ const ProductCard = ({
...
-        background={bgProp}
-        boxShadow="inset 0px -1px 0px rgba(0, 0, 0, 0.1)"
-        minH="200px"
+        className="min-h-[200px] shadow-product-card-img-container"
+        style={{ backgroundColor: bgProp }}
       >
-        <Image
-          src={image}
-          alt={alt}
-          height="100"
```
pettinarip, line 121: We still need this to tell next image the size of the image.

pettinarip (changes requested):
@yashtotla thanks for the PR.

Left a few comments. I've also tagged @nloureiro for design approval since there are a few old colors and shadows that would be nice if we can replace them and match our new DS styles.

src/components/ProductCard.tsx
yashtotla: I will update the box shadow styles once I have more clarity. I have addressed the rest of your comments.

pettinarip (approved):
Thanks @yashtotla

PR #14019, opened 2024-10-02
Data loader

pettinarip:
## Description

Creates a data loader function to handle data caching, error handling and mock data.


pettinarip:
Close in favor of #14179

PR #14021, opened 2024-10-02
Shadcn migration - toc

pettinarip:
## Description

Migrates the ToC component using shadcn dropdown-menu.


wackerow (commented):
https://github.com/user-attachments/assets/8e6c5838-674d-4978-90b1-2878cd95af45

Looking good overall! Noticing we lost the scroll-tracking though. Were we looking to include that here?

src/components/TableOfContents/index.tsx
```diff
@@ -65,43 +57,29 @@ const TableOfContents = ({
...
-      top="19" // Account for navbar
-      p={4}
-      pe={0}
-      maxW="25%"
-      minW={48}
-      height={calc.subtract("100vh", "80px")}
-      overflowY="auto"
+    <aside
+      className={cn(
+        "sticky top-19 hidden h-[calc(100vh-80px)] min-w-48 max-w-[25%] flex-col items-start gap-4 overflow-y-auto p-4 pe-0 lg:flex",
```
wackerow, line 62: image

Per discussion in Discord, noted that the `max-w-[25%]` isn't actually aligned with the DS. Looks like we use a fixed with of 16rem, or `w-64`

```suggestion
        "sticky top-19 hidden h-[calc(100vh-80px)] w-64 flex-col items-start gap-4 overflow-y-auto p-4 pe-0 lg:flex",
```

wackerow (commented):
@pettinarip Left a comment about aligning the width with the DS... Not intending to scope creep though, so if you'd prefer we can just adjust this separately, will leave it up to you.

src/components/TableOfContents/index.tsx
minimalsm: should this be part of this PR? there may be potential about side effects.

wackerow: Can easily break it off as a separate DS task, sure.. don't see it as blocking. 

The regression to the scroll-tracking noted above I would still rather we fix before merging though.

wackerow (changes requested):
https://github.com/user-attachments/assets/8e6c5838-674d-4978-90b1-2878cd95af45

Regression to the scroll-tracking

TylerAPfledderer:
@pettinarip FYI I tagged this PR in the Notion page 👍🏼

src/components/TableOfContents/index.tsx
pettinarip: Good call. I'll do that in a separate PR to discuss potential issues with the design and move this forward.

pettinarip:
> Screen.Recording.2024-10-03.at.16.57.36.mov 
> Regression to the scroll-tracking

Good catch! I've fixed this.

pettinarip:
@wackerow this is ready to be reviewed again

wackerow (approved):
🔥

Afeter8 (approved):

Afeter8 (approved):

PR #14029, opened 2024-10-02
Import warning

pettinarip:
## Description

Adds a custom eslint warning when `react-i18next` is used and suggest to use `next-i18next` to avoid hydration issues.


wackerow (approved):
image

Beautiful ❤️

PR #14037, opened 2024-10-02
Fix: Long-string overflow [Fixes #14013]

### LINKED ISSUE #14013: overflow on a dev doc is breaking on small screens
### Describe the bug

on small screens, this page is broken because of an overflow fail

[https://ethereum.org/en/developers/docs/consensus-mechanisms/pow/](https://ethereum.org/en/developers/docs/consensus-mechanisms/pow/)

Screen Shot 2024-10-01 11 22 43 AM

### To reproduce

1. Go to [en/developers/docs/consensus-mechanisms/pow/#finality](https://ethereum.org/en/developers/docs/consensus-mechanisms/pow/)
2. resize the window to a smaller than 500px

3. See the page with horizontal scroll

### Expected behavior

should not have horizontal scroll  

### Screenshots

_No response_

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes
- [X] No

jorgezerpa:
## Description
I Created a new class ".long_ethereum_address" in global.css file. Such class has the property *word-break* set to *break-all* to allow the text to break to the next line if it is necessary.
 
Then I wrapped the address into an *span* element that has the mentionend class. 

Finally I checked the same page for other languages, and found that has the same error. So I repeated the same process on all related files (for each different translation).


wackerow (changes requested):
Hey @jorgezerpa, thanks for jumping on this.

I think this is good, just two things:

1. @lukassim Just noting this is updating some syntax (not content) across these pages if we can pull them to Crowdin when ready
2. @jorgezerpa We have TailwindCSS in ths repo, so let's ditch the custom class and just use the `break-all` class instead ([see docs](https://tailwindcss.com/docs/word-break#break-all))

Thanks again!

ie.
```html
<span class="break-all">`...long-string...`</span>
```

wackerow (commented):
cc: @pettinarip I think using a span here makes sense, but curious if you have any thoughts on this approach. I considered that we could apply this to the `pre` tag that the back-ticks will parse into, but this could result in un-intended bugs for other text that uses this syntax.

pettinarip:
> cc: @pettinarip I think using a span here makes sense, but curious if you have any thoughts on this approach. I considered that we could apply this to the `pre` tag that the back-ticks will parse into, but this could result in un-intended bugs for other text that uses this syntax.

The backtick will use a `code` tag, not the pre tag. So, I'd prefer to style `code` to accept break-word. Styling `pre` should not be necessary since we are already doing that within the Codeblock component.

In short, I would add a new component for `code` in our `MdComponents`, adding this style. What do you think?

jorgezerpa:
Hello @wackerow I'm happy to hear this works!

I replaced the custom class with the tailwind class "break-words". I tried "break-all" but seems to not have effect here🤔 so I tried with "break-words" which has a similar behavior and it works well!

wackerow:
@pettinarip Sorry, `code` not `pre`, but still... Personally don't think applying that on `code` or a custom `Code` is wise, since we rarely want to just chop a code snippet in the middle of a word (let me know if you disagree)... imo these long hashes are unique cases where we don't mind just letting it wrap anywhere.

I like the idea of a simple custom component, but I'd suggest not naming it "Code" and instead perhaps name it functionally as "BreakAll".

jorgezerpa:
@wackerow @pettinarip  I can try create a component "BreakAll" as you mention, please confirm if it's needed 🙌

nloureiro (approved):
From the design side looks good! Thank you!!! @jorgezerpa

corwintines:
https://github.com/ethereum/ethereum-org-website/pull/14050

I think this resolved this more globally, so going to close this out.

PR #14163, opened 2024-10-14
[ShadCN] Migrate Modal component to ShadCN/Tailwind

TylerAPfledderer:
## Description

Migrate `Modal` component to ShadCN to build off of the `Dialog` component.

Replaces only the instances of the `Modal` currently in prod.


pettinarip:
> Currently trying to leverage `tailwind-variants` to come up with a good approach to make the `Modal` styling a priority, and the other styles as variants. particularly with the placement of the close button.

Great! thanks for the update @TylerAPfledderer. I'll have a closer look tomorrow.

src/components/ui/dialog-modal.tsx
```diff
@@ -0,0 +1,227 @@
...
+  const { header, close } = useDialogStyles()
+  return (
+    <div className={cn(header(), className)} {...props}>
+      {children}
+      <Center className="absolute end-0 top-0" asChild>
+        <DialogPrimitive.Close className={close()}>
+          <MdClose />
+          <span className="sr-only">Close</span>
+        </DialogPrimitive.Close>
+      </Center>
```
pettinarip, line 117: To give full control, perhaps we could avoid forcing this and declare it as part of the `DialogClose` button.

Then, we could be more opinionated in the Modal structure and declare it as part of the header.

```tsx
<DialogHeader>
  <DialogTitle>{title}</DialogTitle>
  <Center className="absolute end-0 top-0" asChild>
    <DialogClose />
  </Center>
</DialogHeader>
```

pettinarip: We could use our `Button` to wrap the close button to inherit the styles, right?

```tsx
<DialogPrimitive.Close asChild>
  <Button ...>...
</DialogPrimitive.Close>
```

src/components/ui/dialog-modal.tsx
```diff
@@ -0,0 +1,227 @@
...
+  title?: React.ReactNode
+  onClose?: () => void
+  actionButton?: {
+    label: string
+    onClick: () => void
+  }
+  contentProps?: DialogContentProps
+}
+
+const Modal = ({
```
pettinarip, line 170: Nice. So, this export is to be backwards compatible with the previous instances, right?

Anyway, I think this abstraction makes sense to avoid repeating this code and have more consistency.

pettinarip: Was thinking if this was the correct place to put it or in `src/components/Modal.tsx`.

src/components/ui/dialog-modal.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+import { Button } from "./buttons/Button"
+import { Center, Flex } from "./flex"
+
+import { useDisclosure } from "@/hooks/useDisclosure"
+
+const dialogVariant = tv({
+  slots: {
+    content:
+      "data-[state=open]:animate-contentShow w-full fixed left-1/2 top-1/2 grid -translate-x-1/2 -translate-y-1/2 gap-4 rounded-md bg-white p-8 shadow-[hsl(206_22%_7%_/_35%)_0px_10px_38px_-10px,_hsl(206_22%_7%_/_20%)_0px_10px_20px_-15px] focus:outline-none",
```
pettinarip, line 16: `contentShow`? is that a built-in animation in TW?

src/components/ui/dialog-modal.tsx
TylerAPfledderer: Yea. Prop naming has changed slightly.

I currently have it here just to keep this isolated from the existing `Modal` file along with having all of this isolated from the existing `dialog` file while I try to figure out how all of it should be merged together.

src/components/ui/dialog-modal.tsx
TylerAPfledderer: That is a utility that came from Radix as I was copy/pasting some default code. I'll review this.

TylerAPfledderer:
@pettinarip Not sure what should be done here using this new modal for the wallet simulators. Per how logic works in radix between `DialogTrigger` and the dialog itself, it would require rebuilding the code structure of the simulator content to retain this internal logic and minimize the need for a controlled state from the parent.

I could try and keep the current way the simulator modals are triggered, but could be become async with the intended approach in using this modal in other areas.

Would at least keep this temporary `dialog-modal` file for now, while doing incremental transitioning, and ignore the simulator area for now.

At the time of this comment, I have not viewed other renders that are currently using the existing `Modal`.

TylerAPfledderer:
`Update`: Not sure what I should do here.

This component seems to require in its pattern to have the `DialogTrigger` in order for the open/close logic to function correctly. In this case, the modals related to the simulator, quiz list, `CodeModal`, and a couple of other locations would need to be updated to account for this. That should be a separate PR.

I could for now ignore this and maintain the controlled logic currently present. Not sure if there will be unintended results.

pettinarip:
> `Update`: Not sure what I should do here.
> 
> This component seems to require in its pattern to have the `DialogTrigger` in order for the open/close logic to function correctly. In this case, the modals related to the simulator, quiz list, `CodeModal`, and a couple of other locations would need to be updated to account for this. That should be a separate PR.
> 
> I could for now ignore this and maintain the controlled logic currently present. Not sure if there will be unintended results.

Hey @TylerAPfledderer not sure if I follow you 100% on all you described but you can have a controlled Dialog without a DialogTrigger, its optional. At least in my tests, I didn't get any warnings or errors by doing that and having a fully controlled dialog.

So, for the Simulator, this is how I see it
```ts
// current layout
<grid>
    <buttons>
    <simulatormodal>
        <template ...>
    </simulatormodal>
</grid>

// new layout
<Dialog open={..}>
    <grid>
        <buttons />

        <DialogContent>
            <template ...>
        </DialogContent>
    </grid>
</Dialog>
```

And if you were talking about keeping the same API as the SimulatorModal, then yes, I think we will have to adapt in each case. And I agree with you that this "adaptation work" might be better addressed in another PR.

TylerAPfledderer:
@pettinarip noting that I do not have any objection to your response, and currently addressing the areas using the current custom `Modal` and replacing it. Currently not experiencing issues.

I would still consider revisiting this in a future PR in improving disclosure.

TylerAPfledderer:
@pettinarip I believe I am at a point for a proper review.

I'm thinking that it should be left as a separate `dialog-modal.tsx` file for now. The goal would be to unify the usage of `dialog` right? I only replaced all the current instances of the custom `Modal` and I see that `Dialog` and Chakra's `Modal` are being imported directly in other places.

And aside from that, I think there still might be a benefit to using `DialogTrigger` instead of outside controlled state, and be able to run `trackCustomEvent` via `onOpenChange` prop, along with anything else that relies on the open/close state.

pettinarip:
> @pettinarip I believe I am at a point for a proper review.
> 
> I'm thinking that it should be left as a separate `dialog-modal.tsx` file for now. The goal would be to unify the usage of `dialog` right? I only replaced all the current instances of the custom `Modal` and I see that `Dialog` and Chakra's `Modal` are being imported directly in other places.

Cool, got it.

I think we will need to leave both, the dialog and dialog-modal. `dialog` is useful for the command component (currently used in the lang picker).

The only chakra modal still pending would be the `CodeModal`, right? wondering if you are going to tackle that one in this PR.

> And aside from that, I think there still might be a benefit to using `DialogTrigger` instead of outside controlled state, and be able to run `trackCustomEvent` via `onOpenChange` prop, along with anything else that relies on the open/close state. This way, additional disclosure logic can be removed.

Sounds good, are you planning on doing that in a different PR?

pettinarip (approved):
@TylerAPfledderer gj, this looks solid. I'll see if I can get a review from designers but looks ready to me.

TylerAPfledderer:
@pettinarip 

> The only chakra modal still pending would be the CodeModal, right? wondering if you are going to tackle that one in this PR.

I believe I didn't touch it because it was not connected to the original custom modal. For how old this PR is now, I would look at that in a separate PR to not this one linger for any longer than needed.

And I would like to address the use of `DialogTrigger` in a separate PR as well.

src/components/Simulator/index.tsx
```diff
@@ -41,14 +43,17 @@ export const Simulator = ({ children, data }: SimulatorProps) => {
...
-  const onClose = (): void => {
-    trackCustomEvent({
-      eventCategory: "simulator",
-      eventAction: `${pathId}_click`,
-      eventName: `close-from-step-${step + 1}`,
-    })
-    // Clearing URL Params will reset pathId, and close modal
-    clearUrlParams()
+  const onClose: ModalProps["onOpenChange"] = (open) => {
+    console.log("🚀 ~ Simulator ~ open:", open)
```
TylerAPfledderer, line 47:
```suggestion
```

pettinarip (changes requested):
@TylerAPfledderer on desktop it looks like everything is working fine but in mobile there are a few things left.

1. You can't scroll if the content it too large. In some cases, even the close button is not visible.
2. We had a margin before wrapping the modal container and you could see a bit of the dark overlay, do you think it is possible to replicate?

This happens in the quizzes page or in the simulator
PR
image

Prod
image

Minor request: could it be possible to make the close button icon a bit bigger? as we have in prod? is 32px in prod, in this PR is 16px.

TylerAPfledderer:
Hey @pettinarip! My apologies in delay with in responding to your latest review.

I will be taking a look at these two items right now. I do not expect any issue, but will notify you immediately if I run into a problem I cannot find a quick solution for.

I will also review the design in Figma.

src/components/ui/dialog-modal.tsx
```diff
@@ -0,0 +1,225 @@
...
+  slots: {
+    content:
+      /**
+       * `size-[calc(-2rem_+_100%)]` provides 16px x- and y-spacing
+       * around the container instead of using `margin` to
+       * avoid positioning side-effects.
+       * (The container is fixed-positioned)
+       *
+       * Credit: Chakra v2 takes a similar approach for modal
+       */
```
TylerAPfledderer, line 21: Comment regarding using of `calc` with height and width instead of `margin` to have spacing around the container when the viewport gets too small.

src/components/ui/dialog-modal.tsx
```diff
@@ -0,0 +1,225 @@
...
+      "data-[state=open]:animate-contentShow size-[calc(-2rem_+_100%)] fixed left-1/2 top-1/2 grid -translate-x-1/2 -translate-y-1/2 gap-4 rounded-md bg-background p-8 shadow-[hsl(206_22%_7%_/_35%)_0px_10px_38px_-10px,_hsl(206_22%_7%_/_20%)_0px_10px_20px_-15px] focus:outline-none z-modal  overflow-auto",
```
TylerAPfledderer, line 22: Note that I add `overflow-auto` here in conjunction to setting `height` (see above comment) to allow for scroll when the container has a lot of content.

I'm torn here. When applying this quick fix, this creates space on the top and bottom around the container to prevent the container from hitting the edge of the screen. Prod does not have bottom spacing because the container is allowed to expand below the fold, and you are actually scrolling the rest of the container into view and not only the content.

The prod version allows for scrolling from below the fold into view because there is a wrapper around the container which is not present in the current usage of the radix component. I would have to create this wrapper manually and wrap `DialogPrimitive.Content`. This would also need to update where the styling is assigned to here.

I prefer the UX of prod, but wanted to show you what my quick fix looks like. If you are in agreement to maintain what is in prod, should I address it here or in a separate PR? I do not see this as a big fix, but I don't want to bloat or extend the life of this PR any more than necessary.

Update: I saw in Chromatic that my approach resulted in a side-effect of expanding the height without a max. (you'll see this when you review) This is correct behavior, but should not be employed unless there is a max-height that should be defined from the design. I think this further justifies addressing how to approach adding the content wrapper I described above.

src/components/ui/dialog-modal.tsx
```diff
@@ -0,0 +1,225 @@
...
+  children,
+  ...props
+}: React.HTMLAttributes<HTMLDivElement>) => {
+  const { header, close } = useDialogStyles()
+  return (
+    <div className={cn(header(), className)} {...props}>
+      {children}
+      <Center className={close()} asChild>
+        <DialogPrimitive.Close>
+          <MdClose size="20" />
```
TylerAPfledderer, line 124: This close button is different from prod, because the prod version comes from Chakra, which has it's own close icon, and different than `MdClose`. Therefore, I chose to use react-icons `size` prop to set a different value to match prod.

TylerAPfledderer (commented):
@pettinarip I made updates with comments provided. The comment regarding scroll needs a recommendation, as this PR is not complete at this point.

src/components/ui/dialog-modal.tsx
pettinarip: Looking at this, I think this effect will be easier if we make the Overlay wrap the Content (right now it is a sibling which make it hard to place it).

Following [this guide](https://www.radix-ui.com/primitives/docs/components/dialog#scrollable-overlay), we could do

```tsx
// this
const DialogContent = () => {
  return (
    <DialogPortal>
      <DialogOverlay>
        <DialogPrimitive.Content>
          {children}
        </DialogPrimitive.Content>
      </DialogOverlay>
    </DialogPortal>
  )
})

// instead of
<DialogPortal>
  <DialogOverlay />
  <DialogPrimitive.Content />
</DialogPortal>
```

and then,

```css
overlay => fixed inset-0 overflow-y-auto p-4
content => free, no positioning
```

Of course, this would be only for mobile screens, the previous styles should still apply to desktop resolutions.

src/components/ui/dialog-modal.tsx
pettinarip: Looks good!

src/components/ui/dialog-modal.tsx
TylerAPfledderer: Ah I missed that. No problem!

TylerAPfledderer:
@pettinarip alright, that wrapping approach is better. Looks like `Overlay` needed to use `display: grid` to position the content center and allow for the overflow to be handled correctly with scroll.

Also made a comment in Chromatic regarding padding for the quiz modal (unless I ended up addressing it before you see it)

TylerAPfledderer:
@pettinarip Applied a missing style. Need the Quiz Modal snapshot reviewed.

pettinarip (approved):
Awesome @TylerAPfledderer! gj

PR #14179, opened 2024-10-17
Data loader

pettinarip:
## Description

Creates a data loader function to handle data caching, error handling and mock data.

----

**Important**: the last two commits are for testing purposes. They need to be removed once we get an approval.

Testing params: current build is only building `en` and the home page has a revalidate time of 60 seconds.

TODO
- [ ] remove testing config before merging


src/components/BigNumber/index.tsx
pettinarip, line 1: Changes for testing purposes. This will show the full date and time and will help us to check that the data has been updated.

src/pages/index.tsx
```diff
@@ -108,30 +107,51 @@ const Codeblock = lazy(() =>
...
+  return await fetchRSS(BLOG_FEEDS)
+}
 
 type Props = BasePageProps & {
   metricResults: AllMetricData
   rssData: { rssItems: RSSItem[]; blogLinks: CommunityBlog[] }
 }
 
+// In seconds
+const REVALIDATE_TIME = 60
```
pettinarip, line 121: Revert once we finish testing.

src/lib/utils/data/dataLoader.ts
```diff
@@ -0,0 +1,54 @@
...
+  const cachedLoaders = loaders.map(([key, loader]) => {
+    const cachedLoader = cacheAsyncFn(key, loader, {
+      cacheTimeout,
+    })
+    return async () => {
+      try {
+        if (USE_MOCK_DATA) {
+          console.log("Using mock data for", key)
+          return await loadMockData(key)
+        }
```
wackerow, line 40: This leaves quite a few console logs when using the mock data... Any chance we could move this higher up so it only logs this once?

src/lib/utils/data/dataLoader.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * const [ethPrice, totalEthStaked, totalValueLocked] = await loadData();
+ */
+
+export function dataLoader<T extends unknown[]>(
+  loaders: {
+    [K in keyof T]: [string, DataLoaderFunction<T[K]>]
+  },
+  cacheTimeout?: number
+): () => Promise<T> {
```
wackerow, line 30:
```suggestion
): () => Promise<T> {
  if (USE_MOCK_DATA) console.warn("Using mock data")
```

src/lib/utils/data/dataLoader.ts
wackerow:
```suggestion
        if (USE_MOCK_DATA) return await loadMockData(key)
```

wackerow (commented):
🎉🎉 Overall seems to be working, but sometimes seems to take one or two refreshes more than I would expect, but it eventually updates. I'd be down to pull this in (after reverting the testing commits) and give it a try in staging then prod

src/lib/utils/data/dataLoader.ts
pettinarip: Ahh good suggestion! I think that is better, agree.

wackerow:
Marked as "Blocked" until reviewed, approved, and test commits reverted.

corwintines (approved):
LGTM, seems to be working well in the preview deploy

wackerow (approved):
Reverted the test commits; bringing in

PR #14181, opened 2024-10-17
fix: ethstaker blog xml feed

wackerow:
## Description
- Updates ethstaker blog xml feed url

## Related Issue
- #14171


pettinarip (approved):
Approving as it worked for me locally. We could wait until the build has finished and double check that is working on the preview deploy.

PR #14182, opened 2024-10-17
feat: fetch next year events

wackerow:
## Description
- Adds an attempt to fetch from the same Google Sheet for next years events
- Google Sheet has different pages for each year, matchable by passing the year number
- If results found, adds to `ethereumEvents` array

### DIFF
--- a/src/scripts/events-import.ts
+++ b/src/scripts/events-import.ts
@@ -12,8 +12,17 @@ import "dotenv/config"
   const communityEvents = localEvents as CommunityConference[]
 
   console.log("Community Events Import..")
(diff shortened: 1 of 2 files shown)


wackerow:
@corwintines We'll need a valid `GOOGLE_API_KEY` added to our GH secrets before this one will work... not sure what happened to the existing one, but did you handle this before?

Then we can run the workflow directly from this branch to confirm it works before we merge this.

wackerow:
Bump cc: @corwintines @pettinarip

corwintines (approved):

PR #14188, opened 2024-10-18
Update community-events.json

develo-pera:
Add ETH Belgrade 3rd edition


corwintines:
Thanks @develo-pera! This got imported from Nathan Sexers list, so going to close this out since its listed now.

PR #14197, opened 2024-10-20
Hotfix disable `revalidate`

pettinarip:
Currently, our revalidate process is always failing due to this bug, which is currently fixed in `dev` #14175. This issue causes the revalidate to never be done and always be triggered on every request, resulting in a lot of unnecessary API calls.

## Description

This PR temporarily stops revalidation until we have #14179 (which will allow us to re-enable it).


wackerow (approved):

wackerow (commented):
The GrowThePie endpoint has been disabled for now, and checking the preview these two metrics are showing the error message. Trade-off here if we want to stop the excessive API calls. Going to pull this in, but we can't let this linger.

PR #14244, opened 2024-10-25
Migrate ActionCard.tsx to tailwind

yun-chiao:
## Description
- Wrapped the `ActionCard` inside `CentralActionCard` in `eth.tsx` with a `Box` to ensure parameters like `sx` are correctly passed down.
- Migrated `className` to Tailwind CSS.
- Replaced all components and types from `@chakra-ui/react`.

## Note

I did not replace `useColorModeValue` with the `useColorModeValue` from `@/hooks/useColorModeValue` because I encountered some strange issues.

For example:

```javascript
const descriptionColor = useColorModeValue(
  "lightmode",
  "darkmode"
)
```
When using descriptionColor as a className:

```jsx
<p className={`${descriptionColor}`}></p>
```

On the first render, it always gets `"darkmode"` in className, but when I `console.log(descriptionColor)`, it consistently returns `"lightmode"` from the first print.

After doing some debugging inside `@/hooks/useColorModeValue`, I suspect it is related to the `theme` being `null` on the initial render, which causes the conditional check to fail and return `"darkmode"`. However, I am unsure why `console.log(descriptionColor)` which would be "lightmode" and the actual `descriptionColor` applied to the CSS behave inconsistently.

### localhost
[
截圖 2024-10-26 上午3 10 05
截圖 2024-10-26 上午3 10 23
](url)
### https://ethereum.org/
截圖 2024-10-26 上午3 10 58
截圖 2024-10-26 上午3 10 53

## Related Issue
#13946


src/components/ActionCard.tsx
```diff
@@ -53,64 +37,41 @@ const ActionCard = ({
   isBottom = true,
   ...props
 }: ActionCardProps) => {
-  const descriptionColor = useColorModeValue("blackAlpha.700", "whiteAlpha.800")
+  const descriptionColor = useColorModeValue(
+    "text-black opacity-65",
+    "text-white opacity-80"
+  )
```
pettinarip, line 43: We should not need `useColorModeValue` for this case. We could do
```tsx
className="text-body opacity-65 dark:opacity-80"
```

src/components/ActionCard.tsx
```diff
@@ -53,64 +37,41 @@ const ActionCard = ({
...
-      boxShadow="
-	  0px 14px 66px rgba(0, 0, 0, 0.07),
-    0px 10px 17px rgba(0, 0, 0, 0.03), 0px 4px 7px rgba(0, 0, 0, 0.05)"
-      color="text"
-      flex="1 1 372px"
-      _hover={linkBoxFocusStyles}
-      _focus={linkBoxFocusStyles}
-      className={className}
-      m={4}
+      className={`focus:bg-table-bg-hover m-4 flex flex-none shadow-table hover:scale-[1.02] hover:rounded hover:bg-background-highlight hover:shadow-table-box-hover hover:duration-100 focus:scale-[1.02] focus:rounded focus:shadow-table-box-hover focus:duration-100 ${className}`}
```
pettinarip, line 47: Lets use [the `cn` function](https://github.com/ethereum/ethereum-org-website/blob/dev/src/lib/utils/cn.ts)
```tsx
className={cn("...", className)}
```

src/components/ActionCard.tsx
```diff
@@ -53,64 +37,41 @@ const ActionCard = ({
...
     >
       <Flex
-        h="260px"
-        bg="cardGradient"
-        direction="row"
-        justify={isRight ? "flex-end" : "center"}
-        align={isBottom ? "flex-end" : "center"}
-        className="action-card-image-wrapper"
-        boxShadow="inset 0px -1px 0px rgba(0, 0, 0, 0.1)"
+        className={`action-card-image-wrapper flex h-[260px] flex-row bg-gradient-to-r from-accent-a/10 to-accent-c/10 ${isBottom ? "items-end" : "items-center"} ${isRight ? "justify-end" : "justify-center"} shadow-[inset_0px_-1px_0px_rgba(0,0,0,0.1)]`}
```
pettinarip, line 51: Use `cn` function
```suggestion
        className={cn(
          "action-card-image-wrapper flex h-[260px] flex-row bg-gradient-to-r from-accent-a/10 to-accent-c/10 shadow-[inset_0px_-1px_0px_rgba(0,0,0,0.1)]",
          isBottom ? "items-end" : "items-center",
          isRight ? "justify-end" : "justify-center"
        )}
```

src/pages/eth.tsx
```diff
@@ -237,7 +237,8 @@ const CentralColumn = (props: ChildOnlyProp) => (
 )
 
 const CentralActionCard = (props: ComponentProps<typeof ActionCard>) => (
-  <ActionCard
+  <Box
+    as={ActionCard}
```
pettinarip, line 241: 👍🏼

pettinarip (changes requested):
@yun-chiao thanks for the PR. Left a reply about `useColorModeValue`. LMK what do you think

yun-chiao:
> useColorModeValue

Hi @pettinarip ,

I've updated this PR, and here’s some information about the issue with `useColorModeValue`. Thanks!

I found an example that has been merged and is using `useColorModeValue`: `<EthVideo>`. When I'm in light mode and reload the page, the first render applies the dark mode setting instead of the light mode, as shown in the attached image. However, it behaves correctly when switching between modes.

If I add the following to `useColorModeValue`:

```javascript
if (resolvedTheme == null) {
  return "null_theme";
}
```

I then get `"null_theme"` in `src={src}`, but if I `console.log` related information in <EthVideo>, it returns the light mode variable as expected, e.g., "/images/ethereum-hero-light.mp4", with no trace of `"null_theme"` in the console. I find this behavior odd, as the displayed result in HTML and CSS differs from the variable output.

Thank you!

Snapshoot for `<EthVideo>`.
It should get the src = "/images/ethereum-hero-light.mp4" in the light mode.
截圖 2024-11-10 下午4 10 38

pettinarip:
@yun-chiao uhmm good catch! yeah...looks like a bug. Thanks for letting me know, I'll take a look later.

pettinarip (approved):

PR #14253, opened 2024-10-29
Update community-meetups.json

exyzee:
Added ETHAntwerp


corwintines:
Thanks @exyzee 

Im not seeing any information about upcoming meetups when going to this site, or where to get information on this. If this information becomes more easily accessible then we can reconsider, but for now going to close this out.

PR #14266, opened 2024-10-30
Revalidate pages tasks

pettinarip:
## Description

Creates two tasks (daily & weekly cron jobs) for trigger.dev to revalidate the defined pages.


src/crons/revalidate-daily.ts
```diff
@@ -0,0 +1,14 @@
+import { schedules } from "@trigger.dev/sdk/v3"
```
minimalsm, line 1: Why do we need a 3rd party lib for this?

wackerow: Wondering similar... @pettinarip  is there benefit to this approach over something like a GitHub workflow action? This looks similar, but managed through TypeScript files, so perhaps that is a benefit right there?

pettinarip: @minimalsm @wackerow right, I ended up using this tool because it was just faster to automate the revalidations.

Compared to a GH action, I'd say the main reasons are: error tracking (logging system), retry logic, and dx in general which makes it much easier to know what's going on and act if something goes wrong. We could easily extend the functionality in the future, e.g. send an email when a revalidation fails. We could also use this tool for other things than just revalidating pages.

But I hear you, using it for a simple cron job might be too much for now. I could try replicating this in a GH action, see how it goes and then decide between the two.

pettinarip:
Closing as this will serve as a temporary solution for revalidating pages until we migrate to the app router.

For now, we will keep this source code in the `trigger-revalidation-tasks` branch in case we need to make changes and deploy new tasks.

Issue #14310, opened 2024-11-01
Add the Node Crawler to the "Tracking nodes in the network" section


angaz:
I would like to add the [Node Crawler](https://github.com/ethereum/node-crawler) ([ethernets.io](https://www.ethernets.io)) to the list of projects crawling the execution network [here](https://ethereum.org/en/developers/docs/nodes-and-clients/#network-overview).

angaz:
Still relevant

angaz:
Still relevant

wackerow:
Commented on #14311, but will reiterate to please feel free to re-open if/when comments addressed and can be re-considered.

PR #14311, opened 2024-11-01
add node-crawler to execution node websites

angaz:
## Description

Add the Node Crawler project to the list of projects listing execution nodes.

## Related Issue

https://github.com/ethereum/ethereum-org-website/issues/14310

### DIFF
--- a/public/content/developers/docs/nodes-and-clients/index.md
+++ b/public/content/developers/docs/nodes-and-clients/index.md
@@ -45,6 +45,7 @@ What these implementations have in common is they all follow a single specificat
 
 Multiple trackers offer a real-time overview of nodes in the Ethereum network. Note that due to the nature of decentralized networks, these crawlers can only provide a limited view of the network and might report different results.
 
+- [Node Crawler](https://www.ethernets.io/) ([GitHub](https://github.com/ethereum/node-crawler)) a free and open source Ethereum network crawler
 - [Map of nodes](https://etherscan.io/nodetracker) by Etherscan
 - [Ethernodes](https://ethernodes.org/) by Bitfly
 - [Nodewatch](https://www.nodewatch.io/) by Chainsafe, crawling consensus nodes


angaz:
Still relevant

angaz:
Still relevant

wackerow (commented):
Hey @angaz, thanks for the post and apologies for the delays!

Overall this looks great! Appreciate another vantage point on the state of nodes on Ethereum

Couple concerns though:

- Site is quite slow, at least in my owb testing. Initial load takes about 3 seconds or so, and I'm getting about the same speeds to switch table views, but switching pages is taking closer to _17-20s, and often times out_:

image

- Not very mobile-friendly:

image

...though this could possibly be improved upon quite with one line inside the `head` element:

```html
<meta name="viewport" content="width=device-width, initial-scale=1">
```

- Can you clarify how up-to-date this all is?

I see "Only shows nodes which were last found via discovery in the last 24 hours Updates every 30 minutes" in the footer, but saw items that appeared noted as "last updated >30 days ago". Trying to grab a screenshot, but I honestly can't get the page to load without timing out right now 🙁.

---

In summary, I'd love to add this to the list, but biggest blocking issue would be the slow response times / timeouts. I'm afraid this would lead to a poor user experience to direct folks here as-is, though it seems like something that could be fixed up and reconsidered. 

Repo looks like the last commit was a few months ago, can you help shine some light on the future plans for this project? Thanks!

angaz:
Hi. Thanks for the review.

Regarding the timeouts: yeah, I'm currently battling with lots of scrapers trying to scrape every page of the crawl history even though it's got nothing useful for them and I've marked those links and pages as no-index, no-follow. The worst are the AI scrapers. It's currently dealing with ~10rps. It's a problem because I don't want to block all bots because I want it indexed by Google for example. So this is putting quite a strain on the database.
The stats page and the nodes page should still be functional. It comes and goes based on IOPS availability, which is pretty maxed out recently. I updated the nginx config to allow for more time for the data to load. It's not a good solution, but it's something. The stats/home page should always load. There is a lot of database caching going on for that page. 3+ seconds is not good, but unfortunately, server resources are pretty limited.

> Not very mobile friendly

Yeah I am aware. I'm not the best at CSS, if it's not clear :joy:. It's possible that I could get some AI to help me to improve it. I haven't given that a try yet.

> Only shows nodes which were last found via discovery in the last 24 hours Updates every 30 minutes.

This is valid for the stats page. The stats are showing "live" nodes, so those are ones which are still found in the discovery network. On the Nodes page, that is the history of all of the nodes since the beginning of the database, which is over 6 million nodes now.
Maybe you make a good point that the data on each page should be explained a bit more.

I am still working on the project. I have not abandoned it. I'm keeping the server up to date every 2 weeks or so. I'm just spread a bit thin the last few months. I need to do some work to upstream my changes, add things like the grafana dashboards and such.

I can ping you again when I've had some time to improve things if you would like. I could implement some ways to kill the bots, but also the scraping is pretty intensive, and I think it could be tuned down so the database is hit less with updates. This will also certainly help with the disk IOPS usage and so, load times.

wackerow:
Thanks @angaz!

Btw, is there another repo that represents the front end? I'm trying to understand the role of https://github.com/ethereum/node-crawler since the front end it serves appears completely different:

image

(dummy data)

Is this repo just being used for the crawler?

wackerow:
Hey @angaz! Circling back on this. Going to close this and the issue out for now, but please feel free to re-open the issue and ping when some of the above has been addressed. Cheers!

PR #14332, opened 2024-11-08
Grammatical and typo corrections

mdqst:
### Added typo corrections to documentation for the following files

`ethereum-org-website/blob/dev/docs/applying-storybook.md`

The word "on" should be "one." The corrected sentence is:
There may be some exceptions, but generally each component should have only one story file.

Corrected.

`ethereum-org-website/blob/dev/docs/deploy-process.md`

The typical workflow will be as follows:
It should be:
The typical workflow is as follows:

This is because the description refers to the current process, so "is" would be the correct tense instead of "will be."

Corrected.

`ethereum-org-website/blob/dev/docs/event-tracking.md`

"Each event comprises of 4 hierarchical values:"
The correct phrase should be:
"Each event comprises 4 hierarchical values:"
The word "comprise" already means "to include," so "of" is unnecessary.

Corrected.

`ethereum-org-website/blob/dev/docs/review-process.md`

There’s a small typo in the "Adding products" section:
"... ensure we not sending our users to any dubious or unsafe products."
It should be:
"... ensure we are not sending our users to any dubious or unsafe products."
The word "are" is missing.

Corrected.

### Goal

Correct grammatical errors and improve readability of the documentation.

### DIFF
--- a/docs/applying-storybook.md
+++ b/docs/applying-storybook.md
@@ -22,7 +22,7 @@ It's as easy as running `yarn storybook` to boot up a dedicated localhost to see
 
 A Storybook "story" is an instance of a component in a certain state or with certain parameters applied to show an alternative version of the component.
 
-There may be some exceptions, but generally each component should have only on story file.
+There may be some exceptions, but generally each component should have only one story file.
 
 The stories file will reside with each component. So the base folder structure in `src` will look like this:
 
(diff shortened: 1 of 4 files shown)


wackerow (approved):
Looks good, thanks @mdqst! Appreciate the clean up

wackerow:
@all-contributors please add @mdqst for bug patch

PR #14338, opened 2024-11-09
Typo Update accordion.tsx

0xwitty:
## Description

There is a minor typo in the `className` of the `ChevronNext` icon inside the `AccordionTrigger` component.

- In this line:
  ```tsx
  <ChevronNext className="size-[1em] shrink-0 text-2xl transition-transform duration-200" />
  ```
  The `size-[1em]` class is incorrect because `size-[...]` is not a valid Tailwind CSS class. It should likely be `w-[1em] h-[1em]` to set the width and height to `1em`.

**Correction:**
```tsx
<ChevronNext className="w-[1em] h-[1em] shrink-0 text-2xl transition-transform duration-200" />
```

This change correctly applies `1em` size to both the width and height of the icon.

Fixed.

### DIFF
--- a/tailwind/ui/accordion.tsx
+++ b/tailwind/ui/accordion.tsx
@@ -33,7 +33,7 @@ const AccordionTrigger = React.forwardRef<
     >
       {children}
       {!hideIcon && (
-        <ChevronNext className="size-[1em] shrink-0 text-2xl transition-transform duration-200" />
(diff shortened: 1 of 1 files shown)


wackerow:
Thanks @0xwitty!

image

Unless I'm missing something this appears to be parsing correctly... If I'm mistaken, then we have more to clean up than just that one since we use that pattern in almost 60 places. If that's the case we should open a new issue to track it, but going to close this out in the meantime. Cheers!

Issue #14248, opened 2024-10-28
In the tutorial pages "minute read" appears twice.


qbzzt:
### Describe the bug

Use Chrome to open https://ethereum.org/en/developers/tutorials/develop-and-test-dapps-with-a-multi-client-local-eth-testnet/, for example. See that under the labels you have: "⏱️11 minute read minute read"

### To reproduce

Use Chrome on Windows, for example.

### Expected behavior

Minutes to read should only appear once.

### Desktop (please complete the following information)

- OS: Windows 11
- Browser: Chrome

### Smartphone (please complete the following information)

N.A

### Additional context

N/A

### Would you like to work on this issue?

- [ ] Yes
- [X] No

Signor1:
Can I quickly fix this ?

Signor1:
@qbzzt @chuyeow @shazow

qbzzt:
It's an open source project. If you can fix it, you can and they'll merge
it (I lack the authority).

alaka03aj:
i have raised a pr for this at #14350 . 
@pettinarip kindly review it

wackerow:
Fixed with #15175

PR #14350, opened 2024-11-12
fix: removed redundant minute read text in tutorials [#14248]

alaka03aj:
## Description

Fixed redundancy in the read time display text in tutorials. Previously, the text displayed as "t minute read minute read." This change ensures that the text now displays correctly as "t minute read" without duplication.

## Related Issue

Issue solved: #14248


wackerow:
Hey @alaka03aj! Thank you for the fix here.. apologies for delays but looks like this was fixed with #15175... closing out =)

Issue #14446, opened 2024-12-02
Find Wallet bug: accordion index stays open when switching filters


wackerow:
### Describe the bug

When switching filters or personas on the find wallets page, the open accordions persist at a specific index position instead of the open/close state being attached to a specific wallet listing. 

In screen recording below, see the _seconds_ wallet position stays open, even if the filters change which wallet is in that position. 

### To reproduce

1. Go to https://ethereum.org/en/wallets/find-wallet/
2. Toggle open a position and then change filters.. find a case where the wallet changes but the  accordion stays open

### Expected behavior

The open/close state should stay with a specific wallet when filters change, not to the position in the list.

### Screenshots

https://github.com/user-attachments/assets/4c4ad039-67d5-4361-ae2b-b0b96f57871c

### Would you like to work on this issue?

- [ ] Yes
- [X] No

dbarabashh:
sorry for taking on this task without any prior notice. i decided to create a pull request really quickly, so i'm ready to continue the discussion in the pull request if anything needs to be changed

PR #14466, opened 2024-12-06
Find Wallet bug: accordion index stays open when switching filters

dbarabashh:
fix: preserve wallet expanded state when applying filters in DataTable

## Description

This PR fixes the wallet expansion state behavior in the DataTable component when filters are applied. Previously, the expanded state was not properly preserved when filters changed the visible data. 

Now:
- The expanded/collapsed state is tied to each wallet's name rather than its position in the list
- When a wallet is filtered out and then back in, its previous expanded state is preserved
- Added proper TypeScript types for the expanded state handling

## Related Issue

Bug fix for wallet expansion state behavior #14446


wackerow:
Hey @dbarabashh, apologies for delays on this. Been a lot going on.. this page was recently refactored to improve it's performance... @pettinarip has put up a simplified fix for this in #16285, so going to close this out in lieu of that one. Thanks again for diving in!

PR #14478, opened 2024-12-10
Switch to weekly active addresses and smaller growthepie API endpoint

mseidlx:
Switch to weekly active addresses and smaller growthepie API endpoint

## Description

- switching to "fundamentals" endpoint instead of "fundamentals_full" - it's a smaller endpoint but still has all required data.
- using "aa_last7d" (Active Addresses last 7 days) instead of "daa" (Daily Active Addresses) since you are referring to this datapoint as weekly active addresses on your page
- small naming fix "growthepie" is how we ideally refer to our platform :)


corwintines (approved):
🔥 thanks @mseidlx

corwintines:
@all-contributors can you add @mseidlx for code

PR #14486, opened 2024-12-11
Fix formatting and bug in pseudocode in `patricia-merkle-trie`

Privat33r-dev:
Prior to the change the doc wasn't formatted for Python code, even though mostly Python syntax is used, which might've lead to decrease in comprehension as the syntax wasn't highlighted.

All the code was formatted with the Black formatter to ease comprehension.

Additionally, the following line contained bug:
```python
        if node = '': return ''
```

The comparison operator (==) was likely meant.


Privat33r-dev:
@Aldekein I guess that it is a false positive. It's not even a part of the PR.

wackerow (approved):
Thanks @Privat33r-dev! Sorry for delay, pulling in

Privat33r-dev:
Thanks for the merge! I am just glad that it was merged while still being relevant.

PR #14495, opened 2024-12-12
feat: Migrate SideNavMobile to Shadcn/Tailwind

ashiskumar-1999:
## Description
Migrated the ```Chakra-UI``` imports to the ```Shadcn/Tailwind```
#13946


src/components/SideNavMobile.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useState } from "react"
 import { AnimatePresence, motion } from "framer-motion"
 import { useTranslation } from "next-i18next"
 import { MdChevronRight } from "react-icons/md"
-import { Box, Center, HStack, Icon } from "@chakra-ui/react"
+import { Box } from "@chakra-ui/react"
```
pettinarip, line 5: All the chakra imports need to be migrated

pettinarip (changes requested):

src/components/SideNavMobile.tsx
ashiskumar-1999: There isn't any particular Box component in the /ui/flex.tsx file, so I was confused on migrating this component.

pettinarip: Ah! got it. A `Box` in chakra is going to render a simple `div` element.

```tsx
// So, if you have
<Box mt={4}>

// should be migrated to
<div className="mt-4">
```

in cases where we are using `motion`

```tsx
<Box as={motion.div}>

// should be migrated to
<motion.div>
```

ashiskumar-1999: I understand, I'll make sure to migrate those imports accordingly.

Additionally, I think it would be beneficial to update the instructions to include this approach. This way, anyone needing to migrate the ```Box``` component in the future can follow the same process we discussed.

pettinarip (commented):
Cool @ashiskumar-1999. There are a few more things to polish to finish this. I'll push some color adjustments.

pettinarip (approved):

PR #14516, opened 2024-12-17
doc: Replace `money` with `ether`

BrodaNoel:
## Description
I think that the concept of "money" is quite different with what Ethereum wants to communicate in this documentation.

When there is a transaction, we are not getting "money" from the sender account, but its ether.

## Related Issue

Well... I just realized (while creating this PR) there should be a discussion first... I just created this: https://github.com/ethereum/ethereum-org-website/discussions/14515


corwintines:
Thanks @BrodaNoel 

While I agree this would probably be an improvement, we do not edit the whitepaper, so this is how it was written. Going to close this out.

BrodaNoel:
Screen Shot 2024-12-17 at 14 16 16

Fair enough...

PR #14519, opened 2024-12-17
[DO NOT MERGE] - Test adding informal address to German NFT page 

minimalsm:


wackerow:
cc: @minimalsm @lukassim not sure if we wanted to do anything with this? Going to close this out, but please reopen if we still plan to use.

PR #14521, opened 2024-12-17
fix(ButtonTwoLines): split rendering of button or link

TylerAPfledderer:
Splits `ButtonTwoLines` component to a dedicated `ButtonTwoLines` and `ButtonLinkTwoLines`.

## Current Behavior

The current `ButtonTwoLines` relies on type guarding to check if a button or button link should render and accept the appropriate props.

The unintended side-effect with the current approach is that there is child content with specific props used that should not be sent to the Button/ButtonLink component, adding complexity in addressing unexpected issues.

## New Behavior

Split the component into dedicated renders of either the `Button` or `ButtonLink` and add a creator function (`createSplitProps`) to help split the props so that only the needed ones for the child content are sent down, and the rest go to the parent component. 

Most props are not explicit and are meant for the parent component only, some are meant for both, so need an intermediary to correctly identify who goes to the child content and not to the parent.

## Additional Information

- Removed the old `ButtonTwoLines` file.
- Updated the required `href` prop for the `ButtonLink`.
- Added a `ButtonLinkTwoLines` story. Although it has the same styling, this will confirm the use of `ButtonLink` without error or regression.

I will note that this seems like overengineering at the moment, as we currently do not employ the component in prod, and therefore do not even know if a rendering with `ButtonLink` is necessary. Open to discussion of simplification, reverting to the old approach and modifying that, or something else.


src/components/ui/buttons/Button.tsx
```diff
@@ -126,8 +126,9 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
 )
 Button.displayName = "Button"
 
-type ButtonLinkProps = LinkProps &
+type ButtonLinkProps = Omit<LinkProps, "href"> &
   Pick<ButtonProps, "size" | "variant" | "isSecondary"> & {
+    href: string
```
pettinarip, line 131: just curious, is the intention of this change to make the field required? looks good if so

pettinarip (approved):
LGTM @TylerAPfledderer. I'm fine with this refactor, its cleaner and easier to understand than the previous one IMO.

src/components/ui/buttons/Button.tsx
TylerAPfledderer: @pettinarip yes. I find it weird that it's not a required thing for the anchor element. 😅

Issue #14527, opened 2024-12-18
Error in solidity code 


YakshitAgarwal:
Original solidity code: 

pragma solidity ^0.7.0;

contract MutexPattern {
    bool locked = false;
    mapping(address => uint256) public balances;

    modifier noReentrancy() {
        require(!locked, "Blocked from reentrancy.");
        locked = true;
        _;
        locked = false;
    }
    // This function is protected by a mutex, so reentrant calls from within `msg.sender.call` cannot call `withdraw` again.
    //  The `return` statement evaluates to `true` but still evaluates the `locked = false` statement in the modifier
    function withdraw(uint _amount) public payable noReentrancy returns(bool) {
        require(balances[msg.sender] >= _amount, "No balance to withdraw.");

        balances[msg.sender] -= _amount;
        bool (success, ) = msg.sender.call{value: _amount}("");
        require(success);

        return true;
    }
}

Error in the above code:
bool (success, ) = msg.sender.call{value: _amount}("");

Changed to:
(bool success, ) = msg.sender.call{value: _amount}("");

YakshitAgarwal:
Can someone assign me this issue please?

konopkja:
this is ethereum.org repo how is this related to our site?

YakshitAgarwal:
the code there was incorrect so I correct it.

PR #14528, opened 2024-12-18
error in solidity code

YakshitAgarwal:
Original solidity code:

pragma solidity ^0.7.0;

contract MutexPattern {
    bool locked = false;
    mapping(address => uint256) public balances;

    modifier noReentrancy() {
        require(!locked, "Blocked from reentrancy.");
        locked = true;
        _;
        locked = false;
    }
    // This function is protected by a mutex, so reentrant calls from within `msg.sender.call` cannot call `withdraw` again.
    //  The `return` statement evaluates to `true` but still evaluates the `locked = false` statement in the modifier
    function withdraw(uint _amount) public payable noReentrancy returns(bool) {
        require(balances[msg.sender] >= _amount, "No balance to withdraw.");

        balances[msg.sender] -= _amount;
        bool (success, ) = msg.sender.call{value: _amount}("");
        require(success);

        return true;
    }
}

Error in the above code:
bool (success, ) = msg.sender.call{value: _amount}("");

Changed to:
(bool success, ) = msg.sender.call{value: _amount}("");

Fixes: https://github.com/ethereum/ethereum-org-website/issues/14527

### DIFF
--- a/public/content/developers/docs/smart-contracts/security/index.md
+++ b/public/content/developers/docs/smart-contracts/security/index.md
@@ -346,7 +346,7 @@ contract MutexPattern {
         require(balances[msg.sender] >= _amount, "No balance to withdraw.");
 
         balances[msg.sender] -= _amount;
-        bool (success, ) = msg.sender.call{value: _amount}("");
+        (bool success, ) = msg.sender.call{value: _amount}("");
         require(success);
 
         return true;


konopkja:
this PR is related to https://github.com/ethereum/ethereum-org-website/issues/14527

corwintines (approved):
Thanks @YakshitAgarwal

PR #14616, opened 2025-01-07
txCosts using layer2Data

corwintines:
## Description

- Had an issue where the average txCosts was using all networks from growthepie, but we only show some of the networks. Filtering this down now so that the average costs reflects the networks we show on ethereum.org


Ahmednurmuhiyadin (approved):
INVESTOR AHMEDNUR

pettinarip (approved):

PR #14638, opened 2025-01-10
fix: typos

omahs:
fix: typos


corwintines:
Thanks @omahs, but these should be changed in Crowdin through our translation program. More info here: https://ethereum.org/en/contributing/translation-program/

PR #14679, opened 2025-01-14
fix: OOG cases

colinlyguo:
## Description

When learning this session, I found that if one transaction has a gas limit of 20000 < 21000, it will fail in the transaction validation phase, and not even enter the EVM execution phase. thus this PR fixes this and adds another case without removing the original meaning.


corwintines (approved):
Cool, thanks @colinlyguo!

corwintines:
@all-contributors can you add @colinlyguo for content

PR #14683, opened 2025-01-15
feat: migrate Card to tailwind

corwintines:
## Description

- Migrate `src/components/Card/index.tsx` from Chakra to tailwind
- Fixes uses of this component to also be tailwind

### DIFF
--- a/src/components/Card/index.tsx
+++ b/src/components/Card/index.tsx
@@ -1,40 +1,44 @@
 import { ReactNode } from "react"
-import { Heading, Stack, StackProps, Text } from "@chakra-ui/react"
 
 import Emoji from "@/components/Emoji"
 
-export type CardProps = Omit<StackProps, "children" | "title"> & {
+import { cn } from "@/lib/utils/cn"
+
+export type CardProps = {
   children?: ReactNode
   emoji?: string
   title?: ReactNode
   description?: ReactNode
+  className?: string
 }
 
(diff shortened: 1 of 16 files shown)


src/components/Card/index.tsx
```diff
@@ -1,40 +1,44 @@
...
+  title,
+  description,
+  children,
+  className,
+  ...props
+}: CardProps) => (
+  <div
+    className={cn(
+      "flex flex-col justify-between space-y-4",
+      "rounded-sm bg-ednBackground",
```
pettinarip, line 26:
```suggestion
      "rounded-sm bg-background-highlight",
```

src/components/Card/index.tsx
```diff
@@ -1,40 +1,44 @@
...
+      "border border-solid border-lightBorder",
```
pettinarip, line 27:
```suggestion
      "border border-solid border-low-contrast",
```

src/styles/semantic-tokens.css
pettinarip, line 1: The changes mentioned before would avoid adding these colors as we don't want to support them anymore.

tailwind.config.ts
pettinarip, line 1: Same here

pettinarip (changes requested):
@corwintines left a couple of comments related to colors before we move forward with it.

pettinarip (approved):

Afeter8 (approved):

PR #14684, opened 2025-01-15
feat: migrate CardList to tailwind

corwintines:
## Description

- Migrate `/src/components/CardList.tsx` to tailwind
- Update usage of this component


pettinarip (changes requested):
Noted that we were using LinkOverlay before that made the whole container area clickable and not just the link. We could use https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/ui/link-box.tsx to replicate the same.

And I see this diff in styles vs prod
image

prod:
image

pettinarip (approved):

Afeter8:
Page

Afeter8 (approved):

PR #14688, opened 2025-01-15
docs: Fix incorrect Devcon year Update index.md

0xminds:
## Description

I noticed a small typo in the 2024 section under the Execution upgrade naming table.
For the Cancun upgrade, the Devcon year was listed as 2018, but it should be **2019**. 

Corrected.

P.S. Thanks for Ethereum!


corwintines (changes requested):
Thanks for bringing this up, but I think most of these days are actually wrong based on the archive https://devcon.org/en/past-events/

I can update these dates.

corwintines (approved):

PR #14724, opened 2025-01-22
Update index.md Peera community forums has a new domain and name

byte:
## Description

Small change since Peerhana is now Peera community forums


corwintines (approved):
Thanks @byte

PR #14739, opened 2025-01-25
Migrate community page to Tailwind/Shadcn

Baystef:
## Description
Migrate community page to Tailwind/Shadcn

## Related Issue
#13946


src/pages/community.tsx
```diff
@@ -232,20 +192,9 @@ const CommunityPage = () => {
       />
       <HubHero {...heroContent} />
       <Divider />
-      <Flex
-        bg="homeBoxTurquoise"
```
Baystef, line 236: Hi @pettinarip , is there a figma page where these pages were redesigned? So I can easily go there and know what we're migrating custom background colours like this to. Thanks

pettinarip: Hey unfortunately we haven't done that. What we are doing with these old colors (not supported in the new theme) is to hardcode their values if there is no an equivalent or very close to be equivalent in the new theme.

On these unique cases, we would do something like
```ts
"bg-[#ccfcff] dark:bg-[#293233]"
```

Then, for most of the grays or highlight/hover colors, we are using `bg-background-highlight`.

Baystef: Alright, Thanks. I'll update it

> Hey unfortunately we haven't done that. What we are doing with these old colors (not supported in the new theme) is to hardcode their values if there is no an equivalent or very close to be equivalent in the new theme.
> 
> On these unique cases, we would do something like
> 
> ```ts
> "bg-[#ccfcff] dark:bg-[#293233]"
> ```
> 
> Then, for most of the grays or highlight/hover colors, we are using `bg-background-highlight`.

pettinarip (approved):

PR #14749, opened 2025-01-26
[ShadCN]: Migrate Avatar to ShadCN

TylerAPfledderer:
## Description

Migrates `Avatar` to ShadCN in the design system.

### **Important**: 
`IssuesList`, `FileContributors`, and `Leaderboard` components all use `Avatar` directly from Chakra, not from the custom component. This usage is missing the use of the avatar as a link (no `href` prop used). 

In addition, the general usage of the avatar in prod does not match intent from the design system. 
- In `IssuesList`, the avatar is accompanied by static text of the contributor's name, and not served as a link to the contributor's profile
- In `Leaderboard`, the design of the list item is too complex to be able to use the avatar from the design system.
- In `FileContributors` it is also a little too complex for the avatar from the design system, as it adds static text around the contributor link

Would need some direction on how to further proceed with this PR to address the above use cases, or resolve them in a separate PR.

## Related Issue
#13946


src/components/ui/avatar.tsx
```diff
@@ -0,0 +1,249 @@
...
+const AvatarBase = React.forwardRef<
+  React.ElementRef<typeof AvatarPrimitive.Root>,
+  AvatarBaseProps
+>(({ className, size, ...props }, ref) => (
+  <AvatarStylesContext.Provider value={avatarStyles({ size })}>
+    <AvatarPrimitive.Root
+      ref={ref}
+      style={
+        {
+          "--avatar-base-shadow-color": "hsl(var(--primary-low-contrast))",
```
pettinarip, line 66: Can't we just use `shadow-primary-low-contrast`? or define it in the theme `shadow-avatar`?

pettinarip (commented):
Amazing. LGTM @TylerAPfledderer

src/components/ui/avatar.tsx
TylerAPfledderer: @pettinarip Doing `shadow-primary-low-contrast` will override the default coordinates (essentially rendering no coordinates). And have to explicitly define different coordinates plus the color.

You could certainly create a shadow theme, but you would have to make two here, because you have to still explicitly state the different sizes. (It got tricky keeping the old way with `em` without creating other problems in the short term).

So
```ts
module.exports = {
  theme: {
    extend: {
      boxShadow: {
        // This is based on what I currently see in the Figma file
        'avatar-lg': '4px 4px 0 var(--primary-low-contrast)',
        'avatar-sm': '2px 2px 0 var(--primary-low-contrast)'
      }
    }
  }
}
```

Not a problem for me if you prefer adding these classes. I'm not sure of another way at the moment.

pettinarip: Ok, that makes sense. Will pull this in 🚀

pettinarip (approved):

PR #14769, opened 2025-01-29
Fixing typo

YakshitAgarwal:
Except a dispute arises, the intermediary only has to be contacted when opening or terminating the channel.
This above statement is grammatically incorrect

The correct statement should be:
Unless a dispute arises, the intermediary only has to be contacted when opening or terminating the channel.


corwintines (approved):
Thanks @YakshitAgarwal

Issue #14777, opened 2025-01-30
Bug report: update link on stablecoin page


konopkja:
### Describe the bug

on https://ethereum.org/en/stablecoins/ there is an old link to "Gitcoin bounties" which has been since then rebranded to Buidlbox.

pls replace link and icon with updated versions

New url: https://app.buidlbox.io/

### To reproduce

1. Go to https://ethereum.org/en/stablecoins/
2. Scroll down to "How to get stablecoins"
3. Click on "Earn

![Image](https://github.com/user-attachments/assets/3acd716e-8be2-470d-8a0a-d4536c86865e)

### Expected behavior

replace invalid url with new one: https://app.buidlbox.io/

replace old gitcoin icon with new branding

### Would you like to work on this issue?

- [ ] Yes
- [x] No

rvmzx:
hi
assign it to me, please

rvmz:
hey @konopkja 
I'll fix it

konopkja:
thank you, assigning you the issue

yjoon45:
Hey @konopkja
If this is not fixed yet, can i pick this up?

konopkja:
> Hey [@konopkja](https://github.com/konopkja) If this is not fixed yet, can i pick this up?

Hi, lets give a couple of days more to @rvmz to make a PR first, if not I will reassign it to you the next week.

wackerow:
Fixed with #15515

PR #14805, opened 2025-02-01
fix: update link on stablecoin page (#14777)

rvmz:
updated link from gitcoin.co/explorer to app.buidlbox.io
updated gitcoin image to its' latest version (https://www.gitcoin.co/brand)


corwintines (changes requested):
Thanks @rvmz 

I think we should update the name here to buidlbox as well, instead of Gitcoin, and use the buidlbox logo instead. After those changes should be good to go 🚀

rvmz:
hey @corwintines 
are we good to merge it?

wackerow:
@all-contributors please add @rvmz for maintenance

PR #14814, opened 2025-02-03
[ShadCN]: Implement ShadCN Avatar to Production

TylerAPfledderer:
## Description

Applies the newest Avatar component to all current instances: `IssuesList.tsx`, `FileContributors.tsx`, and `Leaderboard.tsx`

Two points of note related to the Design System:
- It is intentional in `Leaderboard.tsx` to adhere to the DS, where it is required to supply a value to the `href` prop. It is believed that it is better to address this requirement in the short term with a comment and using `#` as the value in the component instance, instead of making the prop an optional type. This should not be noticeable to the user.
- There are multiple instance where the height and width are set to the token value `10`. This value is not being used in any of the `size` variants in the DS, and therefore explicitly set in the given instances.

## Related Issue
Minor impact approaching migrations done for the above affected components in #13946

### DIFF
--- a/src/components/FileContributors.tsx
+++ b/src/components/FileContributors.tsx
@@ -1,5 +1,4 @@
 import { BaseHTMLAttributes, useState } from "react"
-import { Avatar } from "@chakra-ui/react"
 
 import type { ChildOnlyProp, FileContributor } from "@/lib/types"
 
@@ -12,6 +11,7 @@ import { ScrollArea } from "@/components/ui/scroll-area"
 
 import { trackCustomEvent } from "@/lib/utils/matomo"
 
+import { Avatar } from "./ui/avatar"
 import Modal from "./ui/dialog-modal"
 
 import { useBreakpointValue } from "@/hooks/useBreakpointValue"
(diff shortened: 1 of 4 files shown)


src/components/FileContributors.tsx
pettinarip, line 1: I see the image cutting a bit in this case
image

src/components/Leaderboard.tsx
```diff
@@ -95,10 +96,10 @@ const Leaderboard = ({ content, limit = 100 }: LeaderboardProps) => {
                 <Avatar
                   src={avatarImg}
                   name={avatarAlt}
-                  me={4}
-                  h={10}
-                  w={10}
-                  display={{ base: "none", xs: "block" }}
+                  // This meets the Design System requirement, despite not being used in this instance
+                  href="#"
```
pettinarip, line 100: Could we align it with the link for the row?
```suggestion
                  href={hasGitHub ? `${GITHUB_URL}${username}` : "#"}
```

pettinarip (changes requested):

src/components/Leaderboard.tsx
pettinarip: If not the user gets this ux where the avatar link is upside the linkbox making it a weird experience

https://github.com/user-attachments/assets/129fc872-e7d7-4484-96d5-576e640d700b

TylerAPfledderer: Hmmm, I wasn't expecting that! I'll go ahead with your suggestion here. 

For future consideration the avatar shouldn't really be hoverable here unless styled based on hover over the rest of the list item, correct?

src/components/FileContributors.tsx
TylerAPfledderer: The Avatar Root is set as `display: flex` with `align-items` and `justify-content` being set to `center`. The latter is affecting the image cut. (The image is actually slid to the left, and not cropped).

I moved the flexbox styles to the fallback component, because the text in this component needs the centering. This is not needed for the actual rendered image. (Verified in Storybook)

TylerAPfledderer:
@pettinarip should be all set with the changes here!

src/components/FileContributors.tsx
pettinarip: Nice, this looks great now!

src/components/Leaderboard.tsx
pettinarip: Yes, correct. For this transition, we will leave it as is, but ideally only the row should be the element that gets the hover effect.

pettinarip (approved):
LGTM! thanks @TylerAPfledderer

PR #14820, opened 2025-02-04
[WIP] Listen to feature

corwintines:


Afeter8 (approved):

Afeter8 (approved):

konopkja:
Event Category
Audio

Event Action
click 

Event Name
start, pause, play, speed, next, previous, autoplay, minimize, expand


Event Category
Audio

Event Action
automated_next 

Event Name
1,2,3,4,5

konopkja:
on mobile the widget should be centered

konopkja:
mouse over the icons needs descriptions: play, pause, next, previous, minimize, autoplay

konopkja:
the text should be clickable, not just the icon
image

src/components/ListenToPlayer/PlayerWidget/index.tsx
pettinarip, line 1: Would be cool if we could make all these buttons as actual buttons to inherit keyboard navigation + a11y goodies to explain what each button does.

We could add a `tabIndex={0}` to make it focusable (only when it is shown) and an `aria-label` to explain what this is to screen readers.

pettinarip: We could also render this inside a Portal to make it visible early in the focus order. https://www.radix-ui.com/primitives/docs/utilities/portal

src/components/ListenToPlayer/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+    if (!sound) return
+    sound.seek(duration)
+    setTimeRemaining(0)
+  }
+
+  const handlePlaybackSpeed = (speed: number) => {
+    setPlaybackSpeed(speed)
+  }
+
+  return playlist.length > 0 && index !== -1 ? (
```
pettinarip, line 145:
```suggestion
if (!playlist.length) {
    return null
}

return (
    // happy path
)
```

pettinarip (commented):
Looks good @corwintines 

Are there any plans to keep the feedback widget button visible? maybe placing it above it or on the left side?
image

minimalsm (approved):
Nice work :)!

konopkja:
Screenshot 2025-02-26 at 13 55 57

here the audio goes the full width which it shouldnt

konopkja:
> mouse over the icons needs descriptions: play, pause, next, previous, minimize, autoplay

this still needs to be added i think cc @corwintines

konopkja:
Screenshot 2025-02-26 at 14 00 37

this might be too short, could we keep the original width for the background @corwintines  ?

konopkja:
IMG_594201E6D6AF-1

align to center the component

src/data/audio/what-is-ethereum/script.md
pettinarip, line 1: Noting that we don't have the "You are listening to ..." intro text in this case.

src/components/Tooltip/index.tsx
```diff
@@ -105,7 +105,7 @@ const Tooltip = ({
         <Content
           side="top"
           sideOffset={2}
-          className="max-w-80 px-5 text-sm"
+          className="z-[10000] max-w-80 px-5 text-sm"
```
pettinarip, line 108: Could we use one of the tokens?, in this case `z-popover`. Mainly to avoid z index conflicts with other components.

pettinarip (approved):
LGTM!

PR #14853, opened 2025-02-09
[ShadCN]: Migrate MergeInfographic

TylerAPfledderer:


src/components/MergeInfographic/index.tsx
```diff
@@ -1,105 +1,83 @@
...
-    textAnchor: "start",
-    fill: "currentColor",
-  },
-})
+const Text = ({
+  className,
+  ...rest
+}: SVGTextElementAttributes<SVGTextElement>) => (
+  <text
+    className={cn("fill-current [text-anchor:start]", className)}
```
pettinarip, line 18: didn't know that you could do that (arbitrary props) TIL

pettinarip (approved):
Nice job

src/components/MergeInfographic/index.tsx
TylerAPfledderer: @pettinarip ah yes! Applies to any case where you need to explicitly write a CSS property and value. Of course, you lose the ability to apply a token.

This includes creating CSS variables as well.

PR #14872, opened 2025-02-11
refactor(components): replace and remove old components

TylerAPfledderer:
## Description

Replaces and removes old components that were tied to Chakra or were no longer needed.


src/components/ui/buttons/SvgButtonLink.tsx
pettinarip, line 1: 👍🏼

pettinarip (approved):
Great!!! @TylerAPfledderer thanks

PR #14877, opened 2025-02-12
Install OpenZeppelin 4.* in NFT tutorial

gumb0:
`Counter.sol` is imported in the contract, but was removed from OpenZeppelin 5.0.0, so then compilation fails. Installing older version helps.


wackerow:
@all-contributors please add @gumb0 for maintenance

PR #14878, opened 2025-02-12
Update German translations common.json

minimalsm:
The German translations uses the passive voice, which changes the emphasis from the action of adding to a state of being added, potentially leading to confusion.


wackerow:
Can't confirm these personally... should we make these updates in Crowdin and have them reviewed, then pull them in?

wackerow:
@minimalsm Updating strings outside the Crowdin flow will lead to repeated work at the monthly imports—are these updated in Crowdin?

wackerow:
@lukassim When you get a chance, mind checking on this? If we want this changes just need to make sure they're in Crowdin, else can close this out. Thanks!

minimalsm:
@claude can you review

PR #14879, opened 2025-02-12
Update index.md

perriefidelis:
Added punctuations

I only added some punctuations 

## Description

Some of the listed items weren't puntuated


wackerow:
Appreciate this but given the choice, leaving the full stops off if preferred.. thanks!

PR #14905, opened 2025-02-14
refactor(components): remove remaining Chakra imports

TylerAPfledderer:
## Description

Removes any leftover Chakra imports before taking out the theming config and provider. (Keeping these PR short to cover any unexpected issues)

This does not address the `useColorMode` hook in `useNav`, as that will be removed when removing the Chakra Theme Provider.

Also does not remove anything from the Chakra theme config or storybook, as this will also be addressed when removing the theme provider.


pettinarip (approved):
LGTM!

PR #14906, opened 2025-02-14
i18n `next-intl`

pettinarip:
## Description

This PR implements `next-intl` replacing the previous i18n library `next-i18next`. This is the initial step towards the use of the app router.

### Changes for this transition
- Removed the built-in `i18n` mechanism from nextjs. This is all handled by `next-intl` and by the page components using `getStaticPaths`. The load of namespaces logic is now in our control, created a `loadNamespaces` function.
- All pages are located in the folder `[locale]`
- All translation related hooks were migrated to the exported versions from `next-intl`. These are `useTranslation`, `useRouter`, `usePathname`, `useLocale`
- Current limitation: 404 page can't be translated in this transition.
- Changed the env var `BUILD_LOCALES` to `NEXT_PUBLIC_BUILD_LOCALES` since we need to use it to get all locales used in client components.
- Updated the `Link` component to use the i18n navigation components

### Things to test
- Netlify support (bundles sizes, redirects, compression, default locale redirect)
- Static files urls (pdfs, images in md files)
- Edit paths (docs & md pages)
- RTL support
- Missing translations defaults to English or the searched key
- Storybook support

## Preview link
https://deploy-preview-14906--ethereumorg.netlify.app/


next.config.js
```diff
@@ -59,7 +61,6 @@ module.exports = (phase, { defaultConfig }) => {
 
       return config
     },
-    i18n,
```
pettinarip, line 62: Important change. This is not supported in the app router, removing now and building the pages using `getStaticPaths`

src/components/AssetDownload/index.tsx
```diff
@@ -55,22 +56,12 @@ const AssetDownload = ({
         )}
       </div>
       <Flex className="mt-4 gap-5">
-        <ButtonLink
-          href={imgSrc}
-          onClick={matomoHandler}
-          target="_blank"
-          locale={false}
```
pettinarip, line 62: Modified the `Link` component to handle this case.

src/components/ProductTable/index.tsx
pettinarip, line 1: Im leaving this for this transition. Should be fixed once we moved to the app router.

For context, the main purpose of this was to remove the query params from the url once the page loads.

src/components/Translation.tsx
pettinarip, line 1: Important change. Adaptation of the `Translation` component. It changes a bit the API.

src/components/ui/Link.tsx
pettinarip, line 1: Important change.

src/components/ui/Link.tsx
```diff
@@ -49,25 +50,25 @@ export const BaseLink = forwardRef<HTMLAnchorElement, LinkProps>(function Link(
...
+  const isInternalFile = isFile && !isExternal
   const isHash = url.isHash(href)
 
-  // Get proper download link for internally hosted PDF's & static files (ex: whitepaper)
+  // Get proper download link for internally hosted files (ex: whitepaper.pdf)
   // Opens in separate window.
-  if (isInternalPdf) {
-    href = getRelativePath(asPath, href)
+  if (isInternalFile && !href.startsWith("/")) {
+    href = "/" + getRelativePath(pathname, href)
```
pettinarip, line 71: Convert relative paths into absolute paths

src/hooks/useTranslation.ts
pettinarip, line 1: Created this abstraction to add backwards compatibility with the args that we could handle, making the transition easier.

src/i18n/loadNamespaces.ts
pettinarip, line 1: Loading the namespace is not handled by the i18n library.

This logic tries to mimic what we were doing before:
- tries to load the namespace for the locale + the default locale, and merges both to default to english in missing translations
- if the namespace doesn't exist, it loads the default one

src/i18n/request.ts
pettinarip, line 1: Can be ignored. Not important at this point since this is used for app router pages.

src/pages/[locale]/assets.tsx
```diff
@@ -96,7 +98,16 @@ const H3 = (props: ChildOnlyProp) => (
...
-export const getStaticProps = (async ({ locale }) => {
+export async function getStaticPaths() {
+  return {
+    paths: LOCALES_CODES.map((locale) => ({ params: { locale } })),
+    fallback: false,
+  }
+}
+
+export const getStaticProps = (async ({ params }) => {
+  const { locale = DEFAULT_LOCALE } = params || {}
```
Comment (pettinarip) [inline src/pages/[locale]/assets.tsx:109]: The locale is now passed as a param as its not part of the context anymore.

src/pages/_app.tsx
```diff
@@ -26,18 +31,31 @@ const App = ({ Component, pageProps }: AppPropsWithLayout) => {
...
-          {getLayout(<Component {...pageProps} />)}
-        </BaseLayout>
-      </TooltipProvider>
-    </ThemeProvider>
+    <NextIntlClientProvider
+      locale={(router.query.locale as string) || DEFAULT_LOCALE}
+      messages={pageProps.messages || {}}
+      onError={() => {
+        // Suppress errors by default, enable if needed to debug
+        // console.error(error)
```
pettinarip, line 39: Once we fix all our missing translations, we could enable this and let the build failed. For now, we are suppressing them as we are doing currently.

src/pages/_app.tsx
```diff
@@ -26,18 +31,31 @@ const App = ({ Component, pageProps }: AppPropsWithLayout) => {
...
+      }}
+      getMessageFallback={({ key }) => {
+        const keyOnly = key.split(".").pop()
+        return keyOnly || key
```
pettinarip, line 43: If a missing translation, output just the key without the namespace.

Issue #14910, opened 2025-02-14
Change logic of higlighted networks on layer 2 page


konopkja:
on Page: https://ethereum.org/en/layer-2/

We randomly highlight networks for the reader in this section:

<img width="709" alt="Image" src="https://github.com/user-attachments/assets/5761acb6-714b-4f78-8a8a-be51d5198262" />

**Change request**
Change the logic to only showcase the networks that have reached "Maturing" stage (https://ethereum.org/en/layer-2/networks/).

Networks that are less developed and experimental should not be highlighted anymore.

jrudolf:
Not sure if trivial or not, but would be nice if the page also updated on every refresh (so the 3 shown get rotated faster)

![Image](https://github.com/user-attachments/assets/16e36403-0707-4b4f-bdfc-5bf88c1fb143)

SamarthJ2004:
Hi @konopkja , I would like to work on this issue.

konopkja:
@SamarthJ2004  thank you, assigning you the issue

SamarthJ2004:
Also, as of now there are just 2 matured networks but in the future they will increase. So, are all of them to be shown or max 3?

konopkja:
> Also, as of now there are just 2 matured networks but in the future they will increase. So, are all of them are to be shown or max 3?

good consideration! this list should support max 3

SamarthJ2004:
Hi @konopkja , I have open a pull request for the same. Please suggest if any changes are to be made.

corwintines:
Merged the above PR, so going to close this out.

PR #14917, opened 2025-02-15
issue:14910 added functionality to highlight robust and maturing networks

SamarthJ2004:
Added functionality to highlight robust and maturing networks on Page: https://ethereum.org/en/layer-2/

## Description
Added a function to to filter the L2LayerData, if it is "Robust". If the length is 0, then it checks for network that have reached "Maturing" and return max 3 of them at random.

## Related Issue
#14910 

image


konopkja:
@minimalsm  cc, check if logic makes sense to you

wackerow (approved):

PR #14922, opened 2025-02-17
Add Pectra info to bug bounty page

fredriksvantes:
## Description

Add Pectra info to bug bounty page


wackerow (approved):
Looks good! Thanks @fredriksvantes

Seeing a small bug on this component (needs to be full width) which I'll patch separately

PR #14931, opened 2025-02-18
[ShadCN]: remove Chakra Provider and Dependencies

TylerAPfledderer:
## Description

Removes the Chakra Provider, Theming config, and package dependencies.

`useLocaleDirection` now becomes just a trigger with no value returned, as the RTL direction only needs to be injected into the DOM on update.

This should be the final PR for the UI migration. What is left as a non-blocker is to update documentation about using Tailwind and Radix (ShadCN) in the stack, and any particular notes for DX in how they are being applied to this project (if there are any particulars)


wackerow:
Let's gooo!! 🔥

src/components/Nav/useNav.ts
```diff
@@ -476,7 +473,7 @@ export const useNav = () => {
     const targetTheme = resolvedTheme === "dark" ? "light" : "dark"
 
     setTheme(targetTheme)
-    setColorMode(targetTheme)
+    // setColorMode(targetTheme)
```
pettinarip, line 476:
```suggestion
```

tailwind.config.ts
```diff
@@ -5,12 +5,8 @@ import { screens } from "./src/lib/utils/screen"
 
 const config = {
   // TODO: Move to "class" strategy after removing Chakra
```
pettinarip, line 7:
```suggestion
```

src/theme.ts
pettinarip, line 1: Ohhh yes!

pettinarip (approved):
Awesome!

src/styles/global.css
```diff
@@ -81,7 +81,7 @@
 
 @layer base {
   body {
-    @apply !bg-background font-body !text-body;
+    @apply bg-background font-body leading-base text-body;
```
pettinarip, line 84: Good catch

PR #14940, opened 2025-02-19
refactor(tailwind.config): use current breakpoint set in maxWidth scale

TylerAPfledderer:
## Description

Extends the theming for the `maxWidth` utilities to use the current breakpoint values.

(this actually overrides existing token names)

### DIFF
--- a/src/components/Hero/HubHero/HubHero.stories.tsx
+++ b/src/components/Hero/HubHero/HubHero.stories.tsx
@@ -1,7 +1,4 @@
-import type { CSSProperties } from "react"
 import { Meta, StoryObj } from "@storybook/react"
-
-import { screens } from "@/lib/utils/screen"
 
 import { getTranslation } from "@/storybook-utils"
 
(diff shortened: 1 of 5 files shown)


pettinarip (commented):
Aren't we misusing this class?

Looks like what we really want to use is `max-w-screen-` instead of `max-w-`. Wouldn't overriding maxWidth be redundant?

If we override it, shouldn't we also override the `minWidth` classes?

TylerAPfledderer:
> Aren't we misusing this class?
> 
> Looks like what we really want to use is `max-w-screen-` instead of `max-w-`. Wouldn't overriding maxWidth be redundant?
> 
> If we override it, shouldn't we also override the `minWidth` classes?

@pettinarip Yep. I overlooked that utility when going through the docs.

I'll revert and apply this utility.

TylerAPfledderer:
@pettinarip alright, should be good here.

Chromatic regressions shown are outstanding from previous PRs (that I will review). I saw know regressions in the list specific to the changes made here.

pettinarip (approved):
Nice! looks good thanks @TylerAPfledderer

PR #14946, opened 2025-02-19
fix(SendReceive.stories): stop eth price updates during chromatic snapshots

TylerAPfledderer:
## Description

For the `SendReceive` component, this adds a check for chromatic to use the Fallback eth price when rendering the calculated value during the Chromatic snapshots. This will prevent API updates that would otherwise run a new snapshot for multiple stories when chromatic runs.

### DIFF
--- a/src/components/Simulator/screens/SendReceive/index.tsx
+++ b/src/components/Simulator/screens/SendReceive/index.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useMemo, useState } from "react"
+import isChromatic from "chromatic"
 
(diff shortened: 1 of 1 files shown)


TylerAPfledderer:
> Approving as this fixes the problem in the short term but wondering if we couldn't make a mock for the fetch instead, to keep minimal side effects.

@pettinarip currently, I am not sure if there is a way to mock this, since the fetch data is internal via `useEthPrice()`. I can review Storybook docs to see if mocking is possible as is.

I could've put `isChromatic()` inside `useEthPrice() for cleaner code.

Either way, I'll make a note for follow-up, and then check with you on what I find.

PR #14950, opened 2025-02-20
fix: apply default border color to all

wackerow:
## Description
- apply default `border-border` class to all elements

## Related Issue
Fixes recent border-color regressions with Chakra theming deprecation


pettinarip (approved):
Ah! interesting, I thought that this was already the applied default. Ty

PR #14955, opened 2025-02-21
fix Aaderyn -> Aderyn

BOR4:
## Description

Aderyn is the correct name of tool


BOR4:
Added new PR (#14956) where all translations are fixed as well. Closing this one

PR #14971, opened 2025-02-25
Fix Incorrect Terminology and Client Count in Portal Network Documentation

cjustinobi:
## Description

**Changes Proposed:**
1. **Diagram to Table Correction**:
   - The section incorrectly refers to a "diagram" when the content is clearly structured as a table. Updated the terminology to accurately describe the content as a **table**.

2. **Client Count Correction**:
   - The documentation states that there are **three Portal Network clients**, but the list provided includes **four clients** (Trin, Fluffy, Ultralight, and Shisui). Updated the text to reflect the correct number of clients listed on the site.

**Changes Made:**
- Replaced "diagram" with "table" in the relevant section.
- Updated the text to state "four Portal Network clients" instead of "three."


wackerow:
@all-contributors please add @cjustinobi for maintenance

PR #14977, opened 2025-02-26
Update index.md

tracehawk-io:


corwintines:
Thanks @tracehawk-io 

Wasn't able to look at the explorer as it seems I would need to sign up or pay for something first, kind of unclear. We only list open block explorers. Closing out.

PR #14989, opened 2025-02-27
docs: fix grammatical error: "alongside with" → "alongside"

hexcow:
## Description

I noticed a grammatical issue in the text: the phrase **"alongside with"** is incorrect because "alongside" already means "together with," making "with" redundant. I’ve removed "with" to make the sentence grammatically correct.  

Thanks!


corwintines:
Thanks @hexcow,

What is there already is correct so going to leave it as is.

PR #15017, opened 2025-03-02
Update index.md to explain why SHA-256 hashes are used in blockchains…

srishtiii28:
… Updated explanation on why SHA-256 is used in blockchains in context to Anders Brownworth's demo video

## Description

Added a paragraph explaining why SHA-256 is used in blockchains, covering its role in data integrity, security, immutability, and Proof of Work. Also highlighted its resistance to tampering and efficient verification process.

### DIFF
--- a/public/content/developers/docs/intro-to-ethereum/index.md
+++ b/public/content/developers/docs/intro-to-ethereum/index.md
@@ -22,6 +22,13 @@ Watch Anders explain hashes in blockchains:
 
 <YouTube id="_160oMzblY8" />
 
+
+**Why are SHA-256 hashes used in blockchains?**
+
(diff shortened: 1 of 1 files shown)


corwintines (changes requested):
Thanks @srishtiii28 

I am unsure if this is needed on this page, but this is definitely too early for the content on this page. I would be more open to it being appended to some of the definitions at towards the bottom.

wackerow:
Wouldn't be against having a snipper somewhere about how hashes work, but at the top of the "Intro to Ethereum" page would argue this is inappropriate, especially with Ethereum using keccak-256. Closing this out for now but feel free to open an issue to suggest an alternative. Thanks!

PR #15019, opened 2025-03-02
fix: dead link in doc

hunterfarq:
## Description

Hi! While reviewing the documentation, I noticed that a link to Foundry's testing framework (forge) was broken. The original link pointed to an outdated path (https://github.com/foundry-rs/foundry/tree/master/forge), which no longer exists. I’ve updated it to the correct path: https://github.com/foundry-rs/foundry/tree/master/crates/forge.


corwintines (approved):
Thanks @hunterfarq

PR #15032, opened 2025-03-07
improved readability in single slot finality file

YakshitAgarwal:
Link to page: https://ethereum.org/en/roadmap/single-slot-finality/#the-decentralization-time-overhead-tradeoff

Removed redundancy and improved readability without changing the meaning of the statement.

The original statement: 
"Each validating node on the network has to process attestations from other nodes in order to know that a block has, or has not, achieved that 2/3 threshold."

The changed statement: 
"Each validating node on the network must process attestations from other nodes to determine whether a block has achieved the 2/3 threshold."


corwintines:
Thanks @YakshitAgarwal 

This doesn't really change the content and would result in translation costs. Going to close this out.

PR #15051, opened 2025-03-11
docs: spanish - fixing translation

Ariiellus:
fixing translation

current translation title refers there is a type of node which "tracks" other nodes which is wrong

- native spanish speaker from mexico


corwintines:
@lukassim would you like me to bring this in?

Memohalliwell (approved):

Memohalliwell (approved):

PR #15058, opened 2025-03-11
Correctly set HTML lang attribute

minimalsm:
## Description
The `lang` html attribute was always defaulting to "en" regardless of the actual page locale. My understanding is we use the `lang` to split our search indexes, so this bug is causing all languages to be [searchable in English](https://discord.com/channels/420394352083337236/426091468403376128/1349122790258770013), and none of the non-English languages have any search hits at all.

Screenshot 2025-03-11 at 21 17 54

Solution is... maybe hacky? I modified _document.tsx to extract the locale from query parameters in getInitialProps. This ensures the `lang` is properly set for the crawler to see.

Screenshot 2025-03-11 at 22 33 07

After testing/merging this PR, we need to run a [recrawl on Docsearch to fix the bug](https://crawler.algolia.com/admin/crawlers/d9cd038f-c91f-471d-99bc-f9a3f370beb3/overview)

### DIFF
--- a/src/pages/_document.tsx
+++ b/src/pages/_document.tsx
@@ -12,11 +12,15 @@ import { isLangRightToLeft } from "@/lib/utils/translations"
 
 class Document extends NextDocument {
   static async getInitialProps(ctx: DocumentContext) {
-    return await NextDocument.getInitialProps(ctx)
+    const initialProps = await NextDocument.getInitialProps(ctx)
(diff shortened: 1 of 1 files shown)


pettinarip (approved):
Spotted a minor issue: when you switch lang or navigate to a different lang, the attr is not updated but it is when you reload.

Approved since the previous bug is not critical and the issue will be fixed soon when we finish the app router migration.

PR #15097, opened 2025-03-13
AI agents page

corwintines:
https://deploy-preview-15097--ethereumorg.netlify.app/ai-agents


src/components/Content/ai-agents/BuildYourOwnAIAgent.tsx
pettinarip, line 1: Dont have a better solution atm with the current setup but I don't like the direction of these very specific components that we can't reuse and that don't share/reuse other existing styles or components.

Was thinking that maybe we could fix this in the md files itself. Where you can handle translations and images. Something like

```mdx
<Boxes>
  <Box>
    ![AI16Z](./ai16z.png)
    
    blabla
  </Box>
  <Box>
    ![GAME](./game.png)
    
    blabla
  </Box>
</Boxes>
```

Unfortunately, this is not supported in the current mdx version (mdx syntax inside jsx) but will be possible after we merge #14968 ! since we are also upgrading the mdx version.

src/components/ProductList.tsx
```diff
@@ -26,43 +29,57 @@ const ProductList = ({ actionLabel, content, category }: ProductListProps) => {
...
-              )}
-            </Flex>
-          </ListItem>
-        ))}
+              <Flex className="ms-4 w-full flex-col justify-between pb-4 sm:flex-row">
+                <div className="flex flex-1 flex-col gap-2">
+                  <div>{title}</div>
+                  <div className="mb-0 text-sm opacity-60">{description}</div>
+                  <div className="mb-0 flex flex-col gap-2 text-sm opacity-60">
+                    {contentItems}
```
pettinarip, line 66: Given that `contentItems` is optional, shouldn't we consider that case and avoid rendering the wrapping div?

pettinarip (approved):
LGTM!

konopkja (changes requested):

PR #15108, opened 2025-03-15
Fix typo on weak subjectivity article

RogerPodacter:
Change “to” to “as.”


corwintines (approved):
Thanks @RogerPodacter

PR #15137, opened 2025-03-19
Fix listen-to story

pettinarip:
Our current Chromatic builds are failing.

## Description

- Removes the BaseLayout wrapper to test this component in isolation (fixing also the error in chromatic)
- Adds an interactive test to have a snapshot of the player widget
- Convert the ListenToPlayer into a real button element to improve navigation and a11y


src/components/ListenToPlayer/ListenToPlayer.stories.tsx
```diff
@@ -1,46 +1,42 @@
...
+
+export const PlayerWidget: Story = {
+  play: async ({ canvasElement }) => {
+    const canvas = within(canvasElement)
+    const canvasParent = within(canvasElement.parentElement!)
+
+    const playerButton = canvas.getByRole("button")
+    await waitFor(() => {
+      // TODO: hacky way to wait for the sound to be loaded
+      expect(playerButton.textContent).not.toContain("0 min")
```
wackerow, line 31: @pettinarip Would a `10 min` track cause a false positive here?

wackerow: nm, just a story demo, should be good

wackerow (approved):

Afeter8 (approved):

PR #15144, opened 2025-03-20
Added a new asset tokenization use case

janjuanumair:


corwintines:
Not much information here, even looking at the site. 

When looking at analytics of the app, im pretty underwhelmed with 3 listings, 46 holders, and 30k in volume. I don't think this is ready to be listed, especially on the enterprise page.

PR #15166, opened 2025-03-25
update broken TrustToken link to new TUSD website

timbalabuch:
Replaced outdated link (https://www.trusttoken.com/) with the current official TUSD site (https://tusd.io/). The old link was broken.


corwintines (approved):
Good catch, thanks @timbalabuch

PR #15170, opened 2025-03-25
Hide listen to player on non-english pages

corwintines:
## Description

- Hides the listen to player on non english pages since this content is only in english at the moment


wackerow (approved):
Nice, works for now... if new languages get added can find a way to query for which are available.

wackerow (approved):
clean 😎

corwintines:
> Nice, works for now... if new languages get added can find a way to query for which are available.

Yeah, thats a future Corwin problem XD

PR #15177, opened 2025-03-26
Update index.md Fix Hyperlink

Hunter-Plus:
https://www.ethernodes.org/networkType/Hosting the old URL is already deprecated (404). Now we must visit https://www.ethernodes.org/networkType/cl/Hosting to see the hosting distribution.


wackerow:
@all-contributors please add @Hunter-Plus for maintenance

PR #15200, opened 2025-03-28
SEO: filter invalid locales from alternates

pettinarip:
## Description

- adds `validISO639_1` prop to each language config
- updates `PageMetadata` to filter locales from `alternate` metatags


i18n.config.json
```diff
@@ -5,550 +5,619 @@
     "name": "English",
     "localName": "English",
     "langDir": "ltr",
-    "dateFormat": "MM/DD/YYYY"
+    "dateFormat": "MM/DD/YYYY",
+    "validISO639_1": true
```
minimalsm, line 9: curious why include this for all instead of just adding `invalidISO0639_1` to `pcm` and `fil`?

pettinarip: I prefer a consistent scheme for better typing, and usually go with positive names for clearer logic. That said, I'm not strongly attached to it. Happy to switch to your suggestion if you feel strongly about it.

Memohalliwell (approved):

Memohalliwell (approved):

corwintines (approved):
LGTM as is so will approve, but saw @minimalsm comment so will leave open in case you make changes

wackerow (approved):

PR #15202, opened 2025-03-28
Hotfix reading time for tutorials

pettinarip:
## Description

Removes hardcoded read time for tutorials and calculate it using `reading-time` lib.

### DIFF
--- a/app/[locale]/[...slug]/page.tsx
+++ b/app/[locale]/[...slug]/page.tsx
@@ -63,6 +63,7 @@ export default async function Page({
     lastEditLocaleTimestamp,
     isTranslated,
     contributors,
+    timeToRead,
   } = await getPageData({
     locale,
     slug,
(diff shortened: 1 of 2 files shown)


corwintines (approved):

app/[locale]/[...slug]/page.tsx
```diff
@@ -97,8 +98,7 @@ export default async function Page({
         lastEditLocaleTimestamp={lastEditLocaleTimestamp}
         contentNotTranslated={!isTranslated}
         contributors={contributors}
-        // TODO: Remove this once we have a real timeToRead value
-        timeToRead={2}
+        timeToRead={Math.round(timeToRead.minutes)}
```
Comment (vm06007) [inline app/[locale]/[...slug]/page.tsx:101]: might need to add a default value here as a fallback in case `timeToRead.minutes` returns something undefined

Afeter8 (approved):

PR #15220, opened 2025-04-01
fix: incorrect `Promise<>` usage in `params`

barajeel:
## Description

Noticed that `params` was incorrectly typed as `Promise<>`, but in Next.js, it’s always passed as a regular object.
Removed the unnecessary `Promise<>` so the code works as expected.


wackerow:
@all-contributors please add @barajeel for bug fix

PR #15225, opened 2025-04-03
Update index.md

tracehawk-io:
TraceHawk: A white-labeled blockchain explorer designed based on requirements of EVM chains, Appchains, and L2/L3 rollups. With TraceHawk, projects can have their custom block explorer with preferred features, capabilities, and intuitive design. Also, TraceHawk comes with managed and hosted services to streamline explorer maintenance and scaling to ensure performance, security, and compliance to avoid heavy lifting on your end.


corwintines:
Thanks @tracehawk-io

We only list explorers that are open and not behind a paywall for accessibility. Going to close ths out.

PR #15230, opened 2025-04-03
chore: fix `params` type in `Page` and `generateMetadata` functions

0xsenty:
## Description

I’ve corrected an issue where the `params` were being passed as a promise (`Promise<{ locale: Lang }>`), while it should have been passed directly as an object (`{ locale: Lang }`).

This fix ensures the correct type handling in both the `Page` and `generateMetadata` functions, preventing potential runtime errors.

### DIFF
--- a/app/[locale]/community/page.tsx
+++ b/app/[locale]/community/page.tsx
@@ -34,7 +34,7 @@ export default async function Page({
 export async function generateMetadata({
   params,
 }: {
-  params: { locale: Lang }
+  params: { locale: string }
 }) {
   const { locale } = params
 


app/[locale]/community/page.tsx
```diff
@@ -34,9 +34,9 @@ export default async function Page({
 export async function generateMetadata({
   params,
 }: {
-  params: Promise<{ locale: string }>
+  params: { locale: Lang }
```
Comment (wackerow) [inline app/[locale]/community/page.tsx]:
```suggestion
  params: { locale: string }
```

Copilot summarizes nicely:


In Next.js 14, the `params.locale` value is always passed as a plain string from the route segment, not as a custom type.  
- For Page props, using `Lang` can help with type safety if you validate or restrict locales elsewhere in your app.
- For `generateMetadata`, using `string` is standard because Next.js passes the raw route param, and it avoids type errors if an unexpected locale is provided.

**Summary:**  
- Use `string` for `params.locale` in `generateMetadata` for compatibility and to match Next.js expectations.
- Use `Lang` in Page props only if you want stricter type safety and you validate/convert the value before using it.

If you want consistency, you can use `string` everywhere, but using `Lang` in Page props is fine if you handle validation. For `generateMetadata`, prefer `string`.

wackerow:
@all-contributors please add @0xsenty for bug fix

Issue #15233, opened 2025-04-03
Update full node size in whitepaper and add source so reader can see current state at time of reading


johnnieskywalker:
In Ethereum whitepaper there's statement that was not updated for 11 years

```
The Merkle tree protocol is arguably essential to long-term sustainability. A "full node" in the Bitcoin network, one that stores and processes the entirety of every block, takes up about 15 GB of disk space in the Bitcoin network as of April 2014, and is growing by over a gigabyte per month. 
```

I suggest to update it and add trustworthy source for full node size if reader wants to dig deeper.

Suggested update:

```A "full node" in the Bitcoin network, which stores and processes the entirety of every block, requires approximately [600 GB](https://www.blockchain.com/explorer/charts/blocks-size) of disk space as of April 2025, growing by roughly 50-60 GB per year due to the consistent addition of new blocks.
```

corwintines:
The whitepaper has intentionally been left in the state it was written in for citation purposes. Going to close this out.

PR #15234, opened 2025-04-03
Updated the Bitcoin full node size to state from April 2025

johnnieskywalker:
## Description

So in current version of whitepaper, that was frequently update there's still Bitcoin's full node size from April 2014. As 11 years passed I think this sentence deserves update and link to source that allows reader to check current state at the moment of reading. 

The never-updated statement

_The Merkle tree protocol is arguably essential to long-term sustainability. A "full node" in the Bitcoin network, one that stores and processes the entirety of every block, takes up about 15 GB of disk space in the Bitcoin network as of April 2014, and is growing by over a gigabyte per month._ 

Suggested update:

A "full node" in the Bitcoin network, which stores and processes the entirety of every block, requires approximately [600 GB](https://www.blockchain.com/explorer/charts/blocks-size) of disk space as of April 2025, growing by roughly 50-60 GB per year due to the consistent addition of new blocks.

## Related Issue

[Issue 15233](https://github.com/ethereum/ethereum-org-website/issues/15233)


wackerow:
Hey @johnnieskywalker, thanks for the suggestion. Whitepaper is a page we've left as-is for historical purposes so going to close out. Cheers!

Issue #15231, opened 2025-04-03
fix: back to top button in docs


konopkja:
### Describe the bug

The back to top button is broken 

### To reproduce

1. Go to **https://ethereum.org/en/developers/docs/intro-to-ethereum/**
2. Scroll down to bottom part with feedback buttons
3. See error

### Expected behavior

fix how it looks

### Screenshots

![Image](https://github.com/user-attachments/assets/88a7e7c0-c7ea-4ba4-b415-b402c4eed815)

### Would you like to work on this issue?

- [ ] Yes
- [x] No

heronimus:
Hi @konopkja, I’d like to give this issue a try.

heronimus:
Hi @konopkja,

I’d like to confirm my findings after reviewing the layout code:
- The `back-to-top` link was already present but broken as of at least March 8, 2024 (88f0dd7), on [this preview link](https://deploy-preview-12403--ethereumorg.netlify.app/en/developers/docs/) (visible only on narrow windows).
- The current layout has two back-to-top links/buttons:
   - One above the feedback card, which this issue refers to.
   - Another in the footer, which is used globally across other non-developer documentation pages.

![Image](https://github.com/user-attachments/assets/af7e0106-8bd0-4fc7-a269-ccd46c39b337)  ![Image](https://github.com/user-attachments/assets/36092122-eb7c-4324-88f3-43835da3a0ac)


### Proposed Fix
- I already fixed the broken text in my recently opened PR (#15236) by ensuring the correct sentence/translation for the **“Back to Top”** text while maintaining the existing text style. 
  Let me know if we’d prefer changing it to a button or if there’s already a Figma design for it.

- Regarding the duplicate back-to-top buttons, since the footer version is globally used, modifying it now might affect other pages. The best approach would be to leave it as is and create a separate issue for further discussion.

konopkja:
Hi, the fix you created is good, ty <3!

CodeLeom:
If this issue has been fixed, can it be closed? so we can tidy the place

konopkja:
> If this issue has been fixed, can it be closed? so we can tidy the place

lets close it, once the PR is merged

butterpaneermasala:
can i work on this?

wackerow:
This issue is resolved

PR #15236, opened 2025-04-03
fix: developer docs back-to-top link translation id (close: #15231)

heronimus:
## Description
- The `back-to-top` link currently has an invalid translation-id, which prevents the text from displaying correctly. The correct ID should be `page-developers-docs:back-to-top`.

- Before:
image

- After (EN & FR):
image
image

## Related Issue

- Close: #15231
- A potentially related issue was mentioned in the issue thread discussion.


wackerow:
Hey @heronimus, sorry for delay here... appreciate the patch but looks like this has been fixed at this point. Cheers!

PR #15242, opened 2025-04-05
Reword install deps readme section

TheAlmightyMight:


TheAlmightyMight:
Not sure why there are these two other commits but they don't seem to introudce any additional changes. Is that ok?

nilevols:
I don't think your "unclean" PR commits would bother, they are just merge commits.

Issue #15099, opened 2025-03-13
Update or revamp client diversity


konopkja:
https://ethereum.org/en/developers/docs/nodes-and-clients/client-diversity/

Information regarding client diversity seems to be really out of date, also reth is not even mentioned. 

<img width="868" alt="Image" src="https://github.com/user-attachments/assets/8770e6ea-90fa-4dd9-b52b-c5af1f9e3fbb" />

**Potential solutions:**

1. Ideally try to use API to get this data dynamically instead of an image. 
2. update the image and above it add a message "This diagram may be outdated, go to [ethernodes.org](https://ethernodes.org/) and [clientdiversity.org](https://clientdiversity.org/) for up to date informaiton"

LifeofDan-EL:
Yeah, I can work on that and have it as a chore to update the values, weekly, twice a month or monthly.

konopkja:
> Yeah, I can work on that and have it as a chore to update the values, weekly, twice a month or monthly.

thank you so much!!!!!

LifeofDan-EL:
Hello, so an update on this. I wanted something that works with markdown by default and I picked [Mermaid](https://mermaid.js.org/). It is supported by Docusaurus, MkDocs, Hugo, Obsidian, GitLab Wiki by default.

This approach only modifies the index.md file under client diversity, and it can be updated by just changing the values. However, the fall back is the colour coding, which can't be modified from markdown. 

![Image](https://github.com/user-attachments/assets/da64269c-c5d7-477a-80a4-15847e1c5600)

pettinarip:
Hey @LifeofDan-EL, thanks for the update!

Mermaid looks great and it’s definitely nice that it works well with markdown. That said, we already have a charting library installed as part of the UI stack ([shadcn/ui charts](https://ui.shadcn.com/charts)), which is a bit lighter and already integrated with our theming system. Since we’re just adding a single chart, we’d prefer to avoid bringing in an additional dep like Mermaid, which is a bit heavier. Let’s try to stick with the existing charting setup if possible.

Here are some examples that we are currently using. We could create one for PieChart
- https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/BarChart/index.tsx
- https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/AreaChart/index.tsx

LifeofDan-EL:
[...]

PR #15251, opened 2025-04-07
Client diversity piechart update

LifeofDan-EL:
## Description
This PR is to create a new piechart that changes depending on the values, which is to be used to update the client diversity part of the website. 

#15099

Preview: https://deploy-preview-15251--ethereumorg.netlify.app/en/developers/docs/nodes-and-clients/client-diversity/


LifeofDan-EL:
Deploy is failing, at first, I thought it was a lint error, so I fixed that with `npx eslint ./src/components/MdComponents/index.tsx --fix` but it is still failing.

public/content/developers/docs/nodes-and-clients/client-diversity/index.md
```diff
@@ -5,6 +5,27 @@ lang: en
 sidebarDepth: 2
 ---
 
+import { PieChart } from "@/components/PieChart"
```
pettinarip, line 8: This import is redundant now that we are injecting the component in the MdComponents file.

Also, `next-mdx-remote` (the mdx compiler that we use) doesn't support imports or exports, so we need to do something different to pass the below data to the component.

I think the easiest would be to pass the data array inline like this:
```tsx
<PieChart data={[ { name: "Geth", value: 43 }, ... ]} title="Execution Clients" />
```

To clarify more, you can't use variables like `executionData` without exporting them in mdx. But since imports/exports are not supported, we can't use them either.

pettinarip (changes requested):
@LifeofDan-EL left a comment about the issue we have.

On the other hand, there are a few files in this PR that shouldn't be modified/added/deleted like `yarn.lock`, `.env.example`, and some yarn related ones that are not necessary. Can help you clean them up once we have a working preview.

public/content/developers/docs/nodes-and-clients/client-diversity/index.md
LifeofDan-EL: Thanks @pettinarip  for the clarification, it was an oversight pushing the yarn file. Thanks again.

src/components/PieChart/index.tsx
```diff
@@ -0,0 +1,133 @@
...
+    label: "Value",
+    color: "hsl(var(--accent-a))",
+  },
+} satisfies ChartConfig
+
+/**
+ * Color palette for blue shades, ordered from deep blue (largest share) to light blue (smallest share).
+ * Ensure your data is sorted from largest to smallest for this to work as intended.
+ */
+const COLORS = [
```
pettinarip, line 51: There are a few issues with these colors:
- there is not enough contrast in dark mode
- these are not part of our current color palette

This is the current colors that we have https://dev--63b7ea99632763723c7f4d6b.chromatic.com/?path=/story/design-system-colors--primitives

@konopkja what do you think?

.yarn/install-state.gz
pettinarip, line 1: This file is not needed

.yarnrc.yml
pettinarip, line 1: This file is not needed

pettinarip (changes requested):
@LifeofDan-EL good job. The implementation looks good and is functional. Left a comment about the styles to try to align it with our current design system.

.yarn/install-state.gz
LifeofDan-EL: Resolved.

src/components/PieChart/index.tsx
LifeofDan-EL: Yeah, I looked at the preview after deployment and I agree there isn't enough contrast. Let me use values from the colour palette.

LifeofDan-EL:
The latest build currently has the accent colour and I tweaked the legend a bit. What's left is to work on the responsiveness. 

Once that is done, then I can update the content too, to reflect the new values.

LifeofDan-EL:
I am done. I believe the PR is ready for final review now.

image

konopkja:
Screenshot 2025-04-23 at 9 28 43

there is some visual overlap bug at the bottom

could we reduce the line-height?  It seems double than what it should be.


Other than that this is a huge improvement!

LifeofDan-EL:
> Screenshot 2025-04-23 at 9 28 43
> there is some visual overlap bug at the bottom
> 
> could we reduce the line-height? It seems double than what it should be.
> 
> Other than that this is a huge improvement!

Hello @konopkja, when you say line-height, is it the spacing between the legends or? Also, I have taken note of the overlap bug too.

konopkja:
> > Screenshot 2025-04-23 at 9 28 43
> > there is some visual overlap bug at the bottom
> > could we reduce the line-height? It seems double than what it should be.
> > Other than that this is a huge improvement!
> 
> Hello @konopkja, when you say line-height, is it the spacing between the legends or? Also, I have taken note of the overlap bug too.

yes spacing between each row of the text for the legend

konopkja:
Screenshot 2025-04-30 at 16 19 16

noting that lowering the space between legends wont probably solve the issue for mobile view.

LifeofDan-EL:
Hello @pettinarip, I believe this is ready to be reviewed, and you can take the necessary edits to get it merged. Thanks for your patience.

src/components/PieChart/index.tsx
```diff
@@ -0,0 +1,60 @@
+"use client";
+import React from "react";
+import Link from "next/link";
+
+import { PieChart } from "./PieChart";
+
+const PieChartContainer = () => {
```
pettinarip: `PieChartContainer` component is tightly coupled to the client diversity use case making it hard to reuse.

We have different options to fix this. One could be to create a new component called `ClientDiversityChart` with the current layout and copy, making it clearer that this is for the client diversity page.

```tsx
  // current implementaiton
  <PieChartContainer />

  // proposal: data lives in the md file and is wrapped with the new `ClientDiversityChart` component
  // PieChart is already reusable +1
  <ClientDiversityChart>
    <PieChart data={[...]} title="Client Diversity" />
    <PieChart data={[...]} title="Client Diversity" />
  </ClientDiversityChart>
```

pettinarip (changes requested):
@LifeofDan-EL looking good. I think there is one thing we should polish with the chart implementation before we can get this in.

src/components/PieChart/index.tsx
LifeofDan-EL: Okay, you're correct. Let me get to work on this.

LifeofDan-EL:
I had to make sure things were in sync with the current dev before continuing.

wackerow:
Thanks @LifeofDan-EL! Will take a look shortly

LifeofDan-EL:
> Thanks @LifeofDan-EL! Will take a look shortly

Thanks, it is compiling but, the build is failing weirdly and I have no clue how to fix that. I would appreciate the help. Thanks.

wackerow (changes requested):
Phew! 😮‍💨... first of all, appreciate the work you've put into this @LifeofDan-EL! Handful of stuff to address, but it's still a lot of progress and I think we can get through this...

First, there are some extraneous updates in this PR that need to be reverted, and are causing issues:

1. The .prettierrc file has been deleted—this must be reverted
2. That deletion led to auto-formatting updates that must also be reverted, e.g., the addition of trailing semi-colons, which the prettier config would normally prevent
3. The MdComponents/index.tsx should _only_ include the addition of the new component we're building, but everything else should remain unchanged
4. We use `pnpm` for package management, so nothing yarn-related should be in here

Second, file naming and re-usability... This PieChart component looks decent in terms of reusability at this point, but we should rename the file to `PieChart/index.tsx`... I have some suggestions to clean up the errors we're getting, but it was done quickly and I think we need a couple small layout adjustments still, and with those we should be careful to pay attention to how this component will work with _any_ set of data we feed it on _any_ page we place it. Techniques here can include making sure `className` is passed properly as a prop, and merged with the base classes using `cn("base classes", className)` syntax imported from `@/lib/utils/cn`. This lets us customize on a case-by-case bases when we consume this component from a parent page or other component.

Third... typing. We should not ignore the eslint rules, including the bypassing of the `no-explicit-any` rules. There are ways we can make the typing robust to avoid needing to do this.

https://gist.github.com/wackerow/ced8f9db60cdebfe5720234227d24198

Locally I made some adjustments and saved them to this gist which you should be able to view. The exact layout adjustments should be taken with a grain of salt here, but more importantly are the **updates to the typing**, which eliminates the usage of `any` or even `unknown`, since we do indeed know the types we'll need. 

```tsx
import type { Formatter } from "recharts/types/component/DefaultLegendContent"
```

See line 148 for example of how this is used, and the next line for how we can make sure the payload value is indeed a number before proceeding.

It also demonstrates a few other things such as:

- Usage of guard clauses to return early if certain conditions aren't met (Line 170, line 174)
- Usage of tailwind class names instead of style props where possible
- Updates the stroke color on the pie chart to use full digit-character hex codes—the three character shorthand form wasn't working for me
- Removes the angular gap which made things look goofy

Lastly is the custom wrapper component, `ClientDiversityChart/index.tsx`... overall this looks decent but a couple notes:

- Our <Link> component will automatically parse the `href` and apply the `target` and `rel` for any external links, so these additions are redundant and can be removed... Also we should use standard styling, so can also remove the `className="underline"` here.
- A slightly more challenging step which is a nice-to-have for this PR, would be setting these English strings up for translation. The markdown file works a bit differently, where that entire file will be translated and separately translation-specific copies will be imported through a separate work stream, but for React pages/components such as the few English strings in here it works a bit differently. Honestly, I wouldn't worry about this for the sake of this PR, but we'll need to make sure that when this is completed that we tee-up a follow-up task to extract these strings into a JSON file within src/intl/en/**.json where they can then be references via a unique key identifier, instead of hard-coded. Just noting for your understanding.

---

For the sake of this PR, you _may_ want to consider simply closing this branch out and starting fresh by:
1. branching off of an up-to-date `dev` branch
2. creating the new ClientDiversityChart/index.tsx and PieChart/index.ts components fresh
3. Importing what is necessary into MdComponents/indes.tsx
5. Updating the `index.md` file where the chart will be used (that part shouldn't need to change at all from what I can tell)

This will help clean up the rest of the regressions and make things easier to review, and should help the build pass successfully.

Again, thanks so much for the help here... hopefully that gist link helps provide the guidance needed to get this patched up so we can get it over the line. Let us know if you have any other issues!

LifeofDan-EL:
Thanks so much @wackerow  for the detailed eye. I will close this PR and work on it afresh, incorporating the feedback with a new branch.

wackerow:
@LifeofDan-EL Sounds great! Appreciate that, should help tidy things up =)... going to close this one out for now, looking forward to the update

Issue #15278, opened 2025-04-14
Extra 0xc0 byte in RLP example in French and PT-BR versions


antoinelanco:
Hi,

I noticed an issue with the RLP encoding examples in the French (fr) and Portuguese-Brazilian (pt-br) versions of the documentation.

In the example for [ [], [[]], [ [], [[]] ] ], there is an extra 0xc0 byte included in the encoded output.

The current (incorrect) RLP encoding is:
`[ 0xc7, 0xc0, 0xc1, 0xc0, 0xc3, 0xc0, 0xc0, 0xc1, 0xc0 ]`
It should be corrected to:
`[ 0xc7, 0xc0, 0xc1, 0xc0, 0xc3, 0xc0, 0xc1, 0xc0 ]`

You can see the correct examples here:
[English version (line 83)](https://github.com/ethereum/ethereum-org-website/blob/dev/public/content/developers/docs/data-structures-and-encoding/rlp/index.md?plain=1#L83)

You can see the incorrect examples here:
[French version (line 83)](https://github.com/ethereum/ethereum-org-website/blob/dev/public/content/translations/fr/developers/docs/data-structures-and-encoding/rlp/index.md?plain=1#L83)
[Portuguese-Brazilian version (line 83)](https://github.com/ethereum/ethereum-org-website/blob/dev/public/content/translations/pt-br/developers/docs/data-structures-and-encoding/rlp/index.md?plain=1#L83)

Could you please update the French and PT-BR examples to fix this?

Thanks a lot for your work!

antoinelanco:
Hi @isabelladebrito, Thx, I think the [PR](https://github.com/ethereum/ethereum-org-website/pull/15279) is already merged, in any case, the [site](https://ethereum.org/fr/developers/docs/data-structures-and-encoding/rlp/) has been updated.

isabelladebrito:
Oops, you're right, @antoinelanco! I'll this issue then. 

If you haven't collected your ethereum.org GitPOAP yet, I recommend you to check out [this link](https://github.com/ethereum/ethereum-org-website/pull/15279#issuecomment-2808083305), as well as our [contributor badges on Galxe](https://ethereum.org/en/contributing/#oat).

Thanks again for your help! 🙌

PR #15279, opened 2025-04-14
fix rlp example for fr and pt-br

antoinelanco:
## Related Issue
https://github.com/ethereum/ethereum-org-website/issues/15278


corwintines (approved):
Thanks @antoinelanco!

PR #15282, opened 2025-04-14
Fix load messages2

pettinarip:


wackerow (approved):
Looks great, and working well on my end. Thanks @pettinarip! pulling in

PR #15308, opened 2025-04-17
Update index.md

tracehawk-io:
Added TraceHawk to the list of block explorers. 
TraceHawk: TraceHawk is a white-labeled blockchain explorer designed based on requirements of EVM chains, Appchains, and L2/L3 rollups. With TraceHawk, projects can have their custom block explorer with preferred features, capabilities, and intuitive design. Also, TraceHawk comes with managed and hosted services to streamline explorer maintenance and scaling to ensure performance, security, and compliance to avoid heavy lifting on your end.


corwintines:
I can't just use this out of the box it seems. When I go to this website, all I can do is request a demo. For that reason I am going to close this out.

PR #15310, opened 2025-04-17
Roadmap page

corwintines:
## Description

- Implement new design for roadmap landing page


konopkja:
Screenshot 2025-04-28 at 12 11 57
@corwintines

src/data/roadmap/releases.tsx
```diff
@@ -0,0 +1,194 @@
...
+  content: React.ReactNode[]
+  href: string
+}
+
+export const releasesData: Release[] = [
+  {
+    image: DevelopersHubHeroImage,
+    releaseName: "Paris (The Merge)",
+    releaseDate: "2022-09-15",
+    content: [
```
pettinarip, line 23: why is content an array? seems like we are always rendering one node

src/data/roadmap/releases.tsx
```diff
@@ -0,0 +1,194 @@
...
+      <div key="1">
```
pettinarip, line 24: I guess `key` is useless here, no?

app/[locale]/roadmap/_components/ReleaseCarousel.tsx
```diff
@@ -0,0 +1,272 @@
...
+}
+
+const ReleaseCarousel = () => {
+  const todayDate = new Date()
+  const twoMonthsFromNow = new Date()
+  twoMonthsFromNow.setMonth(todayDate.getMonth() + 2)
+
+  const [api1, setApi1] = useState<CarouselApi>()
+  const [api2, setApi2] = useState<CarouselApi>()
+  const [currentIndex, setCurrentIndex] = useState(findLatestReleaseIndex())
```
Comment (pettinarip) [inline app/[locale]/roadmap/_components/ReleaseCarousel.tsx:65]: if the intention here is to have an initial value, we should call it with a function ([initialized function](https://react.dev/reference/react/useState#parameters)). Otherwise, `findLatestReleaseIndex` would be called in every render.
```suggestion
  const [currentIndex, setCurrentIndex] = useState(() => findLatestReleaseIndex())
```

app/[locale]/roadmap/_components/ReleaseCarousel.tsx
```diff
@@ -0,0 +1,272 @@
...
+  CarouselItem,
+  CarouselNext,
+  CarouselPrevious,
+} from "@/components/ui/carousel"
+
+import { cn } from "@/lib/utils/cn"
+
+import { releasesData } from "@/data/roadmap/releases"
+
+const formatReleaseDate = (date: string) => {
```
Comment (pettinarip) [inline app/[locale]/roadmap/_components/ReleaseCarousel.tsx:20]: can we place it in `lib/utils/date` as `formatDate`?

app/[locale]/roadmap/_components/ReleaseCarousel.tsx
```diff
@@ -0,0 +1,272 @@
...
+                        </div>
+                        <div className="flex w-full items-center justify-center text-center">
+                          <div
+                            className={cn(
+                              "flex h-1 flex-1",
+                              index !== 0
+                                ? nextRelease
+                                  ? "bg-gradient-to-r from-primary to-primary-low-contrast"
+                                  : release.releaseDate <
+                                      new Date().toISOString().split("T")[0]
```
Comment (pettinarip) [inline app/[locale]/roadmap/_components/ReleaseCarousel.tsx:159]: we could reuse some of the vars already defined and have a safer comparison with numbers
```suggestion
                                  : releaseDate.getTime() < todayDate.getTime()
```

pettinarip (approved):
nice job @corwintines! looks great. Left a few observations but nothing blocking from my side.

src/data/roadmap/releases.tsx
corwintines: Artifact of old idea for how I was going to do this, will refactor.

nloureiro (approved):

konopkja (approved):

Issue #15312, opened 2025-04-18
Fix Markdown content pages' link references


juanchax:
### Is your feature request related to a problem? Please describe.

Currently, the content pages contain a multitude of hardcoded link references, which of course are duplicated on the translated versions as well.

Causing:
- There's no single source of truth
- brittle approach
- Difficult to know if all links are synched (as in: all have the same url)
- Updating dead links is a laborious task to say the least 

Here's a bird's eye view of the relevant & related items:

| Issue      | Description |
| ----------- | ----------- |
| [Dead links around site #14823](https://github.com/ethereum/ethereum-org-website/issues/14823)      | A neat report of all dead links found across website pages |
| [Fixed broken link on Contributing Page #14605](https://github.com/ethereum/ethereum-org-website/pull/14605/files) | A pr that updated `/public/content/index.md` only |
| [Old link for article about OATs on Contributing page #14200](https://github.com/ethereum/ethereum-org-website/issues/14200) | An open issue for the same link reference (More on OATs) |
| [Bug report #14711](https://github.com/ethereum/ethereum-org-website/issues/14711) | Same issue, different translated page |

But rest assured, <a id=related-issues></a>[many more are coming](https://github.com/search?q=repo%3Aethereum%2Fethereum-org-website%20https%3A%2F%2Fhelp.galxe.com%2Fen%2Farticles%2F7067290-galxe-oats-reward-and-celebrate-achievements&type=code) &#128565;!

## &#128161;  Parameterized Link References

The idea is to apply a _modularity mindset_ and parameterize references to external links, using some sort of variable injection in the `content md` files; thus creating a _single source of truth_ for references to external URLs used across the website, and simplifying maintenance tasks:

* link references updated
* content pages consistent
* periodically check and update dead link references
* reduced duplication, improved reusability

## &#129302; Isolated modular Implementation Built & Ready for Review

I already implemented the following in an isolated repo, so as to minimize introducing errors (between the keyboard and the chair) and refactoring. 

[...]

Issue #14711, opened 2025-01-20
Bug report: outdated link


quantumopticss:
### Describe the bug

The page about OAT is missing 

### To reproduce

1. Go to 'https://ethereum.org/zh/contributing/'
2. Click on 'more about OAT'
3. Scroll down to 'https://help.galxe.com/en/articles/7067290-galxe-oats-reward-and-celebrate-achievements'
4. See error : the page doesn't exist

### Expected behavior

There should be a page about OAT in https://help.galxe.com/en/articles/7067290-galxe-oats-reward-and-celebrate-achievements

### Would you like to work on this issue?

- [ ] Yes
- [ ] No

JoeChenJ:
Hi, I think the correct link should be "https://help.galxe.com/en/articles/9645630-create-quest-rewards#h_1c5d63ba03".

I can fix this problem.

PR #14605, opened 2025-01-06
Fixed broken link on Contributing Page

timbalabuch:
Fixed broken link on [More On OATs]

Old broken link: https://help.galxe.com/en/articles/7067290-galxe-oats-reward-and-celebrate-achievements
New link: https://help.galxe.com/en/articles/9645630-create-quest-rewards#h_1c5d63ba03

Issue #14200, opened 2024-10-20
Old link for article about OATs on Contributing page


UNOFFICIALbgd:
### Describe the bug

Link is not working

### To reproduce

1. Go to 'https://ethereum.org/en/contributing/#oat'
2. Click on 'More on OATs'
3. See error

### Expected behavior

As there is no any article about OATs on https://help.galxe.com/en/, let's use [this article](https://medium.com/galxe-news/introducing-galaxy-oat-on-chain-achievement-token-7e89779242b4) instead

### Would you like to work on this issue?

- [X] Yes
- [ ] No

Nanle100:
i am interested in this.
can i work on it

PR #15323, opened 2025-04-20
[ Enhancement ] Adding Markdown preprocessor

juanchax:
## Description

Updated external url references contained in /public/content/contributing/index.md to {variable}
Populated variables.ts with external url references contained in /public/content/contributing/index.md
Added preprocess.ts with the markdown preprocessing logic
Removed preprocessMarkdown() from compile.ts
Imported preprocessMarkdown() from preprocess.ts

## Related Issue

Background for this enhabcement --> https://github.com/ethereum/ethereum-org-website/issues/15312

Base issues this enhancement is attempting to resolve and prevent future ones -->

https://github.com/ethereum/ethereum-org-website/pull/14605

https://github.com/ethereum/ethereum-org-website/issues/14200

https://github.com/ethereum/ethereum-org-website/issues/14711

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,8 +22,8 @@
 # DUNE_API_KEY=
 
 # Matomo environment (URL and site ID required for analytics)
-NEXT_PUBLIC_MATOMO_URL=
-NEXT_PUBLIC_MATOMO_SITE_ID=
+# NEXT_PUBLIC_MATOMO_URL=
+# NEXT_PUBLIC_MATOMO_SITE_ID=
 
 # Used to avoid loading Matomo in our preview deploys
 IS_PREVIEW_DEPLOY=false
@@ -31,11 +31,11 @@ IS_PREVIEW_DEPLOY=false
 # Build pages only for the specified langs. Leave it empty to build all the langs
 # e.g. `en,fr` will only build English and French pages
 # Note: always include `en` as it is the default lang of the site
-NEXT_PUBLIC_BUILD_LOCALES=
+NEXT_PUBLIC_BUILD_LOCALES=en,es
 
 # If resource constraints are being hit during builds, change LIMIT_CPUS to a
 # fixed number of CPUs (e.g. 2) to limit the demand during build time
-LIMIT_CPUS=
+# LIMIT_CPUS=
 
 # Enables the bundle analyzer
 ANALYZE=false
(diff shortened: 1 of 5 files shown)


wackerow:
Hey @juanchax! Appreciate the PR... I'm not sure this is really the approach we want to take here. It seems to add quite a bit of complexity?

Was the main point here so the URL's would stay consistent when used in other languages? cc: @pettinarip Curious your take here, but I'm concerned this would create more challenges than it would solve.

juanchax:
> Hey @juanchax! Appreciate the PR... I'm not sure this is really the approach we want to take here. It seems to add quite a bit of complexity?
> 
> Was the main point here so the URL's would stay consistent when used in other languages? cc: @pettinarip Curious your take here, but I'm concerned this would create more challenges than it would solve.

Yeah, ok, a little background info here.  
I noticed there are quite a few outdated links in translated pages, on top of the "bugs" related to dead links on several pages & translations that had already been reported.
Did some digging and found that  the urls for the same link on a page might differ from one language to another, as in, they are _hardcoded_ on pages. 

So the overall refactor proposed might seem like a lot when looking at the effort I'd need to put in, but compared to having to regularly fish for dead + outdated links on _all pages_ + _all translation pages_ and go in and fix all instances "one-by-one", it's not that heavy a taask.

- Moving away from hardcoded urls to url references, 
- keeping a config file as the one source of truth regarding urls for inks on pages, 
- still regularly check the configuration file for dead links *but* 
- update those references in the configuration file _only_. 

Then, all pages across all languages will be automagically picking up the latest urls. If that makes sense(?)

PS. I'm offering to do the heavy lifting here, I just wanted to run it by you guys before going all in; specially to get some feedback and also to learn about possible planned refactors that would eliminate the need for this proposed PR.

corwintines:
Personally I don't think this is worth it. The maintenance of these URL's will be a lot, and the added complexity of it doesnt seem worth it to me. 

I do think we could do something about processing the links on these pages and changing them to what is on the english version of the page. But, a whole config file for this seems unnecessary.

corwintines:
Plus, with AI translation flows we are looking into, this should be handled easier that way too.

corwintines:
Going to close this solution out for now, I think we need a more simple solution than a config for this given the scale of our markdown files.

PR #15324, opened 2025-04-21
feat(pages/resources): add pectra countdown and blobscan api

### LINKED ISSUE #15219: Add Pectra countdown and API integration to the dashboards page
the[ dashboard page](https://ethereum.org/en/resources/) was realised as a PoC, and we have planned a series of improvements.

This one is to add the Upgrade to the scaling section.

here is  the figma > https://www.figma.com/design/hjuimvUsUMgpIL7Nr41evG/Dashboard?node-id=794-116189&t=muMYtca9n1lGnNgm-1

Image

TylerAPfledderer:
## Related Issue

Closes #15219

### DIFF
--- a/src/components/Resources/useResources.tsx
+++ b/src/components/Resources/useResources.tsx
@@ -58,7 +58,11 @@ const formatSmallUSD = (value: number, locale: string): string =>
     maximumSignificantDigits: 2,
   }).format(value)
 
-export const useResources = ({ txCostsMedianUsd }): DashboardSection[] => {
+export const useResources = ({
+  txCostsMedianUsd,
+  txFeesSaved,
+  totalBlobs,
+}): DashboardSection[] => {
   const { t } = useTranslation("page-resources")
   const locale = useLocale()
   const localeForNumberFormat = getLocaleForNumberFormat(locale! as Lang)
@@ -355,7 +359,22 @@ export const useResources = ({ txCostsMedianUsd }): DashboardSection[] => {
     },
     {
       title: t("page-resources-blobs-title"),
-      // TODO: Add metric
+      metric: (
+        <div className="flex [&>*]:grid [&>*]:flex-1 [&>*]:place-items-center">
+          <div>
+            <div className="text-[42px] font-bold leading-2xs">
+              {totalBlobs}
+            </div>
+            <div className="text-sm text-body-medium">Total blobs</div>
+          </div>
+          <div>
+            <div className="text-[42px] font-bold leading-2xs">
+              {txFeesSaved}
+            </div>
+            <div className="text-sm text-body-medium">Total Tx Fees Saved</div>
+          </div>
+        </div>
+      ),
       items: [
         {
           title: "Blob Scan",
--- a/app/[locale]/resources/_components/resources.tsx
+++ b/app/[locale]/resources/_components/resources.tsx
@@ -26,13 +26,23 @@ import { useTranslation } from "@/hooks/useTranslation"
 import heroImg from "@/public/images/heroes/guides-hub-hero.jpg"
 interface ResourcesPageProps {
   txCostsMedianUsd: MetricReturnData
+  txFeesSaved: string
+  totalBlobs: string
 }
 
 const EVENT_CATEGORY = "dashboard"
 
-const ResourcesPage = ({ txCostsMedianUsd }: ResourcesPageProps) => {
+const ResourcesPage = ({
+  txCostsMedianUsd,
+  txFeesSaved,
+  totalBlobs,
+}: ResourcesPageProps) => {
   const { t } = useTranslation("page-resources")
-  const resourceSections = useResources({ txCostsMedianUsd })
+  const resourceSections = useResources({
+    txCostsMedianUsd,
+    txFeesSaved,
+    totalBlobs,
+  })
   const activeSection = useActiveHash(
     resourceSections.map(({ key }) => key),
     "0% 0% -70% 0%"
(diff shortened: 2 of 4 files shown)


src/components/Resources/useResources.tsx
```diff
@@ -58,11 +59,22 @@ const formatSmallUSD = (value: number, locale: string): string =>
...
   const locale = useLocale()
   const localeForNumberFormat = getLocaleForNumberFormat(locale! as Lang)
 
+  const ethPrice = useEthPrice()
+  const avgBlobFeeUsd = formatSmallUSD(
+    // Converting value from gwei to USD
+    avgBlobFee * 1e-9 * ethPrice,
+    localeForNumberFormat
+  ).replace(/[A-Za-z]$/, "")
+
```
TylerAPfledderer, line 77: I double-checked in a converter, and this value should be correct, returning `$2.4` here.

src/components/Resources/useResources.tsx
```diff
@@ -343,7 +355,14 @@ export const useResources = ({ txCostsMedianUsd }): DashboardSection[] => {
   const scalingBoxes: DashboardBox[] = [
     {
       title: t("page-resources-roadmap-title"),
-      // TODO: Add metric
+      metric: (
+        <div className="grid place-items-center py-5">
+          <div className="text-sm">Next upgrade</div>
+          <div className="text-5xl font-bold">Pectra</div>
+          {/* TODO: Convert date to a countdown */}
+          <div className="text-xl font-bold text-body-medium">07 May 2025</div>
```
TylerAPfledderer, line 363: Any suggestions on which way to go here? Kinda already doing an approach in `useResources` for "Time to next block". Use that approach for this?

There is also something more robust like the [react-countdown](https://github.com/ndresx/react-countdown) component.

TylerAPfledderer (commented):
Comments for @pettinarip and @wackerow here. The formatted values of the two numbers don't seem right to me. 😅  I had to remove the "billion" abbreviation from the formatted "Average blob fee".

src/components/Resources/useResources.tsx
pettinarip: Hey @TylerAPfledderer I think that the problem is that `avgBlobFee` is in wei, not gwei. That is why we get it in billions. So, if we do

```suggestion
  const ethPrice = useEthPrice()
  const avgBlobFeeUsd = formatSmallUSD(
    // Converting value from wei to USD
    avgBlobFee * 1e-18 * ethPrice,
    localeForNumberFormat
  )
```

we shouldn't need to replace anything.

TylerAPfledderer: @pettinarip ah, that did it! Pushed the change 😄

nloureiro:
@TylerAPfledderer did manage to see why the page doesn't load?

https://deploy-preview-15324--ethereumorg.netlify.app/en/resources/

TylerAPfledderer:
> @TylerAPfledderer, did you manage to see why the page doesn't load?
> 
> https://deploy-preview-15324--ethereumorg.netlify.app/en/resources/
> 
> 
> 

@nloureiro 

I'm not able to check this. Everything works fine in local dev, and when attempting to build locally, I have issues getting the build to complete successfully for unrelated reasons.

I can try that again today

nloureiro:
> > @TylerAPfledderer, did you manage to see why the page doesn't load?
> > https://deploy-preview-15324--ethereumorg.netlify.app/en/resources/
> 
> @nloureiro
> 
> I'm not able to check this. Everything works fine in local dev, and when attempting to build locally, I have issues getting the build to complete successfully for unrelated reasons.
> 
> I can try that again today

nice, thank you!

TylerAPfledderer:
@nloureiro apologies for the delay. I am still not able to recreate the problem locally. When I attempt to run the build locally, I still get unrelated errors that prevent the build from completing.

@pettinarip @wackerow any input here with the resource not working through the deploy link?

src/data/mocks/blobscanOverallStats.json
pettinarip, line 1: added this file. If you run with `USE_MOCK_DATA=true` this will be used

src/components/Resources/useResources.tsx
```diff
@@ -53,16 +54,24 @@ const formatSmallUSD = (value: number, locale: string): string =>
   new Intl.NumberFormat(locale, {
     style: "currency",
     currency: "USD",
-    notation: "compact",
-    minimumSignificantDigits: 2,
-    maximumSignificantDigits: 2,
```
TylerAPfledderer, line 58: Simplify the price formatting, so that two decimal places are always shown, no matter the number of significant figures. (i.e. `$xx.xx` versus `$x.xx`)

nloureiro (approved):

wackerow (approved):
Thanks a lot @TylerAPfledderer! Sorry for delays here, pulling in

wackerow:
Ah spoke too soon, forgot this got refactored to server-side without hooks. Will need to refactor this a bit

wackerow:
Hey folks! Made some updates here:
- Refactored to newer server-side page approach, extracting only what was needed as a client component
- Cleaned up fetching logic/structs for blobstats
- Adjust logic to fetch latest available upgrade from `src/data/networkUpgradeSummaryData.ts` based on latest available `dateTimeAsString` value
- Added/utilized `humanize-duration` package for human readable durations
- Extract i18n strings; utilize i18n support with humanize-duration where available (fallback to English when language not available.

Example screenshots

image
image
image

Going to pull this in and we can iterate from here as needed

PR #15327, opened 2025-04-22
fix: DocLink routing; use internal link component

microHoffman:
## Description

Currently when clicking on internal link in Learn page ( https://ethereum.org/en/learn/ ), it ended up in 404 redirect, because of the missing i18n (links are pointed to `https://ethereum.org/quizzes/` and not to `https://ethereum.org/[locale]/quizzes/`m which resulted in 404 page not found. 

I mean links in these sections:
image

I've fixed it by using the `InlineLink` component in `DocLink` component, which handles the addition of i18n locale to links automatically.

### DIFF
--- a/src/components/DocLink/index.tsx
+++ b/src/components/DocLink/index.tsx
@@ -4,6 +4,7 @@ import { cn } from "@/lib/utils/cn"
 
 import Emoji from "../Emoji"
 import { Center, Flex, Stack } from "../ui/flex"
+import InlineLink from "../ui/Link"
 import { LinkBox, LinkOverlay } from "../ui/link-box"
 
 import { useRtlFlip } from "@/hooks/useRtlFlip"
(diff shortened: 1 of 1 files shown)


taxmeifyoucan:
This should fix #15337 as well. Bunch of internal and external links are broken, can we please get this resolved @wackerow

microHoffman:
> This should fix #15337 as well. Bunch of internal and external links are broken, can we please get this resolved @wackerow

Maybe it will still happen on a few other places as there are multiple components being used for links across the app, but at least for those in the Learn section boxes it should fix it:)

wackerow (approved):
Agree this is a bug that needed to be fixed, since these were not using the appropriate NextLink (which `InlineLink` uses under-the-hood) to properly handle `/locale/` and internal page routing. 

**This does not fix #15337** which is an unrelated, and more global problem where direct links to say https://ethereum.org/learn are not being automatically forwarded to the default locale, ie https://ethereum.org/en/learn

Pulling this in (Thanks @microHoffman!) but keeping that issue open

PR #15339, opened 2025-04-25
Corrected Priority Fee section to reflect validator incentives

nikitakent:
## Summary 

Clarified language to reflect there is **no penalty** for including zero transactions: staking rewards are independent of transaction inclusion. 

## Description

- Validators still have indirect economic incentive to include transactions in block due to the deflationary effects of burning base fees, thus the sentence:"Without tips, validators would _find it economically viable_ to mine empty blocks, as they would receive the _same block reward_." was not completely accurate. 

- - Changes include more specific language to reflect validators do not receive any direct loss or penalty from consensus/execution-layer rewards.

- Block rewards also outdated terminology.

### DIFF
--- a/public/content/developers/docs/gas/index.md
+++ b/public/content/developers/docs/gas/index.md
@@ -81,7 +81,7 @@ It's also important to note it is unlikely we will see extended spikes of full b
 
 ### Priority fee (tips) {#priority-fee}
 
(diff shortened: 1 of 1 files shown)


shoenseiwaso:
Still relevant

wackerow:
@all-contributors please add @nikitakent for content update

PR #15341, opened 2025-04-25
doc: update how to use a bridge with some improvement on the doc and …

CodeLeom:
…also add reference to a glossary definition

## Description
Adjust sentence, and also adding reference to a glossary definition

### DIFF
--- a/public/content/guides/how-to-use-a-bridge/index.md
+++ b/public/content/guides/how-to-use-a-bridge/index.md
@@ -10,7 +10,7 @@ If there is a lot of traffic on Ethereum, it can become expensive. One solution 
 
 **Prerequisite:** 
 
-- have a crypto wallet; If you don't, follow this guide to [create an Ethereum account](/guides/how-to-create-an-ethereum-account/)
+- have a crypto wallet—if you don't, follow this guide to [create an Ethereum account](/guides/how-to-create-an-ethereum-account/)
 - add funds to your wallet
 
 ## 1. Determine which layer 2 network you want to use


public/content/guides/how-to-use-a-bridge/index.md
```diff
@@ -6,11 +6,11 @@ lang: en
...
+If there is a lot of traffic on Ethereum, it can become expensive. One solution to this is to create new "layers": i.e. different networks which operate in similar ways to Ethereum itself. This so-called Layer 2s helps reduce congestion and cost on Ethereum by processing many more transactions at lower fees, and only storing the result of these on Ethereum every so often. As such, these layers 2s enable us to transact with increased speed and decreased costs. Many popular crypto projects are moving to layer 2s because of these benefits. The simplest way to move tokens from Ethereum to layer 2 is to use a bridge.
```
wackerow: Existing is appropriate

```suggestion
If there is a lot of traffic on Ethereum, it can become expensive. One solution to this is to create new "layers": i.e. different networks which operate in similar ways to Ethereum itself. These so-called Layer 2s help reduce congestion and cost on Ethereum by processing many more transactions at lower fees, and only storing the result of these on Ethereum every so often. As such, these layers 2s enable us to transact with increased speed and decreased costs. Many popular crypto projects are moving to layer 2s because of these benefits. The simplest way to move tokens from Ethereum to layer 2 is to use a bridge.
```

public/content/guides/how-to-use-a-bridge/index.md
```diff
@@ -6,11 +6,11 @@ lang: en
...
 
 **Prerequisite:** 
 
-- have a crypto wallet, you can follow this tutorial: [How to create an Ethereum account](/guides/how-to-create-an-ethereum-account/)
+- have a crypto wallet; If you don't, follow this guide to [create an Ethereum account](/guides/how-to-create-an-ethereum-account/)
```
wackerow:
```suggestion
- have a crypto wallet—if you don't, follow this guide to [create an Ethereum account](/guides/how-to-create-an-ethereum-account/)
```

wackerow:
@all-contributors please add @CodeLeom for content patch

Issue #15402, opened 2025-05-05
Add ETHDimsum on Ethereum meetups [/community/events/#meetups]


isabelladebrito:
### Context
Add ETHDimsum to ethereum.org's Ethereum meetups list (https://ethereum.org/en/community/events/#meetups)

### Meetup to be added
ETHDimsum
🌏  Guangdong, Hong Kong, Macao
URL: https://lu.ma/user/usr-Irp0cJhplYWGoK4

### Would you like to work on this issue?

- [ ] Yes
- [X] No

ArunRawat404:
Hello, I'd love to work on this issue.

Deepakraja03:
I would like to contribute to this issue

wackerow:
See comment on #15411

PR #15411, opened 2025-05-07
Add ETHDimsum on Ethereum meetups 

ArunRawat404:
## Description
Added ETHDimsum meetup details on Ethereum [meetups list](https://ethereum.org/en/community/events/#meetups)

## Related Issue
#15402

### DIFF
--- a/src/data/community-meetups.json
+++ b/src/data/community-meetups.json
@@ -574,5 +574,11 @@
     "emoji": ":belgium:",
     "location": "Belgium",
     "link": "https://lu.ma/ethbelgium"
-  }   
+  },
+  {
+    "title": "ETHDimsum",
+    "emoji": ":earth_asia:",
+    "location": "Guangdong, Hong Kong, Macao",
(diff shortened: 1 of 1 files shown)


ArunRawat404:
Hi @wackerow, I think it is a recurrent meetup. The description of the meetup [ETHDimsum](https://luma.com/user/usr-Irp0cJhplYWGoK4) says so **Ethereum Meetup Series.** The issue also says to add to meetup.

wackerow:
@ArunRawat404 Yes, I'm aware of what the issue states, but the link provided (https://lu.ma/user/usr-Irp0cJhplYWGoK4) shows one single event from five months ago:

image

And the twitter profile similarly hasn't been posted to in months. If I were in Guangdong and I went to this link I would have no direction on where or when the next event would be, which is poor UX for our users. Closing this out for now, but please feel free to provide a link with reference to a recurrent event that visitors can attend.

PR #15439, opened 2025-05-14
Fix go to top [Fixes #15412]

### LINKED ISSUE #15412: go to top button npt working
### Describe the bug

the go to top button is not doing anything
Image

### To reproduce

1. Go to the footer
2. Click on Go to top
3. Nothing happens

### Expected behavior

is should scroll to the top of the page

### Screenshots

_No response_

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes
- [ ] No

smithrashell:
## Description

The "Go to top" button was not functioning due to a missing anchor target.

This PR:
- Adds an `id="top"` to the `main` container in `BaseLayout.tsx`
- Updates the `href` used by the "Go to top" button in `Footer.tsx` to point to `#top`

## Related Issue
Fixes [#15412](https://github.com/ethereum/ethereum-org-website/issues/15412)

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -320,7 +320,7 @@ const Footer = ({ lastDeployLocaleTimestamp }: FooterProps) => {
         <Button
           variant="outline"
           isSecondary
-          onClick={() => scrollIntoView("div.mx-auto")}
+          onClick={() => scrollIntoView("body")}
         >
           <IoChevronUpSharp /> Go to top
         </Button>
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -320,7 +320,7 @@ const Footer = ({ lastDeployLocaleTimestamp }: FooterProps) => {
         <Button
           variant="outline"
           isSecondary
-          onClick={() => scrollIntoView("top")}
+          onClick={() => scrollIntoView("div.mx-auto")}
         >
           <IoChevronUpSharp /> Go to top
         </Button>
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -320,7 +320,7 @@ const Footer = ({ lastDeployLocaleTimestamp }: FooterProps) => {
         <Button
           variant="outline"
           isSecondary
-          onClick={() => scrollIntoView("__next")}
+          onClick={() => scrollIntoView("top")}
         >
           <IoChevronUpSharp /> Go to top
         </Button>
--- a/src/layouts/BaseLayout.tsx
+++ b/src/layouts/BaseLayout.tsx
@@ -51,7 +51,7 @@ export const BaseLayout = ({
        * layout on initial load.
        */}
       <SkipLink />
-      <div className="mx-auto max-w-screen-2xl" id="top">
+      <div className="mx-auto max-w-screen-2xl">
         <Nav />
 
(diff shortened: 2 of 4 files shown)


src/layouts/BaseLayout.tsx
```diff
@@ -51,7 +51,7 @@ export const BaseLayout = ({
        * layout on initial load.
        */}
       <SkipLink />
-      <div className="mx-auto max-w-screen-2xl">
+      <div className="mx-auto max-w-screen-2xl" id="top">
```
wackerow, line 54: Want to be a little careful here since this same `id` is already being used on some pages, such as the Docs pages, and this would result in them being duplicated which isn't valid HTML.. I think we can do this another way.

src/components/Footer.tsx
```diff
@@ -320,7 +320,7 @@ const Footer = ({ lastDeployLocaleTimestamp }: FooterProps) => {
         <Button
           variant="outline"
           isSecondary
-          onClick={() => scrollIntoView("__next")}
+          onClick={() => scrollIntoView("top")}
```
wackerow, line 323: This function currently accepts an `id` (`toId`), but we could broaden it's usage by refactoring to use a css selector string with `querySelector`

I would just update `scrollIntoView` to accept a query selector instead of an `id`, and then we can just target the `h1` in this case, and update any other instances passing an id to prefix a `#`

For example: 
```ts
// src/lib/utils/scrollIntoView.ts

export const scrollIntoView = (
  selector: string,
  options: ScrollIntoViewOptions = { behavior: "smooth", block: "start" }
): void => {
  const element = document.querySelector(selector)

  if (!element) return

  element.scrollIntoView(options)
}
```

Then this line becomes

```suggestion
          onClick={() => scrollIntoView("h1")}
```

And I believe the only other place we use this is in `Button`, where we can prefix the id with `#`:

```tsx
// src/components/ui/buttons/Button.tsx:108

      toId && scrollIntoView("#" + toId)
```

wackerow (changes requested):
Hey @smithrashell! Thanks for the fix! This does work, but I think we can take it another step to make sure we don't break any html rules, and simplify the need to add id's. 

Let me know if you have issues!

src/components/Footer.tsx
smithrashell: Hi @wackerow , Thanks for the clear feedback! I've updated the implementation as suggested and just pushed the changes. Let me know if anything else needs adjustment.

src/components/Footer.tsx
```diff
@@ -320,7 +320,7 @@ const Footer = ({ lastDeployLocaleTimestamp }: FooterProps) => {
         <Button
           variant="outline"
           isSecondary
-          onClick={() => scrollIntoView("__next")}
+          onClick={() => scrollIntoView("div.mx-auto")}
```
wackerow, line 323: Ah, I see that `h1` by default doesn't really position things well.. I'm just going to switch this to use `body` here instead then to avoid any weirdness with this selector in the future.

```suggestion
          onClick={() => scrollIntoView("body")}
```

wackerow (approved):
This looks great! Thanks so much @smithrashell... working great for me, pulling in!

wackerow:
@all-contributors please add @smithrashell for bug fix

PR #15449, opened 2025-05-15
remove innacurated content from the roadmap page

nloureiro:
The homepage has a section with What technical upgrades are coming to Ethereum? and based on feedback, we should remove inaccurate boxes.

Maybe add new ones in the future

How is it in production
Screen Shot 2025-05-15 11 56 41 AM

How does it look with this PR
Screen Shot 2025-05-15 11 58 10 AM


wackerow (approved):
Thanks @nloureiro!

PR #15451, opened 2025-05-15
Fix h fonts

wackerow:
## Description
- fix: layout component set logic... Bug existed causing last set of custom components spread into an object to override everything else, causing the markdown components for the "tutorial" layout to be dominant.
- feat: update h2 styling; removes `mono` and `uppercase` font properties
- refactor: use tw classes for default bg and color in Codeblock component (semi-patches a small color-mode bug I'm seeing with the update to the components)


nloureiro:
@wackerow looks great ship it!!!

Screen Shot 2025-05-16 10 03 26 AM

pettinarip (approved):
Nicee! LGTM

PR #15471, opened 2025-05-19
fix: Fix accordion label inconsistency in stablecoins page

### LINKED ISSUE #15398: Bug report: stablecoins page react state management related
### Describe the bug

On the stablecoins page, under How to get stablecoins section, if Swap is expanded, it correctly shows the label as Less. If Buy is expanded after that, Swap is correctly minimized but the label remains as Less instead of switching back to More. So, it creates a scenario where multiple subsections are labeled as Less even if only one is expanded. After that, if any Less labeled subsection is expanded again, it shows the label as More. Thus, completely inconsistent states.

### To reproduce

1. Go to 'https://ethereum.org/en/stablecoins/'
2. Scroll down to 'How to get stablecoins'
3. Click on Swap
4. Click on Buy

### Expected behavior

When a user clicks Buy after clicking Swap, the label of Buy section should be switched to More since it's no longer expanded. There should be maximum one Less labeled section at a time.

### Screenshots

Image

_No response_

_No response_

### Additional context

_No response_

- [x] Yes
- [ ] No

chainide-agent-bob:
# Fix accordion label inconsistency in stablecoins page

## Description
This PR fixes a state management issue in the StablecoinAccordion component where the "More/Less" labels weren't correctly updating when switching between accordion sections.

## Problem
When a user clicks on an accordion section (like "Swap") and then clicks on another section (like "Buy"), the first section would collapse correctly but its label would remain as "Less" instead of switching back to "More". This created an inconsistent UI state where multiple collapsed sections could show "Less" labels simultaneously.

## Solution
The solution centralizes the state management by:
1. Adding a central `openItem` state in the parent `StablecoinAccordion` component to track which accordion item is currently open
2. Modifying the `AccordionCustomItem` component to accept `isOpen` and `onOpenChange` props instead of managing state internally
3. Ensuring that the "More/Less" label is correctly tied to the actual expanded/collapsed state of each section

This ensures that at most one accordion section can be expanded at any time, and the labels will always accurately reflect the current state of each section.

## Related issue
- Fixes #15398

### DIFF
--- a/src/components/StablecoinAccordion/AccordionCustomItem.tsx
+++ b/src/components/StablecoinAccordion/AccordionCustomItem.tsx
@@ -1,4 +1,4 @@
-import { BaseHTMLAttributes, ReactNode, useState } from "react"
+import { BaseHTMLAttributes, ReactNode } from "react"
 
 import type { ChildOnlyProp } from "@/lib/types"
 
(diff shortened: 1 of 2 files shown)


wackerow (changes requested):
Hey @chainide-agent-bob, appreciate the help here.. Couple things:

This seems to add a fair amount of complexity to simply solve for using the "more"/"less" labels as a custom change to our canonical Accordion design system component, which uses the chevron arrow to indicate opening/closing instead.

To simplify, and to align with existing Accordion component usage, I would suggest we do a couple things instead:

1. Remove `hideIcon` prop from the `AccordionTrigger` component inside `src/components/StablecoinAccordion/AccordionCustomItem.tsx`
2. Remove `<MoreOrLessLink isOpen={open} />` from the same file
3. Remove the `md:px-0` class from the `AccordionTrigger` component (redundant), and then update `px-0` to use `ps-0` instead (keeping the original padding at the end)
4. Revert everything else here

These simple changes will fix the logic, simplify greatly, and align with existing design system.

wackerow:
Appreciate the help here—went ahead and made the above fixes in a separate PR (#16289), going to close this one out

PR #15476, opened 2025-05-19
RSS feed error handling

wackerow:
## Description
- Adds try/catch blocks to RSS feed fetching
- If feed not found, or invalid type, throws error up to parent (`fetchRSS`)
- Parent logs any errors thrown, but continues through list without breaking
- Final check added to make sure fetches resulted in at least 6 (`RSS_DISPLAY_COUNT`) items, else it throws a breaking error. This allows flexibility for small outages or link changes, but will block anything larger that results in missing items in the front end.

## Related Issue
Recent homepage outage from an external rss feed url being changed


src/lib/api/fetchRSS.ts
```diff
@@ -18,108 +13,123 @@ export const fetchRSS = async (xmlUrl: string | string[]) => {
...
+      }
+    } catch (error) {
+      console.error(error instanceof Error ? error.message : error)
+      continue
     }
   }
-  return allItems as RSSItem[][]
+
+  if (allItems.length < RSS_DISPLAY_COUNT)
+    throw new Error("Insufficient number of RSS items fetched")
```
wackerow, line 130: This isn't actually breaking the build as-is, since it is contained inside an api function call... 🤔  cc: @pettinarip

src/lib/api/fetchRSS.ts
```diff
@@ -18,108 +12,119 @@ export const fetchRSS = async (xmlUrl: string | string[]) => {
...
+          `Error parsing XML, invalid RSSResult or AtomResult type: ${url}`
+        )
+      }
+    } catch (error) {
+      console.error(error instanceof Error ? error.message : error)
+      continue
     }
   }
-  return allItems as RSSItem[][]
+  return allItems
```
pettinarip, line 127: this function fetch will never throw, is that the intention? if we want to break the build, we should

pettinarip (changes requested):
By design, the fetches under `src/lib/api` should throw. The error handling logic should happen at the page level. Currently, the `dataLoader` layer is capturing these errors and throwing.

The reason this test is not failing is because the flag `USE_MOCK_DATA` is on in preview deploys. To fully test it, you can override the variable in netlify.toml or in netlify UI config.

wackerow:
> By design, the fetches under `src/lib/api` should throw. The error handling logic should happen at the page level. Currently, the `dataLoader` layer is capturing these errors and throwing.
> 
> The reason this test is not failing is because the flag `USE_MOCK_DATA` is on in preview deploys. To fully test it, you can override the variable in netlify.toml or in netlify UI config.

Couple things:

1- Netlify: `USE_MOCK_DATA` is already set to `false` for this branch

image

But, I didn't test this issue specifically through Netlify (truncating the list of RSS feeds so we don't get an adequate number of responses, which imo should break the build....

2- Locally: I did test locally though, and it builds just fine without breaking when I only have a single RSS feed source listed (which should break it)... `USE_MOCK_DATA=false`... No errors throw at build, but when I serve that build the homepage breaks:

image

image

image

image

image

src/lib/api/fetchRSS.ts
wackerow: I don't want a singular feed to break the build (ie. recent example where one blog changed their feed link)... these cases should just skip over that feed.

BUT, in the event something happens where we don't get ANY feed results, then that should break the build... which was the intent of throwing an error at the end if there aren't enough results.

wackerow (commented):
This should be good to go cc: @pettinarip @corwintines

pettinarip (approved):
LGTM!

PR #15485, opened 2025-05-20
fix hero regression on find wallet

corwintines:
## Description

- Fixes regression for find wallet hero for new design


wackerow (approved):

wackerow:
Just noting separately we should replace the disclaimer banner with something else

Afeter8 (approved):

Issue #15489, opened 2025-05-21
Bug report: wrong line-height in cards


konopkja:
### Describe the bug

The text should have 1 rem space

<img width="320" alt="Image" src="https://github.com/user-attachments/assets/afcdd102-f3a7-4ee3-b042-b3409288ee61" />

### To reproduce

1. Go to https://ethereum.org/de/staking/saas/
2. Scroll down to Cards section
3. See error

### Expected behavior

1 rem 

### Would you like to work on this issue?

- [ ] Yes
- [x] No

Anish-Gupta1:
I would love to take this as well :)

konopkja:
Thank you so much!

Anish-Gupta1:
<img width="601" alt="Image" src="https://github.com/user-attachments/assets/44aacd96-4a06-421e-b620-54dae81ae471" />

is this okay ? let me know if this is good @konopkja

konopkja:
please when making PR, link corresponding issue to it

PR #15496, opened 2025-05-22
Fix incorrect line-height in card text

Anish-Gupta1:
## Description

Replaced leading-3 with leading-[1rem] to match the expected line-height

## Related Issue
Fixes: #15489


src/components/Staking/StakingProductsCardGrid/StakingProductCard.tsx
```diff
@@ -191,7 +191,7 @@ export const StakingProductCard = ({
           {data.map(({ label, status }, idx) => (
             <li
               key={idx}
-              className={`my-4 me-0 ms-auto flex items-center gap-1 text-md leading-3 ${status === "false" && "text-body-medium"}`}
+              className={`my-4 me-0 ms-auto flex items-center gap-1 text-md leading-[1rem] ${status === "false" && "text-body-medium"}`}
```
wackerow: A fun shortcut with tailwind, you can append a `text-` sizing class with `/line-height-class`. We're using `text-md` here, which should really be `text-base`, which is 16px. You're assigning 1rem, which is a line height of 100%, or 1. This can be accomplished with tailwind class `leading-none`.

So to consolidate, you can do all this with `text-base/none`

```suggestion
              className={`my-4 me-0 ms-auto flex items-center gap-1 text-base/none ${status === "false" && "text-body-medium"}`}
```

Just thought I'd share! Going to commit that and pull in, thanks @Anish-Gupta1!

wackerow (approved):

PR #15546, opened 2025-05-30
feat: upgrade package manager to pnpm

wackerow:
## Description

- **Primary update: Implements `pnpm` as canonical package manager**. Dependency requirements are a bit different with `pnpm` compared to `yarn`, and all package versions have been updated accordingly to comply. 
- Enables `corepack` for package manager version control; version maintained in `package.json` (Enable with `corepack enable`)
- Updates README and other docs, removing all instructions that utilized `yarn` and updated to `pnpm` instructions
- Updated any workflow package manager usage


src/lib/constants.ts
```diff
@@ -62,16 +60,6 @@ export const CROWDIN_PROJECT_URL = "https://crowdin.com/project/ethereum-org"
 export const CROWDIN_PROJECT_ID = 363359
 export const CROWDIN_API_MAX_LIMIT = 500
 export const FIRST_CROWDIN_CONTRIBUTION_DATE = "2019-07-01T00:00:00+00:00"
-export const REGULAR_RATES: ReportsModel.RegularRate[] = [
```
pettinarip, line 65: not used? 👍🏼

src/scripts/crowdin/reports/reportsHelpers.ts
```diff
@@ -25,22 +24,45 @@ export async function fetchTranslationCostsReport(
 
   // Todo: Remove ts-ignore when this PR gets merged
   // https://github.com/crowdin/crowdin-api-client-js/pull/282
-  const schema: ReportsModel.TranslationCostSchema = {
+  // const schema: ReportsModel.TranslationCostSchema = {
+  const schema: ReportsModel.TranslationCostsPostEndingSchema = {
```
pettinarip, line 26: this doesn't seem the PR to include this change. do we need to test something? not sure what this change is doing

src/scripts/crowdin/reports/reportsHelpers.ts
```diff
@@ -25,22 +24,45 @@ export async function fetchTranslationCostsReport(
 
   // Todo: Remove ts-ignore when this PR gets merged
```
pettinarip: time to ✂️ this comment?

pettinarip (approved):
Good job @wackerow! thanks for the migration and lodash normalization :)

Left a comment about something related to crowdin that seems unrelated to the pnpm migration. Could you elaborate more on those changes?

src/scripts/crowdin/reports/reportsHelpers.ts
wackerow: There were some changes to the data schema required from the crowdin package. We can check with @lukassim but I'm not sure we really even use this anymore, but in the meantime I tried my best o update to match the new schema.

src/scripts/crowdin/reports/reportsHelpers.ts
wackerow: Looks like it... yeah I also see the commented code line below that I can patch, thanks!

pettinarip (approved):
LGTM!

PR #15553, opened 2025-05-31
refactor: homepage to SSR

wackerow:
## Description
- Removes nested client-side homepage `page.tsx`; migrates everything to `app/[locale]/page.tsx`
- Converts to `const t = await getTranslations()` usage for server-side intl string parsing
- Moves all nested usages of intl parsing to server-side page, passing formatted results to child components
- Removes inappropriate Suspense boundary usage; implements `is (!mounted) return <Fallback/>` logic inside client-side components where appropriate for loading state
- Removes `I18nProvider` wrapper as it's no longer needed
- With fallback changes, added some more usage of Skeleton, adding a skeleton "card"
- Bumped swiper package, renamed a couple things (ie. `ActivityStats` and `getActivity`)

## Related Issue
Ongoing performance initiatives


src/lib/utils/direction.ts
```diff
@@ -0,0 +1,27 @@
...
+  direction: "ltr" | "rtl"
+}
+
+/**
+ * Determines the direction and transformation for right-to-left (RTL) languages.
+ * @param locale The language locale
+ * @returns An object containing the Tailwind className, RTL flag, and direction.
+ * @example const { twFlipForRtl } = getDirectionData('ar');
+ */
+export const getDirection = (locale: Lang): DirectionData => {
```
pettinarip, line 19: we could update [this hook](https://github.com/ethereum/ethereum-org-website/blob/dev/src/hooks/useRtlFlip.ts#L20) with same logic

src/components/ui/swiper.tsx
```diff
@@ -1,10 +1,12 @@
+"use client"
```
pettinarip, line 1: this is because we are now consuming it from a server component and the underlying Swiper is a client component, no? just trying to confirm the reasoning

src/components/Image/CardImage.tsx
```diff
@@ -0,0 +1,28 @@
+"use client"
```
pettinarip, line 1: why do we need to declare it as a client component? shouldn't be a problem rendering this on the server

src/components/Image/CardImage.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { ComponentProps } from "react"
+
+import EventFallback from "@/public/images/events/event-placeholder.png"
+
+type CardImageProps = ComponentProps<"img">
+
+const CardImage = ({ src, className, ...props }: CardImageProps) => (
+  // eslint-disable-next-line @next/next/no-img-element
+  <img
```
pettinarip, line 11: I think I remember why we needed to use the `img` tag instead of `Image` but could we rename the component to something more explicit like `RemoteImage`. Don't think the term "Card" is related to the current implementation.

app/[locale]/page.tsx
```diff
@@ -121,18 +402,539 @@ const Page = async ({ params }: { params: Promise<{ locale: Lang }> }) => {
...
-    metricResults,
-    rssData: { rssItems, blogLinks },
-  }
-
   return (
-    <I18nProvider locale={locale} messages={messages}>
-      <HomePage {...props} />
-    </I18nProvider>
+    <MainArticle className="flex w-full flex-col items-center" dir={dir}>
+      <BannerNotification shouldShow={locale === "en"}>
```
Comment (pettinarip) [inline app/[locale]/page.tsx]:
```suggestion
      <BannerNotification shouldShow={locale === DEFAULT_LOCALE}>
```

src/components/Homepage/CodeExamples.tsx
pettinarip, line 1: 👍🏼 nice

src/components/Homepage/RecentPostsSwiper/Fallback.tsx
```diff
@@ -0,0 +1,12 @@
+import { Skeleton, SkeletonCardGrid } from "@/components/ui/skeleton"
+
+const RecentPostsSwiperFallback = () => (
+  <div className="flex flex-col items-center gap-4">
+    <SkeletonCardGrid className="mt-4 w-full md:mt-16" />
+    <Skeleton className="h-4 w-20 rounded-full" />
+  </div>
+)
+
+RecentPostsSwiperFallback.displayName = "RecentPostsSwiperFallback"
```
pettinarip: Notice that you are declaring the `displayName` in these components where the component already infer its name form the function name. Is there another reason for using it? seems redundant and unnecessary.

src/components/Homepage/RecentPostsSwiper/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+}
+
+const RecentPostsSwiper = ({
+  rssItems,
+  eventCategory,
+  className,
+}: RecentPostsSwiperProps) => {
+  const mounted = useIsClient()
+
+  if (!mounted) return <RecentPostsSwiperFallback />
```
pettinarip: shouldn't we solve this by doing

```tsx
<Suspense fallback={<RecentPostsSwiperFallback />}>
  <RecentPostsSwiper />
</Suspense>
```
?

src/components/Homepage/ValuesMarquee/index.tsx
```diff
@@ -170,86 +184,73 @@ const ValuesMarquee = () => {
 
   const { direction, isRtl, twFlipForRtl } = useRtlFlip()
 
+  if (!mounted) return <ValuesMarqueeFallback />
```
pettinarip: same question. seems like an antipattern with the current tools

src/components/Homepage/utils.ts
```diff
@@ -92,13 +101,13 @@ const getPosition = (position: number): string =>
     stylesByPosition.xl[position]
   )
 
-export const useBentoBox = () => {
-  const { t } = useTranslation("page-index")
+export const getBentoBoxItems = async (locale: Lang): Promise<BentoItem[]> => {
```
pettinarip, line 104: nice!

pettinarip (commented):
Great job @wackerow 🔥

src/lib/utils/direction.ts
wackerow: Yes! But, that file is a hook, and uses `useLocale` client-side to get the `locale`... for sake of keeping it scoped, there are ~13 other client-side files that I would have to update to pass `locale` as an arg, so my thought was more that we'll progressively switch to using `getDirection` as we migrate those to server components, then eventually we should be able to get rid of that hook.

src/components/ui/swiper.tsx
wackerow: Correct.. I'm trying to be more explicit with "use client" for files that use hooks, etc, so it's very clear.

This one in particular though, it looks like we only use `useTranslations`... we _could maybe_ pass the `locale` as a prop, conver to `async` and use `const t = await getTranslations({ locale, namespace: "common" })`...

I'd suggest we pull that out as a separate task since `ui/swiper` touches about 5 components right now. We could play around and see if we could get that component to be a fully server component.

src/components/Image/CardImage.tsx
wackerow: It's because we're loading external images and using `onError={handler}`.. that requires the client to process, which was the existing logic to use that fallback image if our client-side fetch for an image fails.

src/components/Image/CardImage.tsx
wackerow: Yes.. in fact, I did lol, and then I switched back since this specific version has a specific fallback with the "EventFallback".. ie, a specific fallback image; it's not just a generic "remote image".... though I'm sure I could adjust to solve for both. Will check it out.

src/components/Homepage/RecentPostsSwiper/Fallback.tsx
wackerow: TIL; out of habit... Pattern I picked up from shadcn; I guess it's not necessary if it's the same name?

src/components/Homepage/RecentPostsSwiper/index.tsx
wackerow: Well, as-is that wouldn't actually work, since nothing is being help in suspense. We would have to load the `swiper` components with lazy loading, or use `dynamic` with `loading`....

Need to think on this; not sure if one is more performant than the other... thoughts @pettinarip?

src/components/Homepage/ValuesMarquee/index.tsx
wackerow: Same as above.. we're currently not actually lazy loading / suspending these, but perhaps we could

src/components/Homepage/utils.ts
wackerow: Yeah; this will likely be a recurring pattern as we migrate to server components... replace hooks that were only used for translation with async functions that accept a locale from the parent to use async `await getTranslations({ locale, namespace: ""} )` instead...

Then it's no longer a hook, and just a function that can be run on the server component to populate the intl strings ahead of time

src/lib/utils/direction.ts
pettinarip: ah no! thats fine. Should have made myself clearer, this is what I had in my mind
```tsx
export const useRtlFlip = () => {
  const locale = useLocale()
  return getDirection(locale)
}
```
to reuse the logic

src/components/Homepage/RecentPostsSwiper/index.tsx
pettinarip: well on the performance side, it uses lazy loading which will improves the initial load/render since it doesn't need to render that immediately. It will also move it out of the main bundle and put it in a different chunk.

there are also benefits in terms of cleaner code, since you are implementing a more "pure" function with fewer side effects + the readability is better with suspense.

I think it is an easy win.

src/components/ui/swiper.tsx
pettinarip: All good.

This should be a pure/dumb ui component (since it is under `/components/ui`). So, it shouldn't know about translations. This component should receive the props already translated, ready to be rendered.

src/components/Image/CardImage.tsx
pettinarip: Ah ok, that refactor make sense if we want to make this more generic and reusable (which is ideal). Otherwise, if this will be used just for the events, then we could name it for that `RemoteEventImage`

src/components/Homepage/RecentPostsSwiper/Fallback.tsx
pettinarip: yea, will display the function/component name. Shadcn uses it because it wraps the components with `forwardRef`s, losing the name of the component.

src/components/Homepage/RecentPostsSwiper/index.tsx
wackerow: Ok, I can try again but that did not work when I tried it. My first approach was using Suspense for all of these.

src/components/Homepage/RecentPostsSwiper/Fallback.tsx
wackerow: Awesome that's good to know, thanks @pettinarip! Can patch those up in a separate task to clean things up

pettinarip (commented):
Great job,

Noting some missing translations re:10years 
image


fixing

pettinarip:
> Great job,
> 
> Noting some missing translations re:10years image
> 
> fixing

@wackerow left a PR with a potential solution https://github.com/ethereum/ethereum-org-website/pull/15653

after this I think we are good to go

PR #15584, opened 2025-06-01
fix: Error (500) and Not Found (404) routing logic

wackerow:
## Description
Fixes logic surrounding 500 error and 404 not-found handing. 
- Disabled `dyanamicParams = false` since that meant encountering an unknown path (a dynamic param) would immediately fallback to Next.js 404. Removing enables our own custom 404 handling with notFound()
- Removed any chance of throwing error when getting metadata; wraps in try/catch, if error then "not found" metadata returned to allow routing by the page component.
- Wraps the root level `NotFoundPage` in `LocaleLayout` to enable the English site layout with header and footer elements
- Updates the error page with refresh button

Examples:
- https://deploy-preview-15584--ethereumorg.netlify.app/derp -> /en/derp -> Our English 404
- https://deploy-preview-15584--ethereumorg.netlify.app/en/derp -> Our English 404
- https://deploy-preview-15584--ethereumorg.netlify.app/es/derp -> Our Spanish 404

## Related Issue
Currently all 404's and 500's result in Next.js pages


src/components/NotFoundPage/index.tsx
```diff
@@ -1,23 +1,21 @@
+"use client"
```
pettinarip, line 1: why do we need to convert this to a client component? couldn't we keep the server component?

app/[locale]/page.tsx
```diff
@@ -142,14 +147,26 @@ export async function generateMetadata({
...
-    locale,
-    slug: [""],
-    title: t("page-index-meta-title"),
-    description: t("page-index-meta-description"),
-  })
+    // Return basic metadata for invalid paths
+    return {
+      title: t("page-not-found"),
+      description: t("page-not-found-description"),
+    }
```
Comment (pettinarip) [inline app/[locale]/page.tsx:168]: Do we need to return this metadata because the 404 page can't handle metadata? or why?

Comment (pettinarip) [reply app/[locale]/page.tsx:168]: If the 404 page is a server component, we might be able to handle this in that component, right?

app/[locale]/[...slug]/page.tsx
```diff
@@ -97,27 +98,45 @@ export default async function Page({
...
+    )
+  } catch (error) {
+    // If content directory doesn't exist (e.g., in Netlify serverless environment),
+    // return empty array to allow ISR to handle all routes dynamically
+    console.warn(
+      "Content directory not found, enabling full dynamic routing:",
+      error
+    )
+    return []
+  }
```
Comment (pettinarip) [inline app/[locale]/[...slug]/page.tsx:118]: not sure I follow this logic, why `getPostSlugs("/")` would fail in Netlify? if it fails means that we don't have any page to build statically and we should break the build, no?

Comment (pettinarip) [reply app/[locale]/[...slug]/page.tsx:118]: also, `getPostSlugs` is now internally handling the same error case, no?

pettinarip (commented):
Looks great @wackerow good job! left a few comments to understand the logic with error handlers mainly and see if we can polish them

src/components/NotFoundPage/index.tsx
wackerow: @pettinarip Without this I was unable to gain any access to the translated versions, so `/es/derp` was loading English content. Took a few approaches to try and properly load the locale before the page rendered using server-side, but in the end the initial load of the page was always going to English—Switching this to "use client" fixed it.

app/[locale]/page.tsx
Comment (wackerow) [reply app/[locale]/page.tsx:168]: @pettinarip Throwing errors from the metadata generation step will stop our logic in it's tracks and default to Next.js 500 page. If we want this to handle as a 404, we need to return valid metadata, not throw an error, and then let the page component handle the routing.

app/[locale]/[...slug]/page.tsx
Comment (wackerow) [reply app/[locale]/[...slug]/page.tsx:118]:
```
Jun 4, 01:59:07 PM: 4b95411b ERROR Error: ENOENT: no such file or directory, scandir '/var/task/public/content/'
at async Object.readdir (node:internal/fs/promises:952:18)
at async h (/var/task/.next/server/app/[locale]/[...slug]/page.js:1:826446)
at async D (/var/task/.next/server/app/[locale]/[...slug]/page.js:1:549543)
at async j (/var/task/.next/server/app/[locale]/[...slug]/page.js:1:548863) {
errno: -2,
code: 'ENOENT',
syscall: 'scandir',
path: '/var/task/public/content/',
digest: '930804571'
}
```

From the Netlify logs it seems that something was triggering this ENOENT error where it couldn't read from this directory in Netlify's serverless environment. This Search suggested that any other hidden system files that end up in there from this environment could be causing this to break when our code attempts to parse... 
Wasn't 100% clear, but this step did seem to again prevent us from being able to properly access our internal 404 route, and wrapping in try/catch resolved here, while all desired pages are still loading correctly.

Can play around a bit more to see if perhaps the error handling inside `getPostSlugs` would be enough for this; that's one thing I hadn't tried yet, but that was the reasoning for this approach.

PR #15591, opened 2025-06-02
Update community-meetups.json - beerfi adjusted name and link

radeksvarz:
we just updated the name and link slightly


wackerow (approved):
Thanks for the polish @radeksvarz!

PR #15592, opened 2025-06-02
[hot-fix] dynamic paths true, and change l2beat response

corwintines:
Sets dynamic paths to true for l2 pages, and fixes l2beat response to parse properly


app/[locale]/layer-2/page.tsx
```diff
@@ -86,7 +85,7 @@ export async function generateStaticParams() {
   }))
 }
 
-export const dynamicParams = false
+export const dynamicParams = true
```
Comment (pettinarip) [inline app/[locale]/layer-2/page.tsx]: default is true

```suggestion
```

PR #15593, opened 2025-06-02
refactor: Morpher as reusable component

wackerow:
## Description
Refactors `Morpher` into a reusable component
- Accepts `words` list and optional character set (for transition animation) as props 
- Updates logic to use the first word in list; fixes bug where first word was never rendering again
- Refactor component to only return the text, un-styled. Existing homepage language morpher button moved into separate component.

### DIFF
--- a/app/[locale]/10years/_components/TenYearHero.tsx
+++ b/app/[locale]/10years/_components/TenYearHero.tsx
@@ -44,7 +44,7 @@ const TenYearHero = () => (
           className="select-none opacity-0 max-md:hidden"
           data-label="space-holder"
         >
-          {words[0]}
+          {WORDS[0]}
         </span>
         <span className="text-3xl font-bold text-accent-b md:absolute md:start-0 md:text-nowrap">
           <Morpher words={WORDS} charSet="abcdfgijklnopqsvwxyz" />


src/components/Homepage/LanguageMorpher.tsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+  const handleDesktopClick = () => {
+    ;(
+      document.querySelector(
+        `button[name="${DESKTOP_LANGUAGE_BUTTON_NAME}"`
+      ) as HTMLButtonElement
+    ).click()
+  }
+
+  const [isLarge] = useMediaQuery(["(min-width: 48rem)"]) // TW md breakpoint, 768px
```
pettinarip, line 35: FYI we could pull this value from https://github.com/ethereum/ethereum-org-website/blob/dev/src/lib/utils/screen.ts

src/components/Homepage/LanguageMorpher.tsx
```diff
@@ -0,0 +1,72 @@
...
+import {
+  DESKTOP_LANGUAGE_BUTTON_NAME,
+  HAMBURGER_BUTTON_ID,
+  MOBILE_LANGUAGE_BUTTON_NAME,
+} from "@/lib/constants"
+
+import { useMediaQuery } from "@/hooks/useMediaQuery"
+
+const LanguageMorpher = () => {
+  const handleMobileClick = () => {
```
pettinarip, line 15: wow this is the first time I see that we have a button here xD (and a non-reactive approach to do it haha but understand that that was previous implementation)

pettinarip (approved):
LGTM! nice refactor

src/components/Homepage/LanguageMorpher.tsx
wackerow: Yeah this was a tricky work-around since we're linking this button to a far-away elements that changes completely between the desktop nav bar and the mobile menu, so we needed a way to effectively programmatically trigger these menus. Subtle little feature/shortcut =)

src/components/Homepage/LanguageMorpher.tsx
wackerow: Thanks! I'll slip this refactor into the homepage refactor (#15553) since that's where this component is being used and I'm already in there.

PR #15612, opened 2025-06-04
remove old info from withdrawal page

konopkja:
there was a banner on top about withdrawal being live which has been 2 years ago at this point


wackerow (approved):
Thanks @konopkja!

PR #15613, opened 2025-06-04
removing mention of shapella upgrade fromm staking as a service 

konopkja:
since its been over 2 years, i dont think it needs to be mentioned. In fact it may be distracting and irrelevant.


wackerow (approved):
Agree, thanks!

PR #15616, opened 2025-06-04
[debug] fix: tutorial availability to Netlify at build

wackerow:
withhold tutorials from exclusions, allow access at page build


.husky/pre-commit
```diff
@@ -1 +1,3 @@
-npx lint-staged
\ No newline at end of file
+npx lint-staged
+pnpm update-tutorials
+git add src/data/internalTutorials.json
```
pettinarip: Im fine with the current solution but not this part. IMO this will affect the DX and encourage ppl to just skip these hooks.

I'd prefer to call this manually as part of the process of adding a new tutorial. We should update the docs to highlight this.

wackerow: Agree with this approach! Will patch

pettinarip (approved):

PR #15618, opened 2025-06-05
fix: react-hooks/exhaustive-deps

wackerow:
Fixes: `React Hook useEffect has a missing dependency: 'updateWidth'. Either include it or remove the dependency array.  react-hooks/exhaustive-deps`


pettinarip (approved):
👍🏼

PR #15629, opened 2025-06-06
refactor: Developers page to SSR

wackerow:
## Description

- Refactors /developers page to SSR, removing I18nProvider and nested client-side `_components/developers.tsx`
- Refactored `Callout` to `Callout` + `CalloutSSR`: SSR version expects final string values for title and description; Callout remains as-named for better backward compatibility, and expects a translation key for title and description to look up the final string client-side, passing that to the SSR version for canonical styling.

## Related Issue

Ongoing performance initiatives


src/components/CalloutSSR.tsx
pettinarip, line 1: In theory we could just use this one in all the instances, right? we should pass the translated strings.

pettinarip (approved):
LGTM

tata35tg (commented):
👍

PR #15630, opened 2025-06-07
feat: elevate speedrunethereum, add quickstarts

wackerow:
## Preview link
https://deploy-preview-15630--ethereumorg.netlify.app/en/developers


app/[locale]/developers/page.tsx
```diff
@@ -63,6 +65,17 @@ const DevelopersPage = async ({
     namespace: "common",
   })
 
+  const quickstarts = [
+    {
+      description: t("page-developers-quickstart-1-description"),
+      command: "npx create-eth@latest",
```
Comment (pettinarip) [inline app/[locale]/developers/page.tsx]: as a developer, imo we should add context to these commands. At least a link to the doc for that command like https://github.com/scaffold-eth/create-eth or maybe better https://docs.scaffoldeth.io/

app/[locale]/developers/page.tsx
```diff
@@ -63,6 +65,17 @@ const DevelopersPage = async ({
...
+    },
+    {
+      description: t("page-developers-quickstart-2-description"),
+      command: "npm init @rainbow-me/rainbowkit@latest",
```
Comment (pettinarip) [inline app/[locale]/developers/page.tsx]: idem

app/[locale]/developers/page.tsx
Comment (konopkja) [reply app/[locale]/developers/page.tsx]: ok lets have links to solidity and scaffold docs, but instead of buttons have it as hyperlink

pettinarip (approved):

PR #15642, opened 2025-06-09
refactor: nav to server component; add loading states

wackerow:
## Description

- Updates primary Nav bar to more granularly use "use client" components only where needed, migrated rest to server page component.
- Create Nav/Client/index.ts for client-side navigation logic.
- Lazy loads potentially-blocking components with `dynamic`, with `loading` placeholder components
- Cleans up Nav bar structure; Reduces scope of `SearchProvider` (renamed from `Search`) to only component that require it.


src/components/Nav/useThemeToggle.ts
```diff
@@ -0,0 +1,53 @@
...
+   * Adds a keydown event listener to toggle color mode (ctrl|cmd + \).
+   * @param {string} event - The keydown event.
+   */
+  useEventListener("keydown", (e) => {
+    if (e.key !== "\\") return
+    e.preventDefault()
+    if (e.metaKey || e.ctrlKey) {
+      toggleColorMode()
+    } else {
+      // TODO add this to the language picker
```
corwintines: Is this conditional necessary? Still seeing this TODO in it

src/layouts/BaseLayout.tsx
```diff
@@ -16,12 +17,14 @@ const FeedbackWidget = dynamic(() => import("@/components/FeedbackWidget"), {
   ssr: false,
 })
 
-export const BaseLayout = ({
+export const BaseLayout = async ({
   children,
   // contentIsOutdated,
   // contentNotTranslated,
   lastDeployLocaleTimestamp,
+  locale,
```
pettinarip: minor suggestion: to avoid prop drilling we could do in server components and maybe adopt the pattern in other cases
```tsx
import {getLocale} from 'next-intl/server';
const locale = await getLocale();
```

src/components/Nav/Client/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+            isSecondary
+            className="group px-2 max-md:hidden xl:px-3 [&>svg]:transition-transform [&>svg]:duration-500 [&>svg]:hover:rotate-12 [&>svg]:hover:text-primary-hover"
+            onClick={toggleColorMode}
+          >
+            <ThemeIcon className="transform-transform duration-500 group-hover:rotate-12 group-hover:transition-transform group-hover:duration-500" />
+          </Button>
+        )}
+
+        {desktopScreen && (
+          <LanguagePicker>
```
pettinarip: We would need to lazy load the lang picker and the search button as well since it will be not interactive in slow connections while loading.

Basically we need to lazy load all the nav components (except the logo link) IMO since they all need js for execution.

src/components/Nav/Client/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+                {!desktopScreen && (
+                  <MobileNavMenu
+                    toggleColorMode={toggleColorMode}
+                    linkSections={linkSections}
+                    toggleSearch={onOpen}
+                    className="flex md:hidden"
+                  />
+                )}
+              </>
```
pettinarip, line 111: maybe there is something we can do here since we have a tiny layout shift in mobile.

https://github.com/user-attachments/assets/a09e61c4-6f1d-4063-88ad-ff027c48b4a2

src/components/Nav/index.tsx
```diff
@@ -1,93 +1,31 @@
...
+import ClientSideNav from "./Client"
 
-// TODO display page title on mobile
-const Nav = () => {
-  const { toggleColorMode, linkSections } = useNav()
-  const { t } = useTranslation("common")
-  const navWrapperRef = useRef(null)
-  const isClient = useIsClient()
-  const desktopScreen = useBreakpointValue({ base: false, md: true })
+const Nav = async ({ locale }) => {
```
pettinarip: `const locale = await getLocale()`

pettinarip (commented):
@wackerow great job! looks fantastic 👍🏼 

@nloureiro @konopkja would like to get the designers' input on the loading skeletons for the menu objects and see if we are all aligned. Dropping a short video demoing what it looks like on slow connections. 

https://github.com/user-attachments/assets/1ee97808-5970-468c-9e93-7ee01d940bb8

src/components/Nav/Client/index.tsx
wackerow: On it =)

src/components/Nav/Client/index.tsx
wackerow: Yeah good call... this displays it quicker, but it's not interactive immediately which hurts us. Will patch

wackerow: From trying this out though, we're already lazy loading the SearchProvider component though, which loads an interactive search component when ready.

The theme toggle as well, seems to be interactive immediately, so I think we're good there.

The LanguagePicker though will be visible without interactivity when it first loads, so I'll lazy load that as well.

src/layouts/BaseLayout.tsx
wackerow: Oh thank you! Can't believe I didn't find this getter before; was the only reason I drilled the locale on all of these... Will definitely start using this instead, starting with this PR.

src/components/Nav/index.tsx
wackerow: idem =) grazi

src/components/Nav/Client/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+                    className="flex animate-fade-in md:hidden"
+                  />
+                )}
+              </>
+            )
+          }}
+        </SearchProvider>
+
+        {desktopScreen && (
+          <Button
```
pettinarip, line 117: I think we also need to display a skeleton for this button. Currently, the button is no responsive when clicked in pre-load time, affecting INP and FID. We can tackle this is a separate PR.

corwintines (approved):

pettinarip (approved):
@wackerow great job, merging.

There are still two buttons above the fold that should be also covered: the color mode toggle and the morpher button. Both should be delayed.

pettinarip:
lol, @corwintines we were in sync with this PR haha

PR #15652, opened 2025-06-11
Fix failing story

pettinarip:
update imports and i18n setup


pettinarip:
auto-merging to unblock `dev` env

PR #15653, opened 2025-06-11
fix 10year banner translations

pettinarip:
wraps client side component with i18n translations


wackerow:
Refactored to pass strings from server; we can/should discuss more on how we want to approach these cases though going forward

PR #15662, opened 2025-06-12
Patch low contrast text in banner

pettinarip:
from:
image

to:
image


wackerow:
Hey @pettinarip, where are you seeing this bug? Which browser? I'm not seeing it, and noting we already have this applied to the `BannerNotification` wrapper:

https://github.com/ethereum/ethereum-org-website/blob/1abbdb8e1fe3e8ad24d8ac80129ec49c4341b878/src/components/Banners/BannerNotification.tsx#L17

The `[&_a]` selector classes were put in place to override the links in this component due to `bg-primary-action`... are we sure this change is needed?

PR #15677, opened 2025-06-17
perf: Homepage image optimization

wackerow:
## Description
- Implements [Art Direction](https://nextjs.org/docs/app/api-reference/components/image#art-direction) approach for loading separate image assets depending on screen size
- Applies this to homepage SectionBanner images and hero
- SectionBanner images now using a "portrait" version on desktop, which is a different version of the image with sides cropped out, optimized for how we display these images on desktop. `srcset` used to tell next/image to target 512px width, since these do not extend full width.
- SectionBanner on mobile uses existing landscape images, optimized for 100vw since they extend full width.
- Apply similar approach to homepage Hero image, but in reverse, where 2xl+ screens can get away with a slightly cropped version (trimming top and bottom), while other screens still need the full image. 
- Quality of images dropped throughout. Hero image has progressive quality drop from 20 on largest, to 10, to 5 on smallest screens. (of course, optimizing for load speed)

## Todo
- [x] Compare Lighthouse performance scores for homepage with production—both desktop and mobile
- [x] Gather feedback on optimal `quality` setting, balancing trade-offs of load speed vs visual quality

**Playground: https://deploy-preview-15677--ethereumorg.netlify.app/en/hero**

If scores show improvement, we can extract a re-usable component that uses this approach.

## Related Issue
Ongoing performance initiatives: Homepage / Lighthouse LCP score


src/components/Homepage/HomepageSectionImage.tsx
```diff
@@ -0,0 +1,97 @@
...
+    <picture className={className}>
+      <source
+        media={`(min-width: ${breakpointAsNumber.md}px)`}
+        srcSet={desktop}
+      />
+      <source
+        media={`(max-width: ${breakpointAsNumber.md - 1}px)`}
+        srcSet={mobile}
+      />
+      <img {...rest} alt={alt} className="h-auto w-full object-cover" />
```
wackerow:
```suggestion
      <img {...rest} alt={alt} className="h-full w-full object-cover" />
```

wackerow:
Noting- Much of this PR was simply testing out various image sizes and won't be merged, though a few aspects here we could still benefit from, such as loading portrait-cropped versions of the homepage section banners for desktop. Lower priority at the moment

wackerow:
Noting that lighthouse scores aren't really responding much to these changes, and we've identified that the Nav component is much more of the problem here; not just the images. That being said, the image improvements will still help contribute to faster loads and better UX. 

Trimmed off the testing commits from this branch and opened those separately for further testing in #16004. This PR keeps the "art direction" changes and quality adjustments.

Could consider backing off on the quality drop here, though after testing it sounds like most folks found the lower-than-normal quality settings to appear perfectly fine without too much pixel-peeping. Feedback still welcome.

Issue #15671, opened 2025-06-15
Bug report - anchor links in sidebar ToC doesn't pull you


nixorokish:
### Describe the bug

the anchor links on [the maxeb page](https://ethereum.org/en/roadmap/pectra/maxeb/) work until you get to the FAQ. once you hit the FAQ, they no longer draw the point of focus to the linked point

### To reproduce

1. Go to https://ethereum.org/en/roadmap/pectra/maxeb/
2. Click on any table of contents item above FAQ to see that it works there
3. Click on any table of contents item below FAQ to see that it doesn't work there
4. Tested on both Firefox & Chrome on Sonoma 14.4.1

### Expected behavior

Draw user to the chosen part of the page

### Desktop (please complete the following information)

- OS: MacOS Sonoma 14.4.1 
- Firefox 139.0.1
- Brave 1.78.97

### Would you like to work on this issue?

- [ ] Yes
- [x] No

joeurassa:
Hi @pettinarip , the fix didn't work.. UInless someone else is working on it. I will make a pr

PR #15679, opened 2025-06-17
Update markdownChecker.ts

joeurassa:
## Trailing White Space Regex Original / wrong code wasn't commented out. 

This PR fixes that.

Reason: Might be causing an issue with the FAQ in parc roadmap.

## Issue: https://github.com/ethereum/ethereum-org-website/issues/15671


wackerow:
Thanks for jumping in @joeurassa! Looks like this was resolved with another PR... Going to close this out so we keep the markdown checker as-is. Cheers!

PR #15681, opened 2025-06-18
Fix home hero

pettinarip:
## Description

- Fixes translations in home hero
- Removes the home hero story


.storybook/main.ts
```diff
@@ -77,5 +77,8 @@ const config: StorybookConfig = {
 
     reactDocgen: "react-docgen-typescript",
   },
+  features: {
+    experimentalRSC: true,
```
pettinarip, line 81: Im adding this experimental flag for future stories that may use RSC but it didn't work for the home hero. That is why Im removing it for now until we have a solution.

ref https://storybook.js.org/docs/get-started/frameworks/nextjs#react-server-components-rsc

src/components/Hero/HomeHero/index.tsx
```diff
@@ -1,21 +1,24 @@
...
-export type HomeHeroProps = Pick<CommonHeroProps, "heroImg"> & ClassNameProp
+export type HomeHeroProps = Pick<CommonHeroProps, "heroImg"> &
+  ClassNameProp & {
+    locale: Lang
+  }
 
-const HomeHero = ({ heroImg, className }: HomeHeroProps) => {
-  const { t } = useTranslation("page-index")
+const HomeHero = async ({ heroImg, className, locale }: HomeHeroProps) => {
+  const t = await getTranslations({ locale, namespace: "page-index" })
```
pettinarip, line 14: Tried to use `getLocale` but it was not matching the locale in the url 🤷🏼

corwintines (approved):

PR #15691, opened 2025-06-19
Start using Blockscout more (it is Open Source and Etherscan is closed source)

austintgriffith:
It would be cool if we started using Blockscout more because they are Open Source and Etherscan isn't 

Also they are sending traffic to SpeedRunEthereum and that is awesome. 

Let's support open source and let's support great ethereum developer education! 

## Description

I replaced Etherscan links in a few places and I added supplemental Blockscout links in a couple places. 

Really just testing the water here to see if you guys are cool with this.


wackerow (approved):
Looks great! Love it... cc: @lukassim Strings updates will need to flow through Crowdin 🙏 

Thanks @austintgriffith!

PR #15716, opened 2025-06-25
[wip] AB testing prototype

wackerow:


minimalsm:
@claude can you review what we have in this draft PR so far

.env.example
```diff
@@ -30,6 +30,27 @@
...
+
+# Example: Homepage Persona CTAs Test
+# ABTEST_HOMEPAGE_PERSONA_CTAS_ENABLED=false
+# ABTEST_HOMEPAGE_PERSONA_CTAS_ID=1
+# ABTEST_HOMEPAGE_PERSONA_CTAS_VARIANTS="original:50,variant_b:50"
+
+# Example: Multi-variant Wallet Layout Test  
+# ABTEST_WALLET_LAYOUT_ENABLED=false
+# ABTEST_WALLET_LAYOUT_ID=2
+# ABTEST_WALLET_LAYOUT_VARIANTS="original:40,list_view:30,carousel:30"
```
pettinarip: I don't think we'll have any issues at first but I'm concerned about how many env vars we might need in the future, which could quickly hit the size limit.

We could think about a more scalable solution. We could:
- store all these configs in a json file hosted in an external cdn/server
- set just one env var pointing to the json file endpoint
- and load it at build time

That way we can keep env vars minimal and we can switch the configs between envs.

docs/ab-testing.md
```diff
@@ -0,0 +1,162 @@
...
+- Use consistent naming between Matomo and code
+- Always provide a fallback for when tests are disabled
+
+### Variant Weights
+- Use meaningful numbers (50:50, 70:30, etc.)
+- Weights don't need to sum to 100
+- Can adjust without code changes
+
+### Testing Strategy
+1. **Local Development**: Test with `ENABLED=true` and different weights
```
pettinarip: not sure about this but I think you are referring to the abtest enable var no?
```suggestion
1. **Local Development**: Test with `ABTEST_[TEST_NAME]_ENABLED=true` and different weights
```

docs/ab-testing.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Performance
+- Server-side rendering means no layout shifts
+- Minimal JavaScript overhead (only tracking)
+- Cookies are lightweight and scoped
+
+## Troubleshooting
+
+### Test Not Showing Variants
+1. Check `ENABLED=true` in env vars
```
pettinarip:
```suggestion
1. Check `ABTEST_[TEST_NAME]_ENABLED=true` in env vars
```

src/lib/ab-testing/config.ts
```diff
@@ -0,0 +1,39 @@
+import { ABTestConfig } from "./types"
+
+function getEnvVar(key: string, defaultValue: string = ""): string {
+  return process.env[key] || defaultValue
```
pettinarip: this could be another potential issue where we store dynamic variables instead of static ones. This could lead to missing vars if the bundler can't statically analyze the vars.

wackerow: gone!

src/lib/ab-testing/index.ts
```diff
@@ -0,0 +1,8 @@
+// Types (safe for both client and server)
+export type { ABTestAssignment, ABTestConfig, ABTestVariants } from "./types"
+
+// Server utilities (only import these in server components)
+export { getABTestAssignment, getABTestConfigs } from "./server"
+
+// Note: Server actions are exported from ./actions, not here
```
pettinarip: ✂️ ?

app/api/ab-config/route.ts
```diff
@@ -0,0 +1,145 @@
+import { NextResponse } from "next/server"
+
+import { IS_PREVIEW_DEPLOY } from "@/lib/utils/env"
```
pettinarip, line 3: Not a request to change, just FYI, I had an issue with imported env vars before. If you hit an issue with this flag, the reason could be that Netlify is expecting you to declare them with `process.env` inside the lambda file.

app/api/ab-config/route.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (data.result === "error" || !Array.isArray(data)) {
+      console.error(
+        "[AB Config] Matomo API error:",
+        data.message || "Invalid response"
+      )
+      return NextResponse.json(
+        {},
+        {
+          headers: {
+            "Cache-Control": "s-max-age=300, stale-while-revalidate=600",
```
pettinarip, line 72: do we need cache control for the error response?

pettinarip (approved):
Much better with the server side approach! nice job @wackerow

Maybe Im missing something but i don't see the panel in the preview deploy and not sure exactly how to test it.

Other than that, the code looks fine 👍🏼

app/api/ab-config/route.ts
wackerow: Okay, will keep an eye on it, but seems to be working in this case:

https://deploy-preview-15716--ethereumorg.netlify.app/api/ab-config

```
{"AppTest":{"id":"3","enabled":true,"variants":[{"name":"Original","weight":48},{"name":"Variation1","weight":52}]},"HomepagePersonaCTAs":{"id":"2","enabled":true,"variants":[{"name":"Original","weight":75},{"name":"Variation1","weight":25}]}}
```

PR #15717, opened 2025-06-25
fix: prevent RadialChart from overlapping sticky navigation

### LINKED ISSUE #15712: “Time to next block” stat renders above sticky submenu on /resources scroll
### Description
On ethereum.org/resources, when the user scrolls and the submenu (Privacy & security, Ethereum resilience, etc.) becomes sticky at the top, the “Time to next block” circular stat remains rendered above the sticky menu.

Image

This results in a visual overlap, where the stat’s animated arc cuts in front of the navigation — which is unexpected and breaks visual hierarchy.

### Steps to Reproduce
1. Go to https://ethereum.org/en/resources
2. Scroll slowly until the submenu (Privacy & security, etc.) becomes sticky
3. Watch how the circular timer from the “Time to next block” section overlaps the sticky menu

### Expected Behavior
The sticky submenu should always render on top of other page content, including animated stat elements.

### Actual Behavior
The circular stat arc renders above the sticky submenu — breaking layering expectations.

minimalsm:
Fixes #15712

This PR resolves the z-index layering issue where the "Time to next block" circular stat was overlapping the sticky submenu on the /resources page.

## Changes
- Added `z-base` class to RadialChart component to set z-index: 0
- Ensures RadialChart renders below sticky navigation (z-index: 1100)

## Testing
- Navigate to `/resources` page
- Scroll to make submenu sticky
- Verify circular timer no longer overlaps the sticky menu

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/components/RadialChart/index.tsx
+++ b/src/components/RadialChart/index.tsx
@@ -73,34 +73,36 @@ const RadialChart = ({
 
   return (
     <div className={cn("flex flex-col items-center", className)}>
-      <div className="relative overflow-hidden">
-        <RadialBarChart
-          width={170}
-          height={90}
-          cx={85}
-          cy={80}
-          innerRadius={70}
-          outerRadius={100}
-          barSize={10}
-          data={data}
-          startAngle={185}
-          endAngle={-5}
-        >
-          <PolarAngleAxis
-            type="number"
-            domain={[0, totalValue || 100]}
-            angleAxisId={0}
-            tick={false}
-          />
-          <RadialBar
-            background={{ fill: "hsla(var(--body-light))", strokeWidth: 1 }}
-            dataKey="value"
-            cornerRadius={16}
-            fill="hsla(var(--accent-a))"
-          />
-        </RadialBarChart>
-        <div className="absolute bottom-0 left-1/2 -translate-x-1/2 transform text-3xl font-black">
-          {displayValue || value}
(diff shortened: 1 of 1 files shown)


minimalsm:
@coderabbitai whatcha think

minimalsm:
@claude this approach you suggested to solve #15712 was incorrect. simply adding a z-index does nothing here. the menu itself that has the overlapping already has a very high (1000+) z index. the issue is specific to something else in the recharts wrapper (the text below the chart does not overlap. plz try again :)

minimalsm:
@claude nice try but also not correct. Last attempt before I uninstall you. Try really hard, understand the issue deeply, and please don't let me down or I'll be sad

minimalsm:
okay @claude you were getting closer. i was unkind in the last comment and I apologise!

here's the dealio:

- the position relative on the wrapper is because we want to absolutely position the number that appears underneath the chart
- however, the position relative means that overflow hidden doesn't actually work on this component.
- we probably need to remove position relative on this wrapper, and come up with a different organisational structure to our divs to ensure the absolute positioning on the text, but dont get the overflow caused by position relative on the wrapper 

this should be enough context for you to find a suitable solution

wackerow (changes requested):
image

🤔 yeah this is a perky one for some reason

wackerow:
See #15889

PR #15736, opened 2025-06-27
fix: use direct import for Button in LanguagePicker to support ref

JoeChenJ:
## Description
The LanguagePicker could not open due to a ref error caused by a lazy loaded Button component. 

To fix this, the Button is now:

- Lazy loaded only for the ThemeSwitcher.
- Directly imported for the LanguagePicker.


wackerow:
Ah, thanks @JoeChenJ! Great minds think alike! This has already been merged from #15737, going to close this out. Appreciate that quick response though! ❤️

PR #15745, opened 2025-06-29
Fix broken link to Optimism dapps portal

MozirDmitriy:
Replaced the outdated Optimism dapps portal link (https://www.optimism.io/apps) with the current official portal (https://community.optimism.io/) in the Turkish translation file for adding Layer 2s. This ensures users are directed to the correct resource for exploring dapps on Optimism.


wackerow:
Closing out per above comment

PR #15781, opened 2025-07-03
Optimize Morpher performance for mobile devices

pettinarip:
## Description

- Fix memory management by using useRef for timeout variables
- Add accessibility support with usePrefersReducedMotion hook
- Reduce animation frame rate from 30 to 24 FPS for better performance
- Implement deferred animation start (2s delay) to improve initial page load
- Prevent multiple simultaneous animations with proper state tracking
- Optimize React re-renders by only updating state when text changes
- Add proper cleanup to prevent memory leaks and runaway animations

Note: part of the efforts to reduce INP and TBT.

### DIFF
--- a/src/components/Morpher.tsx
+++ b/src/components/Morpher.tsx
@@ -1,6 +1,8 @@
 "use client"
 
-import { useEffect, useState } from "react"
+import { useEffect, useRef, useState } from "react"
+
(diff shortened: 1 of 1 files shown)


wackerow:
Noting a bug where the first word duplicates. Unrelated to the PR changes though so brought it in.

PR #15792, opened 2025-07-07
docs: fix syntax error by adding missing equals signs

mdqst:


wackerow (approved):
Good catch! Thanks @mdqst!

PR #15818, opened 2025-07-14
Optimize home svgs

pettinarip:
## Description

Optimizes the svgs used in the home page with [svgo](https://svgo.dev/).


wackerow:
Hm... how significant is the anticipated impact here? Noticing it minifies all the svg code.

It's not every day, but I know I'm personally inside svg code not-infrequently and do appreciate having the elements broken on new lines... I can of course lint it myself as needed, but would add time.

wackerow (approved):

PR #15819, opened 2025-07-14
Optimize values marquee

pettinarip:
This PR prevents the use of animations early on.

## Description

Use Intersection Observer to load the ValuesMarquee only when it is needed.


wackerow (changes requested):
Would request at least the Loading component be addressed

pettinarip:
@wackerow sorry for the delay on this. I'll look into this better and get back to you with metrics and a fix for this (in case this PR is still justified).

wackerow:
Yeah ok.. just rechecked https://deploy-preview-15819--ethereumorg.netlify.app/en/#values and it seems to load appropriately, unlike my previous comment. I suppose if noone else is getting this bug we can disregard and could probably pull this in

PR #15878, opened 2025-07-19
Fix: Updated ERC-20 interface syntax and corrected return types

thesheriflawal:
### **Correct ERC-20 interface syntax and types in documentation**

Basically, what I added was the "memory" keyword to "string" return types ("name( )" and "symbol( )", which is requires in Solidity versions 0.5.0 and above. This matters because the current version on the website uses outdated or incomplete syntac ("string" without memory location) which might cause an error in modern Solidity versions. 

## These changes align with the official [EIP-20] and solidity version 0.5.0's standards
[EIP-20 Token Standard](https://eips.ethereum.org/EIPS/eip-20)
[Solidity 0.5.0 Breaking Changes](https://docs.soliditylang.org/en/v0.5.0/050-breaking-changes.html#explicit-data-location)

### DIFF
--- a/public/content/developers/docs/standards/tokens/erc-20/index.md
+++ b/public/content/developers/docs/standards/tokens/erc-20/index.md
@@ -52,11 +52,11 @@ From [EIP-20](https://eips.ethereum.org/EIPS/eip-20):
 ### Methods {#methods}
 
 ```solidity
-function name() public view returns (string)
-function symbol() public view returns (string)
+function name() public view returns (string memory)
+function symbol() public view returns (string memory)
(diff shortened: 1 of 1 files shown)


wackerow:
Hey @thesheriflawal, thanks for the update... Forgive me since I'm not in the thick of ERC-20's, but the https://eips.ethereum.org/EIPS/eip-20 link still shows these as they were previously written:

image

And spot checking, I don't see mention of "string memory"

image

Just want to make sure we get the best info here; could you help clarify where the info about these changes are located?

PR #15886, opened 2025-07-20
Fix torch history card date display and placeholder messaging

minimalsm:
## Fix 1: Redundant date display for current torchbearers

Problem: Currently, torch history cards display "From [date] to [date]" even when the torchbearer still holds the NFT. When someone is the current holder, both dates are identical, creating confusing and redundant text.

Solution: Conditionally render the date text based on isCurrentHolder prop:

Current holders: Show "From [date]" only
Previous holders: Show "From [date] to [date]" range

## Fix 2: Generic placeholder messaging

Problem: All future torchbearer placeholders show the same generic "Coming soon!" message, providing no indication of when each person will receive the torch.

Solution: Replace generic messaging with specific dates showing when each torchbearer will receive the torch:

Torchbearer 1: "Coming 20th!"
Torchbearer 2: "Coming 21st!"
Torchbearer 3: "Coming 22nd!"
And so on...

image

### DIFF
--- a/app/[locale]/10years/_components/TorchHistorySwiper/index.tsx
+++ b/app/[locale]/10years/_components/TorchHistorySwiper/index.tsx
@@ -19,6 +19,22 @@ type TorchHistorySwiperProps = {
   currentHolderAddress: Address | null
 }
 
+const getOrdinalSuffix = (num: number): string => {
+  const lastDigit = num % 10;
+  const lastTwoDigits = num % 100;
+  
+  if (lastTwoDigits >= 11 && lastTwoDigits <= 13) {
+    return `${num}th`;
+  }
+  
+  switch (lastDigit) {
+    case 1: return `${num}st`;
+    case 2: return `${num}nd`;
+    case 3: return `${num}rd`;
+    default: return `${num}th`;
+  }
+};
+
 const TorchHistorySwiper = ({
   holders,
   currentHolderAddress,
@@ -40,8 +56,8 @@ const TorchHistorySwiper = ({
       // Create placeholder for future holder
       return {
         address: `placeholder-${index}` as Address,
-        name: `Future Bearer ${index + 1}`,
-        role: "Coming soon...",
+        name: `Torchbearer ${index + 1}`,
+        role: `Coming ${getOrdinalSuffix(20 + index)}!`,
         twitter: "",
         event: {
           from: "0x0000000000000000000000000000000000000000" as Address,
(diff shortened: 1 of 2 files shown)


app/[locale]/10years/_components/TorchHistorySwiper/index.tsx
```diff
@@ -19,6 +19,22 @@ type TorchHistorySwiperProps = {
   currentHolderAddress: Address | null
 }
 
+const getOrdinalSuffix = (num: number): string => {
+  const lastDigit = num % 10;
+  const lastTwoDigits = num % 100;
+  
+  if (lastTwoDigits >= 11 && lastTwoDigits <= 13) {
+    return `${num}th`;
+  }
```
Comment (minimalsm) [inline app/[locale]/10years/_components/TorchHistorySwiper/index.tsx]: will never happen as num will range from 20 to 30

pettinarip (dismissed):

app/[locale]/10years/_components/TorchHistorySwiper/index.tsx
```diff
@@ -19,6 +19,22 @@ type TorchHistorySwiperProps = {
...
+  
+  switch (lastDigit) {
+    case 1: return `${num}st`;
+    case 2: return `${num}nd`;
+    case 3: return `${num}rd`;
+    default: return `${num}th`;
+  }
+};
```
Comment (wackerow) [inline app/[locale]/10years/_components/TorchHistorySwiper/index.tsx]:
```suggestion
const getOrdinalSuffix = (num: number): string => {
  const pr = new Intl.PluralRules('en', { type: 'ordinal' });
  const rule = pr.select(num);
  
  switch (rule) {
    case "one": return `${num}st`;
    case "two": return `${num}nd`;
    case "few": return `${num}rd`;
    default: return `${num}th`;
  }
}
```

Consider a JS native approach using Intl.PluralRules

Comment (minimalsm) [reply app/[locale]/10years/_components/TorchHistorySwiper/index.tsx]: Ah yes, good idea @wackerow !

app/[locale]/10years/_components/TorchHistorySwiper/index.tsx
```diff
@@ -40,8 +56,8 @@ const TorchHistorySwiper = ({
       // Create placeholder for future holder
       return {
         address: `placeholder-${index}` as Address,
-        name: `Future Bearer ${index + 1}`,
-        role: "Coming soon...",
+        name: `Torchbearer ${index + 1}`,
+        role: `Coming ${getOrdinalSuffix(20 + index)}!`,
```
Comment (minimalsm) [inline app/[locale]/10years/_components/TorchHistorySwiper/index.tsx]:
```suggestion
        role: `Coming July ${getOrdinalSuffix(20 + index)}!`,
```

minimalsm:
Closing in favour of #15900

PR #15897, opened 2025-07-21
feat: add playwright test for start page [fixes #15896]

### LINKED ISSUE #15896: Add e2e test for start page to test wallet connection

Add a playwright test for the start page to verify that the wallet setup (currently using RainbowKit) is working correctly.

As suggested here: https://github.com/ethereum/ethereum-org-website/pull/15550#issuecomment-2977415669

Add a new playwright test that connects to a default account using [mock connector from wagmi](https://wagmi.sh/react/api/connectors/mock).

Using a real wallet connection (like [synpress](https://github.com/Synthetixio/synpress) does) would be overkill.

### Additional context

_No response_

- [x] Yes
- [ ] No

agualis:
Fixes #15896 by adding a playwright test for start page to test wallet connection

## Description

Adds a new rainbow kit group with wagmi's mock connector (only when the app is running in localhost).

Another approach could be using an env var like `NEXT_PUBLIC_ALLOW_WALLET_MOCK` that must be `true` when running the app to be tested by playwright but I find it less straight forward (so worse DX).


src/config/rainbow-kit.ts
```diff
@@ -9,21 +9,37 @@ import {
...
-        rainbowWallet,
-        metaMaskWallet,
-        zerionWallet,
-        oneKeyWallet,
-        walletConnectWallet,
-      ],
-    },
-  ],
+  wallets: walletGroups,
+  ssr: isLocalhost,
```
agualis: Not using `ssr: true` casues hydration issues. Is there an explicit reason why you are not using it by default?

wackerow:
Thanks @agualis! Will get some eyes on this soon

src/config/rainbow-kit.ts
pettinarip: True, I don't see why not to enable this.

agualis:
Thanks for reviewing @pettinarip 

> The test looks good and is working perfectly for me locally, but I think we could make a small refactor to organize the code better.
> 

> Move `mockWallet` to the fixtures folder. 

Done ✅

>  * Create a `rainbowkitConfig` specifically for the test runtime.
> We could export the config from `rainbow-kit.ts` and then extend it in a `test.beforeEach` block, for example. This way, we keep the code clean and decoupled from the test logic.

I understand your concern but this is not so trivial because, in order to update `rainbowkitConfig` from the tests we should store it in a react `useState` and expose a method to be called from playwright (which is not so straightforward).

If you want to keep the original `rainbow-kit.ts` file clean, we could isolate the test config in its own file like I show [here](https://github.com/agualis/ethereum-org-website/pull/1). 
 
 ✅ Test setup would be more isolated. 
 ❌ We introduce one test related line in `WalletProviders.tsx` to chose between `rainbowkitTestConfig` and `rainbowkitConfig` which is not ideal but much simpler that the mentioned useState approach. 
 
Happy to use an alternative simpler approach if you know one.

src/config/rainbow-kit.ts
wackerow: Assuming we still want this updated?

```suggestion
  ssr: true,
```

wackerow:
@agualis Thanks for the updates here!—any chance you could help clear up the merge conflicts? 🙏 
Pinged @pettinarip for another review

src/config/rainbow-kit.ts
```diff
@@ -9,21 +9,37 @@ import {
...
+      metaMaskWallet,
+      zerionWallet,
+      oneKeyWallet,
+      walletConnectWallet,
+    ],
+  },
+]
+
+const isLocalhost =
+  typeof window !== "undefined" && window.location.hostname === "localhost"
```
pettinarip: Maybe its better to change this to check the env directly. We could control 

```suggestion
// https://github.com/ethereum/ethereum-org-website/blob/dev/src/lib/utils/env.ts
const isDevOrCI = IS_DEV || IS_CI
```

we would need to add `IS_CI` with `process.env.CI`

agualis:
> @agualis Thanks for the updates here!—any chance you could help clear up the merge conflicts? 🙏 Pinged @pettinarip for another review

I don't have time this week but I will try next one.

agualis:
> @agualis Thanks for the updates here!—any chance you could help clear up the merge conflicts? 🙏

@pettinarip @wackerow Cleared up!

corwintines:
@all-contributors can you add @agualis for code

PR #15899, opened 2025-07-21
Hotfix static render 10y page

pettinarip:
## Description

- renders page statically for faster loads
- get current holder from the transfer events to reduce temporary discrepancies between the 2 data sources (contract & etherscan api)

### DIFF
--- a/src/lib/torch/index.ts
+++ b/src/lib/torch/index.ts
@@ -94,22 +94,9 @@ export const getHolderEvents = async (
   })
 }
 
-export const getCurrentHolderAddress = cache(
-  async () => {
-    const publicClient = getPublicClient(config)
-
-    // If not burned, get the current holder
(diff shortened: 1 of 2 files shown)


src/lib/torch/index.ts
```diff
@@ -94,22 +94,9 @@ export const getHolderEvents = async (
...
-      functionName: "currentHolder",
-    })) as Address
-
-    return currentHolderAddress
-  },
-  ["torch-current-holder"],
-  { revalidate: 86400 }
-)
+export const getCurrentHolder = (holderEvents: TorchHolderEvent[]) => {
+  return holderEvents[holderEvents.length - 1]
```
pettinarip, line 98: assumes last transfer event is current holder

pettinarip:
> Seeing this currently
> 
> Screenshot 2025-07-21 at 14 21 55

expected in the preview deploy env

PR #15904, opened 2025-07-22
fix: Matomo ab testing assignment distribution

wackerow:
## Description
- use additional privacy-preserving header data to increase entropy of users when being assigned an AB test group
- Uses improved hash function to help evenly distribute users based on fingerprint

## Related Issue
AB Testing


wackerow:
Closing in lieu of hot-patch: #15927

PR #15935, opened 2025-07-25
let's update some of these links and add new links to blockscout (open source!) 

austintgriffith:
let's update some of these links and add new links to blockscout on some popular pages to drive more traffic to blockscout because they are open source and etherscan is not!

(also blockscout is supporting Speedrun Ethereum and we want to get them more support!)


wackerow (approved):
Thanks @austintgriffith!

wackerow:
@all-contributors please add @austintgriffith for tooling and maintenance

PR #15941, opened 2025-07-26
feat: Enterprise contact form

wackerow:
## Description
- Adds email contact form to Enterprise footer
- Form forwards minimum viable information as email to Enterprise team using [AWS SES](https://docs.aws.amazon.com/ses/latest/dg/send-email-raw.html)

## Preview link
https://deploy-preview-15941--ethereumorg.netlify.app/en/enterprise/#get-in-touch


package.json
```diff
@@ -30,6 +30,7 @@
     "test:e2e:report": "playwright show-report tests/e2e/__report__"
   },
   "dependencies": {
+    "@aws-sdk/client-ses": "^3.859.0",
```
pettinarip, line 33: @wackerow can we avoid installing the sdk and use direct http calls to the API instead? to reduce the amout of total deps it is installing. Unless we need to use the sdk for other reasons that Im not aware of.

app/[locale]/enterprise/_components/ContactForm/index.tsx
```diff
@@ -0,0 +1,303 @@
...
+  "mac.com",
+  "rocketmail.com",
+  "yahoo.co.uk",
+  "googlemail.com",
+  "mailinator.com",
+  "10minutemail.com",
+  "guerrillamail.com",
+]
+
+const sanitizeInput = (input: string): string =>
```
Comment (pettinarip) [inline app/[locale]/enterprise/_components/ContactForm/index.tsx]: this function seems to be duplicated. Can we also place it under the utils folder?

app/[locale]/enterprise/_components/ContactForm/index.tsx
```diff
@@ -0,0 +1,303 @@
...
+  input
+    .replace(/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi, "")
+    .replace(/javascript:/gi, "")
+    .replace(/on\w+\s*=/gi, "")
+    .replace(/&lt;script/gi, "")
+    .replace(/&lt;\/script/gi, "")
```
Comment (pettinarip) [inline app/[locale]/enterprise/_components/ContactForm/index.tsx]: would add a more robust check here. maybe https://github.com/cure53/DOMPurify which is pretty light

app/api/enterprise-contact/route.ts
```diff
@@ -0,0 +1,183 @@
...
+  },
+})
+
+// Log the region being used for debugging
+console.log("Using AWS SES region:", process.env.SES_REGION || "us-east-2")
+
+// Simple in-memory rate limiting (in production, use Redis or similar)
+const requestHistory = new Map<string, number[]>()
+
+function isRateLimited(identifier: string): boolean {
```
pettinarip: due to the nature of serverless functions, any in-memory logic won't work. This could helps us here https://github.com/upstash/ratelimit-js but not totally convinced that it is worth it

src/lib/utils/sanitize.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Lightweight sanitizer for user input to prevent XSS attacks
+ * Specifically designed for plain text content
+ *
+ * @param input - The string to sanitize
+ * @returns Sanitized string with dangerous content removed
+ */
+export function sanitizeInput(input: string): string {
```
wackerow, line 8: @pettinarip per discussion, looks like we could need to install more dependencies that could be overkill for what we're doing here... as a fallback, asked claude to catch the main attack vectors. 

I see this as a solution for the time, but to be investigated further as the repo advances.

Pulled it into lib/utils and removed the duplication.

app/[locale]/enterprise/_components/ContactForm/index.tsx
```diff
@@ -0,0 +1,295 @@
...
+    setSubmissionState("submitting")
+    setErrors({})
+
+    try {
+      const sanitizedData = {
+        email: sanitizeInput(formData.email),
+        message: sanitizeInput(formData.message),
+      }
+
+      const response = await fetch("/api/enterprise-contact", {
```
Comment (pettinarip) [inline app/[locale]/enterprise/_components/ContactForm/index.tsx:183]: good place to use `react-query` with `useMutation`

pettinarip (approved):
LGTM

package.json
wackerow: to simplify the security setup that goes into the aws connect, I think we're stuck with this one for the time

PR #15944, opened 2025-07-27
Minor code enhancements

bskrksyp9:
## Description

Observed that, 

1. only isTorchBurned() with try/catch block, so Wrapped getInnovationCards(), getAdoptionCards(), and getTransferEvents() with try/catch block to

> Prevent runtime crashes on fetch failures
> Log meaningful errors without breaking UI rendering
> Switched from const to let for conditional assignment on failure

2. two times evaluation of Object.entries 

> used `eventsEntries` to make it compute only once

### DIFF
--- a/app/[locale]/10years/page.tsx
+++ b/app/[locale]/10years/page.tsx
@@ -140,6 +140,7 @@ const Page = async ({ params }: { params: Promise<{ locale: Lang }> }) => {
   )
 
   const currentHolder = getCurrentHolder(torchHolders)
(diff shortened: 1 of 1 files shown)


wackerow:
Hey @bskrksyp9, appreciate the PR... Some of this has been patched already (content adjustment) but the other try/catch blocks imo should break the build if they fail for any reason. These do not include any fetching of external data so if it's not working we should be alerted to this before the deploy gets published. Feel free to push back, and @pettinarip feel free to weigh in, but going to close out for now. Thanks again!

Issue #15940, opened 2025-07-26
Minor Typo in Tutoraial


panditdhamdhere:
https://ethereum.org/en/developers/tutorials/ethereum-for-web2-auth In this tutorial, there is a minor typo. ( see the image below )

<img width="909" height="355" alt="Image" src="https://github.com/user-attachments/assets/79d1ed11-b47e-43fa-bf53-887c57d3a5b0" />

ISmiiukha:
I can resolve it

wackerow:
Thanks folks! Closed with #15962

PR #15962, opened 2025-07-29
Fix typo from 'Using Ethereum for web2 authentication'

jorgesumle:
## Description

Just a small text correction.

## Related Issue

#15940


wackerow (approved):
Thanks @jorgesumle!

PR #16010, opened 2025-08-03
feat: language dropdown enhancements

### LINKED ISSUE #15997: change request: language dropdown on find-wallet page enhancements
On https://ethereum.org/en/wallets/find-wallet/

there are 2 change requests in one component

A)
there is a language support dropdown in a sidebar. This contains a lot of options and currently user has to scroll only to find language that they want to pick.

**Change**: 

- [ ] include ability to type in for quicker and easier selection

B)
Current logic is sorting based on number of wallets which isnt suitable for user intent of finding own language, there

**Change**

- [ ] Sort languages alphabetically instead based on number of wallets

Image

JoeChenJ:
## Description

- [x] Makes the dropdown searchable
- [x] Sorts languages alphabetically

## Preview Link
https://deploy-preview-16010--ethereumorg.netlify.app/en/wallets/find-wallet/

## Related Issue
Closes #15997


src/components/FindWalletProductTable/FindWalletLanguageSelectInput.tsx
```diff
@@ -51,15 +54,36 @@ const FindWalletLanguageSelectInput = ({
...
+            const isVisible = language.name
+              .toLowerCase()
+              .includes(searchQuery.toLowerCase())
             return (
-              <SelectItem key={language.langCode} value={language.langCode}>
+              <SelectItem
+                key={language.langCode}
+                value={language.langCode}
+                className={!isVisible ? "hidden" : ""}
+              >
```
JoeChenJ, line 123: Using `hidden` here instead of filtering the array directly, to avoid `SelectValue` losing its value and in turn leads to `Input` losing focus when `searchQuery` changes

JoeChenJ:
Thanks for the feedback @wackerow!  Just pushed a commit for those issues.

I also noticed the same issues in the wallet dropdown on [layer-2/networks page](https://ethereum.org/en/layer-2/networks/). Should I fix that here as well or open another pr?

wackerow:
Hey @JoeChenJ! Sorry for delay, yes please =) A fix for that would be great, but to your point, let's do that in a separate PR.

This looks better, and it's auto-focusing to the input when the menu is clicked... perhaps it's a "nice-to-have", but I'm not able to keyboard navigate with down/up arrows to select my desired option, but it DOES allow me to press "enter" to select the top result when filtering.

I don't see that as a blocker, but will ask @konopkja to chime in if there is any preference

wackerow (approved):

wackerow:
@all-contributors please add @JoeChenJ for code

PR #16024, opened 2025-08-05
Updated info on testnet naming convention

bskrksyp9:
## Description

Ethereum testnets often follow a tradition of being named after metro stations or easily recognizable places. The goal is to make them memorable and distinguishable from mainnet. This naming pattern reflects community culture and developer convenience. These insights are drawn from Ethereum Cat Herders (ECH) developer calls and other public sources.

### DIFF
--- a/public/content/developers/docs/networks/index.md
+++ b/public/content/developers/docs/networks/index.md
@@ -203,5 +203,29 @@ If a public Ethereum network is like the public internet, a consortium network i
 
 ## Further reading {#further-reading}
 
+### 🚉 Why Are Ethereum Testnets Named After Metro Stations?
+
+Many Ethereum testnets are named after real-world metro or train stations. This naming tradition started early and reflects the global cities where contributors have lived or worked. It's symbolic, memorable, and practical just like testnets are isolated from Ethereum mainnet, metro lines run separately from surface traffic.
+
+#### 🚧 Commonly Used and Legacy Testnets
+
+- **Sepolia** — A metro-linked neighborhood in Athens, Greece. Currently used for smart contract and dApp testing.
+- **Hoodi** — Named after Hoodi metro station in Bengaluru, India. Used for validator and protocol upgrade testing.
+- **Goerli** — Named after Görlitzer Bahnhof in Berlin, Germany. Deprecated in 2024.
+- **Rinkeby** — Named after a Stockholm suburb with a metro station. Deprecated.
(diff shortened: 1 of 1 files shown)


wackerow (changes requested):
Hey @bskrksyp9, thanks for this! Learned a few things here =)

Couple things... For this page specifically I wonder if it's worth including testnets that are no longer active, which would be all of them aside from Sepolia, Hoodi and Ephemery. 

This feels kind of like a historical fun-fact, akin to where we explain the naming behind upgrades, [on our history page](https://ethereum.org/history) ("Why do some upgrades have multiple names?").

Curious opinions here, but not necessarily against adding to where you placed it, though at the least I would move this outside (above) the "Further reading" section which follows a pattern on all pages with just links.

---

If we keep this here, would decrement the heading levels and switch the headers to use "Sentence casing" instead of "Title Casing", amd add custom header IDs... ie:

```md
### 🚉 Why Are Ethereum Testnets Named After Metro Stations?
```

Should be

```md
## 🚉 Why are Ethereum testnets named after metro stations? {#why-naming}
```

with similar pattern to the others.

Let me know your thoughts... cc: @konopkja also curious your take on this

bskrksyp9:
Appreciate for your thoughtful feedback 

I see your point about inactive testnets fitting more as historical context, and the notes on placement and heading style make sense. 

I’d just love if we can keep this story somewhere on the site, even if not right here, since it adds background and character worth preserving.

wackerow:
@bskrksyp9 Yeah sounds good... as next steps let's go ahead with keeping it on this page, with the adjustments requested above.

I believe there is some discussion around expanding the "history" part of the site between a "timeline" like we have now at [/history](https://ethereum.org/history) and a "story" approach that is less bulleted. This could potentially be modified then when those changes are being planned/implemented.

wackerow:
Hey @bskrksyp9 just circling back—are you still interested in making these updates? Let us know if you don't have the bandwidth 🙏

bskrksyp9:
> Hey @bskrksyp9 just circling back—are you still interested in making these updates? Let us know if you don't have the bandwidth 🙏 

Hi! we had agreed to keep this along with your suggestion am good with that..
i don't think i can merge it myself..

bskrksyp9:
@wackerow 

> as next steps let's go ahead with keeping it on this page, **with the adjustments requested above.**

Got it, I overlooked this and thought the changes were already committed. I did the changes as requested.

wackerow:
@all-contributors please add @bskrksyp9 for content

PR #16034, opened 2025-08-08
docs: removed dead EtherNow links

mdqst:
## Description

deleted all links to https://www.ethernow.xyz/ because the service is shut down ([details](https://www.blocknative.com/ethernow-sunset)).


wackerow:
@all-contributors please add @mdqst for maintenance

PR #16032, opened 2025-08-08
feat: ContentHero design update

wackerow:
## Description
Updates ContentHero according to new /what-is-ethereum designs

- Removes bg gradient, enlarged font size and increase font weight
- Updated image handling to be more robust with width/height properties
- Aside: switched to png for translatathon hero to comply; about the same size for a 3x version, and offers blurred loading state with sizing dimensions

### Layer 2 pages
- https://deploy-preview-16032--ethereumorg.netlify.app/en/layer-2/learn
- https://deploy-preview-16032--ethereumorg.netlify.app/en/layer-2/networks

### Roadmap layout
- https://deploy-preview-16032--ethereumorg.netlify.app/en/roadmap/future-proofing
- https://deploy-preview-16032--ethereumorg.netlify.app/en/roadmap/scaling
- https://deploy-preview-16032--ethereumorg.netlify.app/en/roadmap/security
- https://deploy-preview-16032--ethereumorg.netlify.app/en/roadmap/user-experience

### Staking layout
- https://deploy-preview-16032--ethereumorg.netlify.app/en/staking/solo
- https://deploy-preview-16032--ethereumorg.netlify.app/en/staking/pools
- https://deploy-preview-16032--ethereumorg.netlify.app/en/staking/saas
- https://deploy-preview-16032--ethereumorg.netlify.app/en/staking/withdrawals

### Translatathon layout
- https://deploy-preview-16032--ethereumorg.netlify.app/en/contributing/translation-program/translatathon
- https://deploy-preview-16032--ethereumorg.netlify.app/en/contributing/translation-program/translatathon/details
- https://deploy-preview-16032--ethereumorg.netlify.app/en/contributing/translation-program/translatathon/terms-and-conditions

### Upgrade layout
- https://deploy-preview-16032--ethereumorg.netlify.app/en/roadmap/beacon-chain
- https://deploy-preview-16032--ethereumorg.netlify.app/en/roadmap/merge

### Use cases layout
- https://deploy-preview-16032--ethereumorg.netlify.app/en/ai-agents
- https://deploy-preview-16032--ethereumorg.netlify.app/en/dao
- https://deploy-preview-16032--ethereumorg.netlify.app/en/decentralized-identify
- https://deploy-preview-16032--ethereumorg.netlify.app/en/defi
- https://deploy-preview-16032--ethereumorg.netlify.app/en/desci
- https://deploy-preview-16032--ethereumorg.netlify.app/en/nft
[...]

PR #16042, opened 2025-08-10
feat: /what-is-ethereum revamp

wackerow:
## Description
Revamps /what-is-ethereum page per latest designs

### Existing components
- ContentHero: #16032 updates the ContentHero to match the design presenting for this page revamp. This PR extends those changes—if this is not desired (or slowing anything down), ping me and I can remove from this PR so it doesn't block.
- TableOfContents (ToC): used existing, applied classes to custom style in this case to match design -- Separate task: Incorporate this set of styling either as the default for the existing ToC, or add it as a simple `variant` that can be applied when using it.
- FileContributors: Un-touched; not currently matching design. Suggesting this as separate task after confirm how we want to handle this component on all pages.
- Quiz: 3 of 5 existing questions still match new content perfectly fine... two others need adjusting. Not currently in this PR, but should be a separate task to get this added back to the bottom of the page.

### TODOs
- [x] Apply correct `sizes` to all images
- [x] FeedbackWidget added to bottom
- [x] Extract strings for intl
- [x] Add custom tracking events
- [x] Fill in remaining / confirm all links are populated and working

## Preview link
https://deploy-preview-16042--ethereumorg.netlify.app/en/what-is-ethereum


konopkja (approved):

wackerow:
Thanks @konopkja! cc: @pettinarip / @corwintines for code review

@konopkja If you could, take a peak at #16032 as well (ContentHero). This PR builds on top of that one, although it doesn't have to in case we aren't ready to proceed with that update (which affects many pages)

app/[locale]/what-is-ethereum/page.tsx
```diff
@@ -172,578 +117,1017 @@ const Page = async ({ params }: { params: Promise<{ locale: Lang }> }) => {
...
+            className="!py-0 [&>div]:mt-0"
+            contributors={contributors}
+            lastEditLocaleTimestamp={lastEditLocaleTimestamp}
+          />
+
+          <ListenToPlayer slug="what-is-ethereum" />
+        </div>
+
+        <div
+          className={cn(
```
Comment (pettinarip) [inline app/[locale]/what-is-ethereum/page.tsx]: why? can't we handle all these styles inside the ToC component to reuse them later? this is pretty hard to read.

app/[locale]/what-is-ethereum/page.tsx
```diff
@@ -172,578 +117,1017 @@ const Page = async ({ params }: { params: Promise<{ locale: Lang }> }) => {
...
+            // `button` targets mobile version
+            "[&_button>span]:flex-none [&_button]:mb-16 [&_button]:justify-center [&_button]:rounded-lg [&_button]:border-border [&_button]:bg-accent-a/10 [&_button]:text-lg [&_button]:font-bold"
+          )}
+        >
+          <TableOfContents items={tocItems} isMobile className="lg:hidden" />
+          <TableOfContents items={tocItems} className="max-lg:hidden" />
+        </div>
+
+        <div
+          data-label="content"
```
Comment (pettinarip) [inline app/[locale]/what-is-ethereum/page.tsx]: curious, what is this for?

app/[locale]/what-is-ethereum/page.tsx
```diff
@@ -1,168 +1,113 @@
...
+      </span>
+      &nbsp;
+      <span className="group-hover:underline">{children}</span>
+    </Link>
+  )
+}
 
-const loadData = dataLoader([["growThePieData", fetchGrowThePie]])
+const Strong = (chunks) => <strong>{chunks}</strong>
+const Emphasis = (chunks) => <em>{chunks}</em>
```
Comment (pettinarip) [inline app/[locale]/what-is-ethereum/page.tsx]: we should create a separate files for these basic reusable components for translation strings

app/[locale]/what-is-ethereum/page.tsx
```diff
@@ -1,168 +1,113 @@
...
+const LinkWithArrow = async ({ href, className, children }: LinkProps) => {
+  const locale = await getLocale()
+  const { twFlipForRtl } = getDirection(locale as Lang)
+  return (
+    <Link
+      href={href}
+      className={cn("group block w-fit no-underline", className)}
+    >
+      <span data-label="link-arrow" className={twFlipForRtl}>
+        →
```
Comment (pettinarip) [inline app/[locale]/what-is-ethereum/page.tsx]: replace this with an icon to keep it consistent across users. This is how I see this which IMO it is misaligned and pretty small
image

app/[locale]/what-is-ethereum/page.tsx
Comment (wackerow) [reply app/[locale]/what-is-ethereum/page.tsx]: Yeah, happy to do that... this initial approach was mainly to avoid altering the ToC until we had consensus on moving forward with this as a DS update. cc: @konopkja IIRC, you're suggesting we have three variations for the Toc?
- Styling used on use-cases template (left side, single depth)
- Styling used in the docs (right side, variable depth)
- And this as a new third variant, which we'll start off by using on this page?

Can make these adjustments early next week

app/[locale]/what-is-ethereum/page.tsx
Comment (wackerow) [reply app/[locale]/what-is-ethereum/page.tsx]: Nada.. can remove; was only to help keep me organized while building but we're not using it as a selector

app/[locale]/what-is-ethereum/page.tsx
Comment (wackerow) [reply app/[locale]/what-is-ethereum/page.tsx]: Agree... cc: @konopkja... had kept to align with Figma and it's slightly lighter-weight, but less consistent amongst browsers

app/[locale]/what-is-ethereum/page.tsx
Comment (pettinarip) [reply app/[locale]/what-is-ethereum/page.tsx]: Sounds good

Comment (wackerow) [reply app/[locale]/what-is-ethereum/page.tsx]: Update: This styling was refactored to a TableOfContents variant named `beginner` in this PR. PR #16121 takes this further to deprecate `LeftNavBar` and incorporate that styling into TableOfContents as well as a new variant `left`, and renames the default variant used in the docs to `docs`.

corwintines (approved):

PR #16043, opened 2025-08-11
Apps revalidate

corwintines:
- Adds revalidate cron job for new apps routes


src/crons/revalidate-apps.ts
corwintines, line 1: Main changes here, the rest was merging dev and the dapps-revamp branches into this one

PR #16050, opened 2025-08-11
No locale prefix for default locale

pettinarip:
Intent to remove unnecessary redirect for most of the visitors.

## Description

Configures next-intl routing to use "as-needed" locale prefix.

## SEO Migration TODO
- [x]  Redirect `/en/*` → `/*`
- [x]  Update `hreflang`s — English should now point to `/` (not `/en`)
- [x]  Update canonical tags
- [x]  Update internal links
- [x]  Update XML Sitemap
    - Remove `/en` URLs.
    - Add `/` URLs for English.
- [ ] Resubmit sitemap in Google Search Console.

## Summary by CodeRabbit

- New Features
  - Cleaner URLs for the default language: removes the default locale prefix from links and sitemap entries (e.g., “/en” no longer shown for English).
  - Localized URLs now appear only when needed; default-language pages use unprefixed paths.

- Refactor
  - Improved routing configuration to better handle conditional locale prefixes.
  - Standardized URL generation to consistently omit the default locale while preserving non-default locales.
  - Harmonized analytics tracking to use normalized, locale-aware paths for more accurate reporting.

### DIFF
--- a/src/lib/utils/matomo.ts
+++ b/src/lib/utils/matomo.ts
@@ -12,9 +12,8 @@ export const MATOMO_LS_KEY = "ethereum-org.matomo-opt-out"
  * but we want to track them as /en paths for analytics consistency.
  */
 export const normalizePathForMatomo = (pathname: string): string => {
-  // If already has a locale prefix (like /es/, /fr/), keep it as is
-  const hasLocalePrefix = LOCALES_CODES.some(
-    (locale) => locale !== DEFAULT_LOCALE && pathname.startsWith(`/${locale}/`)
+  const hasLocalePrefix = LOCALES_CODES.some((locale) =>
+    pathname.startsWith(`/${locale}/`)
   )
 
   if (hasLocalePrefix) {
@@ -22,7 +21,7 @@ export const normalizePathForMatomo = (pathname: string): string => {
   }
 
   // For paths without locale prefix (English content), add /en prefix
-  return pathname === "/" ? "/en" : `/en${pathname}`
+  return `/${DEFAULT_LOCALE}${pathname}`
 }
 
 export interface MatomoEventOptions {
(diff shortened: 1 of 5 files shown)


src/lib/utils/matomo.ts
```diff
@@ -1,9 +1,30 @@
...
+  const hasLocalePrefix = LOCALES_CODES.some(
+    (locale) => locale !== DEFAULT_LOCALE && pathname.startsWith(`/${locale}/`)
+  )
+
+  if (hasLocalePrefix) {
+    return pathname
+  }
+
+  // For paths without locale prefix (English content), add /en prefix
+  return pathname === "/" ? "/en" : `/en${pathname}`
```
wackerow: @pettinarip In current production, when I log `pathname` for the homepage it prints `/en/` with the trailing slash (which is then logged to Matomo)... is there a reason we're forcing `/` => `/en` without the trailing slash here?

I think we can just do this:

```suggestion
  return `/${DEFAULT_LOCALE}${pathname}`
```

So `/` would return `/en/`, or `/some-path` would return `/en/some-path`

pettinarip: True. On both things, good catch, will update

minimalsm:
@coderabbitai do your thing and review

src/lib/utils/matomo.ts
```diff
@@ -1,9 +1,29 @@
...
+    pathname.startsWith(`/${locale}/`)
+  )
+
+  if (hasLocalePrefix) {
+    return pathname
+  }
+
+  // For paths without locale prefix (English content), add /en prefix
+  return `/${DEFAULT_LOCALE}${pathname}`
+}
```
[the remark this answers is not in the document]
wackerow: `pathname` is what it is, I don't think this matters... the helper function here just slips in the English prefix as-needed

PR #16057, opened 2025-08-12
setup /roadmap for translation

corwintines:
- Sets /roadmap page up for translation

## Preview link
https://deploy-preview-16057--ethereumorg.netlify.app/en/roadmap


src/data/roadmap/releases.tsx
```diff
@@ -197,12 +167,22 @@ export const releasesData: Release[] = [
...
   },
 ]
+
+// Legacy export for backward compatibility - uses hardcoded English strings
+export const releasesData: Release[] = getReleasesData((key: string) => {
+  // This is a fallback that returns the key itself if translations aren't available
+  // In practice, this should not be used in the actual app
+  console.warn(`Translation key ${key} used without translation function`)
+  return key
+})
```
wackerow, line 188: Do we need this?

app/[locale]/roadmap/_components/ReleaseCarousel.tsx
```diff
@@ -19,10 +18,15 @@ import {
...
-import { Release, releasesData } from "@/data/roadmap/releases"
+import { getReleasesData, Release } from "@/data/roadmap/releases"
+
+import { useTranslation } from "@/hooks/useTranslation"
 
 const ReleaseCarousel = () => {
   const locale = useLocale()
+  const { t } = useTranslation("page-roadmap")
+
+  const releasesData = useMemo(() => getReleasesData(t), [t])
```
Comment (wackerow) [inline app/[locale]/roadmap/_components/ReleaseCarousel.tsx:29]: Potentially out-of-scope for these changes, but our "data" file seems more like a hook... we could consider refactoring `getReleasesData` to a hook, with the other hooks embedded

ie create `app/[locale]/roadmap/hooks/useReleases.tsx`, which could contain everything in `src/data/roadmap/releases.tsx`, as well as the other client-side hooks here, returning everything we need:

```tsx
const {
  releasesData,
  startIndex,
  setApi1,
  setApi2,
  getStatus,
  getDisplayDate,
  currentIndex,
  t
} = useReleases()
```

wackerow:
Noticed our use of `htmr` here was bypassing our native Link component logic, rendering simple `a` elements which didn't include the external link arrow, and they weren't opening in new tabs... Pushed a commit to refactor using `Translation` component instead.

Also noticed the glossary tooltip links were not properly render the term/definition strings in the popup so I converted to a simple Link for now—can further debug that separately... Letting it build then will pull this in.

PR #16059, opened 2025-08-12
Fix sample contract of RPC API

ilanolkies:
## Description

Using the name of the contract to declare the constructor is not valid anymore.


wackerow:
@all-contributors please add @ilanolkies for maintenance

PR #16107, opened 2025-08-21
Update /developers/docs/nodes-and-clients/node-architecture/

mansurmasama:
There are essentially two types of clients: execution and consensus. A separate “validator client” doesn’t exist. If a node wants to perform validation, it simply needs to run both the execution and consensus clients—there’s no need to set up an additional validator client.

### DIFF
--- a/public/content/developers/docs/nodes-and-clients/node-architecture/index.md
+++ b/public/content/developers/docs/nodes-and-clients/node-architecture/index.md
@@ -4,7 +4,7 @@ description: Introduction to how Ethereum nodes are organized.
 lang: en
 ---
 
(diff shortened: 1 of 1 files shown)


wackerow:
Thanks @mansurmasama, would push back on this though. Despite these being bundled, the validator client is commonly an optional separate process from just the beacon node:

- https://prysm.offchainlabs.com/docs/backup-and-migration/database-backups/#validator-client
- https://nimbus.guide/validator-client.html (optional to run with beacon node, or can run separately)
- https://lighthouse-book.sigmaprime.io/help_vc.html
- https://chainsafe.github.io/lodestar/run/validator-management/vc-configuration
- https://docs.teku.consensys.io/get-started/start-teku#start-the-validator (optional to run with beacon node, or can run separately)

Grandine (https://docs.grandine.io/validator_client.html) does appears to have this baked in as the only option, enabled by just passing signing keys, though it's the only client I don't have personal experience with.

---

Going to close out, but if you'd like can re-open and offer a suggestion to adjust the copy so this is more clear. Thanks!

PR #16111, opened 2025-08-22
Optimize mobile nav

pettinarip:
## Description

- nav — code refactors & reorganization
- menu — ssr content
- menu — replace accordions with collapsibles (collapsible is a primitive of the accordion component) to improve performance with nested trees
- menu — remove Sheet overlay and hamburger animation to reduce main thread usage
- language picker — pre-calculate `languages` on the server
- language picker — pre-calculate `progress` on the server
- language picker — open content inside the existing Sheet container instead of creating a new Dialog


src/components/ClientOnly.tsx
```diff
@@ -0,0 +1,19 @@
+"use client"
+
+import { type ReactNode } from "react"
+
+import { useIsClient } from "@/hooks/useIsClient"
+
+type ClientOnlyProps = {
+  children: ReactNode
+  fallback?: ReactNode
```
wackerow, line 9: Shouldn't block, but perhaps we rename this to `loading` for consistency with other loading components?

wackerow (dismissed):
This is looking great 🔥, well done @pettinarip! Will be curious to see the impact here

src/components/ClientOnly.tsx
pettinarip: Hmm, but the intention of `fallback` is that it can be anything (including `null`). On the other hand, `loading` narrows the scope, leading developers to think they should only use loading UIs, which isn’t correct.

Issue #13467, opened 2024-07-24
Suggest a wallet: Ctrl Wallet (fka XDEFI Wallet)


sean-mc-grath:
### Wallet name

Ctrl Wallet

### What type of wallet?

EOA

### Wallet description

One wallet for all your assets. Multichain wallet that displays all assets and NFTs on 200+ networks. 

### Wallet logo

![App Icon Black](https://github.com/user-attachments/assets/d3e0733c-6ce6-45ea-a70c-ad393e75ee83)

### Background color for brand logo

#FFFFFF

### URL to the project

https://ctrl.xyz/

### When did the wallet go live to users?

XDEFI has been live since 2020, Ctrl rebrand will be live in September 2024

### What are the fees users may be paying for using your wallet product?

No fees on general usage or transactions. 
0.3% fee on in-app swaps & bridges, with discounts for token holders.

### Does the wallet have an active development team?

Yes. https://github.com/XDeFi-tech/ctrl-extension/

### Is the wallet globally accessible?

Globally accessible everywhere crypto is legal.

### Is the wallet available in multiple languages?

Currently available in English with more languages coming soon.

### What social links are there for the project?

https://discord.gg/ctrlwallet
https://x.com/ctrl_wallet
https://t.me/ctrl_wallet

### Does the wallet have a mobile app? If yes, which operating systems are supported (iOS, Android)?

Yes, iOS & Android

### Does the wallet have a desktop app? If yes, which operating systems are supported (Windows, Mac, Linux)?

N/A

### Does the wallet have a browser extension? If yes, which browsers are supported (Chromium, Firefox, Safari)?

Yes, Chromium

### Is it a hardware wallet?

No

### What is the availability of your source code?

Open source

### What license is the wallet software released under?

This project is licensed under the Apache-2.0 License.

### Who holds the private keys?

Ctrl is a self-custody wallet, we never have access to users private keys. Users have full control over their keys.

### Please describe the measures taken to ensure the wallet's security and provide documentation wherever possible

Security audit by FYEO currently underway.
Previously [audited by Kudelski Security](https://www.xdefi.io/wp-content/uploads/2021/07/XDEFI_Audit_Report_Kudelski.pdf).

### Has the wallet's smart contract code or security modules been audited?

https://www.xdefi.io/wp-content/uploads/2021/07/XDEFI_Audit_Report_Kudelski.pdf

### Does the wallet have an internal security team?

No

### Any other security testing that should be noted?

N/A

[...]

PR #16113, opened 2025-08-22
fix: Update Ctrl Wallet's properties

Savio-Sou:
## Description

This PR update a few properties of the Ctrl Wallet based on my personal experience:
- `ios` and `android` to true
  - Mobile apps are available on App Store and Play Store, which the official site also links to (albeit only if you access the site from a mobile device)
- `withdraw_crypto` to false
  - I couldn't find a way to offramp into fiat in the wallet
- `social_recovery` to false
  - I couldn't find a way to set up social recovery in the wallet
  - https://github.com/ethereum/ethereum-org-website/issues/13467 might have misinterpreted social recovery as being able to login with social accounts (e.g. X)


wackerow:
@all-contributors please add @Savio-Sou for maintenance

PR #16121, opened 2025-08-24
refactor: TableOfContents variants; deprecate LeftNavBar

wackerow:
Extends `what-is-ethereum` #16042

## Description
- Update `TableOfContents` variants to include `left` variant, matching style of `LeftNavBar`
- Deprecate `LeftNavBar` replacing with ToC `variant="left"`


wackerow:
Also refactored translatathon page to use TableOfContents component, updating to add a conditional hiding the "Apply" section if the deadline has passed. Added a DRY date utility helper to return a boolean for when a date has been reached.

corwintines (approved):

Issue #16133, opened 2025-08-26
Suggest a dapp


ButlerATTS:
### Name

Alchemix

### Dapp URL

https://alchemix.fi

### Description

Alchemix allows users to deposit ETH or stablecoins and take loans, in the form of synthetic credit, which automatically pay themselves off with the yield generated from their deposit, without having to worry about liquidation or loan repayment. Unlock the potential of your assets with secure and stress-free borrowing

### Logo

<img width="444" height="442" alt="AlchemixLogo" src="https://github.com/user-attachments/assets/fbd7d2c1-c105-43e6-8403-630400174957" />

### Brand color

#f5c59f

### Is your dapp open source?

Yes

### If open source, what is the repo for the dapp?

https://github.com/alchemix-finance/v2-foundry

### Please describe the measures taken to ensure the dapp's security and provide documentation wherever possible

Alchemix v2 was audited by Runtime Verification. Alchemix continues to engage Runtime Verification for additional audits on protocol changes. The v2 audit report can be found here: https://github.com/runtimeverification/publications/blob/main/reports/smart-contracts/Alchemix_v2.pdf​

Alchemix offers a bug bounty program through ImmuneFI. The program can be found here: https://immunefi.com/bounty/alchemix/​

Alchemix ran a one-off code4rena contest. The contest can be found here: https://code4rena.com/contests/2022-05-alchemix-contest

### Can a user access the dapp without creating an account?

Yes

### Are dapp funds non-custodial?

Not entirely. If the team disappears, users can still access and move their funds, however the team maintains the ability to pause contract functions and migrate funds for upgrades. Our team has always operated transparently, only pausing contracts while investigating potential exploits or to protect users, and only migrating for upgrades with many months of advanced notice and communications across all of our socials

### Do you offer community support?

Yes we have a discord support channel where users can post support threads, and a team that answers them and helps users with their issues

### Describe any geographical restrictions on usage of your dapp

The app is globally accessible. No restrictions on usage

### Would you like to work on this issue?

- [x] Yes
- [ ] No

wackerow:
#16133 attempted to add, but this would've only added to the "mock data"... the submission has been added to the backend list and is now live: https://ethereum.org/en/apps/alchemix/

Thanks @ButlerATTS!

PR #16134, opened 2025-08-26
Addresses #16133 - Request to add Alchemix as a listed dApp

ButlerATTS:
## Description
Adds an Alchemix object to the list of DeFi apps

## Related Issue
https://github.com/ethereum/ethereum-org-website/issues/16133


nloureiro:
Thank you for your submission and the app is live here
https://ethereum.org/en/apps/alchemix/

PR #16142, opened 2025-08-27
patch: use fallback/original variant for non-EN A/B tests

wackerow:
## Description
- Within `ABTestWrapper` component, check locale—if non-English, return fallback/original variant
- Adds `enableAllLocales` boolean prop to override and allow test component to render for all locales

## Related Issue
A/B test cases to use English by default—should avoid rendering English test cases on non-English locales


wackerow:
cc: @konopkja Tagging you here, just fyi... By default this will only render test cases for the English locale, but if we have a case that needs to display on all locales, we can override that as-needed.

PR #16157, opened 2025-08-30
refactor: TabNav; deprecate CategoriesNav, ResourcesNav 

wackerow:
## Description
- Refactor reusable `TabNav` component
- Deprecate `CategoriesNav` and `ResourcesNav` components, replaced with `TabNav`
- Adds `StickyContainer` export from `TabNav.tsx` as a wrapper when we desire this nav bar to by sticky
- Maintains parity with existing production for now (`/apps/categories/[category]` version not sticky, primary colored icon, and does not use `useMotion`)


wackerow:
Converting to draft while re-assessing the logic here between usage as anchor links vs JS tabs

corwintines (approved):

PR #16166, opened 2025-09-01
Clean up theme

pettinarip:
Remove unused vars from theme and format `semantic-tokens.css` file

### DIFF
--- a/src/styles/semantic-tokens.css
+++ b/src/styles/semantic-tokens.css
@@ -92,19 +92,25 @@
       #000000 69.77%
     );
 
-    --card-gradient: linear-gradient(123deg, rgba(255, 255, 255, 0.20) 58.99%, rgba(174, 110, 203, 0.13) 104.04%);
-    --card-gradient-secondary: linear-gradient(95deg, rgba(211, 145, 242, 0.12) 0%, rgba(159, 43, 212, 0.12) 102.78%);
-    --card-gradient-secondary-hover: linear-gradient(95deg, rgba(211, 145, 242, 0.2) 0%, rgba(159, 43, 212, 0.2) 102.78%);
-    --ten-year-gradient: linear-gradient(100deg, #F6C9EA 55.38%, #C7A9F1 110.54%);
-
-    /* Rainbow gradients */
-    --rainbow-gradient: conic-gradient(
-      from 0deg,
-      hsla(var(--primary)),
-      hsla(var(--accent-b)),
-      hsla(var(--accent-a)),
-      hsla(var(--accent-c)),
-      hsla(var(--primary))
+    --card-gradient: linear-gradient(
+      123deg,
+      rgba(255, 255, 255, 0.2) 58.99%,
+      rgba(174, 110, 203, 0.13) 104.04%
+    );
+    --card-gradient-secondary: linear-gradient(
+      95deg,
+      rgba(211, 145, 242, 0.12) 0%,
+      rgba(159, 43, 212, 0.12) 102.78%
+    );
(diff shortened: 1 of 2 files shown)


src/styles/semantic-tokens.css
```diff
@@ -92,19 +92,25 @@
...
+    );
+    --card-gradient-secondary-hover: linear-gradient(
+      95deg,
+      rgba(211, 145, 242, 0.2) 0%,
+      rgba(159, 43, 212, 0.2) 102.78%
+    );
+    --ten-year-gradient: linear-gradient(
+      100deg,
+      #f6c9ea 55.38%,
+      #c7a9f1 110.54%
```
wackerow, line 113: Per in-file comment:

```css
    /**
    /* Gradients (radial, conic, or linear over 3 stops)
    /* For linear-gradient up to 3 stops, use Tailwind classes
    /* https://tailwindcss.com/docs/background-image#linear-gradients
    */
```

Noting that these could be simple tailwind classes instead, though if we want to nit-pick on non-standard angles (123deg, 95deg, 100deg), that would have to wait for tailwind v4, where we could do `bg-linear-123` as an example.

Otherwise we could standardize the angles with something like:

`bg-gradient-to-br from-white/20 from-[59%] to-[rgb(174,110,203)]/[0.13] to-[104%]` in situ

or

```css
--card-gradient: @apply bg-gradient-to-br from-white/20 from-[59%] to-[rgb(174,110,203)]/[0.13] to-[104%];
```


@pettinarip Not sure if you have a preference... not a huge deal, but given we're using tailwind I try to stick to classes where possible

src/styles/semantic-tokens.css
```diff
@@ -92,19 +92,25 @@
...
     );
```
wackerow, line 114:
```suggestion
    --card-gradient: @apply bg-gradient-to-br from-white/20 from-[59%] to-[rgb(174,110,203)]/[0.13];
    --card-gradient-secondary: @apply bg-gradient-to-r from-[rgb(211,145,242)]/[0.12] to-[rgb(159,43,212)]/[0.12];
    --card-gradient-secondary-hover: @apply bg-gradient-to-r from-[rgb(211,145,242)]/20 to-[rgb(159,43,212)]/20;
    --ten-year-gradient: @apply bg-gradient-to-r from-[#f6c9ea] from-[55%] to-[#c7a9f1] to-[110%];
```

wackerow: Hm actually, these aren't within a custom class, so that may not work right... but:

1. `bg-card-gradient` is only used in one place that I can find: `app/[locale]/trillion-dollar-security/page.tsx`
We could replace that usage with `bg-gradient-to-br from-white/20 from-[59%] to-[rgb(174,110,203)]/[0.13]` and remove this custom variable from the theme?
2. `bg-ten-year-gradient` is also only used once, in `app/[locale]/10years/_components/TenYearHomeBanner.tsx` attached to `after:`
We could replace that usage with `after:bg-gradient-to-r after:from-[#f6c9ea] after:from-[55%] after:to-[#c7a9f1] after:to-[110%] dark:after:from-[#1e151b] dark:after:from-[55%] dark:after:to-black dark:after:to-[110%]`, and similarly remove from the theme.
3. `bg-card-gradient-secondary` (4 uses) and `bg-card-gradient-secondary-hover` (2 uses)... could consider making custom classes for these two outside of `semantic-tokens.css`?

@pettinarip Just some thoughts on how to iterate on this, but for the sake of not over-complicating, going to accept and pull in what we have here.

PR #16186, opened 2025-09-05
Add schema.org and JSONLD to pages

corwintines:
## Description

Can test schemas on https://search.google.com/test/rich-results

https://technicalseo.com/tools/google-serp-simulator/

More info on structured data: https://developers.google.com/search/docs/appearance/structured-data/search-gallery

More info on Schema.org: https://schema.org/

Info on JSON-LD: https://nextjs.org/docs/app/guides/json-ld


app/[locale]/wallets/_components/wallets.tsx
Comment (pettinarip) [inline app/[locale]/wallets/_components/wallets.tsx:1]: orphan?

app/[locale]/page-jsonld.tsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  // JSON-LD for ethereum.org as an organization
+  const organizationJsonLd = {
+    "@context": "https://schema.org",
+    "@type": "Organization",
+    "@id": "https://ethereum.org",
+    name: "ethereum.org",
+    url: "https://ethereum.org",
+    logo: "https://ethereum.org/images/ethereum-logo.png",
```
Comment (pettinarip) [inline app/[locale]/page-jsonld.tsx]: This image doesn't exist. Maybe `https://ethereum.org/images/eth-org-logo.png`?

app/[locale]/eth/_components/page-jsonld.tsx
Comment (pettinarip) [inline app/[locale]/eth/_components/page-jsonld.tsx:1]: Noted that this one fails because this is used inside a client component. We need to use `useTranslations`.

pettinarip (commented):
@corwintines I've found two small issues. The rest looks great 👍🏼

app/[locale]/wallets/_components/wallets.tsx
Comment (corwintines) [reply app/[locale]/wallets/_components/wallets.tsx:1]: Not sure if this was deleted when you reviewed yet or not, but I refactored this to SSR in this PR for /wallets

app/[locale]/eth/_components/page-jsonld.tsx
Comment (corwintines) [reply app/[locale]/eth/_components/page-jsonld.tsx:1]: Ah going to just pull this out into the page.tsx file instead, doesn't need to be here.

app/[locale]/get-eth/page.tsx
Comment (pettinarip) [inline app/[locale]/get-eth/page.tsx:1]: There is a translation bug in this page
image

app/[locale]/10years/page-jsonld.tsx
```diff
@@ -0,0 +1,143 @@
...
+        height: "417",
+      },
+    },
+    reviewedBy: {
+      "@type": "Organization",
+      name: "ethereum.org",
+      url: "https://ethereum.org",
+      logo: {
+        "@type": "ImageObject",
+        url: "https://ethereum.org/favicon-32x32.png",
```
Comment (pettinarip) [inline app/[locale]/10years/page-jsonld.tsx]: This image doesn't exist. I think we can use `eth-home-icon.png` as a replacement since it is the same image used as a favicon.

PR #16191, opened 2025-09-05
chore: :memo: update docs Prysm domain to prysm.offchainlabs.com

victorelec14:
## Description

Updated all documentation URLs from https://docs.prylabs.network/docs to https://prysm.offchainlabs.com/docs and new urls paths.


corwintines (approved):
Thanks @victorelec14

PR #16224, opened 2025-09-11
[Update Ethereum for Dart developers] Package publisher changed

jorgesumle:
It looks like the package publisher changed.


wackerow:
@all-contributors please add @jorgesumle for maintenance

PR #16225, opened 2025-09-11
docs: fix outdated Hardhat repository link

famouswizard:
## Description

noticed that the Hardhat link in our docs/code was pointing to the old repository.
updated it to the official one at NomicFoundation to ensure references are correct.


wackerow:
@all-contributors please add @famouswizard for maintenance

PR #16232, opened 2025-09-11
docs: fix broken and outdated oracle links

barajeel:
## Description
I noticed one of the referenced links was returning a 404, so I removed it.
Also updated the Better Programming article to its current URL on Medium.


wackerow:
@all-contributors please add @barajeel for maintenance

PR #16252, opened 2025-09-13
docs: fix outdated links

mdqst:
## Description

I went through the docs and replaced broken/outdated `readthedocs.io` links with the current `docs.vyperlang.org` and `docs.soliditylang.org` equivalents.


wackerow:
@all-contributors please add @mdqst for maintenance

PR #16288, opened 2025-09-16
docs: update outdated Alchemy NFT API docs link

mdqst:
## Description

the old link to the NFT API docs was no longer valid.
updated it to the current reference page to avoid broken redirects and ensure developers can find the right documentation quickly.


wackerow (approved):
Thanks again @mdqst!

PR #16233, opened 2025-09-11
Find wallets optimizations

pettinarip:
## Description

This PR reworks the Find Wallets page to improve render performance and interaction responsiveness. It replaces the data table layout with a virtualized, accessible list; eliminates duplicated state that caused double renders; and reduces unnecessary component updates across filters and wallet rows.

### Key changes

- Virtualized list: replaced table with `@tanstack/react-virtual`.
- Filters architecture: `filters` now single source of truth; `presetFilters` derived locally; unified update path; memoized Filter rows; removed hidden desktop duplication.
- Wallet row perf: de-duplicated `WalletInfo` per device; reduced `ChainImages` layout shift; extracted `PersonaTags`.
- Cleanup: refactored ProductTable with children render prop to let consumers use a List or a DataTable to render rows.

## Preview link
https://deploy-preview-16233--ethereumorg.netlify.app/wallets/find-wallet

PR #16311, opened 2025-09-19
Patch findwallets native search

pettinarip:
## Description
With the recent introduction of the virtualized list (#16233), the native browser search (Ctrl+F) no longer works as expected. Items that are not currently visible are unmounted, so they cannot be found via the search.

This PR mitigates the issue by rendering “ghost” elements containing wallet names. This way, all wallet names remain discoverable with the native search. When matched, the user can scroll to the approximate position of the wallet. While not pixel-perfect, this at least ensures that users can locate wallets using the browser’s search functionality.

### TODO
If it works, move the ghost list to its own component and memo it to avoid perf impact.

### DIFF
--- a/src/components/FindWalletProductTable/index.tsx
+++ b/src/components/FindWalletProductTable/index.tsx
@@ -68,6 +68,7 @@ const FindWalletProductTable = ({ wallets }: { wallets: WalletRow[] }) => {
             )}
             filters={filters}
             matomoEventCategory="find-wallet"
+            estimateSize={300}
           />
         </>
       )}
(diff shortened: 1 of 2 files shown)


wackerow:
Hey @pettinarip, curious how you want to move forward with this... as we discussed, this doesn't perfectly target the location of the wallet, but it _does_ at least get spotted by a browser "Find" search, which is arguably better than now, since the user will at least know that the wallet is on the page somewhere. I'd be down to move forward as-is and iterate from here

pettinarip:
@wackerow yeah, thats fair. Not ideal but functionally better. I'll update the PR and get it ready.

pettinarip:
closing in favor of #18067

PR #16357, opened 2025-09-24
staking home -> stkaing homepage

corwintines:
staking home -> staking homepage


wackerow (approved):
Just trying to limit Crowdin challenges so limited to only en and pcm where we use the "home staking" terminology

PR #16396, opened 2025-09-30
Use 301 for default-locale redirects

pettinarip:
## Description

Converts 307→301 when pathname starts with `/${DEFAULT_LOCALE}`.

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -1,8 +1,28 @@
+import { NextRequest, NextResponse } from "next/server"
 import createMiddleware from "next-intl/middleware"
 
 import { routing } from "./src/i18n/routing"
+import { DEFAULT_LOCALE } from "./src/lib/constants"
 
-export default createMiddleware(routing)
+const handleI18nRouting = createMiddleware(routing)
+
+export default function middleware(request: NextRequest) {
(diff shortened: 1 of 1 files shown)


wackerow:
image

`/en` or `/en/` -> `/` showing `301 (Moved Permanently)` for me

image

`/es` -redirects-to-> `/es/` and shows as `308 (Permanent Redirect)`

This (`308`) passes the SEO ranking along, so I think this is fine (ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/308)

middleware.ts
```diff
@@ -1,8 +1,28 @@
+import { NextRequest, NextResponse } from "next/server"
 import createMiddleware from "next-intl/middleware"
 
 import { routing } from "./src/i18n/routing"
+import { DEFAULT_LOCALE } from "./src/lib/constants"
 
-export default createMiddleware(routing)
+const handleI18nRouting = createMiddleware(routing)
```
Kentomarine, line 7: // Example middleware for logging requests
function requestLogger(req, res, next) {
    console.log(`${req.method} ${req.url}`);
    next(); // Call the next middleware
}

PR #16398, opened 2025-09-30
test default locale redirect

corwintines:
test default locale 301 redirect in _redirects file


corwintines:
This fails so going to close it out.

PR #16427, opened 2025-10-02
Update package manager instructions in README.md

khalil7044:
Added a short note at the end of README.md for Ubuntu and Debian users who may need to install nodejs and npm before using yarn install


wackerow:
@all-contributors please add @khalil7044 for infra

PR #16442, opened 2025-10-06
Update link ds-implementation.md

NikolaiKryshnev:
Updated link in docs/ds-implementation.md to point to Chakra UI v2 components documentation.


wackerow (changes requested):
Oh thanks for highlighting this @NikolaiKryshnev... We no longer use Chakra in this repo, so glad you spotted this, but we wouldn't want to continue linking to Chakra docs or referencing. We use the [shadcn/ui](https://ui.shadcn.com/) approach now with all base ui components living inside src/components/ui. Mind updating to remove all references to Chakra, and referencing the shadcn approach instead?

NikolaiKryshnev:
@wackerow done

wackerow:
@all-contributors please add @NikolaiKryshnev for maintenance

PR #16443, opened 2025-10-06
Handle missing dateOfLaunch by displaying dash placeholder [Fixes #16305]

### LINKED ISSUE #16305: when no info temove the NaN
### Describe the bug

on the app explorer if a field is empty don´t display NaN

Image

### To reproduce

https://ethereum.org/apps/kiwi-news/

### Expected behavior

NaN is showing

### Screenshots

Image

_No response_

_No response_

### Additional context

_No response_

- [ ] Yes
- [ ] No

clacladev:
## Description

Many Apps in the App Explorer are missing the Founding date, but no check is done, so it's manipulated and displayed as `NaN`.

Apps where I can reproduce the issue:
https://ethereum.org/apps/eternal-ai/
https://ethereum.org/apps/virtuals/
https://ethereum.org/apps/human-passport/
https://ethereum.org/apps/zk-open-passport/
https://ethereum.org/apps/zk-open-passport/

## Solution

I am going to treat the missing Founding Data the same way Last Update Data is treated: displaying a `-` (dash) as a placeholder.

## Problem and Solution Demo

Before:
Screenshot 2025-10-06 at 15 51 55

After:
Screenshot 2025-10-06 at 15 57 46

## How to test

- Confirm the issue in production [https://ethereum.org/apps/huddle01/](https://ethereum.org/apps/huddle01/)
- Confirm the issue is fixed in preview [https://deploy-preview-16443--ethereumorg.netlify.app/apps/huddle01/](https://deploy-preview-16443--ethereumorg.netlify.app/apps/huddle01/)

## Related Issue

Issue: https://github.com/ethereum/ethereum-org-website/issues/16305#issuecomment-3368374814

### DIFF
--- a/app/[locale]/apps/[application]/page.tsx
+++ b/app/[locale]/apps/[application]/page.tsx
@@ -129,6 +129,11 @@ const Page = async ({
     return t("page-apps-days-ago", { days: diffInDays })
   }
 
+  const getDisplayYear = (dateString: string) => {
+    if (!isValidDate(dateString)) return "—"
+    return new Date(app.dateOfLaunch).getFullYear()
+  }
+
   const commitHistoryCache: CommitHistory = {}
   const { contributors } = await getAppPageContributorInfo(
     "apps/[application]",
@@ -326,11 +331,7 @@ const Page = async ({
                 <p className="text-sm text-body-medium">
                   {t("page-apps-info-founded")}
                 </p>
-                <p className="text-sm">
-                  {app.dateOfLaunch
-                    ? new Date(app.dateOfLaunch).getFullYear()
-                    : "—"}
-                </p>
+                <p className="text-sm">{getDisplayYear(app.dateOfLaunch)}</p>
               </div>
               <div>
                 <p className="text-sm text-body-medium">


app/[locale]/apps/[application]/page.tsx
```diff
@@ -327,7 +327,9 @@ const Page = async ({
                   {t("page-apps-info-founded")}
                 </p>
                 <p className="text-sm">
-                  {new Date(app.dateOfLaunch).getFullYear()}
+                  {app.dateOfLaunch
+                    ? new Date(app.dateOfLaunch).getFullYear()
+                    : "—"}
```
Comment (wackerow) [inline app/[locale]/apps/[application]/page.tsx]: This looks good overall... slight request though to make this a touch more robust:

We have an `isValidDate` helper function inside `src/lib/utils/date.ts` which would not only check if something is "falsy" like this solution, but also makes sure that any string/number we pass to the `Date` constructor will not evaluate to `NaN`.

This function is already being used within this file, so it's already imported and ready for use. Would suggest a pattern similar to the `getTimeAgo` function above on lines 120-121.

```tsx
  const getTimeAgo = (dateString: string) => {
    if (!dateString || !isValidDate(dateString)) return "—"
```

So we could add something like this immediately below that function:

```tsx
  const getDisplayYear = (dateString: string) => {
    if (!isValidDate(dateString)) return "—"
    return new Date(app.dateOfLaunch).getFullYear()
  }
```

and then update the lines here with:

```suggestion
                  {getDisplayYear(app.dateOfLaunch)}
```

(Side note, we wouldn't need the `!dateString ||` check shown in `getTimeAgo` since TypeScript constraints force that this must be a string, and an empty string passed to `isValidDate("")` will properly evaluate to `false`)

wackerow (changes requested):

app/[locale]/apps/[application]/page.tsx
Comment (clacladev) [reply app/[locale]/apps/[application]/page.tsx]: @wackerow you're absolutely right. I should have read better the code to see there was a utility function and a pattern already established.

Anyway, I implemented the suggested changes and it still works as intended. Thanks for the guidance, next time I try to waste less of your time ;)

Screenshot 2025-10-07 at 16 51 33

Comment (wackerow) [reply app/[locale]/apps/[application]/page.tsx]: Hah didn't waste anyones time, happy to help, thanks @clacladev!

wackerow:
@all-contributors please add @clacladev for bug fix

clacladev:
Happy to help @wackerow 
Let me know if I can help with more tickets!

PR #16458, opened 2025-10-07
fix: language picker fallback logic

wackerow:
## Description
When a language name is the same in both English and that language, logic treats this as if the translated form is not available for that locale, and falls back to the `Intl.DisplayName` JavaScript library. In certain cases this is undesired, such as with Tagalog, whose name is the same in English and Taglog, falls back to the JS library and results in "Filipino" (but we already have Filipino specifically).

- Adds a boolean `forceLocalName` to the `I18nLocale` type (adding intellisense documentation for future clarity) to to override this and use the locale name instead
- Adds `"language-tl": "Tagalog",` to both en/common.json and tl/common.json files

## Related Issue
image


wackerow:
Initial build timed out:

```
12:00:42 PM: Failed during stage 'building site': Command did not finish within the time limit
12:00:42 PM: Execution timed out after 1h30m0.454590714s
```

Re-triggered from Netlify -> _build_ completed successfully:
https://app.netlify.com/projects/ethereumorg/deploys/68e5a8f9a59cf2f355ac6f77

(Lighthouse task timed out on `/en/`)

```
6:28:23 PM: Generating Lighthouse reports. This may take a minute…
6:28:23 PM: Running Lighthouse on /en/ (1/7)
6:28:30 PM: Execution timed out after 1h30m0.521147112s
```

**Preview link live**:
https://deploy-preview-16458--ethereumorg.netlify.app/

corwintines (approved):

PR #16480, opened 2025-10-10
docs(tutorials): add 'How Ethereum Addresses Are Generated' tutorial

ndubuisi-ugwuja:
## Description

This pull request adds a new beginner-friendly tutorial titled “How Ethereum Addresses Are Generated.”
It explains step-by-step how Ethereum addresses are derived from private keys using elliptic curve cryptography (secp256k1) and Keccak-256 hashing.

The tutorial covers:

- What a private key is and why it’s important
- How a public key is mathematically derived from the private key using the secp256k1 curve
- How the Keccak-256 hash function is used to generate an Ethereum address
- Why only the last 40 hexadecimal characters of the hash are used as the final address

File added:
public/content/developers/tutorials/how-ethereum-addresses-are-generated/index.md

The goal of this tutorial is to help newcomers and technical learners understand the cryptographic foundations of Ethereum addresses in a clear and beginner-friendly way.

## Related Issue

N/A – this is a new tutorial contribution.

### DIFF
--- a/public/content/developers/tutorials/generating-ethereum-addresses/index.md
+++ b/public/content/developers/tutorials/generating-ethereum-addresses/index.md
@@ -1,0 +1,115 @@
+---
+title: How Ethereum Addresses Are Generated.
+description: A beginner-friendly explanation of how Ethereum addresses are generated from private keys using elliptic curve cryptography and Keccak-256 hashing.
+author: Ndubuisi Ugwuja
+tags: ["elliptic-curve", "secp256k1", "keccak-256", "wallet address"]
+skill: beginner
+lang: en
+published: 2025-10-10
+---
+
+# Where does an Ethereum address come from?
+An Ethereum address comes from a private key → turned into a public key using elliptic-curve math → hashed with Keccak-256 → take the last 20 bytes (40 hex characters) of that hash. Prepend 0x and you have the address.
+
+## What’s a private key?
+A private key is like a super-secret password, but much longer and much harder to guess — think of it as a 256-bit secret number written in hexadecimal. Instead of words you can remember, it’s a string of random bytes (often shown as 0x followed by 64 hex characters) that only you should know. In the world of cryptocurrencies, that secret number is what proves you own an account: when you want to send money or sign a message, you use the private key to create a digital signature that proves you authorized the action without ever revealing the secret itself. The private key is generated so that it’s practically impossible for anyone else to recreate it by luck; if someone gets your private key they can impersonate you and control your funds, so it must be kept private and stored safely.
+
+## How do we get a public key from the private key?
+We use elliptic curve cryptography (ECC). Ethereum uses a curve called secp256k1 (the same curve Bitcoin uses).
+
+The math idea (simple picture):
+- Think of a curve drawn on a plane. Pick a fixed point on that curve (call it G).
+
+- Multiply that fixed point by your secret number (the private key). Multiplication here means repeated addition on the curve (it’s special math, not normal multiplication).
+
+- The result is another point on the curve — that point is the public key.
+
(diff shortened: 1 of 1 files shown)


wackerow:
Thanks @ndubuisi-ugwuja! Will try to get some eyes on this over the next week or two

wackerow:
Hey @ndubuisi-ugwuja, appreciate the work here! Taking a look through, few initial thoughts...

My first take is that the vast majority of this doesn't really seem like a "tutorial." This feels much more like details that we would expect to see in the [Docs](https://ethereum.org/developers/docs)... Typically the tutorials are designated for follow-along guide that developers can walk through to actually _do_ something, and the only part that feels doable here is the "Simple code snippets you can paste and try JavaScript" section, which isn't really guided, but more just told to copy/paste, making assumptions that they know how to set up ethers.js, etc.

There is already a page dedicated to [Accounts on Ethereum](https://ethereum.org/developers/docs/accounts/)... that page introduces a handful of the concepts you're outlining here. Would be careful with potentially repeating the information there, but we could focus on how to expand from there.

IMO, using this to expand on the docs to include more detailed information about how derivations happen, for example from mnemonic > tree of private keys > public keys > address, would be informative, and fill a gap that currently seems to exist in the docs. I'm imagining something that helps explain the steps involved in [Ian Coleman's BIP-39 demo](https://iancoleman.io/bip39/). I would see this living at a new sub-page to the /developers/docs/ pages. I could also see this as a more intermediate-to-advanced topic, imagining some more explanation along the way about how this math works, akin to a page such as  this [Elliptic Curve](https://learnmeabitcoin.com/technical/cryptography/elliptic-curve/) explainer on learnmeabitcoin.com (just an idea for inspiration)

Alternatively, if we want to keep this as a "tutorial" then I would suggest we try trim down explanations that are already explained elsewhere (linking out where needed), and instead adding more details for how a developer could step through some tools to learn the details. To me this would include using libraries that demonstrate what's involved at each step of the derivation, so developers can get a firm grasp of how this works. This could also include the "easy path" of how to set up and use ethers.js to simply generate keys. 

Let me know your thoughts and how you'd like to proceed!

wackerow:
I guess my last thought after posting and re-reading you original post, if you do want to keep this as a tutorial for beginners, again, would just try to focus on clear and simply step-by-step instructions that would get someone started with the local tools needed, with some example inputs, expected outputs, and how they could use this while building. Try to harness the existing docs where possible for the non-guided parts.

joovhie:
Feedback on “How Ethereum Addresses Are Generated” Tutorial

Hi Wackerow,

Thank you so much for taking the time to review my submission and share
such thoughtful feedback. I completely understand your points — after
rereading the piece, I agree that it leans more toward a conceptual
explanation than a step-by-step tutorial.

I’d like to proceed by expanding it into a *Docs sub-page *like you
mentioned  - /developers/docs/ pages , focusing on how derivations happen
from mnemonic → seed → private key → public key → address. I believe this
direction will fit better alongside the existing “Accounts” documentation
and fill the gap you mentioned around derivation and address generation.

I’ll make sure the updated version aligns with the existing docs structure
and references relevant pages where needed.

Please let me know if that approach sounds good before I begin revising.

Best regards,
*Ndubuisi Ugwuja*
@ndubuisi-ugwuja <https://github.com/ndubuisi-ugwuja>

wackerow:
sgtm!... cc: @konopkja @mnelsonBT re: dev docs addition planning, if you have any thoughts to add here

@ndubuisi-ugwuja Would recommend starting with an outline posted as a [new issue](https://github.com/ethereum/ethereum-org-website/issues/new?template=feature_request.yaml).. ping us in there and we can take it from there! Going to close this PR out for now to keep things clean.. Thanks!

PR #16487, opened 2025-10-13
cleanup: remove unused loading state from GasPriceDisplay

eeemmmmmm:
## Description

Strip the unused loading flag so the component always renders the current gas level label


wackerow:
Ah, to remove this we'll need to adjust more than just this component though.. cc: @corwintines any reason you see to keep this, or can we proceed to take out the stuff that was used for the 10-year anniversary?

corwintines:
@wackerow down to just clean up the 10 year page in this PR and clean out whatever isn't in use anymore.

wackerow:
Going to close this one out for now, but we can set up a separate task to clean up the 10 year page, removing all the pieces that are no longer needed.. thanks folks!

PR #16498, opened 2025-10-15
Add alt text to centralized exchange card images

minimalsm:
## Summary
- add descriptive alt text for the centralized exchange cards so the logos announce properly

## Testing
- pnpm lint

------
https://chatgpt.com/codex/tasks/task_e_68f017a436e88322991f8dc11bb5b73f

### DIFF
--- a/src/hooks/useCentralizedExchanges.ts
+++ b/src/hooks/useCentralizedExchanges.ts
@@ -392,7 +392,7 @@ export const useCentralizedExchanges = () => {
             description,
             link: exchanges[exchange].url,
             image: exchanges[exchange].image,
-            alt: exchanges[exchange].name,
+            alt: t("common:item-logo", { item: exchanges[exchange].name }),
           }
         })
     )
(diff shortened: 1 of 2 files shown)


wackerow (commented):
If we want to do this right, we usually make the alt descriptive, such as "{name} logo"...

To do it with i18n in mind, we can create a template string and inject the item name similar to:

https://github.com/ethereum/ethereum-org-website/blob/e8b4b527fa42430dbfcad56e74be6d46f2a63d04/src/intl/en/page-apps.json#L71

https://github.com/ethereum/ethereum-org-website/blob/8ac50bb2a0380f9a29038eb7eeaef5d33827d408/app/%5Blocale%5D/apps/%5Bapplication%5D/page.tsx#L129

image

Above example in prod: https://ethereum.org/apps/aave/

---

We could use those in many places on the site, so adding to common.json would make most sense:

```json
"item-logo": "{item} logo",
```

And in this case, since this uses the client-side `useTranslation` hook instead of the server-side `getTranslations`, we just need to prepend the namespace:

```tsx
alt: t("common:item-logo", { item: exchanges[exchange].name }),
```

image

---

Will push a patch

wackerow (approved):

PR #16499, opened 2025-10-15
Optimize upcoming events list grouping

minimalsm:
## Summary
- replace the lodash-based grouping in the upcoming events list with a native reducer to avoid pulling the full lodash bundle into the client chunk
- add explicit typings for the grouped community events structure to keep TypeScript happy

## Testing
- pnpm lint

------
https://chatgpt.com/codex/tasks/task_e_68f017c879e08322921e3964eaca2079

### DIFF
--- a/src/components/UpcomingEventsList.tsx
+++ b/src/components/UpcomingEventsList.tsx
@@ -1,7 +1,6 @@
 "use client"
 
 import { useEffect, useState } from "react"
-import { groupBy } from "lodash"
 import { useLocale } from "next-intl"
 
 import type { CommunityConference, Lang } from "@/lib/types"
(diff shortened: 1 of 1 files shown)


minimalsm:
@claude review

src/components/UpcomingEventsList.tsx
```diff
@@ -53,31 +60,44 @@ const UpcomingEventsList = () => {
...
+          year: "numeric",
+        }).format(start)
+        const key = `${formatYearMonth}`
+
+        if (!acc[key]) {
+          acc[key] = []
+        }
+
+        acc[key].push(event)
+        return acc
```
wackerow, line 97: Agree with comment..

```tsx
`${}`
```

^ Redundant pattern, but we then don't need to assign to `const key` either

```suggestion

        if (!acc[formatYearMonth]) {
          acc[formatYearMonth] = []
        }

        acc[formatYearMonth].push(event)
        return acc
```

or, just name `formatYearMonth` `key` instead

pettinarip:
@minimalsm would like to better understand the reasoning behind this change.

is this one?:
>  Before: Pulls entire lodash library (~70KB) into client bundle

this doesn’t seem accurate based on what I’ve seen. I ran a bundle analyzer, and only the imported functions are being bundled, not the entire library. So, this code change likely won’t have a real impact on bundle size.

====

One thing we could test, though, is migrating from `lodash` to `lodash-es`, which benefits from some Next.js optimizations: https://nextjs.org/docs/14/app/api-reference/next-config-js/optimizePackageImports

corwintines:
Going to close this out and can address some of this in the events page redesign.

PR #16500, opened 2025-10-16
Improved formatting

Pfed-prog:
## Description

Noticed that improved formatting was possible

image

### DIFF
--- a/public/content/developers/tutorials/downsizing-contracts-to-fight-the-contract-size-limit/index.md
+++ b/public/content/developers/tutorials/downsizing-contracts-to-fight-the-contract-size-limit/index.md
@@ -51,8 +51,6 @@ This one should be obvious. Functions increase a contract size quite a bit.
 
 ### Avoid additional variables {#avoid-additional-variables}
 
-A simple change like this:
-
 ```solidity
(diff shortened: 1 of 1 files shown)


public/content/developers/tutorials/downsizing-contracts-to-fight-the-contract-size-limit/index.md
```diff
@@ -66,15 +66,14 @@ function get(uint id) returns (address,address) {
 }
 ```
 
-makes a difference of **0.28kb**. Chances are you can find many similar situations in your contracts and those can really add up to significant amounts.
+Makes a difference of **0.28kb**. Chances are you can find many similar situations in your contracts and those can really add up to significant amounts.
```
wackerow: Hm, looks like this is structured as if it's the continuation of the previous sentence, with the code example injected... Feels a little off treating this as a new sentence.. cc: @mnelsonBT Thoughts on best way to handle these?

mnelsonBT: (Outlined that in the comment here instead of adding a suggestion so that you can weigh in first on whether or not it makes sense)

PR #16519, opened 2025-10-21
data: adding wallets supporting Unichain and Ink L2 networks

Parizval:
## Description

Since, We have added Unichain and Ink L2 networks on the website. It would be a good time to add wallets for the same. 
Wallets have been added based on the data provided on following platforms. 

https://www.unichain.org/explore

https://inkonchain.com/apps?search=wallet


wackerow:
@all-contributors please add @Parizval for maintenance

PR #16523, opened 2025-10-21
Update index.md

rahimira2:


wackerow:
Don't think this is an appropriate change... @rahimira2 if you'd like to suggest translations, please check out https://ethereum.org/contributing/translation-program/#translation-program as all of these are handled through Crowdin. Thanks!

PR #16560, opened 2025-10-23
fix: rephrase meaning of attestations,signature,beacon_block_root(#16…

### LINKED ISSUE #16503: statements about attestations and signature is imprecise in /block page
### Describe the bug

I think the following statements could be improved.
```
attestations : list of attestations in favor of the current block
signature : aggregate signature of all attesting validators
beacon_block_root : the root hash of the Beacon block containing this object
```

The paragraph is talking about fields within a given block. Most of the fields are related to the current block. But a few fields, from what I recently learn, are against with `previous` block instead of the current one. They're just included by the current one.

`attestations` should be the list of attestations for the previous block(given the inclusion delay is 1), not for the current.
`signature` should be aggregate signature of a set of validators attesting to `previous` block
`beacon_block_root` description is too vague without embodying any connection with the current block. At least we could say it's pointing to the beacon block that validator see as the chain's head(probably the previous block).

### To reproduce

https://github.com/ethereum/ethereum-org-website/blob/78970c04527cbf6a7471d940f0cb55424bef1fb6/public/content/developers/docs/blocks/index.md?plain=1#L60

### Expected behavior

My current understanding may be not very correct. Precise language can be discussed.

- [x] Yes
- [ ] No

baiwfg2:
## Description
I made some changes of description of three `fields` in /block page

- attestations
- signature
- beacon_block_root

@wackerow Would you mind having a check ?

## Related issue
- Closes #16503

### DIFF
--- a/public/content/developers/docs/blocks/index.md
+++ b/public/content/developers/docs/blocks/index.md
@@ -50,23 +50,23 @@ There is a lot of information contained within a block. At the highest level a b
 
 The block `body` contains several fields of its own:
 
-| Field                | Description                                        |
-| :------------------- | :------------------------------------------------- |
-| `randao_reveal`      | a value used to select the next block proposer     |
-| `eth1_data`          | information about the deposit contract             |
-| `graffiti`           | arbitrary data used to tag blocks                  |
-| `proposer_slashings` | list of validators to be slashed                   |
-| `attester_slashings` | list of attesters to be slashed                    |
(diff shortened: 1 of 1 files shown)


wackerow (approved):
Sgtm, thanks @baiwfg2!

wackerow:
@all-contributors please add @baiwfg2 for content patch

PR #16584, opened 2025-10-27
Implement apps of the week feature

corwintines:
## Description

Implements the Apps of the Week feature on the homepage

### DIFF
--- a/app/[locale]/page.tsx
+++ b/app/[locale]/page.tsx
@@ -518,24 +518,27 @@ const Page = async ({ params }: { params: PageParams }) => {
           </Section>
 
           {/* Apps of the week - Discover the best apps on Ethereum */}
-          <Section id="apps-of-the-week">
-            <SectionContent className="flex flex-col gap-4">
-              <div className="flex flex-col items-center text-center">
-                <SectionTag>Apps of the week</SectionTag>
-                <SectionHeader>Discover apps on Ethereum</SectionHeader>
-                <p className="text-lg">Start exploring Ethereum today</p>
-              </div>
-              <AppsHighlight
-                apps={appsOfTheWeek}
-                matomoCategory="apps-of-the-week"
-              />
-              <div className="!mt-8 flex justify-center">
-                <ButtonLink href="/apps" size="lg">
-                  Browse apps <ChevronNext />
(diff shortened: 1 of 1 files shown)


app/[locale]/page.tsx
```diff
@@ -510,6 +517,24 @@ const Page = async ({ params }: { params: PageParams }) => {
...
+                <SectionTag>Apps of the week</SectionTag>
+                <SectionHeader>Discover apps on Ethereum</SectionHeader>
+                <p className="text-lg">Start exploring Ethereum today</p>
+              </div>
+              <AppsHighlight
+                apps={appsOfTheWeek}
+                matomoCategory="apps-of-the-week"
+              />
+              <div className="flex justify-center">
+                <ButtonLink href="/apps">Browse apps &gt;</ButtonLink>
```
Comment (wackerow) [inline app/[locale]/page.tsx]:
```suggestion
                <ButtonLink href="/apps">
                  Browse apps <ChevronNext />
                </ButtonLink>
```

app/[locale]/page.tsx
```diff
@@ -510,6 +517,24 @@ const Page = async ({ params }: { params: PageParams }) => {
...
+            <SectionContent className="flex flex-col gap-4">
+              <div className="flex flex-col items-center text-center">
+                <SectionTag>Apps of the week</SectionTag>
+                <SectionHeader>Discover apps on Ethereum</SectionHeader>
+                <p className="text-lg">Start exploring Ethereum today</p>
+              </div>
+              <AppsHighlight
+                apps={appsOfTheWeek}
+                matomoCategory="apps-of-the-week"
+              />
```
Comment (wackerow) [inline app/[locale]/page.tsx]: image

Hm... looks like this isn't rendering?

wackerow (changes requested):
@corwintines Guessing the parseAppsOfTheWeek function may not be pulling in what we need? didn't look too deep, but doesn't appear to be rendering

app/[locale]/page.tsx
Comment (wackerow) [reply app/[locale]/page.tsx]: Ah, never mind, I see this was just a mock data issue, looks good locally

Comment (corwintines) [reply app/[locale]/page.tsx]: Ah woops I will get mock data updated here

PR #16597, opened 2025-10-29
i18n(ko): Add translations for what-is-ethereum page

DavidBoyun:
## Description

This PR adds Korean translations for several headings and paragraphs on the `what-is-ethereum` page. This improves localization for Korean users who visit one of the most important entry pages.

## Related Issue

N/A


wackerow:
Hey @DavidBoyun, appreciate this, but a couple things:
1- these strings are almost entirely no longer in use as of the latest revamp to the /what-is-ethereum page and wouldn't be useful at this point,
2- non-English translation generally cannot be verified well through GitHub and are to be handled through the Crowdin platform....
3- we're currently revamping the flow of how translations are managed using AI pipelines where humans will be used as editors instead, but still handled through Crowdin for the flow of all the data.

Going to close this out for now, but hopefully can get you involved in this process in the future if you're interested. Thanks again!

PR #16610, opened 2025-11-02
Small text corrections (3)

jorgesumle:
Fixes a typo and removes double white spaces ("  ") .


ethan0xbuilds:
Good catch — small details like this make a big difference for readability and polish. Thanks for taking the time! 🙌

PR #16618, opened 2025-11-04
fix: community events image

graphite-org:
Unclaimed Vercel Instance


wackerow (approved):
Thanks for flagging @eagerapps... replaced with the new og image from the event page

wackerow:
@all-contributors please add @eagerapps for maintenance

Issue #16609, opened 2025-11-02
feat: use CarbonRunner for GitHub action execution


julian-st:
### Is your feature request related to a problem? Please describe.

This repos GitHub Actions (workflows) uses ubuntu-latest but using this tag runs all the jobs in Azure (high-grid intensity) regions ... cough cough (350gCO2/kWh). We can make this better, more carbon-aware.

### Describe the solution you'd like

Swap the runs-on tag to [CarbonRunner's](https://carbonrunner.io/) carbon-aware runner. Finding the most low-carbon region and runs your jobs there. It saves 90% of emissions, 25% per minute cost (free for open-source projects) and is 40% faster than GitHub.

### Describe alternatives you've considered

None

### Would you like to work on this issue?

- [x] Yes
- [ ] No

kaylamarora:
I've read through their docs and it seems we need someone with admin in the repository to do this, but it does seem like a great idea. We (YES WE) all hate Azure and this would help us alot.

Anyways, we need someone with admin.

wackerow:
Hey @julian-st, love the suggestion... don't have any strong info to back up the claims. Would it be possible to place a PR with one workflow to demo this? Our actions are gratis as an open-source repo via GitHub—if this is the case with CarbonRunner I'd be down to start by at least trialing an example workflow. Will assign... if you put up a PR we should (I believe?) be able to trigger an action that hasn't yet been merged from our actions dashboard. If that's not enough I may need to involve devops.

PR #16628, opened 2025-11-06
Change runner to carbonrunner-2vcpu-ubuntu-latest in label

julian-st:
Change runner to carbonrunner-2vcpu-ubuntu-latest in label workflow

## Related Issue

#16609


wackerow:
Thanks @julian-st... Let's actually try this on a workflow that can be dispatched manually.. would recommend using `.github/workflows/update-chains.yml` for this test so I can attempt to dispatch it from your branch.. mind switching this over to that file instead? 🙏

PR #16631, opened 2025-11-07
Data fetch refactor

corwintines:
Create a data layer
Refactor codebase to use data layer

### DIFF
--- a/src/lib/utils/data/clients/redisClient.ts
+++ b/src/lib/utils/data/clients/redisClient.ts
@@ -4,6 +4,7 @@
  */
 
 import { unstable_cache } from "next/cache"
+import { Redis } from "@upstash/redis"
 
 type RedisClient = {
   set: (
@@ -33,7 +34,6 @@ export const getRedisClient = async (): Promise<RedisClient | null> => {
 
   if (upstashUrl && upstashToken) {
     try {
-      const { Redis } = await import("@upstash/redis")
       const upstashClient = new Redis({
         url: upstashUrl,
         token: upstashToken,
--- a/src/lib/utils/data/clients/supabaseClient.ts
+++ b/src/lib/utils/data/clients/supabaseClient.ts
@@ -4,6 +4,7 @@
 
 import { unstable_cache } from "next/cache"
 import type { SupabaseClient } from "@supabase/supabase-js"
+import { createClient } from "@supabase/supabase-js"
 
 let supabaseClient: SupabaseClient | null = null
 let hasWarned = false
@@ -21,7 +22,6 @@ export const getSupabaseClient = async (): Promise<SupabaseClient | null> => {
 
   if (supabaseUrl && supabaseAnonKey) {
     try {
-      const { createClient } = await import("@supabase/supabase-js")
       supabaseClient = createClient(supabaseUrl, supabaseAnonKey)
       return supabaseClient
     } catch (error) {


app/api/external-data/route.ts
pettinarip, line 1: Do we still need this handler?

src/lib/utils/data/extractExternalData.ts
pettinarip: This seems like an extra step in order to get external data that I believe, could be improved.

Ideally, we should get this in one line with the correct typing
```
const gfissues = await getExternalData(["gfissues"], every("day"))
```

src/lib/utils/data/getExternalData.ts
```diff
@@ -0,0 +1,106 @@
...
+  const keysString = sortedKeys.join(",")
+
+  // Use unstable_cache to cache the fetch operation
+  // The cache key includes both the keys and revalidation time to ensure proper caching
+  // Revalidation time matches the requested revalidation period
+  // Tags allow for cache invalidation by key using revalidateTag()
+  // Note: unstable_cache creates a new cached function each time, but Next.js manages
+  // the cache entries based on the keyParts, so this is efficient
+  try {
+    const cachedFetch = unstable_cache(
```
pettinarip: I'm a bit unsure about this implementation since getExternalData creates a new function on each call, it likely won't be cached. The unstable_cache should instead wrap the exported function.

That said, I don't think the unstable_cache belongs at this level. It should be applied within each storage client implementation (eg getRedisData), since each client might use a different caching approach. For example, redis requires unstable_cache but a service using the native fetch API wouldn't need it. By placing the cache here, we're effectively locking ourselves into a specific caching strategy.

src/lib/utils/data/getExternalData.ts
```diff
@@ -0,0 +1,106 @@
...
+      } catch (error) {
+        console.error(`Error fetching data for key "${key}":`, error)
+        return { key, data: null }
+      }
+    })
+  )
+
+  // Build the data map
+  const dataMap = results.reduce((acc, { key, data }) => {
+    if (data !== null && data !== undefined) {
```
pettinarip: Lets make sure we always return one value per key, even if it’s null or undefined, since those might be meaningful to the consumer. Or is that impossible to happen with current logic?

Im in favor of delegating error handling or custom logic to consumers.

src/lib/utils/data/clients/redisClient.ts
pettinarip, line 1: For organization purposes, would recommend placing all storage vendor files in a separate folder like `data/clients`.

src/lib/api/fetchL2beat.ts
```diff
@@ -1,9 +1,30 @@
...
+      return {
+        error: `L2BEAT API responded with ${response.status}: ${response.statusText}`,
+      }
+    }
+
+    const data = (await response.json()) as L2beatResponse
+
+    return {
+      value: data,
+      timestamp: Date.now(),
```
pettinarip: Could this be handled by the `fetchExternalData` function? I think we should separate this strict return type from the fetcher functions.

We could do this automatically inside `fetchExternalData`:
```ts
const result = await service.function()
return { key: service.key, data: { value: result, timestamp: Date.now() } }
```

That way we simplify the usage of the data layer and we garantee that all the calls with have the same data structure.

src/lib/utils/data/extractExternalData.ts
pettinarip, line 1: This feels like an extra layer or glue code just to retrieve external data, and I think it could be simplified.

```ts
const [gfissuesResponse] = await getExternalData(["gfissues"], every("day"))
const { data: gfissues, timestamp, etc } = gfissuesResponse
```

The data you receive here depends entirely on what you return from your fetcher function. I'd still recommend keeping this logic inside each user-defined fetcher instead. That way, interacting with the data layer becomes simpler and more straightforward.

next.config.js
```diff
@@ -94,7 +94,7 @@ module.exports = (phase, { defaultConfig }) => {
         },
         {
           protocol: "https",
-          hostname: "avatars.githubusercontent.com",
+          hostname: "avatars*.githubusercontent.com",
```
pettinarip, line 97: Just to double check if this is a desired change. Asking because it looks a bit out of context.

src/lib/utils/data/clients/redisClient.ts
```diff
@@ -0,0 +1,159 @@
...
+    return redisClient
+  }
+
+  // Try Upstash Redis first (serverless-friendly)
+  const upstashUrl = process.env.UPSTASH_REDIS_REST_URL
+  const upstashToken = process.env.UPSTASH_REDIS_REST_TOKEN
+
+  if (upstashUrl && upstashToken) {
+    try {
+      const { Redis } = await import("@upstash/redis")
```
pettinarip: why do we import this dep like this?

would prefer to do as the docs:
```ts
import { Redis } from "@upstash/redis"

const redis = new Redis({
  url: <UPSTASH_REDIS_REST_URL>,
  token: <UPSTASH_REDIS_REST_TOKEN>,
})
```

src/lib/utils/data/clients/supabaseClient.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (supabaseClient) {
+    return supabaseClient
+  }
+
+  const supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL
+  const supabaseAnonKey = process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY
+
+  if (supabaseUrl && supabaseAnonKey) {
+    try {
+      const { createClient } = await import("@supabase/supabase-js")
```
pettinarip: Same here, imports at the top of the file.

app/api/external-data/route.ts
corwintines: Removed, had a local commit I hadnt pushed up

src/lib/utils/data/getExternalData.ts
corwintines: Thanks for feedback. Pushed up refactor here around this that I believe addresses your points. Good point on making it service level, my bad.

next.config.js
corwintines: I was getting issues with these images coming back from avatars1, avatars2, avatars3, or avatars4.githubusercontent.com, so I had just put this in here for now. I could make a separate PR for this too, but was running into it while developing this.

src/lib/api/fetchL2beat.ts
corwintines: Makes sense. Fixed

corwintines:
Going to close this version out to develop a v2 MVP after discussion with @pettinarip

PR #16634, opened 2025-11-08
Add PriceChart component with live data updateshttps://github.com/ethereum/ethereum-org-website/compare/dev...maxou59-code:ethereum-org-website:patch-3

maxou59-code:
import { Line } from 'react-chartjs-2';
import { useState, useEffect } from 'react';

export default function PriceChart() {
const [labels, setLabels] = useState<string[]>([]); const [values, setValues] = useState<number[]>([]);

useEffect(() => {
const id = setInterval(() => {
setLabels(l => [...l.slice(-50), new Date().toLocaleTimeString()]); setValues(v => [...v.slice(-50), Math.random() * 2000]); }, 1500);
return () => clearInterval(id);
}, []);

const data = {
labels,
datasets: [
{
label: 'Mock price',
data: values,
tension: 0.2,
},
],
};

return <div style={{height:300}}><Line data={data} /></div>; }

### DIFF
--- a/chagede
+++ b/chagede
@@ -1,0 +1,32 @@
+import { Line } from 'react-chartjs-2';
+import { useState, useEffect } from 'react';
+
+
+export default function PriceChart() {
+const [labels, setLabels] = useState<string[]>([]);
+const [values, setValues] = useState<number[]>([]);
+
+
+useEffect(() => {
+const id = setInterval(() => {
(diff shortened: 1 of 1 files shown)


wackerow:
Hey @maxou59-code, unclear what this is attempting to do. Going to close out, but if there is a change you'd like to make would suggest opening an issue first with proper explanation, and/or at the least the code needs to not be a random new unused file in root. Thanks!

PR #16644, opened 2025-11-11
Fusaka banner

corwintines:
## Description

Creates banner for Fusaka

### DIFF
--- a/src/components/Banners/FusakaBanner/index.tsx
+++ b/src/components/Banners/FusakaBanner/index.tsx
@@ -10,7 +10,7 @@ const FusakaBanner = async () => {
 
   return (
     <LinkBox className="w-full bg-[#333369] p-2 text-center text-white md:p-4 md:px-8">
-      <div className="flex flex-col items-center justify-between gap-2 md:flex-row md:gap-16">
+      <div className="flex flex-col items-center justify-center gap-2 md:flex-row md:gap-16">
         <div className="flex flex-col items-center justify-center">
           <p className="text-xl font-extrabold uppercase !leading-none md:text-2xl">
             FUSAKA


src/components/Hero/FusakaHero/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+                br: () => <br className="md:hidden" />,
+              })}
+            </p>
+            <FusakaCountdown />
+          </div>
+        </div>
+      </LinkBox>
+
+      <div className="relative z-0 h-[240px] overflow-hidden md:h-[480px]">
+        <Image
```
pettinarip: This image does not follow the optimizations implemented in the `HomeHero` component. I’d suggest either extending `HomeHero` and passing the Fusaka image as a prop, or updating it here (though that would duplicate some code).

src/components/Hero/FusakaHero/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+import RoadmapFusakaImage from "@/public/images/roadmap/roadmap-fusaka.png"
+
+const FusakaHero = async () => {
+  const locale = getLocale()
+  const t = await getTranslations({ locale, namespace: "page-index" })
+
+  return (
+    <div className="relative w-full">
+      <LinkBox className="bg-[#333369] p-2 text-center text-white">
```
pettinarip: To be honest, I’m not a fan of this banner from a design perspective.
- It doesn’t follow our standard padding
- The justify-between layout feels unbalanced and creates too much empty space
- The background color lacks sufficient contrast and uses a custom value outside the design system


Proposal:
- Use justify-center
- Use bg-primary-low-contrast
- Apply text-body-inverse on the parent and keep defaults for nested elements

image
image

corwintines:
Pushed up based on your feedback minus color changes @pettinarip 

Will bring up with Nuno, but wanting to get this out now for post devconnect so going to bring in.

PR #16648, opened 2025-11-12
Update broken (404) link from home staking guides

jorgesumle:
I've just updated the broken link.


corwintines (approved):
Thanks @jorgesumle

PR #16657, opened 2025-11-15
fix: download buttons for png and svg

corwintines:
## Description

- refactors ButtonLink to Button with download function
- Directly downloads both SVG, PNG, or JPEG files instead of opening image in new tab

### DIFF
--- a/src/components/AssetDownload/index.tsx
+++ b/src/components/AssetDownload/index.tsx
@@ -5,7 +5,7 @@ import type { ImageProps, StaticImageData } from "next/image"
 
 import AssetDownloadArtist from "@/components/AssetDownload/AssetDownloadArtist"
 import AssetDownloadImage from "@/components/AssetDownload/AssetDownloadImage"
-import { Button } from "@/components/ui/buttons/Button"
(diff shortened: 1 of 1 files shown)


src/components/AssetDownload/index.tsx
```diff
@@ -41,6 +40,40 @@ const AssetDownload = ({
...
     })
   }
+
+  const handleDownload = async (url: string, fileExtension: string) => {
+    if (!url) return
+
+    matomoHandler()
+
+    try {
+      const response = await fetch(url)
```
pettinarip: Given that these are same-origin files, we can use the [`download` attr](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#download) in the links instead of doing this complex workaround, we don't need to fetch the file.

```tsx
<ButtonLink href={imgSrc} onClick={matomoHandler} download={`${title.replace(/\s+/g, "-").toLowerCase()}.${fileExtension}`}>
```

pettinarip (approved):
Nice LGTM!

PR #16693, opened 2025-11-25
update: developers/docs/consensus-mechanisms/pos/block-proposal content

rahimira2:

### DIFF
--- a/public/content/developers/docs/consensus-mechanisms/pos/block-proposal/index.md
+++ b/public/content/developers/docs/consensus-mechanisms/pos/block-proposal/index.md
@@ -4,29 +4,51 @@ description: Explanation of how blocks are proposed in proof-of-stake Ethereum.
 lang: en
 ---
 
-Blocks are the fundamental units of the blockchain. Blocks are discrete units of information that get passed between nodes, agreed upon and added to each node's database. This page explains how they are produced.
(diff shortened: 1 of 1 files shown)


corwintines:
Any reason for this big rewrite @rahimira2? On a quick skim, this looks like the same content just reformatted more or less.

This would have large implications on the translations so I don't want to bring in without hearing your reason for rewriting first. 

@mnelsonBT would also be curious what you think about this.

mnelsonBT:
> Any reason for this big rewrite @rahimira2? On a quick skim, this looks like the same content just reformatted more or less.
> 
> This would have large implications on the translations so I don't want to bring in without hearing your reason for rewriting first.
> 
> @mnelsonBT would also be curious what you think about this.

Thanks @corwintines - appreciate you calling out the translation considerations for updating this documentation. 

Some of these edits do help improve the page's clarity and format (for ex. I think it's great to add the bulleted list of resources under Prerequisites). I agree with you though that keeping these foundational docs translate-able is top priority. 
Some changes will be more difficult to translate out of English; removing non-technical descriptors to achieve brevity also removes the context that helps technical jargon translate. A few of these edits also remove background context that is helpful to provide in this "Foundational topics" non-advanced docs section. 

Any info you have here on rationale/goals would be helpful @rahimira2 , thanks! 

I can review to suggest changes on a few spots where it would be helpful to bring back original context or language. It would be helpful to have your feedback/goals in mind when doing so.

corwintines:
Just a bump on this @rahimira2

wackerow:
@rahimira2 Do me a favor and fill in the PR description with the goals of this update, thanks!

wackerow:
Closing out as stale pending requested information

PR #16754, opened 2025-12-01
[hot fix] redirects

wackerow:
## Description
- Fixes several broken redirects related to locale handling
- Refactors all existing redirects to use next.config.js
- Adds lowercasing logic within Next.js middleware
- Removes `/public/_redirects` and all Netlify handling of redirects
- Adds some additional redirects to handle edge cases spotted within Google Search Console, e.g., `/ethereum.html`, usage of `/content/`, `/how`
- Fixes custom locale redirects (`/no` -> `/nb` and `/ph` -> `/fil`)

## Preview link examples
http://deploy-preview-16754--ethereumorg.netlify.app/


public/_redirects
```diff
@@ -1,186 +1,174 @@
...
+/:locale/discord https://discord.gg/ethereum-org 301!
 
-/use /apps/ 301!
+/:locale/pdfs/* /:locale/ 301!
 
-/dapps /apps/ 301!
+/:locale/brand /:locale/assets/ 301!
 
-/beginners /what-is-ethereum/ 301!
+/:locale/ethereum.html /:locale/what-is-ethereum/ 301!
```
corwintines:
```
/:locale/ethereum.html /:locale/what-is-ethereum/ 301!
```

Does this need the .html?

public/_redirects
```diff
@@ -1,186 +1,174 @@
...
+/:locale/layer2/ /:locale/layer-2/ 301!
 
-/developers/docs/mining/ /developers/docs/consensus-mechanisms/pow/mining/ 301!
+/:locale/grants/ /:locale/community/grants/ 301!
 
-/*/beginners /:splat/what-is-ethereum/ 301!
+/no/* /nb/:splat 301!
 
-/*/build /:splat/developers/learning-tools/ 301!
+/ph/* /fil/:splat 301!
```
corwintines:
```suggestion
/ph/* /fil/:splat 301!
```

This has a redirect at the top of the file, can probably removed I think here

public/_redirects
```diff
@@ -1,186 +1,174 @@
...
+/:locale/daos/ /:locale/dao/ 301!
 
-/dart/ /developers/docs/programming-languages/dart/ 301!
+/:locale/layer2/ /:locale/layer-2/ 301!
 
-/developers/docs/mining/ /developers/docs/consensus-mechanisms/pow/mining/ 301!
+/:locale/grants/ /:locale/community/grants/ 301!
 
-/*/beginners /:splat/what-is-ethereum/ 301!
+/no/* /nb/:splat 301!
```
corwintines:
```suggestion
/no/* /nb/:splat 301!
```

Same deal here, this has a redirect at the top of the file, can probably be removed here I think

public/_redirects
wackerow: yeah this was from Google Search Console showing that somewhere people are hitting this URL, so was meant to directly handle that case

next.config.js
```diff
@@ -135,6 +137,146 @@ module.exports = (phase, { defaultConfig }) => {
...
+      const LOCALE_ALTS = i18nConfigJson.map(({ code }) => code).join("|") // e.g. "en|es|fr|..."
+
+      // Helper function to generate both English (no prefix) and locale-prefixed redirects
+      const createRedirect = (from, to, permanent = true) => {
+        // For external URLs, don't modify the destination
+        const isExternal = to.startsWith("http")
+
+        // English / default-locale: no prefix in source or destination
+        const defaultRedirect = {
+          source: `${from}`,
```
pettinarip:
```suggestion
          source: from,
```

public/_redirects
```diff
@@ -1,186 +0,0 @@
-/discord https://discord.gg/ethereum-org 301!
-
-/*/discord https://discord.gg/ethereum-org 301!
-
-/pdfs/* / 301!
-
-/brand /assets/ 301!
-
-/ether /eth/ 301!
```
pettinarip, line 9: double checking, this is now `["/ether", "/what-is-ether/"]`

public/_redirects
```diff
@@ -1,186 +0,0 @@
...
-
-/search / 301!
-
-/use /apps/ 301!
-
-/dapps /apps/ 301!
-
-/beginners /what-is-ethereum/ 301!
-
-/eth2/ /roadmap/ 301!
```
pettinarip, line 27: this one has also changed, now it goes from `eth2` -> `upgrades` -> `roadmap`

next.config.js
```diff
@@ -135,6 +137,146 @@ module.exports = (phase, { defaultConfig }) => {
...
+      }
+
+      return [
+        // Custom locale aliases redirects
+        { source: "/no/:path*", destination: "/nb/:path*", permanent: true },
+        { source: "/ph/:path*", destination: "/fil/:path*", permanent: true },
+
+        // All redirects ([from, to, permanent? (default true)])
+        ...[
+          ["/discord", "https://discord.gg/ethereum-org"],
```
pettinarip: For maintainability, we could consider moving these list to its own file.

```js
module.exports = [
  ["/discord", "https://discord.gg/ethereum-org"],
  // ...
]
```

public/_redirects
wackerow: Okay...

image

Yeah, I see a couple going to /upgrades/ still, I'll patch these

public/_redirects
wackerow: yep! This should be fixed to match the latest changes

image

mnelsonBT:
> @wackerow great job! the redirects are working well. I left a few comments about some rules that differ from the previous setup.
> 
> One more thing to double-check regarding SEO impact: the redirect status code. We're currently using 308, but I'm wondering if we should consider a 301 instead. cc @mnelsonBT

Touched on this in standup @pettinarip @wackerow  but I looked into it further and want to follow up on this tag to confirm. 
Recommended choices:
- Use 301 for all permanent content page redirects (if possible - not a strong SEO issue if it has to be 308 (see below))
- Use 302 for the /fil/ and /pcm/ locale deprecation redirects (we will revisit these as part of larger language convo, and switch to 301 if we are permanently removing the languages) 

Context: 
From a SEO perspective: 
- Client-side redirects = bad (search engine sees it as a weak signal and/or it doesn't carry over to indexing) 
- 301 or 308 redirects = best choices for permanent redirect (tells search engine to permanently save redirect)
- 302 = only use for temporary redirects (search engine will effectively stash the original page in the index, which is a waste of crawl budget if the redirect is actually permanent) 

Within that: 
- 301 = has universal compatibility, but can't pass 'post' requests (for a form, API, etc) 
- 308 = only compatible with modern browsers (http/1.1) but can pass "post" data for APIs, forms, e-commerce checkout etc.
- 301 is recommended for content pages because they don't need any post data requests to carry over, and 301 has universal compatibility (in the case that someone has a very old browser)

Issue #14497, opened 2024-12-12
Update the EVM visualisation


pcaversaccio:
Currently the visualisation of the EVM is slightly outdated as it does not incorporate transient storage [here](https://ethereum.org/en/developers/docs/evm). I would like to suggest to use my updated version I created as part of the EIP-1153 discussion [here](https://ethereum-magicians.org/t/eip-1153-transient-storage-opcodes/553/78): ![7838f70c988ba1d7952f649cb2977ca79c7b5569.png](https://github.com/user-attachments/assets/8c780088-9ca6-4973-b508-f36982706524)

pcaversaccio:
Guys....

pcaversaccio:
This can't be that hard...

pcaversaccio:
Dudes....

pcaversaccio:
![Image](https://github.com/user-attachments/assets/2c00e685-13fd-48da-8dc6-2e3d96fc89bd)

pcaversaccio:
<img width="640" height="490" alt="Image" src="https://github.com/user-attachments/assets/63c0b9a7-cdac-4f2f-89b0-c9124d6b5907" />

pcaversaccio:
Bro

pcaversaccio:
fml

pcaversaccio:
bro

Treetosayoot:
😁

pcaversaccio:
<img width="220" height="290" alt="Image" src="https://github.com/user-attachments/assets/c7eae882-234e-4c5e-b4e3-442c69975c9d" />

corwintines:
@nloureiro would you have time to make this update? Its pretty minimal, mostly looks like wrapping machine state volatile with a persistent part that includes transient storage. 

@pcaversaccio do you think there is content on the page that needs to be updated to reflect this diagram change as well?

<img width="534" height="309" alt="Image" src="https://github.com/user-attachments/assets/b23262b1-8238-4605-803c-fc35a8e07310" />

corwintines:
Left a wip in figma for you @nloureiro

pcaversaccio:
> [@pcaversaccio](https://github.com/pcaversaccio) do you think there is content on the page that needs to be updated to reflect this diagram change as well?

hmm, I guess the [EVM instructions](https://ethereum.org/developers/docs/evm/#evm-instructions) section could be amended to mention transient storage.

pcaversaccio:
[...]

PR #16770, opened 2025-12-02
Update evm diagram

corwintines:
## Description

- Updates EVM diagram to include persistent machine state

## Preview link
https://deploy-preview-16770--ethereumorg.netlify.app/developers/docs/evm

## Related issue
Fixes https://github.com/ethereum/ethereum-org-website/issues/14497#issuecomment-3603185274


wackerow:
image

Just noting the resolution dropped.. we were using 2x before; will place a patch and pull in

PR #16788, opened 2025-12-04
Update wallet-data.ts for Ambire Wallet

dmngdsm:
## Description

Updated information about Ambire Wallet


wackerow:
@all-contributors please add @dmngdsm for maintenance

PR #16801, opened 2025-12-04
[feat] data layer

corwintines:
## Description

Implements:
- Creates abstract setter and getter functions for data layer
- implements netlify blobs as storage
- hourly and daily trigger.dev tasks
- External fetches


src/data-layer/mocks/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Last Generated
+
+2025-12-05T21:54:56.350Z
+
+## Statistics
+
+- Total tasks: 21
+- Successfully generated: 20
+- Not found: 1
+
```
pettinarip: this doesn't look relevant to consumers

src/data-layer/storage/netlifyBlobs.ts
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * Uses Netlify Blobs to store task data with metadata.
+ * The store name "data-layer" acts as a namespace for all task data.
+ *
+ * Store is initialized lazily to avoid errors when environment isn't configured.
+ */
+let store: ReturnType<typeof getStore> | null = null
+
+function getBlobStore() {
+  if (!store) {
```
pettinarip: can use early returns to make this more readable

```ts
if (store) {
  return store
}

// ... rest
```

src/data-layer/storage/netlifyBlobs.ts
```diff
@@ -0,0 +1,121 @@
...
+      // getStore accepts options object with name, siteID, and token
+      store = getStore({
+        name: "data-layer",
+        siteID,
+        token,
+      } as Parameters<typeof getStore>[0])
+    } else {
+      // Try automatic configuration (works in Netlify environment)
+      // This will use NETLIFY_BLOBS_CONTEXT if available
+      store = getStore("data-layer")
```
pettinarip: Not sure if we are ever going to use this use case. Is this something you have in mind for other uses?

IMO we should always have a these vars. If not, throw to let us know something is not working.

src/data-layer/storage/netlifyBlobsStorage.ts
pettinarip, line 1: For consistency and clarity, following the `mockStorage` name, we should call this `netlifyBlobsStorage`.

src/data-layer/storage/netlifyBlobs.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+  }
+  return store
+}
+
+export const netlifyBlobsStorage: StorageImplementation &
+  GetStorageImplementation = {
```
pettinarip: current typing around the storage is a bit fragmented and confusing.

would suggest unifying these types in `types.ts` and have 1 interface for the `Storage`.
```ts
// types.ts file

import type { tasks } from "./registry"

export type TaskId = (typeof tasks)[number]["id"]

/** Metadata stored alongside data */
export interface StorageMetadata {
  storedAt: string
}

/** Result shape when retrieving data with metadata */
export interface StorageResult<T> {
  data: T
  metadata: StorageMetadata
}

/**
 * Unified storage interface.
 * Implementations must provide both get and set methods.
 */
export interface Storage {
  get<T>(taskId: TaskId): Promise<StorageResult<T> | null>
  set(taskId: TaskId, data: unknown, metadata?: StorageMetadata): Promise<void>
}
```

then the implementation would be cleaner

```ts
export const netlifyBlobsStorage: Storage = {
  async get<T>(taskId: TaskId) { ... },
  async set(taskId: TaskId, data: unknown, metadata?: StorageMetadata) { ... },
}
```

src/data-layer/trigger/tasks/daily.ts
```diff
@@ -0,0 +1,36 @@
...
+  // Run daily at midnight UTC
+  cron: "0 0 * * *",
+  run: async () => {
+    const results: Record<string, unknown> = {}
+
+    // Loop through all daily tasks and execute them
+    for (const task of dailyTasks) {
+      try {
+        console.log(`Fetching data for task: ${task.id}`)
+        const data = await task.fetchFunction()
```
pettinarip, line 22: we are running these sequentially which can be a major bottleneck. We can parallelized by using `await Promise.allSettled`

pettinarip (commented):
@corwintines this is looking really good.

I have one last architectural concern about how consumers will interact with the data-layer.

Right now, to get data you'd need to do this:
```typescript
import { getData } from "@/data-layer/storage/getter"
import { FETCH_ETH_PRICE_TASK_ID } from "@/data-layer/api/fetchEthPrice"
```

This leaks internal details (storage paths, task IDs) into app code and loses type safety.

**Suggested approach**

1. Data-layer exports a clean public API (`src/data-layer/index.ts`)

```typescript
export async function getEthPrice(): Promise<EthPriceData | null>
export async function getL2beatData(): Promise<L2beatData | null>
// ...more
```

No Next.js imports here — keeps it framework agnostic so we can extract it later if needed.

2. Next.js adapter handles caching (`src/lib/data/index.ts`)

```typescript
import { unstable_cache } from "next/cache"
import * as dataLayer from "@/data-layer"

export const getEthPrice = unstable_cache(
  () => dataLayer.getEthPrice(),
  ["eth-price"],
  { revalidate: 3600 }
)
```

3. App code uses the adapter
```typescript
import { getEthPrice } from "@/lib/data"
const price = await getEthPrice() // cached + typed ✨
```

**Why**

- Types flow automatically (no generics)
- Data-layer stays isolated, no framework deps
- Caching is the app's concern, not the data-layer's
- Easy to extract to its own service later

Thoughts?

src/data-layer/mocks/README.md
corwintines: Could be useful depending on if we need to keep track of the last time a mocks were created, but will remove for now.

tests/unit/data-layer/getters.spec.ts
```diff
@@ -0,0 +1,335 @@
...
+// Register path aliases
+try {
+  require("tsconfig-paths/register")
+} catch (e) {
+  // tsconfig-paths not available, paths may not resolve
+  console.warn("tsconfig-paths not found, path aliases may not work")
+}
+
+// Set USE_MOCK_DATA before importing data-layer
+process.env.USE_MOCK_DATA = "true"
```
pettinarip: lets avoid setting env vars in code. You are already doing `"test:unit": "USE_MOCK_DATA=true playwright test --project=unit"` in the scripts (which is the correct approach).

tests/unit/data-layer/getters.spec.ts
```diff
@@ -0,0 +1,335 @@
...
+ * 2. Return the expected types and structures
+ * 3. Handle null cases gracefully
+ * 4. Validate data structure when data is present
+ */
+
+import { expect, test } from "@playwright/test"
+
+// Register path aliases
+try {
+  require("tsconfig-paths/register")
```
pettinarip: why do we need to register this? is this necessary? the e2e tests did not require it

playwright.config.ts
```diff
@@ -7,14 +7,14 @@ import { defineConfig, devices } from "@playwright/test"
 dotenv.config({ path: path.resolve(__dirname, ".env.local") })
 
 export default defineConfig<ChromaticConfig>({
-  testDir: "./tests/e2e",
-  outputDir: "./tests/e2e/__results__",
+  testDir: "./tests",
+  outputDir: "./tests/__results__",
```
pettinarip, line 11: this is probably need that we update GH action yaml file as well, or not?

tests/unit/data-layer/getters.spec.ts
```diff
@@ -0,0 +1,335 @@
...
+import * as dataLayer from "@/data-layer"
+
+test.describe("Data-Layer Getters", () => {
+  test.describe("Price & Market Data", () => {
+    test("getEthPrice returns MetricReturnData or null", async () => {
+      const result = await dataLayer.getEthPrice()
+      if (result !== null) {
+        // MetricReturnData is ValueOrError<number>
+        expect(result).toHaveProperty("value")
+        expect(typeof result.value).toBe("number")
```
pettinarip: I get a bunch of TS errors coming mostly from accessing the `value` prop

```sh
tests/unit/data-layer/getters.spec.ts:23:30 - error TS2339: Property 'value' does not exist on type 'MetricReturnData'.
  Property 'value' does not exist on type '{ error: string; }'.
```

src/lib/data/index.ts
```diff
@@ -0,0 +1,402 @@
...
+  async () => {
+    try {
+      return await dataLayer.getEthPrice()
+    } catch (error) {
+      console.error(
+        "[getEthPrice] Error calling dataLayer.getEthPrice():",
+        error
+      )
+      throw error
+    }
```
pettinarip: Why are we adding these try/catch? errors bubble up fine without them and can be handled by consumers. This just seem like unnecessary code.

src/data-layer/storage/getter.ts
```diff
@@ -0,0 +1,73 @@
...
+    console.error(`[Data Layer] Error retrieving data for task "${taskId}":`)
+    console.error(`  Error message: ${errorMessage}`)
+    console.error(`  USE_MOCK_DATA: ${process.env.USE_MOCK_DATA}`)
+    if (errorStack) {
+      console.error(`  Stack trace:`, errorStack)
+    }
+
+    // Re-throw the original error to preserve its message and stack
+    // This ensures mockStorage's detailed error messages are visible
+    throw error
```
pettinarip: Feels unnecessary this try/catch for just logging things. Lets keep it simple. Error will bubble up and we will need to handle the error anyway on the consumer side.

src/data-layer/mocks/fetch-apps.json
pettinarip, line 1: 6k line file, can we simplify this mock?

src/data-layer/mocks/fetch-git-history.json
pettinarip, line 1: 8k line file, can we simplify?

src/data-layer/mocks/fetch-l2beat.json
pettinarip, line 1: 17k file, pls no xD

PR #16804, opened 2025-12-05
Deprecate and remove pcm and fil locales

pettinarip:
## Description

Removes support for `pcm` (Nigerian Pidgin) and `fil` (Filipino) locales and adds temporary 307 redirects to the default locale.

- adds temp redirects to root equivalents
- deletes translation files

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -166,17 +166,6 @@ module.exports = (phase, { defaultConfig }) => {
         ...redirects.flatMap(([source, destination, permanent]) =>
           createRedirect(source, destination, permanent)
         ),
-
-        // Custom locale aliases redirects
-        { source: "/no/:path*", destination: "/nb/:path*/", permanent: true },
-
-        // Deprecated locale redirects
-        { source: "/ph", destination: "/", permanent: true },
-        { source: "/ph/:path*", destination: "/:path*/", permanent: true },
-        { source: "/pcm", destination: "/", permanent: false },
-        { source: "/pcm/:path*", destination: "/:path*/", permanent: false },
-        { source: "/fil", destination: "/", permanent: false },
-        { source: "/fil/:path*", destination: "/:path*/", permanent: false },
       ]
     },
   }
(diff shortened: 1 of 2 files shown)


src/lib/utils/url.ts
wackerow, line 1: @pettinarip Is this being used anywhere?

wackerow: (`getFirstSegment`)

pettinarip: ah good catch, thanks. Will clean this up

next.config.js
```diff
@@ -162,10 +162,6 @@ module.exports = (phase, { defaultConfig }) => {
       }
 
       return [
-        // Custom locale aliases redirects
-        { source: "/no/:path*", destination: "/nb/:path*", permanent: true },
-        { source: "/ph/:path*", destination: "/fil/:path*", permanent: true },
```
pettinarip, line 167: @corwintines @wackerow these redirects for deprecated locales (/pcm, /fil, /ph) and the Norwegian alias (/no → /nb) were working fine locally but not on Netlify.

The reson is that, on Netlify Edge, the next-intl middleware runs before Next.js processes the redirects in next.config.js. So when someone hit /pcm/some-page, the middleware intercepted it first, didn't recognize pcm as a valid locale, and returned a 404 instead of redirecting to the English version.

By handling these locale redirects directly in the middleware (before next-intl does its thing), we ensure they work consistently everywhere — local dev, Netlify preview deploys, and prod.

src/lib/utils/url.ts
pettinarip: @wackerow this utility was left from a previous implementation and now, I had to reimplement it due to an issue with the locale redirects within Nextjs. More details in [this comment](https://github.com/ethereum/ethereum-org-website/pull/16804/files#r2601868145)

wackerow (approved):
Lgtm! Working well in Netlify preview... ph fil and pcm being stripped out appropriately, and any further redirects for paths are being applied. /no going to /nb along with any other applicable redirects. 👍

PR #16830, opened 2025-12-08
/layer-2/networks data layer implementation

corwintines:
## Description

- Implements cachedGetter function for fetching and revalidating data from storage
- Implements cachedGetter on /layer-2/networks page

PR #16839, opened 2025-12-10
Implement data layer across pages

corwintines:
## Description

- Removes old implementation of dataLoader
- Removes old mocks implementation
- Implements data loader across all React pages on ethereum.org

Depends on #16801 and #16830


corwintines:
Refactored the data layer, so going to close this out for that implementation

PR #16892, opened 2025-12-15
i18n: automated Crowdin translation import

wackerow:
## Description

This PR contains automated LLM translations from Crowdin

### Languages translated

- es


minimalsm:
@wackerow can you add lang name or code in the title of the PR in the script plz

wackerow:
@minimalsm Yes yes, and sorry that these aren't automatically being placed in draft, but these are currently still test runs. As they're being translated, the results are being saved in Crowdin though so re-running without prompt changes will effectively grab the existing.

Can chat through how we want to handle some of the organization around the PRs, but eventually we won't necessarily have one-PR-per-language (think a single file change that gets translated for all locales). For now though as we iterate toward the initial translation of everything, I can add the language to the PR title so it's clear.

PR #16897, opened 2025-12-15
fix(e2e): fix mobile browser e2e test failures and race conditions

pettinarip:
Fix flaky e2e tests that were failing on mobile browsers.

## Description
- Add `aria-label` and `data-testid` to mobile menu dialog for reliable test selectors
- Fix race conditions in filter tests by waiting for checkboxes to render before extracting text
- Use `toHaveText` assertion instead of `textContent()` to avoid timing issues
- Update wallet modal test to match actual "Coinbase" label (was "Coinbase Wallet")

### DIFF
--- a/src/components/Nav/MobileMenu/MobileMenuClient.tsx
+++ b/src/components/Nav/MobileMenu/MobileMenuClient.tsx
@@ -44,6 +44,7 @@ const MobileMenuClient = ({
         onOpenChange={setOpen}
         triggerRef={triggerRef}
         aria-label={t("site-title")}
+        data-testid="mobile-menu-dialog"
       >
         {children}
       </PersistentPanel>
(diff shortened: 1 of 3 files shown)


tests/e2e/start.spec.ts
```diff
@@ -19,7 +19,7 @@ test.describe("Start Page", () => {
     const walletModal = page.getByRole("dialog")
     await expect(walletModal.getByText("Connect a Wallet")).toBeVisible()
     await expect(walletModal.getByText("MetaMask")).toBeVisible()
-    await expect(walletModal.getByText("Coinbase Wallet")).toBeVisible()
+    await expect(walletModal.getByText("Coinbase")).toBeVisible()
```
wackerow, line 22: @pettinarip Hm, what am I missing? Looks like the wallet is named "Coinbase Wallet" in our wallet data

image

image

Also, looks like we should probably update that separately: https://www.coinbase.com/wallet ("Coinbase Wallet is now **Base app**")

pettinarip: @wackerow this is for the /start page. The test is matching a few wallet names that appear in the connect modal.

PR #16904, opened 2025-12-16
update(ui): homepage price location below text

wackerow:
## Description
- Moves homepage price metric below text ("What is ETH?" > [answer] > [price] > [cta])
- Refactors price element out of single-use component

image

cc: @konopkja @nloureiro


corwintines:
I see in Discord that there was thumbs up from designers. Going to merge.

PR #16959, opened 2025-12-19
i18n: automated Crowdin translation import (es)

wackerow:
## Description

This PR contains automated LLM translations from Crowdin

### Languages translated

es

#### Markdown changes (`public/content/translations/{locale}/`)

- staking/dvt/index.md
- staking/solo/index.md
- staking/pools/index.md
- staking/withdrawals/index.md
- staking/saas/index.md

### DIFF
--- a/public/content/translations/es/staking/saas/index.md
+++ b/public/content/translations/es/staking/saas/index.md
@@ -17,7 +17,7 @@ summaryPoints:
 
 La participación como servicio (SaaS), representa una categoría de servicios de participación donde deposita 32 ETH para la validación, pero se delegan las operaciones del nodo a un tercero. A cambio se le suele guiar por la configuración inicial, incluida la generación de claves y el depósito, para luego cargar las claves de firma al operador. Esto permite que el servicio maneje su validador en su nombre, generalmente, a cambio de una cuota mensual.
 
-## ¿Por qué debería participar con un servicio? {#por-qué-apostar-con-un-servicio}
+## ¿Por qué debería participar con un servicio? {#why-stake-with-a-service}
 
 El protocolo de Ethereum no respalda originariamente la delegación de participaciones y, por tanto, estos servicios se han creado con el fin de cubrir esta demanda. Si tiene 32 ETH para participar, pero no domina con confianza el hardware, los servicios de participación (SaaS) le permiten delegar la parte técnica mientras gana recompensas de bloques nativos.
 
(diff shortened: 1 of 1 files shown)


public/content/translations/es/staking/pools/index.md
```diff
@@ -17,70 +17,73 @@ summaryPoints:
...
+  <Card title="Baja barrera de entrada" emoji="🐟" description="¿No eres una ballena? No hay problema. La mayoría de los pools de staking te permiten stakear prácticamente cualquier cantidad de ETH uniendo fuerzas con otros stakers, a diferencia del staking en solitario, que requiere 32 ETH." />
+  <Card title="Haz staking hoy" emoji=":stopwatch:" description="Hacer staking con un pool es tan fácil como un intercambio de tokens. No tienes que preocuparte por la configuración del hardware ni por el mantenimiento del nodo. Los pools te permiten depositar tu ETH, lo que permite a los operadores de nodos ejecutar validadores. Las recompensas se distribuyen entre los contribuyentes, restando una comisión por las operaciones del nodo." />
+  <Card title="Tokens de staking" emoji=":droplet:" description="Muchos pools de staking proporcionan un token que representa un derecho sobre tu ETH en staking y las recompensas que genera. Esto te permite utilizar tu ETH en staking, por ejemplo, como colateral en aplicaciones DeFi." />
```
pettinarip, line 29: Perfect 💯

public/content/translations/es/staking/pools/index.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Participación agrupada
-description: Una visión general de cómo empezar con la participación agrupada de ETH
+title: Staking en conjunto
```
pettinarip, line 2: Not sure if we should be the word "pool" in this case but minor issue. The sound and meaning is good.

pettinarip (approved):
Loved it. Lets wait for the build to finish but I couldn't spot any issue.

public/content/translations/es/staking/pools/index.md
```diff
@@ -17,70 +17,73 @@ summaryPoints:
...
 
 <ExpandableCard title="¿Cuándo puedo retirar mi participación?">
 ¡En cualquier momento! La actualización de red Shanghai/Capella se produjo en abril de 2023 e introdujo las retiradas de participaciones. Después de esta actualización, las cuentas de validador que respaldan las reservas de participación tendrán la posibilidad de salir y retirar ETH a su dirección de retirada designada. Esto permitirá la capacidad de canjear su parte de participación por el ETH subyacente. Compruebe con su proveedor la compatibilidad con esta funcionalidad.
 
 Alternativamente, los grupos que usan tókenes de participación ERC-20, permiten a los usuarios operar dicho token en el libre mercado, pudiendo vender la posición en participación, «retirándola» de forma eficaz sin tener que eliminar ETH del contrato de participación.
 
-<ButtonLink href="/staking/withdrawals/">Más sobre retiradas de participaciones</ButtonLink>
+<ButtonLink href="/staking/withdrawals/">
+Más sobre retiros de staking
+</ButtonLink>
```
pettinarip, line 73: Ah @wackerow — not a blocker, but just for future reference: ideally, we shouldn't have these empty spaces if the children are just text.

The previous version was actually correct (in term of spacing not translation).

The extra spaces cause `<p>` tags to be generated in newer MDX versions, which is problematic. If we can avoid them, that would be much better.

public/content/translations/es/staking/pools/index.md
wackerow: Yeah seems like we should... this is something v0.2.0 should address. If it's minor, let's move forward for now.

public/content/translations/es/staking/pools/index.md
wackerow: Yeah good catch—patching this. Plan was to add line returns only if the opening tag was on a new line to prevent things like `<ExpandableCard><p>something</ExpandableCard></p>`.

I just pushed a patch and I'm going to re-run the job to double check this... if that looks good, I think we're set to run this on a larger set of files.

wackerow: image

Fixed in #16967

PR #17020, opened 2026-01-07
Events pages revamp

wackerow:
## Description

Revamps the community events pages with a new API integration and improved UX.

Key changes:
- Integrate Geode Labs API for live event data (replaces static markdown)
- Add dedicated pages for conferences, meetups, and search
- Add client-side filtering by continent, date, and event type
- Add community hub carousel with SF, Berlin, London, Dubai, Lagos hubs
- Add EdgeScroll UI components for horizontal scrolling sections
- Update ui/tag component style options to accommodate tag variations
- Add JSON-LD structured data for SEO
- Add custom Matomo event tracking
- Deprecate legacy markdown events page, code including import system, package script, unused components and old data

Routes:
- /community/events - Main events hub
- / - Events section of homepage (no change, just new source)
- /developers - Hackathons section (new data, refactors EdgeScroll* components, deprecates Swiper)

New routes:
- /community/events/conferences - Conference listings
- /community/events/meetups - Meetup listings
- /community/events/search - Event search

## Preview link
https://deploy-preview-17020.ethereum.it/community/events

### DIFF
--- a/app/[locale]/community/events/search/page.tsx
+++ b/app/[locale]/community/events/search/page.tsx
@@ -6,6 +6,7 @@ import type { EventItem, PageParams } from "@/lib/types"
 import ContentHero from "@/components/Hero/ContentHero"
 import MainArticle from "@/components/MainArticle"
 import { Alert, AlertContent } from "@/components/ui/alert"
+import { Button } from "@/components/ui/buttons/Button"
 import Input from "@/components/ui/input"
 import { Section } from "@/components/ui/section"
 
@@ -33,6 +34,7 @@ const Page = async ({
     locale,
     namespace: "page-community-events",
   })
+  const tCommon = await getTranslations({ locale, namespace: "common" })
 
   const events = mapEventTranslations(_events, t)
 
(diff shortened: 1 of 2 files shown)


wackerow:
> Major perf issue with continent icons. The 7 continent SVGs total ~1MB, with north-america.svg alone at 560KB. Since this is a client component, all SVGs get bundled into the JavaScript that users download on page load.

Heard. Added 22c8ac5646 which refactors all of the svg's, dropping total size to 17,704 bytes (37 KB on disk) for all of them combined. With that, going to close #17079 since that limits the ability to style with hover/active states. Thanks for noting!

Also pushed a patch for the i18n string which needed the `year` variable.

As for the styling issue, cc: @konopkja The top uses existing ContentHero which has x-padding of 2rem on mobile, while the design uses 1rem padding for the body of this page. Small thing, and probably worth a separate task at this point. Let us know if you have thoughts

app/[locale]/page.tsx
```diff
@@ -61,13 +61,22 @@ import { Skeleton, SkeletonCardGrid } from "@/components/ui/skeleton"
...
+import {
+  getAppsData,
+  getAttestantPosts,
+  getBeaconchainEpochData,
+  getEthPrice,
+  getEventsData,
+  getGrowThePieData,
+  getRSSData,
+  getTotalValueLockedData,
+} from "@/data-layer"
```
Comment (pettinarip) [inline app/[locale]/page.tsx]: If we call the fetch functions directly, the data and page will never be revalidated.

By going through `@/lib/data`, we're using unstable_cache, which forces the app to revalidate those data sources periodically.

src/data/mocks/events.json
pettinarip, line 1: legacy mocks, not used anymore

src/data-layer/mocks/fetch-events.json
pettinarip, line 1: This mock is pretty large, we can simplify to reduce size for the repo.

pettinarip (approved):
LFG!

PR #17051, opened 2026-01-11
fix: replace VisuallyHidden with sr-only class

wackerow:
## Summary
- Replace Radix's `VisuallyHidden` component with Tailwind's `sr-only` class
- Add `position: relative` to external link anchors to properly contain the sr-only element
- Change sr-only from `<div>` to `<span>` for proper inline semantics
- Fixes overflow issues in horizontal scroll containers

The Radix VisuallyHidden component has a known bug where `position: absolute` combined with `whitespace: nowrap` but no explicit `top`/`left` values causes content to overflow its container. Additionally, using a block-level `<div>` inside an inline `<a>` element caused layout issues.

See: https://github.com/radix-ui/primitives/issues/3406

## Test plan
- [ ] Verify screen reader text still works for external links
- [ ] Verify no horizontal page overflow on pages with horizontal scroll containers

Generated with Claude Code

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
     "@radix-ui/react-switch": "^1.1.0",
     "@radix-ui/react-tabs": "^1.1.0",
     "@radix-ui/react-tooltip": "^1.1.2",
-    "@radix-ui/react-visually-hidden": "^1.1.0",
     "@rainbow-me/rainbowkit": "^2.2.3",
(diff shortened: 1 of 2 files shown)


wackerow:
Yeah good call.. will remove that package, build quickly locally to make sure no issues, then pull this in with that change, thanks @pettinarip

PR #17079, opened 2026-01-14
perf(events): optimize continent SVGs and add dark mode support

pettinarip:
## Summary
- Move continent SVGs from bundled React components to public folder
- Optimize SVGs with SVGO, reducing total size from ~1MB to ~370KB (62% smaller)
- Icons now load as images instead of being bundled in JS
- Add `dark:invert` CSS class for dark mode support

## Test plan
- [ ] Verify continent icons display correctly in light mode
- [ ] Verify continent icons display correctly in dark mode (should invert)
- [ ] Check network tab to confirm SVGs load from public folder


wackerow (commented):
@pettinarip This works well for basic black/white but doesn't handle the active/hover states very well:
image

What if we were to import the optimized SVGO forms of these?

cc: @konopkja

wackerow:
#17020 commit 22c8ac56466ee0180f8a6e7a92c6b0deb2a40cf0 refactors all the SVGs to be significantly smaller, while allowing active/hover styling. Closing this one out, thanks @pettinarip!

PR #17109, opened 2026-01-19
refactor/feat(ui): ui/card and ui/tag component refactors

wackerow:
## Description
- Refactors for general ui/card and ui/tag components
- Fixes: tailwind class string interpolation
- Adds: ui/tag TagsInlineText component
- Refactors: ui/card components, consolidating CardDescription, CardHighlight and CardSubTitle into single CardParagraph component with cva variants
- General cleanup for other "card" components/sub-components

## Related Issue
In preparation for new /developers/apps pages

### DIFF
--- a/app/[locale]/apps/_components/AppCard.tsx
+++ b/app/[locale]/apps/_components/AppCard.tsx
@@ -9,6 +9,8 @@ import { APP_TAG_VARIANTS } from "@/lib/utils/apps"
 import { cn } from "@/lib/utils/cn"
 import { slugify } from "@/lib/utils/url"
 
+import { SIZE_CLASS_MAPPING } from "@/lib/constants"
+
 interface AppCardProps {
   app: AppData
   imageSize: number
@@ -32,13 +34,6 @@ const AppCard = ({
   matomoCategory,
   matomoAction,
 }: AppCardProps) => {
-  const SIZE_CLASS_MAPPING: Record<AppCardProps["imageSize"], string> = {
-    10: "size-10",
-    12: "size-12",
-    14: "size-14",
-    16: "size-16",
-    24: "size-24",
-  }
   const cardContent = (
     <div
       className={cn(
--- a/src/components/ui/card.tsx
+++ b/src/components/ui/card.tsx
@@ -19,7 +19,7 @@ const Card = React.forwardRef<
     return (
       <BaseLink
         href={href}
-        className={cn(cardVariants({ className }), "group/link", className)}
+        className={cn(cardVariants({ className }), "group/link")}
         customEventOptions={customEventOptions}
         hideArrow
       >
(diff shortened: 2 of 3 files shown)


src/components/ui/card.tsx
```diff
@@ -92,37 +120,37 @@ const CardTitle = React.forwardRef<
...
+>(({ className, variant, size, ...props }, ref) => (
   <p
     ref={ref}
-    className={cn("text-sm text-body-medium", className)}
+    className={cn(paragraphVariants({ variant, size, className }))}
     {...props}
   />
 ))
-CardDescription.displayName = "CardDescription"
+CardParagraph.displayName = "CardDescription"
```
pettinarip:
```suggestion
CardParagraph.displayName = "CardParagraph"
```

src/components/ui/card.tsx
```diff
@@ -5,49 +5,36 @@ import { cn } from "@/lib/utils/cn"
...
+  if (href) {
     return (
-      <div
-        ref={ref}
-        className={cn("group rounded-2xl", className)}
-        {...props}
-      />
+      <BaseLink
+        href={href}
+        className={cn(cardVariants({ className }), "group/link", className)}
```
pettinarip: unnecessary double `className`. Would suggest removing the last one, could cause unexpected overrides.
```suggestion
        className={cn(cardVariants({ className }), "group/link")}
```

src/components/ui/card.tsx
```diff
@@ -5,49 +5,36 @@ import { cn } from "@/lib/utils/cn"
...
+      >
+        <div ref={ref} {...props} />
+      </BaseLink>
     )
   }
-)
+  return (
+    <div
+      ref={ref}
+      className={cn(cardVariants({ className }), "group", className)}
```
pettinarip: same here

src/components/ui/tag.tsx
```diff
@@ -165,4 +165,36 @@ const TagButton = forwardRef<HTMLButtonElement, TagButtonProps>(
 
 TagButton.displayName = "TagButton"
 
-export { Tag, TagButton }
+const listVariants = cva("", {
+  variants: {
+    variant: {
+      light: "text-sm text-body-medium",
```
pettinarip, line 171: shouldn't this style be part of the base style? otherwise, shouldn't we add a `defaultVariants`?

app/[locale]/apps/_components/AppCard.tsx
```diff
@@ -32,39 +31,31 @@ const AppCard = ({
   hoverClassName,
   matomoCategory,
   matomoAction,
-  forceLightMode = false,
 }: AppCardProps) => {
+  const SIZE_CLASS_MAPPING: Record<AppCardProps["imageSize"], string> = {
```
Comment (pettinarip) [inline app/[locale]/apps/_components/AppCard.tsx]: nitpick: this is an static object, should be moved outside of the component.

pettinarip (approved):
@wackerow looks good. Left a few minor comments

src/components/ui/tag.tsx
wackerow: We use both, and felt cleaner to leave the base case as unstyled, so we didn't default to text-sm and text-body-medium to then have to override it to remove it. This leaves default case without additional styling, with option to add the "light" variant

PR #17110, opened 2026-01-19
feat: new /developer/tools pages

wackerow:
## Description

Adds developer tools directory at /developers/tools with categorized tooling

- New developer tools pages and category subpages
- Server side; state tracked by searchParams
- Data layer integration
  - Raw data fetched from BuidlGuidl developer tooling GitHub endpoint
  - Enriches data with GitHub star information from GitHub API and weekly downloads count from NPM API
- Enhanced UX - App modals, tag filtering, pseudo-randomized highlights, and JSON-LD for SEO
- Infrastructure updates - Date/time helpers, retry logic for rate limiting, navigation updates
- JSON-LD included
- Set up for i18n translations

### Sub-PRs:
- #17130 (can merge separately)
- #17131
- #17134 
- #17137 
- #17178 
- #17180 
- #17181 

## Preview link
https://page-developer-apps--ethereumorg.netlify.app/developers/tools

### DIFF
--- a/app/[locale]/developers/tools/[category]/page.tsx
+++ b/app/[locale]/developers/tools/[category]/page.tsx
@@ -143,6 +143,8 @@ export async function generateStaticParams() {
   return DEV_TOOL_CATEGORIES.map(({ slug }) => ({ category: slug }))
 }
 
+export const dynamicParams = false
+
 export async function generateMetadata({
   params,
 }: {


src/data-layer/mocks/fetch-developer-tools.json
pettinarip, line 1: Lets reduce the size of this mock. Can we have a minimal mock so that the UI doesn't break?

app/[locale]/developers/tools/types.ts
```diff
@@ -0,0 +1,114 @@
+import { DeveloperAppsResponse } from "@/lib/types"
+
+import { DEV_APP_CATEGORIES, DEV_APP_CATEGORY_SLUGS } from "./constants"
+
+export type DeveloperAppTag =
```
Comment (pettinarip) [inline app/[locale]/developers/tools/types.ts]: can't we infer all these types from the constants? or maybe just use `string`? is this list needed?

src/data-layer/index.ts
```diff
@@ -15,6 +15,8 @@ import type {
 } from "@/lib/types"
 import type { CommunityEventsReturnType } from "@/lib/interfaces"
 
+import type { DeveloperApp } from "../../app/[locale]/developers/apps/types"
```
pettinarip: lets follow the pattern and declare this type in `/lib/types`. This also fix a circular dep issue

src/data-layer/fetchers/developer-tools/fetchGitHub.ts
```diff
@@ -0,0 +1,159 @@
...
+  } catch {
+    return null
+  }
+}
+
+function buildGraphQLQuery(repos: RepoInfo[]): string {
+  const repoQueries = repos
+    .map(
+      (repo, i) => `
+    repo${i}: repository(owner: "${repo.owner}", name: "${repo.name}") {
```
pettinarip, line 43: Minor security note: the owner/name values from BuidlGuidl get interpolated directly into the query. The practical risk is low, but adding a simple regex check here would be good hygiene:
```ts
const VALID_REPO_PATTERN = /^[a-zA-Z0-9._-]+$/
```
Not blocking, just a nice-to-have.

app/[locale]/developers/apps/page.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+  const activeApp = enrichedData.find((app) => app.id === appId)
+
+  // Clean up invalid appId by redirecting
+  if (appId && !activeApp) {
+    redirect("/developers/apps")
+  }
+
+  // Get dynamic highlights based on stars and recent activity (cached weekly)
+  const highlightsByCategory = await getCachedHighlightsByCategory(enrichedData)
```
Comment (pettinarip) [inline app/[locale]/developers/apps/page.tsx]: Why do we need to cache this? the page is cached already by the enrichedData fetch, right? this is derived data from that result.

app/[locale]/developers/apps/page.tsx
```diff
@@ -0,0 +1,219 @@
...
+  const highlights = getMainPageHighlights(highlightsByCategory)
+
+  // Get randomized previews (5 apps per category) - cached daily
+  const previewsByCategory =
+    await getCachedRandomPreviewsByCategory(dataByCategory)
```
Comment (pettinarip) [inline app/[locale]/developers/apps/page.tsx]: Same here, all of them derive from the same `enrichedData` object

app/[locale]/developers/apps/[category]/page.tsx
```diff
@@ -0,0 +1,241 @@
...
+        category={category}
+        categoryApps={allCategoryData}
+        contributors={contributors}
+      />
+      <ContentHero
+        breadcrumbs={{
+          slug: `/developers/apps/${t(`page-developers-apps-category-${category}-breadcrumb`)}`,
+        }}
+        title={t(`page-developers-apps-category-${category}-title`)}
+        description={t(`page-developers-apps-category-${category}-description`)} // TODO: Confirm
```
Comment (pettinarip) [inline app/[locale]/developers/apps/[category]/page.tsx]: confirmed?

app/[locale]/developers/tools/[category]/page.tsx
```diff
@@ -0,0 +1,241 @@
...
+    params.set("appId", appId)
+    if (validTag) {
+      params.set("tag", validTag)
+    }
+    return `?${params.toString()}`
+  }
+
+  // Get contributor info for JSON-LD
+  const commitHistoryCache: CommitHistory = {}
+  const { contributors } = await getAppPageContributorInfo(
```
Comment (pettinarip) [inline app/[locale]/developers/tools/[category]/page.tsx:76]: The opening of the modal feels slow. Since this page reads the query params, it gets rendered all the type and I suspect this function is triggering every time. Not good.

app/[locale]/developers/apps/utils.ts
```diff
@@ -0,0 +1,276 @@
...
+ *
+ * Uses Next.js unstable_cache to cache the computation for 1 week.
+ * Since the input data (apps array) is already cached via getDeveloperToolsData(),
+ * this primarily caches the sorting/filtering/randomization computation.
+ *
+ * Cache key includes the function name, ensuring cache invalidation when function changes.
+ * Tagged for manual cache invalidation if needed via revalidateTag('developer-apps-highlights').
+ */
+export const getCachedHighlightsByCategory = unstable_cache(
+  async (apps: DeveloperApp[]) => getHighlightsByCategory(apps),
```
Comment (pettinarip) [inline app/[locale]/developers/apps/utils.ts]: `unstable_cache` serializes function arguments as part of the cache key. Passing apps: `DeveloperApp[]` creates:
1. Expensive serialization on every call
2. Large cache key strings
3. Potential cache misses if any app data changes

Not sure if this is even needed.

app/[locale]/developers/tools/_components/HighlightsSection.tsx
Comment (pettinarip) [inline app/[locale]/developers/tools/_components/HighlightsSection.tsx:1]: From the UI perspective,

This is `/developers/apps`
image

This is `/apps`
image

Shouldn’t we reuse the same component? They look like essentially the same thing. Also, I’d argue against making the image smaller, keeping it the same size feels more consistent with the rest of the UI.

src/lib/utils/fetch.ts
pettinarip, line 1: This is going to be tech debt. I don't like this, I feel like we should:
1. solve this at the infra level
2. listen to 429 errors and rate limit headers to retry more efficiently

app/[locale]/developers/apps/[category]/page.tsx
```diff
@@ -0,0 +1,241 @@
...
+                <SubpageCard
+                  key={slug}
+                  title={t(`page-developers-apps-category-${slug}-title`)}
+                  description={t(
+                    `page-developers-apps-category-${slug}-description`
+                  )}
+                  icon={<Icon className="size-8" />}
+                  href={`/developers/apps/${slug}`}
+                  matomoEvent={{
+                    // TODO: Confirm all
```
Comment (pettinarip) [inline app/[locale]/developers/apps/[category]/page.tsx]: confirmed?

app/[locale]/developers/apps/page.tsx
```diff
@@ -0,0 +1,219 @@
...
+        <Section id="apps" className="space-y-4">
+          <h2>{t("page-developers-apps-applications-title")}</h2>
+          <EdgeScrollContainer>
+            {DEV_APP_CATEGORIES.map(({ slug, Icon }) => (
+              <EdgeScrollItem
+                key={slug}
+                asChild
+                className="ms-6 w-[calc(100%-4rem)] max-w-md md:min-w-96 md:flex-1 lg:max-w-[33%]"
+              >
+                {/* // TODO: */}
```
Comment (pettinarip) [inline app/[locale]/developers/apps/page.tsx]: Can we cleanup these comments?

app/[locale]/developers/apps/_components/TagFilter.tsx
Comment (pettinarip) [inline app/[locale]/developers/apps/_components/TagFilter.tsx:1]: This looks very similar as the apps table with the filter. I'd strongly advocate to reuse existing components.

Comment (konopkja) [reply app/[locale]/developers/apps/_components/TagFilter.tsx:1]: the reason why we chose different dropdown here is because of UX - there are too many options to choose so basic dropdown isnt sufficient, but discussed with paul that actually original app explorer also grew and we could replace app explorer dropdown this new

app/[locale]/developers/tools/_components/HighlightsSection.tsx
Comment (wackerow) [reply app/[locale]/developers/tools/_components/HighlightsSection.tsx:1]: Yeah, as mentioned off GH, all the external images being pulled in are 4x1 aspect ratio. #17134 creates a patch for this using the same height, while improving visibility of the banners without chopping out 50% of each image

app/[locale]/developers/tools/[category]/page.tsx
Comment (pettinarip) [inline app/[locale]/developers/tools/[category]/page.tsx:1]: I think we should `export const dynamicParams = false` to throw the 404s. Right now everything is accepted:
https://page-developer-apps--ethereumorg.netlify.app/developers/tools/notexist

Comment (wackerow) [reply app/[locale]/developers/tools/[category]/page.tsx:1]: ✅

PR #17131, opened 2026-01-21
feat: simple markdown

wackerow:
## Extends `page-developer-apps` branch
- #17110

## Description

- add simple markdown parser 
  - For use with in-page markdown content parsing, not full pages with front matter
  - Simplifies use of headers to bold paragraphs
  - Accepts optional component overrides
- add stripMarkdown helper util 
  - Removes markdown syntax without parsing

### For `/developers/apps/` pages

- parse descriptions for markdown in app modal

Fixes:

image

To this:

image

- use stripMarkdown on app card descriptions

Fixes:

image

To this:

image

### DIFF
--- a/app/[locale]/developers/apps/_components/AppModalContents.tsx
+++ b/app/[locale]/developers/apps/_components/AppModalContents.tsx
@@ -37,7 +37,7 @@ const AppModalContents = async ({ app }: { app: DeveloperApp }) => {
     h5: BoldedParagraph,
     h6: BoldedParagraph,
     img: () => null,
-  } as unknown as MDXRemoteProps["components"]
+  } as MDXRemoteProps["components"]
 
   return (
     <div className="bg-background">


src/lib/utils/md.ts
```diff
@@ -139,3 +139,34 @@ export const checkPathValidity = (
...
   validPaths.some((path) => path.slug.join("/") === slugArray.join("/"))
+
+/**
+ * Strips markdown syntax from text, leaving plain text
+ * For preview/snippet text where markdown shouldn't be visible
+ *
+ * @param text - Text with markdown syntax
+ * @returns Plain text with markdown markers removed
+ */
+export function stripMarkdown(text: string): string {
```
wackerow, line 150: @pettinarip As an alternative to this, we could use [`remove-markdown`](https://www.npmjs.com/package/remove-markdown) (runtime dependency) which is what we use on the blog to clean up the preview text for the posts. It's fairly small and [a bit more robust](https://github.com/zuchka/remove-markdown/blob/main/index.js) but also potentially overkill for what we need here. Let me know if you think we should switch.

src/lib/md/renderSimple.tsx
wackerow, line 1: Trimmed this down to just compile basic markdown strings using our existing `htmlElements` as the default component styling. When consumed, we can pass custom overrides, which in this case for the App Modal I've just overridden the heading levels to prevent any semantic html issues, and hidden any potential img tags for simplicity.

@pettinarip Could also just move this inside `compile.ts` as well, and maybe rename to `compileSimple` (similar function to `compile` just without the locale/scope/slugArray handling, and doesn't deal with front matter).

PR #17157, opened 2026-01-23
perf: dynamically import Contributors to reduce bundle size

### LINKED ISSUE #17156: perf: Contributors component bundles 420KB for all MDX pages
## Problem

The `Contributors` component bundles the entire `.all-contributorsrc` file (377KB, 1,225 contributor avatars) into a 420KB client-side chunk that is loaded as an **initial chunk** for ALL MDX pages via the `[...slug]` route.

However, `Contributors` is only used on the `/contributing` page.

### Evidence

```bash
# 420KB chunk contains all 1,225 contributor avatar URLs
$ grep -o 'avatars.githubusercontent.com' .next/static/chunks/8084-*.js | wc -l
1225

# Matches exactly the source file
$ grep -o 'avatars.githubusercontent.com' .all-contributorsrc | wc -l
1225

# Chunk is in initial load for ALL MDX pages
$ cat .next/app-build-manifest.json | grep -l "8084"
# Shows 8084 IS in initial chunks for [locale]/[...slug]/page
```

### Root Cause

`src/components/Contributors/index.tsx`:
```tsx
"use client"
import data from "!!raw-loader!@/../.all-contributorsrc"
const allContributors = JSON.parse(data)
```

1. Component is `"use client"` 
2. Uses `raw-loader` to import entire JSON at build time
3. Component is exported from `MdComponents` (base components for all MDX)
4. Webpack bundles it for all pages even though only `/contributing` uses it

### Impact

- **420KB** downloaded on every MDX page load
- ~6,500+ MDX pages affected
- Users never see this data unless they visit `/contributing`

### Suggested Fix Options

1. **Move to page-specific component** (like #17153 for PieChart)
   - Remove from `MdComponents`, add to a contributing-specific layout
   
2. **Server-side data loading** (better)
   - Make Contributors a server component
   - Pass contributor data as props from the page
   - No client bundle impact

3. **Dynamic import with next/dynamic** (best)
   - Only load Contributors component on pages that use it
   - Zero impact on other pages

Related: #17153 (PieChart move to docs-specific components)

minimalsm:
## Summary

Use `next/dynamic` to lazy-load the Contributors component, avoiding bundling 420KB of contributor data for all MDX pages.

### Before
- Contributors imports `.all-contributorsrc` (377KB, 1,225 avatars) via raw-loader
- Compiled to 420KB client chunk
- Chunk was in **initial load** for ALL `[...slug]` MDX pages
- Only `/contributing` page actually uses the component

### After
- Contributors is dynamically imported
- 420KB chunk only loads when `/contributing` page is visited
- All other MDX pages (~6,500) no longer download this data

Closes #17156

### DIFF
--- a/src/components/MdComponents/index.tsx
+++ b/src/components/MdComponents/index.tsx
@@ -1,4 +1,5 @@
 import { type ComponentProps, type HTMLAttributes } from "react"
+import dynamic from "next/dynamic"
 
 import type { ChildOnlyProp } from "@/lib/types"
 
@@ -6,7 +7,8 @@ import Card from "@/components/Card"
 import { RestakingList } from "@/components/Content/restaking/RestakingList"
 import BrowseApps from "@/components/Content/what-are-apps/BrowseApps"
 import WhatAreAppsStories from "@/components/Content/what-are-apps/WhatAreAppsStories"
-import Contributors from "@/components/Contributors"
+
+const Contributors = dynamic(() => import("@/components/Contributors"))
 import DocLink from "@/components/DocLink"
 import Emoji from "@/components/Emoji"
 import ExpandableCard from "@/components/ExpandableCard"


minimalsm:
Closing - not needed. Production already handles this correctly via RSC. See #17156.

PR #17168, opened 2026-01-26
Fix Blobscan API format change breaking hourly task

pettinarip:
## Summary
- Fix Trigger.dev hourly task failing with `TypeError: Cannot read properties of undefined (reading 'totalBlobs')`
- Blobscan API changed response format from `[{stats}]` to `{data: [{dimension, metrics, updatedAt}]}`
- Update fetcher to aggregate "other" + "rollup" categories for network totals
- Simplify type to only include fields actually used (`totalBlobs`, `avgBlobFee`, `updatedAt`)

## Test plan
- [x] Verify `pnpm lint` passes
- [x] Verify Trigger.dev task runs successfully after deploy
- [x] Verify blob stats display correctly on /resources page

### DIFF
--- a/src/data-layer/fetchers/fetchBlobscanStats.ts
+++ b/src/data-layer/fetchers/fetchBlobscanStats.ts
@@ -47,9 +47,16 @@ export async function fetchBlobscanStats(): Promise<BlobscanStats> {
     throw new Error("Blobscan API missing category data")
   }
 
+  const totalBlobs = other.metrics.totalBlobs + rollup.metrics.totalBlobs
+
+  const avgBlobFee =
+    (other.metrics.avgBlobFee * other.metrics.totalBlobs +
+      rollup.metrics.avgBlobFee * rollup.metrics.totalBlobs) /
+    totalBlobs
+
   const stats: BlobscanStats = {
-    totalBlobs: other.metrics.totalBlobs + rollup.metrics.totalBlobs,
-    avgBlobFee: (other.metrics.avgBlobFee + rollup.metrics.avgBlobFee) / 2,
+    totalBlobs,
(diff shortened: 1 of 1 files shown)


src/data-layer/fetchers/fetchBlobscanStats.ts
```diff
@@ -17,18 +31,29 @@ export async function fetchBlobscanStats(): Promise<BlobscanOverallStats> {
...
+    (d) => d.dimension.type === "category" && d.dimension.name === "rollup"
+  )
+
+  if (!other || !rollup) {
+    throw new Error("Blobscan API missing category data")
+  }
+
+  const stats: BlobscanStats = {
+    totalBlobs: other.metrics.totalBlobs + rollup.metrics.totalBlobs,
+    avgBlobFee: (other.metrics.avgBlobFee + rollup.metrics.avgBlobFee) / 2,
```
wackerow: @pettinarip This `avgBlobFee` calculation weights "rollup" and "other" blob fees equally.. we should weight them by their total blob count to get a more accurate average.. consider:

```suggestion
  const totalBlobs = other.metrics.totalBlobs + rollup.metrics.totalBlobs

  const avgBlobFee =
    (other.metrics.avgBlobFee * other.metrics.totalBlobs +
      rollup.metrics.avgBlobFee * rollup.metrics.totalBlobs) /
    totalBlobs

  const stats: BlobscanStats = {
    totalBlobs,
    avgBlobFee,
```

pettinarip: Ah good catch, ty

0xsarawut (approved):

PR #17171, opened 2026-01-26
fix: include meetup groups in events search results

minimalsm:
## Summary
- Fixed count mismatch between "See all (X)" button and actual search results
- Search page now includes meetup groups from `community-meetups.json`, matching the preview component

## Problem
When searching for events by location (e.g., "Italy"):
- Preview showed 6 events with "See all (8)" button
- Clicking showed only 4 results

The preview component received combined data (API events + meetup groups), but the search page only fetched API events.

## Test plan
- [ ] Go to `/community/events`
- [ ] Search for "Italy" in "Find events near you"
- [ ] Verify count in "See all" button matches results on search page
- [ ] Verify Italian meetup groups (Milan, Venice, Naples, Udine) appear in search results

### DIFF
--- a/app/[locale]/community/events/search/page.tsx
+++ b/app/[locale]/community/events/search/page.tsx
@@ -14,7 +14,7 @@ import { getMetadata } from "@/lib/utils/metadata"
 
 import EventCard from "../_components/EventCard"
 import OrganizerCTA from "../_components/OrganizerCTA"
-import { mapEventTranslations, sanitize } from "../utils"
+import { getMeetupGroups, mapEventTranslations, sanitize } from "../utils"
 
 import { getEventsData } from "@/lib/data"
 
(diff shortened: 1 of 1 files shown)


wackerow:
image

Seems to be missing the "conference" results now when testing the "Italy" query

minimalsm:
The conferences appear to be missing in the preview because preview deploys use mock data (`src/data-layer/mocks/fetch-events.json`) which only contains 6 events—none of which are in Italy. Production uses Netlify Blobs with the full event dataset.

The fix is correct: the search page now combines API events with meetup groups, matching what FilterEvents receives. In production, searching "Italy" should show both the conferences (ETH Milan, Farcon Rome, etc.) and the meetup groups (Web3Mi, Ethereum Venezia, NapulETH, udinETH).

To verify, we'd need to either:
1. Merge and test on production/staging
2. Or add Italian events to the mock data file for local testing

nloureiro:
Backlog-sweep note: this is a real bug fix (meetups missing from events search, causing the count mismatch) that only needs a rebase — it's been conflicted since February. @minimalsm still planning to land it, or should someone adopt it?

nloureiro:
Closing in favor of #18692, which adopts this fix (with credit) updated for the current `getMeetupGroups(locale)` / `mapEventTranslations(..., locale)` signatures. You diagnosed this correctly — the search page was the only events surface not merging meetup groups, and it still was until today. Thanks Joe! 👋

PR #17174, opened 2026-01-27
Move GitHub contributors fetching to scheduled data-layer task

pettinarip:
## Summary
- Replace per-request GitHub API calls with pre-fetched data stored in Netlify Blobs
- Add `fetchGitHubContributors` scheduled task running weekly to fetch all contributor data
- Use git/trees API for single-request content file listing (replaces recursive directory walks)
- Fetch commits in parallel batches (20 concurrent requests) for performance
- Update `contributors.ts` to read from data-layer instead of making live API calls
- Bump `@trigger.dev/sdk` to 4.3.3

## Test plan
- [x] Verify the scheduled task runs successfully and stores data in Netlify Blobs
- [x] Verify contributor information displays correctly on content pages
- [x] Verify contributor information displays correctly on app pages
- [x] Verify build succeeds without errors

### DIFF
--- a/src/data-layer/fetchers/fetchGitHubContributors.ts
+++ b/src/data-layer/fetchers/fetchGitHubContributors.ts
@@ -115,20 +115,21 @@ async function fetchCommitsForPath(
     return []
   }
 
-  // Transform to FileContributor format and deduplicate
-  const contributors = commits
-    .filter((commit: { author?: unknown }) => !!commit.author)
-    .map(
-      (commit: {
-        author: { login: string; avatar_url: string; html_url: string }
-        commit: { author: { date: string } }
-      }) => ({
-        login: commit.author.login,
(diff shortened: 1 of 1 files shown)


pettinarip:
@wackerow thanks for the review! all three items addressed.

`new Date().toISOString()` fallback — now it hides the date entirely when contributor data is missing. `lastEditLocaleTimestamp` is optional throughout the type chain and the UI conditionally renders it. Showing any fallback date felt misleading regardless of source.

Additionally, investigating the missing dates uncovered a deeper bug: commits authored with emails not linked to a GitHub account return `author: null` from the API. The fetcher was filtering these out entirely, so pages with only such commits (e.g. `community/support/faq`) had no contributor data at all. Fixed by falling back to the git commit author name when the GitHub author is null.

PR #17175, opened 2026-01-27
fix(pl): remove typo/garbage text in defi page

Pytkopff:
Removed "oSearch TMut" from the "What is DeFi?" section in the Polish translation. It appears to be corrupted text or a copy-paste error.


wackerow:
@all-contributors please add @Pytkopff for bug patch

PR #17184, opened 2026-01-28
Add homepage persona modal A/B test variant

pettinarip:
## Summary
- Adds a new A/B test variation for the homepage CTAs section
- Replaces the four CTA grid with a single "Start here" button that opens a persona selection modal
- Modal allows users to self-identify as beginners, developers, or enterprise and routes them to relevant pages
- Also updates the hero section title/description for the persona variant

## Test plan
- [x] Verify original variant (four CTAs grid) still renders correctly
- [x] Verify persona variant displays "Start here" button
- [x] Test modal opens on button click and tracks Matomo event
- [x] Test each persona link navigates correctly and closes modal
- [x] Verify external links open in new tab with proper attributes
- [x] Test responsive layout on mobile and desktop
- [ ] Verify A/B test debug panel shows correct variants

### DIFF
--- a/app/[locale]/page.tsx
+++ b/app/[locale]/page.tsx
@@ -12,12 +12,14 @@ import type {
 } from "@/lib/types"
 import { CodeExample } from "@/lib/interfaces"
 
+import ABTestWrapper from "@/components/AB/TestWrapper"
 import ActivityStats from "@/components/ActivityStats"
 import { ChevronNext } from "@/components/Chevron"
 import HomeHero from "@/components/Hero/HomeHero"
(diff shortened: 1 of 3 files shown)


src/components/AB/TestDebugPanel.tsx
pettinarip, line 1: maybe out of scope but this is setting the debug panel to render in a portal to avoid pending an extra hidden node in the place where the ABTestWrapper is used.

pettinarip:
New variant approved by @konopkja. Merging after getting a successful build.

PR #17221, opened 2026-02-03
fix(data-layer): add CoinGecko demo API key to prevent rate limiting

pettinarip:
## Summary
- Added `x_cg_demo_api_key` parameter to all CoinGecko API calls to prevent 429 rate limiting errors
- Updated ETH price, ETH market cap, and stablecoins fetchers to use `COINGECKO_API_KEY` environment variable
- Added `COINGECKO_API_KEY` to `.env.example`

## Test plan
- [x] Verify `COINGECKO_API_KEY` is set in production environment
- [x] Confirm ETH price fetcher task runs successfully without 429 errors
- [x] Confirm ETH market cap fetcher task runs successfully
- [x] Confirm stablecoins fetcher runs successfully

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -27,6 +27,9 @@
 # Dune Analytics API key (required for total eth staked)
 # DUNE_API_KEY=
 
+# CoinGecko API
+# COINGECKO_API_KEY=
+
 # Matomo environment (URL and site ID required for analytics)
 NEXT_PUBLIC_MATOMO_URL=
 NEXT_PUBLIC_MATOMO_SITE_ID=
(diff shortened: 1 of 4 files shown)


.env.example
```diff
@@ -27,6 +27,9 @@
 # Dune Analytics API key (required for total eth staked)
 # DUNE_API_KEY=
 
+# CoinGecko API
+# COINGECKO_API_KEY=
```
wackerow, line 31: Not seeing this in Netlify, should it be?

pettinarip: It is set in the trigger.dev env where this fetches are used.

PR #17228, opened 2026-02-04
Equal bug bounty pay for german hunters.

riproprip:
For whatever reason the bug bounty amount on the german page is 50k usd while other languages have 250k.

This adjusts that.


wackerow:
@all-contributors please add @riproprip for maintenance

PR #17299, opened 2026-02-13
[MUST MERGE BEFORE DEPLOY] Revert emergency rate-limit fix — restore contributor fetch with proper solution

minimalsm:
## DO NOT MERGE YET — needs a proper fix first

I pushed a hotfix directly to dev (84c0dceed3) that removes the `while(true)` rate limit retry loop in `fetchWithRateLimit()`. This was blocking every single build — the loop would wait 22+ minutes for GitHub API rate limits to reset, triggering Next.js's 60-second worker timeout and killing all static page generation.

**This PR reverts that hotfix**, restoring the original blocking behavior. It should NOT be merged until a proper solution is implemented that:

1. Adds a module-level cache so contributor fetches are shared across locale renders (currently 25 locales × same paths = 25× redundant API calls)
2. Adds in-flight request deduplication
3. Handles rate limits gracefully instead of blocking or silently dropping data

## Why the hotfix was needed

- 39 pages call `getAppPageContributorInfo`, each making up to 12 GitHub API calls
- 25 locales build independently with no shared cache → ~11,700 API calls per build
- GitHub rate limit is 5,000/hour → exhausted immediately
- `while(true)` loop blocks all workers for 1,300+ seconds waiting for reset
- Next.js kills workers after 60 seconds, restarting in an infinite loop

Build log:
```
8:23:20 PM: Rate limit exceeded, waiting for 1331 seconds (×25)
8:24:20 PM: ⚠ Sending SIGTERM signal to Next.js build worker due to timeout of 60 seconds
8:24:20 PM: ⚠ Restarted static page generation for /en/10years...
(every page, every locale)
```

## What's degraded on dev right now

Contributor info on app pages may be incomplete if the GitHub API rate limit is hit during build. Pages still render fine — the contributor display handles empty arrays.

## Test plan

- [ ] Implement proper caching/dedup solution
- [ ] Verify builds complete with contributor data intact
- [ ] Then merge this revert

### DIFF
--- a/src/lib/utils/gh.ts
+++ b/src/lib/utils/gh.ts
@@ -122,29 +122,33 @@ async function fetchWithRateLimit(filepath: string): Promise<Commit[]> {
   // If no token available, return empty array
   if (!gitHubToken) return []
 
-  const response = await fetch(url.href, {
-    headers: { Authorization: `token ${gitHubToken}` },
-  })
+  /* eslint-disable no-constant-condition --
+   * eslint does not like while(true)
+   **/
+  while (true) {
+    const response = await fetch(url.href, {
+      headers: { Authorization: `token ${gitHubToken}` },
+    })
 
-  if (
-    response.status === 403 &&
-    response.headers.get("X-RateLimit-Remaining") === "0"
-  ) {
-    console.warn(`GitHub API rate limit exceeded for ${filepath}. Skipping.`)
-    return []
-  }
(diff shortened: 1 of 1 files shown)


wackerow:
`fetchWithRateLimit` has been deprecated; `getGitHubContributors` is now handled via data-layer, found in `src/lib/data/index.ts` -- closing out

PR #17562, opened 2026-02-15
fix(i18n): set correct robots meta for untranslated locale pages

minimalsm:
## Summary
- Flips `index: true` to `index: false` for untranslated locale pages
- Canonical tag pointing to the English version is preserved
- `follow: true` is kept so crawlers still follow links

## Context
Untranslated locale pages (e.g., `/fr/some-page/` showing English content) were explicitly being indexed, creating massive duplicate content across 60+ locale prefixes. This will reduce indexed page count significantly (expected and correct).

## Test plan
- [x] Visit an untranslated locale page — should have `<meta name="robots" content="noindex, follow">`
- [x] Translated pages are unaffected (still indexed)
- [x] Canonical tag to English version is still present

### DIFF
--- a/src/lib/utils/metadata.ts
+++ b/src/lib/utils/metadata.ts
@@ -141,7 +141,7 @@ export const getMetadata = async ({
   }
 
   if (!isCurrentPageTranslated) {
-    return { ...base, robots: { index: true, follow: true } }
+    return { ...base, robots: { index: false, follow: true } }
   }
 
   return base


mnelsonBT:
Hi @minimalsm - the 'Pages to Noindex sheet indicates "Remove canonical tag and add meta noindex tag" for rows 7 - 2040 (rows 2040 - 5581 are app URLs, which only have "Add meta noindex tag," likely because they were never marked with a canonical)

PR #17564, opened 2026-02-15
refactor(nav): use CSS responsive classes instead of client-only rendering

minimalsm:
## Summary
- Removes `ClientOnly` and `MediaQuery` wrappers from the header navigation
- Both desktop and mobile nav are now in the server-rendered HTML
- CSS responsive classes (`hidden md:contents` / `md:hidden`) control visibility
- Deletes `src/components/Nav/loading.tsx` (skeleton fallbacks no longer needed)

## Context
The entire header navigation was wrapped in `ClientOnly` + `MediaQuery`, meaning Google saw zero nav links — only skeleton placeholders. This replaces JS conditional rendering with standard CSS responsive design.

## Test plan
- [ ] View page source — all nav `<a>` tags present in SSR HTML
- [ ] No hydration mismatch errors in console
- [ ] Mobile: hamburger menu still opens/closes
- [ ] Desktop: dropdown menus still work
- [ ] Visual regression check (skeleton states removed)


mnelsonBT:
Hi @minimalsm, did the search feature work correctly in both viewports when you built this out locally? 
It looks like both mobile & desktop nav use the same <Search /> component, which could mean the search would be duplicated if both menus are rendering in the DOM. 

Otherwise LGTM though if there were no issues

pettinarip:
Closing as this is not rendering the links on the HTML since the navs are mounted client-side on user's interaction. Alternative to cover these goals and keep current perf optimizations #17928

PR #17566, opened 2026-02-15
feat(footer): add homepage link

minimalsm:
## Summary
- Adds an "Ethereum.org" link pointing to `/` above the footer link sections
- Renders as a crawlable `<a>` tag on every page

## Test plan
- [x] Footer contains a visible "Ethereum.org" link to `/`
- [x] Link is a crawlable `<a>` tag in page source
- [x] Appears on all pages

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -333,6 +333,15 @@ const Footer = ({ lastDeployLocaleTimestamp }: FooterProps) => {
         </Button>
       </div>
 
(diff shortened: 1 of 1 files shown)


wackerow:
image

just an opinion, but this looks off... doesn't really match the theming of anything else in the footer. If main goal is SEO, could we consider making this the same font size and color as the other footer nav links? Also, branding technique has been to use lowercase ethereum.org cc: @minimalsm @konopkja

PR #17589, opened 2026-02-18
perf: replace dynamic imports with fs.readFile for i18n and md content

pettinarip:
## Summary

- Replace dynamic `import()` with `fs/promises.readFile()` for loading translation JSON files and markdown content
- Reduces webpack memory consumption during build by avoiding chunk creation for thousands of potential import paths

## Motivation

The build was exceeding memory limits on Netlify after recent i18n imports added translations for Bengali, Polish, Urdu, Ukrainian, German, Marathi, and Turkish.

**Root cause**: Dynamic imports like `await import(.../${locale}/${ns}.json)` force webpack to analyze and create chunks for all possible paths at build time:
- 60+ locales × 50 namespaces = ~3,000 JSON import possibilities
- 4,255 MD files with locale variants = thousands more

**Solution**: Use `fs.readFile()` which is a runtime operation, bypassing webpack's module resolution entirely.

## Test plan

- [x] Verify build completes without memory errors on Netlify
- [x] Verify translations load correctly on dev and production
- [x] Verify markdown content renders correctly

### DIFF
--- a/src/lib/i18n/loadMessages.ts
+++ b/src/lib/i18n/loadMessages.ts
@@ -1,14 +1,19 @@
 import fs from "fs"
+import { readFile } from "fs/promises"
 import path from "path"
+
+interface IntlMessages {
+  [key: string]: string | IntlMessages
+}
 
 function getNamespaces(localePath: string): string[] {
   return fs
-    .readdirSync(localePath)
-    .filter((file) => file.endsWith(".json"))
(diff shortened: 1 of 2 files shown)


Blondie0123:
Replace dynamic import() with fs/promises.readFile() for loading translation JSON files and markdown content

wackerow (approved):

PR #17599, opened 2026-02-19
fix(seo): remove duplicate @id in tutorials page JSON-LD

wackerow:
## Summary
- The tutorials page JSON-LD had two `ItemList` nodes sharing the same `@id` (`#tutorials`), which is invalid in a JSON-LD `@graph`
- Merged into a single `ItemList` with `ListItem` entries wrapping nested `Course` items, preserving both list and course structured data signals
- Uses per-tutorial `skill` for `educationalLevel` instead of a hardcoded `"beginner-intermediate"`


myelinated-wackerow:
## PR #17599 Review: `fix(seo): remove duplicate @id in tutorials page JSON-LD`

**Author**: wackerow | **File**: `app/[locale]/developers/tutorials/page-jsonld.tsx` | **+17 / -31**

### What it does

The existing code has two `ItemList` nodes in the `@graph` array that share the identical `@id` (`${url}#tutorials`). That's invalid JSON-LD — when two nodes in a `@graph` share an `@id`, parsers/crawlers may only see one or behave unpredictably. One list had plain `ListItem` entries, the other had `Course` entries.

The fix merges them into a single `ItemList` where each `ListItem` wraps a nested `Course` via the `item` property. This is structurally correct per schema.org — `ListItem.item` is the canonical way to attach a typed entity to a list position.

### What's good

1. **Correct fix for a real SEO bug.** Duplicate `@id` in a `@graph` is genuinely invalid and could confuse Google's structured data parser.
2. **Net deletion of code** (-14 lines). Simpler is better.
3. **`educationalLevel` now uses `tutorial.skill`** instead of the hardcoded `"beginner-intermediate"` string. The `Skill` enum values (`"beginner"`, `"intermediate"`, `"advanced"`) are valid `educationalLevel` values per schema.org. The `?? "beginner"` fallback is reasonable since `skill` is optional on `ITutorial`.

### Issues / Questions

1. **`description` dropped from the outer `ListItem`.**
   The old first `ItemList` had `description` on each `ListItem`. The new code removes it from the `ListItem` level (it's only on the nested `Course` now). This is fine per spec — `description` lives better on the `Course` — but worth flagging as intentional.

2. **`position` removed from inner `Course`.**
   The old second `ItemList` had `position` on each `Course`. Now `position` only exists on the `ListItem` (which is correct — `position` belongs to `ListItem`, not `Course`). This is actually a fix of a minor misuse, so this is good.

3. **`numberOfItems` still references `internalTutorials.length` but list is `.slice(0, 10)`.**
   This was a pre-existing issue, not introduced by this PR. The `numberOfItems` says e.g. 150 but only 10 items are in the array. Worth a follow-up but not a blocker.

4. **`"beginner-intermediate"` was not a valid schema.org value anyway.** The old code used a made-up compound level. Switching to the enum values is strictly better. No concern here.

### Verdict

**Approve.** The fix is correct, minimal, and improves the structured data validity. The only pre-existing oddity is the `numberOfItems` mismatch (item #3 above), which I'd suggest filing separately.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code) (Opus 4.6)

PR #17635, opened 2026-02-23
patch: tutorials list ui and header issues

wackerow:
## Description
- Cleans up classNames for UI for tutorials list
- Fixes mobile spacing around skill tag and search input.
- Fixes header levels, adding `sr-only` h2 when results are available, with h3 used for title
- Deprecated redundant `Text` component; `space-y-6` used from parent instead
- i18n: provides translations for new `sr-only` h2 string

## Preview link
https://deploy-preview-17635.ethereum.it/developers/tutorials

cc: @konopkja

### DIFF
--- a/app/[locale]/developers/tutorials/_components/tutorials.tsx
+++ b/app/[locale]/developers/tutorials/_components/tutorials.tsx
@@ -27,8 +27,9 @@ import Translation from "@/components/Translation"
 import { getSkillTranslationId } from "@/components/TutorialMetadata"
 import TutorialTags from "@/components/TutorialTags"
 import { Button } from "@/components/ui/buttons/Button"
-import { Flex, FlexProps } from "@/components/ui/flex"
+import { Flex } from "@/components/ui/flex"
 import Input from "@/components/ui/input"
+import { BaseLink } from "@/components/ui/Link"
 import TabNav from "@/components/ui/TabNav"
 import { Tag, TagButton } from "@/components/ui/tag"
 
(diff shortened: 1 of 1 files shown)


wackerow:
## Critical Issue Found                                                                                                                                                                                                                                                                                             

Empty `<h2>` tag at the top of the container — This is the main problem I see.

At the top of the tutorials wrapper (line ~273 in the diff), there's:
`<h2 className="sr-only"></h2>`

This heading has no text content. It's always rendered regardless of filter state. Screen readers will announce "heading level 2" with nothing after it — that's an accessibility anti-pattern and arguably worse than having no heading at all. Contrast this with the conditional `<h2>` further down that correctly uses `{t("page-tutorials-available")}`. I'd guess the intent was to put a translated label here too but it was missed.

## Minor Observations (non-critical)

1. block on a LinkFlex component — The class changed from flex-col justify-between to block space-y-6. Since LinkFlex wraps a Flex component (which likely sets display: flex), the block class needs to actually override that. Worth verifying it renders correctly and doesn't create a specificity conflict with Tailwind.
2. text-border removed from LinkFlex — The old classes had text-border on the card links. Removed without replacement. If that was providing text color on the card, the color may now fall through to an inherited value. Likely intentional given the PR description, but worth a visual check on the preview.
3. Bengali (bn) file — Has a lot of invisible character normalization changes (zero-width joiners or similar). Not harmful, but not really a content change either.

## Everything Else Looks Clean

- Removing the Text component and using space-y-6 from the parent is a good simplification
- `<p>` to `<h3>` for tutorial titles improves heading semantics correctly
- Ternary replacing && for the empty state is fine
- All 25 i18n files correctly add the new "page-tutorials-available" key
- Unused cn and HTMLAttributes imports properly cleaned up

---
_Reviewed by Claude (Opus 4.6)_

wackerow:
Patched above issues
- Removes stray h2
- Since "LinkFlex" was being forced back to a block, removed it's usage and replaced with proper BaseLink component
- Refactored to implicit return for arrow function mapping
- Reverted superfluous `bn` translation updates, good to go

PR #17637, opened 2026-02-23
docs(revoke-access): correct wrong Blockscout URL

Uaitt:
The [how to revoke smart contract access article](https://ethereum.org/guides/how-to-revoke-token-access/) provides a Blockscout URL that returns 404. This PR provides the right URL instead.


wackerow:
@all-contributors please add @Uaitt for maintenance

PR #17651, opened 2026-02-24
docs(stablecoins): correct typo

Uaitt:


src/intl/en/page-stablecoins.json
```diff
@@ -137,7 +137,7 @@
   "page-stablecoins-usdt-banner-swap-button": "Swap ETH for USDT",
   "page-stablecoins-usdt-banner-learn-button": "Learn about USDT",
   "page-stablecoins-usdt-logo": "The USDT logo",
-  "page-stablecoins-usds-banner-body": "USDS is the successor to Dai, fully backed by crypto and designed for onchain savings and rewards. Widely used in DeFi for while keeping users in full control of their funds.",
+  "page-stablecoins-usds-banner-body": "USDS is the successor to Dai, fully backed by crypto and designed for onchain savings and rewards. Widely used in DeFi for keeping users in full control of their funds.",
```
wackerow: good catch; small nit-pick, would ditch "for" and keep "while"


```suggestion
  "page-stablecoins-usds-banner-body": "USDS is the successor to Dai, fully backed by crypto and designed for onchain savings and rewards. Widely used in DeFi while keeping users in full control of their funds.",
```

wackerow: "while keeping" reads more naturally — it conveys two concurrent facts: USDS is widely used in DeFi, and it keeps users in control. The "while" links them smoothly.

"for keeping" in A is slightly awkward because it implies the reason USDS is widely used is specifically for the purpose of keeping users in control. That's a narrower claim and reads a bit clunky — "widely used for keeping" isn't a natural construction.

_Reviewed by Claude_

wackerow:
@all-contributors add @Uaitt for bug patch

PR #17661, opened 2026-02-25
perf: lazy-load mobile menu content to reduce RSC payload

pettinarip:
## Summary

- Move mobile menu content from server component to client-side lazy-loaded component
- This avoids shipping all navigation translations and language display info in the initial RSC payload since the mobile menu is hidden behind a click
- Saves ~82KB from the initial HTML payload by deferring navigation data until the user actually opens the mobile menu

## Changes

- **MobileMenuContent.tsx**: New client component with the actual menu content
- **MobileMenuClient.tsx**: Uses `React.lazy()` to load content on first open
- **MobileMenu/index.tsx**: Simplified to just render MobileMenuClient
- **LvlAccordion.tsx**: Remove `async` keyword that caused repeated Suspense
- **useLanguagesDisplayInfo.ts**: New client hook to generate language info

## Test plan

- [x] Verify mobile menu still opens and functions correctly
- [x] Verify language selection works properly
- [x] Verify navigation links work in mobile menu
- [ ] Check that initial page load doesn't include mobile menu data in RSC payload
- [x] Test accordion navigation levels open/close correctly

### DIFF
--- a/src/components/Nav/MobileMenu/MobileMenuClient.tsx
+++ b/src/components/Nav/MobileMenu/MobileMenuClient.tsx
@@ -93,7 +93,6 @@ const MobileMenuClient = ({ className, side }: MobileMenuClientProps) => {
           ref={triggerRef}
           className={cn("-me-2", className)}
           isMenuOpen={open}
-          onPointerEnter={() => lazyImport()}
         />
       </SheetTrigger>
 
@@ -110,19 +109,19 @@ const MobileMenuClient = ({ className, side }: MobileMenuClientProps) => {
           <ErrorBoundary
             fallback={({ reset }) => (
               <div className="flex flex-1 flex-col items-center justify-center gap-4 p-8 text-center">
-                <p className="text-body-medium">Failed to load menu</p>
+                <p className="text-body-medium">{t("loading-error")}</p>
                 <div className="flex gap-3">
                   <button
                     className="rounded-md bg-primary px-4 py-2 text-sm text-white hover:bg-primary-hover"
                     onClick={reset}
                   >
-                    Try again
+                    {t("try-again")}
                   </button>
                   <button
                     className="rounded-md border border-body-light px-4 py-2 text-sm text-body hover:bg-background-highlight"
                     onClick={() => handleOpenChange(false)}
                   >
-                    Close
+                    {t("close")}
                   </button>
                 </div>
               </div>
--- a/src/intl/en/common.json
+++ b/src/intl/en/common.json
@@ -471,6 +471,7 @@
   "translation-program": "Translation Program",
   "translator": "Translator",
   "trillion-dollar-security": "Trillion dollar security project",
+  "try-again": "Try again",
   "try-using-search": "Try using search to find what you're looking for or",
   "tutorials": "Tutorials",
   "up": "Up",


wackerow (approved):
@pettinarip Thanks for this! Looks great overall, left the claude review above.. Curious your take on the Error boundary. The skeleton looks fine imo. Approving as-is, but ofc feel free to update if you think it's worth it.

pettinarip:
@wackerow thanks for the review. I've applied most of them. Also, improved a bit the skeleton.

src/components/Nav/MobileMenu/MobileMenuClient.tsx
```diff
@@ -12,41 +14,125 @@ import HamburgerButton from "./HamburgerButton"
 import { useCloseOnNavigate } from "@/hooks/useCloseOnNavigate"
 import { useTranslation } from "@/hooks/useTranslation"
 
+// Lazy-load the menu content to avoid including it in initial RSC payload
+// This saves ~82KB by not SSR'ing navigation data that's hidden behind a click
+const lazyImport = () => import("./MobileMenuContent")
+const MobileMenuContent = React.lazy(lazyImport)
+
+function MobileMenuContentSkeleton() {
```
wackerow, line 22: Nice, looks great now 🔥

src/components/Nav/MobileMenu/MobileMenuClient.tsx
```diff
@@ -12,41 +14,125 @@ import HamburgerButton from "./HamburgerButton"
...
+  // Track if menu has ever been opened to keep content loaded after first open
+  const [hasBeenOpened, setHasBeenOpened] = React.useState(false)
+
+  // Prefetch the menu chunk after the page is idle
+  React.useEffect(() => {
+    if (hasBeenOpened) return
+    if (typeof window.requestIdleCallback === "function") {
+      const id = window.requestIdleCallback(() => lazyImport())
+      return () => window.cancelIdleCallback(id)
+    }
```
wackerow, line 78: Solid

src/components/Nav/MobileMenu/MobileMenuClient.tsx
```diff
@@ -12,41 +14,125 @@ import HamburgerButton from "./HamburgerButton"
...
 
   return (
-    <Sheet open={open} onOpenChange={setOpen}>
+    <Sheet open={open} onOpenChange={handleOpenChange}>
       <SheetTrigger asChild>
         <HamburgerButton
           ref={triggerRef}
           className={cn("-me-2", className)}
           isMenuOpen={open}
+          onPointerEnter={() => lazyImport()}
```
wackerow: Tiny note, not sure how much this will do on a mobile device, but perhaps it jumpstarts the process on "tap down" for mobile? Either way, not doing any harm

src/components/Nav/MobileMenu/MobileMenuClient.tsx
```diff
@@ -12,41 +14,125 @@ import HamburgerButton from "./HamburgerButton"
...
         triggerRef={triggerRef}
         aria-label={t("site-title")}
         data-testid="mobile-menu-dialog"
       >
-        {children}
+        {hasBeenOpened && (
+          <ErrorBoundary
+            fallback={({ reset }) => (
+              <div className="flex flex-1 flex-col items-center justify-center gap-4 p-8 text-center">
+                <p className="text-body-medium">Failed to load menu</p>
```
wackerow: Looks good, just not quite sure how to test this. 

Only note would be use of English strings here. I think this is an edge case so don't think it needs to block the PR, but separately we may want to consider i18n here.

src/components/Nav/MobileMenu/MobileMenuClient.tsx
pettinarip: Ah, good call! this doesn’t make sense for mobile. I’ll remove it and stick with automatic prefetching during browser idle event.

src/components/Nav/MobileMenu/MobileMenuClient.tsx
pettinarip: The only way to manually test this would be to run it locally and throw an error from the lazy loaded component, in this case the content component.

re: i18n, true. I'll double check if we can reuse an existing string. otherwise, I'll add one

PR #17662, opened 2026-02-25
perf: remove unused keys from common.json namespace

pettinarip:
## Summary

- Remove 40 unused translation keys from common.json that are not referenced anywhere in the codebase
- Move 4 page-specific keys to their respective namespaces (page-community-events, page-get-eth)

## Changes

- **40 unused keys removed** from common.json across all 26 locales
- **Keys relocated** to proper namespaces:
  - `item-logo`, `support` → page-community-events.json
  - `listing-policy-disclaimer` → page-get-eth.json

## Impact

Reduces the common namespace payload by ~4KB per locale, improving initial page load performance.

## Test plan

- [x] Verify build passes
- [x] Spot check pages that use relocated keys (community events, get-eth)

### DIFF
--- a/src/intl/ar/common.json
+++ b/src/intl/ar/common.json
@@ -448,6 +448,7 @@
   "what-is-ethereum": "ما هو إيثريوم؟",
   "what-is-the-ethereum-network": "ما هي شبكة الإيثريوم؟",
   "withdrawals": "عمليات سحب تجميد العملات",
+  "wrapped-eth": "Wrapped Ether",
   "yes": "نعم",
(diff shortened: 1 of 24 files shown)


myelinated-wackerow (changes requested):
_duplicate -- removed_

myelinated-wackerow (dismissed):
_duplicate -- removed_

wackerow:
Actually, a follow-up to that last comment... A better solution for that page specifically is to simply remove the alt text:

image

These icons are decorative only and should be `alt=""` anyway. I can patch that so we can get this in.

PR #17671, opened 2026-02-26
docs: correct Github typo across whole website

Uaitt:
Github should actually be written as GitHub. This typo appears in many places across the whole website, even in different translations.

That's why there are many files changed :)


wackerow (commented):
Hey @Uaitt! Sorry for delay here.. appreciate this polish. 

We've ended up with a bunch of merge conflicts, but they're primarily from the non-English version of these pages.

For context, a translation pipeline is being implemented for the site which will soon be run to generate a fresh set of translations based on the English. Subsequent to that it should detect changes to the English that keeps the rest up to date, and we'll be mostly hands-off the non-English content files.

As such, I would **recommend we revert the changes to the non-English files in here, and keep it to English only**. The changes won't propagate to non-English immediately, but will be coming shortly. That should bring the diff down to about five (5) files -- I can help take care of that to get this over the line.

---

If you're curious, when the translations are run again, the other files should be corrected with the new setup, enforced with the following:

https://github.com/ethereum/ethereum-org-website/blob/f027e11672c5362e0b69e38cbab94a8ae92c40f4/src/scripts/intl-pipeline/intl-sanitizer.ts#L744-L754

wackerow (approved):
Went ahead and reverted the non-English changes. Thanks again @Uaitt ! Pulling in

wackerow:
@all-contributors please add @Uaitt for bug fix

PR #17672, opened 2026-02-26
SEO - missing image alt text 

mnelsonBT:
Resolves missing alt text on images across ethereum.org 

## Description

- Added alt props directly to React component source code (changes auto-propagate to all locale pages)
- Updated type definitions where needed (e.g. BentoItem) to accommodate the new alt field 
- Identified ~36 unique (page, image) pairs in locale-specific markdown files missing alt tex
- Translated approved English alt text into all affected locales and applied translated alt text to each per-locale


app/[locale]/bug-bounty/page.tsx
```diff
@@ -159,58 +159,69 @@ export default async function Page({ params }: { params: Promise<Params> }) {
       title: "Besu",
       link: "https://besu.hyperledger.org/en/stable/",
       image: besu,
+      alt: "Besu logo",
```
Comment (wackerow) [inline app/[locale]/bug-bounty/page.tsx]: Image

⚠️ These `clients`, `specs` and `languages` arrays are the ones that render here in this screenshot, and already have associated text for the brand. In these cases I'd suggest keeping them off (at `alt=""` which is the default fallback).

app/[locale]/bug-bounty/page.tsx
```diff
@@ -305,39 +318,47 @@ export default async function Page({ params }: { params: Promise<Params> }) {
         </Text>
         <ClientRow>
           <Client>
-            <Image src={besu} alt="" {...iconImageProps()} />
+            <Image src={besu} alt="Besu logo" {...iconImageProps()} />
```
Comment (wackerow) [inline app/[locale]/bug-bounty/page.tsx]: Image

👍 These look good, as they're associated with the eleven (11) client logos that do not have any associated text already, nested under "Clients featured in the bounties"

Comment (wackerow) [reply app/[locale]/bug-bounty/page.tsx]: BUT: We also need to keep non-English in mind. We have an entry in the `common` namespace `"item-logo": "{item} logo",` that allows us to use the `item-logo` key and provide an `item` name (the brand name) so it interpolates into each language accordingly.

Currently, this means we need to load the common namespace first

```tsx
// below: const t = await getTranslations({ namespace: "page-bug-bounty" })
const tCommon = await getTranslations({ namespace: "common" })
```

And then refactor the `alt` attributes to:

```tsx
// Using Erigon as an example:
alt={tCommon("item-logo", { item: "Erigon" })}
```

This namespace may undergo some changes soon (see #17662) but in the meantime, if we want to do this correctly, that would be the approach.

src/components/Homepage/BentoCard.tsx
wackerow, line 1: Would revert these additions ✂️

src/components/Homepage/utils.ts
wackerow, line 1: Same ✂️ 

The component this data is passed to have `alt=""` hard coded already, so not needed

app/[locale]/community/events/page.tsx
```diff
@@ -219,7 +219,7 @@ const Page = async ({ params }: { params: PageParams }) => {
                       <div className="grid size-fit shrink-0 place-items-center overflow-hidden rounded-full">
                         <Image
                           src={banner}
-                          alt=""
+                          alt={`${location} Ethereum community hub banner`}
```
Comment (wackerow) [inline app/[locale]/community/events/page.tsx]: Image

Would revert this one back... decorative logo, with a title already available 

```suggestion
                          alt=""
```

wackerow (changes requested):
cc: @mnelsonBT Dropped some comments.. Lemme know if you need a hand!

mnelsonBT:
> @mnelsonBT Lgtm... Applied those patches. Take another look and if you're good with this can pull in for next weeks deploy

LGTM @wackerow - thank you for doing so!

PR #17682, opened 2026-02-27
perf: deduplicate common namespace serialization across layout and page providers

pettinarip:
## Summary

- Eliminate double-serialization of the `common` i18n namespace in HTML payload
- Layout provider serializes "common" once; page providers now merge parent messages instead of replacing them
- No changes needed to any of the 29 page files — the merge happens automatically in `I18nProvider`

## Problem

The `common` namespace (~26KB) was serialized twice on every page:
1. Layout-level `NextIntlClientProvider` in `providers.tsx` serialized it for layout components
2. Page-level `I18nProvider` also included it via `getRequiredNamespacesForPage()`

Per next-intl behavior, nested providers treat `messages` as atomic replacements — the page provider's messages completely replaced the layout's. This forced pages to re-include "common".

## Solution (3 files)

1. **`providers.tsx`** — Use `NextIntlClientProvider` directly as root provider (no merging needed)
2. **`I18nProvider.tsx`** — Call `useMessages()` to get parent provider's messages and merge with page-specific messages on the client
3. **`translations.ts`** — Remove "common" from `getRequiredNamespacesForPage()` base namespaces

## How it works

```
Layout: serializes { common: {...} } once → NextIntlClientProvider
  └── Page: serializes only { "page-wallets": {...} } → I18nProvider
        ├── useMessages() returns { common: {...} } from parent
        ├── merged = { common: {...}, "page-wallets": {...} }
        └── NextIntlClientProvider gets merged messages
```

## Test plan

- [x] TypeScript type check passes (`pnpm tsc --noEmit`)
- [x] Production build succeeds (`pnpm build`)
- [ ] Desktop nav: click through all 5 sections, verify labels render
- [ ] Mobile nav: open hamburger menu, verify labels, theme toggle, search
- [ ] Search modal: verify all UI strings on desktop and mobile
- [ ] FeedbackWidget: click floating dot, verify prompt and yes/no buttons
- [ ] RTL locale (e.g. `/ar/`): verify nav direction and layout
- [ ] HTML payload: verify "common" namespace appears only once

### DIFF
--- a/app/[locale]/providers.tsx
+++ b/app/[locale]/providers.tsx
@@ -1,8 +1,7 @@
 "use client"
 
-import { type AbstractIntlMessages } from "next-intl"
+import { type AbstractIntlMessages, NextIntlClientProvider } from "next-intl"
 
-import I18nProvider from "@/components/I18nProvider"
 import ThemeProvider from "@/components/ThemeProvider"
 import { TooltipProvider } from "@/components/ui/tooltip"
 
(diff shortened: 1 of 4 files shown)


pettinarip:
Closing this — re-measured on current `dev` with a proper before/after build and it's basically a no-op now (~0 gzip delta, `common` still serialized twice either way). React's flight serializer already dedupes the page provider's `common`, so the double-serialization this targeted is gone.

The real duplication that's left is the *layout* provider's `common` shipping twice in the flight stream — different problem, chasing that one separately.

PR #17691, opened 2026-03-02
Doces:Update applying-storybook.md

buzhidao380:


wackerow:
If you're learning git, please mention as such, else please avoid spam; only warning 🙏

PR #17583, opened 2026-02-16
Redesign /community/support/ as crisis-first support hub

konopkja:
## Summary
- Replaces the static markdown FAQ (73% negative rating, No.1 worst page on site) with a component-based card layout focused on the actual audience: scam victims (~40%), confused newcomers (~25%), and users with wallet issues (~15%)
- Adds dedicated subpages for scams, FAQ, and common misconceptions with actionable guidance
- Removes developer/node operator sections (content already lives at /developers/ and /run-a-node/) since <5% of visitors are developers
- Uses modern Card components with colored accent shadows (matching /stablecoins/ editors choice pattern)

## Context

Analysis of 17 direct Paperform submissions and 143 cross-site support-seeking submissions showed the page was designed for developers but visited almost exclusively by people in crisis. The emotional mismatch was the core design failure — a scam victim who lost their life savings would land on a page listing Alchemy University and CryptoDevs Discord.

### What changed
- **New**: `app/[locale]/community/support/` — server + client component replacing the markdown-only page
- **New**: `public/content/community/support/scams/index.md` — scam help, reporting, fund tracing
- **New**: `public/content/community/support/faq/index.md` — wallet, transaction, staking FAQ
- **New**: `public/content/community/support/misconceptions/index.md` — identity clarification
- ~~**Modified**: `public/content/community/support/index.md` — simplified to subpage links~~ (Removed old markdown lander in lieu of react page)

### Page structure
1. Hero with identity clarification (ethereum.org is not a wallet/exchange/platform)
2. Decentralization callout
3. **Get help** — "Something went wrong" + "Protect yourself" (colored shadows)
4. **Learn** — "Using Ethereum" + "Common misconceptions" (colored shadows)
5. "Still need help?" CTA pointing to ethereum.org Discord
6. Feedback card

## Test plan
- [x] Verify page renders at `/community/support/`
- [x] Verify all 3 subpages render: `/scams/`, `/faq/`, `/misconceptions/`
- [x] Verify all internal links resolve correctly
- [x] Verify external links (revoke.cash, Chainabuse, IC3, etc.) are live
- [x] Check dark mode rendering
- [x] Check mobile responsiveness
- [x] Verify colored card shadows display correctly in both themes

## Preview link
https://deploy-preview-17583.ethereum.it/community/support

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #17694, opened 2026-03-02
i18n: extract support page strings

myelinated-wackerow:
## Extends branch
- #17583

## Summary
- Created `page-community-support.json` namespace with 39 translation keys for the `/community/support/` page
- Replaced all hardcoded English strings in `page.tsx` with `getTranslations()` calls using `next-intl`
- Registered `/community/support/` prefix in `translations.ts` namespace mapping (before the `/community/` catch-all)

## TODO
- [x] Add non-English translations

## Question for @konopkja

The `customEventOptions` values (`eventCategory`, `eventAction`, `eventName`) are currently hardcoded English strings used as analytics identifiers. Should these remain as-is (since they're tracking keys, not user-facing text), or would you prefer them extracted to the i18n namespace as well?

## Test plan
- [x] Verify page renders identically with translated strings
- [x] Confirm `generateMetadata` pulls title/description from namespace
- [x] Check that `/community/support/` resolves to `page-community-support` namespace


konopkja:
tracking - Keep them as hardcoded English strings.

konopkja:
as always though im looking at english only

PR #17721, opened 2026-03-05
fix(seo): fix JSON-LD structured data on /wallets/find-wallet/

### LINKED ISSUE #17719: fix(seo): JSON-LD structured data contains untranslated i18n keys on /wallets/find-wallet/
## Description

The JSON-LD structured data on `/wallets/find-wallet/` outputs raw i18n translation keys instead of translated text. This breaks rich result eligibility and sends malformed data to search engines.

### Current (broken)

```json
{
  "@type": "WebPage",
  "name": "page-find-wallet-meta-title",
  "description": "page-find-wallet-meta-description"
}
```

The BreadcrumbList and Article schema have the same problem:
- `BreadcrumbList` item #3 name: `page-find-wallet-meta-title`
- `Article.headline`: `page-find-wallet-title`
- `Article.description`: `page-find-wallet-meta-description`

### Expected

```json
{
  "@type": "WebPage",
  "name": "List of Ethereum Wallets",
  "description": "Find and compare Ethereum wallets based on the features you want."
}
```

### The HTML meta tags are fine

The regular `<title>`, `<meta name="description">`, and Open Graph tags all output correctly translated text. The bug is specifically in the JSON-LD generation path — the translation function is applied to HTML meta tags but skipped when the same values are injected into the JSON-LD `<script type="application/ld+json">` block.

### Impact

- Google's Rich Results testing tool would flag this as malformed structured data
- The page is ineligible for enhanced search appearances (rich snippets, knowledge panels)
- Raw i18n keys in structured data may signal low quality to search engines

### Likely affected scope

This is probably not limited to `/wallets/find-wallet/` — any page using the same JSON-LD generation pattern with translatable fields may have untranslated keys in structured data. Worth auditing site-wide.

### Suggested fix

In the component that generates the JSON-LD script tag, ensure the translation function (e.g., `t()` or equivalent) is called on all translatable fields before they are serialized into the JSON-LD object.

konopkja:
## Summary

- **Fix i18n namespace bug**: `page-jsonld.tsx` used namespace `"page-find-wallet"` instead of `"page-wallets-find-wallet"`, causing all translation keys to output as raw strings (e.g. `"page-find-wallet-meta-title"` instead of `"List of Ethereum Wallets"`) in the JSON-LD script tag
- **Replace incorrect `Article` schema type**: the find-wallet page is an interactive comparison tool, not an article. Changed to `CollectionPage` + `ItemList` with `SoftwareApplication` entries per wallet, matching the pattern already used on `/layer-2/networks/`
- **Add wallet metadata to structured data**: each wallet now includes platform availability, open source status, self custody, hardware wallet support, L2 support, staking, and NFT support as `additionalProperty` values

## Context

The HTML `<title>` and `<meta>` tags were correctly translated — only the JSON-LD block had the wrong namespace, so the bug was invisible in the browser but broken for search engines.

The `Article` → `CollectionPage` + `ItemList` + `SoftwareApplication` change follows the same approach as `app/[locale]/layer-2/networks/page-jsonld.tsx` and gives Google actionable structured data about the actual wallets on the page.

Closes #17719

## Test plan

- [ ] Verify JSON-LD output contains translated text (not raw i18n keys) by viewing page source or using [Google Rich Results Test](https://search.google.com/test/rich-results)
- [ ] Verify `CollectionPage` and `ItemList` render correctly in structured data
- [ ] Verify wallet count in `numberOfItems` matches displayed wallets
- [ ] Check non-English locale (e.g. `/es/wallets/find-wallet/`) also outputs translated JSON-LD

### DIFF
--- a/app/[locale]/wallets/find-wallet/page-jsonld.tsx
+++ b/app/[locale]/wallets/find-wallet/page-jsonld.tsx
@@ -52,6 +52,7 @@ export default async function FindWalletPageJsonLD({
         "@id": url,
         name: t("page-find-wallet-meta-title"),
         description: t("page-find-wallet-meta-description"),
+        image: "https://ethereum.org/images/wallets/wallet-hero.png",
         url: url,
         inLanguage: locale,
         contributor: contributorList,


myelinated-wackerow:
## Review Summary

### What this PR fixes well

**Namespace bug fix** -- The core fix (`"page-find-wallet"` -> `"page-wallets-find-wallet"`) is correct and important. The `page-find-wallet` namespace doesn't exist as a translation file, so every `t()` call in the JSON-LD was emitting raw key strings to search engines. The browser-visible `<title>` and `<meta>` tags were unaffected since `page.tsx` and `generateMetadata` already used the correct namespace -- making this bug invisible to humans but broken for crawlers.

**Schema type change** -- Replacing `Article` with `CollectionPage` + `ItemList` of `SoftwareApplication` entries is semantically accurate for an interactive comparison tool, and consistent with the existing `layer-2/networks/page-jsonld.tsx` pattern.

**Wallet metadata in structured data** -- Adding platform availability, open source status, self custody, L2 support, staking, and NFT support as `additionalProperty` values gives Google actionable structured data about each wallet.

---

### Suggestions before merging

**1. `chromium` maps to `"Chrome OS"` -- this is inaccurate**

```tsx
if (wallet.chromium) os.push("Chrome OS")
```

The `chromium` field indicates a Chromium browser extension, not Chrome OS operating system support. Consider `"Chromium (Extension)"` or similar. As-is, this would misrepresent wallet capabilities to search engines.

**2. Dropped `image` property -- SEO regression**

The old `Article` node had `image: "https://ethereum.org/images/wallets/wallet-hero.png"`. The new schema drops this entirely. Google's structured data guidelines favor entries with images for rich results. Consider keeping an `image` on the `CollectionPage` node, or adding per-wallet images using `wallet.image` on each `SoftwareApplication`.

**3. Minor: `CollectionPage` vs `WebPage` inconsistency with reference**

The first `@graph` node was changed from `"WebPage"` to `"CollectionPage"`. The layer-2/networks reference keeps `"WebPage"`. `CollectionPage` is arguably more accurate here (it's a valid subtype), but the two list pages should probably be consistent -- either update both or keep both as `WebPage`.

Items 1 and 2 are worth addressing before merge. Item 3 is minor.

---

*Reviewed by Claude (Opus 4.6)*

PR #17740, opened 2026-03-09
Increase Sentry pageload sample rate for reliable Web Vitals

pettinarip:
## Summary
- Replace flat `tracesSampleRate: 0.01` with a `tracesSampler` function in `instrumentation-client.ts`
- Pageload transactions sampled at 10% (up from 1%) for more reliable Web Vitals data
- All other transactions remain at 1% to avoid quota impact

## Context
Sentry Web Vitals insights showed many pages with exactly "100 pageloads" — this was Sentry extrapolating from a single captured trace (1 × 1/0.01 = 100). Drilling into individual pages confirmed only 1 actual trace, making performance scores statistically unreliable.

## Test plan
- [ ] Verify Sentry Web Vitals insights show more pageload samples after deploy
- [ ] Monitor Sentry quota usage to ensure 10% pageload sampling doesn't exceed limits


wackerow:
bump @pettinarip

pettinarip:
@wackerow thanks. Here is the analysis I made and test to double check it works

---

The conclusion doesn't hold for `@sentry/nextjs` specifically.

The review correctly traces the generic `@sentry/browser` code path, but `@sentry/nextjs` takes a different path — it explicitly includes `sentry.op` in the `attributes` object before the span is created, so it's available at sampling time.

Verified locally by logging `samplingContext` inside `tracesSampler`:

```
{
  name: '/',
  sentry.op: 'pageload',
  sentry.origin: 'auto.pageload.nextjs.app_router_instrumentation',
  sentry.source: 'url'
}
```

The current implementation works as intended.

PR #17741, opened 2026-03-09
fix: prevent dev/staging subdomains from being indexed by search engines

pettinarip:
## Summary

- Adds `X-Robots-Tag: noindex, nofollow` HTTP headers in `netlify.toml` for branch deploys, deploy previews, and the `dev`/`staging` branches
- Sets `NEXT_PUBLIC_SITE_URL` per branch (`dev` → `https://dev.ethereum.org`, `staging` → `https://staging.ethereum.org`) so canonical URLs and noindex meta tags resolve correctly at build time
- Extracts `IS_PRODUCTION_DEPLOY` constant from `SITE_URL` hostname check, used by both `robots.ts` and `metadata.ts` to gate indexing

### Problem

`SITE_URL` was resolving to `https://ethereum.org` on dev/staging deploys at SSR runtime because `DEPLOY_PRIME_URL` (a build-time-only Netlify env var) was unavailable, falling through to `URL` which always points to the production domain. This caused:

1. **"Duplicate without user-selected canonical"** — dev/staging pages had canonical URLs pointing to `ethereum.org`, creating cross-domain duplicate signals
2. **"Indexed, though blocked by robots.txt"** — `robots.txt` correctly blocked crawling but pages lacked `<meta name="robots" content="noindex">`, so Google indexed URLs it couldn't crawl

### Fix layers

| Layer | Mechanism | Protects against |
|-------|-----------|-----------------|
| `X-Robots-Tag` header | HTTP header via `netlify.toml` | Indexing of any resource (HTML, PDF, images) |
| `NEXT_PUBLIC_SITE_URL` per branch | Build-time env var | Wrong canonical URLs, missing noindex meta tags |
| `IS_PRODUCTION_DEPLOY` check | Code-level hostname comparison | Meta robots tag omission on non-production |

## Test plan

- [ ] Verify `dev.ethereum.org` returns `X-Robots-Tag: noindex, nofollow` header
- [x] Verify `dev.ethereum.org` pages have `<meta name="robots" content="noindex, nofollow">`
- [x] Verify `dev.ethereum.org` canonical URLs point to `dev.ethereum.org`, not `ethereum.org`
- [ ] Verify `staging.ethereum.org` has the same protections
- [ ] Verify `ethereum.org` (production) is unaffected — no noindex, correct canonicals

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -63,8 +63,8 @@
   [context.staging.headers.values]
     X-Robots-Tag = "noindex, nofollow"
 
-# Set NEXT_PUBLIC_SITE_URL per branch so canonical URLs and noindex
-# meta tags resolve correctly at build time.
+# Override SITE_URL for named branches with custom subdomains so canonical
+# URLs and OG metadata use the custom domain instead of *.netlify.app.
 [context.dev.environment]
   NEXT_PUBLIC_SITE_URL = "https://dev.ethereum.org"
 
--- a/next.config.js
+++ b/next.config.js
@@ -35,9 +35,18 @@ module.exports = (phase, { defaultConfig }) => {
     ...defaultConfig,
     reactStrictMode: true,
     env: {
-      // Context is used to determine the environment for Sentry
+      // Netlify build-time vars inlined so they're available at SSR runtime.
       // ref. https://docs.netlify.com/configure-builds/environment-variables/#build-metadata
       NEXT_PUBLIC_CONTEXT: process.env.CONTEXT,
+      // Resolve site URL once at build time. NEXT_PUBLIC_SITE_URL (set in
+      // netlify.toml for dev/staging) takes precedence, then Netlify's
+      // deploy-specific URLs, falling back to the production domain.
+      NEXT_PUBLIC_SITE_URL:
+        process.env.NEXT_PUBLIC_SITE_URL ||
+        process.env.DEPLOY_PRIME_URL ||
+        process.env.DEPLOY_URL ||
+        process.env.URL ||
+        "https://ethereum.org",
     },
     webpack: (config) => {
       config.module.rules.push({
--- a/src/lib/constants.ts
+++ b/src/lib/constants.ts
@@ -24,13 +24,9 @@ export const LOCALES_CODES = BUILD_LOCALES
   ? BUILD_LOCALES.split(",")
   : i18nConfig.map(({ code }) => code)
 
-// Site urls - auto-detect from Netlify deploy context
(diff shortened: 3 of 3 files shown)


myelinated-wackerow (approved):
Nice PR -- the layered approach (HTTP header + build-time env + code-level check) is the right strategy here. Two suggestions:

### 1. Prefer `const` with IIFE over `export let`

Minor, but `export let` allows external reassignment. An IIFE keeps it immutable:

```ts
export const IS_PRODUCTION_DEPLOY = (() => {
  try {
    return new URL(SITE_URL).hostname === "ethereum.org"
  } catch {
    return false
  }
})()
```

### 2. Code-level protection may leak on arbitrary branch deploys and deploy previews

I'm fairly confident (though not 100%) that for branches other than `dev` and `staging`, the code-level layer doesn't fire correctly. Since `NEXT_PUBLIC_SITE_URL` is only set for those two branches, arbitrary branch deploys and deploy previews would fall through the `SITE_URL` chain. `DEPLOY_PRIME_URL` / `DEPLOY_URL` / `URL` are build-time-only Netlify vars -- I believe they're unavailable at SSR runtime in Netlify Functions -- so `SITE_URL` would resolve to `"https://ethereum.org"`, making `IS_PRODUCTION_DEPLOY` true. That means no `noindex` meta tag, `robots.txt` allows crawling, and canonical URLs point to `ethereum.org`.

The `X-Robots-Tag` HTTP header from `context.branch-deploy` / `context.deploy-preview` does block indexing at the HTTP level, so these deploys aren't unprotected. But the canonical URL leak could still send confusing duplicate signals to crawlers.

One possible fix: rather than deriving `IS_PRODUCTION_DEPLOY` from `SITE_URL`, use a dedicated boolean that only production sets:

```toml
[context.production.environment]
  NEXT_PUBLIC_IS_PRODUCTION = "true"
```

```ts
export const IS_PRODUCTION_DEPLOY = process.env.NEXT_PUBLIC_IS_PRODUCTION === "true"
```

Since it's `NEXT_PUBLIC_*`, it gets inlined at build time. It defaults to `false` everywhere except production, and doesn't touch `SITE_URL` -- so no risk of breaking AB testing, OG images, JSON-LD, or other `SITE_URL`-dependent functionality.

That said, this is a suggestion -- the HTTP header layer already covers the indexing concern, so this is about closing the canonical URL gap for non-dev/staging deploys. Your call on whether that's worth the extra env var.

---

*Reviewed by Claude Opus 4.6*

pettinarip:
@wackerow thanks. I refactor the code to stop using the SITE_URL to calc production env flag...I just use the flag that we already have available. Safer and simpler.

PR #17745, opened 2026-03-09
Update dev docs smart-contracts/security

hidargmax27-cmyk:
Added a detailed section explaining the differences, pros, and cons of Transparent Proxy and UUPS (EIP-1822) patterns to address function clashing in smart contract upgrades. Inserted after the basic proxy explanation to provide deeper architectural context for developers.

### DIFF
--- a/public/content/developers/docs/smart-contracts/security/index.md
+++ b/public/content/developers/docs/smart-contracts/security/index.md
@@ -135,6 +135,25 @@ Accounts interact with the proxy contract, which dispatches all function calls t
 
 Delegating calls to the logic contract requires storing its address in the proxy contract's storage. Hence, upgrading the contract's logic is only a matter of deploying another logic contract and storing the new address in the proxy contract. As subsequent calls to the proxy contract are automatically routed to the new logic contract, you would have “upgraded” the contract without actually modifying the code.
 
+### Advanced Proxy Patterns: Transparent vs. UUPS {#advanced-proxy-patterns}
+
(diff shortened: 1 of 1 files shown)


myelinated-wackerow (changes requested):
Thanks for the contribution! The content itself is solid -- good explanation of Transparent vs. UUPS patterns with a clear pros/cons structure. However, this needs to live on a different page.

**Wrong page**: This content belongs on the [upgrading smart contracts page](/developers/docs/smart-contracts/upgrading/) rather than the security page. The security page intentionally keeps the proxy discussion brief as part of "disaster recovery plans" and links out to the upgrading page for details. That upgrading page already covers proxy patterns (Mechanism #3) and even links to an article on Transparent vs. UUPS in its Further Reading. Adding this here duplicates that material and breaks the disaster recovery narrative flow.

**Suggestion**: Rework this as an expansion of the "Proxy patterns" section on `/developers/docs/smart-contracts/upgrading/index.md` instead. It would fit naturally after the existing proxy discussion there.

A few additional notes for the revision:

- **Heading hierarchy**: The `###` (H3) heading is nested inside a `####` (H4) section, which breaks document structure. Make sure heading levels are consistent with wherever this lands.
- **Function clashing description**: The EVM doesn't face ambiguity here -- Solidity matches the proxy's local function by selector and never hits the fallback. The issue is that the logic contract's matching function becomes *unreachable*, not that the EVM "wouldn't know" which to call.
- **Tone**: "Highly gas-inefficient" and "catastrophic risk" are a bit strong for this site's voice. Consider dialing back slightly.
- **Formatting**: The blank line before the `[More on upgrading contracts]` link got removed, which would merge it into the preceding paragraph.

---
Reviewed by Claude Opus 4.6

nloureiro:
Checking in as part of a PR-backlog sweep: the review feedback from March 10 (this content belongs on a different page) is still unanswered. If you'd like to continue, please respond or update within 14 days — otherwise we'll close this. Thanks!

wackerow:
@hidargmax27-cmyk Closing this one out. Content on this topic would still be useful, but this would need a re-write. If you want to go further with the contribution, I would recommend opening an issue to draft an outline of what you're looking to add, with details on where you'd like it to go for feedback first. Thanks again!

PR #17755, opened 2026-03-11
fix: add ErrorBoundary to Search modal and fix MobileMenu error recovery

pettinarip:
## Summary

- Wrap the dynamically-imported `SearchModal` with an `ErrorBoundary` so that a `ChunkLoadError` shows a recovery dialog ("Refresh page" / "Close") instead of crashing the entire app to a black screen
- Fix the `MobileMenu` error fallback to use `window.location.reload()` instead of `reset`, since `reset` just re-triggers the same failed dynamic import

## Context

The webpack runtime in Next.js 14 App Router uses `[chunkhash]` instead of `[contenthash]` ([vercel/next.js#78756](https://github.com/vercel/next.js/issues/78756)), so its filename can stay the same across deploys even when chunk mappings change. Browsers that cached the old runtime serve stale chunk hashes, causing 404s on lazy imports like `SearchModal`. This is fixed upstream in Next.js 15.3.4+ ([vercel/next.js#80153](https://github.com/vercel/next.js/pull/80153)).

## Test plan

- [x] Click Search button — modal opens normally
- [x] Simulate chunk load failure (e.g. block the chunk URL in DevTools) — error dialog appears with "Refresh page" and "Close" buttons instead of black screen
- [x] Click "Close" — dialog dismisses, page remains usable
- [x] Click "Refresh page" — page reloads
- [x] Open mobile menu — menu works normally
- [x] Simulate chunk failure on mobile menu — same recovery dialog behavior

### DIFF
--- a/src/components/Nav/MobileMenu/MobileMenuClient.tsx
+++ b/src/components/Nav/MobileMenu/MobileMenuClient.tsx
@@ -107,15 +107,15 @@ const MobileMenuClient = ({ className, side }: MobileMenuClientProps) => {
       >
         {hasBeenOpened && (
           <ErrorBoundary
-            fallback={({ reset }) => (
+            fallback={() => (
               <div className="flex flex-1 flex-col items-center justify-center gap-4 p-8 text-center">
                 <p className="text-body-medium">{t("loading-error")}</p>
                 <div className="flex gap-3">
                   <button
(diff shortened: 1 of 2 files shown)


wackerow:
Looks like we may be duplicating much of the error fallback between `Search` and `MobileMenu` (same structure, slightly different container styles). Could consider extracting that into a shared component (e.g., `ChunkLoadErrorFallback`)... just an idea, didn't need to block

PR #17756, opened 2026-03-11
Add new bounty hunter 'Revofusion' - 4x High 40000 points

0xMushow:


wackerow:
@all-contributors add @0xMushow for maintenance

PR #17776, opened 2026-03-13
fix: withdrawal credentials widget

wackerow:
## Description
- fix: removes client-side fetch logic from `WithdrawalCredentials`; replace with direct link to `beaconcha.in/validator/<index>#deposits` (mainnet) and `hoodi.beaconcha.in/validator/<index>#deposits` (testnet)
- content: update preceding paragraph text with guidance (UI label strings are intentionally hard-coded in English to match destination site)
- remove deprecated strings
- i18n: update translations for new string (Gemini, utilizing eth glossary)

image

## Related Issue
image

### DIFF
--- a/src/components/Staking/WithdrawalCredentials.tsx
+++ b/src/components/Staking/WithdrawalCredentials.tsx
@@ -20,9 +20,8 @@ const WithdrawalCredentials: FC = () => {
 
   const mainnetHref = `https://beaconcha.in/validator/${inputValue}#deposits`
   const testnetHref = `https://hoodi.beaconcha.in/validator/${inputValue}#deposits`
-  const disabledClass = !inputValue.length
-    ? "pointer-events-none opacity-50"
-    : ""
+  const isDisabled = !inputValue.length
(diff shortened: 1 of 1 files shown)


src/components/Staking/WithdrawalCredentials.tsx
```diff
@@ -150,28 +35,22 @@ const WithdrawalCredentials: FC = () => {
...
-            {isLoading.mainnet && <Spinner />}
-          </Button>
-          <Button
-            onClick={() => checkWithdrawalCredentials(true)}
-            disabled={!inputValue.length}
+          </ButtonLink>
+          <ButtonLink
+            href={testnetHref}
             variant="outline"
+            className={disabledClass}
```
pettinarip, line 50: this one is tricky for screen readers and tabbing because the link exists and the readers will announce them.

would recommend either conditionally render the href (pass `href={inputValue ? mainnetHref : undefined}`) or add `aria-disabled="true"` and `tabIndex={-1}` alongside the CSS classes.

wackerow:
Cerberus, block Pum482930

PR #17797, opened 2026-03-18
feat: hide contributors if edit button hidden

wackerow:
## Description
- Hides `FileContributors` component on pages marked with `hideEditButton`

Used on only a few pages, the `hideEditButton` flag is available for certain more-protected pages which intentionally limit community contributions for editing. In these cases, showing the file contributors component feels less appropriate.

### Current affected pages
- /community/events/organizing/
- /cookie-policy/
- /foundation/
- /privacy-policy/
- /terms-of-use/
- /whitepaper/

cc: @konopkja


wackerow:
Side note, the current usage appears to be exclusively via the "static" layout. I've applied to patch to the three layouts that contained this component, but raises the question if it's worth even having `hideEditButton` as a front matter prop for any template other than static (i.e., docs or tutorials). I don't see a case when we'd want to hide that button for anything using those layouts, so maybe worth just removing.

PR #17815, opened 2026-03-20
Raise minimum cache revalidation to daily, move gas table to client-side API

pettinarip:
## Summary
- Raised all data-layer cache revalidation intervals from hourly to daily (86,400s) to reduce CDN cache miss rate and improve TTFB
- Added `fetchGasPrice` to the data layer (hourly Trigger.dev task fetching from Etherscan), so gas data is always available in Netlify Blobs
- Created `/api/gas-eth-price` API route (hourly revalidation) that combines gas price from the new data source with ETH price from CoinGecko (single source of truth)
- Converted `GasTable` to a client component that renders with server-provided initial data, then refreshes client-side via the API route
- Translations are resolved server-side and passed as props to avoid client i18n namespace issues

## Test plan
- [x] Verify homepage, staking, stablecoins, layer-2, and resources pages load correctly
- [x] Verify `/what-is-ether` page renders gas table with data on initial load (no flash of "—")
- [x] Confirm `/api/gas-eth-price` returns valid `gasPrice` and `ethPriceUSD`
- [x] Check that no pages have revalidation intervals below daily
- [x] Verify gas table works in dev with mock data (`src/data-layer/mocks/fetch-gas-price.json`)

### DIFF
--- a/app/[locale]/what-is-ether/_components/GasTable.tsx
+++ b/app/[locale]/what-is-ether/_components/GasTable.tsx
@@ -45,7 +45,7 @@ const GasTable = ({ labels, locale, initialData }: GasTableProps) => {
       .then((fresh) => {
         if (fresh) setData(fresh)
       })
-      .catch(() => {})
+      .catch((err) => console.warn("Failed to refresh gas/ETH price", err))
   }, [])
 
   const calculateCost = (gasUnits: number) => {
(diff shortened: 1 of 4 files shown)


pettinarip:
@wackerow thanks for the review! one adjustment: instead of bumping the route revalidation to daily, I kept it at 1hr and created hourly-cached versions of the getters (`getHourlyEthPrice`, `getHourlyGasPriceData`) so the client API actually serves fresh data each hour.

PR #17823, opened 2026-03-22
Letting your gasless users hold tokens and call contracts

### LINKED ISSUE #17614: Tutorial: Letting your gasless users hold tokens and call contracts
### Tutorial title

Letting your gasless users hold tokens and call contracts

### Tutorial description

A previous article (#17575) discussed using gasless access to your own application using EIP-712 signatures. However, this only works for direct access to your own smart contracts.  Using [account abstraction](https://docs.erc4337.io/index.html), we can create smart contract wallets that accept two types of transactions, and relay them to a requested destination:

1. Transactions sent by a specific EOA (which require ETH)
2. Transactions sent from anywhere, but signed by the same EOA.

This way, we can provide a gasless way for an account to hold assets (tokens, etc.) and perform all the functions an EOA with gas can. This is not centralized, because any relay can relay the transactions, and without a relay it's still possible to do everything directly.

### Tutorial tags

gasless, account abstraction

### Skill level

Intermediate

### Hosted on ethereum.org or hosted elsewhere?

Hosted on ethereum.org

### For tutorials to be hosted on Ethereum.org: Tutorial Content

1. Introduction: Why is this useful
1.1. Why certain requests cannot be relayed natively (a.k.a. why do we check `msg.sender` rather than `tx.origin`.
2. The wallet smart contract
3. Building a relayer
4. The user interface
5. Conclusion: When is this enough? The advantages and disadvantages of 7702 

### For tutorials hosted elsewhere: URL to tutorial

_No response_

- [x] Yes
- [ ] No

qbzzt:
## Description

New tutorial: Using account abstraction we can create smart contract wallets that accept transactions that either were sent by a specific EOA, or which were signed by that EOA. These smart contracts can then own tokens, which are under the EOA's control.

[You can see the rendred version here](https://deploy-preview-17823.ethereum.it/developers/tutorials/gasless-token).

## Preview link
https://deploy-preview-17823.ethereum.it/developers/tutorials/gasless-token

## Related Issue
- Closes #17614


qbzzt:
@minimalsm, @wackerow, @pettinarip, sorry to bug you, but could you please review this?

public/content/developers/tutorials/gasless-token/index.md
```diff
@@ -0,0 +1,788 @@
...
+
+  const proxyTransfer = async () => {
+    const calldata = encodeFunctionData({
+      abi: Erc20.abi,
+      functionName: 'transfer',
+      args: [transferTo, BigInt(transferAmount) * 10n**18n],
+    })
+
+    const {v, r, s} = await signMessage(proxyAddr, calldata)
+    messageUserProxy(proxyAddr, faucetAddr, calldata, v, r, s)
```
wackerow: `proxyTransfer` passes `faucetAddr` as the target, but earlier (where `transferToken` is introduced) the prose says "This may be FaucetToken, but it does not have to be."

As written, the "Token to transfer" input is ignored on the proxy path and transfers always hit `FaucetToken`. Is the intent that the target should be `transferToken` here, or should the prose be narrowed to say the proxy transfer is FaucetToken-only?

public/content/developers/tutorials/gasless-token/index.md
```diff
@@ -0,0 +1,788 @@
...
+
+   ```sh
+   git clone https://github.com/qbzzt/260315-gasless-tokens.git
+   cd 260315-gasless-tokens
+   forge build
+   cd server
+   npm install
+   ```
+
+3. Edit `.env` to set `SEPOLIA_PRIVATE_KEY` to a wallet that has ETH on Sepolia. If you need Sepolia ETH, [use a faucet](https://cloud.google.com/application/web3/faucet/ethereum/sepolia) to get it. Ideally, this private key should be different from the one you have in your browser wallet.
```
wackerow: Mind if we keep the faucet suggestion away from Google, and internal to our own listings?

```suggestion
3. Edit `.env` to set `SEPOLIA_PRIVATE_KEY` to a wallet that has ETH on Sepolia. If you need Sepolia ETH, [use a faucet](/developers/docs/networks/#sepolia) to get it. Ideally, this private key should be different from the one you have in your browser wallet.
```

wackerow (commented):
Thanks @qbzzt!

See commits 1f0f2f2d87d478e37a09fc5a09ab60d8f5452829 for a few clerical patches (2434c70de85defd2658770fee3e91a4a15cf7618 contains some auto-formatting)

Dropped a couple questions then I think we can get this in

public/content/developers/tutorials/gasless-token/index.md
qbzzt: Thank you. I'll fix it over the weekend.

public/content/developers/tutorials/gasless-token/index.md
wackerow: ❤️

public/content/developers/tutorials/gasless-token/index.md
qbzzt: Thank you. Fixed in both source code and article.

wackerow (approved):
Looks good! Thanks @qbzzt!

PR #17832, opened 2026-03-23
fix(data-layer): beaconchain 429 rate limit resilience

### LINKED ISSUE #17825: [Recovery Agent] fetch-beaconchain persistent 429 rate limiting (66 failures/24h)
## Error Summary

- **Source:** trigger-dev
- **Item ID:** trigger-fetch-beaconchain-beaconcha-in-epoch-responded-with-status-n
- **Task:** fetch-beaconchain
- **Status:** 429 (Too Many Requests)
- **Hit count:** 66 failed runs in 24 hours (3 retry attempts per run)
- **First seen:** 2026-03-23T15:00:04.844Z
- **Dashboard:** https://cloud.trigger.dev/projects/v3/proj_jnkvoplayqkrsnovqsdr/runs/run_cmn3bba7k4ayu0on2iksz4qcz

## Analysis

The `fetch-beaconchain` task is scheduled hourly and fetches data from two Beaconcha.in API endpoints (`/api/v1/epoch/latest` and `/api/v1/ethstore/latest`). The task is consistently failing with HTTP 429 on the epoch endpoint before even reaching the ethstore call.

With 66 failures over 24 hours (the task runs hourly with up to 3 retries = ~72 max attempts), this indicates **nearly 100% failure rate** — the API key's rate limit is being exceeded persistently.

The fetcher already includes a 1-second delay between the two API calls (`fetchBeaconChain.ts:42`), but since the 429 occurs on the **first** request, the inter-request delay is irrelevant. The issue is likely at the hourly scheduling level or the API key's overall rate limit tier.

**Stack trace:**
```
Error: Beaconcha.in epoch responded with status 429
    at fetchBeaconChain (src/data-layer/fetchers/fetchBeaconChain.ts:35:11)
```

## Affected Files

- `src/data-layer/fetchers/fetchBeaconChain.ts` — fetcher implementation
- `src/data-layer/tasks.ts` — task scheduling (hourly, line 92)

## Suggested Approach

Several mitigation options to investigate:

1. **Check API key status** — Verify the `BEACONCHAIN_API_KEY` is still valid and check its rate limit tier on Beaconcha.in
2. **Reduce fetch frequency** — Move from hourly to every 2-4 hours; beacon chain data (validators, staked ETH, APR) doesn't change rapidly enough to justify hourly fetches
3. **Add graceful fallback** — On 429, return the last cached/stored value instead of throwing, so stale data is served rather than no data
4. **Increase retry backoff** — Current retry config is 2-30s which is too aggressive for 429s; consider longer initial delay (e.g., 60s minimum) with proper `Retry-After` header parsing
5. **Deduplicate with fetchTotalEthStaked** — `fetchTotalEthStaked.ts` may also be hitting Beaconcha.in, contributing to rate limit exhaustion

## Confidence Assessment

Low confidence because:
- Root cause is external (Beaconcha.in rate limiting), not a code bug per se
- Multiple valid fix approaches exist with different tradeoffs
- Need to verify API key status before determining the right solution
- Reducing frequency or adding fallback behavior are architectural decisions that need team input

---

pettinarip:
## Summary
- Add shared `fetchRetry` utility wrapping Trigger.dev's `retry.fetch` with automatic 429 and 5xx backoff retry
- Replace raw `fetch()` with `fetchRetry` in beaconchain fetcher, removing the hardcoded 1s sleep between calls
- Move beaconchain task from hourly to daily schedule (1,000 req/month quota was being exhausted at ~1,440 req/month)
- Align task-level and global retry timeouts

Closes #17825

## Test plan
- [x] Verified `fetchRetry` retries 429 responses with exponential backoff (2 attempts)
- [x] Verified beaconchain task runs on daily schedule
- [x] TypeScript compilation passes
- [ ] Lint and prettier pass

### DIFF
--- a/src/data-layer/fetchers/fetchBeaconChain.ts
+++ b/src/data-layer/fetchers/fetchBeaconChain.ts
@@ -5,11 +5,13 @@ import type {
   MetricReturnData,
 } from "@/lib/types"
 
+import { fetchRetry } from "./fetchRetry"
+
 export type BeaconChainData = BeaconchainEpochData & { apr: MetricReturnData }
 
 /**
  * Fetch beaconchain data from Beaconcha.in API.
- * Combines epoch and ethstore endpoints (sequential to respect 1 req/sec limit).
+ * Combines epoch and ethstore endpoints sequentially.
  */
 export async function fetchBeaconChain(): Promise<BeaconChainData> {
   const base = "https://beaconcha.in"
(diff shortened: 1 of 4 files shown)


src/data-layer/tasks.ts
```diff
@@ -86,10 +86,10 @@ const DAILY: TaskDef[] = [
   [KEYS.EVENTS, fetchEvents],
   [KEYS.DEVELOPER_TOOLS, fetchDeveloperTools],
   [KEYS.TRANSLATION_GLOSSARY, fetchTranslationGlossary],
+  [KEYS.BEACONCHAIN, fetchBeaconChain],
```
pettinarip, line 89: most important change, from hourly to daily

src/data-layer/fetchers/fetchBeaconChain.ts
```diff
@@ -27,31 +29,26 @@ export async function fetchBeaconChain(): Promise<BeaconChainData> {
...
-    const error = `Beaconcha.in epoch responded with status ${status}`
-    throw new Error(error)
+    throw new Error(`Beaconcha.in epoch responded with status ${status}`)
   }
   const epochJson: EpochResponse = await epochResponse.json()
   const { validatorscount, eligibleether } = epochJson.data
   const totalEthStaked = Math.floor(eligibleether * 1e-9)
 
-  // Wait 1s to respect rate limit
-  await new Promise((r) => setTimeout(r, 1000))
```
wackerow, line 42: Worth double-checking: with this sleep removed, the ethstore fetch fires immediately after the epoch fetch. If beaconcha.in enforces 1 req/sec, the second call will routinely 429 and depend on `fetchRetry` backoff to recover. We sure this will work as expected?

pettinarip: Yes, tested. Its already working as expected.

PR #17836, opened 2026-03-24
feat/fix: update gemini-translations

wackerow:
## Description

- **Code comment translation bug fixes**: fence language labels with metadata (e.g., `sh copy`) now correctly map to their comment syntax family instead of falling through to JS default. Translated code comments now replace English originals instead of being prepended alongside them.
- **Gemini API logging**: timestamped REQUEST/RESPONSE logging with model, duration, and token counts. Per-language timing and formatted token usage summary table at pipeline end. Verbose prompt logging behind VERBOSE flag.

### `sh copy` fence label misidentified as JS (`code-block-extractor.ts`)
`getCommentSyntax("sh copy")` fell through to the default `"js"` family because the full string didn't match `"sh"`. This caused `//` in URLs like `https://github.com/...` inside shell code blocks to be treated as line comments and corrupted during translation.

**Fix:** Strip metadata after the language name before matching: `language.split(/\s+/)[0]`

### Translated comments appended instead of replacing English (`gemini-translate.ts`)
`restoreComments()` was called with the original code block content (English comments intact) instead of the stripped version (comments removed). This produced duplicate comment blocks -- translated Indonesian prepended above the original English.

**Fix:** Pass `strippedCode` (from `extractComments()`) to `restoreComments()` instead of `block.content`.

## Test plan

- [ ] Verify `getCommentSyntax("sh copy")` returns `"shell"` (new test)
- [ ] Verify multi-line comment restore into stripped code has no English duplication (new test)
- [ ] Verify extract-translate-restore round trip produces clean NatSpec (new test)
- [ ] Run full test suite: `npx playwright test --project=unit tests/unit/sanitizer/code-block-extractor.spec.ts`
- [ ] Run translation on a test file with `sh copy` fence to confirm no URL corruption


wackerow:
Closing in lieu of upcoming `intl-pipeline-v5` branch changes

PR #17850, opened 2026-03-25
Refactor all fetchers to use fetchRetry

pettinarip:
## Summary
- Replace plain `fetch()` with `fetchRetry()` across all 26 data-layer fetchers for automatic retry on transient HTTP errors (408, 409, 429, 500-599)
- Remove custom `retry()` wrapper usage in `fetchGitHub` and `fetchNpmJs`, consolidating retry logic into the shared `fetchRetry` utility
- Delete the now-unused `src/lib/utils/fetch.ts` (old retry/sleep utilities)
- Add per-caller override support to `fetchRetry` for custom retry strategies (e.g. `"headers"` strategy for APIs with `Retry-After`)

## Details
- `fetchRetry` wraps Trigger.dev's `retry.fetch` with exponential backoff (3 attempts) on 408, 409, 429, and 500-599
- GitHub-specific 403 rate-limit handling preserved in `fetchGitHubContributors` and `fetchGitHistory` (GitHub uses 403, not 429)
- Inter-batch pacing delays preserved where they existed (GitHub GraphQL 1s, npm 15s)
- `sleep` utility exported from `fetchRetry.ts` for shared use

## Test plan
- [x] TypeScript compiles cleanly (`npx tsc --noEmit`)
- [x] ESLint + Prettier pass (via pre-commit hooks)
- [x] Verify fetchers succeed in Trigger.dev dev environment
- [ ] Monitor first production run for any regression in data fetching


src/data-layer/fetchers/fetchRetry.ts
```diff
@@ -2,31 +2,50 @@
...
-    maxAttempts: 2,
-  },
+type RetryFetchInit = NonNullable<Parameters<typeof retry.fetch>[1]>
+type RetryByStatus = NonNullable<
+  NonNullable<RetryFetchInit["retry"]>["byStatus"]
+>
+
+/** Sleep for specified milliseconds. */
+export const sleep = (ms: number) =>
+  new Promise((resolve) => setTimeout(resolve, ms))
```
wackerow, line 22: Could consider leaving this back in the fetch util, unless it needs to be inside data-layer dir.. not a blocker

PR #17860, opened 2026-03-25
Developer tooling page suggestions

escottalexander:
@wackerow @konopkja and I had talked about some changes to the developer tooling page and Paul asked me to submit a draft PR with a first iteration of my ideas for improvements.

## Description
Open to feedback if something doesn't look/feel right.

Main improvements:
- Switched to the new [ethereum/builder-resources](https://github.com/ethereum/builder-resources) repo as the data source for tools
- Restructured categories and added subcategories
- UI is updated to expose all the tools on a single page but with search and easy category drill-down
- Data is now sorted based on a combination of the [deepfunding ](https://github.com/deepfunding/dependency-graph/blob/main/datasets/gg24-phase2/seedReposWithDependenciesAndWeights.json) research results and GitHub/NPM data. This way top results are determined by an algorithm( see src/data-layer/fetchers/developer-tools/ranking.ts)  that shows the most popular resources first.
- Added "Suggest a resource" to the bottom of the page with a link to the builder-resources repo issue creation form.

Thank you for checking it out.


wackerow:
Thanks @escottalexander! Will try to find some time to sync on this next week

src/intl/en/page-developers-tools.json
```diff
@@ -1,52 +1,51 @@
 {
```
konopkja, line 1: Suggestion: The "Resources" label in the breadcrumb path (`pathRootLabel`) feels redundant since the page is already clearly about developer tools/resources. Consider removing it or replacing with something more contextual. The word "Resources" doesn't add information the user doesn't already have from being on this page.

app/[locale]/developers/tools/_components/ToolsCatalog.tsx
```diff
@@ -0,0 +1,530 @@
...
+          } else {
+            toolsBySubcategory.set(tool.subcategory_id, [tool])
+          }
+        }
+
+        const subcategoryGroups = category.subcategories
+          .map((subcategory) => {
+            const subcategoryTools =
+              toolsBySubcategory.get(subcategory.id)?.slice() || []
+            subcategoryTools.sort((a, b) => {
```
Comment (konopkja) [inline app/[locale]/developers/tools/_components/ToolsCatalog.tsx:230]: Suggestion: These category headings (e.g. "Contract tooling", "Security & testing") are the primary content sections of the page. They'd benefit from being `<h2>` instead of `<h3>` for better visual hierarchy and semantic structure. The subcategory headings below them (`<h4>`) could then become `<h3>`.

```suggestion
                    <h2 className="text-xl font-bold">
```

konopkja (commented):
A few UI suggestions for the new tooling page layout:

app/[locale]/developers/tools/_components/ToolsCatalog.tsx
```diff
@@ -0,0 +1,530 @@
...
+      {hasBeenVisible ? (
+        <AppCard
+          key={toolKey}
+          name={tool.name}
+          nameClassName="text-base"
+          description={plainDescription}
+          descriptionClassName="text-sm"
+          descriptionMaxLines={2}
+          descriptionExpandable={false}
+          thumbnail={tool.thumbnail_url ?? undefined}
```
Comment (konopkja) [inline app/[locale]/developers/tools/_components/ToolsCatalog.tsx:105]: Suggestion: The search `<Input>` should be full-width of the sidebar. It looks like it may be constrained by the sticky container. Adding `className="w-full"` on the Input (if not already default) or ensuring no padding/margin is limiting it would make this feel more polished.

app/[locale]/developers/tools/_components/ToolsCatalog.tsx
```diff
@@ -0,0 +1,530 @@
...
+              if (scoreDiff !== 0) return scoreDiff
+              return a.name.localeCompare(b.name)
+            })
+            return {
+              subcategory,
+              tools: subcategoryTools,
+            }
+          })
+          .filter((group) => group.tools.length > 0)
+
```
Comment (konopkja) [inline app/[locale]/developers/tools/_components/ToolsCatalog.tsx:241]: Following from the h3 -> h2 suggestion above for category headings: these subcategory headings should then become `<h3>` instead of `<h4>` to maintain proper heading hierarchy.

```suggestion
                            <h3 className="text-sm font-semibold text-body-medium">
```

app/[locale]/developers/tools/_components/ToolsCatalog.tsx
```diff
@@ -0,0 +1,530 @@
...
+        <AppCard
+          key={toolKey}
+          name={tool.name}
+          nameClassName="text-base"
+          description={plainDescription}
+          descriptionClassName="text-sm"
+          descriptionMaxLines={2}
+          descriptionExpandable={false}
+          thumbnail={tool.thumbnail_url ?? undefined}
+          href={`?tool=${encodeURIComponent(toolKey)}`}
```
Comment (konopkja) [inline app/[locale]/developers/tools/_components/ToolsCatalog.tsx:109]: Suggestion: The short description text on each tool card in the main grid could use a secondary/muted color to create better visual hierarchy between the tool name and its description. Something like `text-body-medium` instead of the default `text-body`.

```suggestion
          descriptionClassName="text-sm text-body-medium"
```

(This is just for the catalog cards on the main page, not the modal popup.)

konopkja:
Screenshot 2026-04-08 at 19 53 13

this white space on top in the pop up should be removed

konopkja:
Screenshot 2026-04-08 at 19 54 38
 the number counter needs to be left aligned with the content in brackets

konopkja:
Screenshot 2026-04-08 at 20 06 47

should probably be grey not so strong

escottalexander:
> Screenshot 2026-04-08 at 19 54 38 the number counter needs to be left aligned with the content in brackets

@konopkja  Do you mean like this?
image

escottalexander:
> Screenshot 2026-04-08 at 20 06 47
> should probably be grey not so strong

Do you mean when hovering the grey should be lighter? Or are you saying the description text should be grey?

konopkja:
> > Screenshot 2026-04-08 at 19 54 38 the number counter needs to be left aligned with the content in brackets
> 
> @konopkja Do you mean like this? image

yep exactly, just the numbers should be centered

konopkja:
> > Screenshot 2026-04-08 at 20 06 47
> > should probably be grey not so strong
> 
> Do you mean when hovering the grey should be lighter? Or are you saying the description text should be grey?

ah, i meant the purple icon should be by default probably grey

src/intl/en/page-developers-tools.json
escottalexander: I ended up just removing it.

escottalexander:
> A few UI suggestions for the new tooling page layout:

Thank you @konopkja. All your feedback has been resolved in my latest commit.

pettinarip (commented):
Hey @escottalexander! liked the new direction of the page.

I did a quick perf/seo pass against prod and wanted to flag a few SEO things worth a look. @mnelsonBT @konopkja would be nice to have your eyes on this too

1. we lose the static category URLs. The 7 SSG pages (`/developers/tools/security`, `/sdks`, etc.) now collapse into `?category=` query params on a single page. Google generally doesn't index query-param variants as distinct pages, so we'd be giving up 7 indexable URLs that currently rank. The legacy slugs are redirected (good 👍), but the new category URLs the page itself links to only resolve via a 307 temporary redirect. Worth making those permanent if we keep them.

2. the catalog isn't crawlable. Because tools render client-side via `LazyToolCard` (IntersectionObserver), the server HTML ships 0 rendered tool cards. The full dataset also roughly *doubles* the HTML payload (+95% gzip) while showing less actual content.

3. JSONLDs, mostly good with one gap. The main page's structured data is correctly updated to the new taxonomy, nice. Two things: (a) `[category]/page-jsonld.tsx` is now dead code since those routes just redirect, so we've effectively dropped the per-category structured data; and (b) the `ItemList` lists categories but its URLs point at the redirecting paths rather than the real `?category=` destinations.

Just want to make sure we don't quietly drop SEO surface.

escottalexander:
Thanks for the thorough SEO/perf pass @pettinarip. Really appreciate you digging into this, these were great
  catches. I've reworked the page to address all three but I am super open to any more feedback you may have.

  1. Static category URLs are back as real pages

  The category routes are no longer `?category=` query params on a single page: `/developers/tools/[category]` is once again a proper SSG route with one statically-generated page per category and daily revalidate in case the tool source repo gets updates. The catalog sidebar now links directly to those real `/developers/tools/{category}/` paths rather than query strings, so we keep all of the indexable URLs that currently rank. Legacy slugs (e.g. `/developers/tools/security` > `/developers/tools/security-testing/`) are permanently redirected in `redirects.config.js`, so there are no temporary 307s anymore.

  2. The catalog is now crawlable

  I dropped the client-side LazyToolCard/IntersectionObserver approach; all tool cards now render in the server HTML, so crawlers get the full dataset on first load. To keep the off-screen rendering cost down without pulling cards out of the DOM, the cards use CSS content-visibility: auto, which lets the browser skip layout/paint for off-screen cards while leaving them fully present in the markup. **If you can see a better approach that helps reduce the gzip size while also making it crawlable, please advise.**

  3. JSONLD

  - `[category]/page-jsonld.tsx` is no longer dead code since the category routes are real pages again, the per-category structured data renders on each of them as intended.
  - The main page's ItemList URLs now point at the real `/developers/tools/{category}/` destinations rather than the redirecting paths.

  I've rebased onto the latest dev as well. Would love another look when you
  have a moment and @mnelsonBT @konopkja, happy to hear any further thoughts.
  
  🙏🙏🙏

mnelsonBT:
> Thanks for the thorough SEO/perf pass @pettinarip. Really appreciate you digging into this, these were great catches. I've reworked the page to address all three but I am super open to any more feedback you may have.

Hi @escottalexander, apologies for neglecting to respond on GH here. Pablo was looking at a fix to use the  /apps/categories/ and AppCard pattern before going ooo; I'll follow up with @wackerow today to see if your update here will work or if he wants to refactor to that approach. Thank you!

pettinarip:
Closing this in favor of **#18537**, which continues and finishes this work.

Thanks @escottalexander — your three original commits are preserved verbatim in #18537 (same commit authorship, not squashed), so your contribution is fully credited there. #18537 layers URL-structure, intercepting-route modal, catalog perf, and design-system fixes on top, targets `dev`, is approved, and is being merged.

Nothing is lost by closing here — the merged result on `dev` will include your commits.

PR #17875, opened 2026-03-27
Fix React DOM warnings for blurWidth/blurHeight image props

pettinarip:
## Summary
- Strip `blurWidth`/`blurHeight` from `getImageProps()` rest props before spreading onto `` elements — these are internal Next.js blur placeholder dimensions, not valid HTML attributes ([vercel/next.js#56511](https://github.com/vercel/next.js/issues/56511))
- Add `images.qualities` config to `next.config.js` for Next.js 16 compatibility

## Test plan
- [x] Verified warnings no longer appear in browser console on `/en`
- [x] Verified no `blurwidth`/`blurheight` attributes on DOM `` elements
- [x] ESLint passes on all changed files

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -98,7 +98,7 @@ module.exports = (phase) => {
     },
     trailingSlash: true,
     images: {
-      qualities: [5, 10, 20, 35, 40, 75],
+      qualities: [5, 10, 20, 35, 40, 75, 90, 100],
       deviceSizes: [640, 750, 828, 1080, 1200, 1504, 1920],
       remotePatterns: [
         {


next.config.js
```diff
@@ -98,6 +98,7 @@ module.exports = (phase) => {
     },
     trailingSlash: true,
     images: {
+      qualities: [5, 10, 20, 35, 40, 75],
```
wackerow: @pettinarip 

```suggestion
      qualities: [5, 10, 20, 35, 40, 75, 90, 100],
```

In the repo:
- quality={100} in src/components/ui/avatar.tsx (lines 175, 198)
- quality={90} in src/components/Homepage/TrustLogos.tsx (line 64)
- quality={90} in src/components/Homepage/SavingsCarousel.tsx (line 288)

Would suggest being safe and adding 90 and 100 to this set. Details below 👇

---

https://nextjs.org/docs/15/app/api-reference/components/image#qualities

> ```
> module.exports = {
>   images: {
>     qualities: [25, 50, 75],
>   },
> }
> ```
> In the example above, only three qualities are allowed: 25, 50, and 75. If the [quality](https://nextjs.org/docs/15/app/api-reference/components/image#quality) prop does not match a value in this array, the image will fail with a `400` Bad Request

Saw this in the docs, but oddly I'm not getting that result... https://deploy-preview-17875.ethereum.it/bug-bounty uses Avatar but seems to [load them just fine](https://deploy-preview-17875.ethereum.it/_next/image/?url=https%3A%2F%2Fgithub.com%2Fholiman.png%3Fsize%3D40&w=128&q=100), and the URLs in the srcset load without any issue even with `q=100`. That being said, would be a good safeguard in case that behavior changes, OR, we consider setting those cases to `quality={75}`

pettinarip: makes sense,thanks!

PR #17874, opened 2026-03-27
Deploy v11.1.0

pettinarip:
***
### ⚡️ Changes
- fix: resolve structured data validation errors across app, tutorial, and event pages (#17863) @pettinarip
- Update decentralized-identity/index.md (#17014) @zulu0echo
- Add internal infrastructure section to AGENTS.md (#17849) @pettinarip
- Add new bounty hunter 'Grego AI' - Low 1000 points (#17844) @0xMushow
- fix(data-layer): beaconchain 429 rate limit resilience (#17832) @pettinarip
- feat: hide contributors if edit button hidden (#17797) @wackerow

## 🌐 Translations

- deprecate(i18n): translation progress from picker (#17852) @wackerow
- i18n: Gemini translations (ta) (#17871) @wackerow
- i18n: Gemini translations (bn) (#17866) @wackerow
- i18n: Gemini translations decentralized-identity (all languages) (#17867) @wackerow
- i18n: Gemini translations roadmap/zkevm (all languages) (#17861) @wackerow
- Add zkEVM for L1 block verification page (#16975) @minimalsm
- i18n: select re-translations (sw) (#17857) @wackerow
- fix: replace broken YouTube embed on /desci/ page (#17851) @pettinarip
- i18n: Gemini translations (ur) (#17854) @wackerow
- i18n: re-translations for select Urdu files (ur) (#17859) @wackerow
- i18n(sw): automated Crowdin translation import (part 01/05/07/10/11/12 of 13) (#17521) @minimalsm
- i18n(sw): automated Crowdin translation import (part 12 of 13) (#17532) @minimalsm
- i18n(sw): automated Crowdin translation import (part 11 of 13) (#17531) @minimalsm
- i18n(sw): automated Crowdin translation import (part 10 of 13) (#17530) @minimalsm
- i18n(sw): automated Crowdin translation import (part 07 of 13) (#17527) @minimalsm
- i18n(sw): automated Crowdin translation import (part 05 of 13) (#17525) @minimalsm
- i18n(pl): automated Crowdin translation import (part 03/07/09/10 of 13) (#17445) @minimalsm
- i18n: Gemini translations (de) (#17842) @wackerow
- i18n(mr): automated Crowdin translation import (part 04/07/09/10 of 13) (#17511) @minimalsm
- i18n: Gemini translations creating-a-wagmi-ui-for-your-contract (mr) (#17845) @wackerow
- i18n: Gemini translations (mr) (#17846) @wackerow
[...]

PR #17878, opened 2026-03-27
fix: pin @docsearch/css to 3.9.0 to match @docsearch/react v3

pettinarip:
## Summary

- Pins `@docsearch/css` to `3.9.0` to match the version expected by `@docsearch/react@3.9.0`
- PR #17818 introduced `@docsearch/css@^4.6.0` which resolved to v4, but the React component is v3 — the v4 CSS caused style regressions

## What was broken

1. **Search icon in primary/purple color** — v4 CSS doesn't override the button's inherited `text-primary`, while v3 sets body color on `.DocSearch-Search-Icon`
2. **Constant spinner in search modal** — v4 CSS shows `.DocSearch-LoadingIndicator` by default (`display: flex`), while v3 hides it unless `.DocSearch-Container--Stalled` is present
3. **Hover/highlight style issues** — v4 CSS variables and selectors don't align with the v3 React component's DOM structure

Reported in #17874 (comment): https://github.com/ethereum/ethereum-org-website/pull/17874#issuecomment-4141205224

## Test plan

- [x] Search icon in nav should be body color (dark), not purple
- [x] Opening search modal should show no spinner when idle
- [x] Search results should have correct hover and highlight colors

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   "dependencies": {
     "@aws-sdk/client-s3": "^3.982.0",
     "@aws-sdk/client-ses": "^3.859.0",
-    "@docsearch/css": "^4.6.0",
+    "@docsearch/css": "3.9.0",
     "@docsearch/react": "^3.5.2",
     "@hookform/resolvers": "^3.8.0",
     "@netlify/blobs": "^10.4.1",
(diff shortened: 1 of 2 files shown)


wackerow:
Thanks @pettinarip, yeah would've been from #17818 🙏

PR #17898, opened 2026-03-30
refactor: upgrade Next.js 15.5 to 16.2 with next-intl v4

pettinarip:
## Summary

- Upgrade Next.js from 15.5.14 to 16.2.1
- Upgrade next-intl from v3 to v4.8.3
- Upgrade @sentry/nextjs to 10.46.0, @next/bundle-analyzer to 16.2.1
- Rename `middleware.ts` → `proxy.ts` (Next.js 16 convention)
- Add Turbopack loader rules alongside existing webpack config
- Update lint scripts from `next lint` to `eslint .` (`next lint` removed in v16)

## Key decisions

- **Webpack stays default** for both dev and build — Turbopack has Tailwind CSS selector parsing issues that need a Tailwind upgrade to resolve. Turbopack rules are configured and available via `dev:turbo` / `build:turbo`
- **ESLint 8 kept** — the ESLint 9 + flat config migration is deferred to a follow-up PR to reduce blast radius

## Test plan

- [x] `npx tsc --noEmit` passes (no new errors)
- [x] `pnpm build` (webpack) succeeds
- [x] `pnpm dev` (webpack) starts cleanly
- [x] Lint-staged hooks pass on commit
- [x] Netlify deploy preview builds and serves pages correctly
- [x] Static content pages don't 404 (ISR avoidance preserved)
- [x] Locale switching works across all routes
- [x] ~~`pnpm build-storybook` succeeds~~ **Known issue**: `@storybook/nextjs@8.6` does not support Next.js 16 (`next/config` was removed). Requires Storybook 10.3+ upgrade — tracked in follow-up PR.

## Follow-up PRs

- ESLint 9 + flat config migration
- Turbopack as default bundler (after Tailwind upgrade)
- **Storybook 10 upgrade** (`@storybook/nextjs@8.6` → `10.3+` for Next.js 16 compat)
- `cacheComponents` / PPR adoption (after `crypto.randomUUID()` investigation)


src/i18n/routing.ts
```diff
@@ -8,4 +8,5 @@ export const routing = defineRouting({
   localePrefix: "as-needed",
   alternateLinks: false,
   localeDetection: false,
+  localeCookie: false,
```
pettinarip, line 11: it was part of `localeDetection` before. Now there is a separate option to disable cookies.

package.json
```diff
@@ -4,11 +4,13 @@
...
   "private": true,
   "scripts": {
-    "dev": "next dev",
-    "build": "next build",
+    "dev": "next dev --webpack",
+    "dev:turbo": "next dev",
+    "build": "next build --webpack",
+    "build:turbo": "next build",
     "start": "next start",
-    "lint": "next lint",
```
pettinarip, line 10: `next lint` was removed from next 16

next.config.js
```diff
@@ -240,5 +252,4 @@ module.exports = withSentryConfig(module.exports, {
   silent: true,
   widenClientFileUpload: true,
   disableLogger: true,
-  automaticVercelMonitors: true,
```
pettinarip, line 243: vercel only option, not needed

src/components/Hero/HomeHero/index.tsx
```diff
@@ -23,7 +23,7 @@ const HomeHero = async ({
   image2xl,
   alt: altProp,
 }: HomeHeroProps) => {
-  const locale = getLocale()
+  const locale = await getLocale()
   const t = await getTranslations({ locale, namespace: "page-index" })
```
wackerow: From the pattern I'm seeing, looks like we could just remove this locale assignment entirely?


```suggestion
  const t = await getTranslations("page-index")
```

next.config.js
```diff
@@ -96,6 +96,18 @@ module.exports = (phase) => {
 
       return config
     },
+    // Turbopack loader equivalents for the webpack() config above
```
wackerow, line 99: 🚀

src/components/Hero/HomeHero/index.tsx
wackerow: (If we do -- `getLocale` import needs to be removed too, forgot to code-suggest that)

pettinarip: yeah, wasn't sure if I wanted to refactor all of them because it was going to add a lot of noise but I just did. Pushed one last commit to refactor all and use the short/inferred version.

wackerow: Thanks, yeah I think it's worth it while we're doing this to just clean them all up 👍

next.config.js
```diff
@@ -96,6 +96,18 @@ module.exports = (phase) => {
...
+      rules: {
+        "*.yaml": { loaders: ["yaml-loader"], as: "*.js" },
+        "*.yml": { loaders: ["yaml-loader"], as: "*.js" },
+        "*.svg": { loaders: ["@svgr/webpack"], as: "*.js" },
+        "*.md": { loaders: ["raw-loader"], as: "*.js" },
+        "*.mp3": { as: "*.static" },
+      },
+    },
+    // Replaces config.externals.push("pino-pretty", "lokijs", "encoding")
+    serverExternalPackages: ["pino-pretty", "lokijs", "encoding"],
```
wackerow, line 110: Claude noting:

> Nit: The webpack callback still has `config.externals.push("pino-pretty", "lokijs", "encoding")`. Both `serverExternalPackages` and the webpack externals push do the same thing. The `serverExternalPackages` is the Next.js 16 way, and the webpack push is the legacy way.
> 
> Having both isn't harmful (double-externalization is a no-op), but it's technically redundant when running webpack. Could clean up the webpack push in a follow-up, but not blocking.

package.json
```diff
@@ -107,8 +108,9 @@
     "@chromatic-com/storybook": "1.5.0",
     "@google/genai": "^1.46.0",
     "@netlify/plugin-nextjs": "^5.15.9",
-    "@next/bundle-analyzer": "^15.5.12",
+    "@next/bundle-analyzer": "^16.2.1",
     "@playwright/test": "^1.52.0",
+    "@sentry/nextjs": "^10.46.0",
```
wackerow: Claude noting:

> `@sentry/nextjs` moved from dependencies to devDependencies: This works because Next.js bundles everything at build time, and Netlify installs all deps during build. But it's unconventional and could confuse future maintainers.
> 
> Sentry runs at runtime (initialization in `sentry.server.config.ts` / `sentry.edge.config.ts`). I'd consider keeping it in dependencies for clarity, though functionally it works either way since the bundler resolves it at build time.

next.config.js
pettinarip: Correct. This is part of the Turbopack adoption. Since we are still using webpack, I just prefer to not touch these configs yet. We can deal with this in the other PR

package.json
pettinarip: Agreed. Patched

PR #17906, opened 2026-04-01
refactor: adopt Turbopack as default bundler

pettinarip:
## Summary

Adopts Turbopack as the default bundler. `--webpack` becomes the opt-out (`dev:webpack` / `build:webpack` kept as fallback). Along the way, fixes several patterns that didn't survive Turbopack's stricter module graph and file tracer.

## What's in this PR

**Default bundler flip**
- `dev` / `build` scripts drop `--webpack`.
- `@next/bundle-analyzer` removed (webpack-only).
- `Contributors` drops `!!raw-loader!` inline syntax.
- Sentry: dropped webpack-only `treeshake.removeDebugLogging`; silenced dev-time debug logging.

**Turbopack-specific fixes**
- **Tutorial MDX loader**: replaced `await import('.../${locale}/${slug}/index.md')` with `fs.readFile()`. Turbopack statically registers every template match (~143k candidate modules), which froze `pnpm dev`.
- **Path constants in the MDX pipeline**: Turbopack's file tracer does not fold cross-module string constants, so `join(CONTENT_DIR, ...slug)` traces ~4× wider than `join("public/content", ...slug)`. Inlined at hot sites; residual warnings silenced via `turbopack.ignoreIssue`. See `docs/solutions/build-errors/turbopack-file-tracer-doesnt-fold-cross-module-string-constants.md` so this isn't reverted later.

**Contributors + Staking refactor**
- `Contributors` is now a server-only component that reads `.all-contributorsrc` directly, strips unused fields, and shuffles once per SSG build. Deletes the `all-contributors.json` duplicate (it was already drifting from the bot-maintained source).
- `/staking` converted from a client component to a server component (`getTranslations` instead of `useTranslation`). `StakingHierarchy` explicitly marked `"use client"`.

## Behavior changes

- Contributors list renders in SSR HTML (no empty-then-filled hydration flicker). Shuffle order changes per deploy rather than per visit.
- `.all-contributorsrc` (bot-maintained) is the single source of truth. No possibility of drift.

## Test plan

- [x] `pnpm dev` starts with Turbopack
- [x] `pnpm build` succeeds with Turbopack
- [x] `pnpm build:webpack` still works as fallback
- [x] `pnpm build-storybook` succeeds
- [x] Netlify deploy preview builds
- [x] Contributors, staking pages render
- [x] Sentry source-map upload completes

## Known gaps / follow-ups (not in this PR)

- **`public/content/**` mixes static assets and source markdown**, which forces the file tracer to consider ~200k files in scope. Moving `content/` out of `public/` (or migrating to a proper content layer like Contentlayer/Velite) would eliminate most of the Turbopack band-aids in this PR. Bigger effort; worth scoping separately.
- **`turbopack.ignoreIssue` config.** Cosmetic suppression of "Overly broad patterns" warnings that `outputFileTracingExcludes` already prevents from over-bundling. Would go away with the content-pipeline migration above.


myelinated-wackerow (approved):
LGTM. Few nice-to-haves:

- `TRANSLATIONS_DIR` in `src/lib/constants.ts` looks orphaned now that `translationRegistry.ts` inlines the literal — please double-check and remove if confirmed unused
- Nit on `ContributorsView`: `className="h-[132px] w-[132px]"` could just be `size-[132px]`, and the `width={132} height={132}` attrs were only required for `next/image` — the plain `` doesn't need them unless I'm missing a CLS reason

pettinarip:
@myelinated-wackerow thanks for the review!

> `TRANSLATIONS_DIR` in `src/lib/constants.ts` looks orphaned now that `translationRegistry.ts` inlines the literal — please double-check and remove if confirmed unused

Confirmed orphaned (no imports anywhere in `src/` or `app/`; the only other `TRANSLATIONS_DIR` matches were local constants defined inside `.markdownlint-rules/*.js`, which never imported from `@/lib/constants`). Removed the export.

> Nit on `ContributorsView`: `className="h-[132px] w-[132px]"` could just be `size-[132px]`, and the `width={132} height={132}` attrs were only required for `next/image` — the plain `` doesn't need them unless I'm missing a CLS reason

You weren't missing anything — it's a plain ``. Switched to `size-[132px]` and dropped the `width`/`height` holdovers. The class still reserves the box, so CLS is unchanged.

Both addressed in 00ed4adc6e.

PR #17912, opened 2026-04-02
perf: improve INP by deferring Matomo tracking and virtualizing Select

pettinarip:
## Summary

Improves Interaction to Next Paint (INP) across the site by addressing the two highest-impact bottlenecks identified via CrUX data, Sentry performance monitoring, and local web-vitals measurement.

### 1. Defer Matomo tracking to `requestIdleCallback`

Every link click, button press, and tab switch was synchronously calling `trackCustomEvent`, which read `localStorage`, parsed JSON, and pushed to the Matomo queue — all blocking the main thread during user interactions.

Now the tracking call is deferred to `requestIdleCallback` (with `setTimeout(0)` fallback), and the opt-out status is cached in memory to avoid repeated `localStorage` reads.

### 2. Virtualize react-select MenuList for large option lists

The `/get-eth/` country picker renders 240 options via react-select. Opening the dropdown mounted all 240 `Option` components as DOM nodes, causing **>1s INP**.

Added a virtualized `MenuList` using `@tanstack/react-virtual` (already a project dependency) that only renders visible items. Kicks in automatically when the option count exceeds 30.

### Measured INP on `/get-eth/` (production build, web-vitals library)

| Version | INP | Rating |
|---------|-----|--------|
| Before (240 DOM options) | 1048ms | poor |
| After (virtualized) | 248ms | needs-improvement |

Other pages (homepage, /what-is-ethereum, /staking, /what-is-ether) measured between 48–112ms (good).

## Test plan

- [x] Visit `/get-eth/`, open the country picker, search and select a country — verify it works and feels responsive
- [x] Check other pages with react-select (`/staking/` launchpad widget) still work — the virtualization only kicks in above 30 options
- [x] Verify Matomo events still fire in production (check network tab for Matomo requests after clicking links)
- [x] Test the Matomo opt-out toggle on `/privacy-policy/` — toggling should take effect immediately #17922

### DIFF
--- a/src/lib/utils/matomo.ts
+++ b/src/lib/utils/matomo.ts
@@ -41,11 +41,12 @@ export const trackCustomEvent = ({
 
   if (isOptedOut()) return
 
+  // Capture URL synchronously — client-side navigations can change
+  // window.location before the idle callback fires
+  const currentUrl = window.location.href.split(/[?#]/)[0]
+
   scheduleIdleCallback(() => {
-    // Set custom URL removing any query params or hash fragments
-    if (window) {
-      push([`setCustomUrl`, window.location.href.split(/[?#]/)[0]])
-    }
+    push([`setCustomUrl`, currentUrl])
     push([`trackEvent`, eventCategory, eventAction, eventName, eventValue])
   })
 }


myelinated-wackerow (approved):
@pettinarip Nice PR -- clean scope, solid INP improvement, well-measured.

One suggestion: capture `window.location.href` synchronously before deferring to `requestIdleCallback`. In production, fast client-side navigations could update the URL before the idle callback fires, causing `setCustomUrl` to record the destination page instead of the source page:

```ts
const currentUrl = window.location.href.split(/[?#]/)[0]

scheduleIdleCallback(() => {
  push([`setCustomUrl`, currentUrl])
  push([`trackEvent`, eventCategory, eventAction, eventName, eventValue])
})
```

Zero perf cost -- the expensive work stays deferred.

Also minor: PR description says virtualization kicks in above 50 options but `VIRTUAL_THRESHOLD` is 30.

Otherwise LGTM.

---
Reviewed by Claude (Opus 4.6)

pettinarip:
> capture `window.location.href` synchronously before deferring to `requestIdleCallback`

Fixed in 3dcaef9 — URL is now captured synchronously before the idle callback.

> PR description says virtualization kicks in above 50 options but `VIRTUAL_THRESHOLD` is 30

Fixed the PR description.

PR #17921, opened 2026-04-03
refactor: decouple e2e tests from chromatic, run unit tests on PRs

pettinarip:
## Summary

- Decouples E2E tests from Chromatic's Playwright integration — tests now use standard `@playwright/test` imports
- Adds a `unit-tests` CI job that runs on every PR (previously no tests ran on PRs)
- Restricts E2E tests to `workflow_dispatch` only (manual trigger) since they require a Netlify deploy
- Removes the `chromatic` CI job that depended on Playwright artifacts
- Removes `@chromatic-com/playwright` dependency from `package.json`
- Updates `docs/e2e-testing.md` to reflect the new setup

## Why drop the full-page visual snapshots?

The E2E Chromatic snapshots captured full pages (home, find-wallet, Arabic RTL) that render live data — wallet listings, ETH price, events, stats. This data changes between every deploy, so snapshots always produced diffs that were auto-accepted (`exitZeroOnChanges: true`). They never caught real regressions, only data churn.

We also have a limited Chromatic snapshot quota. Running full-page captures across 4 browser/device configs on every dispatch burned budget without value. Component-level visual regression via the Storybook Chromatic workflow (`chromatic.yml`) is a better use of that quota — stories use controlled data and actually catch meaningful UI changes.

The functional E2E assertions (page loads, navigation, filters, search, RTL layout checks) are preserved — only the `takeSnapshot()` calls and the Chromatic pipeline job are removed.

## Test plan

- [x] Verify `unit-tests` job runs on PR open
- [x] Verify `e2e-tests` job does NOT run on PR open
- [x] Verify `e2e-tests` job runs via manual workflow dispatch
- [x] Confirm `pnpm test:unit` passes locally


myelinated-wackerow (commented):
@pettinarip Looks good -- clean separation of unit vs e2e, and the Chromatic E2E removal rationale makes sense.

Two small notes:

1. **Consider adding pnpm store cache to `unit-tests`** -- the `e2e-tests` job already uses `actions/cache` for this. Without it `unit-tests` will re-download all deps on every PR. Not blocking, just slower than it needs to be.

2. The renamed workflow (`tests.yml`) won't appear in the Actions tab until it's merged into `dev` -- just a heads up so nobody wonders where it went.

---
Reviewed by Claude (Opus 4.6)

wackerow (approved):

pettinarip:
> 1. **Consider adding pnpm store cache to `unit-tests`** -- the `e2e-tests` job already uses `actions/cache` for this. Without it `unit-tests` will re-download all deps on every PR.

Good catch — added `cache: pnpm` to `actions/setup-node` in the `unit-tests` job (and reordered so `pnpm/action-setup` runs first, since setup-node needs to call `pnpm store path` for the cache probe). See 5b11af8.

> 2. The renamed workflow (`tests.yml`) won't appear in the Actions tab until it's merged into `dev`

Yep, aware — thanks for the heads up.

PR #17922, opened 2026-04-03
fix: respect matomo opt-out for page view tracking

pettinarip:
## Summary

- The Matomo opt-out toggle on `/privacy-policy/` only gated custom events (`trackCustomEvent`), while page view tracking in `Matomo.tsx` continued regardless of opt-out status
- This is a gap between the privacy policy (section 3.3) which implies opting out stops all Matomo tracking, and the actual code behavior
- Exports the existing `isOptedOut()` helper from `matomo.ts` and checks it before sending page view events in `Matomo.tsx`

Depends on #17912 (branches off `perf/improve-inp-core-web-vitals`).

## Test plan

- [x] Visit `/privacy-policy/`, opt out via the Matomo checkbox
- [x] Navigate to other pages and confirm no `matomo.php` requests appear in the Network tab
- [x] Opt back in and confirm page view tracking resumes on navigation

### DIFF
--- a/src/components/Matomo.tsx
+++ b/src/components/Matomo.tsx
@@ -15,7 +15,7 @@ export default function Matomo() {
   const [previousPath, setPreviousPath] = useState("")
 
   useEffect(() => {
-    if (!matomoInitialized) {
+    if (!matomoInitialized && !isOptedOut()) {
       init({
         url: process.env.NEXT_PUBLIC_MATOMO_URL!,
         siteId: process.env.NEXT_PUBLIC_MATOMO_SITE_ID!,
--- a/src/components/Matomo.tsx
+++ b/src/components/Matomo.tsx
@@ -4,6 +4,8 @@ import { useEffect, useState } from "react"
 import { usePathname } from "next/navigation"
 import { init, push } from "@socialgouv/matomo-next"
 
+import { isOptedOut } from "@/lib/utils/matomo"
+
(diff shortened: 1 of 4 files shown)


myelinated-wackerow (approved):
@pettinarip Looks good -- approving.

The opt-out guard placement is correct: it checks before any `push()` calls fire, still updates `previousPath` so referrer stays accurate if the user opts back in, and cache invalidation via `clearMatomoOptOutCache()` ensures the toggle takes effect immediately on next navigation.

One thing worth considering as a follow-up: `Matomo.tsx` unconditionally calls `init()` on mount, which loads the Matomo tracker script regardless of opt-out status. Even without a `trackPageView`, the script load itself may make a request to the Matomo server. An early `if (isOptedOut()) return` before the `init()` block would close that gap. Not a blocker for this PR.

Reviewed by Claude (Opus 4.6)

pettinarip:
> One thing worth considering as a follow-up: `Matomo.tsx` unconditionally calls `init()` on mount, which loads the Matomo tracker script regardless of opt-out status. Even without a `trackPageView`, the script load itself may make a request to the Matomo server. An early `if (isOptedOut()) return` before the `init()` block would close that gap.

Addressed in 1850940 — added `!isOptedOut()` guard to the `init()` block so the tracker script is never loaded for opted-out users.

PR #17928, opened 2026-04-06
add server-rendered hidden nav for search engine crawlability

pettinarip:
## Summary

- Adds a `CrawlableNav` server component that renders all navigation links as plain `<a>` tags in the initial HTML
- The interactive desktop nav (Radix NavigationMenu) unmounts dropdown content when inactive, and the mobile nav is lazy-loaded on click — neither exposes links to crawlers
- `CrawlableNav` uses the same `buildNavigation()` data source so links stay in sync automatically
- Rendered as a sibling `<nav>` with `sr-only` (visually hidden, present in DOM)
- Zero client-side JS — fully server-rendered

## Test plan

- [x] Verify nav links appear in the page source HTML (View Source or `curl`)
- [x] Confirm the hidden nav is not visually visible on any viewport
- [x] Check that interactive desktop and mobile menus still work as before
- [x] Validate no duplicate `key` warnings in console

### DIFF
--- a/src/components/Nav/CrawlableNav.tsx
+++ b/src/components/Nav/CrawlableNav.tsx
@@ -1,0 +1,52 @@
+import { getTranslations } from "next-intl/server"
+
+import type { NavItem } from "./types"
+
+import { Link } from "@/i18n/navigation"
+import { buildNavigation } from "@/lib/nav/buildNavigation"
+
+function collectLinks(items: NavItem[]): { label: string; href: string }[] {
+  const links: { label: string; href: string }[] = []
+  for (const item of items) {
+    if ("href" in item && item.href) {
+      links.push({ label: item.label, href: item.href })
+    }
+    if ("items" in item && item.items) {
+      links.push(...collectLinks(item.items))
+    }
+  }
+  return links
+}
+
+/**
+ * Server-rendered hidden nav for search engine crawlability.
+ *
+ * The interactive desktop nav uses Radix NavigationMenu which unmounts
+ * dropdown content when inactive, and the mobile nav is lazy-loaded on
+ * interaction. Neither makes nav links available in the initial HTML.
+ *
+ * This component renders all nav links as plain <a> tags in the server
+ * response so crawlers can discover and follow them.
+ */
+const CrawlableNav = async () => {
+  const t = await getTranslations("common")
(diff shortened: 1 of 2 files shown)


pettinarip:
@claude review

src/components/Nav/CrawlableNav.tsx
```diff
@@ -0,0 +1,52 @@
...
+const CrawlableNav = async () => {
+  const t = await getTranslations("common")
+  const sections = buildNavigation(t)
+
+  const allLinks = Object.values(sections).flatMap((section) =>
+    collectLinks(section.items)
+  )
+
+  return (
+    <nav aria-hidden="true" className="sr-only" tabIndex={-1}>
```
myelinated-wackerow:
```suggestion
    <nav inert className="sr-only">
```

myelinated-wackerow (changes requested):
@pettinarip `aria-hidden="true"` + `tabIndex={-1}` on the `<nav>` doesn't prevent keyboard focus on the child `<Link>` elements -- `aria-hidden` only affects the AT tree, and `tabIndex` only applies to the element itself, not children. Keyboard users will tab through ~50 invisible links.

Replace with [`inert`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert):

Single attribute that removes from AT, prevents all child focus, and prevents interaction. Supported in all modern browsers.

---
*Reviewed by Claude/Opus*

src/components/Nav/CrawlableNav.tsx
pettinarip: > `aria-hidden="true"` + `tabIndex={-1}` on the `<nav>` doesn't prevent keyboard focus on the child `<Link>` elements -- `aria-hidden` only affects the AT tree, and `tabIndex` only applies to the element itself, not children. Keyboard users will tab through ~50 invisible links.

Addressed: replaced `aria-hidden="true"` and `tabIndex={-1}` with `inert`. Single attribute handles AT removal, focus prevention on all descendants, and interaction blocking.

pettinarip: Great catch btw, didn't know about `inert`

wackerow (approved):

PR #17930, opened 2026-04-06
Fix nested button hydration error in Tooltip components

pettinarip:
## Summary
- Added `asChild` prop to the custom `Tooltip` component, forwarded to Radix's `TooltipTrigger`
- Applied `asChild` to `PlayerButton` and standalone `Tooltip` usages in `PlayerWidget` and `Stat` that wrap `<button>` children, so the child button *becomes* the trigger instead of being nested inside one
- Fixes `<button> cannot be a descendant of <button>` hydration error on `/what-is-ethereum` and anywhere else these components are used

## Test plan
- [ ] Verify no hydration errors in console on `/what-is-ethereum`
- [ ] Verify player widget tooltips still appear on hover (desktop) and tap (mobile)
- [ ] Verify Stat tooltip icon buttons still work
- [ ] Check no visual regression in Chromatic

### DIFF
--- a/src/components/ListenToPlayer/PlayerWidget/index.tsx
+++ b/src/components/ListenToPlayer/PlayerWidget/index.tsx
@@ -26,7 +26,9 @@ const PlayerButton = ({
   return isMobile() ? (
     children
   ) : (
-    <Tooltip content={tooltipContent}>{children}</Tooltip>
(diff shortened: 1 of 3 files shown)


wackerow (approved):
Lgtm! We could even consider defaulting asChild to true on `Tooltip`, but good as it is

PR #17929, opened 2026-04-06
feat: auto-deploy data layer tasks on merge to master

pettinarip:
## Summary

- Adds a GitHub Actions workflow (`deploy-data-layer.yml`) that automatically deploys Trigger.dev data-layer tasks to production when code is merged to `master`
- Pins `trigger.dev` CLI as a devDependency (`4.3.3`) instead of using `npx trigger.dev@latest`
- Updates `trigger:dev` and `trigger:deploy` scripts to use the local binary

## Details

The data-layer has 27 scheduled fetchers (hourly/daily/weekly) deployed to Trigger.dev Cloud. Previously, deploying required running `pnpm trigger:deploy` manually — risking production tasks drifting from the codebase.

The workflow:
- Triggers on push to `master` with path filters: `src/data-layer/**`, `trigger.config.ts`, `package.json`
- Uses `pnpm trigger:deploy` with `TRIGGER_SECRET_KEY` from GitHub Actions secrets
- Runs independently of the Netlify build

## Test plan

- [ ] Verify workflow appears in Actions tab after merge
- [ ] Verify pushes to master with data-layer changes trigger the workflow
- [ ] Verify pushes with only content changes do NOT trigger the workflow
- [x] Verify `pnpm trigger:dev` works locally with the new script

PR #17870, opened 2026-03-27
refactor: videos page revamp

wackerow:
> **Note:** This is a draft preview PR with intent to extend PR:
- #17602

## Summary

Refactors the video data architecture and page structure to align with site conventions.

### Data layer
- Moved all video metadata from `videos.json` into markdown frontmatter (`index.md`)
- Renamed `transcript.md` to `index.md` for each video
- Rewrote data layer to read from filesystem instead of JSON
- Added `VideoFrontmatter` type, replaced old `Video`/`VideoMeta`/`VideoWithMeta` types
- Deleted `videos.json` and `videoTranscripts.ts` (merged into `videos.ts`)
- Added `breadcrumb` frontmatter to all 42 videos
- Used shared `stripMarkdown` instead of duplicate `stripMdx`

### Components and organization
- Moved page-scoped components out of `src/components/` into `app/[locale]/videos/_components/`
- Inlined transcript accordion and markdown rendering directly into `page.tsx`
- Moved `VIDEO_CATEGORIES` to colocated `constants.ts`, `getVideosByCategory` to `utils.ts`
- Updated `VideoWatch` shared component for new data layer
- Cleaned up unused i18n keys
- Removed PR-added unit and e2e tests

### SEO preserved
- JSON-LD `VideoObject` schema on every video detail page
- Transcript text stays in DOM when accordion is closed via `forceMount`
- `generateStaticParams` pre-renders all video pages at build time
- Per-video `generateMetadata` for title and description meta tags
- Breadcrumbs with translatable labels
- Sitemap inclusion for video pages
- Semantic HTML structure

## Preview links
- https://deploy-preview-17870.ethereum.it/nft
- https://deploy-preview-17870.ethereum.it/videos
- https://deploy-preview-17870.ethereum.it/videos/privacy-is-existential/

**Schema Markup Validator**
- https://validator.schema.org/#url=https%3A%2F%2Fdeploy-preview-17870.ethereum.it%2Fvideos
- https://validator.schema.org/#url=https%3A%2F%2Fdeploy-preview-17870.ethereum.it%2Fvideos%2Fprivacy-is-existential%2F

PR #17968, opened 2026-04-13
build(deps): remove trigger.dev CLI from devDependencies

myelinated-wackerow:
## Summary

- Removes `trigger.dev` CLI (4.3.3) from `devDependencies` to reduce Netlify build disk usage
- Reverts `trigger:dev` and `trigger:deploy` scripts to use `npx trigger.dev@4.3.3` (pinned)
- Updates `deploy-data-layer.yml` GitHub Action to use `npx` instead of `pnpm exec`

## Context

Netlify production builds started failing with `ENOSPC: no space left on device` after the videos PR (#17870) merged into staging. The combination of ~1,050 new statically generated video pages and the `trigger.dev` CLI package (added in #17929) exceeded the build environment's disk limit.

The CLI is only used for deploying tasks to Trigger.dev Cloud (via GH Action) and local dev -- it's never imported or bundled. Removing it from `devDependencies` drops ~1,277 lines from the lockfile and avoids it being copied into the Netlify serverless function handler.

No functional change -- the CLI is still available via `npx` at the same pinned version.

## Test plan

- [ ] Netlify deploy preview builds successfully
- [ ] `pnpm trigger:dev` works locally
- [ ] GH Action deploys tasks on merge to master

### DIFF
--- a/.github/workflows/deploy-data-layer.yml
+++ b/.github/workflows/deploy-data-layer.yml
@@ -30,6 +30,6 @@ jobs:
         run: pnpm install
 
       - name: Deploy to Trigger.dev
-        run: pnpm exec trigger deploy
+        run: npx trigger.dev@4.3.3 deploy
         env:
           TRIGGER_SECRET_KEY: ${{ secrets.TRIGGER_SECRET_KEY }}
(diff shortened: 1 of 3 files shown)


wackerow:
Closing for now -- could re-assess it's value if we see the need to trim further

PR #17988, opened 2026-04-17
seo(jsonld): initialize author profiles

myelinated-wackerow:
## Description

- Adds a registry of schema.org Person profiles and resolves them into page JSON-LD via frontmatter (`authors`) or direct imports
- Works with legacy `author` fields through an auto-generated alias map (full name or GitHub handle)
- Falls back to the community organization when no author is declared
- Author attribution strengthens E-E-A-T and author-credibility signals for search engines and LLM citation
- Registry will expand iteratively over time

## Testing plan
- [ ] Spot check updated pages against https://validator.schema.org/ to confirm presence of linked data


src/lib/jsonld/utils.ts
```diff
@@ -0,0 +1,95 @@
...
+  const entities = values
+    .map((v): KnownEntity | null => ENTITY_ALIASES[v.toLowerCase()] ?? null)
+    .filter((e): e is KnownEntity => e !== null)
+
+  return {
+    authorGraphNodes: entities,
+    authorIds:
+      entities.length > 0
+        ? entities.map((e) => ({ "@id": e["@id"] }))
+        : [REFERENCE.ETHEREUM_COMMUNITY],
```
pettinarip: Silent fallback on unrecognized authors. Consider a dev-only console.warn (or Sentry breadcrumb) when a frontmatter value produces zero matches — otherwise authors will think they're attributed when they aren't

app/[locale]/[...slug]/page-jsonld.tsx
```diff
@@ -49,18 +50,23 @@ export default async function SlugJsonLD({
...
+      ...authorGraphNodes,
       {
         "@type": "WebPage",
         "@id": url,
         name: frontmatter.title,
         description: frontmatter.description,
         url: url,
         inLanguage: locale,
-        author: [REFERENCE.ETHEREUM_COMMUNITY],
+        author: authorIds,
```
Comment (pettinarip) [inline app/[locale]/[...slug]/page-jsonld.tsx:70]: should we keep the `ETHEREUM_COMMUNITY` ref always set as a default in md pages?

I'm asking because (if I'm not mistaken) setting `authors: ["Andreas Antonopoulos"]` in the glossary (`public/content/glossary/index.md`) incorrectly attributes the entire glossary to a single person.

Comment (wackerow) [reply app/[locale]/[...slug]/page-jsonld.tsx:70]: @pettinarip yeah good call.. Okay, @mnelsonBT please feel free to chime in, but in the meantime:

- If we force "community" as a baseline contributor on all pages, we loose granularity
- We can add "Ethereum Community" to the front matter authors list to include as needed when other authors are listed
- If NO "authors" front matter available, falls back to community
- I've updated the /glossary/ page to add "Ethereum Community" explicitly

So as examples, as of c1c16b2 the /whitepaper/ page will show Vitalik _without_ the community as "author", but the /glossary/ page will show Andreas _with_ the community.

Comment (wackerow) [reply app/[locale]/[...slug]/page-jsonld.tsx:70]: Okay, after chatting with @mnelsonBT -- going to revert c1c16b2 and KEEP the community as a baseline author. This is more accurate to the default case and there are very few exceptions where that may not be fully accurate (a couple pages max?)... Requiring the addition of "Ethereum Community" on each markdown page that include the "authors" front matter prop would add unnecessary burden, and risk that we do not include them, when the vast majority of cases it is the desired path. The risk of including on the couple pages that it may be less appropriate is minimal. Patch incoming.

src/lib/jsonld/utils.ts
wackerow: Okay thanks!... As noted below, refactored this to always include the community as an author. 

For the "unrecognized" author logging, I'd like to think on that a little... this could create a lot of noise, since currently there are many existing author entries that do no have entries in KNOWN_AUTHORS just yet. Will break that off as a separate task from this PR.

wackerow:
cc: @pettinarip and @mnelsonBT per conversations, going to bring this in for this release -- please let me know if any issues and I can iterate to fix separately.

wackerow (approved):

PR #17997, opened 2026-04-20
fix(intl-pipeline): pending-branch baseline + docs refresh

myelinated-wackerow:
## Summary
- Use the `intl/pending-<base>` branch as baseline on re-runs so non-deterministic LLM output doesn't re-translate unchanged sections or produce merge conflicts. Merges base into pending first and fails fast on conflict; temp branches now branch off pending and are cleaned up after successful merge.
- Document the orchestration layer in `tests/specs/PIPELINE-SPEC.md`: pending-branch lifecycle, temp-branch lifecycle, base-moved-during-run handling, and the non-English file edit policy (manifest-mapping rule, with the `stamp_only` workflow input as the escape hatch).
- Consolidate the Internationalization section of `AGENTS.md` to capture the single-propagator rule and point agents at the spec.
- Migrate the `/review-translations` slash command from the deprecated `ethereum.org/api/glossary` endpoint to ETHGlossary. Base URL resolves at runtime from `GLOSSARY_API_URL` env var or the `config.ts` default so it stays correct when the host changes. Prefers `POST /filter` for per-file scoped lookups.
- Prune `FUTURE.md`: remove items that have shipped; relabel split-PRs as nice-to-have with a note on the pending-branch orchestration nuance.

## Test plan
- [ ] Trigger `intl-pipeline.yml` against a test base with no existing pending branch (first-run path).
- [ ] Trigger again without any English changes (idempotency: zero drift, zero LLM calls, clean no-op merge).
- [ ] Trigger after a new English change lands on base between runs (pending-as-baseline path: base merged into pending, only the delta translated, clean merge back).
- [ ] Simulate a base-into-pending conflict and confirm the run aborts fast with no translation work done.
- [ ] Run `/review-translations` against a PR and confirm ETHGlossary endpoints resolve via `GLOSSARY_API_URL` and `llms.txt` fetch succeeds.

### DIFF
--- a/.claude/commands/review-translations.md
+++ b/.claude/commands/review-translations.md
@@ -224,38 +224,39 @@ Read `.claude/translation-review/known-patterns.md` — this contains all issue 
 
 ### Translation Glossary (AUTHORITATIVE SOURCE)
 
-The EthGlossary API (`https://ethereum.org/api/glossary`) is the **authoritative source** for all Ethereum term translations across the entire pipeline. Community-voted glossary terms are not suggestions — they are the required translations.
+**ETHGlossary** is the authoritative source for Ethereum term translations. Deviations are critical issues, not warnings.
 
-**Fetch live from the API first, fall back to cache only if the API is unreachable:**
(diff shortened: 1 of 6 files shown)


pettinarip (approved):
LGTM

PR #18005, opened 2026-04-21
fix(Simulator): use live gas price for send transaction fee

pettinarip:
## Summary

- The simulator's network fee was hardcoded at `0.00042 ETH` (21k gas × 20 gwei), which over-represents real costs by ~20× at today's ~1 gwei base fee. At the current fallback ETH price this shows as `$0.42` when realistic values are closer to `~$0.05`.
- `/api/gas-eth-price` already returns `gasPrice` alongside `ethPriceUSD`; the simulator just wasn't consuming it.
- Renamed `useEthPrice` → `useGasEthPrice` to return `{ ethPrice, gasPrice }`, and derive the transfer fee in `SendReceive` as `ETH_TRANSFER_GAS_UNITS * gasPrice * 1e-9`.
- Replaced `ETH_TRANSFER_FEE` constant with `ETH_TRANSFER_GAS_UNITS = 21_000` and `FALLBACK_GAS_PRICE_GWEI = 1` (reflects typical post-Pectra base fee, keeps Chromatic snapshots deterministic via the existing `isChromatic()` guard).

## Test plan

- [x] On `/` homepage simulator, step through the send flow to step 5 ("You are sending") and verify network fee shows a realistic value (`~0.000021 ETH` at 1 gwei, with USD equivalent proportional to live ETH price)
- [x] Verify post-send balance on step 6 reflects the smaller fee deduction
- [x] Chromatic snapshot for `SendReceive / SendSummary` remains stable (uses 1 gwei fallback deterministically)
- [x] `/wallets` page wallet simulator still works end-to-end
- [x] Check that the `ConnectWeb3` screen still renders the final wallet state correctly (hook signature changed)


wackerow (approved):
@pettinarip Lgtm, thanks! Though the `ConnectWeb3` phase goes back to the original "$50" balance—can't recall if this was the intended behavior or not, but also doesn't necessarily need to block this improvement

pettinarip:
> the `ConnectWeb3` phase goes back to the original "$50" balance—can't recall if this was the intended behavior or not

That's pre-existing behavior — each simulator phase (CreateAccount, SendReceive, ConnectWeb3) is a self-contained demonstration and doesn't thread balance state from the previous phase. `ConnectWeb3` computes its ETH balance fresh from `USD_RECEIVE_AMOUNT / ethPrice` in `src/components/Simulator/screens/ConnectWeb3/index.tsx`. This PR only touches the gas-price-derived fee, so it doesn't change that. Worth revisiting separately if we want continuity across phases, but agreed it shouldn't block this fix.

PR #17978, opened 2026-04-15
Trailing seo fixes 

mnelsonBT:
Fix: trailing SEO remediations from last audit recommendations (404s, broken images, HTTP link normalization, and keyword optimization)

## Description

SEO improvements remaining from the recent SEO audit. The primary focus is on target keyword optimization, eliminating 404/4XX errors, preventing server-less 502 cascades, and improving link health across localized content. 
* Keyword optimization: Updated site title tag to lead with 'Ethereum'; updated a few page headers per auditor recommendation to better align with semantic search intents.
* Bulk HTTP to HTTPS link normalization: Bulk find & replace to move insecure `http://` ecosystem links with `https://` across MDX and translation files. Updated `src/components/ui/Link.tsx` to ensure any future HTTP links are handled at the component level.
* UI Fallbacks for broken assets:
  * Implemented a native fallback in `src/components/ui/avatar.tsx` for broken Crowdin contributor avatars to prevent 403/404 image load errors.
  * Added a dynamic `MapPin` UI fallback mechanism to the `EventCard` component to gracefully catch and handle broken remote event images.
* Serverless 502 Cascade Prevention: Updated the error handling for missing/broken app pages so they gracefully degrade to a standard 404 response instead of triggering cascading 502 server errors.
* Routing & translation fixes: Resolved cross-locale 404s by fixing broken redirect relative links and sanitizing malformed curly quotes in the Portuguese (`pt-br`) and Russian (`ru`) glossary source files that were causing localized redirect loops.

Changes addressed remaining audit export critical errors and low-hanging high impact warnings. Remaining flags on title length, duplicate H1s from frontmatter & title tag, image size, etc. will require another round of updates with more tactical review.

PR #18006, opened 2026-04-21
fix(seo): wrap remaining generateMetadata notFound() calls in try/catch

pettinarip:
## Summary

Two remaining `generateMetadata` functions on `dev` still call `notFound()` directly, which in Netlify's serverless runtime surfaces as a **502** instead of a 404 — poisoning hreflang reports (see Ahrefs export from 2026-04-17, 335 pages flagged with "Some pages are not returning 200 status code"). This patches both:

- `app/[locale]/developers/tools/[category]/page.tsx` — wraps the category-slug validation in try/catch, falls back to basic "page not found" metadata. Directly explains the 3 `/developers/tools/.../?toolId=…` 502s in the Ahrefs export.
- `app/[locale]/apps/categories/[catetgoryName]/page.tsx` — same pattern applied to the two `notFound()` calls (invalid slug + invalid enum). Not in the current Ahrefs export (category slugs come from `generateStaticParams`, so misses are rare), but the bug class is identical and worth closing while we're here.

The main `Page()` functions keep their existing `notFound()` calls untouched — they handle the actual 404 HTTP status correctly. Only `generateMetadata` needs the fallback.

## Companion to #17978

PR #17978 (`trailingSEO`) applied the same pattern to `app/[locale]/apps/[application]/page.tsx`, which accounts for the bulk of the 502 cascade (all `/apps/[slug]/` URLs). This PR closes the remaining gaps I found while sweeping the codebase for `generateMetadata` + `notFound()` combinations. After both land, a repeat Ahrefs crawl should show the hreflang-to-5xx counts collapse.

## Test plan

- [ ] CI green (typecheck + lint pass locally)
- [ ] Hit an invalid dev-tools category slug in a preview deploy — confirm 404, not 502
- [ ] Hit an invalid apps category slug in a preview deploy — confirm 404, not 502
- [ ] After merge + deploy, re-run Ahrefs crawl and confirm hreflang-to-5xx count drops

### DIFF
--- a/app/[locale]/apps/categories/[catetgoryName]/page.tsx
+++ b/app/[locale]/apps/categories/[catetgoryName]/page.tsx
@@ -36,8 +36,6 @@ import { slugify } from "@/lib/utils/url"
 
 import { appsCategories } from "@/data/apps/categories"
 
-import { DEFAULT_LOCALE } from "@/lib/constants"
-
 import AppsHighlight from "../../_components/AppsHighlight"
 import AppsTable from "../../_components/AppsTable"
 import SuggestAnApp from "../../_components/SuggestAnApp"
@@ -151,7 +149,7 @@ const Page = async (props: {
                   <BreadcrumbItem>
                     <BreadcrumbLink href="/">Ethereum.org</BreadcrumbLink>
                   </BreadcrumbItem>
-                  <BreadcrumbSeparator className="me-[0.625rem] ms-[0.625rem] text-gray-400">
+                  <BreadcrumbSeparator className="ms-[0.625rem] me-[0.625rem] text-gray-400">
                     /
                   </BreadcrumbSeparator>
                   <BreadcrumbItem>
@@ -159,7 +157,7 @@ const Page = async (props: {
                       {t("page-apps-all-apps")}
                     </BreadcrumbLink>
                   </BreadcrumbItem>
-                  <BreadcrumbSeparator className="me-[0.625rem] ms-[0.625rem] text-gray-400">
+                  <BreadcrumbSeparator className="ms-[0.625rem] me-[0.625rem] text-gray-400">
                     /
                   </BreadcrumbSeparator>
                   <BreadcrumbItem>
@@ -257,10 +255,7 @@ export async function generateMetadata(props: {
       description,
     })
   } catch {
-    const t = await getTranslations({
-      locale: DEFAULT_LOCALE,
-      namespace: "common",
-    })
+    const t = await getTranslations("common")
 
     return {
       title: t("page-not-found"),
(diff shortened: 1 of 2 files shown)


app/[locale]/apps/categories/[catetgoryName]/page.tsx
```diff
@@ -223,35 +225,48 @@ export async function generateMetadata(props: {
...
+      locale,
+      slug: ["apps", "categories", normalizedSlug],
+      title,
+      description,
+    })
+  } catch {
+    const t = await getTranslations({
+      locale: DEFAULT_LOCALE,
+      namespace: "common",
+    })
```
Comment (wackerow) [inline app/[locale]/apps/categories/[catetgoryName]/page.tsx]:
```suggestion
    const t = await getTranslations("common")
```

app/[locale]/developers/tools/[category]/page.tsx
```diff
@@ -152,20 +154,32 @@ export async function generateMetadata(props: {
...
+      title: t(`page-developers-tools-category-${category}-title`),
+      description: t(
+        `page-developers-tools-category-${category}-meta-description`
+      ),
+    })
+  } catch {
+    const t = await getTranslations({
+      locale: DEFAULT_LOCALE,
+      namespace: "common",
+    })
```
Comment (wackerow) [inline app/[locale]/developers/tools/[category]/page.tsx]:
```suggestion
    const t = await getTranslations("common")
```

wackerow (approved):
@pettinarip Looks good! I believe we should be able to preserve the `locale` though instead of forcing English. Suggestions below, let me know if you disagree -- otherwise good to merge!

app/[locale]/apps/categories/[catetgoryName]/page.tsx
Comment (pettinarip) [reply app/[locale]/apps/categories/[catetgoryName]/page.tsx]: > Despite use of `locale: DEFAULT_LOCALE` this does seem to return the Spanish strings 🤷‍♂️

Good catch — I dug into why: `src/i18n/request.ts` builds the messages map once per request from `requestLocale` (deep-merging the request locale over English). Passing an explicit `locale` to `getTranslations` only changes the *formatting* locale (Intl date/number), not the messages that get looked up. So `locale: DEFAULT_LOCALE` was a no-op for strings on this project — Spanish came through either way.

Applied the suggestion in 46629213dc — `getTranslations("common")` is honest about what was actually happening and preserves the user's locale for the `page-not-found` copy, which is the better UX anyway.

app/[locale]/developers/tools/[category]/page.tsx
Comment (pettinarip) [reply app/[locale]/developers/tools/[category]/page.tsx]: > Despite use of `locale: DEFAULT_LOCALE` this does seem to return the Spanish strings 🤷‍♂️

Good catch — I dug into why: `src/i18n/request.ts` builds the messages map once per request from `requestLocale` (deep-merging the request locale over English). Passing an explicit `locale` to `getTranslations` only changes the *formatting* locale (Intl date/number), not the messages that get looked up. So `locale: DEFAULT_LOCALE` was a no-op for strings on this project — Spanish came through either way.

Applied the suggestion in 46629213dc — `getTranslations("common")` is honest about what was actually happening and preserves the user's locale for the `page-not-found` copy, which is the better UX anyway.

PR #18007, opened 2026-04-21
perf: lazy-load layer-2 networks table

pettinarip:
## Summary

- Lazy-load `Layer2NetworksTable` on `/layer-2/networks` using the project's established `lazy.tsx` + `loading.tsx` convention (same pattern as `FindWalletProductTable`)
- Drops the SSR + hydration cost of the interactive table; hero, callouts, and `NetworkMaturity` remain server-rendered so SEO is unaffected
- As a side effect, fixes a hydration mismatch inside `ProductTable` → `MediaQuery` on first paint (server renders without the desktop filter column, client was trying to mount it)
- Removes the passthrough `_components/networks.tsx` wrapper and the now-pointless `<Suspense>` around the client tree
- Adds an explicit `ExtendedRollup` cast on `mainnetData` at the page boundary — the old untyped wrapper had been masking that `ethereumNetworkData` doesn't fully satisfy the type (runtime behavior unchanged)

## Test plan

- [ ] Visit `/layer-2/networks` on desktop — no hydration error in console, skeleton flashes briefly before the table mounts
- [ ] Visit on mobile width — mobile filter toggle + row list render correctly
- [ ] Expand a network row, open mobile filters, apply a filter — behavior matches `dev`
- [ ] `pnpm build` passes, route First Load JS for `/layer-2/networks` drops vs `dev`

## Preview link
https://deploy-preview-18007.ethereum.it/layer-2/networks


src/components/Layer2NetworksTable/hooks/useNetworkColumns.tsx
```diff
@@ -79,17 +79,12 @@ export const useNetworkColumns: ColumnDef<ExtendedRollup & { id: string }>[] = [
...
-                      $
-                      {numberFormat(meta.locale as Lang, {
+                  {row.original.txCosts
+                    ? `$${numberFormat(meta.locale as Lang, {
                         minimumFractionDigits: 2,
                         maximumFractionDigits: 3,
-                      }).format(row.original.txCosts || 0)}
-                    </>
-                  ) : (
-                    <p>-</p>
```
pettinarip, line 91: this was causing a hydration issue (p inside a p)

app/[locale]/layer-2/networks/_components/networks.tsx
```diff
@@ -1,109 +0,0 @@
...
-    title: t("common:nav-networks-explore-networks-label"),
-    description: t("page-layer-2-networks-hero-description"),
-  }
-
-  return (
-    <>
-      <MainArticle className="relative flex flex-col">
-        <ContentHero {...heroProps} />
-
-        <Layer2NetworksTable
```
Comment (pettinarip) [inline app/[locale]/layer-2/networks/_components/networks.tsx:32]: move all into the server-page level component, except form `Layer2NetworksTable` that is lazy loaded. This way we can have the h1 and content server rendered

wackerow (approved):

PR #18009, opened 2026-04-21
test(visual): add Playwright + Chromatic page visual regression suite

pettinarip:
## Summary

- Add a Playwright-driven page visual regression suite that renders real Next.js pages (covering each layout under `src/layouts/`) at three viewports and uploads DOM archives to Chromatic.
- Wire a dedicated `chromatic-pages` GitHub Actions workflow that builds the production app with mock data, runs Playwright, and hands off archives to Chromatic — aligned with the official Playwright + Chromatic pattern.
- Build via `pnpm build` with `USE_MOCK_DATA=true` and `NEXT_PUBLIC_BUILD_LOCALES=en` so tests exercise the same production build the site ships (Turbopack since #17906), deterministically.
- Introduce a `data-slot="loading"` contract on the shared `Skeleton` and `Spinner` primitives so tests wait on one signal for all loaders to settle before snapshotting — no per-page hydration selectors.
- Simplify `EthPriceCard` to use the cached `useGasEthPrice` hook — removes the bespoke CoinGecko fetch and **drops the 24hr % change indicator** (arrow + percentage + colored gradient) on the homepage. The 24hr value drifted every snapshot and was the dominant flake source for the homepage layout.

## Test plan

- [x] CI: `chromatic-pages` workflow succeeds on this PR and publishes a Chromatic build
- [x] Chromatic build shows snapshots for each layout under `src/layouts/`
- [ ] No regressions in the existing `chromatic.yml` Storybook workflow
- [x] Re-running the workflow produces stable snapshots (no flakiness from hydration / network)


myelinated-wackerow (approved):
LGTM on the visual-test infrastructure. A few actionables:

- **PR description says "tests exercise the same webpack output the site ships"** — stale after the #17906 merge. The workflow runs `pnpm build`, which is now Turbopack. Either update the description, or pin the workflow to `pnpm build:webpack` if webpack-specific output matters for snapshot fidelity.
- **`last-24-hrs` translation key is orphaned** — `EthPriceCard` no longer renders the 24hr % change, so the key is unused in source code but still present in `src/intl/{locale}/common.json` across all 25 locales. Remove it from the namespace if the deprecation is intentional.
- **EthPriceCard quietly drops the 24hr % change indicator** (arrow + percentage + colored gradient). Probably worth a one-line mention in the PR body so it's not invisible to reviewers and stakeholders watching the homepage.
- **Naming nit on `maybeShuffle`** — `safeShuffle` reads better; the "safe" framing makes the intent (deterministic for visual tests) clearer than the conditional-sounding "maybe."
- **Minor: `reuseExistingServer: true` in `playwright.visual.config.ts`** — fine for CI, but locally a `pnpm dev` on :3000 will be silently used in place of the production build the tests assume. Worth a one-liner in the skill doc's "Common situations" section.

Skill doc is genuinely useful — covers the non-obvious gotchas (1024 viewport, dual config, `data-slot` contract, `HOME: /root`).

_Reviewed by Claude (Opus 4.7)_

pettinarip:
> **PR description says "tests exercise the same webpack output the site ships"** — stale after the #17906 merge. The workflow runs `pnpm build`, which is now Turbopack. Either update the description, or pin the workflow to `pnpm build:webpack` if webpack-specific output matters for snapshot fidelity.

Addressed: updated the PR description to say "the same production build the site ships (Turbopack since #17906)". The workflow intentionally uses `pnpm build` so the visual tests stay aligned with whatever build the site actually ships — pinning to `build:webpack` would defeat that.

pettinarip:
@wackerow FYI, I've refactored the `EthPriceCard` component again because I noticed I had removed the "last 24hrs change" from the UI. Reverted the UI changes and also added the data in the data layer.

PR #18011, opened 2026-04-21
fix(seo): JSON-LD validation and consistency patches

myelinated-wackerow:
## Summary

Four JSON-LD fixes packaged together.

## Changes

- **Correct `reviewedBy` domain.** Per schema.org, `reviewedBy` is valid only on `WebPage` and its subtypes. ~~Swapped to `editor` on `Article` and `VideoObject` nodes (both `CreativeWork`), and~~ removed from `ItemList` nodes (no equivalent property applies). _(Edit: `editor` does not accept `Organization` type, only `Person` -- has been removed)_

- **Drop `/en/` prefix from JSON-LD URLs.** Hardcoded `https://ethereum.org/${locale}/...` URLs now route through `normalizeUrlForJsonLd`, which respects `localePrefix: "as-needed"` (no prefix for the default locale).

- **Extract `/staking/deposit-contract/` JSON-LD to a sibling file.** Previously inline inside the React component; now follows the standard `page-jsonld.tsx` pattern alongside every other route. Uses the canonical EF/Community organization entities instead of a bogus inline "ethereum.org" Organization.

- **DRY common `@graph` entities.** Extracts the ethereum.org `WebSite` entity and the trio of top-level graph nodes (EF Organization, Community Organization, WebSite) into shared constants. Each page now spreads `...baseGraphNodes` instead of repeating them, and references the WebSite via `@id` in `isPartOf`.

## Note

Parallel to the author attribution work in:
- #17988

Both touch `src/lib/utils/jsonld.ts` and many `page-jsonld.tsx` files but are not dependent -- whichever lands first, the other will need a merge resolution.

---

Generated by Claude Opus 4.7


pettinarip (commented):
@myelinated-wackerow nice cleanup

**Blockers / needs clarification**

1. **PR description vs. diff mismatch on `editor`.** The summary says `reviewedBy` was "swapped to `editor` on `Article` and `VideoObject` nodes," but I can't find a single `editor:` insertion in the diff — `reviewedBy` is simply removed from those nodes.
2. **`deposit-contract/page-jsonld.tsx` drops the `Article` node.** Every other extracted page emits both a `WebPage` and an `Article`; this new file emits only a `WebPage`.

**Nit**

3. `app/[locale]/staking/deposit-contract/page-jsonld.tsx` doesn't thread `contributors` through, so the `WebPage` lacks the `contributor` array the other extracted pages carry. Easy parity win now that it's a proper sibling file.

mnelsonBT (approved):
LGTM @wackerow , love seeing these graphs in tight shape! Defer to your call on the deposit contract page; I'm aligned with keeping that one clean as it's a core source for the contract.

wackerow (approved):
Thanks @pettinarip!

> 1. **PR description vs. diff mismatch on `editor`.** The summary says `reviewedBy` was "swapped to `editor` on `Article` and `VideoObject` nodes," but I can't find a single `editor:` insertion in the diff — `reviewedBy` is simply removed from those nodes.

Good catch, this was updated after the PR was posted—I've updated the PR description. Schema.org requires `editor` to be a `Person` not an `Organization` so this has been removed. The `reviewedBy` remains inconsistent with `Article`, so remains removed as well.

> 2. **`deposit-contract/page-jsonld.tsx` drops the `Article` node.** Every other extracted page emits both a `WebPage` and an `Article`; this new file emits only a `WebPage`.

This page isn't actually an article so semantically correct to leave off `Article` type for this page (likely others we could fine-tune as we keep iterating on this) cc: @mnelsonBT 

> 3. `app/[locale]/staking/deposit-contract/page-jsonld.tsx` doesn't thread `contributors` through, so the `WebPage` lacks the `contributor` array the other extracted pages carry. Easy parity win now that it's a proper sibling file.

Similarly, this page is one that doesn't really accept contributors (no `FileContributors` component nor "Edit" button) given the nature of this page, so was intentionally kept off.

---

This refactor causes breaking changes for the Authors changes in PR #17988 so I'm going to pull this in and update the other changes from there. 

Thanks folks!

PR #18012, opened 2026-04-21
fix(deps): patch fast-xml-parser critical CVE (GHSA-m7jm-9gc2-mpf2)

myelinated-wackerow:
## Summary

Resolves Dependabot alert [#240](https://github.com/ethereum/ethereum-org-website/security/dependabot/240) — **GHSA-m7jm-9gc2-mpf2 / CVE-2026-25896** (CVSS 9.3, critical) in `fast-xml-parser`.

Lockfile-only change via `pnpm update fast-xml-parser` and `pnpm update @aws-sdk/client-s3`. Resolved version moves from **5.3.4 → 5.5.8**, which is above the `5.3.5` patch floor.

## Why this fix

- **The vulnerability**: a `.` in a DOCTYPE entity name is treated as a regex wildcard during entity replacement, allowing an
attacker to shadow built-in XML entities (`&lt;`, `&gt;`, `&amp;`, `&quot;`, `&apos;`) with arbitrary values. This bypasses entity
encoding and leads to XSS when parsed output is rendered. Patched upstream in 5.3.5.
- **How it reaches us**: `fast-xml-parser` is a pure transitive dependency — the only consumer chain is `@aws-sdk/xml-builder` via
`@aws-sdk/client-s3` and `@aws-sdk/credential-provider-node`, and it resolves to a single version. We do not import it directly
in app code.
- **Why no pnpm.overrides entry**: `@aws-sdk/xml-builder@3.972.4` declares `fast-xml-parser: ^5.0.0`, so normal resolution picks
`5.5.8` without needing a permanent override. Keeps the diff minimal — lockfile only, no package.json change.

## Breaking-change risk

Low.

- `fast-xml-parser` 5.3.4 → 5.5.8 is a minor bump within the same major; no documented breaking changes in that window.
- The library is only exercised inside AWS SDK internals (S3 XML response parsing). Our only AWS usage is the data layer's
S3-compatible storage (`s3-dcl1.ethquokkaops.io`) serving app screenshots for `/dapps`.
- No new peer-dependency conflicts were introduced.

## Test plan

- [ ] CI passes (lint, typecheck, build)
- [ ] Netlify preview deploy succeeds
- [ ] `/dapps` and app listing pages load images correctly (verifies S3 data-layer read path still works)
- [ ] Dependabot alert [#240](https://github.com/ethereum/ethereum-org-website/security/dependabot/240) closes after merge

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "trigger:deploy": "trigger deploy --env-file src/data-layer/.env.local"
   },
   "dependencies": {
-    "@aws-sdk/client-s3": "^3.982.0",
+    "@aws-sdk/client-s3": "^3.1035.0",
     "@aws-sdk/client-ses": "^3.859.0",
     "@docsearch/css": "3.9.0",
     "@docsearch/react": "^3.5.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -16,8 +16,8 @@ importers:
   .:
     dependencies:
       '@aws-sdk/client-s3':
-        specifier: ^3.982.0
-        version: 3.982.0
+        specifier: ^3.1035.0
+        version: 3.1035.0
       '@aws-sdk/client-ses':
         specifier: ^3.859.0
         version: 3.859.0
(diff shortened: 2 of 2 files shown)


pettinarip (commented):
@myelinated-wackerow quick question before merge: do we need the override? `@aws-sdk/xml-builder@3.972.4` already declares `fast-xml-parser: ^5.0.0`, so `pnpm update fast-xml-parser` (or `pnpm update @aws-sdk/client-s3`) should resolve to 5.7.1 through normal resolution — just a lockfile change, no permanent `package.json` entry.

wackerow:
@pettinarip yeah good call, cleaner -- updated PR and description

pettinarip (approved):

Issue #18023, opened 2026-04-23
[Bug report] /zh/contributing/ page: #new-to-open-source anchor link broken — heading missing id attribute


jzhishu:
### Describe the bug

On the `/zh/contributing/` page, the heading "新接触开源项目？ 开源新手？" is missing its anchor `id`, causing in-page links that reference `#new-to-open-source` to not work.

### To reproduce

1. Go to `/zh/contributing/`
2. Try navigating to `/zh/contributing/#new-to-open-source`
3. The page does not scroll to the corresponding heading

### Expected behavior

The heading should have `id="new-to-open-source"` so that anchor links like `/zh/contributing/#new-to-open-source` correctly jump to that section, consistent with the behavior on other language versions of the same page.

### Screenshots

<img width="1936" height="1048" alt="Image" src="https://github.com/user-attachments/assets/7336fc88-9d49-4c8a-8f59-43c57d8187ab" />

### Desktop (please complete the following information)

- OS: macOS
- Browser: Chrome / Safari

### Additional context

Affected file: `public/content/translations/zh/contributing/index.md`, line 88

```markdown
## 新接触开源项目？ 开源新手？{#new-to-open-source}
```

### Would you like to work on this issue?

- [x] Yes
- [ ] No

jzhishu:
I will fix this issue and submit the pr soon

wackerow:
Closed with:
- #18024

PR #18024, opened 2026-04-23
fix: miss heading id

jzhishu:
## Description

Fixes a broken anchor link on the `/zh/contributing/` page.
The heading on line 88 of `public/content/translations/zh/contributing/index.md` was missing a space before `{#new-to-open-source}`, causing the `remark-heading-id` plugin to silently drop the ID. As a result, the heading was rendered without an `id` attribute and anchor links to `#new-to-open-source` did not work.

## Related Issue

[https://github.com/ethereum/ethereum-org-website/issues/18023](issues-18023)

## Follow-up suggestion
Perhaps a new markdownlint rule could be added to handle this and prevent recurrence.


wackerow (approved):
Thanks @jzhishu, good catch!

And to your comment about a markdownlint rule, yes! That has since been implemented, along with several updates to our translation pipeline.

The existing content was produced prior to those updates, but soon we should be running another full pass of the website which should enforce the exact thing your fixing and the rule your commenting about. Thanks again!

wackerow:
@all-contributors please add @jzhishu for bug fix

PR #18026, opened 2026-04-23
seo(jsonld): VideoGallery + WebPage for video pages

wackerow:
## Summary
- `/videos/`: promoted `CollectionPage` to `VideoGallery` (more specific subclass per Schema.org's `CollectionPage > MediaGallery > VideoGallery` hierarchy). `ItemList` retained as `mainEntity` for carousel rich-result eligibility, and the 10-item slice was dropped so the full catalog is exposed to crawlers.
- `/videos/[slug]/`: wrapped the previously orphaned `VideoObject` in a proper `@graph`. Spreads `BASE_GRAPH_NODES`, adds a `WebPage` node that owns the 3-item breadcrumb and points `mainEntity` at the `VideoObject`, and adds `VideoObject.isPartOf` → the gallery. `VideoObject.publisher` switched from the full `ORGANIZATION.ETHEREUM_FOUNDATION` object to `REFERENCE.ETHEREUM_FOUNDATION` (the full object is already declared in `BASE_GRAPH_NODES`).
- Added a minimal `{"@type": "VideoGallery", "@id": videoGalleryUrl}` stub on the viewing page so `VideoObject.isPartOf`'s cross-page `@id` resolves to a typed entity instead of the generic `CreativeWork` fallback strict validators infer from an unresolved reference.

## Key decisions
- **`VideoGallery` over `CollectionPage`** — more specific, and the subclass chain (`VideoGallery is-a CollectionPage is-a WebPage`) keeps it backward-compatible with anything that only understood the parent.
- **`WebPage` wrapper on video pages** — matches the pattern used by every other page on the site. The previously bare `VideoObject` worked for Google's video rich result but was semantically orphaned from the site graph.
- **`VideoObject.isPartOf` points to the `VideoGallery`**, not the `WebPage` or `WebSite`. A video is naturally a part of the gallery collection; pointing at the page or the site is technically valid but semantically weak.
- **Minimal type-assertion stub for `VideoGallery`** on the viewing page — `@id` references work across documents, but not every parser reconciles cross-document graphs. A `{@type, @id}` stub keeps the viewing page's graph self-contained without duplicating the canonical definition from `/videos/page-jsonld.tsx`.

## Test plan
- [x] Render `/videos/` locally, paste the `application/ld+json` from the DOM into validator.schema.org — expect `VideoGallery` with `ItemList` nested via `mainEntity`, no dangling references.
- [x] Render `/videos/[slug]/` for at least one slug, paste into validator.schema.org — expect `WebPage` with `VideoObject` nested via `mainEntity`, `VideoGallery` stub nested under `VideoObject.isPartOf`, and no generic `CreativeWork` fallback.
- [x] Browser smoke-test the gallery and at least one video page.
- [x] Verify breadcrumb renders correctly on the video page UI.

_Message generated by Claude (Opus 4.7)_

### DIFF
--- a/app/[locale]/videos/[slug]/page-jsonld.tsx
+++ b/app/[locale]/videos/[slug]/page-jsonld.tsx
@@ -1,3 +1,5 @@
+import { getTranslations } from "next-intl/server"
+
 import type { VideoFrontmatter } from "@/lib/interfaces"
 
 import PageJsonLD from "@/components/PageJsonLD"
@@ -6,9 +8,10 @@ import { stripMarkdown } from "@/lib/utils/md"
 import { toIsoDuration } from "@/lib/utils/time"
 import { normalizeUrlForJsonLd } from "@/lib/utils/url"
 import { getDefaultThumbnailUrl } from "@/lib/utils/videos"
-import { ORGANIZATION } from "@/lib/jsonld/constants"
 
-export default function VideoPageJsonLD({
+import { BASE_GRAPH_NODES, REFERENCE } from "@/lib/jsonld/constants"
+
+export default async function VideoPageJsonLD({
   locale,
   slug,
   frontmatter,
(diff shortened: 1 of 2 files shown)


mnelsonBT (approved):
Looks excellent - thank you @wackerow !

wackerow:
@pettinarip Going to pull this one in per @mnelsonBT's approval so it can be part of the deploy. Please note if you see any issues though and I'll make sure to patch in a follow up.

PR #18039, opened 2026-04-24
feat(videos): sync thumbnails to S3

myelinated-wackerow:
## Summary

Adds a daily Trigger.dev task that downloads video thumbnails from YouTube (or customThumbnailUrl) to our S3 bucket, and wires the /videos gallery to serve those S3-hosted images instead of hitting img.youtube.com directly.

## Motivation

YouTube's thumbnail CDN is inconsistent:
- `sddefault.jpg` (best quality) returns 404 for some videos
- `hqdefault.jpg` is the documented "guaranteed" fallback but has been flaky in practice (broken thumbnails observed on production /videos recently)
- Direct YouTube URLs go through Next.js' image optimization proxy, which errors server-side on 404s before client-side `onError` can fire -- so a transient YouTube 404 becomes a broken image users see

Syncing to S3 solves this with a single reliable origin, and gives us the progressive-quality pick (sd > hq) at sync time instead of render time.

Follows the existing `fetchApps` / `fetchEvents` / developer-tools pattern.

### New

- `src/data-layer/fetchers/fetchVideoThumbnails.ts` -- daily task. For each video: uploads `customThumbnailUrl` if present, else tries `sddefault.jpg` and falls back to `hqdefault.jpg`. Keyed by slug so each video maps to exactly one S3 object.
- `VIDEO_THUMBNAILS` key + registration in `src/data-layer/tasks.ts` DAILY
- `getVideoThumbnails()` getter in `src/data-layer/index.ts`
- Cached wrapper in `src/lib/data/index.ts`

### Modified

- `src/data-layer/s3.ts` -- `uploadToS3` gains an optional `customFilename` 3rd param so callers with a stable ID (e.g., slug) can override the URL-derived hash key with a human-readable filename. Existing callers unaffected.
- `src/lib/utils/videos.ts` -- `getVideos()` fetches the thumbnail map in parallel with the slug scan; `toVideoCardData()` uses the S3 URL. `getDefaultThumbnailUrl()` removed.
- `app/[locale]/videos/[slug]/page-jsonld.tsx` -- `thumbnailUrl` inlined to point at the canonical YouTube URL (JSON-LD is plain string metadata for crawlers, doesn't pass through Next.js Image).
- `next.config.js` -- `img.youtube.com` removed from `remotePatterns` (no longer rendered via Next.js Image anywhere).

## Ops notes

- First deploy: trigger the task manually once so the map populates before users hit the gallery. The getter in `getVideos()` has a graceful fallback (`.catch(() => null)`), so a missing/empty map doesn't break the page -- but cards won't show thumbnails until the sync has run.

## Test plan

- [x] Run `fetch-video-thumbnails` task in Trigger.dev dev environment
- [x] Confirm S3 bucket contains `videos/thumbnails/<slug>.jpg` for each video
- [x] Spot-check a few S3 URLs render in the browser
- [x] Load `/videos` locally and confirm gallery cards use S3 URLs
- [x] Load a `/videos/[slug]` page and inspect JSON-LD -- `thumbnailUrl` should be a YouTube URL
- [x] Re-run the task; verify no redundant uploads (HEAD check short-circuits)
- [ ] After merge, trigger the task in production; verify /videos renders from S3

---


pettinarip (approved):
LGTM!

PR #18062, opened 2026-04-28
fix(intl-pipeline): support directory paths in TARGET_PATH

myelinated-wackerow:
## Summary

Restores historical behavior: passing a directory as `target_path` (e.g. `public/content/videos/`) now expands to all `.md` files under it (or `.json` files under `src/intl/en/`) and translates each. Previously the pipeline tried `readFileSync` on the directory and threw `EISDIR`.

## Test plan

- [ ] Dispatch `intl-pipeline.yml` with `target_path=public/content/videos/`, single locale → expansion logged, all files translated.
- [ ] Single-file `target_path` (existing behavior) still works.
- [ ] Comma-separated mix of files + directories works.

Generated by Claude Opus 4.7

### DIFF
--- a/src/scripts/intl-pipeline/main.ts
+++ b/src/scripts/intl-pipeline/main.ts
@@ -155,6 +155,25 @@ async function loadGlossary(
   }
 }
 
+/**
+ * Recursively collect files under `dir` whose name ends with `ext`.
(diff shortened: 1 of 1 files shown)


wackerow (approved):
cc: @pettinarip pulling this patch in -- "directory" approach was lost in the migration (only comma separated files were working) -- this restores that functionality.

PR #18067, opened 2026-04-29
fix(seo): ssr the wallet list on /wallets/find-wallet

### LINKED ISSUE #17717: fix(seo): /wallets/find-wallet/ serves no wallet content to search engines (ssr: false)
## Description

The `/wallets/find-wallet/` page renders **zero wallet content in its initial HTML**. Google and other search engines see a title, one sentence of description, and 124 skeleton placeholder divs with `animate-pulse` — but no actual wallet listings.

### Root cause

`src/components/FindWalletProductTable/lazy.tsx` uses:
```js
export default dynamic(() => import("."), { ssr: false, loading: Loading })
```

This explicitly disables server-side rendering for the entire wallet product table. Combined with `react-virtual` (added Sep 2025), even after JavaScript execution only visible-viewport wallets exist in the DOM — Googlebot doesn't scroll.

### What Google sees

- Page title: "List of Ethereum Wallets | ethereum.org"
- Heading: "Choose your wallet"
- One sentence: "Wallets store and transact your ETH..."
- **124 empty skeleton placeholders**
- Zero wallet names, zero descriptions, zero links to wallets

All wallet data (MetaMask, Ledger, Rabby, Rainbow, etc.) exists only inside `self.__next_f.push()` RSC script payloads — not in the rendered DOM.

### Impact

- Page likely classified as **thin content** by Google's indexer
- YMYL (Your Money Your Life) finance page with no substantive content in HTML
- Google's December 2025 core update specifically targeted thin comparison/tool pages
- The page gets ~35-40k monthly visits but could rank significantly higher with proper SSR

### Suggested fix

The wallet data is already fetched server-side in `page.tsx` and passed as props. Remove `ssr: false` from the dynamic import so the wallet table renders on the server. For the virtual list, render at minimum the first 20-30 wallet items as real DOM elements server-side, then virtualize on client after hydration.

### Verification

Fetch the page with curl or `view-source:` and search for any wallet name (e.g., "MetaMask") — you'll find it only in `<script>` tags, never in rendered HTML.

pettinarip:
## Summary

Fixes #17717 — Googlebot was indexing the loading skeleton on `/wallets/find-wallet` (zero wallet content in SSR HTML). This branch restores SSR coverage for SEO, then defends INP against the regression that SSR introduces.

**SSR coverage**
- Render the wallet list and filter sidebar on the server: drop `ssr: false`, drop `MediaQuery` wrappers, drop the dynamic-import skeleton, replace `useSearchParams()` with `window.location.search` to avoid Next's silent CSR-bailout on this SSG'd page.

**INP defense**
- Drop list virtualization; lazy-mount the heavy expanded sub-component per row.
- Replace Radix `Collapsible` with native `<details>`/`<summary>` (with the same `focus-visible` ring as `Button`).
- Unify `WalletInfo` desktop + mobile trees behind a single CSS Grid (was two parallel hidden trees, both hydrating).
- Hide-not-unmount filtered rows via a `matchedIds: Set<string>` — filter clicks no longer churn 52 subtrees.
- `Row` is `memo`'d and the lazy sub-component is passed via `children`, so closed rows skip re-render on filter changes.
- Tracking effect on the expanded sub-component fires once per expand (was once per render).

## Verification

| | dev.ethereum.org (deployed) | this branch |
|---|---|---|
| Wallet names in SSR HTML | **0** | **52** |

INP (Playwright iPhone 12, 4× CPU throttle):

| Scenario | dev.ethereum.org | this branch |
|---|---|---|
| Filter open — steady p75 | 24 ms | 32 ms |
| Wallet expand — steady p75 | 40–48 ms | 48–56 ms |
| Preset filter — steady p75 | 0–16 ms | 0–16 ms |
| Filter open — early p75¹ | 24–32 ms | 432 ms |
| Wallet expand — early p75¹ | 96–112 ms | 456–488 ms |

¹ dev's "early" measurement happens post-hydration since the page renders client-side; not directly comparable. Steady-state INP is equivalent. The remaining early-click cost is concentrated in the hydration window of the 52 wallet trees the page now ships.

Wire size (brotli, deploy-preview-18067 vs prod): 78 kb → 107 kb (+29 kb). Uncompressed grows 535 kb → 1452 kb because we now ship 52 cards' worth of markup + inline lucide SVGs; brotli compresses the repetition well, so wire impact is small.

## Test plan

- [x] SSR HTML contains all 52 wallets
- [x] Visual parity at desktop + mobile, open + closed
- [x] Filter toggle keeps 52 rows in DOM, hides non-matches
- [x] URL params pre-filter correctly
- [x] No-results UI + reset still work
- [x] `Layer2NetworksTable` (other `ProductTable` consumer) unaffected
- [x] Visit-website link inside `<summary>` doesn't toggle the row (manual)
- [x] TypeScript + ESLint pass
- [x] CI green on Netlify
- [ ] CrUX / RUM after merge to confirm real-world INP impact


src/components/FindWalletProductTable/WalletInfo.tsx
```diff
@@ -59,120 +57,79 @@ const WalletInfo = ({ wallet }: WalletInfoProps) => {
   }, [wallet.supportedLanguages])
 
   return (
-    <div className="flex flex-col gap-4">
+    <div className="relative flex flex-col gap-4">
+      {/* Open-state stripe (desktop only), sits in the image-column gutter. */}
+      <div
+        aria-hidden
+        className={`pointer-events-none absolute top-14 -bottom-9 left-7 hidden w-1 -translate-x-1/2 lg:group-[[open]]/collapsible:block ${wallet.twBackgroundColor}`}
```
wackerow: Image

wackerow: Patched in 9923abf778f90201f4555cc0cccf392bf860f3c6

wackerow (approved):
Looks good @pettinarip! Spotting a could things we could clean up, but they don't need to block.

The gutter decoration line loads a little buggy when expanding the card—could be nice to see if there is a cleaner way to have this load as a single dom element, perhaps even with a smooth animation so it's not so abrupt. 

Also some clean up here with string interpolating classes–could use `cn()` to pass this as it's own arg.

Pushed a commit to patch an RTL bug, adjusting the positioning of that gutter decoration. Pulling in!

PR #18071, opened 2026-04-29
fix: hide misleading update date and empty contributors modal when data is missing

### LINKED ISSUE #18070: App Router pages show "Page last update: January 1, 1970" and empty "See contributors" modal
### Describe the bug

App Router pages that use `getAppPageContributorInfo()` show two related visible defects:

1. **"Page last update: January 1, 1970"** — the Unix epoch, instead of a real date.
2. **"See contributors" opens an empty modal** — the avatar group, "Thanks to..." text, and an empty list, with no contributors shown.

The two symptoms share a single root cause: `appPages[pagePath]` resolves to an empty array for these pages, and the downstream code does not handle the empty case defensively.

### Affected pages (sample, likely more)

- https://ethereum.org/what-is-ethereum/
- https://ethereum.org/what-is-ether/

I have not enumerated every affected page, but the pattern applies to App Router pages calling `getAppPageContributorInfo(...)` — see `app/[locale]/what-is-ethereum/page.tsx`, `app/[locale]/what-is-ether/page.tsx`, `app/[locale]/gas/page.tsx`, `app/[locale]/learn/page.tsx`, `app/[locale]/roadmap/_vision/page.tsx`, etc.

### To reproduce

1. Open https://ethereum.org/what-is-ethereum/ in any browser.
2. Look at the top of the article. You will see "Page last update: January 1, 1970".
3. Click "See contributors". A modal opens with the "Thanks to..." text and an empty list — no avatars, no usernames.

Confirmed in production on 2026-04-29.

### Expected behavior

- The page should not display "Page last update: January 1, 1970" — that's the Unix epoch and is clearly wrong. It should either show the real last-edit date, or hide the line entirely if the date is unknown.
- The "See contributors" affordance should not exist when there are no contributors to show — clicking a button that opens an empty modal is broken UX.

#### Symptom-level (UI / utility code)

Both symptoms reduce to: `getAppPageContributorInfo(pagePath, locale)` returns `contributors = []` and the chain that consumes it is not empty-safe.

**`src/lib/utils/gh.ts:3`** — `getAppPageLastCommitDate` reduces an empty array starting from `new Date(0)` and `.toString()`s the result. So given `[]`, it returns the Unix epoch as a real, formattable date string. Downstream `getLocaleTimestamp` then formats this as "January 1, 1970", which is truthy, so the FileContributors component renders it.

```ts
// src/lib/utils/gh.ts
export const getAppPageLastCommitDate = (
  gitHubContributors: FileContributor[]
) =>
  gitHubContributors
    .reduce((latest, contributor) => {
      const commitDate = new Date(contributor.date)
      return commitDate > latest ? commitDate : latest
    }, new Date(0))         //  ← returns epoch when array is empty
    .toString()
```

**`src/components/FileContributors.tsx:127`** — the component renders the "See contributors" button and modal unconditionally, with no guard for `contributors.length === 0`. So an empty `contributors` array still produces a clickable button that opens an empty modal.

#### Underlying data cause (suspected, needs verification by maintainers)

I believe `appPages["what-is-ethereum"]` (and likely most or all App Router pages) is **missing or empty in the GitHub-contributors blob storage** that backs `dataLayer.getGitHubContributors()`. The fetcher at `src/data-layer/fetchers/fetchGitHubContributors.ts:443` populates this from the GitHub commits API on a weekly schedule and writes to Netlify Blobs. If that data is stale, missing, or never populated for App Router pages, every page using `getAppPageContributorInfo` will show these symptoms.

In `fetchGitHubContributors.ts:535`, an `if (contributors.length > 0)` filter omits empty results from the stored map — so the absence of a key is indistinguishable from a fetch failure.

I do not have access to production blob storage to verify which case applies. **This part needs a maintainer with access to inspect.**

### Proposed fix (symptom-level, defensive)

I'm planning to open a PR with two small, defensive changes that make the UI behave correctly even when the contributors array is empty:

1. `getAppPageLastCommitDate` returns `null` (or the function is updated and its caller propagates `null`) when given an empty array, instead of fabricating an epoch timestamp.
2. `FileContributors` returns `null` when `contributors.length === 0` and there is no `lastEditLocaleTimestamp`, so the entire block (including the broken modal trigger) is hidden.

This **does not** address the underlying data-pipeline issue — once the contributors blob is repopulated correctly, the real dates and avatars will surface. The defensive fix simply prevents misleading information from being shown when data is missing for any reason (stale cache, fetch failure, new page not yet picked up).

### Additional context

This bug is high-visibility — it appears at the top of major pages including "What is Ethereum?" — and undermines trust in the freshness of the content.

Reported by an external contributor with confirmation from production at 2026-04-29.

konopkja:
## Description

Closes #18070.

App Router pages currently render two visible defects when contributor data is missing:

1. **"Page last update: January 1, 1970"** — the Unix epoch.
2. **"See contributors" opens an empty modal.**

Both symptoms share a single trigger: `appPages[pagePath]` resolves to `[]` in the GitHub-contributors blob storage, and the consumer code does not handle the empty case defensively.

This PR adds the defensive handling so empty data hides cleanly instead of fabricating a misleading date and an empty modal trigger. It does **not** address why the underlying data is missing — that root cause is flagged in the linked issue for maintainer follow-up.

### Repro (before this PR)

- Visit https://ethereum.org/what-is-ethereum/ → see "Page last update: January 1, 1970".
- Click "See contributors" → empty modal.

### After this PR

When `contributors` is empty:
- The `Page last update` line is hidden (no fake date).
- The "See contributors" button is hidden (no empty modal).
- If both contributors AND timestamp are missing, the entire `FileContributors` block returns `null`.
- In JSON-LD, `dateModified` is omitted entirely instead of emitting an epoch string for SEO.

When data is available, the UI is unchanged.

### Core fix
- **`src/lib/utils/gh.ts`** — `getAppPageLastCommitDate` now returns `null` for an empty contributors array instead of `new Date(0).toString()`. The previous `.reduce(..., new Date(0))` silently produced the Unix epoch when there was nothing to reduce over, which then formatted as "January 1, 1970".
- **`src/lib/utils/contributors.ts`** — `getAppPageContributorInfo` propagates `null` instead of formatting `null` through `getLocaleTimestamp`.
- **`src/components/FileContributors.tsx`** — accepts `string | null` for `lastEditLocaleTimestamp`, returns `null` when both contributors and timestamp are missing, and conditionally hides the "See contributors" button + modal when contributors is empty.

### Type propagation
- **`src/lib/types.ts`** — `PageWithContributorsProps.lastEditLocaleTimestamp` accepts `string | null`.
- **`src/layouts/ContentLayout.tsx`** — `lastEditLocaleTimestamp` accepts `string | null`.
- **11 `page-jsonld.tsx` files** in `app/[locale]/...` — accept `string | null` and conditionally include `dateModified` only when truthy, so structured data does not advertise an epoch timestamp to search engines when the date is unknown.

## Out of scope (flagged in #18070)

The underlying data-pipeline question — *why* `appPages[pagePath]` is empty for these pages in the first place — is **not addressed** here. Verifying that requires inspecting the production Netlify Blobs that back `getGitHubContributors()` or running the scheduled fetcher (`src/data-layer/fetchers/fetchGitHubContributors.ts`) against production. I do not have access to either.

This PR makes the UI honest in the meantime: instead of showing fake data, it shows nothing. Once the data pipeline is fixed, the real dates and avatars will surface automatically — no further code changes needed.

## Verification

- `npx tsc --noEmit` — passes
- `npx eslint` on touched files — passes
- `pnpm test:unit` — 840 tests pass
- `npm run build` — passes locally

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## The change has been tested on:

- [x] Desktop
- [ ] Mobile
- [ ] Tablet

## Required

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes


nloureiro:
Backlog-sweep note: this UX fix has had merge conflicts since May. @konopkja worth a rebase so it can get reviewed — or close if it's been overtaken.

wackerow:
Posted a superceding PR as an up-to-date and cleaner approach to this solution that doesn't require touching so many files and data types:
- #18838

Closing this out in lieu of that one

PR #18066, opened 2026-04-29
docs(data-layer): add fetcher isolation rule

pettinarip:
Adds a new rule to the `data-layer` skill (Rule 5) and a short pointer in `src/data-layer/docs.md` (Rule 4):

- Fetchers must stay isolated from the Next.js app — no `process.cwd()` reads, no imports from `app/`, `src/components/`, or `src/lib/data/`.
- Allowed: types, pure constants, pure utility functions.
- If a fetcher needs app data, fetch it via the GitHub API. Don't bundle app files via `additionalFiles` in `trigger.config.ts`.
- Points at `fetchGitHubContributors.ts` as the canonical pattern.

PR #18072, opened 2026-04-29
fix(videos): fetch frontmatter via GitHub API

myelinated-wackerow:
## Summary

Rewrites `fetchVideoThumbnails` to read video frontmatter from GitHub instead of the local filesystem. Discovery happens via the trees API; each video's `index.md` is pulled from `raw.githubusercontent.com`. S3 upload chain (customThumbnailUrl > sddefault > hqdefault, slug-keyed filename) is unchanged.

## Motivation

The original implementation (#18039) read `public/content/videos/` from the filesystem, which works locally and in `pnpm trigger:dev` but fails on Trigger.dev production with `ENOENT: no such file or directory, scandir '/app/public/content/videos'`. Trigger.dev only ships the data-layer bundle, not the app's content tree.

This aligns with the fetcher isolation rule from #18066: data-layer fetchers are runtime-isolated from the Next.js app and should reach app data via the GitHub API, mirroring `fetchGitHubContributors`. Bundling app files into the trigger.dev deploy via `additionalFiles` would re-create the coupling we want to avoid.

## Changes

- `src/data-layer/fetchers/fetchVideoThumbnails.ts`
  - Removed `fs/promises`, `path`, `@/lib/constants` imports
  - Added `discoverVideoSlugs(token)` — single `git/trees/master?recursive=1` call, filters for `public/content/videos/<slug>/index.md`
  - Added `fetchFrontmatter(slug, token)` — pulls raw markdown via `raw.githubusercontent.com` and parses with gray-matter
  - Throws if `GITHUB_TOKEN_READ_ONLY` is missing
  - S3 upload logic unchanged

## Test plan

- [x] `pnpm trigger:dev` against the dev environment: task completed in ~2m46s, `Video thumbnail sync complete: 58/58 uploaded`
- [x] Spot-checked S3 URLs render in the browser (e.g. `videos/thumbnails/ethereum-basics-intro.jpg`)
- [ ] After merge: existing `deploy-data-layer.yml` workflow auto-deploys to Trigger.dev prod when this lands on master. Manually trigger `fetch-video-thumbnails` once afterward to populate the production blob, then redeploy Netlify so the gallery picks up the populated map.

## Related

- #18039 — original sync-thumbnails-to-S3 PR (introduced the filesystem-based fetcher this fixes)
- #18066 — adds the data-layer fetcher isolation rule this PR aligns with

---

### DIFF
--- a/src/data-layer/fetchers/fetchVideoThumbnails.ts
+++ b/src/data-layer/fetchers/fetchVideoThumbnails.ts
@@ -1,15 +1,24 @@
-import { readdir, readFile } from "fs/promises"
-import { join } from "path"
-
 import matter from "gray-matter"
 
 import type { VideoFrontmatter } from "@/lib/interfaces"
 
-import { CONTENT_DIR } from "@/lib/constants"
-
 import { uploadToS3 } from "../s3"
 
+import { fetchRetry } from "./fetchRetry"
+
+const GITHUB_API_BASE =
+  "https://api.github.com/repos/ethereum/ethereum-org-website"
+const RAW_BASE =
+  "https://raw.githubusercontent.com/ethereum/ethereum-org-website/master"
+const VIDEOS_PATH_PREFIX = "public/content/videos/"
+const VIDEO_INDEX_SUFFIX = "/index.md"
+
 const THUMBNAIL_PREFIX = "videos/thumbnails"
+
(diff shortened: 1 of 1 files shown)


pettinarip:
forgot to approve before merge lol ✅

PR #18095, opened 2026-05-01
fix(developers): add ISR revalidation to prevent stale hackathon events

### LINKED ISSUE #18086: bug(developers): past hackathon events still showing as upcoming — stale static build
## Bug Report

**Page:** https://ethereum.org/developers/
**Found via:** Automated weekly QA review (2026-04-27)

### Problem

The Hackathons section of the Developers page currently lists events that have already passed:

| Event | Dates | Status |
|-------|-------|--------|
| ETHGlobal Cannes | Apr 3–5, 2026 | ⛔ Past |
| ETHSilesia | Apr 16–19, 2026 | ⛔ Past |

### Root cause

Events are fetched from the Geode Labs Supabase API at **build time** via `src/data-layer/fetchers/fetchEvents.ts`, which correctly filters by `endTime >= now`. However, the static build has not been refreshed since those events were upcoming, so they remain in the deployed output.

### Options to fix

1. **Trigger a new deployment** (quickest one-time fix)
2. **Enable ISR** on the Developers page so events auto-refresh periodically without a full rebuild
3. **Move hackathon fetching to client-side** so it always reflects the live API

This issue will recur with every event cycle unless option 2 or 3 is implemented.

nloureiro:
Fixes #18086

## Problem

The /developers/ hackathons section shows events from a Geode Labs Supabase API. The filter logic in `fetchEvents.ts` correctly removes past events — but only at **build time**. Between deployments, old events linger in the static output, as happened with ETHGlobal Cannes (Apr 3–5) and ETHSilesia (Apr 16–19) appearing as upcoming after they had passed.

## Fix

Add Next.js ISR (`export const revalidate = 86400`) to the developers page so it regenerates at most every 24 hours automatically, keeping the events list current without waiting for a full deployment.

## Changes

- `app/[locale]/developers/page.tsx` — add `export const revalidate = 86400`

## Why 24 hours

Events are typically announced days/weeks in advance so a 24 h window won't miss anything meaningful, while preventing the multi-week staleness seen now. This can be tuned down if needed.


myelinated-wackerow:
Thanks for digging into this, @nloureiro -- the diagnosis is spot on. The page renders statically, so every past-event filter (`fetchEvents.ts`, plus both checks in `getHackathons`) freezes `new Date()` at build time and can't re-filter until the next deploy. ISR is the right instinct.

The concern with `export const revalidate = 86400` at the page level is that it's **route-wide**, not scoped to the events section -- and this route reads `public/content` at render time:

```
page.tsx:151  const recentPosts = (await getBlogPostsData(locale)).slice(0, 3)
   -> src/lib/utils/md.ts:104  join(process.cwd(), "public/content") -> fsp.readFile(...)
```

`public/content` is CDN-served and **not bundled into the Netlify serverless function**. It reads fine at build, but on the ISR re-render `getBlogPostsData` can't find the files -- each read is caught and returns `null`, so the "recent blog posts" section silently renders empty ~24h after each deploy. (No 404 here, since the page shell is TSX rather than MDX -- just a blank section.)

This is a documented, already-bitten failure mode: `docs/solutions/integration-issues/netlify-isr-404-async-server-components.md`, and the codebase already guards against it -- see `getStaticDeveloperToolsData` in `src/lib/data/index.ts` (`revalidate: false`, with a comment explaining exactly this).

Worth noting the events data itself would survive ISR fine -- it's read from Netlify Blobs via the data-layer (`getEventsData`), which *is* available at runtime. It's only the `public/content` markdown read that breaks. Tying revalidation to an on-demand `revalidatePath` wouldn't help either, since any runtime re-render of the route hits the same wall.

Suggested alternative that keeps the page fully static and sidesteps the landmine: render the hackathons list as a small client component seeded with the build-time list (HTML still has content + SEO), then re-fetch from a tiny `/api/hackathons` route and re-filter with a live `now` on mount -- stale entries self-heal client-side without making the whole page ISR. Pushing that to this branch shortly for review.

-- Opus 4.8/@wackerow

myelinated-wackerow:
Pushed the client-filter approach to this branch. The hackathons list is now a small client component (`HackathonsList`) seeded with the build-time list (SSR keeps content + SEO); it re-filters with a live clock on mount and pulls the freshest list from a new `/api/hackathons` route (reads the data-layer / Blobs, available at runtime).

One correction to my framing above, now that I've checked the build output: `/developers` is *already* an ISR route (`revalidate: 1d`) -- not from a page-level export, but inherited from `getEventsData` (`unstable_cache`, 24h) via `getHackathons()`. So the original `export const revalidate = 86400` was redundant with the value the route already had. The reason events still go stale is that the daily revalidation isn't refreshing the served page on Netlify (the serverless re-render can't read `public/content` -- the failure mode in `netlify-isr-404-async-server-components.md` -- so the last successful build keeps being served). The client-side filter sidesteps that entirely: events self-heal in the browser regardless of whether/when server ISR fires.

Two related notes for review:

1. Making the page *fully* static (to close the latent `public/content`-on-ISR risk for the recent-posts section) would take a `revalidate: false` events getter + `force-static`, mirroring `getStaticDeveloperToolsData`. Left out here to keep this PR scoped to the events bug.

2. The same build-time-freeze pattern lives elsewhere and is worth a separate follow-up:
   - `/community/events` -- closest sibling, same event data; the static page bakes the past/upcoming split at deploy, so a passed conference lingers under "Upcoming" until the next deploy.
   - `/staking` battle-tested flags (`getBattleTestedFlag`) -- rolling 6mo/1yr window computed at build; drifts until rebuild. Low severity.
   - `parseAppsOfTheWeek` -- latent (no current caller), same freeze if wired onto a static page.

The clean systemic fix is on-demand revalidation from the daily Trigger.dev tasks (ping the existing `/api/revalidate` after each Blobs write), but that's only safe once markdown reads are decoupled from runtime. Capturing it here; happy to open a tracking issue.

cc: @pettinarip for review

-- Opus 4.8/@wackerow

PR #18119, opened 2026-05-04
refactor(skills): split performance into references

pettinarip:
## Summary

- Split the monolithic `performance` skill into a knowledge skill with chunked references following the [Agent Skills](https://agentskills.io) progressive-disclosure pattern.
- Top-level `SKILL.md` is a thin index; rule content lives in `references/` (one file per topic: `diagnose-table`, `edge-caching`, `bundle`, `build`, `rsc`, `inp`, `images`, `data-fetching`, `anti-patterns`, `shipped`, `external-overrides`).
- Marked `user-invocable: false` so the model can auto-load the skill on perf topics but it stays out of the `/` autocomplete menu.
- All knowledge consolidated inside the skill — no pointers to repo docs or memory files.

## Test plan

- [x] \`npx skills-ref validate .claude/skills/performance\` passes (only \`user-invocable\` flagged — documented Claude Code extension)
- [x] Claude Code skill listing includes \`performance\` description
- [x] \`/performance\` does not appear in Claude Code's slash autocomplete
- [x] Asking a perf question (e.g., "what's our rule on edge caching?") triggers the model to load \`references/edge-caching.md\`
- [x] \`grep -r 'docs/' .claude/skills/performance/\` returns nothing — knowledge is self-contained

### DIFF
--- a/.claude/skills/performance/SKILL.md
+++ b/.claude/skills/performance/SKILL.md
@@ -1,0 +1,39 @@
+---
+name: performance
+description: Performance patterns and anti-patterns for ethereum.org covering TTFB, LCP, INP, CLS, bundle size, build memory, and RSC payload. Use when diagnosing a perf regression or reviewing proposed code changes for perf landmines.
+user-invocable: false
+---
+
+# Performance Knowledge
+
+Knowledge base. No runtime actions; authoring conventions for maintainers below. Read the relevant `references/*.md` files directly via the Read tool.
+
+## Two entry points
+
+- **Diagnosing** an existing regression → `references/diagnose-table.md` (symptom → bucket).
+- **Reviewing** a proposed change before merge → `references/review-checklist.md` (diff pattern → rules to load).
+
+Both flows always also load `references/anti-patterns.md`.
+
+## File map
+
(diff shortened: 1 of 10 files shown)


.claude/skills/performance/SKILL.md
```diff
@@ -0,0 +1,39 @@
+---
+name: performance
+description: Performance patterns and anti-patterns for ethereum.org covering TTFB, LCP, INP, CLS, bundle size, build memory, and RSC payload. Use when diagnosing a perf regression or reviewing proposed code changes for perf landmines.
+user-invocable: false
```
wackerow, line 4: Image

Noting this _does_ show up when you type the full `/performance` but anything less and it does _not_ show the auto-complete -- should be find

wackerow (approved):
```bash
$ npx skills-ref validate .claude/skills/performance
Validation failed for .claude/skills/performance:
  - Unexpected fields in frontmatter: user-invocable. Only allowed-tools, compatibility, description, license, metadata, name are allowed.
```

Per test note in PR body, this looks correct -- only `user-invocable` being flagged.

Lgtm, thanks @pettinarip

Issue #18121, opened 2026-05-05
feat(design-system): Add Storybook stories for canonical UI primitives


myelinated-wackerow:
## Why

The `.claude/skills/design-system/` skill (PR #18120) describes our canonical UI primitives in text. Storybook stories give them a *visual* reference -- so designers can see what's canonical, agents can be pointed at Storybook URLs in PR review, and new variants get visual review before they land.

This issue tracks adding stories for the `ui/` primitives that don't have them today. Items are grouped into five scoped PRs to keep each review tractable.

## For each PR

1. Look at existing high-quality stories as exemplars: `Button.stories.tsx`, `Input.stories.tsx`, `Tag.stories.tsx`, `Table.stories.tsx`.
2. Pick ONE of the five PR groups below. Open a PR scoped to just that group.
3. Reference this issue in the PR description (don't close the issue).

## Standards for new stories

The north star: **a designer browsing Storybook should be able to see every variant that exists** for a given primitive. Not exhaustive combinations -- just enough that nothing is hidden.

Each story file should include:

- **One `Default` story** -- canonical, simplest valid usage
- **One story per variant prop** -- showing all the prop's options together (e.g., a single `Sizes` story rendering each size side-by-side). For compositional primitives like `Card`, this includes sub-component variant props (e.g., `CardBanner` `background`, `CardTitle` weight)
- **One story per meaningfully-different sub-component** when the sub-components have distinct shapes (e.g., `Skeleton` vs `SkeletonLines` vs `SkeletonCardGrid`)
- **2-3 realistic composite stories** -- full configurations that model real-world use (e.g., `HeroCard` showing Banner + Title + Paragraph + Footer all together)

What to **avoid**:

- Exhaustive cross-product combinations (Card with every banner background x every title weight x every layout = bloat)
- Single "kitchen sink" stories that lump everything together -- prefer multiple focused stories

What to **include in every story**:

- **Title pattern**:
  - `UI / Primitives / [Name]` for shadcn-regenerable primitives (Button, Card, Sheet, Dialog, Input, etc.)
  - `UI / [Name]` for custom project primitives (Tag, Section, TabNav, EdgeScrollContainer, etc.)
- **a11y addon passes** -- Storybook will flag obvious violations

Light/dark mode and locale toggling are available via Storybook's built-in toolbar -- no per-story config required for those.

### Strings and translations

Most `ui/` primitives accept strings as `children` or props. Pass static English in stories -- no intl wiring needed.

[...]

PR #18123, opened 2026-05-05
feat(storybook): expand form input stories

myelinated-wackerow:
## Summary

Adds and expands Storybook stories for the form input primitives in `src/components/ui/`. This is PR 2 of 5 for #18121, scoped to form inputs. One commit per checkbox in the issue.

- **Textarea** (new) -- Default, Sizes, HasError, WithoutPlaceholder, WithHelperText, Disabled. Title `UI / Primitives / Textarea`.
- **Input** (expanded) -- adds HasError, Disabled (with and without value), AutoFocus, WithHelperText. Retitled to `UI / Primitives / Input`.
- **Checkbox** (expanded) -- split kitchen-sink story into focused stories per state, adds Indeterminate. Retitled to `UI / Primitives / Checkbox`.
- **Switch** (expanded) -- split into Off, On, Disabled, Invalid, WithLabel, and a SettingsList composition. Retitled to `UI / Primitives / Switch`.

The four stories live under `UI / Primitives / [Name]` to match the hierarchy established in #18122 (PR 1). Title alignment is intentionally consistent within this cleanup; the team can decide on a final hierarchy in a future pass.

## Removed

- **RadioGroup** -- expansion was added (8cccb5b695) and then removed (2021192a8a), along with the original story file that lived on dev. Component is being removed in a separate cleanup; the story would be orphaned, so it shouldn't ship.

## Notes for review

- Textarea and Input share the same `cva` variant shape (`size`, `hasError`); the two stories are deliberately structurally similar so designers see the parity at a glance.
- `Checkbox` indeterminate state addresses the gap called out in #18121.
- `Switch` `WithLabel` story uses nested-label association (no `htmlFor` needed); `SettingsList` models a realistic settings-panel use.

## Test plan

- [ ] Eyeball each new/expanded story in Storybook (light + dark, locale toggle)
- [ ] a11y addon passes for each story

Refs: #18121


pettinarip (approved):
LGTM!

PR #18136, opened 2026-05-06
fix(a11y): use aria-labels where appropriate over sr-only

wackerow:
## Description
- use aria-labels where appropriate over sr-only
- Keep aria-describedby where appropriate, adding missing aria-label
- intl: rem hard-coded english strings, replace with intl strings


src/components/ui/alert.tsx
```diff
@@ -81,17 +82,20 @@ AlertDescription.displayName = "AlertDescription"
...
-    variant="ghost"
-    className={cn("-me-4 rounded-full text-body", className)}
-    {...props}
-  >
-    <X className="h-6 w-6" />
-    <span className="sr-only">Close</span>
-  </Button>
-))
+>(async ({ className, ...props }, ref) => {
+  const t = await getTranslations("common")
```
pettinarip: not convinced about this. I don't think we should handle translations in these pure/dumb UI components. Maybe a convention to set.

also, getTranslations make this components server-only components. I think UI components should be env agnostic.

wackerow: @pettinarip Okay, thinking on how to handle the aria-label here then, since we wouldn't want to have to pass that every time we use it. Perhaps this specific component (`AlertCloseButton`) gets broken out as it's own file?

pettinarip: @wackerow splitting it into its own file won't really help here. I'd rather we settle on one convention, since we're already doing it two ways in this PR (`getTranslations` here vs `useTranslation` in `dialog-modal.tsx`):

my preference: keep the primitive agnostic. `aria-label` as a prop with an english fallback, caller translates when it has `t`:
```tsx
({ "aria-label": ariaLabel = "Close", ...props }, ref) => (
  <Button ref={ref} aria-label={ariaLabel} ...>
```

Otherwsie, if we must keep this translated, self-translate with the client hook (`useTranslation`, like `dialog-modal`). No caller involvement, but locks it to client-only.

wackerow: Okay -- yeah I agree it's not worth pigeon-holing this into either client OR server-only. I'd rather not hardcode the English string though since that's prime to be forgotten. Making `aria-label` a required prop would force the consumer to provide a properly translated label here, while keeping the ui/alert components dumb. 

There are a couple other out-of-scope things we should address in this component as well, but I'll leave those for another branch.

pettinarip (approved):

PR #18173, opened 2026-05-12
Bounty form update

0xFredrik:
## Description

Replaced previous google bbp form with locally encrypted form with built in fee burn to reduce spam


wackerow (approved):
Thanks @fredrik0x!

PR #18180, opened 2026-05-13
fix(qa): filter past hackathons and add site name to page titles

nloureiro:
## Summary
Fixes three issues found in the weekly automated QA review (2026-05-11).

### Changes
- **Past hackathons on /developers/**: `getHackathons()` now applies a runtime date guard so events past their end date are never shown, even if the cached event data is stale.
- **Missing \`| ethereum.org\` in page titles**: Added consistent site-name suffix to the meta titles for \`/developers/\`, \`/get-eth/\`, and \`/layer-2/\`, matching the pattern already used on the homepage and roadmap pages.

### Not included
The "roadmap for 2025" heading on \`/what-is-ethereum/\` is already fixed on \`dev\` (says 2026) — it just hasn't shipped to production yet.

### DIFF
--- a/src/intl/en/page-developers-index.json
+++ b/src/intl/en/page-developers-index.json
@@ -1,5 +1,5 @@
 {
-  "page-developer-meta-title": "Ethereum Developer Resources",
+  "page-developer-meta-title": "Ethereum Developer Resources | ethereum.org",
   "page-developers-account-desc": "Contracts or people on the network",
   "page-developers-accounts-link": "Accounts",
   "page-developers-api-desc": "Using libraries to interact with smart contracts",
--- a/app/[locale]/developers/utils.tsx
+++ b/app/[locale]/developers/utils.tsx
@@ -103,9 +103,14 @@ export const getVideoCourses = async (): Promise<VideoCourse[]> => {
 export const getHackathons = async (): Promise<EventItem[]> => {
   const events = await getEventsData()
   if (!events) return []
-  return events.filter(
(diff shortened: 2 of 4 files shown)


src/intl/en/page-developers-index.json
```diff
@@ -1,5 +1,5 @@
 {
-  "page-developer-meta-title": "Ethereum Developer Resources",
+  "page-developer-meta-title": "Ethereum Developer Resources | ethereum.org",
```
wackerow, line 2: @nloureiro I see we're using this pattern inconsistently -- I think a cleaner solution would be to bake this into the src/lib/utils/metadata.ts `getMetadata` helper function. We have `siteTitle` in there which displays `ethereum.org`, which we could automatically append to the end of the titles on all pages.

- Avoids repeating ourselves
- Avoids unnecessary intl complications
- Avoids inconsistency for any new pages

The homepage uses "Ethereum - The complete guide from Ethereum.org" which 1- we should lowercase "ethereum.org", and 2- we can simply opt that page out of automatically appending ` | ethereum.org` to the end since it'd be redundant.

If you're okay with that, perhaps we revert these from this PR, merge in the hackathon changes, and we can set up an issue to fix up the og:title metadata in a separate PR.

wackerow: Issue #18185 posted with refactor summary -- No harm in just pulling these in and a follow-up task can address more completely.

wackerow (approved):

PR #18194, opened 2026-05-14
codeblock: replace prism-react-renderer with shiki

pettinarip:
## Summary

Replaces `prism-react-renderer` with [`shiki`](https://shiki.style/) for code block syntax highlighting. The new `Codeblock` is an async server component; the client wrapper only owns the collapse + copy controls.

## Why

- **Correctness across the languages we actually use.** Our content has 13k+ Solidity blocks, 4.9k Python, 2.1k bash/shell, 1k Rust, plus JS/TS/TSX/JSX. `prism-react-renderer` (and JS-only alternatives like `sugar-high`) mis-tokenise everything outside the JS family on the tutorials and docs pages — exactly where highlighting matters most. Shiki ships 692 grammars and handles all of them.
- **Zero client JS.** Highlighting runs server-side; the resulting HTML string is handed to a tiny `"use client"` wrapper via `dangerouslySetInnerHTML`. Grammars/themes never cross the RSC boundary, so the client bundle does not grow.
- **Build-time cost, not request-time.** Cost is ~10–16 ms per block on cold call, amortised across statically-generated tutorial/docs routes — well under the noise floor next to MDX compilation.
- **Theme switching with no re-highlight.** Shiki emits inline `--shiki-light` / `--shiki-dark` CSS variables; the dark theme is selected purely via the `.dark` ancestor.

## Changes

- New `src/components/Codeblock/` (server `Codeblock.tsx` + client `CodeblockClient.tsx`) replaces the old single-file `src/components/Codeblock.tsx`.
- `src/lib/shiki.ts` — singleton highlighter with a fixed lang/theme allowlist (one-light + one-dark-pro) to avoid lazy-import latency on cold starts.
- `src/styles/codeblock.css` — token styles + dark-mode variable swap, imported once from `global.css`.

## Test plan

- [x] Tutorials render correctly in light + dark mode (`/developers/tutorials/uniswap-v2-annotated-code` — 149 blocks, good stress test)
- [x] Solidity, Python, bash, Rust blocks tokenise (not just JS)
- [x] Copy button + collapse still work
- [ ] No `shiki` / `codeToHtml` / `@shikijs` symbols leak into `.next/static/chunks/*.js`


src/components/Codeblock/Codeblock.tsx
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  const matches = langClass?.match(/language-(.*)/)
+  const language = matches?.[1] || ""
+  const resolvedLang = resolveLang(language)
+
+  const html = await highlight(codeText, language)
+
+  const shouldShowCopyWidget =
+    resolvedLang !== "text" && COPY_WIDGET_LANGS.has(resolvedLang)
```
wackerow: Looks like this is parity with existing, but wondering if we should just make the copy button available for all of these -- especially with this design, it doesn't really get in the way much.

src/components/Codeblock/index.tsx
wackerow, line 1: Curious if there is a reason for not placing this directly inside `index.tsx` (vs the `index.ts` exporter)?

wackerow (approved):
Couple non-blocking comments, but this looks great to me @pettinarip, thanks! This paired with #18195 will be a nice polish to all the codeblocks.

src/components/Codeblock/Codeblock.tsx
pettinarip: Yeah...probably not worth this logic. I'll remove it and keep everything consistent.

src/components/Codeblock/index.tsx
pettinarip: Not a convention we are following. I'll move it to the index

src/components/Codeblock/Codeblock.tsx
pettinarip: > wondering if we should just make the copy button available for all of these

Addressed: removed the `COPY_WIDGET_LANGS` allowlist; the copy button now shows on every non-homepage codeblock.

src/components/Codeblock/index.tsx
pettinarip: > Curious if there is a reason for not placing this directly inside `index.tsx` (vs the `index.ts` exporter)?

Addressed: moved the component into `index.tsx` and removed the `index.ts` re-exporter.

PR #18196, opened 2026-05-14
refactor: centralize site title suffix in metadata

### LINKED ISSUE #18185: Centralize site name suffix in page titles via getMetadata
## Context

The `| ethereum.org` suffix on `<title>` and `og:title` is currently inconsistent across pages: ~9 EN intl keys bake it into the translatable string, while ~30+ pages omit it entirely. PR #18180 added it to three QA-flagged pages by editing intl JSON, which is the wrong layer -- "ethereum.org" is a brand/domain constant, not translatable content. Treating it as a translated string forces 25 locales to carry the same token and creates drift.

`src/lib/utils/metadata.ts` already uses `siteTitle = t("site-title")` for OG `siteName` and Twitter `site`/`creator`. The page `title` field is the only place it isn't applied.

### 1. Promote `site-title` from intl to a typed constant

`"ethereum.org"` is the same string in every language; it does not belong in `common.json`. Move it to a constant (e.g. `SITE_TITLE` in `src/lib/constants.ts`) and replace `t("site-title")` callers (`getMetadata` + any others -- grep first to be sure).

```ts
// src/lib/constants.ts
export const SITE_TITLE = "ethereum.org"
```

### 2. Append the brand suffix centrally in `getMetadata`, wrapped in bidi isolates

In `src/lib/utils/metadata.ts`, replace the bare `title` pass-through with logic that appends ` | ethereum.org` (wrapped in U+2066 LRI / U+2069 PDI) to all three places it's used (`title`, `openGraph.title`, `twitter.title`). The isolates ensure the brand is always treated as a single LTR run regardless of surrounding locale -- zero-width and no-op in LTR contexts, correct rendering in RTL.

Skip appending when the title already mentions the brand inline (the homepage's `"Ethereum - The complete guide from ethereum.org"` -- and any future brand-y title -- naturally falls through):

```ts
const titleAlreadyHasBrand = title
  .toLowerCase()
  .includes(SITE_TITLE.toLowerCase())
const finalTitle = titleAlreadyHasBrand
  ? title
  : `${title} | ⁦${SITE_TITLE}⁩`
```

The single `includes` check on the bare `SITE_TITLE` constant covers two cases at once:

- **Inline brand mentions** (homepage): `"... from ethereum.org"` already contains the brand -- no append.
- **Stale suffix in intl** (cleanup drift): `"... | ethereum.org"` already contains the brand -- no double-suffix.

No opt-out prop, no locale conditional, no hardcoded path check.

### 3. Strip the suffix from EN intl strings (one-time cleanup)

Remove ` | ethereum.org` from the ~9 EN keys that currently bake it in -- run `grep -rE '"[a-z-]+meta-title".*\| ethereum.org' src/intl/en/` for the authoritative list. Let the intl pipeline propagate to non-English locales; do not hand-edit translated files (per CLAUDE.md).

Safe to land in either order relative to #2 -- the `includes` check makes both states valid.

### 4. Lowercase the homepage's brand reference

Update `page-index-meta-title` in `src/intl/en/page-index.json` from `"Ethereum - The complete guide from Ethereum.org"` to `"Ethereum - The complete guide from ethereum.org"` to match the brand constant casing.

## Acceptance criteria

- [ ] `SITE_TITLE` exported from a constants file; `site-title` removed from `common.json` and all 24 translated `common.json` files (via pipeline, not hand-editing)
- [ ] Every page rendered via `getMetadata` outputs `<title>` and `og:title` ending in ` | ethereum.org` (visually), except the homepage
- [ ] Homepage `<title>` reads `Ethereum - The complete guide from ethereum.org` (no suffix, lowercase brand) with no special-casing in `getMetadata`
- [ ] No EN intl key contains ` | ethereum.org` literally (`grep` returns zero matches in `src/intl/en/`)
- [ ] No double-suffix anywhere in deploy preview
- [ ] Visual check in `ar` and `ur` deploy previews: brand renders left-to-right inside the RTL title bar
- [ ] Quick scan for any in-house code (analytics, logs, snapshot tests) that does literal substring matching on page titles -- the U+2066/U+2069 wrappers will break naive comparisons against `"ethereum.org"`. Fix or note any hits.

## Out of scope

- Renaming or recasing the brand elsewhere (homepage hero, footer, etc.)
- Page-level `<title>` overrides outside `getMetadata` (none expected; confirm via grep)
- Replacing LRI/PDI with HTML-level `<bdi>` -- the `<title>` element doesn't accept inline markup, so Unicode isolates are the correct tool here

## Related

- PR #18180 (QA-driven fix this generalizes)
- `src/lib/utils/metadata.ts`
- `src/intl/en/common.json` -- `site-title` key to be removed

---

_Generated by Claude Opus 4.7_

kafeelraza:
## Description

This PR centralizes the `ethereum.org` site title suffix handling in `getMetadata` instead of hardcoding it in translation strings.

### Changes

* Added `SITE_TITLE` constant in `src/lib/constants.ts`
* Replaced `t("site-title")` usages with `SITE_TITLE`
* Centralized title suffix appending in `metadata.ts`
* Added bidi isolate wrapping for RTL-safe rendering
* Removed hardcoded `| ethereum.org` suffixes from EN intl files
* Updated homepage metadata title casing to `ethereum.org`

### Notes

* Homepage avoids duplicate suffixes automatically because the brand is already present in the title.
* Build failure during prerendering appears unrelated to this PR (`ENOENT` missing content file).

## Related issue
- Closes #18185

### DIFF
--- a/src/lib/utils/metadata.ts
+++ b/src/lib/utils/metadata.ts
@@ -4,6 +4,7 @@ import { getTranslations } from "next-intl/server"
 import {
   DEFAULT_OG_IMAGE,
   IS_PRODUCTION_DEPLOY,
+  SITE_TITLE,
   SITE_URL,
 } from "@/lib/constants"
 
@@ -65,7 +66,12 @@ export const getMetadata = async ({
   const t = await getTranslations("common")
 
   const description = descriptionProp || t("site-description")
-  const siteTitle = t("site-title")
+
+  const titleAlreadyHasBrand = title
+    .toLowerCase()
+    .includes(SITE_TITLE.toLowerCase())
+
+  const finalTitle = titleAlreadyHasBrand ? title : `${title} | ⁦${SITE_TITLE}⁩`
 
   // Auto-detect translated locales if not provided
   const finalTranslatedLocales =
@@ -93,7 +99,7 @@ export const getMetadata = async ({
     : []
 
   const base: Metadata = {
-    title,
+    title: finalTitle,
     description,
     formatDetection: { telephone: false },
     metadataBase: new URL(SITE_URL),
@@ -109,12 +115,12 @@ export const getMetadata = async ({
       }),
     },
     openGraph: {
-      title,
+      title: finalTitle,
       description,
       locale,
       type: "website",
       url,
-      siteName: siteTitle,
+      siteName: SITE_TITLE,
       images: [
         {
           url: ogImage,
@@ -122,11 +128,11 @@ export const getMetadata = async ({
       ],
     },
     twitter: {
-      title,
+      title: finalTitle,
       description: twitterDescription || description,
       card: "summary_large_image",
-      creator: author || siteTitle,
-      site: author || siteTitle,
+      creator: author || SITE_TITLE,
+      site: author || SITE_TITLE,
       images: [
         {
           url: ogImage,
--- a/src/lib/utils/metadata.ts
+++ b/src/lib/utils/metadata.ts
@@ -71,7 +71,9 @@ export const getMetadata = async ({
     .toLowerCase()
     .includes(SITE_TITLE.toLowerCase())
 
-  const finalTitle = titleAlreadyHasBrand ? title : `${title} | ⁦${SITE_TITLE}⁩`
+  const finalTitle = titleAlreadyHasBrand
+    ? title
+    : `${title} | \u2066${SITE_TITLE}\u2069`
 
   // Auto-detect translated locales if not provided
   const finalTranslatedLocales =
--- a/src/components/Nav/MobileMenu/MenuHeader.tsx
+++ b/src/components/Nav/MobileMenu/MenuHeader.tsx
@@ -1,5 +1,7 @@
 import { SheetClose, SheetTitle } from "@/components/ui/sheet"
 
+import { SITE_TITLE } from "@/lib/constants"
+
 import { useTranslation } from "@/hooks/useTranslation"
 
 const MenuHeader = () => {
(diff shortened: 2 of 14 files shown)


src/lib/utils/metadata.ts
```diff
@@ -65,7 +66,12 @@ export const getMetadata = async ({
   const t = await getTranslations("common")
 
   const description = descriptionProp || t("site-description")
-  const siteTitle = t("site-title")
+
+  const titleAlreadyHasBrand = title
+    .toLowerCase()
+    .includes(SITE_TITLE.toLowerCase())
+
+  const finalTitle = titleAlreadyHasBrand ? title : `${title} | ⁦${SITE_TITLE}⁩`
```
wackerow: For the BiDi handling here, would consider the escape-sequence form so the intent is visible in source:

```suggestion
  const finalTitle = titleAlreadyHasBrand ? title : `${title} | \u2066${SITE_TITLE}\u2069`
```

Output should be identical

wackerow (changes requested):
Thanks @kafeelraza! Looks great! Nice and clean

Left one inline code suggestion, but also, `src/intl/en/page-layer-2-learn.json` line 2 also has the ` | ethereum.org` appendage... would you mind patching that one as well? Then I think this is good to go

src/lib/utils/metadata.ts
Blondie0123: ⬇️ Suggested change
-  const finalTitle = titleAlreadyHasBrand ? title : `${title} | ⁦${SITE_TITLE}⁩`
+  const finalTitle = titleAlreadyHasBrand ? title : `${title} | \u2066${SITE_TITLE}\u2069`

kafeelraza: Patched the remaining hardcoded suffix in `page-layer-2-learn.json` and updated the BiDi isolates to the escaped unicode form (`\u2066` / `\u2069`). Thanks for the review!

wackerow (approved):
Looks good! Thanks for the help @kafeelraza!

wackerow:
@all-contributors please add @kafeelraza for code

kafeelraza:
Thanks @wackerow and everyone for the review and guidance. Happy to contribute to ethereum.org

PR #18197, opened 2026-05-15
chore(translatathon): deprecate pages, archive leaderboard

myelinated-wackerow:
## Summary

- Deprecate the three `template: translatathon` markdown pages (`/contributing/translation-program/translatathon`, `/details`, `/terms-and-conditions`) by deleting them across the English source and all 24 translated locales. Permanent redirects from each path send users to `/contributing/`.
- Drop the dedicated `src/layouts/md/Translatathon.tsx` layout, its registration entries in `src/layouts/{index.ts,md/index.ts}`, the `page-translatathon` i18n namespace and path mapping, and the orphaned content-placeholder JSONs under `src/data/placeholders/`.
- Reshape the `/translatathon/leaderboard` route as a historical archive: drop the Translatathon dropdown menu, left-side TOC, hero apply button, and time-sensitive intro copy; trim the JSON-LD breadcrumb to remove the now-deleted hub crumb; convert the `Leaderboard` component to a server component and remove the show-more pagination so all 217 rows render up front. With nothing else importing them, the entire `src/components/Translatathon/` directory is removed.
- Tidy `public/images/translatathon/`: move the still-useful `walking.png`, `round-table.png`, and `settlement.png` up to `public/images/` (and fix the `walking.png` import in `app/[locale]/what-is-ethereum/page.tsx`); delete the duplicate `man-baby-woman.png` and the now-orphan `translatathon_dolphin.png`; remove the directory.

## Note on inbound links

The leaderboard archive is intentionally left at its original URL so external links continue to resolve; it is no longer linked from anywhere in the site. A repo-wide grep turned up zero internal hrefs to any of the deprecated paths (the only mention was a non-clickable line in `docs/gemini-translation-roadmap.md`).

## Test plan

- [ ] Visit `/contributing/translation-program/translatathon`, `/details`, and `/terms-and-conditions`, plus a locale-prefixed variant (e.g. `/fr/contributing/translation-program/translatathon`); confirm each 301s to `/contributing/`.
- [ ] Visit `/contributing/translation-program/translatathon/leaderboard` and confirm it renders the archive intro copy, no Translatathon dropdown menu, no apply CTA, and the full list of rows up front.
- [ ] Load `/what-is-ethereum` and confirm the "when and who" banner image still renders after the import path move.

🤖 Generated with Claude Code

PR #17994, opened 2026-04-19
AI content updates

mnelsonBT:
- Adds the AI agents builder hub (/ai-agents/) with a landing page and nine sub-pages
- Replaces the AI agents use case page with a Decentralized AI use case page
- Updates site navigation per #17615 
- Video embeds: Integrated five educational video resources via <VideoWatch> across getting-started/, ethereum/, identity/, and decentralized-ai/

### Builder pages added 
- /ai-agents/ — AI agents builder hub landing page
- /ai-agents/ethereum/ — Why Ethereum for AI agents
- /ai-agents/getting-started/ — Four-step developer on-ramp
- /ai-agents/frameworks/ — Framework directory and decision guide
- /ai-agents/wallets/ — Smart accounts, session keys, key management
- /ai-agents/verification/ — zkML, TEEs, onchain audit trails
- /ai-agents/payments/ — Machine payments with x402
- /ai-agents/identity/ — Agent identity with ERC-8004
- /ai-agents/use-cases/ — What agents are doing on Ethereum today
- /ai-agents/l2s/ — Choosing an L2 for agent workloads

### Layout and components 
- Custom AiAgentsLayout template (src/layouts/md/AiAgents.tsx) with dropdown sub-navigation, ContentHero, and ToC sidebar
- HubCard component built from Card/CardBanner/CardContent/CardTitle/CardParagraph/ButtonLink primitives for the landing page capability grid
- pre → Codeblock override so fenced code blocks render with syntax highlighting, line numbers, copy button, and language label (matching /developers/docs/* behavior)

### Navigation changes: 
- Updated buildNavigation.ts to add AI agents under the Build menu
- Replaced the AI agents callout on the Use Cases hub with a Decentralized AI callout
- Added all required i18n keys to common.json

### Deviations from original Issue scope: 
- Added a BREADCRUMB_LABEL_OVERRIDES map in page-jsonld.tsx so the slug-derived breadcrumb label renders as "AI agents" instead of "Ai agents" in the BreadcrumbList JSON-LD, fixing incorrect casing visible in Google's breadcrumb rich results, per Jakub feedback
- Distributed the centralized FAQ page (from #17627) into per-page ExpandableCard FAQ sections. Each page carries its own contextual Q&A rather than linking to a single FAQ index
<ExpandableCard> body text to match faqItems schema answers, eliminating "on this page" references, and removing substantive content (contract addresses, SDK lists) that only appeared in one location
[...]

PR #18198, opened 2026-05-15
feat(layouts): consolidate topic page layouts

myelinated-wackerow:
## Summary

Folds Staking, UseCases, Roadmap, and Upgrade markdown layouts into a single async server `TopicLayout` driven by per-topic config in `src/data/topics/<key>.ts`. Each section becomes a small config file (translation namespace, dropdown items, hero dimensions, edit banner) instead of a near-duplicate layout component. The four old layout files become MDX component bundles only and move to `src/components/MdComponents/topics/`.

Companion changes in the same PR:

- Migrates 375 markdown files (15 EN + 360 translations across 24 locales) from `summaryPoint1..4` to a single `summaryPoints` array; drops the dead `emoji` frontmatter field. Safe to land as a flat find-replace because none of the touched paths are ingested by the intl-pipeline yet (verified via `.manifests/public/content/` check).
- Hoists the `Codeblock`-wrapped `Pre` handler -- previously duplicated inline in Docs and Tutorial layouts -- into `MdComponents.htmlElements` as the site-wide default for fenced code blocks.

## Behavior preservation

Dropdown labels, Matomo event names/categories, hero image dimensions, the edit banner (UseCases topic), and Staking's community callout are preserved verbatim. The community callout is passed as an `afterContent` slot prop by the slug router for `template: staking` pages.

## Related

- Translatathon deprecation handled separately in #18197
- AiAgents PR #17994 rebases onto this
- `TopicLayout` Storybook story deferred -- server-component `getTranslations` needs a request context that the project's `storybook-next-intl` setup doesn't provide; worth its own discussion

## Test plan

- [ ] `/staking/`, `/staking/solo/`, `/staking/saas/`, `/staking/pools/`, `/staking/withdrawals/` render with the staking dropdown, hero, and the community callout below content
- [ ] All `/use-cases/` subpages render with the use-cases dropdown, hero, and the "edit this page" banner
- [ ] `/roadmap/security/`, `/roadmap/scaling/`, `/roadmap/user-experience/`, `/roadmap/future-proofing/` render with the roadmap dropdown
- [ ] `/roadmap/beacon-chain/` and `/roadmap/merge/` (`template: upgrade`) render with the upgrade dropdown and show "page last updated: ..." under the summary list
- [ ] Spot-check one non-English locale (e.g. `/es/staking/solo/`) -- dropdown labels translate
- [ ] Spot-check one RTL locale (e.g. `/ar/staking/solo/`) -- layout is correctly mirrored
- [ ] Pages with fenced code blocks (e.g. `/developers/docs/getting-started/`) still render with syntax highlighting and the copy-to-clipboard widget

🤖 Generated with Claude Code


src/layouts/ContentLayout.tsx
```diff
@@ -27,10 +28,12 @@ export const ContentLayout = ({
   heroSection,
   contributors,
   lastEditLocaleTimestamp,
+  editBanner,
   ...props
 }: ContentLayoutProps) => {
   return (
     <div {...props}>
+      {editBanner}
```
pettinarip: this is a parent implementation problem that should live here. This banner or whatever the parent want to render should probable be part of `heroSection` and keep the API clean. Small issue but leaves an unnecessary prop imo

src/layouts/index.ts
```diff
@@ -2,34 +2,38 @@ import { MDXRemoteProps } from "next-mdx-remote"
...
 export const layoutMapping = {
   static: StaticLayout,
-  "use-cases": mdLayouts.UseCasesLayout,
-  staking: mdLayouts.StakingLayout,
-  roadmap: mdLayouts.RoadmapLayout,
-  upgrade: mdLayouts.UpgradeLayout,
+  "use-cases": TopicLayout,
+  staking: TopicLayout,
+  roadmap: TopicLayout,
+  upgrade: TopicLayout,
```
pettinarip, line 23: nice

app/[locale]/[...slug]/page.tsx
```diff
@@ -79,6 +81,40 @@ export default async function Page(props: { params: Promise<SlugPageParams> }) {
   const requiredNamespaces = getRequiredNamespacesForPage(slug, layout)
   const messages = pick(allMessages, requiredNamespaces)
 
+  if (topicConfig) {
+    const afterContent =
+      layout === "staking" ? (
+        <StakingCommunityCallout className="my-16" />
```
Comment (pettinarip) [inline app/[locale]/[...slug]/page.tsx:87]: sidecomment, we should probably need to think about a global callout component (in base layout perhaps) to be used for any page that needs something like this

app/[locale]/[...slug]/page.tsx
```diff
@@ -79,6 +81,40 @@ export default async function Page(props: { params: Promise<SlugPageParams> }) {
...
+            slug={slug}
+            frontmatter={frontmatter}
+            tocItems={tocItems}
+            lastEditLocaleTimestamp={lastEditLocaleTimestamp}
+            contentNotTranslated={!isTranslated}
+            contributors={contributors}
+            config={topicConfig}
+            afterContent={afterContent}
+          >
+            {content}
```
Comment (pettinarip) [inline app/[locale]/[...slug]/page.tsx:108]: instead of adding a new prop `afterContent` we could just do
```suggestion
            {content}
            {afterContent}
```

src/data/topics/roadmap.ts
```diff
@@ -0,0 +1,38 @@
...
+        matomoEvent: "clicked roadmap user experience home",
+      },
+      {
+        textKey: "nav-roadmap-future-proofing",
+        href: "/roadmap/future-proofing",
+        matomoEvent: "clicked roadmap future-proofing home",
+      },
+    ],
+  },
+  heroImage: { width: 1456, height: 816 },
```
pettinarip: I'd strip this `heroImage` data from here. This is a UI related data that the Layout knows better the size of the hero image to be displayed.

pettinarip (commented):
@myelinated-wackerow great work overall!

src/data/topics/roadmap.ts
wackerow: @pettinarip thanks for this... yeah this was faulty anyway—image sizes didn't necessarily match but also, it's not appropriate to apply to the full layout since the front matter dictates the actual image (where aspect ratios can/will vary).

Replaced with usage of `sizeOf` as a build-time image-size probe in `compile.ts` so `next/image` gets the correct aspect ratio for each -- similar to our `blurDataURL` -- Double check me that this will work with our setup, but I think that should do it.

(Commit: c930c4c882168a3686e05680f77b8ebaff055540)

app/[locale]/[...slug]/page.tsx
Comment (wackerow) [reply app/[locale]/[...slug]/page.tsx:87]: Agree -- can separate this task

src/layouts/ContentLayout.tsx
wackerow: Agree -- patched

app/[locale]/[...slug]/page.tsx
Comment (wackerow) [reply app/[locale]/[...slug]/page.tsx:108]: Agree -- patched

pettinarip (approved):

Issue #18191, opened 2026-05-14
chore(design-system): add Storybook stories for reusable feature components


myelinated-wackerow:
The companion issue to #18121. That issue tracks Storybook stories for `ui/` primitives. This one tracks the **reusable feature components** in `src/components/` -- multi-page-used components that deserve standalone visual reference but don't live in `ui/`.

Background: a component-disposition audit identified ~22 components in `src/components/` that deserve standalone visual reference: multi-page-used feature components (PRs 1-3), Callout/Banner variants slated for unification (PR 4), and shared-infra layout components (PR 5). A small number have only one consumer today but are explicitly designed for reuse (`ListingMethodology`); those are included. Three of them (`Callout`, `CalloutBanner`, `CalloutSSR`) plus `BannerNotification` and `TranslationBanner` are slated for consolidation into a unified `Callout` per #18133 -- we're storying them *before* that work lands so the consolidation has visual reference for the differences. PR 4 isolates them.

## How to pick up a PR group

1. Look at existing high-quality feature-component stories as exemplars: `AppCard.stories.tsx`, `DocLink.stories.tsx`, `Tooltip.stories.tsx`, `Hero/*/...stories.tsx`.
2. Pick ONE of the five PR groups below. Open a PR scoped to just that group.
3. For each component in your group:
   - If currently a loose `src/components/Foo.tsx`, migrate to `src/components/Foo/index.tsx` first (so the story can live alongside)
   - Add `src/components/Foo/Foo.stories.tsx`
4. Reference this issue in the PR description (don't close the issue).

## Standards for new stories

Follow the same conventions as #18121:

- One `Default` story (canonical, simplest valid usage)
- One story per variant prop (showing all options together)
- 2-3 realistic composite stories
- Title pattern `Components / [ComponentName]` (or `Components / [Category] / [Name]` if a natural grouping exists)
- a11y addon passes; no `chromatic.modes` opt-in unless genuinely needed

**Chromatic snapshots**: per the team convention established with #18121, apply `parameters: { chromatic: { disableSnapshot: true } }` at the meta level on each new story. Snapshots can be re-enabled per-story later once we assess which views need regression coverage.

## Out of scope

- Stories for `UpgradeStatus.tsx` -- pending refactor into `ui/alert` (#18132). Story will come with the refactor.
[...]

Issue #18133, opened 2026-05-06
feat(design-system): unify Callout / CalloutBanner / BannerNotification into single component


myelinated-wackerow:
## Why

The codebase currently has seven overlapping components for "highlighted message" UI:

- `Callout.tsx` (client wrapper around CalloutSSR)
- `CalloutSSR.tsx` (server)
- `CalloutBanner.tsx` (client wrapper, deprecated)
- `CalloutBannerSSR.tsx` (server, canonical post-cva)
- `BannerNotification.tsx` (top-of-page stripe)
- `DismissableBanner/` (wraps `BannerNotification` with localStorage state)
- `TranslationBanner.tsx` (floating bottom banner on non-English non-translated pages -- has a session-only "X" plus a persistent "Don't show again" with localStorage keyed per-path)

These differ in placement and chrome but cover the same conceptual surface. Consolidate into a **single server-renderable `Callout`** with sub-component composition (Card-style) plus a small client island for dismissibility.

## Target API: sub-component composition (Card-style)

The unified Callout follows `ui/card`'s composition pattern -- discrete sub-components the consumer pieces together. This avoids hard-coding heading levels and gives consumers full control over content layout.

```tsx
<Callout size="large" variant="default">
  <CalloutImage src={img} />
  <CalloutContent>
    <CalloutEyebrow>Optional uppercase eyebrow</CalloutEyebrow>
    <CalloutTitle asChild>
      <h2>Title text -- consumer picks the heading level</h2>
    </CalloutTitle>
    <CalloutDescription>Body content.</CalloutDescription>
    <CalloutFooter>
      <Button>CTA</Button>
    </CalloutFooter>
  </CalloutContent>
  <CalloutClose />
</Callout>
```

### Sub-components

- **`Callout`** -- root server component with size + variant props
- **`CalloutImage`** -- optional image decoration; lays out per existing `CalloutSSR` / `CalloutBannerSSR` when present
- **`CalloutContent`** -- text column wrapper
- **`CalloutEyebrow`** -- optional small uppercase leader text
- **`CalloutTitle`** -- the headline. Supports `asChild` (Radix Slot) so consumers control the heading level (`<h2>`, `<h3>`, etc.) instead of the component hard-coding it
- **`CalloutDescription`** -- body paragraph slot
- **`CalloutFooter`** -- CTA / button area
- **`CalloutClose`** -- the dismiss button (client island; details below)

### Root variants

[...]

PR #18202, opened 2026-05-15
chore(design-system): shared layout Storybook stories (PR 5 of #18191)

myelinated-wackerow:
## Summary

PR 5 of #18191. Adds Storybook stories for the two top-level layout components rendered site-wide via `BaseLayout`. Taken before PR 4 so the Callout/Banner work can wait on #18133 coordination.

Components covered:
- `Nav` (async server top-level wrapper) -- `Default` desktop story plus a `MobileViewport` story. Per the issue rubric, sub-components (`DesktopNav`, `MobileNav`, `MobileMenu`, `ThemeToggleButton`) stay un-storied since the top-level Nav composition is the visual canon.
- `Footer` (async server) -- `Default` plus `MobileViewport`. `GoToTopButton` is intentionally left without a standalone story per the issue's "optional if visual is non-trivial" guidance -- its visual is a labelled outline button, which the Footer story already exhibits in context.

Each story uses the `Components / Layout / <Name>` title pattern and opts out of Chromatic snapshots at the meta level per the #18121 convention. Both meta entries use `layout: "fullscreen"` so the bars render against the iframe edge.

Related to #18191. Does not close the issue.

## Test plan

- [ ] `pnpm storybook` -- both new entries appear under `Components / Layout / *` and render without console errors at desktop and `mobile1` viewports
- [ ] a11y addon panel is clean on each story
- [ ] Footer: `GoToTopButton` renders and labels correctly inside the Footer composition
- [ ] Nav: locale switch via storybook-next-intl re-renders the labels (English / German etc.) without breaking the layout

Claude (Opus 4.7) co-authored these changes.

### DIFF
--- a/src/components/Footer/Footer.stories.tsx
+++ b/src/components/Footer/Footer.stories.tsx
@@ -1,0 +1,37 @@
+import { Meta, StoryObj } from "@storybook/nextjs"
+
+import Footer from "."
+
+const meta = {
+  title: "Components / Layout / Footer",
+  component: Footer,
+  parameters: {
+    chromatic: { disableSnapshot: true },
+    layout: "fullscreen",
+    docs: {
+      description: {
+        component:
+          "Site footer rendered through `BaseLayout`. Async server component (`getTranslations`), so the linked labels come from the active locale. Surfaces five link-section columns, social icons, dipper links (privacy, terms, etc.), and the `GoToTopButton`. `lastDeployLocaleTimestamp` is a pre-formatted string -- the caller is responsible for locale-formatting before passing it in.",
+      },
+    },
+  },
+} satisfies Meta<typeof Footer>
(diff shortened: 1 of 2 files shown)


pettinarip (approved):

src/components/Footer/Footer.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: {
+    lastDeployLocaleTimestamp: "May 15, 2026, 9:42 AM",
+  },
+}
+
+export const MobileViewport: Story = {
```
pettinarip, line 30: this case should not be a story itself. Viewports are handled in the SB UI

PR #18208, opened 2026-05-15
Refactor opcodes table into separate component and fix table layout

### LINKED ISSUE #10886: DS implementation - Table layout `fixed`
_This is a part of implementing [the new DS](https://github.com/ethereum/ethereum-org-website/issues/9548)_

***

TODO
1. Set the Table component's layout as `fixed`.
2. Refactor the table in [the opcodes page](https://ethereum.org/en/developers/docs/evm/opcodes/) by moving the table to its own page.
3. Mark the following [TODO comment](https://github.com/ethereum/ethereum-org-website/blob/dev/src/components/Table/index.tsx#L15) as done

kafeelraza:
## Description

* implemented fixed table layout for the opcodes page
* extracted the opcodes table into a dedicated React component
* moved opcode data into a separate data file
* resolved the existing TODO related to table layout
* improved table responsiveness and overflow behavior
* preserved links inside gas, notes, and mem/storage columns

Closes #10886


myelinated-wackerow (changes requested):
Thanks for taking this on @kafeelraza! Issue #10886 has been sitting for a while, appreciate you diving into this, and the direction you're heading looks good. The `Table` primitive cleanups (fixing the broken `[&_tr:last-child]:border-0` selector in the product variant, adopting `table-fixed`, dropping the resolved TODO) are good catches. Few things to tighten up before this is ready to merge—most are mechanical, and the main one is that the component isn't actually wired up yet. I've put it all in a checklist below so it's easy to track.

## Heads-up: the new component isn't reachable yet

`OpcodesTable` is registered in `baseComponents` in `page.tsx`, but `public/content/developers/docs/evm/opcodes/index.md` still contains the original inline `| Stack | Name | ... |` markdown table — it's what's rendering on the preview deploy. The visual changes you're seeing come from your `table.tsx` updates being picked up by the existing markdown table mapping, not from your new component. Fixing this is item 1 below.

## Repairs checklist

**Out-of-scope changes to revert**

- [ ] **Remove debug log** in `generateStaticParams()` (`app/[locale]/[...slug]/page.tsx`) — `console.log("ALL SLUGS:", slugs)`. It'll spam every build.
- [ ] **Revert all 12 `src/data/placeholders/content-*-data.json` files.** These were generated automatically by `pnpm build` running on Windows — note the backslash-escaped paths (`\\content\\translations\\pt-br\\...`). Path lookups go through `toPosixPath`, so those keys are unreachable. A `git checkout origin/dev -- src/data/placeholders/` should drop them cleanly.
- [ ] **Revert `src/lib/utils/md.ts`.** That change isn't part of this refactor — happy to look at it in a separate PR if you'd like to send one.

**Wire the component up**

- [ ] In **the English file only** — `public/content/developers/docs/evm/opcodes/index.md` — replace the markdown table with `<OpcodesTable />`. Leave the translated copies under `public/content/translations/[locale]/developers/docs/evm/opcodes/` untouched; the intl-pipeline will replicate the change once English moves.

**Component cleanup (`src/components/Table/OpcodesTable.tsx`)**

- [ ] **Drop `"use client"`** (line 1). There's no state, effects, or browser APIs here, and the `Table` primitive already handles its own client boundary. As a Server Component the 1,500-line `OPCODES` array stays out of the JS bundle.
- [ ] **Replace the raw `<a>`** in `renderCell` with `InlineLink` from `@/components/ui/Link`. `InlineLink` handles external-link safety and tracking.
- [ ] **Rename `interface Opcodes` → `Opcode`** in `src/data/evm/opcodes.ts` (it represents a single row).
- [ ] **Unify the cell shape.** `type Cell = { label: string; href?: string }`, then use `Cell` for `gas`, `memStorage`, and `notes`. Right now only `gas` is typed with `href?`, so `renderCell(opcode.memStorage)` and `renderCell(opcode.notes)` can't actually render links even when the source markdown does (e.g., the `[sign extend](...)` link on `SIGNEXTEND`'s notes).
- [ ] Drop the `if (!cell) return ""` guard once the type is required, or make the field optional — pick one.
- [ ] **Match the existing column header names** from the markdown: **Initial Stack** (not Inputs) and **Resulting Stack** (not Outputs).

**i18n (translations)**

For a self-contained component, the convention is a dedicated namespace JSON in `src/intl/en/`:

- [ ] Create `src/intl/en/component-opcodes-table.json` with one key per header (`opcode`, `name`, `gas`, `initial-stack`, `resulting-stack`, `mem-storage`, `notes`).
- [ ] In `OpcodesTable.tsx`, read them via `getTranslations("component-opcodes-table")` (works once `"use client"` is removed).
- [ ] Leave the opcode `notes` strings in the TS data (e.g., `"Halt execution"`) as English for now — translating that body is a separate, larger conversation. Worth noting in the PR description so it's clear the regression is intentional and bounded.

**`table.tsx` cleanups**

- [ ] **Restore the compact single-line class strings** in the `baseStyles` object. Repo convention is inline strings — easier to diff. Your semantic additions (`text-sm`, `font-semibold`, `whitespace-normal`, `break-words`) are great; just keep them inline.
- [ ] **Restore the JSDoc comment** above `CellPropsWithAlign` — it explains why the `align` prop diverges from the native HTML `align` attribute (deprecated).
- [ ] **Move `min-w-[1100px]` off the base `table` slot.** As written it forces *every* Table on the site (every variant) to a minimum 1100px width, which will cause horizontal overflow on narrow 2–3 column tables across the site. The opcodes table is what needs it; isolate it there:
  - Apply on the `<OpcodesTable>` instance: `<Table className="min-w-[1100px]">`.
  - Or, even better, add a `<colgroup>` inside `OpcodesTable` with per-column widths — `table-fixed` will respect those, and it'll fix the "Notes column wraps while Opcode is sparse" visual issue without needing `min-w` at all.

---

Let us know if you hit issues. Can give it another review when ready and we'll try to get this in. Thanks again!

kafeelraza:
Thanks for the detailed review! I've addressed the requested changes, including wiring up the new `<OpcodesTable />` component, cleaning up the table primitives, adding i18n headers, reverting unrelated files, and updating the markdown page to use the component. Latest changes are pushed in `dd5b078`.

kafeelraza:
Friendly ping @pettinarip @wackerow 👋

This PR has been open for a couple of weeks. When you have time, could you please take a look and let me know if any changes are needed from my side?

Thank you!

nloureiro:
Apologies for the silence since your June 3 ping — that's on us, and a maintainer response on the open review points is owed here. From your side, two things would unblock it: resolving the merge conflicts with `dev`, and addressing the May 20 review comments. If both sides move, this can get back on track — flagging it for maintainer attention as well.

kafeelraza:
Hi @myelinated-wackerow @wackerow @pettinarip @nloureiro, I’ve resolved the merge conflicts with dev, fixed the workflow parse error from the latest merge, and pushed the updates.

CI lint/type-check and unit tests are passing; Netlify deploy preview is still pending. Could you please take another look when you have a chance?

Thanks!

nloureiro (changes requested):
Thanks for the rebase and the patience — you've been responsive and this deserved a faster re-review from our side. The May-20 checklist is nearly all addressed (server component, i18n namespace, `InlineLink`, renames, instance-scoped `min-w`, placeholders/md.ts reverted — all confirmed), and CI is green. A deep verification pass against the original markdown table found three blocking items, though:

**1. Seven rows are missing.** The original table has 156 data rows; `src/data/evm/opcodes.ts` has 149. The dropped rows are the `_invalid_` range rows: `0C-0F`, `1E-1F`, `21-2F`, `4B-4F`, `A5-EF`, `F6-F9`, `FB-FC`. Please restore them (or state explicitly why they're omitted).

**2. All non-gas links were dropped.** The 30 `gas.md` links survive, but every notes/mem-storage link is gone: `[sign extend]` (SIGNEXTEND), EIP-141 (INVALID), EIP-155 (CHAINID), EIP-1153 ×2 (TLOAD/TSTORE), EIP-4844 (BLOBHASH), EIP-5656 (MCOPY), EIP-7516 (BLOBBASEFEE), and the EIP-1559 GASPRICE footnote. The PR description says these were preserved — and rendering them was the specific reason the unified `Cell` type was requested in May. The `href` field already supports it.

**3. Site-wide hover regression in `ui/table.tsx`:** `baseStyles.tbody` changed from `[&_tr:hover]:bg-background-highlight` (highlight the hovered row) to `hover:[&_tr]:bg-background-highlight` (hovering anywhere highlights **all** rows) — this hits every non-product table on the site. Looks accidental; please revert.

Quick nits (non-blocking): SIGNEXTEND `outputs: ["SIGNEXTEND(x", "b)"]` should be one item `"SIGNEXTEND(x, b)"`; export `Cell` from the data file instead of declaring it twice; trailing newlines on the three new files; and the `claude-review-translations.yml` change looks like a genuine fix (comments inside the `if: |` expression) but please mention it in the description so it can be validated independently.

The refactor's architecture is right and this is close — happy to re-review quickly once the three blockers land.

nloureiro:
@kafeelraza rather than send you on another round trip after the seven-week wait, I've pushed the review fixes directly to your branch (maintainer-edit was enabled — hope that's okay):

- restored the **7 `_invalid_` range rows** — the data now matches the original table's 156 rows
- restored the **9 dropped links** (sign-extension, EIP-141/155/1153×2/4844/5656/7516, and the EIP-1559 gasprice footnote) via the `Cell.href` field
- reverted the tbody hover to per-row highlight
- fixed the SIGNEXTEND outputs split, exported `Cell` from the data file, trailing newlines, removed the blank-line churn in `table.tsx`

`pnpm type-check` passes. One judgment call to note for the maintainer who merges: the original markdown linked only a fragment of some cells (e.g. `[sign extend]`), but `Cell` supports one href per cell, so those cells now render wholly as links — the URLs are all preserved; richer per-fragment links would need a `Cell` shape change and didn't seem worth blocking on.

Please give the commit a quick look and re-request review — from my side the three blockers are resolved and this is ready for a maintainer pass.

nloureiro (commented):
Dismissing my own changes-requested by noting: all three blockers from the re-review are resolved by the pushed commit (row count verified at 156, links spot-checked at their original columns, hover class reverted). Since the fixing commit is mine, final approval should come from another maintainer — flagging as ready for that pass.

kafeelraza:
Thanks @nloureiro for the fixes and follow-up.

I’ve reviewed the latest commit and everything looks good from my side. Since the blockers are resolved and all checks are passing, could @myelinated-wackerow @wackerow @pettinarip please take another look when you have a chance?

Thanks!

wackerow:
@kafeelraza Thank you for the continued help here... Taking another look through this, I think I glossed over the intl/translation issues too much in my previous review. As it stands, this would cause a significant regression in our non-English version of this page, and we'd end up with a bunch of hard-coded English strings throughout the table cells.

@pettinarip -- Issue #10886 notes a few requests, notably:

> Refactor the table in [the opcodes page](https://ethereum.org/en/developers/docs/evm/opcodes/) by moving the table to its own page.

Could you help clarify this? Did you explicitly want an abstracted component?

Alternatively, I think we could accomplish this much simpler:

```tsx
// New file: src/components/ui/wide-table.tsx

import { type ChildOnlyProp } from "@/lib/types"

type WideTableProps = ChildOnlyProp & {
  className?: string
}

// Opt-in wrapper for wide markdown tables (e.g. the EVM opcodes reference).
// Gives the inner table a legible min-width with fixed layout so it scrolls
// horizontally on narrow viewports instead of cramming.
const WideTable = ({ className, ...props }: WideTableProps) => (
  <div
    className={cn("[&_table]:min-w-[1100px] [&_table]:table-fixed", className)}
    {...props}
  />
)

export default WideTable
```

```tsx
// Inside: src/components/MdComponents/index.tsx

import WideTable from "@/components/ui/wide-table"

// ...and add WideTable to the `export const reactComponents` object
```

Then just wrap the existing markdown table with `<WideTable></WideTable>` inside `public/content/developers/docs/evm/opcodes/index.md`:

```mdx
<WideTable>

| Stack | Name           |                                               Gas                                               | Initial Stack                                                                             | Resulting Stack                 | Mem / Storage                                                                 | Notes                                                                                                                            |
...

</WideTable>
```

image

- Keeps existing content unchanged, including all the link snippets throughout the different cells.
- Allows a flexible `WideTable` wrapper to be re-used elsewhere if needed
- Can be applied to English only to start, without affecting or regressing the non-English versions

@pettinarip Let me know if you had something in mind here, or if this trimmed down approach looks good.

kafeelraza:
@wackerow updated based on your feedback. I removed the hard-coded OpcodesTable/data approach, restored the original markdown opcode table for translation compatibility, and wrapped it with a new WideTable MDX component for the layout fix. I also verified the table still has 156 data rows and the previously restored links/invalid ranges are preserved.

pettinarip:
@wackerow I'm fine with that approach. The idea there was to move the table to its own separate page for better UX, since it's too large and users have to scroll both vertically and horizontally. Maybe even replace the table with something easier to visualize.

Not a blocker though. I think we can drop that point, it's pretty stale at this point.

wackerow (approved):
Looks good! Thanks @kafeelraza!

PR #18214, opened 2026-05-18
fix(ui): ExpandableCard layout and spacing

wackerow:
## Description

- fix: remove mb-4 hard-coding on ExpandableCard (fixes /staking/solo/)
- refactor: allow consumer to lay out cards as desired
- refactor: Update non-markdown consumers to dictate spacing from parent wrapper
- feat: add AccordionContainer for consistent layout (single-column for stacks of ExpandableCards, `max-w-5xl`) -- wrap existing stacks with new component
- refactor: Update markdown version to apply bottom margin for parity; max-width managed via markdown page component width
- fix: spacing on /roadmap/, /run-a-node/


src/components/ui/accordion.tsx
```diff
@@ -5,6 +5,15 @@ import { ChevronNext } from "@/components/Chevron"
 
 import { cn } from "@/lib/utils/cn"
 
+const AccordionContainer = ({
+  className,
+  ...props
+}: React.BaseHTMLAttributes<HTMLDivElement>) => {
+  return (
+    <div className={cn("w-full max-w-5xl space-y-4", className)} {...props} />
```
wackerow, line 13: Applies standard container for accordions/ExpandableCards, listing as single column, with consistent max width and spacing; discussed with @konopkja

src/components/MdComponents/index.tsx
```diff
@@ -166,6 +168,13 @@ export const ContentContainer = (props: ComponentProps<"article">) => {
   )
 }
 
+export const ExpandableCardWithMargin = ({
+  className,
+  ...props
+}: ExpandableCardProps) => (
+  <ExpandableCard className={cn("mb-4", className)} {...props} />
```
wackerow, line 181: Alternatively here, if we want to only apply `mb-4` to those that have another ExpandableCard immediately following it, we could use:

```suggestion
  <ExpandableCard
    className={cn(
      "expandable-card has-[+.expandable-card]:mb-4",
      className
    )}
    {...props}
  />
```

(In this case, it's just applying a unique class name `expandable-card` for each instance, then the `has-[+.expandable-card]` it checking if the next sibling also has that class, and applying the `mb-4` for those cases)

---

Taking it another direction, we could consider forcing a margin on the bottom AND top for all of these:

```suggestion
  <ExpandableCard className={cn("my-4", className)} {...props} />
```

This would make sure there is always a little space left before and after, and any margins that preceded or follow it would simply collapse, and not add together. This would probably be my preference for markdown usage.

pettinarip: No strong opinion. If there is no regression with current approach, I'd ship as is for now. Later we can consider these other approaches if we see any issues.

pettinarip (commented):
@wackerow LGTM!

Nice to have: noticed that we have an animation only when the card is collapsed. Not when we expand it. Current behavior feels too abrupt. Would be nice to have a tiny quick animation on open event.

pettinarip (approved):
forgot to approve :P

PR #18217, opened 2026-05-19
patch: app networks and layout

wackerow:
## Description
- Add support for networks that are not shown on /layer-2/ but used by gaming applications, including image assets for each
- Patch UI for application pages, including alignment and adding conditionals for data that isn't always available
- Extract `by {company}` intl string


pettinarip (approved):
LGTM!

PR #18223, opened 2026-05-19
refactor: migrate TypeScript script runners from ts-node to tsx

pettinarip:
## Summary

Migrates every TypeScript script runner in the repo from `ts-node` to [`tsx`](https://tsx.is). `tsx` starts faster, handles CJS/ESM automatically (no more `-O '{"module":"commonjs"}'` flag), and resolves our `@/*` tsconfig path aliases out of the box.

## Changes

| File | Change |
|------|--------|
| `package.json` | `update-tutorials` script + devDependency: `ts-node` → `tsx` |
| `.github/workflows/intl-pipeline.yml` | run the pipeline via `pnpm exec tsx` |
| `.github/workflows/update-chains.yml` | run via `pnpm exec tsx`; drop the now-unnecessary `pnpm add -g ts-node` step |
| `.github/workflows/i18n-sync-external-descriptions.yml` | align onto `pnpm exec tsx` (was `npx tsx`) so all workflows match |
| `.claude/commands/fix-sanitizer-bug.md` | update all 4 `ts-node` invocations, including the inline `-e` evals |
| `pnpm-lock.yaml` | regenerated — `ts-node` tree out, `tsx` in |

## Why this is safe in CI

`pnpm exec` runs the local binary and won't auto-fetch a missing one — but `tsx` is now a devDependency and every workflow runs a plain `pnpm install` (no `--prod`, no `NODE_ENV=production`, no `.npmrc` override) before its `tsx` step, so `node_modules/.bin/tsx` is always present. Verified locally: `tsx -e` supports the inline eval + `require()` pattern used in the sanitizer doc, and `tsx` resolves the `@/*` alias used by `update-chains.ts`.

## Test plan

- [ ] `pnpm update-tutorials` runs without errors
- [ ] `intl-pipeline` workflow passes
- [ ] `update-chains` workflow passes


wackerow:
@pettinarip Note there is another change in progress in PR #18628 that will help address the Windows issue, and the fragile `-O '{ "module": "commonjs" }'`, but keeps `ts-node`. If that gets merged fix, would need rebasing here and would trim the effect of this PR to mainly the `tsx` swap
- #18628

wackerow (approved):

PR #18225, opened 2026-05-20
bug-bounty: add EngineAPI as explicit out-of-scope item

0xMushow:
Added EngineAPI as out-of-scope as it's considered to be trusted, right after the line already mentioning other stuff like API / JSON-RPC.


wackerow (approved):
Looks good @0xMushow, thanks! I pushed one commit adjusting this a bit to minimize the diff -- This will render the new string in the same location, but will prevent triggering repeat translations for the other strings by keeping their key/value pairs unchanged.

0xMushow:
Thanks! I initially didn't want to mix the `li-10` in the middle of everything. LGTM!

PR #18239, opened 2026-05-20
fix: untrack next-env.d.ts and generate via next typegen

pettinarip:
`next-env.d.ts` has shown up dirty after every `dev`/`build` since #18139. Next 16 writes a different path in dev vs build (`.next/dev/types/...` vs `.next/types/...` — [vercel/next.js#85738](https://github.com/vercel/next.js/issues/85738)), so no committed version is stable. [Official guidance](https://nextjs.org/docs/app/api-reference/config/typescript#next-envdts) is to gitignore it.

- Re-add `next-env.d.ts` to `.gitignore` (reverts that piece of #18139).
- `pnpm type-check` → `next typegen && tsc --noEmit`. `next typegen` (Next 15.5+) regenerates the file in ~600ms with no full build.

### DIFF
--- a/.claude/skills/design-system/references/new-component-checklist.md
+++ b/.claude/skills/design-system/references/new-component-checklist.md
@@ -225,6 +225,6 @@ Title hierarchy:
 - [ ] No commented-out code
 - [ ] No `console.log` debug statements
 - [ ] No TODO comments without an owner / issue link
-- [ ] Type errors clean (`npx tsc --noEmit`)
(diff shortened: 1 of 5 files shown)


wackerow (approved):
Thanks @pettinarip, looks good! Just added a commit updating one skill and the agents documentation to point to use of `pnpm type-check` over `npx tsx --noEmit` to avoid errors on fresh clones/worktrees.

PR #18242, opened 2026-05-20
Add copy page button to markdown-generated pages

pettinarip:
## Visual changes

**New: "Copy .md" button**
- Appears at the top of every markdown page (docs, guides, tutorials, topic hubs)
- Dropdown menu with: *View as Markdown*, *Open in ChatGPT*, *Open in Claude*
- Has copy / loading / copied states

**Unified "page actions" row under the H1**
- Same row across all markdown-driven layouts (Docs, Static/Guides, Topic, Tutorial), so the top of every article looks consistent
- Contains: Copy .md + Edit on GitHub link + Listen-to-article player (where applicable)
- Wraps cleanly on mobile

**Edit on GitHub restyled**
- Was: a full outline button inside the desktop TOC
- Now: a small inline text link with GitHub icon, living in the actions row
- Removed from the right-hand TOC sidebar entirely

**Listen-to-article player redesigned**
- Was: large block-style bar embedded in the article body
- Now: a compact secondary button sitting in the actions row
- On mobile, label collapses to just "Listen"

**Static / Guide pages — top section tightened**
- Markdown H1 is now the page title (one title instead of separate header + body H1)
- "Last updated: …" timestamp removed from the top
- Mobile TOC moved above the H1 (was below)
- Less vertical air at the top of the page

**Docs pages — contributors moved**
- Contributors block moved from top of article (under H1) to the bottom, separated by a top border
- H1 sits tighter to the actions row

## Test plan

- [x] Visit a page on each layout and confirm the actions row appears: `/en/staking/solo/` (Topic), `/en/about/` (Static), `/en/developers/docs/intro-to-ethereum/` (Docs), `/en/developers/tutorials/erc20-annotated-code/` (Tutorial)
- [x] Click "Copy .md" → markdown lands on clipboard, button transitions to "Copied"
- [x] Dropdown: *View as Markdown* opens the raw `.md`; *ChatGPT* / *Claude* open with the page URL pre-filled
- [x] Switch to a non-English locale (e.g. `/es/staking/solo/`) — markdown link points at the translated file, falls back to English when missing
- [x] Listen-to-article player shows in the actions row on pages that have a playlist; collapses to "Listen" on mobile
- [x] Resize to mobile: actions row wraps cleanly, dropdown still works
- [x] Dark mode renders correctly


src/components/CopyPageButton/icons.tsx
```diff
@@ -0,0 +1,13 @@
+import type { SVGAttributes } from "react"
```
myelinated-wackerow: The project convention (per `CLAUDE.md` and the `src/components/icons/` directory) is to declare SVG icons as `.svg` files loaded via `@svgr/webpack`, then import them as React components -- the same way `PageActions/index.tsx` imports `Github from "@/components/icons/github.svg"` in this same PR. The only `.tsx` files under `src/components/icons/` are compound icons built on top of lucide-react (e.g. `CheckCircle.tsx`), not raw SVGs.

Any reason not to add these as `src/components/icons/chatgpt.svg` and `src/components/icons/claude.svg` and import them the standard way? Keeps them consistent with the rest of the icon set and shareable if another component ever needs them.

src/components/CopyPageButton/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+  title,
+  description,
+  eventOptions,
+}: CopyPageMenuItemProps) => (
+  <DropdownMenuItem asChild>
+    <a
+      href={href}
+      target="_blank"
+      rel="noopener noreferrer"
+      onClick={() => trackCustomEvent(eventOptions)}
```
myelinated-wackerow, line 71: Could these use the project's `InlineLink` (`src/components/ui/Link.tsx`) instead of raw `<a>` tags? It already handles `target="_blank"`, `rel="noopener noreferrer"`, the screen-reader `(opens in a new tab)` hint, and Matomo tracking via `customEventOptions`. For the `mdPath` item, `InlineLink` would also auto-detect it as an internal file (since `isFile` returns true for any extension) and route via `NextLink`.

Related question on the `ArrowUpRight` icon inside each item: if we use `InlineLink`, it appends its own external-link arrow inline with the text. Hand-rolling an `ArrowUpRight` and then passing `hideArrow` to suppress the built-in one feels like fighting the abstraction -- worth doing only if the visual placement (top-right of the title row, aligned with the title rather than after the description) is intentional and the inline arrow looks wrong. If the standard inline arrow is acceptable, you could drop both the manual icon and `hideArrow`.

src/components/CopyPageButton/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+  const locale = useLocale()
+  const { onCopy, hasCopied } = useClipboard({ timeout: 2000 })
+  const [isLoading, setIsLoading] = useState(false)
+
+  const cleanSlug = slug.replace(/^\/+|\/+$/g, "")
+  const useTranslatedPath = locale !== DEFAULT_LOCALE && isTranslated
+  const mdPath = useTranslatedPath
+    ? `/content/translations/${locale}/${cleanSlug}/index.md`
+    : `/content/${cleanSlug}/index.md`
+  const pagePath = `/${locale}/${cleanSlug}/`
```
myelinated-wackerow: `mdPath` correctly maps English -> `/content/` and non-English -> `/content/translations/{locale}/`, which matches the actual filesystem layout. Good.

For `pagePath` though: `src/i18n/routing.ts` uses `localePrefix: "as-needed"`, so the canonical English URL is `ethereum.org/smart-contracts/`, not `ethereum.org/en/smart-contracts/`. Right now `pagePath` always prepends the locale, which means the URL we feed to ChatGPT/Claude for English pages is a redirect URL rather than the canonical one. Probably still works, but some LLM fetch tooling chokes on redirects. Could the locale prefix be skipped when `locale === DEFAULT_LOCALE` (similar to `mdPath`)?

src/components/PageActions/index.tsx
```diff
@@ -0,0 +1,62 @@
+"use client"
```
myelinated-wackerow, line 1: Question on the `"use client"` directive: the only client-only bits are `usePathname()` and `useTranslation()`. Both could be replaced -- pathname derived from `slug` + locale at the (server-component) layout level and passed in as a prop, and `getTranslations("common")` used server-side for the `edit-page` label. That'd leave `PageActions` as a server wrapper around the `CopyPageButton`/`ListenToPlayer` client islands and ship slightly less JS for the wrapping div + edit link. Not a blocker since the children are already client -- more curious whether you considered it and the marginal gain wasn't worth the refactor.

src/components/PageActions/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+        className
+      )}
+    >
+      {children}
+      <CopyPageButton slug={slug} isTranslated={isTranslated} />
+      {!hideEditButton && editPath && (
+        <InlineLink
+          href={editPath}
+          hideArrow
+          className="inline-flex items-center gap-2 text-xs text-body! no-underline! hover:text-primary-hover! [&>svg]:size-4"
```
myelinated-wackerow, line 47: The trail of `!`-important utilities (`text-body!`, `no-underline!`, `hover:text-primary-hover!`) suggests we're fighting `InlineLink`'s default styling. Per the design-system guidance ("use a variant, not a new component"), would it make sense to add a `variant` to `InlineLink` for this "quiet / unstyled inline icon-link" case? If only one caller needs it, leave as-is; if a second caller appears, that's the trigger to extract.

src/components/ui/button-group.tsx
```diff
@@ -0,0 +1,89 @@
+import * as React from "react"
+import { cva, type VariantProps } from "class-variance-authority"
```
myelinated-wackerow: `CLAUDE.md` says new components should use `tailwind-variants` (`tv`) rather than `cva`. Looks like this file was lifted from shadcn upstream (which still ships `cva`), so I get the temptation -- just confirming whether you considered porting it to `tv` for consistency, or whether you'd argue the shadcn-parity is worth more here. Not a strong opinion either way.

myelinated-wackerow (commented):
Hey @pettinarip -- really clean implementation overall, especially the unified actions row across layouts and the mobile-collapsing listen player. Left some inline comments -- all non-blocking suggestions and questions, none are deal-breakers.

One observation that didn't fit a single line, on the ListenToPlayer placement: `Static.tsx` gates `<ListenToPlayer />` on `getPlaylistBySlug(slug).index !== -1` and renders it inside `PageActions` as children, but `Topic.tsx` / `Tutorial.tsx` / `Docs.tsx` don't pass `ListenToPlayer` at all. The current playlist slugs all happen to be Static or custom pages, so it isn't broken today -- just worth flagging that if a playlist entry is ever added for a Topic/Tutorial/Docs slug, the player will silently not render. Would it be cleaner to centralize the playlist-gating-and-render inside `PageActions` itself so all four layouts behave consistently?

src/layouts/Topic.tsx
```diff
@@ -134,6 +135,12 @@ export const TopicLayout = async ({
       heroSection={heroSection}
       showDropdown={frontmatter.showDropdown ?? true}
     >
+      <PageActions
+        slug={slug}
+        isTranslated={!contentNotTranslated}
+        editPath={getEditPath(slug)}
+        className="-ms-2 !mt-8 mb-8 lg:!mt-0 [&+h2]:!mt-0"
```
myelinated-wackerow, line 148: Could you add a short comment explaining what these `!`-important overrides are fighting (looks like default prose-margin / first-child rules from `ContentLayout` or the markdown stylesheet)? Otherwise the next person to touch this layout will have a hard time figuring out which rule wins and why.

src/components/CopyPageButton/icons.tsx
pettinarip: Good call — moved them to `chatgpt.svg`/`claude.svg` under `src/components/icons/`. Done in 7419a6c.

src/components/CopyPageButton/index.tsx
pettinarip: Fair point on `InlineLink` handling the boilerplate. The `ArrowUpRight` placement here is intentional though — it sits at the top-right of the title row (aligned with the title, not after the description), so the inline arrow would land in the wrong spot. Leaving the raw `<a>` + manual icon for that reason, but happy to revisit if the layout changes.

src/components/CopyPageButton/index.tsx
pettinarip: Nice catch — fixed in 7419a6c, `pagePath` now mirrors `mdPath` and drops the locale prefix for the default locale so we feed the canonical URL.

src/components/PageActions/index.tsx
pettinarip: Yeah, considered it — `usePathname` and `getTranslations` could both go server-side and leave just the children as client islands. Skipped it for now since the gain is marginal (one div + edit link) and the children are client anyway, but it's a sensible refactor if we revisit this area.

src/components/PageActions/index.tsx
pettinarip: Agreed it's a bit smelly. Leaving as-is since this is the only caller — happy to extract a `quiet`/`icon-link` variant on `InlineLink` the moment a second one shows up.

src/components/ui/button-group.tsx
pettinarip: Yeah, it was straight from shadcn upstream. Ported to `tv` in 7419a6c — consistency wins.

src/layouts/Topic.tsx
pettinarip: Added a comment in 7419a6c explaining they're fighting `ContentContainer`'s `*:first:mt-0` plus the prose top margin on the next h2.

pettinarip:
Thanks for the thorough review! Addressed in 7419a6c:

- **ListenToPlayer placement** — great catch on the latent inconsistency. Centralized the playlist gate + render inside `PageActions` (using the lazy loader to avoid bloating layouts without playlist slugs), and dropped the children pattern from `Static.tsx`. All four layouts now behave the same.
- **Icons** — moved to `chatgpt.svg`/`claude.svg` under `src/components/icons/`.
- **`pagePath`** — canonical URL, drops the locale prefix when `locale === DEFAULT_LOCALE`.
- **`button-group`** — ported from `cva` to `tv`.
- **`!`-overrides** — added an explainer comment in `Topic.tsx`.

Pushed back on a few (raw `<a>` for the arrow placement, `"use client"` on `PageActions`, `InlineLink` variant) — replied inline with the reasoning.

pettinarip:
@wackerow merging this

PR #18220, opened 2026-05-19
Automate weekly release preparation

pettinarip:
## Summary

- New `weekly-release.yml` workflow: Thursday cron + manual dispatch that runs `/prepare-release`, waits for CI, runs `/review-release --post-comment`, and posts status to Discord by editing one rolling message.
- New composite actions: `discord-notify` (create/edit webhook message) and `wait-for-pr-checks` (poll `statusCheckRollup` with transient retry).
- Refactored `ci.yml` to use `wait-for-pr-checks` instead of inline curl+python.
- `/review-release`: dropped auto-approve/request-changes paths; added opt-in `--post-comment` flag for posting the summary as a non-review PR comment.

## Test plan

- [ ] Trigger `weekly-release.yml` via `workflow_dispatch` on a test run and confirm Discord receives a single rolling status message (created, then edited as the job progresses).
- [ ] Verify `/prepare-release` step opens the deploy PR as expected.
- [ ] Verify `wait-for-pr-checks` correctly polls `statusCheckRollup` and tolerates transient API failures.
- [ ] Run `/review-release --post-comment` against a real deploy PR and confirm it posts a comment (not a review) and does not approve or request changes.
- [ ] Re-run `ci.yml` and confirm the refactored `wait-for-pr-checks` step behaves the same as the previous inline curl+python.

PR #18257, opened 2026-05-22
fix(weekly-release): grant id-token permission and require RELEASE_BOT_TOKEN

pettinarip:
## Summary

Two fixes to `weekly-release.yml` after the first manual test run on `dev` ([run 26290559253](https://github.com/ethereum/ethereum-org-website/actions/runs/26290559253)) failed:

- **Grant `id-token: write`** at the workflow level. `anthropics/claude-code-action@v1` authenticates with Anthropic via GitHub OIDC and the failed run logged `Could not fetch an OIDC token. Did you remember to add 'id-token: write' to your workflow permissions?`
- **Drop the `secrets.GITHUB_TOKEN` fallback** for `RELEASE_BOT_TOKEN`. The fallback looked safe but silently degraded: with `GITHUB_TOKEN` authoring the deploy PR, no `pull_request` workflows would fire on it (CI, Netlify preview), while Discord still reported success. Better to fail loudly if the bot token isn't set.

## Test plan

- [ ] Confirm `RELEASE_BOT_TOKEN` is set in repo secrets (currently a PAT scoped to `ethereum/ethereum-org-website` with Contents, PRs, Workflows R/W).
- [ ] Re-trigger `weekly-release.yml` via `workflow_dispatch` and confirm the `Run /prepare-release` step no longer fails on OIDC.
- [ ] Confirm the deploy PR (if one is opened) shows the bot identity as author and triggers downstream CI + Netlify preview.
- [ ] If `RELEASE_BOT_TOKEN` is unset, confirm the workflow fails fast at the checkout / `gh` step instead of silently degrading.

### DIFF
--- a/.github/workflows/weekly-release.yml
+++ b/.github/workflows/weekly-release.yml
@@ -20,9 +20,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     env:
-      # Use a bot token so the auto-created deploy PR triggers downstream
-      # pull_request workflows (GITHUB_TOKEN-authored PRs don't).
-      GH_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN || secrets.GITHUB_TOKEN }}
+      # Required: deploy PR must be authored by a bot token so downstream
+      # pull_request workflows fire.
+      GH_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN }}
     outputs:
       skipped: ${{ steps.skip_check.outputs.skip }}
       skip_reason: ${{ steps.skip_check.outputs.reason }}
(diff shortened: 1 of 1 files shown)


.github/workflows/weekly-release.yml
```diff
@@ -19,9 +20,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     env:
-      # Use a bot token so the auto-created deploy PR triggers downstream
-      # pull_request workflows (GITHUB_TOKEN-authored PRs don't).
-      GH_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN || secrets.GITHUB_TOKEN }}
+      # Required: deploy PR must be authored by a bot token so downstream
+      # pull_request workflows fire.
+      GH_TOKEN: ${{ secrets.RELEASE_BOT_TOKEN }}
```
wackerow, line 25: Nice, I see this loaded now, thanks

.github/workflows/weekly-release.yml
```diff
@@ -9,6 +9,7 @@ on:
 permissions:
   contents: write
   pull-requests: write
+  id-token: write
```
wackerow, line 12: 👍

wackerow (approved):
Looks good @pettinarip, sorry for the premature merge on #18220 but I think this should do it

PR #18261, opened 2026-05-22
fix(storybook): use useArgs in FilterBar story

pettinarip:
## Summary

- Replace inline React `useState` in `FilterBar.stories.tsx` render functions with `useArgs` from `storybook/preview-api`.
- This unblocks the Chromatic build, which was failing with Storybook 10's *"When combining Storybook hooks (e.g. useArgs) with framework hooks (e.g. React's useState, useEffect, useRef) in the same render function, use Storybook's equivalents from 'storybook/preview-api' instead"* error.
- As a bonus, the Controls panel now drives the FilterBar state instead of being decorative — round-trip Storybook integration.

## Test plan

- [x] `pnpm type-check` passes
- [x] `pnpm build-storybook` succeeds locally
- [ ] Chromatic build goes green

### DIFF
--- a/.storybook/withNextThemes.ts
+++ b/.storybook/withNextThemes.ts
@@ -1,10 +1,10 @@
-import { useEffect } from "react"
 import { useTheme } from "next-themes"
 import {
   type DataAttributeStrategyConfiguration,
   DecoratorHelpers,
 } from "@storybook/addon-themes"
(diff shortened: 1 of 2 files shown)


pettinarip:
Superseded by a new PR that contains only the real fix (the decorator hook import). Closing this one to keep the FilterBar useArgs refactor separate from the actual Chromatic build fix.

PR #18295, opened 2026-05-28
fix(docs): correct smart-contracts example URL in style guide

nloureiro:
## Summary

- The "Linking to internal pages" section of the content standardization style guide showed `/docs/developers/smart-contracts/` as the good example
- The actual path for smart contracts docs is `/developers/docs/smart-contracts/`
- Fixed the good and bad examples to use the correct path structure

## Context

Found while reviewing the style guide for internal linking guidance. The guide correctly explains not to hard-code `/en/` prefixes, but the example URL itself used the wrong path order.

## Test plan

- [ ] Verify the updated example URL `/developers/docs/smart-contracts/` resolves correctly on the site


wackerow (approved):
😆 wow

PR #18298, opened 2026-05-28
patch: rm noir from post tags

myelinated-wackerow:
## Description
- Remove noir from post tags

cc: @mnelsonBT


mnelsonBT (approved):
LGTM, thanks @wackerow !

wackerow (approved):

PR #18302, opened 2026-05-28
fix: add alt text to contributor avatar images on /contributing/

nloureiro:
## Summary

- Ahrefs flagged 73 main-site pages (`Warning-Missing_alt_text.csv`) for images missing alt text
- All flagged images were GitHub avatar `` elements rendered by `ContributorsView` on every locale variant of `/contributing/`
- The component was using `alt=""` (empty string); changed to `alt={contributor.name}` so each avatar has a meaningful description for both SEO crawlers and screen readers

## Test plan

- [ ] `/contributing/` — inspect contributor avatar images in DevTools, confirm `alt` attributes contain contributor names
- [ ] After next Ahrefs crawl: `Warning-Missing_alt_text.csv` row count for `ethereum.org` should drop from 73 to 0

### DIFF
--- a/src/components/Contributors/ContributorsView.tsx
+++ b/src/components/Contributors/ContributorsView.tsx
@@ -27,12 +27,12 @@ const ContributorCard = ({ contributor }: { contributor: Contributor }) => {
       <img
         className="size-[132px]"
         src={contributor.avatar_url}
-        alt=""
+        alt={contributor.name}
         loading="lazy"
         decoding="async"
       />
       <div className="p-4">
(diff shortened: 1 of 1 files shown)


wackerow:
@nloureiro Can you check details on what flagged this? The alt text isn't missing, it's intentionally left as an empty string since it's a decorative element.. the users name is displayed for any screen readers right next to the avatar.

wackerow:
Closing this one out:

1. Avatar is decorative; contributor name already rendering next to it
2. With that, link has a valid accessible name
3. This may create a warning, but should not be an error, Empty string. (`""`) is a valid `alt`

PR #18312, opened 2026-05-29
fix(ui): trim surrounding whitespace from image assets

myelinated-wackerow:
## Description
- Whitespace embedded in images is variable, and causes less precise handling of how the images render
- CSS padding can be used to more precisely apply whitespace around images as-needed, 
- PR trims whitespace surrounding images, allowing subject of images to properly fill their containers in a more reliable and  consistent manner

Out-of-scope follow-up tasks:
- Remove duplicated images and settle on a single canonical file for any given image asset (#18331)
- Rename images to be descriptive of what the image actually contains, not based on how it's used (#18331)
- Organize image assets into proper folder structure to improve the ability to find an asset when needed
- Update /assets/ page with full set and develop plan for continuous maintenance for this page

## Related Issue
design system / ui clean-up


pettinarip (approved):
Great!

PR #18315, opened 2026-05-29
feat(stories): Ethereum adoption map section (d3-geo)

pettinarip:
> Draft for a **preview deploy**. Base is `dev` for now to get a Netlify preview; I'll retarget the base to the `stories` branch (this stacks on the `/stories` page work, not on `dev` directly).

## What this adds
Re-adds the "Ethereum adoption on planet earth" section to `/stories` (position 3, between Discover and More stories) as an interactive world choropleth.

- **Rendering:** `d3-geo` (flat **Mercator**) + `topojson-client` over public-domain **world-atlas 110m** geometry. Projection centered on ~11°E so the antimeridian seam falls in the Bering Strait (Russia stays whole); far-western Aleutian polygons that cross the seam are dropped. Paths computed at module scope → SSR-stable.
- **Color:** single-hue **indigo** intensity scale matching Figma, on a rounded lavender panel. Driven by `--map-ink` / `--map-surface` CSS vars that flip in **dark mode** (no white box). No-data countries use a visible light-indigo baseline (not gray).
- **Performance:** below the fold, so the map is **lazy-loaded** (`dynamic`, `ssr: false`) and gated on `IntersectionObserver`; the heavy geometry/JS is a non-blocking chunk. Initial `/stories` HTML drops ~43%. Reserved aspect-ratio box → no CLS.
- **Interaction:** hover/tap tooltip (localized country name via `Intl.DisplayNames` + `numberFormat`); hovered country outline drawn as a top overlay so shared borders aren't occluded.
- **a11y:** heading, and an `sr-only` country→rate list are server-rendered; map `aria-hidden`.
- **Data is mock** (`src/data/storiesAdoption.ts`) pending a real adoption metric.

## Notes
- Map geometry is MIT/public-domain (world-atlas / Natural Earth) — **not** amCharts (CC BY-NC, rejected).
- `react-simple-maps` is unusable on React 19 (`defaultProps`); `d3-geo` directly is the modern replacement.

## Verify
- `pnpm type-check` ✓, `eslint` ✓
- Preview: `/stories` → scroll to the adoption section (light + dark).


konopkja:
Screenshot 2026-06-10 at 15 33 41

lets hide dates here for now since all is last year

wackerow:
@pettinarip Tried to catch this up with latest `dev` -- feel free to revert to commit ae29681. Keeping in draft, but lemme know if we have plans for this

wackerow:
@konopkja Any thoughts on wanting to move forward with this? Would require some work to get it functional for mobile. Currently in a semi-abandoned state

konopkja:
i think its still relevant enhancement to the page, not the main priority based on the page's popularity so id say it depends on effort.

> https://deploy-preview-18315.ethereum.it

PR #18329, opened 2026-06-01
refactor(ui): standardize headings on font-black

myelinated-wackerow:
## Summary

Standardizes all site headings on `font-black` (weight 900). The base `h1`-`h6` default and `CardTitle` now set `font-black`, and the many components that hardcoded `font-bold`/`font-semibold` on headings have had those weights removed so they inherit the new default.

## What changed

- `src/styles/base.css`: `h1`-`h6` default weight `font-bold` -> `font-black`.
- `src/components/ui/card.tsx`: `CardTitle` drops its per-variant weight; the `variant` (semibold/bold/black) prop becomes a `size` prop (sm/lg) and the title inherits `font-black`. Call sites migrated accordingly.
- Removed hardcoded `font-bold` from headings across 23 files (incl. the shared MDX `commonHeadingAttributes` helper) so markdown content and component headings inherit the base weight.
- Removed hardcoded `font-semibold` from headings across 11 files (incl. the MDX `Heading4` helper), kept as a separate commit for clean revertability.
- Intentional `font-normal`/`font-medium` eyebrows, kickers, and empty-state headings were deliberately left untouched.

## Preview link
https://deploy-preview-18329.ethereum.it/

## Test plan

- [ ] Chromatic/Storybook visual regression for `Card` and heading-bearing components.
- [ ] Spot-check rendered heading weight on: a markdown content page, `/developers`, `/what-is-ethereum`, `/community/support`, and a `Card`-heavy page.
- [ ] Verify heading weight on an RTL locale (`ar`/`ur`), where Inter's Latin subset hands off to a fallback font that may not carry weight 900.


konopkja:
Consider for non black:
- FAQ expandable cards
- developers hub page: second section benefits should not be headers?

app/[locale]/developers/page.tsx
```diff
@@ -118,7 +118,7 @@ const WhyGrid = () => {
         <div className="flex gap-1.5" key={heading}>
           <CheckCircle />
           <div className="space-y-1">
-            <h3 className="text-lg">{heading}</h3>
+            <h3 className="text-lg font-bold">{heading}</h3>
```
Comment (pettinarip) [inline app/[locale]/developers/page.tsx:122]: calling out this one...shouldn't we default to `-black` instead?

pettinarip (approved):
@myelinated-wackerow Niceee!

- Possible straggler — `src/components/ExpandableCard.tsx:87` still has `<h3 className="text-xl font-semibold">`

app/[locale]/developers/page.tsx
Comment (wackerow) [reply app/[locale]/developers/page.tsx:122]: Yeah good question -- Chatted with @konopkja and the ExpandableCards and the cards you're highlighting here were both called out as ones we decided to keep as just `bold` instead.

Overall approach is to standardize the defaults, but still allow for overrides within specific components. Likely worth even more discussion on just how liberal we want to be with that.

konopkja (approved):

PR #18334, opened 2026-06-01
feat(ui): add CardButtonFake

wackerow:
## Description
Adds "fake" button to ui/card exports for use when a button is visually desired inside a full-card link

- Semantic `div` with styling mirroring `ui/button` for use nested inside `ui/card` when `Card` component passed an `href` (thus semantically an anchor element, disallowing the nesting of another anchor or button)
- Enabled `hover-link` custom tailwind variant for proper hover handling on "buttons" that end up  either `button`, `a` or `div` element variations
- Implements on homepage cards
- Handled as it's own server-side component in lieu of using `useContext` to decide based on surrounding div if a "ButtonLink" should be re-styled or not.

## Related Issue
Homepage cards: request to use button for visual recognition inside full-card link -- cc: @konopkja

### DIFF
--- a/src/components/Homepage/GetStartedGrid.tsx
+++ b/src/components/Homepage/GetStartedGrid.tsx
@@ -5,6 +5,7 @@ import { Image } from "@/components/Image"
 import {
   Card,
   CardBanner,
+  CardButtonFake,
   CardContent,
   CardFooter,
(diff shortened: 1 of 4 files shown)


pettinarip (approved):
LGTM!

PR #18346, opened 2026-06-03
feat(search): evaluate Typesense as Algolia DocSearch replacement

pettinarip:
## What

Evaluation spike swapping Algolia DocSearch for **Typesense** via the `typesense-docsearch-react` drop-in fork. The search modal, styling and keyboard UX are untouched — only the backend client changes, so this is an apples-to-apples comparison of the two engines.

**Draft / not for merge.** Production stays on Algolia.

## Key changes

- `@docsearch/{react,css}` → `typesense-docsearch-{react,css}`; Typesense server config + `filter_by: language:=<locale>`
- `docsearch:language` meta tag (`src/lib/utils/metadata.ts`) — the self-hosted scraper does **not** derive language from `<html lang>` the way Algolia's hosted crawler does, so the field has to be fed explicitly
- `typesense/` — in-repo scraper config + run instructions
- `NEXT_PUBLIC_TYPESENSE_*` env vars

## Status

Frontend verified locally against the live index: modal, highlighting, typo tolerance, clean result URLs, 0 console errors.

Relevance is **not** measurable yet — a re-scrape is needed first (`language` is empty on all records, and ar/zh are under-crawled). Browser access also needs CORS enabled on the Typesense host.

## Verdict criteria

Relevance first (≥ Algolia on a fixed query set incl. ar/zh), then latency, with ops/cost as tiebreakers — noting that adopting Typesense means owning the crawler and losing Algolia's global edge.


wackerow:
Closing in lieu of:
- #19194 
- #19231

PR #18357, opened 2026-06-05
redesign(ui): FeedbackCard with text-h* utility classes

myelinated-wackerow:
## Description
Update FeedbackCard component to latest designs

- feat: create custom tailwind utility classes (`utilities.css`) for all h1-6 `font-weight` and `line-height` styles, which can be applied using `text-h*`. Enables easy application of a specific header side to component elements.
- refactor: updated `base.css` to point h1-6 base `font-weight` and `line-height` styling to new utilities classes (e.g., `h1 { @apply text-4xl lg:text-5xl; }` -> `h1 { @apply text-h1; }`
- redesign: `FeedbackCard` to remove border, background and padding. Apply `text-h3` styling to `h2` element. Tidy spacing between elements, placing paragraph text closer to `h2` (1rem) than to the buttons (2rem). 

## Related Issue
Design system/theming

### DIFF
--- a/src/styles/utilities.css
+++ b/src/styles/utilities.css
@@ -45,63 +45,27 @@
 }
 
 @utility text-h1 {
-  font-size: var(--text-4xl);
-  line-height: var(--tw-leading, var(--text-4xl--line-height));
-
-  @media (width >= theme(--breakpoint-lg)) {
-    font-size: var(--text-5xl);
-    line-height: var(--tw-leading, var(--text-5xl--line-height));
-  }
+  @apply text-4xl lg:text-5xl;
 }
 
 @utility text-h2 {
-  font-size: var(--text-3xl);
-  line-height: var(--tw-leading, var(--text-3xl--line-height));
-
-  @media (width >= theme(--breakpoint-lg)) {
-    font-size: var(--text-4xl);
-    line-height: var(--tw-leading, var(--text-4xl--line-height));
-  }
+  @apply text-3xl lg:text-4xl;
 }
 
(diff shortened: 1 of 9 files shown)


src/styles/utilities.css
```diff
@@ -44,6 +44,66 @@
   background-image: var(--ten-year-gradient);
 }
 
+@utility text-h1 {
```
pettinarip, line 47: wondering if we can just do `@utility text-h1 { @apply text-4xl lg:text-5xl; }` to:
- make it cleaner
- reuse the breakpoint
- reuse the already defined pairs in theme.css

pettinarip (commented):
@myelinated-wackerow looks good!

nit: the PR scope is bigger than just the redesign :)

**kind of a blocker:** update the skills and docs to enforce the usage of `text-h*` to not end having both systems in the codebase and losing the abstraction.
> heading-like text always uses text-h*, never the raw size utilities

src/styles/utilities.css
wackerow: Ah syntax highlighter got me -- I definitely agree and tried that first but it appeared in the code as if it was invalid syntax. It does appear to work though 🔥

wackerow:
@pettinarip updated using simplified `@apply` syntax, and added the design-system skill updates.

nloureiro (approved):

wackerow:
Thanks @nloureiro!

@pettinarip Going to pull this in, but please let me know if any other issues you see and I'll be happy to make follow-up adjustments

PR #18359, opened 2026-06-05
fix: wrap sub-filter options to prevent desktop filter sidebar overflow

### LINKED ISSUE #18358: Find wallet page: filter sidebar clips text and breaks layout when Desktop OS sub-filters expand (desktop)
## Description

On the [Find a wallet](https://ethereum.org/wallets/find-wallet/) page, opening the filter sidebar on desktop and toggling the **Desktop** device filter (which reveals Linux/Windows/macOS sub-checkboxes) breaks the layout:

1. **Text clipping** — filter option labels and descriptions are truncated on the right side with no wrapping (visible in the sidebar for labels like "Buy crypto with fiat directly in th…", "You can connect to applications…", etc.)
2. **Layout overflow** — the sidebar content doesn't wrap properly when the OS sub-filter checkboxes (Linux / Windows / macOS) appear below the Desktop toggle, causing the filter panel to overflow its container and clip against the wallet results column.

## Steps to reproduce

1. Go to https://ethereum.org/wallets/find-wallet/
2. On a desktop viewport, open the **Filters** panel
3. Under **Device**, toggle **Desktop** on
4. Observe the Linux/Windows/macOS sub-checkboxes appearing and the sidebar layout breaking

## Expected behaviour

Filter text wraps correctly; sidebar maintains its width and does not clip content or overlap the results column.

## Actual behaviour

Text is clipped on the right edge of the sidebar. The OS sub-filter row overflows the container width.

## Environment

- Production: https://ethereum.org/wallets/find-wallet/
- Desktop viewport (observed on macOS)
- Affects the sidebar filter panel when the Desktop OS sub-filters are visible

nloureiro:
## Summary

Fixes #18358

Two layout bugs in the desktop filter sidebar on the Find a Wallet page (and any other `ProductTable` page).

### 1. Sub-filter options row overflows (`Filter.tsx`)

When the **Desktop** device filter is toggled on, a row of sub-option checkboxes (Linux / Windows / macOS) is rendered with `flex flex-row` and no wrapping. Inside the `w-80` sidebar (with `p-6` padding, leaving ~272px of usable width), the three checkboxes overflow horizontally — pushing the sidebar beyond its intended width and shoving the wallet results column off-screen.

**Fix:** `flex-row` → `flex-row flex-wrap` with `gap-x-6 gap-y-2` so the sub-option row wraps cleanly.

### 2. Filter description text clips on the right (`SwitchFilterInput.tsx`)

`Field` applies `[&>*]:w-full` to all its direct children. `FieldDescription` had `ms-8` (margin-start), which shifts the element 32px from the left *but keeps its width at 100% of the container* — overflowing 32px past the right edge.

**Fix:** `ms-8` → `ps-8` (padding-start). Same visual indent (text aligns under the label, past the icon), but the element stays within its container bounds.

## Test plan
- [ ] Go to https://ethereum.org/wallets/find-wallet/
- [ ] Open the **Filters** panel on a desktop viewport
- [ ] Toggle **Desktop** on under Device — Linux / Windows / macOS checkboxes should wrap to the next line instead of overflowing the sidebar
- [ ] Wallet results column should remain fully visible alongside the open filter panel
- [ ] Filter descriptions (e.g. "You can connect to applications that support WalletConnect…") should wrap within the sidebar without clipping on the right edge

## Preview link
https://deploy-preview-18360.ethereum.it/wallets/find-wallet

### DIFF
--- a/src/components/ProductTable/Filter.tsx
+++ b/src/components/ProductTable/Filter.tsx
@@ -102,7 +102,7 @@ const Filter = ({ filter, filterIndex, onChange }: FilterProps) => {
       className="bg-background-highlight p-6"
     >
       <AccordionTrigger className="border-b md:px-0">
-        <p className="text-base font-bold text-body">{filter.title}</p>
+        <p className="text-body text-base font-bold">{filter.title}</p>
       </AccordionTrigger>
       <AccordionContent className="p-0 md:p-0">
         <FieldSet className="gap-0">
@@ -120,7 +120,7 @@ const Filter = ({ filter, filterIndex, onChange }: FilterProps) => {
                   handleChange
                 )}
                 {item.inputState === true && item.options.length ? (
-                  <FieldSet className="flex flex-row gap-6 px-2 pb-4">
+                  <FieldSet className="flex flex-row flex-wrap gap-x-6 gap-y-2 px-2 pb-4">
                     {item.optionsLegend && (
                       <FieldLegend className="sr-only">
                         {item.optionsLegend}
--- a/src/components/ProductTable/Filter.tsx
+++ b/src/components/ProductTable/Filter.tsx
@@ -102,7 +102,7 @@ const Filter = ({ filter, filterIndex, onChange }: FilterProps) => {
       className="bg-background-highlight p-6"
     >
       <AccordionTrigger className="border-b md:px-0">
-        <p className="text-body text-base font-bold">{filter.title}</p>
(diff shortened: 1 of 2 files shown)


wackerow (approved):
Lgtm! Thanks @nloureiro

Small unrelated bug spotted while reviewing this:

Image

Need some space above the footer filter menu button on mobile

PR #18362, opened 2026-06-05
fix: prevent EF logo from stretching on foundation page

### LINKED ISSUE #18361: Foundation page: EF logo image stretched/distorted horizontally
## Description

The Ethereum Foundation logo on [/foundation](https://ethereum.org/foundation/) is stretched horizontally, causing visible distortion.

## Root cause

The `Logo` component (`src/components/Logo/index.tsx`) renders the logo with `className="h-[100px] w-auto"`. However, Next.js `<Image>` renders as `display: block`, which causes `width: auto` to resolve to the full container width (~768px) rather than the aspect-ratio-derived width. With the default `object-fit: fill`, the image is then stretched to fill 768×100px.

Confirmed via runtime inspection:
- Natural rendered size: 160×51px (800×259px source, scaled by Next.js)  
- Actual rendered size: **768×100px**
- Computed `object-fit`: `fill`

The source image `public/images/ef-logo.png` is also an 800×259px canvas, meaning the diamond logo only occupies a portion of the image — the wide canvas amplifies the distortion.

## Steps to reproduce

1. Go to https://ethereum.org/foundation/
2. Observe the EF logo below the page title — it appears squashed/flat instead of the correct diamond shape

## Expected behaviour

The EF logo renders at the correct aspect ratio (~100px tall, proportional width).

## Fix suggestion

In `src/components/Logo/index.tsx`, either:
- Change `w-auto` to `w-fit` (CSS `width: fit-content`) so the element shrinks to its intrinsic width, or
- Add `object-contain` and constrain the width explicitly, or
- Wrap the `<Image>` in an inline container so the block behaviour doesn't stretch it

Longer term, cropping `ef-logo.png` to the actual diamond bounds (removing extra canvas whitespace) would make the sizing more predictable.

nloureiro:
## Summary

Fixes #18361

The EF logo on [/foundation](https://ethereum.org/foundation/) was rendering stretched horizontally. The `Logo` component used `w-auto`, but Next.js `<Image>` renders as `display: block`, causing `width: auto` to fill the container (~768px) rather than computing from the aspect ratio. With the default `object-fit: fill`, the image was stretched to 768×100px.

**Fix:** `w-auto` → `w-fit` (`width: fit-content`), which shrinks the element to its intrinsic width instead of filling the container.

## Test plan
- [ ] Go to https://ethereum.org/foundation/
- [ ] The EF diamond logo below the page title should render at correct proportions (not squashed/flat)
- [ ] Verify in both light and dark mode

### DIFF
--- a/src/components/Logo/index.tsx
+++ b/src/components/Logo/index.tsx
@@ -14,7 +14,7 @@ const Logo = () => {
   return (
     <Image
       src={image}
-      className="h-[100px] w-auto"
+      className="h-[100px] w-fit"
       alt={t("ethereum-foundation-logo")}
     />
   )


wackerow (approved):
@nloureiro Lgtm.. not the root cause of why this happened (addressed in #18373), but this is still a reasonable fallback

PR #18363, opened 2026-06-05
patch(ui): text-body for all headers inside CardContent

wackerow:
## Description
- Expands text-body override to headers along with strong tags inside `CardContent` sub-component
- Fixes /bug-bounty/ MarkdownCards which contain sub-headers that were having text-body-medium applied

image

## Related Issue
Fixes:
image

### DIFF
--- a/src/components/ui/card.tsx
+++ b/src/components/ui/card.tsx
@@ -117,7 +117,7 @@ const CardContent = React.forwardRef<
     className={cn(
       childSpacingVariants({ spacing }),
       "flex-1 space-y-(--content-space) p-(--card-pad)",
-      "text-body-medium **:data-[label=card-title]:text-body **:[strong]:text-body",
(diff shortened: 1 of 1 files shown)


myelinated-wackerow (approved):
Approve. Expanding the `text-body` override from `strong` to `:is(h2,h3,h4,h5,h6,strong)` is the right call -- headers and strong tags inside `CardContent` should be full-contrast, not muted. Verified the `[:is(...)]` arbitrary-variant syntax already compiles elsewhere (QuizRadioGroup), and it's a pure color override on child elements so no specificity conflict. Site-wide blast radius is intentional; worth a quick Chromatic glance, but the change is correct and minimal.

_Reviewed by Claude Opus 4.8_

wackerow:
@pettinarip pulling in this small patch

PR #18369, opened 2026-06-07
refactor(ui): rounded-base implementation

myelinated-wackerow:
## Description

Introduces a single `--radius-base` design token as the canonical corner radius and migrates components to it, standardizing rounding across the site. Removes the redundant custom `--radius-4xl` override (Tailwind v4 already ships `4xl` = 2rem by default). Bundles in related cleanups to the list/card container pattern, RTL handling, and a few component refactors touched along the way.

## Changes

**Radius token**
- Add `--radius-base` (`= var(--radius-2xl)`, 1rem) in `theme.css`; remove the redundant custom `--radius-4xl` declaration.
- Migrate call sites from `rounded-2xl` (and assorted `rounded-xs`/`lg`/`3xl`) to `rounded-base`.
- Migrated `ui/alert` component to use `rounded-base` for consistency.
- Note: several `rounded-base` usages already existed on `dev` but were silently no-ops (the token did not exist yet); this PR is what makes them resolve.

**List / card containers**
- `CardList` and `Leaderboard` now render a single rounded, bordered, `overflow-hidden` container with `not-last:border-b` row dividers, replacing per-item borders and bottom margins.
- Rename `CardList`'s internal `Card` to `Row` to reflect its new role.
- The two translation-program leaderboards drop their no-op `text-foreground` / `bg-muted` classes as part of this.

**RTL cleanup**
- Drop the `useRtlFlip` hook from `CardList`, `DocLink`, `DocsNav`, `Leaderboard`, and `QuizWidget` in favor of inline `rtl:-scale-x-100` and logical (`inset-s`/`inset-e`) utilities. (Same direction as the RTL infrastructure work in #18384 -- these are two halves of one migration.)

**Related component refactors**
- `Leaderboard`: converted to an async server component (`getTranslations` instead of the client hook).
- `DocLink`: simplified to a single `Link` with hover scale/shadow; removed the `Emoji`/`Flex`/`Center`/`Stack`/`LinkBox` scaffolding.
- `DocsNav`: handles `null` doc data gracefully, uses `customEventOptions` for tracking, `fill` -> `stroke` hover, and modern Tailwind v4 class syntax (`m-0!`, `wrap-break-word`).
- `TranslationBanner`: restructured as a fixed-position `aside`.
- `StakingConsiderations`: new concave "fillet" tab corners via radial-gradient pseudo-elements driven by a local `--tab-r` property, with JS hover tracking so adjacent highlighted tabs merge without a seam.

**Knowledge base**
- Update the `design-system` skill to match: document `--radius-base` / `rounded-base` as the canonical corner radius (replacing the stale `--radius-4xl` custom-radius entry) and refresh the `rounded-2xl` example snippets.

## Notes
- No functional/content changes; visual + structural only.
- The `bug-bounty` severity-card refactor and the `wallets` seed-phrase tweak were moved out of this PR to keep it scoped to the radius token + component work -- they now live in the page-layouts PR (#18409).
- Recommend Chromatic review across the affected component stories.

## Related Issue
Design system


src/components/DocLink/index.tsx
```diff
@@ -1,56 +1,41 @@
...
-          )}
-        />
-      </Flex>
-    </LinkBox>
+      <ScrollText className="shrink-0 self-center" />
+      <p className="flex-1 font-semibold">{children}</p>
+      <ArrowRight
+        className={cn(
+          "mx-6 shrink-0 self-center",
+          "transition-tranform group-hover:transition-tranform group-hover:scale-120 group-hover:text-primary",
```
pettinarip: Typo: `transition-tranform` (missing the `s`, in both spots). Since it's not a real class, no `transition-property: transform` gets emitted and `transition-all` on the parent `<Link>` doesn't reach the arrow — so the hover `scale-120`/rotate snaps instead of animating. Should be `transition-transform`.

src/components/CardList.tsx
```diff
@@ -97,23 +92,28 @@ const CardList = ({
...
-          <Card {...listItem} imageWidth={imageWidth} />
+        <LinkBox key={id || idx} className={itemClasses}>
+          <Row {...listItem} imageWidth={imageWidth} />
         </LinkBox>
       ) : (
-        <div key={idx}>
-          <Card
+        <div key={idx} className={itemClasses}>
+          <Row
             onClick={() => {
```
pettinarip: Heads-up on dropping `"use client"`: when a server page renders `CardList`, this inline `onClick` on the non-link branch gets passed to a DOM element, which throws in an RSC. It's safe today only because the lists that reach here all have `link` set (and the filtered exchanges consumer is itself a client component), but it'd crash the first time a link-less item renders from a server page. `clickHandler`/`customEventOptions`/`trackCustomEvent` are also effectively dead now since no consumer passes them — might be cleanest to drop the click path entirely, or keep the component client.

pettinarip (approved):
Nice cleanup! Approving. I'll exec a manual chromatic tests just to see if there is any obvious major regression

src/components/CardList.tsx
wackerow: Good catch, agree! `clickHandler` not being used so I'll deprecate that and just pass `customEventOptions` through to `Row` and then `BaseLink` just in case we want to use that in the future, which allows us to remove the `onClick` entirely.

wackerow (approved):

PR #18377, opened 2026-06-08
Add Fastnode.io to dev docs /nodes-and-clients/nodes-as-a-service/

GeorgeWilstein:
Add RPC nodes provider Fastnode

## Description

Add RPC nodes provider Fastnode


nloureiro:
Thanks for the submission. Closing against the [listing criteria](https://ethereum.org/en/contributing/adding-developer-tools/): every entry on the nodes-as-a-service page carries a documentation link, and we couldn't locate developer docs for Fastnode (the site renders no content without JavaScript); the feature claims ("50+ blockchain networks", archive access) are unverifiable without them, and the feature list as written doesn't differentiate from the ~28 providers already listed. The diff also has formatting issues that would fail lint. A resubmission with a working docs URL and concrete, verifiable features would be reviewed on its merits.

PR #18158, opened 2026-05-09
a11y: ProductTable persona cards semantics

myelinated-wackerow:
## Summary

Refs #18114 (PR 2 of 2). Persona-card refactor on /wallets/find-wallet -- the remaining half of the accessibility/semantic-HTML pass.

- Persona grid is wrapped in `<FieldSet>` + sr-only `<FieldLegend>` ("Filter wallets by user type"). Each card is now a `<FieldLabel>` wrapping a sr-only Radix `<Checkbox>`; the description sits as a sibling `<FieldDescription>` referenced by `aria-describedby`, so screen readers announce the title as the accessible name and the description as the accessible description (no run-on).
- The `<h3>` inside each card is gone -- title becomes a styled `<span>`, fixing the document-outline jump.
- The visible "(123)" persona count is hidden from screen readers via `aria-hidden`; a sibling sr-only span renders an ICU-substituted string so locales can position the number per their own grammar (English: "{count} available").
- Visual layout, hover, focus, and click-anywhere-on-card behavior are unchanged.

## Test plan

- [ ] /wallets/find-wallet: persona grid renders inside a `<fieldset>` with sr-only `<legend>` "Filter wallets by user type".
- [ ] No `<h3>` inside any persona card; the document outline jumps from h1 directly to the h2 sr-only added in #18112, with no further headings inside the finder UI.
- [ ] Each persona is a checkbox: Tab reaches it, Space toggles it, focus ring renders around the visible card via `has-[:focus-visible]:`.
- [ ] Clicking the description area or any card padding still toggles the persona.
- [ ] Screen reader announces persona title as the accessible name and description as the accessible description; count is announced as "{count} available" instead of "(123)".
- [ ] Chromatic shows zero visual diffs.

Generated by Claude Opus 4.7

## Related issue
- Closes #18114

PR #18397, opened 2026-06-12
fix(e2e): update find-wallet persona filter locator

myelinated-wackerow:
## Description

The e2e suite currently fails on PRs into `staging`, blocking the v11.10.0 deploy CI (see #18393 / #18395). Root cause: #18158 rewrote the find-wallet persona cards from `<button>`s into `<FieldLabel>`s wrapping a visually hidden Radix checkbox (`role="checkbox"`), but the Playwright page object still targeted `getByRole("button", ...)`, so `find-wallet.spec.ts:18` times out on every browser project. The e2e job only runs for PRs into `master`/`staging`/`test/**`, so the mismatch never surfaced on the original dev PR.

Test-only change to `tests/e2e/pages/FindWalletPage.ts`:

- Locate the persona card `<label>` and click it -- native label forwarding toggles the checkbox, the same path a real user hits (see the comment in `PresetFilters.tsx` about label clicks forwarding to the checkbox)
- Assert the checkbox is actually checked after the click, which also guards against the double-toggle regression the component comment warns about
- Read the `(N)` counter from the card label text (the parenthesized count only exists in the visual span, so the regex stays unambiguous)

This fixes 4 of the 6 deploy-blocking e2e failures (`find-wallet.spec.ts:18` across all four projects). The two mobile-chrome-only failures at `:29`/`:46` (pointer interception / element instability) are a separate issue and not addressed here.

Note: this targets `dev`; to green the deploy CI it also needs to reach `staging` (cherry-pick or retarget as preferred).

### DIFF
--- a/tests/e2e/pages/FindWalletPage.ts
+++ b/tests/e2e/pages/FindWalletPage.ts
@@ -56,19 +56,27 @@ export class FindWalletPage extends BasePage {
    * Click on a persona filter and return the expected count
    */
   async selectPersonaFilter(personaName: string) {
-    const personaButton = this.presetFiltersContainer.getByRole("button", {
-      name: new RegExp(personaName, "i"),
+    const personaNamePattern = new RegExp(personaName, "i")
+
+    // Persona cards are <label>s wrapping a visually hidden checkbox
+    // (PresetFilters.tsx); clicking the label forwards to the checkbox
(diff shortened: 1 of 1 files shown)


pettinarip (approved):
Nice!

PR #18311, opened 2026-05-29
feat: add /stories page and individual story pages

pettinarip:
## Summary

Adds a `/stories` landing page that brings Ethereum human-interest stories together in one place, plus individual long-form story pages. Stories come in three forms, each its own section:

- **Discover** — long-form written stories at `/stories/[slug]`, rendered through the shared markdown/tutorial pipeline. The section is data-driven: card title, description and cover image are read straight from each story's frontmatter (no hardcoded lists or image imports).
- **More stories** — short video stories, sourced from the existing `community-stories` videos.
- **Community stories** — the short story cards reused from the 10years page, now with a toggleable category filter.

A contribution guide (`/contributing/adding-community-stories`) and `/stories` nav + footer links are included.

## Notable changes

- **Markdown video clips** — markdown can now embed silent, looping clips via ``. A new `MarkdownVideo` component sizes them by fixed aspect ratio (orientation from a `-portrait` filename suffix), plays only while on-screen, and respects `prefers-reduced-motion`. The image pipeline (`rehypeImg`) skips dimension-probing and blur placeholders for video, dropping the previous ffprobe dependency.
- **`StoryCard`** extracted into a reusable component (flip between English/original, optional read-more expander, optional date) shared by `/stories` and `/10years`.
- Shared helpers `getStorySlugs` / `getStoriesData` in `@/lib/utils/md`, new `Story.category` field and `StoryPreview` type.
- `TweetEmbed` component for inline tweet embeds in stories.

## Preview link
https://deploy-preview-18311.ethereum.it/stories

## Follow-ups

- Component file-naming refactor (separate PR): kebab-case all components + flatten single-file folders, in one sweep across the whole codebase ([discussion](https://github.com/ethereum/ethereum-org-website/pull/18311#discussion_r3456241888)).

PR #18405, opened 2026-06-12
Cleanup - Tighten video tag system to 1:1 shelf-key model

mnelsonBT:
Replaces the many-to-one video tagging system, where each gallery filter shelf matched against 5–18 secondary tags, with a strict 1:1 mapping where each shelf key is the only tag that maps to it.
- Brings unique tags across all videos down from 63 tags to 9 tags, and brings orphan tags (on videos but not mapped to any shelf) down from 17 to 0 

Changes:
- VIDEO_CATEGORIES in constants.ts: each shelf's tags array now contains only its own key (e.g. how-ethereum-works shelf matches only the how-ethereum-works tag)
Split privacy-and-security into separate privacy and security shelves
- Added events shelf (downstream of re-categorization done in https://github.com/ethereum/ethereum-org-website/pull/18311 ) 
- Updated topic: frontmatter on all 58 existing videos to use shelf-key tags only
- Updated contributing guide (adding-videos/index.md) Step 4 with a clean tag↔filter table
- Updated i18n keys in page-videos.json

Rationale: 
The old system had many ad-hoc tags (consensus, blockchain, defi, rollups, authentication, etc.) that were only used to loosely bucket videos into shelves. Many tags existed on videos with no corresponding shelf at all. When working on PR #18311 I noticed that this mess  had masked that some videos were lacking filter views entirely, some were miscategorized, etc. 
 This cleanup makes the system self-documenting. Contributors only need to know the shelf names, not memorize secondary tags. 

To address the original goal of 1:many shelf key config: If we want more dynamic tag-driven features in the future (e.g. pulling videos by topic into "learn more" carousels on educational pages), we can introduce new tags intentionally based on those filter keys rather than inheriting a sprawl of unused tags and working backward to build filters from there.


wackerow (approved):
Looks good... There will be conflicts with #18311 after this but completely manageable. 

non-English versions will be updated separately with intl-pipeline

PR #18414, opened 2026-06-16
refactor(staking): source staking stats from Dune, replace validator count with % staked

pettinarip:
## Summary

Beaconcha.in retired its free API keys (May 2025), which broke the staking page's data fetch (`fetchBeaconChain`). This migrates all staking stats to **Dune Analytics** and reorganizes the data-layer fetchers around it.

The `/staking` stats box now shows: **Total ETH staked** · **Percent of ETH staked** · **Current APR**.

## What changed

**Data layer**
- New `fetchDune.ts` shared helper (`fetchDuneCounter`) — reads a single-value Dune query via the free cached results endpoint.
- New `fetchStakedPercentage.ts` — `% of ETH staked` (Dune query `1933048`, normalized to a fraction).
- `fetchTotalEthStaked.ts` now reads **net staked ETH** (deposits − full withdrawals, Dune query `1933035`) instead of cumulative deposits — the same flows the % stat is built on, so the two stats stay consistent.
- Removed `fetchBeaconChain.ts`, its mock, and the dead beaconcha-only types (`BeaconChainData`, `BeaconchainEpochData`, `EpochResponse`, `EthStoreResponse`, `EthStakedResponse`) + the `BEACONCHAIN` key/getters.

**Staking page / UI**
- Replaced the **Total validators** stat with **Percent of ETH staked**. The old validator count was really `net stake ÷ 32`, which is a misleading headcount post-Pectra (EIP-7251 raised max effective balance to 2048 ETH).
- APR is now **derived** from total staked via `computeStakingApr` (consensus issuance formula) — no external APR feed. Tracks the old ETH.STORE rate within ~0.05pp.
- Total ETH staked now renders as an integer (no decimals).
- Stats tooltips + homepage stat attribution updated to credit **Dune Analytics**.

## Notes / follow-ups
- The Dune queries currently point at hildobby's public dashboard queries. Before relying on this long-term, fork them into the ethereum.org Dune workspace and put them on a refresh schedule (the API only returns the last execution).
- New `fetch-staked-percentage` blob must be populated (daily task) before the page renders live data; it fails loud if missing.

## Testing
- `tsc` passes clean.
- Updated the data-layer getters unit test (`getBeaconchainData` → `getStakedPercentageData`).


src/lib/utils/staking.ts
wackerow, line 6: Should we worry about leap seconds?? 😆 (https://en.wikipedia.org/wiki/Leap_second)

wackerow, line 8: No strong preference, and fine with me if we keep it here, but just wondering if we should move these to a `constants.ts` file

wackerow (approved):
Nice! Thanks @pettinarip.. left one small question but otherwise looks good

src/lib/utils/staking.ts
pettinarip: hmm these are only used by the apr calc and are meaningless without it. If we see in the future that we need to reuse some of these constants, then it would make more sense to move. For now, I'd keep them colocated.

PR #18167, opened 2026-05-11
Fix horizontal page overflow from EdgeScrollContainer

pettinarip:
## Summary

- Fixes #17777: `/community/events` had ~2300px of horizontal page overflow on mobile (and other viewports).
- Root cause: the inner `overflow-x: auto` flex scroll area inside `EdgeScrollContainer` was leaking into the document's scroll width despite ancestors having `overflow: visible`. The 8-card hubs instance had enough internal scroll width to make the leak user-visible; smaller instances on `/developers/tools`, `/developers`, and `/community/events/conferences` stayed below the threshold but were similarly affected.
- Fix: add `relative overflow-x-clip` on the `EdgeScrollContainer` outer wrapper. `position: relative` is required — `overflow-x: clip` alone does not contain the leak in Chromium (it prevents `window.scrollTo` from moving but real wheel/touch scrolling still produces overflow).

## Test plan

- [x] Mobile (375px): `/community/events` no longer scrolls horizontally
- [x] `document.documentElement.scrollWidth === clientWidth` after load
- [x] Hubs cards still scroll horizontally inside the container
- [x] Card shadows / vertical padding still visible (no clipping of `overflow-y`)
- [x] No regression on `/developers/tools`, `/developers`, `/community/events/conferences`
- [x] Visual: deploy preview screenshot of hubs section looks correct

PR #18417, opened 2026-06-16
fix: prevent find-wallet persona cards from causing horizontal page scroll on mobile

pettinarip:
### Description

A visual regression on **`/wallets/find-wallet/`** (mobile) shipped in v11.10.0 (#18393), detected by Chromatic — the persona filter row makes the whole page scroll horizontally.

### Root cause

The persona-card row is an intentional horizontal-scroll strip (`grid-flow-col` + `overflow-x-auto`). Two issues let it leak into the **document** scroll width on mobile:

1. **`<fieldset>` min-content (the Chromatic flag).** #18158 (`a11y: ProductTable persona cards semantics`) wrapped the strip in a `<fieldset>` (`FieldSet`). A `<fieldset>` has a UA-default `min-width: min-content`, so it refused to shrink below its content width (~1096px), pushing the document wider than the viewport.
2. **Inner scroll-area leak (what reviewers still saw on the preview).** Even once the fieldset was constrained, the inner `overflow-x-auto` strip leaked its scroll width into the document — a real, draggable horizontal page scroll on WebKit/iOS Safari.

### Fix

`relative min-w-0 overflow-x-clip` on the `FieldSet`:

```diff
-    <FieldSet className="gap-0">
+    <FieldSet className="relative min-w-0 gap-0 overflow-x-clip">
```

- `min-w-0` — overrides the `<fieldset>` `min-width: min-content` so it shrinks to the viewport width.
- `relative overflow-x-clip` — contains the inner scroll area so it can't leak into page scroll width. This is the **same pattern as `EdgeScrollContainer`** in #18167; `position: relative` is required for `overflow-x: clip` to contain the leak in Chromium.

The persona strip still scrolls horizontally **within itself** — only the document-level overflow is removed. Card shadows / vertical padding are unaffected (`overflow-y` stays visible).

### Verification

- `document.documentElement.scrollWidth === clientWidth` after load ✅
- `window.scrollX` stays `0` after `scrollBy({ left: 5000 })` ✅
- Persona strip still scrolls horizontally internally ✅
- Card shadows / layout visually unchanged vs. baseline ✅
- Added `/wallets/find-wallet/` to `tests/e2e/horizontal-overflow.spec.ts` (the #18167 regression suite) — passes on `e2e-mobile-chrome`.

### DIFF
--- a/src/components/ProductTable/PresetFilters.tsx
+++ b/src/components/ProductTable/PresetFilters.tsx
@@ -241,7 +241,7 @@ const PresetFilters = ({
   )
 
   return (
-    <FieldSet className="gap-0">
+    <FieldSet className="relative min-w-0 gap-0 overflow-x-clip">
       <FieldLegend className="sr-only">
         {t("page-find-wallet-persona-legend")}
       </FieldLegend>
--- a/tests/e2e/horizontal-overflow.spec.ts
+++ b/tests/e2e/horizontal-overflow.spec.ts
@@ -3,21 +3,24 @@ import { expect, test } from "@playwright/test"
 /**
  * Regression test for https://github.com/ethereum/ethereum-org-website/issues/17777
  *
- * Pages that use EdgeScrollContainer must not leak horizontal overflow to the
+ * Pages with inner horizontal scroll areas (EdgeScrollContainer, the
+ * find-wallet persona strip, etc.) must not leak horizontal overflow into the
  * document scroll area. Asserting on both `scrollWidth` and an actual
  * `scrollBy` is intentional: each catches a different class of regression
  * (layout vs. containment).
  */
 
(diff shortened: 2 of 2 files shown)


src/components/ProductTable/PresetFilters.tsx
```diff
@@ -241,7 +241,7 @@ const PresetFilters = ({
   )
 
   return (
-    <FieldSet className="gap-0">
+    <FieldSet className="relative min-w-0 gap-0 overflow-x-clip">
```
wackerow, line 244: Image

Much better than in production, but noting a smaller bug here where the overflow-x-clip usage chops off the shadow -- Imo an acceptable trade-off for now to fix the main bug here

wackerow (approved):

Issue #18388, opened 2026-06-10
Suggestion: Add GitHub issue templates for new content types


mnelsonBT:
PR https://github.com/ethereum/ethereum-org-website/pull/18387 adds contribution guidelines for the new content page types:
- reports
- builder articles 

It would be nice to add new issue templates in GitHub for these new content types. 
The below suggestions reference the format of the [suggest a video issue template](https://github.com/ethereum/ethereum-org-website/issues/new?template=suggest_video.yaml).

Reports template - suggested fields to include: 
- Link to report .pdf 
- Publisher name
- Original publish date 
- Title 
- Why should this report be listed? _Briefly explain how this report helps ethereum.org users learn about Ethereum_ 

Builder article template - suggested fields to include: 
- Author (name) and team (optional) 
- Proposed article title 
- Proposed article topic 
- Brief description of the article _1–3 sentences summarizing what the article will cover_
- Language _Which language will the article be written in?_
- Why should this article be published? _Briefly explain how this article helps Ethereum builders or the broader community understand today's landscape_

PR #18425, opened 2026-06-18
feat(latest): redesign /latest into a community updates hub

pettinarip:
## Summary

Redesigns `/latest` from a flat list of builder blog posts into a **"Latest Ethereum updates"** hub that merges first-party builder articles with a curated community RSS feed.

### Page
- **Hero** with a shared CTA pair (Submit a resource / Propose an article)
- **Highlights** — editorial cards
- **Latest Ethereum articles** — filterable grid (category facets + topic tags) with load-more
- **Read more on these websites** — source directory with per-source logos
- **Suggest a resource** CTA banner

Highlights and the grid share one `LatestCard` (whole-card link), composed from design-system card primitives.

### Data
- Unified `LatestArticle` model (builder posts + RSS), sorted newest-first; highlights de-duped from the grid by href
- `LATEST_SOURCES` — 18 curated feeds, the single source of truth for both RSS ingestion and the website directory — with per-source category facets, `<category>` filtering (Besu), and a link rewrite for sources that moved domains (Vitalik)
- `fetchRSS`: trailing 180-day window, plain-text excerpt + cover-image extraction; per-source display cap applied at the consumer

### i18n
- RSS shown in English across all locales with a section-level disclaimer in non-EN; copy keys in `page-latest.json` (EN only — the intl pipeline propagates)

### Deferred
- `Submit a resource` links to the GitHub issue chooser as an interim target — swap to the news-source contributing guide once it lands (currently a Google Doc; see #18388)
- Per-locale translated feeds

## Before merge
- [x] Simplify `src/data-layer/mocks/fetch-rss.json` — the current mock is extensive for testing purposes; trim it to a shorter, lighter fixture before merging

## Test plan
- [x] `pnpm type-check` + ESLint pass
- [x] `/latest` renders all sections; filters and load-more work; grid degrades to builder-only when the RSS cache is empty

## Preview link
https://deploy-preview-18425.ethereum.it/latest


pettinarip:
- [x] **TODO (depends on #18472):** once the shared `TagFilter` component lands on `dev`, refactor this page's tag filter to reuse it (`@/components/ui/tag-filter`) instead of the bespoke filtering logic — multi-select with **OR** match semantics (matching done here in the parent), per the extraction plan.

app/[locale]/latest/_components/LatestArticlesGrid.tsx
```diff
@@ -0,0 +1,195 @@
...
+                  {article.timeToRead ? (
+                    <>
+                      {" · "}
+                      {t("page-latest:page-latest-minute-read", {
+                        minutes: article.timeToRead,
+                      })}
+                    </>
+                  ) : null}
+                </>
+              ) : undefined
```
Comment (wackerow) [inline app/[locale]/latest/_components/LatestArticlesGrid.tsx]: Was going to suggest using the "tags inline text" component here..

```suggestion
                <TagsInlineText
                  list={[
                    formatDate(article.date),
                    article.timeToRead &&
                      t("page-latest:page-latest-minute-read", {
                        minutes: article.timeToRead,
                      }),
                  ]}
                />
```

..but realizing a _tiny_ bug with this component you'd have to patch as well, just changing `const renderList = list.filter(String)` to `const renderList = list.filter(Boolean)` (the `String` filter isn't working as expected.. Boolean more appropriate). Not really a big deal though.

But with that, it'd handle the "undefined" cases and just render a bullet-separated list of whatever values are available. `[undefined, undefined]` would just render null.

wackerow (approved):
@pettinarip Looks good!

I did a quick test locally to see how a new internal post would show up, setting the date to yesterday, but I'm not seeing it show up.. assuming perhaps something with the data-layer? Not necessarily a blocker, just caught my attention.

Also, the "Highlighted" posts at the top, in particular the `latest/privacy-apps-on-ethereum` entry.. I don't see the "time to read" on the card -- should it be there?

Nothing blocking here though

pettinarip:
@wackerow just tested locally and adding a post works. The tricky part is that you need to add the slug into a special file under `src/data/blogPosts.json` which is a setup that we already had. I think we could probably get rid of that file and read everything that exists under `/latest`. I'll check if that refactor can be done safely and will push it. It will make it easier to maintain.

wackerow:
> @wackerow just tested locally and adding a post works. The tricky part is that you need to add the slug into a special file under `src/data/blogPosts.json` which is a setup that we already had. I think we could probably get rid of that file and read everything that exists under `/latest`. I'll check if that refactor can be done safely and will push it. It will make it easier to maintain.

Ah that'll do it. Yeah, certainly doesn't need to go in this PR, but thanks for the update

wackerow (approved):
🚀

PR #18427, opened 2026-06-18
refactor: start page to latest design system

myelinated-wackerow:
## Description

Refactors the `/start` page and its `StartWithEthereumFlow` sub-components to align with the latest design system, co-locates the components with the route, and fixes several layout/a11y bugs surfaced during the cleanup.

## Changes

**Structure**
- Moved `StartWithEthereumFlow` (and its `ConnectYourWallet`, `DownloadAWallet`, `LetUseSomeApps`, `ShareModal` children) from `src/components/` into `app/[locale]/start/_components/`, co-located with the only route that uses them.

**Spacing / layout**
- Adopted the responsive `--spacing-page` token (`px-page`: 1rem mobile / 2rem desktop) for the page's horizontal padding.
- Fixed the Swiper full-bleed breakout: the hardcoded `-mx-8` / `px-4 sm:px-8` assumed a fixed 2rem page padding and broke once `px-page` became responsive (zero left padding below `sm`, excess right-side space below `md`). Now uses `-mx-page` + `px-page` so the breakout always cancels the parent padding exactly at every breakpoint.
- Fixed the active Swiper card clipping its bottom padding: `getStyleFromIndex` was applying a `maxHeight` derived from the slide's border-box `offsetHeight` (padding included) to the inner content, so tall content (the first card's wallet list) overflowed past the content box and got clipped at the bottom. The active slide now renders at natural height; only inactive slides keep the clamp for the peek-behind effect.
- Start-aligned the "Sign in with Ethereum" CTA on the second card (was centered via `items-center`).

**Design-system / best-practice cleanup**
- Replaced a hard-coded `"Explore more"` string with the existing `page-start-apps-explore-more` translation key.
- Typed the Swiper event handlers (`SwiperType`) instead of implicit `any`.
- Standardized `rounded-xl` to the `rounded-base` token.
- Hardened external `window.open` calls with `"noopener,noreferrer"` (Twitter share + dapp links).

**Accessibility**
- Replaced the `<div onClick>` "I have a wallet" toggle rows with a proper `Checkbox` + `<label htmlFor>` association (toggle/tracking moved to `onCheckedChange`), making them keyboard-operable and screen-reader-announced. Distinct ids per breakpoint copy.

## Testing
- [ ] Verify `/start` hero + intro padding at mobile / `sm` / `md` / desktop widths
- [ ] Step through all three Swiper cards; confirm symmetric card padding and the peek-behind effect on cards 2 and 3
- [ ] Confirm the "Sign in with Ethereum" button is start-aligned on desktop
- [ ] Toggle the "I have a wallet" checkbox via keyboard (Tab + Space) and by clicking its label
- [ ] Confirm share modal + dapp links still open correctly

## Preview link
https://deploy-preview-18427.ethereum.it/start

## Related task
Design system


app/[locale]/start/page.tsx
```diff
@@ -48,48 +50,35 @@ const Page = async (props: { params: Promise<PageParams> }) => {
...
+        className="mb-space-3x h-60 w-full object-cover md:h-96"
+        preload
+      />
+      <div className="flow mx-auto max-w-3xl px-page text-center">
+        <h1>{t("page-start-title")}</h1>
+        <p>{t("page-start-subtitle")}</p>
+      </div>
+
+      <I18nProvider locale={locale} messages={messages}>
+        <MainArticle className="space-y-space-2x overflow-x-hidden px-page py-page-2x">
```
Comment (pettinarip) [inline app/[locale]/start/page.tsx:68]: obeservations about not having the h1 inside the main article, not blocking:
- algolia and semantics related, h1 might be missing as it is missplaced
- "skip to content" button jumps to this container and miss the h1 from the viewport

should we rework this part?

pettinarip (approved):
LGTM @myelinated-wackerow left a question that might also impact other pages we are refactoring

app/[locale]/start/page.tsx
Comment (wackerow) [reply app/[locale]/start/page.tsx:68]: Hm.. personally kinda feel like that's preferred, but it's a good question that reaches well beyond this page. Effectively all of our pages have the h1 in a hero, which comes before the MainArticle. For now I would say to keep as-is, but maybe worth a discussion.

cc: @mnelsonBT - thoughts on the approach we're taking here? It would be a significant overhaul to change this throughout the site (moving the hero's inside the main `article` element)

---

I'm also noting a small bug with that "skip to content" link -- my keyboard focus stays in the nav bar, which is not the desired behavior. It's supposed to move keyboard focus so the user "skips" having to tab through all the nav elements.. anyone else experiencing this? Out-of-scope for this, but can address it separately.

Comment (mnelsonBT) [reply app/[locale]/start/page.tsx:68]: Just to follow-up here on chat with @wackerow - agree with keeping as-is for now since the pages already had the H1 in the hero. If we see any new issue flags for missing H1s / other negative site health impact after deploy then we can update, but it sounds like this config was intentional for the keyboard tabs.

PR #18440, opened 2026-06-20
refactor(ui): ContentLayout asidePosition variant + content-page primitives

myelinated-wackerow:
## Description

Infrastructure for the content/article page refactors: expands `ContentLayout` with an `asidePosition` variant and tidies the primitives those pages rely on. No page is switched over here -- these are the shared building blocks (the page swaps land in the follow-up PR).

- **`ContentLayout` -- `asidePosition`.** New `asidePosition?: "left-bottom" | "right-top"` (+ optional `listenSlug`). `"left-bottom"` (default) is the existing shape -- left TOC rail, contributors at the article foot -- so every current consumer (`TopicLayout`, `/learn/`, `/staking/`, `/use-cases/`) is unchanged. `"right-top"` is the new standalone-article shape: card TOC in a right-hand `<aside>`, a compact contributors + optional `ListenToPlayer` byline up top, capped reading width.
- **`FileContributors`.** Flattened the nested `Flex` wrappers to one element; added a `variant` (`base` / `compact`) so callers stop overriding padding with `!`/`[&>div]` hacks; dropped the now-unused `asChild`.
- **`ListenToPlayer` / `TopOfPagePlayer`.** Threads an optional `className` through to the player button (position passed in, not baked).
- **`PageActions`.** a11y: `role="toolbar"` + `aria-label`.
- **`TableOfContents` / `Topic.tsx` / `common.json`.** Minor supporting tweaks.
- **Skill docs.** Documents `asidePosition` and corrects the `ContentLayout` consumer list in `layouts.md`.

## Related Issue

Design system

---

Reviewer note: `left-bottom` consumers (`TopicLayout` -> roadmap/staking/upgrade/use-cases, plus `/learn/`) warrant a visual pass -- `ContentLayout`'s internal structure changed even though the default shape didn't.


src/layouts/ContentLayout.tsx
```diff
@@ -4,19 +4,28 @@ import { FileContributor } from "@/lib/types"
...
     children: React.ReactNode
     tocItems: TableOfContentsProps["items"]
     heroSection: React.ReactNode
     contributors: FileContributor[]
     lastEditLocaleTimestamp?: string
+    /**
+     * `"left-bottom"` — Contributors at the foot of the article, TOC rail on the left.
+     * `"right-top"` — Compact contributors + listen byline up top, card TOC on the right.
+     */
+    asidePosition?: "left-bottom" | "right-top"
```
pettinarip, line 27: my take on `asidePosition`: it describes where things sit, but the prop is really picking a whole page persona: it flips the TOC between rail and card, caps the article width, and moves the contributors/listen byline.

So I'd lean toward a semantic name like `variant="hub"|"article"` (or `kind`). Hides the layout mechanics, and a future third arrangement won't turn the name into a lie.

Fine as-is too, but since nothing uses `right-top` yet, now's the cheap moment to switch.

pettinarip (approved):
@myelinated-wackerow Left an opinion on the new prop name but overall it looks good

src/layouts/ContentLayout.tsx
wackerow: Yeah I guess couple thoughts here

"hub" vs "article" -- they aren't necessarily hubs (in the way we've used that term elsewhere).. It's primarily the "Topic" pages.. and they're _all_ articles. That was part of the thoughts process here.

But also, it sounds like there are some design considerations in-flight to be decided soon as to the path forward with migrating the Topic pages to use the newer approach, which would effectively put a short-lived expiration date on this entire prop.

pettinarip: Yeah thats fine

myelinated-wackerow:
Closing in favor of #18468

PR #18458, opened 2026-06-21
fix API3 DAO link

### LINKED ISSUE #18457: API3 DAO link uses non-resolving www.api3.org host
## Description

The oracles documentation links to `https://www.api3.org/` for API3 DAO, but the `www` host no longer resolves. The root domain `https://api3.org/` is accessible, so this appears to be a straightforward URL update.

Affected source file:

- `public/content/developers/docs/oracles/index.md`

## Example I verified

| Source file / context | URL | Observed result |
| --- | --- | --- |
| `oracles/index.md` line 405 | `https://www.api3.org/` | DNS failure: could not resolve host |

I also checked the likely replacement:

```bash
curl -L -I https://api3.org/
```

Observed result: `HTTP 200`.

## Reproduction

```bash
curl -L -I https://www.api3.org/
```

Observed result: `Could not resolve host: www.api3.org`.

## Expected behavior

The API3 DAO link should point to a resolving API3 URL.

## Suggested fix

Change `https://www.api3.org/` to `https://api3.org/`. Translated copies also include the same URL and may need refreshing.

alerodriargui:
## Description

Updates the broken API3 DAO link in the canonical English oracles documentation from `https://www.api3.org/` to `https://api3.org/`. The replacement URL was verified to return `200 OK`.

## Related Issue

Fixes #18457


nloureiro (dismissed):
Verified ✅ — `www.api3.org` no longer resolves; `api3.org` returns 200.

---

myelinated-wackerow (approved):

wackerow (approved):

PR #18469, opened 2026-06-22
fix(intl-pipeline): support blank/root TARGET_PATH for full-tree runs

myelinated-wackerow:
## Problem

Running the intl pipeline with a blank `TARGET_PATH` — documented in the workflow input as *"blank for all"* — errors with `TARGET_PATH is required`. Passing the source roots explicitly (`public/content,src/intl/en`) also fails: `normalizeTargetPath` mangles them into `public/content/public/content` and `public/content/src/intl/en`, which then fail the on-disk existence check.

## Fixes

**`main.ts`**
- A blank `TARGET_PATH` now defaults to the two source roots (`config.jsonRoot`, `config.mdRoot`) instead of erroring; the existing directory-expansion pass resolves them to every source file.
- `walkForExt` skips the `translations/` subtree, so expanding `public/content` doesn't pull in translated files (which `validateTargetPath` rejects).

**`config.ts`**
- `normalizeTargetPath` now leaves paths already at/under a source root untouched — including the bare roots `public/content` (no trailing slash) and `src/intl/en` (a non-`.json` directory that was misclassified as markdown). Only bare relative paths get prefixed.

## Result

Blank **and** explicit-root runs both resolve to **62 JSON namespaces + 324 markdown files** (329 English minus the 5 `DO_NOT_TRANSLATE` legal pages), with no `translations/` leakage.

## Verification

- `normalizeTargetPath` checked across 8 cases (bare roots, rooted files, bare relative files, `translations/` strip, non-en strip) — all pass.
- Full-tree expansion replicated on the real dirs → 386 source files, zero `translations/` entries.
- eslint clean on both files.

### DIFF
--- a/src/scripts/intl-pipeline/config.ts
+++ b/src/scripts/intl-pipeline/config.ts
@@ -173,11 +173,18 @@ export function normalizeTargetPath(
     p = `src/intl/en/${intlMatch[2]}`
   }
 
-  // 3. Auto-prefix by file type
-  const isJson = p.endsWith(".json")
-  if (isJson && !p.startsWith("src/intl/en/")) {
+  // 3. Auto-prefix by file type. Paths already at/under a source root (the
+  //    bare roots "src/intl/en" and "public/content" included) are left as-is;
+  //    only bare relative paths get prefixed. Note "public/content" has no
+  //    trailing slash, so the root itself must be matched explicitly.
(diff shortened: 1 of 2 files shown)


wackerow (approved):
Head's up @pettinarip, pulling in a small patch to allow us to run the intl-pipeline without any arguments so it processes all files.

All files with manifests that match the latest English will be skipped, so this should catch everything that remains, and surface any remaining bugs we'll need to work through

PR #18472, opened 2026-06-23
feat(ui): extract reusable TagFilter component

pettinarip:
## Summary

Extracts the tag-filtering logic from `/developers/tutorials` into a reusable, controlled `TagFilter` component plus a generic `getTagCounts` util. This is **PR 1 of 2** — a follow-up PR will adopt `TagFilter` on the new `/latest` page (with OR match semantics).

## What's new

- **`src/lib/utils/tags.ts`** — generic `getTagCounts<T>(items, getTags)` returning count-descending `[name, count]` entries. No tag normalization (each data source owns its own hygiene).
- **`src/components/ui/TagFilter/`** — controlled, presentational multi-select chip filter:
  - `value` / `onChange` controlled selection; match semantics (AND/OR) stay in the parent.
  - `defaultVisible` cutoff with a show-more/less expander; selected tags beyond the cutoff stay pinned-visible while collapsed so an active filter never disappears.
  - `showCount` (default true); counts rendered via `numberFormat(locale)` (honours Urdu/Arabic numbering rules).
  - Expander chrome reads `show-more` / `show-less` from the `common` namespace; `aria-pressed` on chips, `aria-expanded` on the expander.
- **`TagFilter.stories.tsx`** — Storybook coverage (collapsed, expanded, with-selection, active-hidden-tag, no-count).

## Refactor

- **`tutorials.tsx`** — replaces the inline `FilterTag` + popular/niche tag-splitting with `<TagFilter>`. The skill + tags + search matching `useEffect` is unchanged (AND semantics preserved). Matomo tracking moved into a single-delta `onChange` handler.
- **`tutorial.ts`** — removes the now-unused `getSortedTutorialTagsForLang` (no remaining callers).
- **`page-developers-tutorials.json`** — drops the orphaned `page-tutorial-more-tags` key. The expander now reads "Show more (N)" (from `common`) instead of "+N more". Non-English locales will be reconciled by the intl-pipeline.

## Verification

- `pnpm type-check` ✓ · `pnpm lint` ✓
- Storybook render verified for all states (count-desc ordering, cutoff + correct hidden count, active styling, pinned-visible selection).
- Note: the live tutorials page can't be rendered locally (requires Netlify Blobs secrets); page-level visual coverage is left to Chromatic on this PR.

### DIFF
--- a/.claude/skills/design-system/references/components.md
+++ b/.claude/skills/design-system/references/components.md
@@ -391,6 +391,25 @@ Big variant matrix.
 **`variant`**: `subtle | high-contrast | solid | outline`
 **`size`**: `small | medium`
 
+### `TagFilter`
+
+```tsx
+import TagFilter from "@/components/ui/tag-filter"
+```
+
+Controlled, presentational multi-select chip filter built on `TagButton` -- a wrapping chip row with an optional show-more/show-less expander. **Reach for this instead of hand-rolling a `TagButton` row** whenever a page filters a list by tags. Selection and match semantics (AND vs OR) live in the parent; the component only renders chips and reports toggles.
+
+```tsx
+<TagFilter
+  tags={getTagCounts(items, (i) => i.tags)} // [name, count][], caller pre-sorts/filters
+  value={selectedTags}
+  onChange={setSelectedTags}
+  defaultVisible={12} // chips before the expander; selected-but-hidden tags stay pinned-visible
+/>
+```
+
+**Props**: `tags` (`[name, count][]`, rendered as-is), `value` / `onChange` (controlled), `defaultVisible` (cutoff; omit to show all), `showCount` (default `true`, formats via `numberFormat(locale)`), `className`. Pair with `getTagCounts` from `@/lib/utils/tags` to build count-descending entries from any item list.
+
 ### `Alert`
 
 ```tsx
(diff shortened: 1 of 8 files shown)


src/components/ui/tag-filter.tsx
```diff
@@ -0,0 +1,119 @@
...
+          aria-expanded={expanded}
+          className="inline-flex items-center gap-1 rounded-full border border-dashed border-body-medium px-3 py-0.5 text-xs text-body-medium uppercase transition-colors hover:border-primary hover:text-primary"
+        >
+          {expanded ? (
+            <>
+              {t("show-less")} <ChevronUp className="size-3" />
+            </>
+          ) : (
+            <>
+              {t("show-more")} ({formatCount(tags.length - visible.length)}){" "}
```
wackerow, line 109: Pushed one commit to patch this tiny logic bug with the "Show more (n)" count... If you expanded, selected a tag, then "Show less", the selected tag remains visible but still show "Show more (n)"... using `visible.length` fixes so it displays "Show more (n - 1)" (if one previously-hidden tag is selected).

.claude/skills/design-system/references/components.md
wackerow, line 1: Had Claude update the design-system Skill docs

wackerow (approved):
@pettinarip Looks good! Added two small commits, details below -- pulling in!

PR #18474, opened 2026-06-23
add new-page skill

pettinarip:
## Summary

Adds two user-invoked skills for building ethereum.org pages, reuse-first against the design system. Draft — opening for feedback.

- **`/new-page`** — interview → scaffold → QA. Forks first on **markdown vs App Router** (an authoring decision, asked upfront), then picks the layout/hero, maps the Figma design to existing components, and builds. Key gate: an **exhaustive component-map table** shown for user sign-off before any code.
- **`/review-page`** — adversarial review loop (design-system + copy reviewers) used as `/new-page`'s QA pass.

## Notes

- Scope: **only** `.claude/skills/{new-page,review-page}/`. No app code — pages built while testing the skills live on separate branches.
- User-invoked (`disable-model-invocation: true`).


wackerow (approved):

wackerow:
Let's pull in and we can iterate as it's utilized

PR #18486, opened 2026-06-24
fix broken ZK-Kit link

### LINKED ISSUE #18485: Broken ZK-Kit link on privacy page
### Describe the bug

The ZK-Kit resource on the privacy page points to `https://zkkit.pse.dev/`, but that domain no longer resolves.

The same outdated URL is present in translated copies of the page.

### To reproduce

1. Go to https://ethereum.org/privacy/
2. Find **Zk-kit** in the privacy resources list.
3. Open the link.
4. See that the destination cannot be reached.

### Expected behavior

The resource should open the current ZK-Kit website at:

https://zkkit.org/

### Additional context

The outdated URL appears in `public/content/privacy/index.md` and 11 translated copies.

Yes

alerodriargui:
## Description

Replaces the broken ZK-Kit URL with the current official website on the canonical English privacy page.

## Related Issue

Fixes #18485


nloureiro (dismissed):
Verified ✅ — `zkkit.pse.dev` no longer resolves; `zkkit.org` returns 200.

---

myelinated-wackerow (approved):

wackerow (approved):

PR #18492, opened 2026-06-24
remove dead Pectra.wtf links

### LINKED ISSUE #18491: Dead Pectra.wtf links across Pectra pages
### Describe the bug

The `https://pectra.wtf` domain no longer resolves, leaving broken resource links on the Ethereum forks and Pectra roadmap pages.

The same dead links are also present in translated copies of both pages.

### To reproduce

1. Go to https://ethereum.org/ethereum-forks/ or https://ethereum.org/roadmap/pectra/
2. Find the **Pectra.wtf** resource link.
3. Open the link.
4. See that the domain cannot be reached.

### Expected behavior

The dead `Pectra.wtf` resource entries should be removed. Both pages already include other active resources about the Pectra upgrade, so a replacement is not necessary.

### Additional context

The dead URL appears in:

- `public/content/ethereum-forks/index.md`
- `public/content/roadmap/pectra/index.md`
- 32 translated copies of these pages

Yes

alerodriargui:
## Description

Removes dead Pectra.wtf resource links from the canonical English Ethereum forks and Pectra roadmap pages.

## Related Issue

Fixes #18491


nloureiro (dismissed):
Verified ✅ — `pectra.wtf` is fully dead (no DNS/connection), and there's no live replacement, so removing the two dead resource bullets is the right call.

---

myelinated-wackerow (approved):

wackerow (approved):

PR #18500, opened 2026-06-24
fix(a11y): move focus to main content on skip link

myelinated-wackerow:
## Description

Fixes the skip-to-main-content link so it actually moves keyboard focus, not just scroll position.

Previously, activating the skip link scrolled the page to the main content but left keyboard focus stranded in the nav bar, so a keyboard user still had to tab through every nav item to reach the content. Two root causes:

1. The target `<article id="main-content">` was not focusable, so the browser's fragment navigation could scroll to it but could not move focus there.
2. The link routed through `next-intl`'s `Link`, which pushed `#main-content` into the URL (where it could stack with later hash links), and the focus-induced scroll ignored `scroll-margin-top`, leaving the heading clipped behind the fixed nav.

## Changes

- **`MainArticle`** is now focusable via `tabIndex={-1}` (with `outline-none` to suppress the ring on the whole region) so it can receive focus as a skip target.
- **`SkipLink`** is reworked to treat activation as a focus action instead of navigation: it `preventDefault`s the fragment nav, moves focus with `focus({ preventScroll: true })`, then uses the shared `scrollIntoView` util so `scroll-margin-top` clears the nav and no hash is written to the URL. The `href="#main-content"` remains as a no-JS fallback.
- The component is split across the server/client boundary (label fetched via `getTranslations`, interaction client-side) and co-located under `src/components/SkipLink/` (`index.tsx` + `SkipLinkContent.tsx`). The public import path is unchanged.

## Testing

- Verified in-browser with keyboard navigation: activating the skip link moves focus into the main content, the next Tab lands on the first interactive element inside the article, the heading clears the fixed nav, and no `#main-content` hash is added to the URL.
- `pnpm type-check` passes.


src/components/SkipLink/index.tsx
pettinarip, line 1: this file is unnecessary, right? the `common` namespace is always pushed to the client. So, we could merge it with src/components/SkipLink/SkipLinkContent.tsx

pettinarip (approved):

PR #18502, opened 2026-06-24
fix: button dropdown scroll

myelinated-wackerow:
## Description
- Caps ButtonDropdown component height to that available on screen
- Adds collisionPadding to pad against bottom/top of screen
- Adds up/down chevron indicators when more items are available that are not visible
- Enables scrolling within dropdown menu

## Related Issue
Bug: larger dropdown menu contents leave items inaccessible on shorter screens


pettinarip (changes requested):
The scroll/chevron behavior works well, but I think we're solving it at the wrong level and I'd like to reconsider the approach before merging.

The root issue isn't specific to `ButtonDropdown`. It's that our `ui/dropdown-menu.tsx` is a stale shadcn snapshot. Our `DropdownMenuContent` uses `overflow-hidden` with no max-height, so *any* dropdown with a long list overflows the viewport and gets clipped on short screens.

Current upstream shadcn already fixes exactly this on the shared content:

```
max-h-(--radix-dropdown-menu-content-available-height) overflow-x-hidden overflow-y-auto
```

The new added here is the chevron/fade, vanilla shadcn just shows a native scrollbar. That's a nice touch, but as a separate `ButtonDropdown/` folder it duplicates the dropdown and leaves us with two similar components, which I think will confuse future contributors about which to reach for.

My suggestion:
1. Bring `DropdownMenuContent` in line with upstream (`max-h` + `overflow-y-auto`) so scrolling works everywhere.
2. Drop the parallel component. If we want to keep the chevron affordance, fold it into the shared content behind an opt-in prop (e.g. `<DropdownMenuContent scrollAffordance>`) so there's still one component and one mental model.

pettinarip (approved):
Looks great! gj

was wondering if in the future we could just use `/ui/dropdown-menu.tsx` for everything and drop `ButtonDropdown.tsx` completely 🤔

PR #18506, opened 2026-06-24
patch: rm invalid outline-color button variant

myelinated-wackerow:
## Description

Several `<ButtonLink>` instances in the roadmap content used `variant="outline-color"`, which is not a valid button variant. It only rendered as an outline by coincidence: `cva` silently ignores the unknown value, and the button's base styling already is the outline pattern (the real `outline` variant is an empty string). Because these live in MDX content, TypeScript never caught the invalid prop.

This swaps all occurrences to the proper `variant="outline"` (identical render, correct semantics) and cleans up a few docs/tests that referenced the invalid name as incidental sample data.

The non-English copies of these pages still carry the old value; the intl-pipeline will propagate the `outline` fix to them deterministically (it's an inert, non-translatable attribute, so no re-translation is triggered).

## Changes

- `public/content/roadmap/{scaling,security,statelessness,user-experience}/index.md` — `outline-color` -> `outline`
- `src/scripts/intl-pipeline/intl-sanitizer.ts` — update docstring example
- `tests/unit/intl-pipeline/sanitizer/standalone-fixes.spec.ts` — update test fixture (still passes)
- `docs/solutions/integration-issues/sanitizer-test-research.md` — update bug-pattern example


pettinarip (approved):
Perfect

PR #18507, opened 2026-06-25
refactor(ui): roadmap page to latest design system

myelinated-wackerow:
## Description

Refactors the `/roadmap` page onto the latest design-system patterns, matching the recently refactored `community` / `community/events` hubs. No copy or content changes — this is structure, layout primitives, and a couple of visual fixes.

**Page architecture**

- Converted the page to a Server Component: the body now lives in `page.tsx` on the standard `main > MainArticle.flow > Section` skeleton, dropping the old `"use client"` `_components/roadmap.tsx` wrapper and the page-wide `I18nProvider`.
- `ReleaseCarousel` is now the only client island, wrapped in a scoped `I18nProvider` (same pattern as events' `FilterEvents`), and its 4-deep wrapper nesting was collapsed.
- End-of-page `ContentFeedback` moved outside `MainArticle`; every `<Section>` carries a stable `id`.

**Primitives over one-offs**

- "Changes coming" and "Technical upgrades" cards rebuilt on `ui/card` + `Grid` (replacing the custom card divs / `SubpageCard`).
- Semantic headings and `flow` / `space` / `page` spacing tokens throughout (no more hand-rolled margins or arbitrary padding).

**Visual fixes**

- "Why we need a roadmap" image: a new portrait crop (`community-hero-portrait.png`) is served on desktop via `<picture>` (landscape on mobile), so only the rendered variant downloads. It's full-height and right-aligned to the gutter; mobile height is capped.
- Fixed the release-timeline connector gradient: it now fades `primary` → `primary-low-contrast` at the true shipped-release boundary instead of only when the next release has a `soon` status. Previously the boundary rendered as a hard edge whenever the upcoming release had no scheduled date (the current case).
- The `extra-security` roadmap icon now uses `currentColor`.

**i18n**

- Extracted the remaining hardcoded English in the Fusaka release card (Blob Parameter Only forks, Gas Limit & DoS hardening) to `page-roadmap` translation keys so they flow through the intl pipeline.

## Related Issue

N/A


src/components/icons/roadmap/extra-security.svg
pettinarip, line 1: showing in black
Image

pettinarip (commented):
LGTM, one small issue detected on the color of the shield icon

pettinarip (approved):

src/components/icons/roadmap/extra-security.svg
wackerow: good catch—patched

PR #18518, opened 2026-06-25
refactor(ui): PageHero - drop header prop, add component slot

myelinated-wackerow:
## Summary

Cleans up `PageHero` and extends it to cover more layouts, then migrates the Bug Bounty page (which had reimplemented a hero by hand) onto it.

### `PageHero` API changes

- **Removed the `header` eyebrow prop.** `title` is now **always** the page `<h1>` (it previously demoted to `<h2>` when a `header` was passed). The four pages still using `header` (`run-a-node`, `staking`, `wallets`, `_vision`) now pass `breadcrumbs` instead; their breadcrumb labels already exist in the `common` namespace.
- **Added an optional `eyebrow` slot** (`ReactNode`) rendered between the breadcrumbs and the title — for a status indicator, tag, or label. Does not affect heading hierarchy.
- **Added a `heroComponent` slot.** `heroImg` and `heroComponent` are now a **discriminated union** (mutually exclusive at the type level; passing both is a compile error). An image stacks **above** the text on mobile; a component folds **below** it. Both sit beside the text on desktop. Either may be omitted for a text-only hero.
- Simplified the flex layout: row is the default (desktop); only the mobile axis is set per-instance (`max-lg:flex-col` / `max-lg:flex-col-reverse`). Dropped a redundant local `--space` declaration that duplicated the `:root` token in `base.css`.

### Bug Bounty migration

The hand-rolled hero now uses `<PageHero>`: breadcrumbs, the status dot via `eyebrow`, the leaderboard preview via `heroComponent`, and the gradient title applied through `className` (`**:[h1]:…bg-clip-text text-transparent`, replacing the inline WebKit styles).

### Housekeeping

- Normalized prop order at the `latest`, `stories`, and `stablecoins` call sites to match the component's argument order.
- Added Storybook stories for the `heroComponent` and `eyebrow` slots.
- Updated the `design-system` skill/KB to reflect the new API (no `header` prop, `eyebrow`/`heroComponent` slots).

## Review notes

- **Bug Bounty hero spacing/gradient** is the one spot worth eyeballing — the `heroComponent` wrapper padding and the gradient h1 descender handling (`pb-1`) were tuned without a live render. Chromatic on the new stories + a look at `/bug-bounty` will confirm.
- The Bug Bounty hero kept `variant="no-divider"` to preserve its original border-less look; flag if the unified divider is preferred.

## Test plan

- [ ] Chromatic: new `PageHeroWithComponent` / `PageHeroWithEyebrow` stories render correctly (light/dark, RTL, mobile/desktop).
- [ ] `/bug-bounty` — eyebrow, gradient title, leaderboard aside (right on desktop, below on mobile), buttons.
- [ ] `/run-a-node`, `/staking`, `/wallets` — breadcrumb now shows in the hero (was the uppercase eyebrow); title is the `<h1>`.
- [ ] A text-only hero (e.g. `/developers/tools`) and an image hero (e.g. `/stablecoins`) are unchanged.

## Preview link
https://deploy-preview-18518.ethereum.it/bug-bounty (hero component)
https://deploy-preview-18518.ethereum.it/staking (hero image, updated breadcrumb)
https://deploy-preview-18518.ethereum.it/communty/support (no hero image or component)


src/components/Hero/PageHero/index.tsx
```diff
@@ -59,57 +74,36 @@ const PageHero = ({
...
+            breadcrumbs
+          ) : (
+            <Breadcrumbs {...(breadcrumbs as BreadcrumbsProps)} />
+          )}
         </div>
 
-        <PrimaryHeading className="text-4xl font-black not-last:mb-space lg:text-6xl">
+        {eyebrow && <div className="mb-space-2x">{eyebrow}</div>}
+
+        <h1 className="text-4xl font-black not-last:mb-space lg:text-6xl">
```
pettinarip, line 104: preexisting but related: could we remove the hardcoded font sizes and use the globals?

pettinarip (approved):
LGTM! left one question that might be worth including in the PR but not blocking

src/components/Hero/PageHero/index.tsx
wackerow: We can, but the desktop size does not yet match on all pages. More below, but I'd suggest we set that aside as a separate task (easy patch if we end up reaching consensus)

@konopkja @nloureiro Our globals on desktop use text-5xl (48px), while the pages using the PageHero use text-6xl (60px). This hero was previously using text-7xl (64px) and we brought it to 60px during the refactor, but going another step drops by 12px more.

60px kept the PageHero h1's larger than normal, while avoiding making every h1 on the entire site that size. PageHero is used on the majority of pages, but not all. There are still one-off custom h1 setups on around a dozen pages still.

Option 1: reduce PageHero h1 to the global 48px (down from current 60px, previously 64px)
Option 2: increase global to 60px, defaulting to the larger size on all pages
Option 3: increase global to 60px, and go through on the dozen-or-so pages that don't use this and consider overriding with `text-h2` if we want to shrink those back down (though they'd end up smaller: `text-3xl lg:text-4xl` is 30px and 36px, mobile and desktop respectively)
Option 4: keep as-is with the global at 48px, and the PageHero overrides with a larger h1

I'd lean towards either 4 or 3... I think the 60px looks good, but it's not appropriate for every page imo. I think we should have two sizing options for the h1 depending on the page/hero.

PR #18519, opened 2026-06-25
refactor(ui): community subpages to latest design system

myelinated-wackerow:
## Description

Refactors the community subpages to the latest design system, aligning them with the conventions used across the other refactored page layouts. Component files are renamed to kebab-case, page layouts are simplified, and several event `_components` are reworked (footer CTA, no-results alert, organizer CTA).

## Scope

- `community/events/` -- page, conferences, meetups, search, and `_components` (event card, continent tabs, filter events/meetups, footer CTA, no-results alert, organizer CTA)
- `community/page.tsx` and `community/support/page.tsx`
- `src/lib/utils/date.ts` -- supporting tweak for date formatting

## Notes

- Component files renamed from PascalCase to kebab-case per project file-naming convention.
- No content/translation changes; English-only UI refactor.

## Preview
<https://deploy-preview-18519.ethereum.it/community/support>
<https://deploy-preview-18519.ethereum.it/community/events>
<https://deploy-preview-18519.ethereum.it/community/events/conferences>
<https://deploy-preview-18519.ethereum.it/community/events/meetups>
<https://deploy-preview-18519.ethereum.it/community/events/search>


app/[locale]/community/events/_components/continent-tabs.tsx
```diff
@@ -126,8 +126,15 @@ export default function ContinentTabs({
...
+      {/* Persistent polite live region: announces the empty-filter result to screen readers */}
+      <div role="status" className="sr-only">
+        {filteredEvents.length === 0 ? noEventsMessage : ""}
+      </div>
+
       {filteredEvents.length === 0 ? (
-        <p className="py-8 text-center text-body-medium">{noEventsMessage}</p>
+        <p className="mx-auto py-8 text-center text-body-medium">
+          {noEventsMessage}
+        </p>
```
Comment (pettinarip) [inline app/[locale]/community/events/_components/continent-tabs.tsx:137]: isn't this duplicated to sr? couldn't we remove the sr-only region in favor of this?

pettinarip (approved):
One tiny comment, the rest looks solid, gj!

app/[locale]/community/events/_components/continent-tabs.tsx
Comment (wackerow) [reply app/[locale]/community/events/_components/continent-tabs.tsx:137]: This is intentional actually..

The conditional on this `<p>` element means it won't mount _until_ that condition is met (a "no results" query)... most screen readers won't announce an element that is inserted into the DOM at the same moment as its content.. the region has to already exist and then change.

The `role="status"` block is always present, so when a "no results" query is reached, a screen reader already has access to this (the assistive technology registers it on load), the content changes from an empty string to a proper message, and the user is notified immediately. Adding `role="status"` to the `<p>` would basically be a no-op and nothing would be announced. 

Admittedly, it's basically a new pattern to the repo as we don't really do this elsewhere. I'll keep it for now... It's a better pattern imo, perhaps we can have a separate task to improve this functionality elsewhere. Overall lower priority at the moment though.

PR #18521, opened 2026-06-25
Show wallet fees on the find-wallet page

### LINKED ISSUE #18520: Show wallet fees on the find-wallet page to help users choose a provider
## UX enhancement

The [find a wallet](https://ethereum.org/wallets/find-wallet) page lets users compare wallets on platforms, languages, features, and security, but **not on cost**. Fees are often a deciding factor when choosing a wallet, and today a user has to leave the page and dig through each wallet's site or terms to find them.

## Proposal

Surface a concise, consistent fee summary on each wallet card (a new row beside the existing language/device rows), e.g.:
- `Swap fee: 0.85%`
- `Swap/bridge fee: 0.875%, Buy/sell fee: 1%`
- `Device: $149` (hardware)
- `Swap fee: 0%` / `Swap fee: variable`

This lets users weigh cost alongside the other attributes and make a more informed decision without leaving the page.

## Notes

- Fee data is sourced from the wallet-suggestion issues (which ask submitters about fees) and cross-checked against each wallet's official fee/ToS/pricing pages.
- Figures are point-in-time and self-reported in places, so they should be open to correction by wallet teams and maintainers.
- Initial version uses English strings; localization can follow.

A PR implementing this follows.

konopkja:
## Description

Adds a concise **fee summary** to each wallet card on the find-wallet page, so users can compare what a wallet costs (swap/bridge fees, buy/sell fees, or hardware price) alongside the other attributes when choosing a provider.

Closes #18520

## Changes

- `WalletData`: new optional `fee_summary` field (`src/lib/types.ts`)
- `wallet-data.ts`: `fee_summary` populated for 48 wallets
- New price-tag `FeeIcon` in the wallet icon set, matching the existing icon style
- `WalletInfo`: renders the fee row below the languages row (only when a summary exists)

## How the summaries were built

- Sourced from the wallet-suggestion issues (which ask submitters "what are the fees users may be paying?") and cross-checked against each wallet's official fee / ToS / pricing pages.
- Normalised to one consistent grammar: `Swap fee: X%`, `Swap/bridge fee: X%`, `Buy/sell fee: X%`, `Device: $X`. Free swaps show `Swap fee: 0%`; provider-set fees show `variable`; wallets that don't swap show no swap line.

## Caveats for reviewers

- Fee figures are point-in-time and partly self-reported. A few are `variable` or `not disclosed` where no official rate is published. Worth confirming with wallet teams; happy to adjust.
- Strings are English-only for now — localization can be a follow-up.

## Testing

Ran locally against the dev server; the fee row renders on every applicable wallet card and the page compiles without errors.


wackerow (changes requested):
These won't translate; we should find an approach where we can use intl-friendly strings

myelinated-wackerow (approved):
Approving -- the intl refactor resolves my earlier request. All copy now lives in `src/intl/en/page-wallets-find-wallet.json` as ICU messages, values format through our `numberFormat`/`formatPriceUSD`/`numberToPercent` wrappers (so ur/ar numbering is correct), and the fee list joins via `Intl.ListFormat`. Rendered all 46 wallets locally in en/de/ar/ur -- no missing keys, no double units.

Pushed one commit: dropped the `lower-for-token-holders` qualifier + its string (it only served Ctrl Wallet, removed in #18721, so nothing rendered it), and narrowed the range whitespace normalization so it stops downgrading the NBSP between value and unit.

Handoff notes, since these changes are going into the find-wallet revamp PR rather than straight to `dev`:

1. **Fee data still needs a wallet-team pass.** 46 of 48 listings carry figures that are point-in-time and partly self-reported, some on listings whose `last_updated` is years old (Rainbow: `2022-06-30` + "Swap fee: 0.85%"). This is the item most at risk of getting lost in the transfer.
2. **Methodology copy.** Worth one sentence in the existing `page-find-wallet-methodology-*` block saying fees are provider-reported and subject to change.
3. **No visual coverage.** The mocks in `FindWalletProductTable.stories.tsx` have no `fees`, so the row never reaches Chromatic. One `fees: [...]` line on a mock fixes that in the revamp.
4. **Qualifier taxonomy won't scale.** 7 of 8 qualifiers are used exactly once and 4 encode brand semantics (FOX, TPT, Premium). Each new caveat costs a union member, an intl key, and 24 machine translations -- so a data-only contribution can't express one. Consider dropping qualifiers from the card and letting the wallet's own page carry the detail.
5. **Optional guard.** `WalletFeeTFunc` types the key as `string`, so a typo'd `type`/`qualifier` renders the raw key at runtime with no build error. A unit test asserting every union member has a matching intl key would have caught the dead one above.

-- Opus 5

wackerow (approved):
cc: @pettinarip

pettinarip (approved):

PR #18536, opened 2026-06-26
feat: revamp /get-eth page to match new design

pettinarip:
## Description

Revamps the `/get-eth` page to match the new Figma design.

### Highlights
- **Hero**: live ETH price card (`EthPriceCard`, restyled to the new neutral/centered design) overlaid on the coin cluster, via a new optional `imageOverlay` slot on `PageHero`.
- **Ways to get ETH**: card grid (CEX, earn, receive, DEX, wallets, staking), DEX section removed per design.
- **Find a local crypto exchange**: light-blue section with a top-anchored world-map background (light + dark PNG variants), copy refreshed.
- **Wallets / Keeping your ETH safe**: grey-panel section with white (`nested`) cards.
- Copy updated throughout to match Figma (`page-get-eth.json`).

### Notes
- ETH price is wired to live data (`useGasEthPrice` → `/api/gas-eth-price`).
- `PageHero` gains an optional, backward-compatible `imageOverlay?: ReactNode` prop (no effect on existing callers).

## Checklist
- [x] `pnpm type-check` passes
- [x] ESLint clean
- [x] Verified in browser (light + dark, hero + collapsed/expanded exchange section)

Preview: https://deploy-preview-18536.ethereum.it/get-eth


app/[locale]/get-eth/page.tsx
```diff
@@ -124,278 +154,230 @@ export default async function Page(props: { params: Promise<PageParams> }) {
...
+        <I18nProvider locale={locale} messages={messages}>
+          <MainArticle className="flow">
+            <Section id="ways">
+              <h2 className="text-center text-h1">
+                {t("page-get-eth-ways-you-can-get-eth")}
+              </h2>
+              <Grid columns={3}>
+                {waysToGetEth.map(
+                  ({ icon: Icon, title, description, linkText, href }) => (
+                    <Card key={href + String(title)}>
```
Comment (wackerow) [inline app/[locale]/get-eth/page.tsx]:
```suggestion
                    <Card key={href + String(title)} hoverLift>
```

cc: @konopkja This card has a single action -- Maybe worth chatting if we want to apply similar logic to this type of card as we do for cards that have a single CTA button. This one uses a link, specifically `LinkWithArrow`. If we want the whole card to be clickable for these too, could set up a variant to the `CardButtonFake` so we can use the "link with arrow" as well as a "fake link", while the entire card is clickable. Non-blocking for this, just noting.

app/[locale]/get-eth/page.tsx
```diff
@@ -124,278 +154,230 @@ export default async function Page(props: { params: Promise<PageParams> }) {
...
-              <p className="mb-8 max-w-3xl text-center">
-                {t("page-get-eth-exchanges-intro")}
+                </em>
               </p>
+            </Section>
+
+            <Section>
+              <Callout
+                image={handEth}
+                alt=""
```
Comment (wackerow) [inline app/[locale]/get-eth/page.tsx]:
```tsx
// src/components/ui/callout.tsx

// type:
alt?: string

// props:
alt = "",
```

Should be able to just leave this off


```suggestion
```

app/[locale]/get-eth/page.tsx
```diff
@@ -47,49 +52,74 @@ import { exchangesByCountryLastUpdated } from "@/data/exchangesByCountry"
...
 import wallet from "@/public/images/wallet.png"
 
-type CardProps = {
-  children: ReactNode
-  emoji: string
+const Eyebrow = ({ children }: { children: ReactNode }) => (
+  <p className="text-sm font-bold tracking-wide text-primary-high-contrast uppercase">
+    {children}
+  </p>
+)
```
Comment (wackerow) [inline app/[locale]/get-eth/page.tsx]: Might be worth considering abstracting this as it's own reusable ui component

src/components/CentralizedExchanges/index.tsx
wackerow, line 89: Image

Out-of-scope, but noticing this `Select` has a background that in this case obliterates when it unfolds onto the next section with the same bg color. cc: @konopkja Could consider at a minimum a border on this to help clearly define it visually

app/[locale]/get-eth/page.tsx
```diff
@@ -124,278 +154,230 @@ export default async function Page(props: { params: Promise<PageParams> }) {
...
+                      {t("page-get-eth-protect-eth-in-wallet")}
+                    </CardTitle>
+                    <CardParagraph>
+                      {t("page-get-eth-protect-eth-desc")}
+                    </CardParagraph>
+                    <InlineLink href="/wallets/">
+                      {t("page-get-eth-your-address-wallet-link")}
+                    </InlineLink>
+                  </CardContent>
+                </Card>
```
Comment (wackerow) [inline app/[locale]/get-eth/page.tsx]: @konopkja The first and third cards here each contain links -- per our chats we would logically add the `hoverLift` to these, but noting that the second card does not contain an action, so it would not. Curious if you any concerns with placing the `hoverLift` on two cards while the third in the group does not have it.

app/[locale]/get-eth/page.tsx
```diff
@@ -124,278 +154,230 @@ export default async function Page(props: { params: Promise<PageParams> }) {
...
+                    {t("page-get-eth-description")}
+                  </p>
+                  <p className="text-lg text-body-medium">
+                    {t("page-get-eth-security")}
+                  </p>
+                </div>
+              </Grid>
+
+              <Grid columns={3} className="mt-space-2x">
+                <Card variant="nested">
```
Comment (wackerow) [inline app/[locale]/get-eth/page.tsx]:
```suggestion
                <Card variant="nested" hoverLift>
```

app/[locale]/get-eth/page.tsx
```diff
@@ -124,278 +154,230 @@ export default async function Page(props: { params: Promise<PageParams> }) {
...
+                        {t("page-get-eth-do-not-copy")}
+                      </p>
+                    </div>
+                    <CardParagraph>
+                      {t("page-get-eth-your-address-desc-3")}
+                    </CardParagraph>
+                  </CardContent>
+                </Card>
+
+                <Card variant="nested">
```
Comment (wackerow) [inline app/[locale]/get-eth/page.tsx]:
```suggestion
                <Card variant="nested" hoverLift>
```

app/[locale]/get-eth/page.tsx
```diff
@@ -124,278 +154,230 @@ export default async function Page(props: { params: Promise<PageParams> }) {
...
-          </p>
-
-          <EthPriceCard className="mb-space" />
-          <ButtonLink
-            href="#country-picker"
-            customEventOptions={{
+      <PageHero
+        variant="no-divider"
+        breadcrumbs={{ slug: "get-eth" }}
+        heroComponent={
```
Comment (wackerow) [inline app/[locale]/get-eth/page.tsx:155]: ❤️ wonderful

wackerow (approved):
@pettinarip Lookin good! Few tidy up suggestions and a couple design questions, otherwise lgtm

app/[locale]/get-eth/page.tsx
Comment (pettinarip) [reply app/[locale]/get-eth/page.tsx]: Added `hoverLift`. Whole-card-clickable would be a nice follow-up via a `CardButtonFake` + link-with-arrow variant.

app/[locale]/get-eth/page.tsx
Comment (pettinarip) [reply app/[locale]/get-eth/page.tsx]: Done — `alt` is optional, defaults to `""`.

app/[locale]/get-eth/page.tsx
Comment (pettinarip) [reply app/[locale]/get-eth/page.tsx]: Extracted to `@/components/ui/eyebrow`.

src/components/CentralizedExchanges/index.tsx
pettinarip: Out of scope 👍 — worth a focused follow-up with @konopkja.

app/[locale]/get-eth/page.tsx
Comment (pettinarip) [reply app/[locale]/get-eth/page.tsx]: Added `hoverLift` to the two link cards, left the address card flat (no action). @konopkja lmk if you'd prefer uniform.

app/[locale]/get-eth/page.tsx
Comment (pettinarip) [reply app/[locale]/get-eth/page.tsx]: Done.

app/[locale]/get-eth/page.tsx
Comment (pettinarip) [reply app/[locale]/get-eth/page.tsx]: Done.

PR #18599, opened 2026-06-27
Nested tooltip/popover variant + arrow-wrapping outline

myelinated-wackerow:
## Summary

Tooltips and popovers now draw their border + elevation as a single `drop-shadow` stack (`popover-outline` utility) so the outline wraps the Radix `Arrow` instead of a plain `border` severing it at the content-box edge. Adds a `nested` prop to `PopoverContent`/`TooltipContent` (threaded through the `Tooltip` wrapper and `ChainImages`) that swaps the surface + arrow fill to the base `bg-background`, so a tooltip opening over an already-`bg-background-highlight` container (e.g. a wallet table row on hover) stays legible instead of blending in.

Shadow opacity is themed via `--body` at `0.1` (light) / `0.15` (dark), declared inside the utility. Rationale that previously lived in oversized inline comments now lives in the design-system skill (Floating surfaces).

## Commits

1. `feat: nested popover/tooltip + outline utility` -- the `popover-outline` utility and the `nested` prop on the ui primitives.
2. `feat: wire nested tooltips into wallet table` -- threads `nested` through `Tooltip`/`ChainImages` and opts the wallet table's chain icons + filter tooltips into it.
3. `docs: add floating-surface notes to skill` -- moves the rationale into `design-system` references.

## Test plan

- [ ] Wallet product table: chain-icon and filter tooltips use the base background over highlighted rows and stay legible.
- [ ] Default tooltips/popovers elsewhere: unchanged highlight background, visible 1px border that wraps the arrow.
- [ ] Dark mode: shadow lift reads as expected at 0.15 opacity.
- [ ] Chromatic: confirm 1px border coverage at rounded corners and the arrow tip.


pettinarip (approved):
LGTM

PR #18471, opened 2026-06-23
i18n: translation pipeline, loose ends batch 1 (all languages)

myelinated-wackerow:
## Full-tree translation update

Automated `intl-pipeline` full-tree run (blank `TARGET_PATH`) across all 24 target languages. The pipeline's own PR step failed because its auto-generated body exceeded GitHub's 65,536-character limit (it lists every changed file); this PR was opened in recovery with a consolidated summary instead.

**Run:** https://github.com/ethereum/ethereum-org-website/actions/runs/27984239227  -  ~$6.26, 200 Gemini calls, ~42 min.

### Scale
- **939 files** changed vs `dev`
  - 467 translated markdown (`public/content/translations/<lang>/`)
  - 266 translated JSON namespaces (`src/intl/<lang>/`)
  - 206 manifests (`.manifests/`)

### Per-language (content files)

| Lang | MD | JSON | Total |
|---|---|---|---|
| ar | 20 | 11 | 31 |
| bn | 19 | 11 | 30 |
| cs | 19 | 11 | 30 |
| de | 19 | 11 | 30 |
| es | 19 | 11 | 30 |
| fr | 20 | 11 | 31 |
| hi | 19 | 11 | 30 |
| id | 21 | 11 | 32 |
| it | 19 | 12 | 31 |
| ja | 19 | 11 | 30 |
| ko | 19 | 11 | 30 |
| mr | 19 | 11 | 30 |
| pl | 20 | 11 | 31 |
| pt-br | 20 | 11 | 31 |
| ru | 20 | 11 | 31 |
| sw | 19 | 12 | 31 |
| ta | 19 | 11 | 30 |
| te | 20 | 11 | 31 |
| tr | 20 | 11 | 31 |
| uk | 20 | 11 | 31 |
| ur | 19 | 11 | 30 |
| vi | 20 | 11 | 31 |
| zh-tw | 19 | 11 | 30 |
| zh | 19 | 11 | 30 |

### Markdown by content area

| Area | Files |
|---|---|
| guides | 144 |
| developers | 81 |
| community | 48 |
| roadmap | 48 |
| staking | 48 |
| real-world-assets | 24 |
| web3 | 24 |
| zero-knowledge-proofs | 24 |
| whitepaper | 21 |
| ethereum-forks | 1 |
| videos | 1 |
| energy-consumption | 1 |
| what-are-apps | 1 |
| contributing | 1 |

_Per-file list omitted to stay within GitHub's PR body limit; see the Files Changed tab._

[...]

PR #18609, opened 2026-06-28
fix(intl): make the translation pipeline reliable (manifest race, crash-safety, PR body, RECITATION)

myelinated-wackerow:
## Summary

Fixes four intl-pipeline bugs that made full-tree runs unreliable, in one branch. The headline bug silently desynced manifests so the *next* run re-translated everything — blocking the goal of running the pipeline on every merge to `dev`.

## The bugs

**1. Committer ref race dropped manifests (root cause).** `SharedCommitter` advanced the branch ref on every per-file commit. Under the task pool's concurrency those ref updates raced and returned `422 not-a-fast-forward`. A content commit that threw on the 422 aborted *before* recording its manifest blob; the end-of-run squash rebuilds from recorded blobs, so content shipped **without its manifest**. Run [27984239227](https://github.com/ethereum/ethereum-org-website/actions/runs/27984239227) / PR #18471: ~733 files translated, ~0 markdown manifests committed (100% correlation: e.g. `ar` had 20/20 markdown content commits hit 422 and 0 manifests landed). The next run then sees every file as stale and re-translates all of them.

Fix: `commitFile` now only creates + records a blob and never touches the branch ref; `squashByLanguage` force-updates the ref once. Manifests are also built before the content blob is recorded, so a builder error strands neither.

**2. No crash-safety after #1.** Moving all persistence to the end-of-run squash meant a mid-run crash lost that run's costly LLM output. Restored incremental durability without the race: `commitFile` flushes everything recorded so far to the branch every 50 files (the squash run mid-stream), serialized through a single-writer lock so no two ref updates overlap. Per-file granularity (not per-language), non-fatal on failure, and the final squash still collapses to one commit per language.

**3. PR body exceeded GitHub's 65,536-char limit.** Full-tree runs failed to post their PR (the body listed every failure and grew unbounded across runs), forcing a hand-opened recovery PR (#18471). Failure lists are now capped at 40 and the body is clamped to 60,000 chars (tail-preserving) at both create and update.

**4. RECITATION blocked our own content deterministically.** Gemini requests were pinned to `temperature: 0`, so a `finishReason=RECITATION` block (false-positive on our open-licensed whitepaper/consensus docs) reproduced identically on every retry and fallback model. Retries now escalate temperature (0 -> 0.5 -> 1.0) to clear the recitation match. (Consolidated from work originally on `review-18471`.)

## Testing

- New unit specs: `tests/unit/intl-pipeline/commit-ref-race.spec.ts` (5 tests: no per-file ref update, squash lands content+manifest together, interleaved cross-language commits drop nothing, periodic checkpoint serialized force-update, auto-checkpoint every `CHECKPOINT_EVERY`, failed-checkpoint non-fatal) and `pr-body-limit.spec.ts` (3 tests: clamp + failure cap).
- Full intl-pipeline unit suite green (904 passed), `tsc` clean, `eslint` clean.
- Three adversarial code-review passes (final one over the whole branch, focused on the concurrent-checkpoint design): no blockers.

## Follow-up (not in this PR)

Once merged: close PR #18471 and delete `intl/pending-dev` (its content is salvageable but its manifests are drifted), then re-run the pipeline blank-target from clean `dev`. With these fixes a blank-target run skips files whose manifests match current English, translates the rest, and lands them all in one PR with correct manifests — gracefully skipping (and reporting) any RECITATION/transient failures instead of corrupting state.

### DIFF
--- a/.claude/skills/intl-pipeline/references/gotchas.md
+++ b/.claude/skills/intl-pipeline/references/gotchas.md
@@ -17,10 +17,12 @@ The Gemini adapter at `src/scripts/intl-pipeline/lib/llm/gemini.ts` checks `resp
 - `STOP` — normal completion
 - `MAX_TOKENS` — output truncated; section probably too large
 - `SAFETY` — content filter blocked it. Safety settings are `BLOCK_NONE` in the adapter, but blocks can still trigger on some edge content (mining/attack descriptions in certain non-Latin languages). If `BLOCK_NONE` doesn't help, the prompt or content needs rework, not the safety settings.
-- `RECITATION` — model declined to reproduce training data. Rare; restart usually resolves.
+- `RECITATION` — model declined to reproduce training data. **Deterministic per file+language**, NOT transient: the adapter retries the byte-identical prompt up to 3x and gets the identical `RECITATION` every time (`tokens_out=0`), then gives up. The file+lang is then skipped — it ships untranslated (keeps its prior/English state), is recorded as a failed task, and listed in the PR body's failure block. Restarting the whole run will hit the exact same combos. Recurring victims are long reference docs (consensus-mechanisms `pos`/`poa`, `defi`, `ethash`, whitepaper) in fr/es/pt-br. The real fix is upstream of a plain retry: mutate before retrying (smaller chunks, secondary model, reworded prompt) or accept the skip and handle those files out-of-band. Don't burn time expecting a re-run to clear them.
 - `OTHER` — bucket catchall. Log shows full response; debug case-by-case.
 
 Non-STOP finish reasons are logged at WARNING level. Search workflow logs for `FINISH_REASON` if a section seems to be missing translation output.
+
+**Beware: a "success" workflow conclusion does not mean a clean run.** Per-file+lang tasks that fail (RECITATION, blob/build errors) are emitted as `[WARN]`, listed in the PR body's "N task(s) failed" block, and the run still goes green. Always read the PR body's failure block and grep the log before trusting a green run. See `recovery.md` -> "Diagnosing a completed run."
 
 ## `intl-content-tree` is a separate npm package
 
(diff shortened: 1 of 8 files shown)


myelinated-wackerow:
## Review: Approve

Reviewed the full diff plus the surrounding code each change depends on. All four fixes are correct, and the core concurrency argument holds. No blockers.

### Verified
- **Ref-race fix.** `commitFile` now only creates+records a blob; the sole ref update is `_squashNow`'s force-update, serialized through `refLock`. The safety precondition holds: `main.ts` `await pool.drain()` fully completes before `squashByLanguage()`, with nothing touching `commitFile` in between -- so no concurrent append can race the final squash. The race surface is gone.
- **Checkpoint idempotency.** Every checkpoint and the final squash rebuild the chain from base and force-update, so the end state is independent of how many checkpoints ran. Counter reset-before-fire correctly avoids a queued burst.
- **PR body.** Only unbounded sources are the failure list (capped at 40) and accumulated run summaries (`clampBody`, tail-preserving). `generateRunSummary` uses counts, not a per-file enumeration. Fully addressed.
- **RECITATION.** `MAX_RETRIES=3` -> temps escalate 0 -> 0.5 -> 1.0 in the loop that re-issues on RECITATION; attempt 1 stays deterministic. Low-risk.
- **Rate limits.** This *lowers* GitHub write volume vs the design already in production: ~3-3.5k writes (1 blob/file + ~36 checkpoints) vs the old ~6-8k (blob+tree+commit+ref per file), well under the 5k/hr primary limit. `fetchWithRetry` backs off 60s+ on any secondary-limit graze and checkpoint failures are non-fatal.

### Non-blocking residuals
1. **`main.ts` comment slightly overstates the invariant.** It says a failed task "records NEITHER content nor manifest." True if the throw is in manifest *building* (correctly moved first), but a content blob POST can succeed and then the manifest blob POST throw -- stranding content without its manifest. Low-probability and self-healing (re-translates next run), so content-first is the right order; just tighten the wording.
2. **Watch the first full-tree run** for `[RATE LIMIT]` lines around checkpoint bursts. If they appear, bump `CHECKPOINT_EVERY` 50 -> ~100 rather than touching retry logic.

Optionally tighten the one comment before merge; everything else is follow-up monitoring. Ship it.

-- Claude

wackerow (approved):
@pettinarip Heads up, pulling this in to move forward on the remaining translations

Issue #18593, opened 2026-06-27
Tech debt: Table component has cell spacing issues and deprecated align attribute


alerodriargui:
﻿## Description

In src/components/ui/table.tsx there is a comment at line 10 about cell spacing issues and at line 80 about a deprecated align attribute.

## Issues

1. Cell spacing issues with some table content
2. Deprecated align attribute in use (HTML align attribute is deprecated in favor of CSS)

## Impact

Table rendering may be inconsistent, and the deprecated attribute may trigger browser warnings.

PR #18621, opened 2026-06-29
Fix table spacing and align handling

alerodriargui:
## Description

Updates the shared table component to address the table tech debt noted in #18593.

Changes include:

- Explicitly resetting table border spacing with `border-collapse border-spacing-0`
- Removing the stale TODO about table cell spacing
- Keeping Markdown `align` support while applying it through CSS `text-align`
- Preserving any existing inline `style` values passed to `TableHead` or `TableCell`

## Related Issue

Ref: #18593

### DIFF
--- a/src/components/ui/table.tsx
+++ b/src/components/ui/table.tsx
@@ -6,12 +6,6 @@ import { tv, type VariantProps } from "tailwind-variants"
 
 import { cn } from "@/lib/utils/cn"
 
-/**
- * TODO: Currently, there are cell spacing issues with some table content.
- * Prefer `table-fixed` utility class in the future when content has been addressed
- * to provide equal cell widths.
- */
-
 const baseStyles = {
   th: "text-start border-b border-body text-body normal-case align-bottom p-4",
   tr: "not-[:last-of-type]:[&_th]:border-e-2 not-[:last-of-type]:[&_th]:border-e-background not-[:last-of-type]:[&_td]:border-e-2 not-[:last-of-type]:[&_td]:border-e-background",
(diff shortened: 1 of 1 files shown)


nloureiro (changes requested):
There's one genuinely good fix in here worth keeping, but the headline change is a no-op and a couple of details need adjusting:

**Keep:** the `withTextAlign` style-merge. On `dev`, `style={{ textAlign: align }}` followed by `{...props}` means a caller-passed `style` silently drops the alignment — merging fixes a real (if latent) bug.

**Please change:**
1. `border-collapse border-spacing-0` has no effect: Tailwind v4 preflight (imported in `src/styles/global.css`) already applies `border-collapse: collapse` to all tables, and `border-spacing` is ignored under collapse. Drop it (or explain what it fixes visually).
2. The deprecated-`align` handling this targets was already done on `dev` (`ui/table.tsx` strips `align` and applies CSS `text-align`), and the deleted TODO tracked *future* `table-fixed` work this PR doesn't implement — please restore the TODO (or implement it) and change "Fixes #18593" to "Refs #18593" so the issue doesn't auto-close unresolved.
3. The `TableCaption` class reorder fights `prettier-plugin-tailwindcss` — dev's order is the plugin's sorted output, so the next `pnpm format` reverts yours.

With the style-merge kept and the rest reverted this becomes a small, clean fix.

wackerow (approved):

wackerow:
Pulling in, but I pushed a commit to revert the removal of the TODO, and changed the "Fixes" PR body to "Ref:" so it doesn't auto-close that issue

PR #18622, opened 2026-06-29
Update media query listener handling

### LINKED ISSUE #18592: Tech debt: Deprecated addListener/removeListener in useMediaQuery
﻿## Description

In src/hooks/useMediaQuery.ts:55, the deprecated addListener/removeListener API is used for media query change events.

## Impact

This API is deprecated in modern browsers and may be removed in future versions. While kept for Safari < 14 support, it should be migrated to the standard addEventListener API when Safari < 14 support is dropped.

## Note

This is intentionally kept for Safari < 14 compatibility, but should be tracked for migration.

alerodriargui:
## Description

Updates `useMediaQuery` to prefer the standard `MediaQueryList.addEventListener("change", ...)` API for media query subscriptions.

Changes include:

- Using feature detection for `addEventListener`
- Keeping a legacy Safari fallback isolated behind a typed `LegacyMediaQueryList`
- Removing the previous `try/catch` fallback around listener registration

## Related Issue

Fixes #18592

### DIFF
--- a/src/hooks/useMediaQuery.ts
+++ b/src/hooks/useMediaQuery.ts
@@ -51,13 +51,19 @@ export function useMediaQuery(
 
 type MediaQueryCallback = (event: MediaQueryListEvent) => void
 
+type LegacyMediaQueryList = MediaQueryList & {
+  addListener: (callback: MediaQueryCallback) => void
+  removeListener: (callback: MediaQueryCallback) => void
+}
+
 function listen(query: MediaQueryList, callback: MediaQueryCallback) {
-  // Use deprecated `addListener` and `removeListener` to support Safari < 14 (#135)
-  try {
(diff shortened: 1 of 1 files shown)


nloureiro (approved):
Verified: the old try/catch was an intentional Safari <14 shim (#135), and this PR **keeps** that support — it swaps the try/catch for `typeof query.addEventListener === "function"` feature detection with the `addListener/removeListener` fallback intact. Functionally equivalent, standard pattern. Non-blocking nit: the `LegacyMediaQueryList` intersection type is redundant (lib.dom already declares the deprecated methods) but harmless. CI green. Human merge pending.

wackerow (approved):
Lgtm, thanks @alerodriargui!

PR #18623, opened 2026-06-29
Fix network upgrade date formatting

### LINKED ISSUE #18591: Bug: dateTimeAsString needs fixing in NetworkUpgradeSummary
﻿## Description

In src/components/History/NetworkUpgradeSummary.tsx:37 there is a TODO: "// fix dateTimeAsString"

## Impact

The date/time formatting for network upgrades may be displaying incorrect or improperly formatted dates to users.

## Suggested action

Investigate and fix the dateTimeAsString formatting logic.

alerodriargui:
## Description

Fixes `NetworkUpgradeSummary` date formatting for network upgrade metadata.

Changes include:

- Removing the stale `dateTimeAsString` TODO
- Validating the upgrade date with `getValidDate` before rendering
- Formatting the timestamp directly in UTC using the shared `dateTimeFormat` helper
- Including the UTC time zone via `timeZoneName` instead of appending `+UTC`
- Removing the unnecessary client-side state/effect used only for date formatting

## Related Issue

Fixes #18591


nloureiro (approved):
Verified: `getValidDate` exists in `src/lib/utils/date.ts`, and formatting during render matches the established pattern in `LocaleDateTime.tsx` (the old client-only `useEffect` deferral was the outlier, not the norm). Rendering is now also gated on a *valid* date instead of just a truthy string, and `timeZoneName: "short"` replaces the hand-rolled `+UTC` suffix. Nice cleanup of #18591. Human merge pending.

wackerow (approved):
Looks good! Separately, would be nice to add some normalization for whitespace in these date strings, since Node's ICU and the browser's ICU can emit different whitespace causing hydration issues. Should be okay in this branch, but I'll post a follow-up issue.

PR #18626, opened 2026-06-29
Use country codes for centralized exchange exceptions

### LINKED ISSUE #18588: Tech debt: useCentralizedExchanges hook may be unused; needs i18n support
﻿## Description

The hook useCentralizedExchanges at src/hooks/useCentralizedExchanges.ts:5 has a self-identified TODO: "// Remove unused?" suggesting it may be dead code.

## Additionally

At line 389 there is: "// Set up for i18n support; currently all country names in English"

## Impact

If unused, this hook contributes to bundle size and maintenance burden. If still needed, it requires i18n support for country names.

alerodriargui:
## Description

- Confirms `useCentralizedExchanges` is still used by the Get ETH centralized exchanges UI and removes stale unused-code TODO comments.
- Stores the source country code on each select option so U.S.-specific exchange exceptions are detected by `US` instead of comparing against an English country label.
- Removes the obsolete i18n TODO now that localized country labels can still trigger the U.S. exception logic correctly.

## Related Issue

Fixes #18588

## Preview link
https://deploy-preview-18626.ethereum.it/get-eth

### DIFF
--- a/src/hooks/useCentralizedExchanges.ts
+++ b/src/hooks/useCentralizedExchanges.ts
@@ -2,14 +2,6 @@ import { useState } from "react"
 import { shuffle } from "lodash"
 import { useLocale } from "next-intl"
 
-// TODO: Remove unused?
-// import argent from "@/public/images/wallets/argent.png"
-// import binanceus from "@/public/images/exchanges/binance.png"
-// import imtoken from "@/public/images/wallets/imtoken.png"
-// import mycrypto from "@/public/images/wallets/mycrypto.png"
-// import myetherwallet from "@/public/images/wallets/myetherwallet.png"
-// import squarelink from "@/public/images/wallets/squarelink.png"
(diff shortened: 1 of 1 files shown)


nloureiro (approved):
Verified — and this is a better find than the issue suggested: on current `dev`, option labels come from `Intl.DisplayNames` (`getCountryCodeName`), which returns "United States" — never equal to the literal `"United States of America (USA)"` being compared. So the USA state-exception descriptions are **dead code in every locale today**, not just non-English ones. `countryCode === "US"` fixes it; the `US` key exists in `exchangesByCountry.ts`, the `Object.entries` cast is sound for this const object, and the removed commented-out imports settle the stale part of #18588. CI green incl. type-check. Sequencing: #18661 touches the same hook file — whichever merges second needs a trivial rebase.

wackerow (approved):
Looks good, thanks for the clean up @alerodriargui

PR #18627, opened 2026-06-29
Sentry runtime logging for copy failure 

### LINKED ISSUE #18586: Tech debt: ~109 console.warn/error calls without centralized logging service
﻿## Description

There are ~109 console.warn and console.error calls throughout the codebase (excluding scripts), used for error handling without a centralized logging service.

## Notable locations

- src/scripts/i18n/**/* — ~50 occurrences
- src/data-layer/**/* — ~10 occurrences
- src/hooks/useLocalStorage.ts — 3 console.warn
- src/hooks/useCentralizedExchanges.ts — ~2
- src/components/EthPriceCard.tsx — console.error
- src/components/Codeblock.tsx — console.error
- src/components/AB/TestWrapper.tsx — console.warn
- src/lib/utils/rss.ts — console.error
- src/lib/utils/md.ts — console.warn/error
- app/api/*/route.ts — Several console.error
- app/[locale]/*/page.tsx — Several console.warn/error

## Impact

Console-based logging is not suitable for production monitoring. Errors may go unnoticed, and there is no structured logging for debugging or alerting.

alerodriargui:
## Description

- Adds a shared `logger` utility that reports warnings/errors to Sentry while keeping console output for local development.
- Migrates a representative set of runtime logging call sites across API routes, AB testing, hooks, UI components, JSON-LD, and markdown/code helpers away from direct `console.warn` / `console.error` calls.
- Leaves scripts and the remaining larger logging cleanup for follow-up work, but establishes the centralized pattern requested by the issue.

## Related Issue

- #18586


pettinarip (changes requested):
@alerodriargui  thanks for picking this up!

Heads up on our side: we're close to our Sentry quota most months, and `logger.warn` → `captureMessage` makes every warning an Issue against it. The missing-`href` warning in `Link.tsx` fires per render, so it'd add up fast. Most of these are also fine as plain `console.*`, server-side they already land in our Netlify logs.

One thing I do see worth it is `CopyPageButton`. The catch only logs today, so the button just resets with an empty clipboard and nobody ever reports it. Your `{ mdPath }` context is exactly right there.

Would you mind trimming down to just that one, calling `Sentry.captureException` directly instead of the wrapper?

wackerow (approved):
Trimmed to scope outlined by @pettinarip -- Updated description to avoid closing related issue. Pulling in

PR #18628, opened 2026-06-29
Make tutorial update script cross-platform

### LINKED ISSUE #18585: Bug: markdown-checker script broken on Windows (cross-platform compatibility)
﻿## Description

The markdown-checker script in package.json uses inline JSON with single quotes (-O "{module: "commonjs"}"), which is not valid on Windows PowerShell.

## Error

```
SyntaxError: Unexpected token """ ""{ is not valid JSON
```

## Impact

Developers on Windows cannot run pnpm markdown-checker. This is a cross-platform compatibility issue in the project.

## Suggested fix

Use a config file instead of inline JSON, or use cross-platform compatible quoting.

alerodriargui:
## Description

- Replaces the inline `ts-node -O` JSON compiler options in `update-tutorials` with a dedicated `tsconfig.scripts.json`.
- Avoids shell-specific JSON quoting so the script runs correctly in Windows PowerShell as well as POSIX shells.

Note: `package.json` no longer contains a `markdown-checker` script, but the same broken inline JSON pattern from the issue is present in the current `update-tutorials` script.

## Related Issue

Fixes #18585

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "chromatic:pages": "chromatic --playwright --zip --exit-zero-on-changes",
     "lint:md": "markdownlint-cli2 \"public/content/**/*.md\" \"!public/content/translations/**\"",
     "lint:md:fix": "markdownlint-cli2 --fix \"public/content/**/*.md\" \"!public/content/translations/**\"",
-    "update-tutorials": "ts-node -O '{ \"module\": \"commonjs\" }' src/scripts/update-tutorials-list.ts",
+    "update-tutorials": "ts-node --project tsconfig.scripts.json src/scripts/update-tutorials-list.ts",
     "prepare": "husky",
     "test": "pnpm test:unit && pnpm test:e2e",
     "test:unit": "USE_MOCK_DATA=true playwright test --project=unit",
--- a/tsconfig.scripts.json
+++ b/tsconfig.scripts.json
@@ -1,0 +1,8 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "commonjs",
+    "moduleResolution": "node"
+  },
+  "include": ["src/scripts/**/*.ts", "src/lib/**/*.ts"]
+}


nloureiro (approved):
Verified empirically: ran `ts-node --project tsconfig.scripts.json src/scripts/update-tutorials-list.ts` in a scratch worktree at this PR's head — exit 0, "Updated src/data/internalTutorials.json with 58 tutorials". Moving the compiler options from inline shell JSON (the actual Windows breakage) into `tsconfig.scripts.json` is the right fix, and the PR body correctly notes the original issue (#18585) referenced a script that no longer exists. CI green.

myelinated-wackerow (changes requested):
Thanks for this — the diagnosis is spot on. The `-O '{ "module": "commonjs" }'` inline JSON genuinely breaks on Windows (neither `cmd.exe` nor PowerShell handle the POSIX single-quote + `\"` escaping), so moving the compiler options out of the shell is the right instinct. A couple of suggestions to make it even cleaner before we merge:

**1. Avoid the new root file — ts-node can read the options from the tsconfig it already loads.**

ts-node natively honors a top-level `ts-node` key in `tsconfig.json`, so instead of a separate `tsconfig.scripts.json`, you can add the override directly to the existing `tsconfig.json` (as a sibling to `compilerOptions`):

```jsonc
{
  "compilerOptions": { /* ... existing ... */ },
  "ts-node": {
    "compilerOptions": {
      "module": "commonjs",
      "moduleResolution": "node"
    }
  }
}
```

This key is read only by ts-node at runtime — `tsc --noEmit` and the Next.js build ignore it — so it's inert outside script execution. Nice bonus: it also repairs a latent inconsistency. The base config uses `moduleResolution: "bundler"`, which is invalid paired with `module: "commonjs"` (bundler resolution requires an ESM `module`), so setting `moduleResolution: "node"` here makes the pair coherent.

The script then simplifies to no shell quoting at all:

```jsonc
"update-tutorials": "ts-node src/scripts/update-tutorials-list.ts"
```

**2. The same pattern lives in two CI workflows.**

`ts-node` is used in exactly three places in the repo, all wanting CommonJS — so once the `ts-node` key is in place, these can drop their inline `-O` too:

- `.github/workflows/update-chains.yml:47` → `npx ts-node ./src/scripts/update-chains.ts`
- `.github/workflows/intl-pipeline.yml:84` → `npx ts-node ./src/scripts/intl-pipeline/main.ts`

(They run on Ubuntu so they don't *break* today, but folding them in kills the brittle inline JSON everywhere and keeps a single source of truth.)

Net result: no new files, cross-platform everywhere, and the fragile inline JSON is gone from all three call sites. Want to give that a go?

_— Opus 4.8 / @wackerow_

wackerow:
Closing in lieu of:
- #18223

PR #18630, opened 2026-06-30
refactor(ui): stablecoins page to latest design system

myelinated-wackerow:
## Summary

Refactors the `/stablecoins` page to current design-system conventions and extends `ui/card` with the interaction props that work required.

### Page (`/stablecoins`)

- Vertical rhythm is driven by a single `.flow` on `MainArticle`; shared section column classes (`mx-auto w-full max-w-7xl px-page`) are hoisted via `*:[section]:`, and spacing uses `space`/`page` tokens instead of hand-rolled margins.
- Content is split into semantic `<Section>` chunks, each with a stable `id`; `scroll-mt` comes from the `Section` primitive (no per-section overrides).
- `h2`s render at `text-h1` size via `**:[h2]:text-h1` (ContentFeedback sits outside `MainArticle`, so it's unaffected).
- The two gradient bands ("choose a stablecoin", "earn interest") are full-bleed, with their content contained at `max-w-7xl`.
- `EdgeScrollContainer` edge spacing tracks `px-page` at every breakpoint, fixing horizontal overflow on mobile.
- The "apps to earn interest" cards are now whole-card links (`href` on `Card`) using `CardButtonFake` instead of a nested `ButtonLink`.
- `tag-green`/`tag-yellow` statuses; decorative tool logos use empty `alt` (their names are the visible titles).
- `StablecoinsTable` is co-located at `app/[locale]/stablecoins/_components/stablecoins-table.tsx`.

### `ui/card`

- Adds the `border` / `hoverOutline` / `hoverLift` boolean props (replacing the old `hoverEffect`); `hoverOutline` auto-enables when the `Card` has an `href`.
- `pathway-card` and the homepage `GetStartedGrid` are migrated to the new props.
- Documents the props in the design-system skill (`card-walkthrough.md`).

## Test plan

- [ ] `/stablecoins` renders correctly in light and dark, mobile through desktop, with no horizontal page scroll.
- [ ] Both horizontal card rows scroll cleanly and their cards align to the `max-w-7xl` column.
- [ ] Whole-card links in "apps to earn interest" navigate; the quiz and the market-cap table still work.
- [ ] Cards using the new `border`/`hoverOutline`/`hoverLift` props (stablecoins, pathway-card, homepage Get Started grid) show the expected resting border and hover treatment.

## Preview link
https://deploy-preview-18630.ethereum.it/stablecoins


src/components/ui/card.tsx
```diff
@@ -36,9 +36,9 @@ const cardVariants = cva(
         sm: "[--card-pad:--spacing(2.5)] [--content-space:--spacing(2.5)]",
         xs: "[--card-pad:--spacing(0)] [--content-space:--spacing(1)]",
       },
-      hoverEffect: {
-        lift: "hover:shadow-md hover:scale-[1.005]",
-      },
+      hoverLift: { true: "hover:shadow-md hover:scale-[1.005]" },
+      hoverOutline: { true: "ring ring-transparent hover:ring-primary-hover" },
```
pettinarip: not sure I see the benefits of this change. I read your comment about having "one-or-the-other", but is that really a problem or something we actually want.

Would we want a card to have both hover effects at the same time or prevent that?

src/components/cards/pathway-card.tsx
```diff
@@ -64,10 +67,10 @@ const PathwayCard = ({
       )}
       <CardContent spacing="none" className="flex items-center gap-4">
         <div className="flex-1">
-          <div className="mb-1 flex flex-wrap items-center gap-x-3 gap-y-1">
+          <div className="mb-1 flex flex-wrap-reverse items-center gap-x-3 gap-y-1">
             <CardTitle>{title}</CardTitle>
             {badge && (
-              <Tag status={badge.status ?? "success"} size="small">
+              <Tag status={badge.status ?? "tag-green"} size="small">
```
pettinarip, line 73: I've just realized we have now "tag-green". What about just using "success"?

this opens other thoughts not related to this PR:
- could we unify and use "success" with new colors instead?
- the `tag-` prefix is a bit redundant

app/[locale]/community/page.tsx
```diff
@@ -157,12 +157,7 @@ export default async function Page(props: { params: Promise<PageParams> }) {
             </div>
             <Grid balanced={4}>
               {cards.map((card, idx) => (
-                <Card
-                  key={idx}
-                  href={card.href}
-                  variant="nested"
-                  hoverEffect="lift"
-                >
+                <Card key={idx} href={card.href} variant="nested" hoverLift>
```
Comment (pettinarip) [inline app/[locale]/community/page.tsx:160]: this one also has the outline on hover but here we just have `hoverLift` 🤔 

Image

pettinarip (approved):
@myelinated-wackerow LGTM, left a few thoughts but not blockers

src/components/cards/pathway-card.tsx
wackerow: Yeah I mentioned this to @konopkja yesterday -- the `success` variant isn't very dark-mode friendly—I'd be down to consider unifying in some form

The tag-_color_ variants were brought in at some point when we simply needed a variety of colors that didn't really correspond to a "success/warn/error" status (and we needed more than the defaults offered).

konopkja: id be for unification

app/[locale]/community/page.tsx
Comment (wackerow) [reply app/[locale]/community/page.tsx:160]: That's from the automatic application of `hoverOutline` if an `href` is supplied to the card

```tsx
hoverOutline: !!href || hoverOutline;
```

As we discussed, given this pattern, it's likely cleaner to just update ui/card to remove that `hoverOutline` "variant" and simply apply that as the default styling for the root `Card` if an `href` is provided, since that's really the only case we want to use this effect.

I'll address that in a separate PR to avoid blocking this one.

PR #18632, opened 2026-06-30
fix typescript errors

### LINKED ISSUE #18583: Tech debt: Fix suppressed @ts-expect-error comments (3 occurrences)
﻿## Description

There are 3 suppressed TypeScript errors using @ts-expect-error that should be properly fixed:

1. **src/config/rainbow-kit.ts:80** — "// @ts-expect-error - TODO: fix this" — Type error in RainbowKit configuration
2. **src/components/LanguagePicker/index.tsx:58** — "// @ts-expect-error" — Generic type mismatch
3. **app/[locale]/[slug]/page.tsx:60** — "// @ts-expect-error Incompatible component function signatures" — Component typing issue

## Impact

These suppress actual type errors, which could lead to runtime bugs. They should be properly typed rather than silenced.

alerodriargui:
## Description

Removes the remaining suppressed TypeScript errors from the RainbowKit and language picker code paths.

- Types `getTargetChains()` as a non-empty chain tuple for RainbowKit configuration.
- Adds a chain-name type guard when reading `NEXT_PUBLIC_CHAIN_NAMES`.
- Types `LocaleDisplayInfo.localeOption` as `Lang`.
- Updates the language picker navigation to use the current localized pathname with a validated selected locale instead of suppressing the `next-intl` route type mismatch.
- Removes now-redundant locale casts in `useLanguagePicker`.

## Related Issue

Fixes #18583


nloureiro (changes requested):
The substance here is good and I could verify it, but CI is genuinely red:

**Blocking:** ESLint fails on `src/config/rainbow-kit.ts:1` — the new `import type { Chain } from "viem"` isn't sorted (`simple-import-sort/imports`). Run `pnpm lint:fix` and push. (The CI type-check step never ran because ESLint failed first.)

**Verified while reviewing** (local `tsc --noEmit` at this PR's head passes):
- `LocaleDisplayInfo.localeOption: Lang` is safe — the only producer already takes `Lang`.
- `getTargetChains(): [Chain, ...Chain[]]` with the `isChainName` guard properly fixes the one suppression issue #18583 confirmed still valid.
- The LanguagePicker change is the documented next-intl pattern and safe at runtime (`routing.ts` has no `pathnames` config, so `usePathname()` + `push(pathname, { locale })` is correct).

One import-order fix away from approvable.

wackerow (approved):
Looks good! Got it up to speed with dev and cleaned up a couple tiny things. Thanks @alerodriargui

PR #18634, opened 2026-06-30
fix: improve staking considerations accessibility

### LINKED ISSUE #18582: Accessibility: StakingConsiderations needs a11y and mobile responsive improvements
﻿## Description

In src/components/Staking/StakingConsiderations/index.tsx there are two known issues:

1. **Line 72**: "// Improve a11y" — Accessibility improvements needed
2. **Line 76**: "// Make mobile responsive" — Not properly responsive on mobile devices

## Impact

Users with assistive technologies may have difficulty using this component. Mobile users may have a poor experience.

alerodriargui:
## Description

Improves accessibility and mobile responsiveness for the StakingConsiderations component.

- Converts the desktop considerations list into an accessible tab interface.
- Adds `tablist`, `tab`, `tabpanel`, `aria-selected`, `aria-controls`, and panel labelling.
- Adds keyboard navigation with Arrow Up, Arrow Down, Home, and End.
- Uses real buttons for selectable tab items while preserving the existing visual styling.
- Makes the mobile dropdown full width.
- Improves mobile layout by reducing the panel minimum height, using responsive padding, and stacking indicator groups on small screens.
- Marks decorative icons as `aria-hidden`.
- Removes the stale a11y and mobile responsive TODO comments.

## Related Issue

Fixes #18582

### DIFF
--- a/src/components/Staking/StakingConsiderations/index.tsx
+++ b/src/components/Staking/StakingConsiderations/index.tsx
@@ -1,8 +1,8 @@
 "use client"
 
-import { useState } from "react"
+import { type CSSProperties, type KeyboardEvent, useId, useState } from "react"
 
-import type { StakingPage } from "@/lib/types"
+import type { MatomoEventOptions, StakingPage } from "@/lib/types"
 
 import ButtonDropdown from "@/components/ButtonDropdown"
 import {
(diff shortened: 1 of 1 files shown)


nloureiro (approved):
Verified the ARIA rework: `role="tablist"` + `aria-orientation="vertical"`, `role="presentation"` on `li`s, real `<button role="tab">` with `aria-selected`, roving `tabIndex`, shared `aria-controls`, and Up/Down/Home/End keyboard nav — matches the WAI-ARIA vertical-tabs pattern. RTL-safe (logical properties; vertical arrows are direction-agnostic). CI fully green.

Non-blocking nits + one flag for the merging maintainer:
1. `selectConsideration` now fires the Matomo event on every arrow-key traversal (selection-follows-focus) — consider tracking only click/Enter to avoid inflating analytics.
2. APG recommends `tabIndex={0}` on the tabpanel when it has no focusable content (this one has only h3/p/indicators).
3. **The PR bundles deliberate visual changes beyond a11y** (panel `min-h-96`→`min-h-80`, mobile `p-4`, indicator groups stacked, `mt-auto`→`mt-8` below md) — it resolves the old mobile-responsive TODO, but wants a Chromatic/visual look before merge, not just a11y sign-off.

wackerow:
Appreciate the go at this @alerodriargui, but a couple things:

- Leading with the reality that this entire page (and in particular, this component) is being refactored within the next week or so. At that point, all of this will be moot. 
- For these changes as they are, they could do the trick, but they're a touch over-engineered when we should be relying more on the Radix framework that can help handle a lot of this logic for us. 

Given these two points, I'm going to close this one out for now, but in the event anything changes in the plans, we can always reopen this and work to get the patch in. Thanks again!

PR #18636, opened 2026-06-30
fix: stabilize product table filter state

### LINKED ISSUE #18581: Bug: ProductTable guard removed to avoid infinite re-renders
﻿## Description

In src/components/ProductTable/index.tsx:79, there is a comment indicating that a guard was removed specifically to avoid infinite re-renders:

"// removed to avoid infinite re-renders"

## Impact

This is a workaround that masks an underlying re-render issue. The root cause (why infinite re-renders occur) may still exist and could cause performance issues or unexpected behavior in other scenarios.

## Suggested action

Investigate the root cause of the infinite re-renders and add a proper fix rather than removing the guard.

alerodriargui:
## Description

Fixes the ProductTable filter synchronization workaround by addressing the unstable filter inputs that could cause repeated renders.

- Adds semantic comparison for `initialFilters` before syncing ProductTable state.
- Reintroduces guarded filter state synchronization without relying on array identity.
- Updates URL query hydration to use the current filter state and removes the exhaustive-deps workaround.
- Stabilizes `useTranslation()` so its returned `t` function can be used safely in memo dependencies.
- Memoizes wallet and layer 2 filter definitions to avoid recreating filter arrays every render.

## Related Issue

Fixes #18581

### DIFF
--- a/src/components/FindWalletProductTable/hooks/useWalletFilters.tsx
+++ b/src/components/FindWalletProductTable/hooks/useWalletFilters.tsx
@@ -1,4 +1,4 @@
-import { useRef } from "react"
+import { useMemo, useRef } from "react"
 import { Brain, LayersPlus } from "lucide-react"
 import { useLocale } from "next-intl"
 
@@ -21,12 +21,13 @@ export const useWalletFilters = (): FilterOption[] => {
   const locale = useLocale()
   const { t } = useTranslation("page-wallets-find-wallet")
   const prevNetworkArray = useRef<string[]>([])
-  return [
-    {
-      title: t("page-find-wallet-device"),
-      showFilterOption: true,
-      items: [
-        {
+  return useMemo(
+    () => [
+      {
+        title: t("page-find-wallet-device"),
+        showFilterOption: true,
+        items: [
+          {
           filterKey: "mobile",
           filterLabel: t("page-find-wallet-mobile"),
           description: "",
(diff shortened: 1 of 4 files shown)


nloureiro:
Reviewed in depth — not approving yet, but not blocking either; this needs a product/QA decision more than code changes:

**What checks out**: the refactor is sound. The reset is value-keyed (`getInitialFiltersKey` + `useRef` guard) so it's correct even where memoization is ineffective; the mount-hydration switch to a functional `setFilters` update is strictly better; the `useMemo` deps cover everything captured. CI green.

**What the reviewer should know**: the instability described in #18581 doesn't reproduce on `dev` — the unstable arrays only feed `useState` (ignored after mount) and a `useCallback` dep, causing re-renders but not incorrect filter state. So this is a preventive refactor, not a bug fix.

**Real behavior change needing manual QA before merge**: if `initialFiltersKey` changes at runtime (e.g. locale switch without remount under App Router), **all user filter selections including URL-hydrated state are wiped**. Arguably correct (labels must re-translate), but please QA: apply filters + URL params on `/wallets/find-wallet`, switch language, confirm no reset loop and that the reset semantics are acceptable.

**Two asks**: (1) move the per-render `JSON.stringify` of the ~1050-line filter structure into `useMemo` — it currently runs on every keystroke; (2) `pnpm format` — the filter bodies weren't re-indented inside the new `useMemo` wrappers.

wackerow:
@pettinarip With the recent find-wallet changes, is this anything that would still apply or should it be closed out?

PR #18638, opened 2026-06-30
fix: narrow tsconfig include patterns

### LINKED ISSUE #18578: tsconfig include patterns override node_modules exclusion, causing ~700 noise TS errors
## Description

The tsconfig.json include patterns (**/*.ts, **/*.tsx) override the exclude: [node_modules] setting, causing TypeScript to check .ts files inside node_modules.

## Impact

~700 noise errors from node_modules/ox including:
- TS2737: BigInt literals not available when targeting below ES2020
- TS2791: Exponentiation cannot be performed on bigint values unless target is ES2016+
- TS2322, TS2769 in various ox library files

## Root cause

The include array has **/*.ts and **/*.tsx which are redundant with ./src/**/* but also accidentally include node_modules.

## Suggested fix

Remove the **/*.ts and **/*.tsx patterns from include (they are redundant with ./src/**/*), or add node_modules/** to the exclude list.

alerodriargui:
## Description

Narrows the TypeScript project include patterns to avoid checking TypeScript files inside `node_modules`.

- Removes broad `**/*.ts` and `**/*.tsx` include patterns.
- Adds explicit project scopes for `app`, `tests`, root TypeScript files, and existing generated Next.js types.
- Keeps `src` coverage unchanged.
- Preserves the existing `node_modules` exclusion.

## Related Issue

Fixes #18578

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -42,9 +42,11 @@
   },
   "include": [
     "./src/**/*",
+    "./app/**/*",
+    "./tests/**/*",
     "next-env.d.ts",
-    "**/*.ts",
-    "**/*.tsx",
+    "./*.ts",
+    "./*.tsx",
     ".next/types/**/*.ts",
     ".next/dev/types/**/*.ts"
   ],


nloureiro:
Reviewed with an empirical check: `tsc --listFilesOnly` under the old and new `include` produces **byte-identical file sets** (6,331 files each) — so this changes nothing about type-check coverage. The description's premise doesn't hold: `node_modules` is already excluded (tsconfig.json:52), and the node_modules `.d.ts` files in the program come from import resolution, which `include` can't affect (the paired issue #18578 was closed as disproven — `tsc` on dev is clean).

It also introduces a small future footgun: today `**/*.ts` catches any TS file added outside `src/`/`app/`/`tests`/root (a future `scripts/` or `netlify/` dir); after this change such files silently skip type-checking.

Recommendation for the maintainers: close as a no-op, or merge only under an explicit hygiene framing with the footgun accepted knowingly. Nothing is broken — CI is green — but it shouldn't land labeled as a fix.

nloureiro:
Closing as a verified no-op, per the review above: `tsc --listFilesOnly` produces byte-identical file sets (6,331 files) under the old and new `include` patterns, the premise about node_modules doesn't hold (`exclude` already covers it; the `.d.ts` files in the program come from import resolution, which `include` can't affect), and the narrowed globs would let future TS files outside `src`/`app`/`tests` silently skip type-checking. The paired issue #18578 was closed earlier for the same reason. Thanks for the audit sweep — several of its other PRs are landing.

PR #18640, opened 2026-06-30
Handle stablecoins market data errors

### LINKED ISSUE #18576: Bug: marketsHasError variable set but never handled in stablecoins page
## Description

In \pp/[locale]/stablecoins/page.tsx\, the variable \marketsHasError\ is set at lines 90 and 130 but the error is never handled or displayed to the user.

## Impact

If the stablecoin markets API fails, the error is silently swallowed and users may see incomplete/incorrect data without any feedback.

## Code references

/app/[locale]/stablecoins/page.tsx:90\ and \:130\

## Expected behavior

When \marketsHasError\ is true, an error state or fallback UI should be shown to the user.

alerodriargui:
## Description

Fixes the stablecoins market data error state by showing an explicit error alert when the stablecoin markets API fails.

`marketsHasError` is already passed from the stablecoins page into `StablecoinsTable`; this updates the table to render an accessible `Alert` with the existing translated error message instead of leaving the failure as an understated table row.

## Related Issue

Fixes #18576

### DIFF
--- a/src/components/StablecoinsTable.tsx
+++ b/src/components/StablecoinsTable.tsx
@@ -8,6 +8,7 @@ import { Image } from "@/components/Image"
 
 import { cn } from "@/lib/utils/cn"
 
+import { Alert } from "./ui/alert"
 import { Button } from "./ui/buttons/Button"
 import { Flex } from "./ui/flex"
 import InlineLink from "./ui/Link"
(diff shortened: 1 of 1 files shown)


nloureiro:
This one has merge conflicts with `dev` — could you rebase? Small, sensible error-state fix; happy to review once it's green.

wackerow:
Appreciated @alerodriargui but going to close this one out.. Couple things

- This has been refactored significantly now on `dev` branch and issue has been resolved
- A translated error is already handled by the server here
- `role="alert"` -- This wouldn't really be appropriate here. This would be more appropriate for cases with screen readers and forms, where a user needs to be alerted that there was an error with an action they took, but this is simply data fed from the server and would never be announced, since these would only announce when something changes that triggers it.

Cheers!

PR #18641, opened 2026-06-30
Fix rehype image buffer hashing type

### LINKED ISSUE #18575: TypeScript error: Buffer type incompatibility in rehypeImg.ts and crowdin/files.ts
## Description

Two TypeScript errors caused by Buffer<ArrayBufferLike> not being assignable to BufferSource/BodyInit.

## Files affected

1. **src/lib/md/rehypeImg.ts:106** — \TS2345\: \Buffer<ArrayBufferLike>\ is not assignable to \BufferSource\
2. **src/scripts/i18n/lib/crowdin/files.ts:330** — \TS2769\: No overload matches for \etch()\ call. \Buffer<ArrayBufferLike>\ not assignable to \BodyInit\

## Root cause

Node.js \Buffer\ type incompatibility with newer \@types/node\. The \Buffer<ArrayBufferLike>\ type from newer Node.js typings doesn't match the older \BufferSource\ union type.

## Steps to reproduce

\\\
npx tsc --noEmit
\\\

## Expected fix

Cast to appropriate types or update to use \Uint8Array\ instead of \Buffer\ where possible.

alerodriargui:
## Description

Fixes the TypeScript `Buffer<ArrayBufferLike>` incompatibility in `rehypeImg.ts` by converting the file `Buffer` to a `Uint8Array` before passing it to `getHashFromBuffer`.

The original `Buffer` is still used for `getPlaiceholder`, while the hash path now uses a standard `BufferSource`-compatible byte array.

Note: the issue also references `src/scripts/i18n/lib/crowdin/files.ts`, but that file does not exist in the current repository tree.

## Related Issue

Fixes #18575

### DIFF
--- a/src/lib/md/rehypeImg.ts
+++ b/src/lib/md/rehypeImg.ts
@@ -108,9 +108,10 @@ const setImagePlaceholders = async (
 
     // Load image data from file system as buffer
     const buffer: Buffer = fs.readFileSync(path.join("public", src))
+    const imageBytes = Uint8Array.from(buffer)
 
     // Get hash fingerprint of image data (no security implications; fast algorithm prioritized)
-    const hash = await getHashFromBuffer(buffer, {
+    const hash = await getHashFromBuffer(imageBytes, {
       algorithm: "SHA-1",
       length: 8,
     })


nloureiro (approved):
Verified: `getHashFromBuffer` takes `BufferSource`, and `Uint8Array.from(buffer)` resolves the Node-Buffer-vs-web-typings mismatch with a copy — negligible cost in a build-time placeholder path. Ready to merge.

wackerow (approved):
Lgtm, thanks @alerodriargui!

PR #18645, opened 2026-06-30
Derive SideNav open state from active route

### LINKED ISSUE #18642: Tech debt: SideNav parent sections do not derive open state from active child route
## Description

`src/components/SideNav.tsx` has a TODO noting that the tree state should be based on whether the current path is a child of a parent section.

Currently, top-level `path` items default to open and only collapse after user interaction. This means the side navigation state is not fully derived from the active route hierarchy.

## Code reference

`src/components/SideNav.tsx`:

```ts
// TODO set tree state based on if current path is a child
// of the given parent. Currently all `path` items default to open
// and they only collapse when clicked on.
```

## Impact

The developer docs side navigation can show more sections expanded than necessary, which makes long navigation trees harder to scan. It can also make active-route behavior inconsistent between direct page loads and user-driven collapses.

## Suggested action

Update `SideNav` so parent sections derive their initial and route-change open state from whether the current path matches one of their descendants, while preserving explicit user toggles where appropriate.

alerodriargui:
## Description

Fixes the developer docs side navigation so parent sections derive their open state from the active route hierarchy instead of opening every top-level section by default.

This adds recursive route matching for nav items and removes the stale TODO about deriving tree state from child routes. The SideNav now opens only the section branches that contain the active page, while preserving user-controlled expand/collapse behavior between route changes.

## Related Issue

Fixes #18642


nloureiro (approved):
Verified: manual toggle UX is preserved — the effect deps are `[isLinkInPath]`, so it fires only when the route relationship changes; a user can still collapse a section while on a page inside it. Path matching is stricter and safer than the old substring `includes` (normalized trailing slash, anchored `startsWith`), and dropping the `item.path` check is fine since directory entries are detected via recursion into children hrefs. Resolves the TODO at SideNav.tsx:132. CI green.

One deliberate UX change to sign off on before merging: previously all top-level sections defaulted open on every docs page; now **only the section containing the current page opens**, and on `/developers/docs/` itself everything starts collapsed. That's the TODO's stated intent, but it visibly changes the docs sidebar everywhere — expect visual-suite churn and confirm the default-collapsed landing state is wanted.

pettinarip (approved):
Resolved the merge conflict in `src/components/SideNav.tsx` by keeping this PR's recursive active-route matching (`isItemInPath`), adapted to the `useTranslations` (next-intl) API that landed on `dev`. The recursive approach correctly opens only the section branches containing the active page, which is the intended fix for #18642. Type-check passes. Thanks @alerodriargui!

PR #18662, opened 2026-07-01
fix: use console.warn instead of console.log in useLocalStorage catch block

### LINKED ISSUE #18658: Fix: Use console.warn instead of console.log in useLocalStorage catch block
## Description

In \src/hooks/useLocalStorage.ts\, the \catch\ block that handles \JSON.parse\ failures uses \console.log\ to log the error. Error conditions should use \console.warn\ or \console.error\ so they are visually distinct in developer tools and can be filtered properly.

## Location
\src/hooks/useLocalStorage.ts:104\

## Suggested fix
Replace \console.log\ with \console.warn\ on line 104.

alerodriargui:
Closes #18658

## Description

Replaces \console.log\ with \console.warn\ in the \JSON.parse\ catch block in \src/hooks/useLocalStorage.ts\. Error conditions should use \console.warn\ or \console.error\ so they are visually distinct and can be filtered properly.

### DIFF
--- a/src/hooks/useLocalStorage.ts
+++ b/src/hooks/useLocalStorage.ts
@@ -101,7 +101,7 @@ function parseJSON<T>(value: string | null): T | undefined {
   try {
     return value === "undefined" ? undefined : JSON.parse(value ?? "")
   } catch {
-    console.log("parsing error on", { value })
+    console.warn("parsing error on", { value })
     return undefined
   }
 }


nloureiro:
Closing in favor of #18677, which edits the same lines of `useLocalStorage.ts` — removing the log entirely rather than promoting it to `console.warn` (a parse failure here already fails safe, so there's nothing actionable to warn about). The corresponding issue #18658 was closed as a duplicate of #18674 for the same reason. Thanks — the pair just needed deduping.

PR #18666, opened 2026-07-02
perf(netlify): trim SSR function bundle 211MB → 115MB unzipped

pettinarip:
The Netlify SSR function (`___netlify-server-handler`) is capped at **250MB unzipped**. It had grown as pages were added. This trims it **211MB → 115MB** (measured via local `netlify build`), no runtime behavior change.

## Changes
- **Drop the vestigial `developers/tutorials` file-tracing include (~37MB).** The hub and every tutorial page render `revalidate:false` (fully prerendered, CDN-served), so the translated MD is read at build only. Individual tutorial pages already render include-free under the same pattern. Verified the hub still prerenders after removal.
- **Exclude `public` mp4/videos (~5.4MB)** — mp4 was missing from the media exclude list.
- **Exclude dev-only root files (~3.5MB)** — `tests`, `docs`, `README.md`, `tsconfig.tsbuildinfo`, `.all-contributorsrc`, `deno.lock`, `AGENTS.md`.
- **Dedup `sharp` 0.34.3 → 0.34.5 (~16MB).** `plaiceholder` pinned 0.34.3 while `next` pulls 0.34.5, installing two full sharp/libvips versions. Matching `next` drops one (`plaiceholder` accepts `>=0.30.6`).
- **Exclude sharp's musl binaries.** Dead on glibc Lambda; the Netlify plugin already strips them in CI, so this just makes local measurements match production.

## Fix: anchored `docs/**` `tests/**` (post-review)
The bare `"docs"`/`"tests"` excludes matched as **substrings** during Turbopack's trace pruning, so `"docs"` also matched the `124~_@docsearch_react...` SSR chunk and dropped it from the function — ISR pages using DocSearch (e.g. `/developers`) 502'd with `ChunkLoadError` on on-demand re-render. Anchored to `"docs/**"` / `"tests/**"` (they match the repo dirs, not `@docsearch`). Verified: chunk restored (54 trace refs, matching base), repo dirs still excluded, bundle unchanged. Reproduced and fixed via `revalidate` testing on the deploy preview.

## Notes
- `sharp` is build-time only (image opt is offloaded to Netlify's Image CDN; `plaiceholder` runs during MD compile), so we keep only its glibc binary.
- Production is ~33MB lower than the local numbers above, since the plugin's musl strip is CI-only (~178MB → ~116MB, 46% of limit).
- Not here: `shiki` bundles all languages (~7MB); fully excluding sharp (~16MB) needs `plaiceholder`'s import made lazy.
- Separately worth tracking: Turbopack production trace pruning matching excludes as substrings is a sharp edge (this is why the bare `docs` bit us). Force-revalidating build-time-static content pages (`/videos`, `/stories`, `/developers/tutorials`) is also unsupported (content excluded from the function) — refresh those via redeploy.


wackerow (approved):
🔥

PR #18671, opened 2026-07-02
refactor(ui): bake href hover outline into Card

myelinated-wackerow:
## Description

Follow-up to [this discussion on #18630](https://github.com/ethereum/ethereum-org-website/pull/18630#discussion_r3515142162): the hover outline was only ever wanted on linkable cards, and `href` already auto-enabled it, so the manual `hoverOutline` variant was redundant surface area.

- Removes the `hoverOutline` variant from `ui/card`; the ring (`ring ring-transparent hover:ring-primary-hover`) is now applied directly in the `href` render path, so every linkable Card gets the affordance and it can no longer be set on non-link cards
- `border` still composes: `ring-border` at rest, brightening to `ring-primary-hover` on hover
- `hoverLift` remains the opt-in for cards that hold their own actions (multiple buttons or text-link actions) where the card itself isn't clickable
- Drops the now-redundant `hoverOutline` from `PathwayCard` and migrates `GetStartedGrid` from a manual `border transition-colors hover:border-primary-hover` class chain to the `border` prop + automatic ring
- Updates the design-system skill docs to match (and fixes stale `hoverEffect="lift"` references)

## Test plan

- [x] `GetStartedGrid` and `PathwayCard` cards show the same rest border and primary hover ring as before
- [x] Non-link Cards render no hover ring


pettinarip (approved):
Nice!

PR #18672, opened 2026-07-02
feat(ui): intrinsic sizing for markdown video

myelinated-wackerow:
## Description

Markdown videos (``) were forced into a fixed 16:9 or 9:16 box with `object-contain`, so any clip that wasn't exactly the standard ratio letterboxed inside the box — and the rounded corners from #18501 landed on the empty box edges instead of the clip itself.

This sizes the `<video>` element to the clip's own intrinsic aspect ratio so the box hugs the video: landscape clips fill the content width, portrait clips stay height-capped (mirroring the YouTube Shorts embed). With the box accurate to the content, `rounded-base` corners now always land on the video.

Implementation:

- `rehypeImg` now probes video dimensions at build time via a small pure-JS reader of the mp4/mov `tkhd` box (`image-size` can't read video, and `ffprobe` isn't guaranteed in the build image). Results are memoized per build, keyed by path + size + mtime so `next dev` edits aren't served stale.
- `MarkdownVideo` sets an explicit CSS `aspect-ratio` from the probed dimensions so layout space is reserved up front (no CLS). The `width`/`height` attributes alone are unreliable for this on `<video>`.
- Fail-safe throughout: any parse issue, unsupported container (webm etc.), remote source, or oversized file falls back to intrinsic CSS sizing without breaking the build.

Stacked on #18501 (markdown rounded corners); diff shown here is just the video-sizing commit.

## Related Issue

None — follow-up to #18501.

### DIFF
--- a/src/components/Image/MarkdownImage.tsx
+++ b/src/components/Image/MarkdownImage.tsx
@@ -10,9 +10,12 @@ import { toPosixPath } from "@/lib/utils/relativePath"
 import { CONTENT_IMAGES_MAX_WIDTH } from "@/lib/constants"
 
 interface MarkdownImageProps extends Omit<ImageProps, "width" | "height"> {
-  width: string
-  height: string
-  aspectRatio: string
+  // Optional: `rehypeImg` omits these when dimensions can't be probed (e.g. a
+  // video whose container it can't read), in which case the renderer falls back
+  // to intrinsic sizing.
+  width?: string
+  height?: string
+  aspectRatio?: string
 }
 
 const MarkdownImage = ({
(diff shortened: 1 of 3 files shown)


pettinarip (approved):
Approving. It works and it's already done, so no reason to block. 👍

One thing though: feels like a lot of code complexity just to get rounded corners on odd-ratio clips… which we basically don't have yet (?). For the few clips we'll realistically have, a filename/frontmatter convention would've done the job with way less to maintain IMO. Something like `` to inherit the not supported ratio.

wackerow:
Closed in lieu of:
- #18744

PR #18677, opened 2026-07-03
Remove localStorage parse debug log

### LINKED ISSUE #18674: Remove debug console.log from localStorage JSON parsing helper
## Description

`src/hooks/useLocalStorage.ts` logs every JSON parse failure from `parseJSON`:

```ts
} catch {
  console.log("parsing error on", { value })
  return undefined
}
```

This hook is shared UI infrastructure, so a malformed localStorage value can produce noisy browser console output and expose the raw stored value in logs. The surrounding hook already falls back safely when localStorage reads fail.

## Suggested fix

Remove the debug `console.log`, or replace it with a guarded warning that does not include the raw localStorage value.

While touching this file, the warning strings above this helper also appear to contain mojibake smart quotes (`â€œ` / `â€`) and could be normalized to plain quotes.

## File

- `src/hooks/useLocalStorage.ts`

alerodriargui:
## Description

Removes the debug `console.log` from the shared `useLocalStorage` JSON parsing helper. Malformed localStorage values still fall back to `undefined`, but the raw stored value is no longer printed to the browser console.

Also fixes the nearby helper comment typo while touching the same small block.

Closes #18674

## Validation

- Compared the branch against `ethereum:dev`; the PR contains one commit and only changes `src/hooks/useLocalStorage.ts`.
- No local test run: `git`/`gh` are not available in this shell, and this is a small console cleanup.

### DIFF
--- a/src/hooks/useLocalStorage.ts
+++ b/src/hooks/useLocalStorage.ts
@@ -96,12 +96,11 @@ export function useLocalStorage<T>(
   return [storedValue, setValue]
 }
 
-// A wrapper for "JSON.parse()"" to support "undefined" value
+// A wrapper for "JSON.parse()" to support "undefined" value
 function parseJSON<T>(value: string | null): T | undefined {
   try {
     return value === "undefined" ? undefined : JSON.parse(value ?? "")
   } catch {
-    console.log("parsing error on", { value })
     return undefined
   }
 }


pettinarip (approved):

nloureiro (approved):
Verified: removes the debug log in `parseJSON`'s catch (parse failures already fail safe by returning `undefined`) and fixes the doc-comment typo. Preferred over #18662, which touches the same lines but keeps the logging — closing that one in favor of this. Note: the red CI checks are **cancelled** runs from the bulk-push queue, not failures — needs a re-run before merge.

PR #18686, opened 2026-07-04
fix: improve markdown video accessibility controls

### LINKED ISSUE #18682: a11y: markdown video (MarkdownVideo) accessibility gaps
Markdown video clips (`MarkdownVideo`, rendered from ``) have a few accessibility gaps worth addressing. Surfaced during review of the markdown video sizing work (#18672); these are pre-existing in the component, so tracking them separately.

- **Pause control (WCAG 2.2.2 Pause, Stop, Hide):** clips `loop` and autoplay while in view, running well past 5s, but controls are exposed only under `prefers-reduced-motion`. Users without that setting have no way to pause/stop the motion. Fix: always-available pause/controls, or stop after a single pass under 5s.
- **No poster / blank first frame:** clips render empty until in-view playback starts (`preload="metadata"`, no poster passed). A build-time first-frame poster would remove the blank flash and also give the reduced-motion branch something to show instead of a bare control bar.
- **Missing playback opt-outs:** no `disablePictureInPicture` / `disableRemotePlayback` (and `controlsList`), so browsers offer PiP/cast affordances for small decorative UI clips.

Component: `src/components/Image/MarkdownVideo.tsx`

alerodriargui:
## Description

Addresses #18682

Updates `MarkdownVideo` so markdown-authored video clips expose playback controls for everyone, not only users with `prefers-reduced-motion`. This gives users a direct way to pause looping motion while preserving the existing in-view autoplay behavior for users who have not requested reduced motion.

The video element now also:

- preloads video data instead of metadata only, improving the initial/reduced-motion frame before playback starts
- disables picture-in-picture and remote playback affordances for these small markdown clips
- hides download/remote playback options through `controlsList`

## Validation

- Verified the remote branch diff is a single-file change to `src/components/Image/MarkdownVideo.tsx`
- Verified fork branch compare is `ahead_by: 1` and `behind_by: 0` from upstream `dev` base commit `db3a095448ed1b3e67f1a63d086f769f4b0c0c84`
- Could not run `pnpm lint` or `pnpm type-check` locally because `pnpm` is not installed in this PowerShell environment


src/components/Image/MarkdownVideo.tsx
```diff
@@ -64,9 +64,12 @@ const MarkdownVideo = ({
         loop
         muted
         playsInline
-        preload="metadata"
+        preload="auto"
```
pettinarip: lets keep `metadata` as the default

pettinarip (changes requested):

nloureiro (changes requested):
The motivation is legitimate — an autoplaying looping clip >5s with no visible pause for non-reduced-motion users is a real WCAG 2.2.2 gap that #18672 doesn't address — but as written this needs rework:

1. **Rebase onto #18672 first**: that PR is approved and rewrites the same doc-comment block and `<video>` attribute region of `MarkdownVideo.tsx`, and the two diverge semantically (#18672 keeps `preload="metadata"` + `controls={showControls}`; this deletes `showControls` and forces `controls`). Guaranteed conflict once it merges.
2. **`preload="auto"` is a bandwidth regression**: every markdown clip downloads in full at page load, even offscreen, defeating the in-view play gating — and the `poster` already covers first-frame display. Please revert to `metadata`.
3. **User pause is silently overridden**: the play effect calls `video.play()` whenever the clip scrolls back into view, so a user who pauses via the newly always-visible controls gets the video restarted — defeating the control this PR adds (2.2.2 requires the pause to stick). Needs a "user paused" flag that suppresses the auto-play branch.

Also note always-visible control chrome changes the look of every markdown video — wants visual QA. Happy to re-review after the rebase + those two fixes.

wackerow (approved):
@nloureiro and @pettinarip Fixed your above points. To the UI note: testing in Brave, Safari and Firefox the UI controls hide unless hovering.

wackerow:
Note this doesn't fix the `poster` part of the related issue. Removed the auto-close comment from the description so the issue stays open.

PR #18696, opened 2026-07-06
chore: let the stale bot close issues and PRs after a 30-day warning

nloureiro:
## Description

The stale workflow has labeled items `Status: Stale` since 2021 but never closed anything (`days-before-close: -1`). That's how 106 stale items accumulated (oldest from 2022) before the July 2026 backlog sweep closed them by hand. This makes the bot finish the job so the backlog doesn't regrow:

- **30 days silent → stale warning → 30 more days → auto-close** (60 days total), with messages that explain exactly how to keep an item open and that closing is not a judgment on the idea.
- **Exempt labels** for long-horizon work: `pinned 📌` (new — just created), `good first issue`, `help wanted`, `epic 💪`. GFIs are exempt so the newly restocked good-first-issue list (22 issues) doesn't get culled.
- Closes issues as `not_planned` so they don't read as completed.
- Upgrades `actions/stale` v3 → v9 (input names compatible) and adds the explicit `permissions` block v9 expects.
- Fixes the PR warning message, which previously said "This issue…".

## Things to decide in review

- Are the four exempt labels the right set? (e.g. should `awaiting PR` or `request for comments 🗣️` be exempt too?)
- 30+30 days is the proposed cadence — easy to tune.

Part of the backlog-hygiene follow-ups from the 2026-07-06 triage.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -5,18 +5,45 @@ on:
     - cron: "0 8 * * *"
   workflow_dispatch:
 
+permissions:
+  issues: write
+  pull-requests: write
+
 jobs:
   stale:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/stale@v3
+      - uses: actions/stale@v9
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           stale-issue-label: "Status: Stale"
           stale-pr-label: "Status: Stale"
-          stale-issue-message: "This issue is stale because it has been open 30 days with no activity."
-          stale-pr-message: "This issue is stale because it has been open 30 days with no activity."
+          stale-issue-message: >
+            This issue is stale because it has been open 30 days with no
+            activity. It will be closed in 30 days unless there is new
+            activity, the stale label is removed, or an exempt label
+            (`pinned 📌`, `good first issue`, `help wanted`, `epic 💪`) is
+            added.
+          stale-pr-message: >
+            This pull request is stale because it has been open 30 days with
+            no activity. It will be closed in 30 days unless there is new
+            activity, the stale label is removed, or the `pinned 📌` label is
+            added.
+          close-issue-message: >
+            This issue was closed automatically after 60 days without
+            activity. This keeps the issue list actionable — it is not a
+            judgment on the idea. If it is still relevant, please leave a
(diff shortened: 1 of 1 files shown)


wackerow:
Not sure recommend this -- there are plenty of cases where things go stale for a couple months and get revisited and merged. If they get closed, they'll easily become invisible and forgotten.

nloureiro:
Fair concern — closed items do lose visibility, and "stale then revived and merged" is a real pattern here. Rather than push this as-is, here are the two configs I'd offer, pick whichever matches your instinct:

**A. Issues-only close** (`days-before-close: 30` for issues, `-1` for PRs): PRs keep the current label-only behavior — nothing that could still be revived-and-merged ever auto-closes — while issue rot (which was 90% of the backlog problem: listing requests and reports nobody will revisit) still gets cleaned. The revive-and-merge cases you're describing are almost always PRs.

**B. Label-only everywhere, keep the rest**: drop the auto-close entirely and keep this PR's other fixes (v9 upgrade, permissions block, corrected PR message, exempt labels for when you do want closing later). The counterweight would be periodic maintainer/agent sweeps like the one that just closed ~70 rotted items by hand — that works, but it's exactly the manual labor this tried to automate.

My lean is **A** — the July sweep found the oldest stale *issue* was from 2022 with zero chance of revival, while several stale *PRs* were indeed worth salvaging, which matches your concern precisely. Happy to push either config; which do you prefer?

wackerow:
I'm not sure I'd auto-close issues either, but if we want to consider it, I'd push it well beyond 30 days. Again there are plenty of cases that these lower priority issues sit, and can be picked up at later dates... If we really want to prevent things from going very stale then I may suggest an auto close after at least 6 months if not a year.

nloureiro:
> I'm not sure I'd auto-close issues either, but if we want to consider it, I'd push it well beyond 30 days. Again there are plenty of cases that these lower priority issues sit, and can be picked up at later dates... If we really want to prevent things from going very stale then I may suggest an auto close after at least 6 months if not a year.

Thanks. I’ve updated the workflow to keep PRs label-only and move issue auto-close to a 180-day total window: 
issues are marked stale after 30 days, then closed 150 days later >>> @wackerow want to make it 360days?
PRs are never auto-closed.

nloureiro:
@wackerow, the bot said it looks mergeable 😁

PR #17155, opened 2026-01-23
Accessibility Violations - Add Accessible Labels to Swiper Navigation…

huangkevin-apr:
This PR fixes accessibility violations in the Swiper component where the previous/next navigation buttons lack accessible labels, making them unusable for screen reader users.

## Description

The swiper navigation buttons (`SwiperPrevButton` and `SwiperNextButton`) are icon-only controls without any text alternatives. While they may be visually clear through directional arrow icons, assistive technologies have no way to convey their purpose to users.

Added internationalized `aria-label` attributes to both navigation buttons:

Previous Button: Added `aria-label={t("previous")}` for localized "Previous" label
Next Button: Added `aria-label={t("next")}` for localized "Next" label
i18n Integration: Imported `useTranslation` hook to provide translated labels that match the site's current language setting

## Related Issue

- Closes #17154

PR #18698, opened 2026-07-06
fix: add missing i18n keys for swiper a11y slide messages

### LINKED ISSUE #18695: Swiper a11y slide messages use missing i18n keys ("previous"/"next" not in common.json)
## Description

`src/components/ui/swiper.tsx` (~lines 151–154) configures the swiper a11y module with:

```tsx
a11y={{
  prevSlideMessage: t("previous"),
  nextSlideMessage: t("next"),
}}
```

where `t` is the `common` namespace — but neither `"previous"` nor `"next"` exists in `src/intl/en/common.json`. next-intl falls back to rendering the raw key, so screen-reader slide announcements are the literal strings "previous"/"next" in every locale — untranslated in all 24 non-English languages (including RTL ar/ur) — plus missing-message errors in dev.

## Proposed fix

1. Add `"previous"` and `"next"` (or better, `"swiper-previous-slide"` / `"swiper-next-slide"` with descriptive values) to `src/intl/en/common.json` and update the component to match.
2. Translations propagate via the normal intl flow.

## Found while

reviewing #17155 (swiper a11y labels, closed as redundant) — the visible-button `aria-label`s were fixed in a2198fa3b0, but these a11y-module message keys were missed.

nloureiro:
## Description

`src/components/ui/swiper.tsx` configures the swiper a11y module with `t("previous")` / `t("next")` from the `common` namespace — but neither key exists in `common.json`. next-intl falls back to rendering the raw key, so screen-reader slide announcements are the literal strings "previous"/"next" in every locale (untranslated in all 24 non-English languages, including RTL ar/ur), plus missing-message errors in dev.

This renames them to descriptive keys (`swiper-previous-slide` / `swiper-next-slide` — the generic names were too collision-prone for the shared namespace) and adds the English strings alphabetically in `common.json`. Translations propagate via the normal intl pipeline.

Found while reviewing #17155 — the visible-button `aria-label`s were fixed in a2198fa3b0, but these a11y-module message keys were missed.

Fixes #18695


wackerow (commented):
@pettinarip Thoughts on this adding to the common.json namespace? We only use this component on a few pages, personally would lean to a component-dedicated namespace instead of adding to the already very long list of strings in common.json.

wackerow (approved):
Migrated these strings to new `component-swiper` namespace. Translations completed/review/approved in #18799 to be merged in tandem. Docs updated with details on intl coverage for this component.

PR #18700, opened 2026-07-06
fix: return 400 for malformed percent-encoding instead of 500

### LINKED ISSUE #17967: [Recovery Agent] Unhandled 'failed to decode param' error in [...slug] route
## Error Summary

- **Source:** grafana-logs
- **Item ID:** grafana-fn-error-failed-to-decode-param-at-module-o-next-server-app-locale-s
- **Path/URL:** `app/[locale]/[...slug]/page.tsx` (catch-all content route)
- **Status:** 500 (unhandled server error)
- **Hit count:** 28
- **First seen:** 2026-04-10 (timestamp 1776103884868)
- **Last seen:** (within last monitoring window)

## Analysis

Next.js throws `Error: failed to decode param` when a URL path segment contains malformed percent-encoding (e.g. `%GG`, a bare `%`, or `%%`). In the App Router, dynamic params like `[...slug]` are decoded with `decodeURIComponent` at the framework level — before any page component code runs.

The stack trace points to `.next/server/app/[locale]/[...slug]/page.js`, which is the compiled catch-all markdown content page. Because `generateStaticParams` in that file returns `[]` in the Netlify serverless environment (no content directory at runtime), the route accepts all paths dynamically. This means any URL reaching the server — including bot/scanner traffic with malformed encoding — hits this route and triggers the error.

The error propagates before the `error.tsx` boundary can catch it (that boundary handles client-side and post-render errors, not framework-level param decoding failures), resulting in a 500 response and an ERROR-level log entry.

## Affected Files

- `app/[locale]/[...slug]/page.tsx` — catch-all page with dynamic routing fallback
- No `middleware.ts` exists — nothing currently intercepts malformed requests upstream

## Suggested Approach

**Add a `middleware.ts`** at the repo root that validates URL path encoding before Next.js processes params. If decoding fails, return a `400 Bad Request` response to stop the error from bubbling:

```ts
import { NextRequest, NextResponse } from "next/server"

export function middleware(request: NextRequest) {
  try {
    decodeURIComponent(request.nextUrl.pathname)
  } catch {
    return new NextResponse("Bad Request: Invalid URL encoding", { status: 400 })
  }
  return NextResponse.next()
}

export const config = {
  matcher: ["/((?!_next/static|_next/image|favicon.ico).*)"],
}
```

**Risks / things to verify before shipping:**
1. **New infrastructure** — no middleware exists today. Verify that adding one doesn't break locale routing (the project uses `createNextIntlPlugin`, not the middleware-based next-intl approach, so locale detection/redirect shouldn't be affected, but needs a smoke test).
2. **Edge runtime behavior** — `request.nextUrl.pathname` uses the WHATWG URL parser which may normalize some malformed sequences silently; test with URLs like `/en/foo%GGbar` to confirm the `catch` fires correctly.
3. **Static asset handling** — ensure the `matcher` pattern correctly excludes `_next/` paths and public assets.
4. **Sentry integration** — confirm the middleware doesn't interfere with Sentry's edge instrumentation (`sentry.edge.config.ts`).

An alternative (lighter) approach: check whether `notFound()` in the page component can be called before Next.js decodes params — but this is not possible since decoding happens at the router level, before user code runs.

## Confidence Assessment

Low — root cause is clear (malformed URL encoding hitting the dynamic catch-all route) but the fix requires introducing middleware as a new architectural component. The interaction with `next-intl`'s plugin approach and the Sentry edge config needs careful verification. Risk of unintended side effects on locale routing or asset serving warrants human review before merging.

---

nloureiro:
## Description

Requests whose path contains malformed percent-encoding (`/en/%GG`, a bare `%`) hit the `[locale]/[...slug]` catch-all, where Next's framework-level `decodeURIComponent` throws `failed to decode param` — an unhandled 500 that the `error.tsx` boundary can't catch. Production logs show 28 occurrences since April, essentially all bot/scanner traffic (#17967).

Fix: a guard at the top of the existing `proxy.ts` — `try { decodeURIComponent(pathname) } catch { return 400 }` — before any locale/redirect logic runs. No new infrastructure: the proxy already runs on these routes, so the added cost is one try/catch per request.

## Verified locally (dev server + curl)

| Path | Before | After |
|---|---|---|
| `/en/%GG` | 500 | **400** ("Bad Request: malformed URL encoding") |
| `/en/foo%` | 500 | **400** |
| `/en/` → `/` redirect chain | unchanged | unchanged |
| `/nl/` (deprecated-locale 301) | unchanged | unchanged |

Also ran `pnpm type-check` — clean.

Note for reviewers: the issue's suggested fix predates the middleware→proxy rename (repo is on Next 16); this lands the same logic in the existing `proxy.ts` rather than adding a new `middleware.ts` (which Next 16 rejects when `proxy.ts` exists).

Fixes #17967

### DIFF
--- a/proxy.ts
+++ b/proxy.ts
@@ -69,16 +69,23 @@ function redirectTo(request: NextRequest, pathname: string, status: number) {
 export default function proxy(request: NextRequest) {
   const { pathname } = request.nextUrl
 
-  // Reject malformed percent-encoding (e.g. `%GG`, a bare `%`) with a 400
-  // before route params are decoded. Otherwise the framework-level
-  // decodeURIComponent in the [locale]/[...slug] catch-all throws
-  // "failed to decode param", surfacing as an unhandled 500 (#17967).
+  // Catch malformed percent-encoding (e.g. `%GG`, a bare `%`) before route
+  // params are decoded. Otherwise the framework-level decodeURIComponent in
+  // the [locale]/[...slug] catch-all throws "failed to decode param",
+  // surfacing as an unhandled 500 (#17967). Redirect to a well-formed slug
+  // that fails validation in the catch-all, which calls notFound() — so the
+  // visitor lands on the styled, localized 404 page. This must be a
+  // redirect, not a rewrite: on a rewrite Next still decodes the original
+  // request URL and throws the same 500, whereas the redirect produces a
+  // fresh request with a clean path. (A bare NextResponse with status 400
+  // would render as a browser-default dead-end page.)
   try {
     decodeURIComponent(pathname)
   } catch {
-    return new NextResponse("Bad Request: malformed URL encoding", {
-      status: 400,
-    })
+    return NextResponse.redirect(
+      new URL(`/${DEFAULT_LOCALE}/__bad-request__`, request.url),
+      302
+    )
   }
 
   const lowerPath = pathname.toLowerCase()
--- a/proxy.ts
+++ b/proxy.ts
@@ -69,6 +69,18 @@ function redirectTo(request: NextRequest, pathname: string, status: number) {
 export default function proxy(request: NextRequest) {
   const { pathname } = request.nextUrl
 
+  // Reject malformed percent-encoding (e.g. `%GG`, a bare `%`) with a 400
+  // before route params are decoded. Otherwise the framework-level
+  // decodeURIComponent in the [locale]/[...slug] catch-all throws
+  // "failed to decode param", surfacing as an unhandled 500 (#17967).
+  try {
+    decodeURIComponent(pathname)
+  } catch {
+    return new NextResponse("Bad Request: malformed URL encoding", {
+      status: 400,
+    })
+  }
+
   const lowerPath = pathname.toLowerCase()
   if (pathname !== lowerPath) {
     return redirectTo(request, lowerPath, 301)


wackerow (commented):
Image

Seems fine, but results in a browser-default dead-end page, unlike a 500 or 404 which provides a site-themed page result

myelinated-wackerow:
One refinement to consider before merge. Since the response is created in the proxy before any React runs, the plain-text body renders as a browser-default dead-end page. Next has no "400 page" concept (no error boundary catches a middleware-crafted response), but we can hand these requests to the styled 404 instead:

```ts
try {
  decodeURIComponent(pathname)
} catch {
  // Hand off to the styled, localized 404 page rather than a bare
  // plain-text response. The rewritten slug fails validation in the
  // [locale]/[...slug] catch-all, which calls notFound().
  return NextResponse.rewrite(
    new URL(`/${DEFAULT_LOCALE}/__bad-request__`, request.url)
  )
}
```

The status becomes 404 instead of 400, which still satisfies the goal of #17967 (getting these out of the 500/error bucket), and 404 is defensible semantics for garbage URLs. The rewritten request does not re-enter the proxy, so there is no loop.

For what it is worth, `NextResponse.rewrite(url, { status: 400 })` looks like it would give the best of both, but the rendered route status wins on page rewrites outside the Vercel edge runtime -- a long-standing inconsistency -- so I would avoid relying on it for Netlify.

_- Claude_

nloureiro:
Good call on the styled page — implemented, with one empirical correction: **the rewrite variant doesn't work**. On a rewrite, Next still decodes the original request URL and throws the same `failed to decode param` 500 — the dev log shows the error attributed to `page: '/en/__bad-request__'` even though the rewrite target is clean. A **302 redirect** produces a fresh request with a clean path, so I've pushed that instead: `/en/%GG` → 302 → slug fails validation in the catch-all → `notFound()` → styled, localized 404, final status 404.

Verified on a local dev server: malformed paths land on the real "Page not found" page (curl follows to final 404), healthy routes and locale redirects unchanged. Trade-off vs the rewrite: the address bar changes to `/__bad-request__/` — arguably a feature for garbage URLs. Status semantics match your note (404 instead of 400, still satisfies #17967's goal of emptying the 500 bucket).

PR #18702, opened 2026-07-06
perf(find-wallet): server-render wallet list + rebuild filter architecture to fix mobile INP

pettinarip:
## Why

`/wallets/find-wallet` is one of our most-visited pages and has poor INP on slow mobile devices (Sentry mobile: p75 400ms, p90 ~1.27s over the last 30d). Reproduced locally at 4× CPU throttle (≈ mid-range Android): **any tap during the first ~1.5s of page life took 1.0–1.5s to respond**, on production and on a local build of `dev` alike.

### Root cause

The entire table was a single client island (`FindWalletProductTable`, `"use client"` at the root):

- 49 rows + ~250 Radix tooltips (5 chain icons/row) + two always-mounted filter panels + preset cards hydrated in one burst that saturated the main thread
- The full wallet dataset shipped **three times**: SSR markup, RSC flight props, and a static `wallet-data.ts` import in the client bundle (via `getLanguageCountWalletsData` in the language select)
- Filter state was a nested `FilterOption[]` structure containing **JSX render-closures**, stored in `useState` — every switch/preset tap re-created it and re-rendered both panels + all rows
- Expanding a row mounted `WalletSubComponent`, which re-ran the 1,050-line `useWalletFilters()` and mounted ~19 more Radix tooltips (Sentry: `summary` taps p75 1000ms, up to 2750ms)
- Detail panels only existed after a client-side expand → **not crawlable**

## What changed

`FindWalletProductTable/` is replaced by `src/components/FindWallets/` with an inverted client/server split. The rendered UI is unchanged.

- **Rows are pure server HTML** inside native `<details>`: expansion is the browser toggle (zero React work), and the per-wallet feature matrix is now in the SSR HTML → **newly crawlable**
- **`WalletListController`** (renders `null`) is the only client code touching rows: toggles `.hidden` from filter state, delegates expand-once tracking and `data-matomo` link analytics
- **Flat filter state** (`{toggles, language, layer2}`) in a small context provider; updates wrapped in `startTransition`. Pure logic in `src/lib/utils/findWalletFilters.ts` — no wallet-data import, the client only receives compact boolean entries
- **All strings resolved server-side** and passed as props → dropped the page-level `I18nProvider` and its serialized messages blob
- **One delegated `TooltipLayer`** replaces ~1,200 Radix tooltip instances (triggers are `<button data-tooltip-ref aria-describedby>`, so descriptions reach screen readers without JS)
- Repeated check/warning/info icons → **SVG `<use>` sprite** (they appeared 931× each); 24px chain logos drop blur placeholders
- **`content-visibility: auto` on rows**: initial layout of all 49 rows was a single ~1s main-thread task at 4× throttle *even with JS disabled*; the browser now skips layout for offscreen rows
- `buttonVariants` extracted to server-safe `ui/buttons/button-variants.ts` (re-exported from `Button.tsx`, no call-site changes) so server components can style anchors as buttons

`ProductTable/` is untouched (still used by `Layer2NetworksTable`).

## Measured (4× CPU throttle ≈ mid-range Android, local prod build vs production)

| | prod | this PR |
|---|---|---|
| Early-click INP p75 (expand / preset / filters) | 984 / 1072 / 952ms | **680 / 664 / 648ms**, before `content-visibility` |
| Steady-state INP | 40–64ms | 24–112ms (expand is now a native toggle) |
| Main-thread blocking, first 4s | — | 2.4s → **1.7s** (with `content-visibility`); longest task 963ms → ~500ms |
| HTML transferred (gzip) | 214 KB | **172 KB (−20%)** |
| Client JS chunks (gzip) | 528 KB | **484 KB (−44 KB)** |
| Wallet names in SSR HTML | 49 | 49 (+ detail panels now also crawlable) |

Raw HTML grows (1.0 → ~3 MB) because detail panels are server-rendered and RSC ships markup twice (HTML + flight) — but transferred bytes *shrink* and none of it hydrates. App-shell control: early-click on `/en/about` is 24ms, so the remaining early cost is page content parse, not React.

## Testing

- `pnpm type-check`, ESLint, Prettier, full production build ✅
- All find-wallet e2e specs pass (desktop + mobile-chrome). One page-object locator scoped now that "Links" headings exist in every row's (closed) details
- Browser-verified: filtering (49→38), personas (→5), reset, URL-param hydration (`?android=true&hardware=true` → 6), mobile sheet flow, delegated tooltips, deep-scroll rendering with `content-visibility`, expanded panels pixel-faithful

## Notes for review

- Matomo events preserved (categories/actions/names), with two deliberate deltas: a few inconsistent legacy `eventAction` strings standardized to `` `${locale} - ${label}` ``, and the `eventValue: JSON.stringify(filters)` blob on social links dropped
- `FindWalletProductTable.stories.tsx` deleted without replacement — the new root is an async server component; page-level coverage exists via the Playwright+Chromatic visual suite. Component stories for the client leaves can follow up
- Tapping a chain icon / info icon now opens its tooltip without toggling the row (previously it could do both)


pettinarip:
Closing as we are going to revamp this page. Improvements found here will land in that revamp.

PR #18703, opened 2026-07-06
Fix - Report publisher schemas

mnelsonBT:
## Description

Issue:
- Each report's publisher on the /reports page was emitted as a node with {"@type": "Organization", "name": "..."}, no @id, url, or sameAs. The site has a two-tier entity system in src/lib/jsonld/ — KNOWN_PERSONS/KNOWN_ORGANIZATIONS registries of rich profiles (stable @id, url, sameAs) that pages emit once in the @graph and point at via @id references, with an auto-built alias map resolving authors. The /reports list page was an outlier by emitting one-off publisher nodes instead of using the registry.
- The 2 EF reports emitted an "Ethereum Foundation" node in the same @graph that already contains the rich EF entity from BASE_GRAPH_NODES, creating two competing EF nodes in the graph's id references setup 
- Compound publishers ("Etherealize, Nethermind and L2BEAT", "PSE and Shutter") were each emitted as a single Organization whose name is multiple organizations. Schema.org publisher accepts an array. 

Approach in this PR:
- src/lib/jsonld/organizations.ts - Registered the 16 report publishers in KNOWN_ORGANIZATIONS
- src/lib/jsonld/utils.ts - Extracted the alias-map lookup into resolveKnownEntities(); resolveAuthorsFromFrontmatter now wraps it, unchanged in behavior. Publishers can't use the author function directly because it appends the Ethereum.org  Community reference for the publisher (correct for authors of content on the eth-org site but the wrong attribution for the third-party publishers of external reports)
- app/[locale]/reports/data.ts - The report's publisher display string resolves through the same alias map as markdown authors: frontmatter (key, display name, or GitHub handle). Only the 5 entries whose display text can't resolve (compound or prose names) carry an optional publishers: string[] override, mirroring the authors/author frontmatter convention.
- app/[locale]/reports/page-jsonld.tsx - Each report item's publisher is now @id reference(s) (array for compound publishers); the full org nodes are spread into the @graph, deduplicated across reports and against BASE_GRAPH_NODES (fixes the duplicate-EF issue). Unresolvable publishers fall back to the previous anonymous node.


myelinated-wackerow (approved):
Reviewed the registry approach and verified the deploy preview's emitted @graph directly: all publisher references resolve within the graph, no duplicate @ids (double-EF fixed), compound publishers emit arrays, and the anonymous fallback works as intended. Pushed two polish commits (typed publisher keys, redundant override removed, parentOrganization pattern for subsidiary orgs, trimmed unused registry entries) -- all validated on the rebuilt preview. LGTM.

mnelsonBT:
>  -- all validated on the rebuilt preview. LGTM.
Thanks team @wackerow - marked as ready for review

wackerow (approved):

PR #18704, opened 2026-07-06
fix: homepage remittances copy says seconds, not minutes

### LINKED ISSUE #17998: suggestion: homepage 12 minutes to 12 seconds
## Description
New homepage shows a carousel item stating "Send money home in 12 minutes." This is the average time to _finality_. 

Given the target audience, this may be better suited to state "Send money home in 12 _seconds_" instead, which is closer to the time for inclusion in a block. This is likely more aligned with what the average user will be interested in, and prevents the misconception that a user would have to wait for 12 minutes while transferring funds. 

cc: @konopkja

nloureiro:
## Description

The homepage savings-carousel remittances card says **"Send money home in 12 minutes"** — that's time-to-*finality*, but for a remittance the recipient effectively has the funds at inclusion (~12 seconds, one slot). The card's own description already says *"They receive the funds almost instantly"*, so the 12-minutes headline undercuts its own message, as #17998 argued (cc @konopkja).

Changes:
- `page-index-carousel-remittances-title`: "…in {seconds} seconds"
- `page-index-carousel-remittances-ethereum-value` (the WIRE TRANSFER vs ETHEREUM comparison): "{seconds} seconds"
- Placeholder renamed `minutes` → `seconds` in both strings and the component, so the variable name matches the meaning; the value is still the locale-formatted `12`.

English-only change; translations follow via the intl pipeline (the key values changed, so the manifests will pick them up).

Fixes #17998


wackerow (changes requested):
Image

Altering the variable from `minutes` to `seconds` will break every non-english version of this since it's trying to inject `seconds` into their versions of the string, which only has a `minutes` placeholder still. 

Before we can merge this we'll need to fund the intl-pipeline on this file. I can address that

wackerow (approved):

PR #18725, opened 2026-07-08
feat(apps): add subcategory prop to CategoryAppsGrid

myelinated-wackerow:
## Description

Adds an optional `subcategory` prop to `CategoryAppsGrid` so content pages can embed a grid narrowed to a single subcategory, e.g. `<CategoryAppsGrid category="privacy" subcategory="identity" />`. When set, the grid is pre-filtered server-side and rendered static: no filter bar, no client component in the RSC payload.

- Matching is slug-based (`slugify`) against the raw English tags before translation, so it behaves identically across all 25 locales and accepts `"identity"`, `"Identity"`, or `"liquid-staking"`. Valid slugs are the same set as the `app-subcategories` intl keys.
- Unknown or empty-result subcategories log a warning and render nothing, mirroring the existing unknown-category behavior.
- The prop is typed `string` rather than the per-category `subCategory` unions in `types.ts` -- those unions are stale relative to the live data (e.g. `PrivacyApp` lists only `Pools | Payments | RPC`), and MDX passes strings anyway.
- `limit` and `hideFilter` behavior unchanged; `FilterableCategoryAppsGrid` untouched.

## Test plan

- [x] Manually verified against mock data with privacy apps across `identity`, `payments`, `pools`, `rpc`, and `stealth-address` filters, plus the unknown-subcategory path
- [x] `pnpm type-check`, ESLint, and Prettier pass locally

### DIFF
--- a/src/components/Content/apps/CategoryAppsGrid.tsx
+++ b/src/components/Content/apps/CategoryAppsGrid.tsx
@@ -39,6 +39,12 @@ interface CategoryAppsGridProps {
   /** The app category slug (e.g. "defi", "gaming"). Derived from AppCategoryEnum values. */
   category: Lowercase<AppCategory>
   /**
+   * Optional subcategory to narrow the grid (e.g. "identity", "liquid-staking").
+   * Matched by slug against the category's subcategory tags. When provided,
+   * the grid is pre-filtered to that subcategory and the filter UI is hidden.
+   */
+  subcategory?: string
+  /**
    * Maximum number of apps to display. Defaults to 9.
    * Pass `Infinity` (or `"Infinity"`) to show all apps without a limit.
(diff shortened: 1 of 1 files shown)


tapheret2 (commented):
### Review: feat(apps): add subcategory prop to CategoryAppsGrid

**Files:** `src/components/Content/apps/CategoryAppsGrid.tsx`  
**Size:** +22 / -1

**Notes**
- Focused diff from a static pass; no obvious red flags, but I only reviewed the patch text.

**Nits**
- Any user-facing change that should be mentioned in the PR description?
- Need a changelog/release note for the next tag?

Thanks @myelinated-wackerow — independent review pass on the diff.

pettinarip (approved):

PR #18733, opened 2026-07-09
fix: guard Intl formatters against invalid locale tags

### LINKED ISSUE #18013: [Recovery Agent] RangeError: Incorrect locale information provided at Intl.DateTimeFormat (SSR)
## Error Summary

- **Source:** grafana-logs (Netlify function logs)
- **Item ID:** `grafana-fn-rangeerror-incorrect-locale-information-provided-at-intl-datetimeformat-a`
- **Error:** `RangeError: Incorrect locale information provided`
- **Digest:** `3823164852` (Next.js ignore-listed error hash — same source location on every hit)
- **First seen:** 2026-04-22 ~11:33 UTC
- **Hit count:** 10 on one request, 33 on another within seconds (~43 total, same digest). A companion queue entry (`...-at-intl-datetimeformat`) is the same digest; treat as duplicate.
- **Request ID:** `01KPTFCH77K54F187CXQH9B6N7`

### Stack trace (minified, SSR chunk)

```
RangeError: Incorrect locale information provided
    at Intl.DateTimeFormat (<anonymous>)
    at <unknown> (.next/server/chunks/ssr/_9450b360._.js:2:1659)
    at ae (.next/server/chunks/ssr/_5ed8cb7a._.js:2:24260) {
  digest: '3823164852'
}
```

No application-code frames survive in the minified stack — both frames land in compiled SSR chunks.

## Analysis

`Intl.DateTimeFormat(locale, ...)` throws this specific `RangeError` when `locale` is not a well-formed BCP 47 tag: empty string, `undefined` coerced to `"undefined"`, an underscore form (`en_US`), a crowdin code with an invalid region subtag (e.g. `es-EM`), a path segment, a translated string, etc.

The project wraps `Intl.DateTimeFormat` in `src/lib/utils/date.ts` (`dateTimeFormat()`), and that wrapper does **not** validate the locale before passing it through. Callers include:

- `src/lib/utils/date.ts` — `formatDate`, `formatDateRange`, `getLocaleYear`, `getLocaleFormattedDate`
- `src/lib/utils/time.ts` — `getLocaleTimestamp` (used in `app/[locale]/layout.tsx:61` on **every** page)
- `src/components/LocaleDateTime.tsx`
- `src/components/BigNumber/index.tsx`
- `src/components/History/NetworkUpgradeSummary.tsx`
- `src/components/RadialChart/index.tsx`
- `src/components/Content/what-are-apps/WhatAreAppsStories.tsx`
- `app/[locale]/roadmap/_components/ReleaseCarousel.tsx`
- `app/[locale]/resources/_components/UpgradeCountdown.tsx`
- `app/[locale]/collectibles/_components/CollectiblesProgress/index.tsx`
- `app/[locale]/community/events/_components/EventCard.tsx` (via `formatDate`/`formatDateRange`)
- `app/[locale]/community/events/_components/ContinentTabs.tsx`

`app/[locale]/layout.tsx:50` does call `notFound()` for locales outside `routing.locales`, so direct route typos with bad slugs should be caught before `getLocaleTimestamp` runs at line 61. That suggests the bad locale is arriving via a channel that bypasses the layout guard — possibilities worth investigating:

1. A **code path that passes a non-locale string where locale is expected** — e.g., a crowdin-style code (`es-EM`, `ur-IN`, `pt-BR` uppercased) leaking through `src/lib/types.ts#LangType` into a formatter. The 2026-04-20 homepage redesign merge (#17972) and the subsequent `intl/pending-homepage` merges (#17992, #17996) introduced large intl-pipeline changes — worth checking whether any of the new components read `crowdinCode` or a raw translation value and pass it as a locale.
2. A **Server Action or API route** that renders a dateless component but invokes `dateTimeFormat` with data from query params or headers.
3. A **middleware-less route** — `src/i18n/routing.ts` sets `localePrefix: "as-needed"` with `localeDetection: false`, and no `middleware.ts` exists at the repo root. Anywhere `[locale]/...` renders outside the validated `LocaleLayout` (e.g., `sitemap.ts`, `robots.txt`, `/api/*`) could pass an unchecked value.
4. A **client-side `useLocale()`** that hydrates with a server-rendered stale value — but the error is server-side so this is unlikely.

Without a sourcemap lookup or a Sentry event with unminified frames, the exact call site is guesswork. That's why this is filed as an issue rather than a PR.

## Suggested Approach

1. **Get a sourcemap-resolved stack.** If Sentry is capturing this (it is not in the current queue but may be under a separate issue), look up digest `3823164852` or search for `RangeError: Incorrect locale information`. Alternatively, pull the minified bundle for build `_9450b360` / `_5ed8cb7a` from the Netlify deploy artifacts and map offsets 1659 / 24260 back to source.
2. **Reproduce.** Check `beforeSend`/`onError` hooks to log the offending `locale` value just before the call in `src/lib/utils/date.ts:31`. Redeploy with extra context and wait for a recurrence.
3. **Defensive fix (medium risk).** Wrap `new Intl.DateTimeFormat(locale, finalOptions)` and `new Intl.NumberFormat(locale, finalOptions)` in a try/catch that falls back to `DEFAULT_LOCALE` when `locale` isn't canonical:

   ```ts
   try {
     Intl.getCanonicalLocales(locale)
   } catch {
     locale = DEFAULT_LOCALE
   }
   return new Intl.DateTimeFormat(locale, finalOptions)
   ```

   This converts a 500 into a silent fallback. It is defensive hardening, not a root-cause fix — should be paired with step 2 so the real source gets found and fixed, not just papered over.
4. **Audit call sites that read non-validated input.** Grep for `crowdinCode`, `locale as Lang`, and any place where `locale` is assembled from a translation string, URL param, or user-provided cookie.

## Affected Files

Primary suspects (entry points that call `dateTimeFormat` on every render):

- `src/lib/utils/date.ts` (wrapper — add validation here)
- `src/lib/utils/time.ts#getLocaleTimestamp` (called from `app/[locale]/layout.tsx:61` on every page)

Callers to audit for where `locale` originates:

- `src/components/LocaleDateTime.tsx`
- `src/components/BigNumber/index.tsx`
- `src/components/History/NetworkUpgradeSummary.tsx`
- `src/components/RadialChart/index.tsx`
- `app/[locale]/roadmap/_components/ReleaseCarousel.tsx`
- `app/[locale]/resources/_components/UpgradeCountdown.tsx`
- `app/[locale]/collectibles/_components/CollectiblesProgress/index.tsx`
- `app/[locale]/community/events/_components/EventCard.tsx`
- `app/[locale]/community/events/_components/ContinentTabs.tsx`

## Confidence Assessment

**Low confidence** — the error is real and persistent (not a one-off), and the wrapper in `src/lib/utils/date.ts` is definitely the throw site. But:

- Stack is minified with no app-code frame, so the actual caller is unknown.
- The layout already validates `locale` against `routing.locales`, which rules out the most obvious bug class.
- A defensive fallback in the wrappers is easy but hides the real bug instead of fixing it. Fixing the root cause requires either a sourcemap lookup or reproducing the malformed locale value.

Action needed: a human should (a) resolve the sourcemap to identify the caller, and (b) decide whether to land a defensive fallback as a stopgap while the root cause is investigated.

---

nloureiro:
## Description

Production logs (#18013) show `RangeError: Incorrect locale information provided` bursts (28+ hits) from the events pages. Root cause, verified locally:

- V8's Intl constructors throw a RangeError for **structurally invalid** BCP 47 tags — `"undefined"`, `"wp_admin"`, `""` — but not for merely unknown ones (`"xx-XX"` is fine).
- In the App Router, `generateMetadata` runs **concurrently with** the layout render, so the layout's `routing.locales` guard can't protect it. The three events pages call `getLocaleYear(params.locale)` → `dateTimeFormat` directly on the raw param, so a crawler hitting `/undefined/community/events/` 500s.
- Repro: `node -e 'new Intl.DateTimeFormat("undefined", {calendar:"gregory", numberingSystem:"latn", year:"numeric"})'`

## Fix (two layers)

1. **`safeLocale()`** in `src/lib/utils/date.ts`: falls back to `DEFAULT_LOCALE` when `Intl.getCanonicalLocales` throws — verified it rejects exactly the same input class as the constructors. Applied inside both `dateTimeFormat()` and `numberFormat()` (the repo's two mandated formatter wrappers), so no other call site can hit this again.
2. **Root cause**: the layout's locale guard (`routing.locales.includes` + `notFound()`) added at the top of the three events `generateMetadata` functions, so invalid-locale requests become proper 404s instead of English-rendered 200s.

`pnpm type-check` passes; guard behavior verified in node (invalid → `en` fallback, valid tags incl. `pt-br` pass through untouched).

Complements #18700 (which handles malformed percent-encoding at the proxy layer; this handles well-formed-but-invalid locale segments at the formatter layer).

Fixes #18013


tapheret2 (commented):
### Review: fix: guard Intl formatters against invalid locale tags

**Files:** `app/[locale]/community/events/conferences/page.tsx`, `app/[locale]/community/events/meetups/page.tsx`, `app/[locale]/community/events/page.tsx`, `src/lib/utils/date.ts`, `src/lib/utils/numbers.ts`  
**Size:** +37 / -2

**Notes**
- Focused diff from a static pass; no obvious red flags, but I only reviewed the patch text.

**Nits**
- Any user-facing change that should be mentioned in the PR description?
- Need a changelog/release note for the next tag?

Thanks @nloureiro — independent review pass on the diff.

PR #18745, opened 2026-07-10
Mark dropdown sub-items with dash instead of padding

myelinated-wackerow:
## Description

Designer-requested tweak to the topic-page dropdowns (`ButtonDropdown`, used via `ContentLayout`): nested sub-items are now marked with a leading en dash (`–`) instead of extra inline-start padding.

- All depths share the base `ps-8`; the depth-based `ps-12`/`ps-16` padding map is removed
- Levels deeper than one indent the dash with an em space per extra level (only one level is used today, e.g. `/roadmap/*`)
- The marker is `aria-hidden`, so accessible names remain just the item text
- Logical margin (`me-2`) keeps RTL rendering correct with no special casing
- Marker and indent characters are single constants for easy future adjustment

## Test plan

- Open any `/roadmap/*` page and check the section dropdown: sub-items show `– ` prefix at uniform padding (mobile viewport for the sticky bottom variant)
- Verify an RTL locale (e.g. `/ar/roadmap/`): dash renders on the right of the item text

### DIFF
--- a/src/components/ButtonDropdown.tsx
+++ b/src/components/ButtonDropdown.tsx
@@ -24,9 +24,13 @@ export interface ListItem {
     eventName: string
   }
   callback?: (idx: number) => void
-  /** Nested sub-items rendered indented beneath this item. */
+  /** Nested sub-items rendered beneath this item, marked with a leading dash per level. */
   items?: Array<ListItem>
 }
+
+/** Visual nesting marker; excluded from accessible names. Levels past the first indent the dash with an em space per extra level. */
+const SUB_ITEM_MARKER = "–" // en dash
+const SUB_ITEM_INDENT = " " // em space
 
 export interface List {
   text: string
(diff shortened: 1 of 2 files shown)


tapheret2 (commented):
### Review notes on “Mark dropdown sub-items with dash instead of padding”

Touched files (2): `src/components/ButtonDropdown.tsx`, `src/lib/utils/topicDropdown.ts`
Diff size (approx): +19 / -8

- Change set is focused. I don’t see an obvious correctness red flag from the diff alone.

**Questions / nits**
- Does this need a changelog / release note entry for the target version?
- Any user-facing behavior change that should be called out in the PR body?

Thanks @myelinated-wackerow for the contribution — left this as an independent review pass.

konopkja (approved):

pettinarip (approved):

PR #18749, opened 2026-07-10
refactor: migrate Staking i18n to next-intl useTranslations (#18742 pattern PR)

myelinated-wackerow:
# Pattern PR for #18742 — Staking directory

First batch of the `@/hooks/useTranslation` → `useTranslations` migration. Staking was chosen as the pattern batch because it contains every migration trap in the repo-wide inventory — including the **only** two root-dotted `t.rich` passthroughs and the only two embedded-HTML messages in all 103 affected files. Later batches reference this PR as the recipe.

## Migration recipe

**1. Hook swap** (every file):

```tsx
// Before
import { useTranslation } from "@/hooks/useTranslation"
const { t } = useTranslation("page-staking")
// After
import { useTranslations } from "next-intl"
const t = useTranslations("page-staking")
```

**2. Colon-form keys** → already namespace-bound, drop the prefix; a *different* namespace gets a second binding:

```tsx
// Before
t("page-staking:comp-withdrawal-credentials-verify", { network: "Mainnet" })
t("common:data-provided-by")
// After
t("comp-withdrawal-credentials-verify", { network: "Mainnet" })
const tCommon = useTranslations("common")
tCommon("data-provided-by")
```

**3. Root-dotted `t.rich`/`t.raw` passthroughs** → drop the namespace segment (the binding supplies it):

```tsx
// Before
t.rich("page-staking.page-staking-launchpad-widget-p1", { ... })
// After
t.rich("page-staking-launchpad-widget-p1", { ... })
```

**4. Plain keys** → unchanged call, but the semantics change from `t.raw()` (legacy wrapper) to cooked ICU `t()`. This is safe **only if the message contains no ICU args and no HTML** — verified per key against `src/intl/en/*.json` before the swap (see audit below). Messages with markup would instead need `t.raw`/`t.rich`; none were needed in this batch outside the existing `t.rich` sites.

## Why this is behavior-preserving

- Every plain-key message referenced by these six files (including the dynamic `t(title)`/`t(content)`/`t(linkText)` key sets in `StakingComparison`, which are statically enumerable) was classified against the English catalogs: all plain text, no ICU args, no markup, and zero ICU-quote-risk strings (`''` or `'` adjacent to `{}`) in `page-staking` or `common`.
- Missing-key behavior is unchanged: `src/i18n/request.ts` already suppresses `onError` and supplies `getMessageFallback` globally, so next-intl degrades the same way the wrapper's try/catch did.
[...]

Issue #18742, opened 2026-07-09
Migrate legacy useTranslation hook call sites to next-intl useTranslations


myelinated-wackerow:
## Context

Team decision (July 2026, codified in #18723): client components use namespace-bound `useTranslations` from `next-intl` directly, with one binding per namespace (`const t = useTranslations("page-x")`, plus e.g. `const tCommon = useTranslations("common")` for a second namespace). The custom `@/hooks/useTranslation` wrapper — which added `namespace:key` cross-namespace syntax and raw-message defaults — is now marked `@deprecated` (see #18723) but still has 103 importing files.

## Why migrate (not just deprecate)

The wrapper's semantics diverge from next-intl in ways that hide bugs: its plain-key form returns `t.raw()` and **silently ignores a `values` argument** (see the sr-only quiz bug, #18743), it swallows missing keys, and its `.rich`/`.raw` passthroughs are bound to the root namespace (callers pass full dotted keys). Two translation APIs with different behavior is a standing footgun.

## Current inventory (as of #18723)

- 103 files import `@/hooks/useTranslation` (58 named import, 22 default import, rest mixed)
- 33 call sites use the `namespace:key` colon form → each becomes a second namespace-bound function
- 1 plain-key call passes `values` that are silently dropped (`QuizRadioGroup.tsx:67` — tracked in #18743)
- 2 files call `t.rich` through the wrapper with root-dotted keys (`src/components/Staking/StakingLaunchpadWidget.tsx`, `WithdrawalsTabComparison.tsx`)

## Migration mechanics (not a mechanical sed)

Per file:
1. `const { t } = useTranslation("ns")` → `const t = useTranslations("ns")`; colon-form calls → additional namespace bindings.
2. **Embedded-HTML strings**: the wrapper returns raw messages by default, so many call sites render Crowdin-era strings containing `<b>`/`<a href>` via htmr. Plain next-intl `t()` breaks on these — use namespace-bound `t.raw("key")` (or `t.rich`). Identify by checking the message content, not the call site.
3. While touching a file, check whether it should be a Server Component using `await getTranslations` instead (see design-system skill, server-vs-client.md).
4. Audit for latent `t("plain-key", { values })` calls — these were silently non-interpolating and may reveal more hidden bugs like the quiz one.

## Completion

- Zero imports of `@/hooks/useTranslation`; delete the hook (it is already marked `@deprecated` pointing here).
[...]

PR #18753, opened 2026-07-10
refactor: migrate site shell (Nav/Search/LanguagePicker/TOC/Breadcrumbs) to useTranslations (#18742)

myelinated-wackerow:
# Site-shell batch for #18742

Third batch, applying the #18749 recipe to the navigation chrome: **Nav** (mobile menu ×5, `useNavigation`, `useThemeToggle`), **Search** ×3, **LanguagePicker** ×2, **TableOfContents** ×2, **Breadcrumbs** — 15 files, every binding `common`.

## Notes for review

- **Nav label paths unified**: `CrawlableNav` (server) already called `buildNavigation` with a namespace-bound `getTranslations("common")`, while `useNavigation` (client) passed the legacy wrapper's raw-returning `t`. Both sides now have identical semantics; `buildNavigation`'s 190 keys are all plain text (pre-verified — no ICU, no markup).
- **Fixes a live missing-key bug**: the search modal and mobile menu lazy-load error states called `t("loading-error")` — a key that exists in **no** catalog, so prod renders the literal string `loading-error` when those chunks fail to load. Both now use the existing `loading-error-refresh` key ("Error, please refresh."), already translated in all 25 locales. (One of the 14 missing-key call sites flagged in the #18749 audit.)
- **Dynamic keys**: Breadcrumbs' `t(path)` translates URL segments and falls back to the raw segment for unknown ones — behavior identical post-migration (`getMessageFallback` returns the key, same as the wrapper's catch). `ThemeToggleFooterButton`'s `t(themeLabelKey)` takes only `dark-mode-aria-label` / `light-mode-aria-label`, both verified plain.
- One colon-form call (`common:languages` in LanguagePickerMenu) was already namespace-local — prefix dropped.

## Test plan

- [x] `pnpm type-check`, `pnpm lint` clean locally
- [ ] Deploy-preview sweep: shell strings on `/` + a docs page (breadcrumbs, TOC) × 25 locales (posted as comment)
- [ ] Playwright: open search modal, mobile menu, and language picker on the preview (en, ar, zh) — covers the lazy-loaded/interaction-gated strings incl. the `loading-error-refresh` path components

— Fable

### DIFF
--- a/src/components/Breadcrumbs/index.tsx
+++ b/src/components/Breadcrumbs/index.tsx
@@ -1,5 +1,5 @@
 import { Fragment } from "react"
-import { useLocale } from "next-intl"
+import { useLocale, useTranslations } from "next-intl"
 
 import type { Lang } from "@/lib/types"
 
@@ -15,8 +15,6 @@ import {
   BreadcrumbProps,
   BreadcrumbSeparator,
 } from "../ui/breadcrumb"
-
-import { useTranslation } from "@/hooks/useTranslation"
 
 export type BreadcrumbsProps = BreadcrumbProps & {
   slug: string
@@ -44,7 +42,7 @@ type Crumb = {
 //   { fullPath: "/eth2/proof-of-stake/", text: "PROOF OF STAKE" },
 // ]
 const Breadcrumbs = ({ slug, startDepth = 0, ...props }: BreadcrumbsProps) => {
-  const { t } = useTranslation("common")
+  const t = useTranslations("common")
   const locale = useLocale()
   const dir = isLangRightToLeft(locale! as Lang) ? "rtl" : "ltr"
   const normalizedSlug = normalizeSlug(slug)
--- a/src/components/LanguagePicker/LanguagePickerMenu.tsx
+++ b/src/components/LanguagePicker/LanguagePickerMenu.tsx
@@ -1,3 +1,5 @@
+import { useTranslations } from "next-intl"
+
 import type { LocaleDisplayInfo } from "@/lib/types"
 
 import {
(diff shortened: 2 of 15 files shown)


myelinated-wackerow:

Against [deploy-preview-18753.ethereum.it](https://deploy-preview-18753.ethereum.it):

**1. Fetch sweep — 75/75 pages clean** (25 locales × `/`, `/developers/docs/intro-to-ethereum/`, `/staking/solo/`). Asserts on visible SSR HTML: locale-resolved top-level nav labels, TOC "on this page", breadcrumb segment translation, no un-interpolated ICU args, no escaped markup, no dotted missing-key fallbacks. (Nav dropdown content and button aria-labels are hydration-gated and never appear in SSR HTML — those are covered by layer 2.)

**2. Playwright interaction checks (en, ar, zh) — 0 failures**: opens the search modal and asserts the DocSearch input placeholder equals the locale's `search-ethereum-org` message; opens the mobile hamburger menu, navigates to the language picker, and asserts the translated filter placeholder renders.

Both check suites were first calibrated against production (identical results there), so failures here would have been true regressions.

### About the two failed deploys on this PR

The first two Netlify builds failed because of a defect in my batch transform, not a flake: it inserted the `next-intl` import **above** the `"use client"` directive in seven files, and prettier then rewrote the displaced directive into a no-op `;("use client")` — silently stripping the client boundary. Fixed in aa27f860ef (directive restored as first statement in all affected files; the two client components in the follow-up tables batch were double-checked for the same defect). The third build, containing the fix, is the green one this verification ran against.

### For human eyeballs

- [ar mobile menu](https://deploy-preview-18753.ethereum.it/ar/) — hamburger → menu drawer, languages entry, theme toggle: RTL layout + translated labels
- [zh search](https://deploy-preview-18753.ethereum.it/zh/) — search button → modal placeholder and DocSearch footer labels
- The `loading-error` → `loading-error-refresh` fix only renders when a lazy chunk fails to load, so it has no happy-path surface to click; the key's presence in all 25 catalogs is verified and the string is exercised by the same code path as before.

— Fable

wackerow (approved):
Clean; lgtm -- preview checks out

PR #18759, opened 2026-07-10
Replace Blobscan with growthepie as blob stats data source

pettinarip:
## Description

Replaces the `fetch-blobscan-stats` data source (Blobscan) with growthepie's `da_fundamentals.json`, which the data layer already depends on for three other fetchers.

**Why:** Blobscan's API failed 124 of ~720 hourly runs (~17%) over the last 30 days — every failure a 503 — including multi-day outages (June 13–15 ~53h, June 28–29 ~34h, July 5–6 ~19h, July 8–10 intermittent). Each failed run had already survived 2 task attempts plus internal retry backoff, so these were real sustained outages, and each one fired a Discord alert.

## Changes

- `fetchBlobStats.ts` (renamed from `fetchBlobscanStats.ts`): sums growthepie's `da_ethereum_blobs` daily rows — total blobs = Σ `da_blob_count`, avg blob fee = Σ `da_fees_eth` ÷ total blobs, converted to wei so the existing wei→USD conversion on `/resources` is unchanged
- Task moved from `HOURLY` to `DAILY` schedule to match growthepie's daily update cadence
- Task id / storage key renamed `fetch-blobscan-stats` → `fetch-blob-stats`; getters renamed to `getBlobStats`; `/resources` page, mock data, and unit test updated
- The Blobscan link card on `/resources` stays — only the data source changes

## Deploy note

The new storage key starts empty. After the trigger.dev deploy, trigger `fetch-blob-stats` (or `daily-data-fetch`) once in the dashboard so the key is populated before the next site deploy — otherwise `/resources` reads null until the midnight UTC run.

## Verification

- Ran the new fetcher end-to-end against the live API through the real `fetchRetry` wrapper: `totalBlobs: 20,651,704`, `avgBlobFee: 2.29e14 wei` — consistent with the live site's current values (20M blobs, $1.1 avg fee at current ETH price)
- `pnpm type-check` ✓, ESLint clean on changed files ✓, all 20 data-layer unit tests pass ✓


src/data-layer/tasks.ts
```diff
@@ -54,7 +54,8 @@ export const KEYS = {
   RSS: "fetch-rss",
   GITHUB_REPO_DATA: "fetch-github-repo-data",
   EVENTS: "fetch-events",
-  BLOBSCAN_STATS: "fetch-blobscan-stats",
+  // Key value predates the growthepie migration; kept for storage continuity
```
pettinarip:
```suggestion
```

pettinarip:
**Note: the displayed "Average Blob Fee" will drop from ~$1.10 to ~$0.41 after this deploys. Expected, not a data bug.**

The sources use different denominators: Blobscan's `avgBlobFee` is per blob **transaction**, growthepie's is per **blob** — and one tx carries ~2.5 blobs on average (20M blobs / 8M txs), which exactly explains the ratio. Total fees agree between the sources within ~5% (≈4.7–4.9k ETH all-time), and blob counts within ~4%.

Since the stat is labeled "Average Blob Fee", the new per-blob value is the correct one — the old number was per-transaction wearing the wrong label.

wackerow (approved):

PR #18761, opened 2026-07-10
fix(simulator): modal not closing on direct load

### LINKED ISSUE #18760: Wallet simulator modal cannot be closed when landing directly on /wallets/?sim=...
### Describe the bug

Opening the wallet simulator via a direct link with the `sim` query param (e.g. https://ethereum.org/wallets/?sim=connect-web3) renders the modal, but it cannot be closed: the X button, backdrop click, and Escape all fail. Opening the simulator by clicking a path button on `/wallets/` works normally, including closing.

### To reproduce

1. Navigate directly to https://ethereum.org/wallets/?sim=connect-web3 (fresh page load)
2. Click the X, the backdrop, or press Escape
3. Modal stays open; URL keeps `?sim=connect-web3`

### Root cause

Close calls `clearUrlParams()` in `src/components/Simulator/index.tsx`, which uses `router.replace(pathname)`. The App Router serves same-pathname navigations from the client cache entry created at initial load and restores that entry's URL. On a direct load, that URL includes `?sim=...`, so the param never leaves `useSearchParams` and the modal's `open` state (derived from it) stays true. When the modal is opened client-side from the page, the cache entry was created from the param-less URL, which is why that flow closes fine.

### Fix

Use the native History API (`window.history.replaceState`) for opening/closing, the documented App Router pattern for shallow query-param updates (synced to `useSearchParams` since Next 14.1). Mechanism verified against the production build. Fix in progress on `fix/simulator-modal-close`.

-- Fable

myelinated-wackerow:
## Description

Loading `/wallets/?sim=<path-id>` directly (fresh page load with the query param already attached) made the wallet simulator modal impossible to close: the X button, backdrop click, and Escape all failed, and the URL kept the `?sim=` param. Opening the simulator from the page worked fine.

The close handler used `router.replace(pathname)` to drop the param, but the App Router serves same-pathname navigations from the client cache entry created at initial load and restores that entry's URL. On a direct load that URL includes `?sim=`, so the param never leaves `useSearchParams`, and the modal's `open` state (derived from it) stays true.

This switches `openPath`/`clearUrlParams` to native `window.history.replaceState`, the documented App Router pattern for shallow query-param updates (synced to `useSearchParams` since Next 14.1). Since the `sim` param is consumed only client-side, this also drops the unnecessary RSC round-trip on open/close.

Verified against the production build with Playwright: native `replaceState` closes the modal on a direct `?sim=` load, re-opening afterwards works, and the open-from-page flow is unchanged.

## Related issue

- Closes #18760

### DIFF
--- a/src/components/Simulator/index.tsx
+++ b/src/components/Simulator/index.tsx
@@ -23,16 +23,12 @@ import { Template } from "./Template"
 import type { PathId, SimulatorData } from "./types"
 import { getValidPathId, isValidPathId } from "./utils"
 
-import { usePathname, useRouter } from "@/i18n/navigation"
-
 type SimulatorProps = {
   children: ReactNode
   data: SimulatorData
 }
 export const Simulator = ({ children, data }: SimulatorProps) => {
   const t = useTranslations("component-wallet-simulator")
-  const router = useRouter()
-  const pathname = usePathname()
   const searchParams = useSearchParams()
 
   // Track step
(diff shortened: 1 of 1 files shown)


pettinarip (approved):
Perfecto

PR #18769, opened 2026-07-10
FeedbackWidget popup UI polish

myelinated-wackerow:
## Description

UI adjustments to the FeedbackWidget popup (the dialog that opens from the floating dot):

- `rounded-base` border radius and halved padding (`p-4`)
- Close button moved to its own row above the title, rendering the standard 24px lucide `X`, with a translatable `aria-label`
- All popup text centered on both screens
- `mt-space-half` rhythm above the thank-you subtitle and timing lines; dropped stray `leading-`/`tracking-` overrides
- Semantic cleanup: subtext divs are now `p` elements, and the title is an `h2` referenced by `aria-labelledby` on the dialog

Note on the `h2`: it never enters a page's heading outline during normal browsing — Radix fully unmounts the portaled popover content when closed, so the heading only exists in the DOM while the popup is open. While open, a heading title referenced by `aria-labelledby` is the WAI-ARIA dialog pattern recommendation, and APG's dialog examples use `h2`.

-- Fable

### DIFF
--- a/src/components/FeedbackWidget/index.tsx
+++ b/src/components/FeedbackWidget/index.tsx
@@ -49,35 +49,41 @@ const FeedbackWidget = () => {
         </PopoverTrigger>
 
         <PopoverContent
-          className="mx-2 w-80 max-w-[calc(100vw_-_1rem)] rounded bg-background p-4 sm:p-8"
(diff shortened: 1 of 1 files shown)


konopkja (approved):

pettinarip (approved):
🚀

PR #18770, opened 2026-07-10
refactor: make community stories translatable

myelinated-wackerow:
## Summary

The 26 community stories shown on `/stories` and `/10years` (and 3 of them again on `/what-are-apps`) were hard-coded English (`storyEnglish` in `src/data/tenYearStories.ts`) and never translated. This PR sets them up for the intl-pipeline:

- **New `community-stories` namespace** (`src/intl/en/community-stories.json`): all story copy moved here as `story-*` keys. Data entries now reference the copy via `storyKey`; `loadMessages` picks the namespace up automatically and non-English locales fall back to English until the pipeline translates it.
- **`storyOriginal` is preserved verbatim** as submitted. Stories submitted in English now carry the English text there explicitly (previously blank).
- **`originalLocale`** (BCP-47) tags each submission's language. When it matches the viewing locale, `getCommunityStories` serves the author's original words instead of a round-trip machine translation (e.g. Korean readers get Inhwan's own Korean), and the flip toggle hides since both sides would be identical.
- **Server-side resolution**: `getCommunityStories(locale)` in `src/lib/utils/stories.ts` (replaces `parseStoryDates`, which lived in `app/[locale]/10years/_components/`) resolves story copy + locale-formatted dates before props reach client components, so no extra namespaces ship to the client.
- **StoryCard**: renders `story` (localized) instead of `storyEnglish`; hides the flip when the displayed copy equals the original; flip labels are now language-neutral (`story-card-translation` / `story-card-show-translation` replace the English-specific keys, since the non-flipped side is no longer necessarily English).
- **WhatAreAppsStories** split into a server wrapper (resolves its three featured stories from the same namespace, same pattern as sibling `BrowseApps`) + a client list holding the expand state. Its duplicated hard-coded texts, hard-coded `en-US` date formatting, and literal "Read more"/"Read less" labels (new `story-card-read-less` key) are all localized now.

### Data/bug fixes along the way

- React list keys now use `storyKey` — `/10years` keyed by `story.name` and there are two "Nico" entries (duplicate-key bug).
- Lida's `twitter` field contained the Persian story text instead of a URL; cleared.
- Alex's date `"5.5"` -> `"5.5."` so it parses like the other spreadsheet dates; trimmed stray tabs/trailing whitespace in several `name`/`twitter` fields.

## Translation follow-up

The translation source strings are `src/intl/en/community-stories.json` (story copy) and `src/intl/en/component-story-card.json` (card labels). The companion intl-pipeline PR #18772 (`intl/pending-stories-intl`) already carries the translations for both namespaces across the translated locales and is approved -- merge it in tandem with this PR.

## Test plan

- [x] Scripted diff against HEAD confirms all 26 story texts and originals moved byte-identically (only intentional whitespace trims)
- [x] All 26 namespace strings render verbatim through next-intl's ICU parser (no brace/apostrophe escaping surprises)
- [x] `pnpm type-check`, ESLint, Prettier pass
- [ ] Deploy preview: `/stories`, `/10years`, `/what-are-apps` render unchanged in English (flip toggle appears only on the five stories with non-English originals)


myelinated-wackerow:
## Review -- `refactor: make community stories translatable`

Reviewed the full diff and traced the i18n plumbing end-to-end. Applied two small hardening patches (`67dd739`) and merged latest `dev` (`802c7bea0e`). Verdict below.

### Patches applied during review (`67dd739`)
- **Locale fallback drift** (`getCommunityStories`): added a per-key `hasTranslation` check (via the locale's own `loadMessages`, before the English merge). English submissions on an untranslated locale now render the verbatim `storyOriginal` instead of a potentially-drifted English intl value. Scoped deliberately -- **non-English** originals keep `t()` as a readable English fallback, so an untranslated locale never shows a reader an unreadable original with the flip hidden.
- **Silent card drop** (`WhatAreAppsStories`): a `FEATURED_STORIES` entry whose `storyKey` no longer exists in `tenYearStories` now logs a build-time `console.warn` instead of silently rendering fewer cards (matches the `CategoryAppsGrid` pattern).

### Merged `dev` (`802c7bea0e`)
Brought in #18698 (`fix/swiper-a11y-i18n-keys`) + translations. One conflict in `translations.ts`, resolved by unioning namespaces: `/10years/` now carries both `component-swiper` (18698) and `component-story-card` (this branch); `/apps/` keeps `component-swiper`; `/stories/` + `/what-are-apps/` preserved. The `UserStories` / `CommunityStories` client components auto-merged cleanly, keeping both dev's `useTranslations("common")` migration and this branch's `key={story.storyKey}` fix.

### Verified
- **storyKey <-> `community-stories.json` parity: 26/26**, no empty values, no duplicate keys.
- **Fallback chain**: `i18n/request.ts` merges default-locale messages as the base, so the en-only namespaces resolve for every locale until translated.
- **Client delivery**: `component-story-card` ships to `/10years`, `/stories`, and `/what-are-apps` (exact-path map + slug normalization confirmed); `community-stories` is resolved server-side and correctly never shipped to the client.
- **Locale-match "author's own words"** fires for es/fr/ko/zh (real site codes); `fa` isn't a site locale, so Lida's story is English-with-flip-to-Persian, as intended.
- **Bug fixes confirmed**: duplicate-key (two "Nico" -> unique `storyKey`s), Lida's `twitter` field (Persian prose -> cleared), Alex `5.5` -> `5.5.` (now parses), whitespace trims. `region` safely dropped (no consumer); `category` preserved for the `/stories` filter and guarded (`raw ?? ""`).
- `pnpm type-check`, ESLint, Prettier clean post-merge; diff vs `origin/dev` is exactly the 16 stories files.

### Verdict: approve
Solid, well-scoped refactor; server-side resolution avoids the client `new Date()` hydration trap the old what-are-apps had, and the bug fixes are genuine. Two merge notes:
- **Merge in tandem with #18772** (`intl/pending-stories-intl`), which carries both new namespaces' translations across the translated locales.
- Remaining check: deploy-preview render of `/stories`, `/10years`, `/what-are-apps`, especially the flip toggle on the 5 non-English stories.

-- Opus 4.8

pettinarip (approved):
Reviewed — solid refactor, no blocking issues. Traced the localize/flip logic across cases (English→English hides flip, non-English origin flips to verbatim, fallback vs real translation via `loadMessages`) and it's correct. Verified `community-stories` stays server-only (`pick(allMessages, requiredNamespaces)`), featured what-are-apps cards render equivalently, and all three routes remain **statically rendered** (`setRequestLocale` + explicit-locale `getTranslations`; `getLocale()` in the MDX-embedded component is static-safe because the catch-all route calls `setRequestLocale` first).

LGTM.

PR #18734, opened 2026-07-09
fix: RTL support for Embla carousels (roadmap ReleaseCarousel)

nloureiro:
## Description

Fixes #18580: the `ui/carousel` Embla wrapper always ran LTR physics, and the roadmap release carousel additionally hard-forced `dir="ltr"`, so RTL locales (ar/ur) got a backwards timeline with inverted keyboard arrows.

| File | Change |
|---|---|
| `src/components/ui/carousel.tsx` | Pass `direction` from `useRtlFlip()` into the Embla options (explicit `opts.direction` still wins); set `dir` on the region root so `rtl:` variants work from first paint; swap ArrowLeft/ArrowRight semantics in RTL (in Embla RTL the visually-left slide is *next*); `-ml-4` → `-ms-4`; mirror both chevrons (`rtl:-scale-x-100`); fix vertical-orientation button centering (`rtl:translate-x-1/2`). |
| `ReleaseCarousel.tsx` | Remove both hardcoded `direction: "ltr"` opts + the stale TODO; boundary gradient gets `rtl:bg-linear-to-l`. |
| `roadmap/page.tsx` | Remove the forced `dir="ltr"` on the releases `<Section>` — this was added after the earlier fix attempt branched and would have kept the bug alive. |

**Design note:** direction comes from the locale (`useRtlFlip`/`useLocale`), *not* `document.dir` — in this app `dir` is set client-side in an effect (`useLocaleDirection`), so it's empty during hydration and unusable for this.

**Verified** on a local dev server (`USE_MOCK_DATA=true`): `/ar/roadmap` renders both carousel roots with `dir="rtl"` and logical margins; `/en/roadmap` keeps `dir="ltr"`; no hydration errors. The scroll interaction itself can't be exercised locally — the nav buttons are disabled with mock data **on unmodified dev as well** (verified via stash-control), so please do the interactive check on the deploy preview: on `/ar/roadmap` the timeline should start from the right, ArrowLeft should advance, chevrons mirrored, and the shipped→upcoming gradient should fade right-to-left. Swiper-based carousels (homepage) need no change — Swiper auto-detects RTL and was verified out of scope in the investigation.

Supersedes #18637, which had the right idea (credit @alerodriargui) but was based on pre-#18468 code, relied on `document.dir` (unset at hydration), and missed the `dir="ltr"` that had moved to `roadmap/page.tsx`.

Fixes #18580

PR #18771, opened 2026-07-10
refactor: make Translation component self-contained, deprecate for new code (#18742)

myelinated-wackerow:
# Translation component port + deprecation (#18742)

Makes `Translation.tsx` self-contained: the legacy wrapper's dispatch moves inline — root `useTranslations()` binding, legacy `namespace:key` ids become dot paths, `values` triggers cooked ICU interpolation, and the no-values path fetches the **raw** message so Crowdin-era embedded HTML (real tags with attributes, e.g. `<a href>`) survives for `html-react-parser`. Non-string raw lookups fall back to the id, matching the wrapper's catch.

- **No rendering change**: zero of the 69 consumer files pass `values` (verified), so every live call takes the raw path exactly as before. If `values` is ever passed it now actually interpolates instead of being silently dropped.
- **Marked `@deprecated` for new code** (per review discussion): new strings should be plain ICU or `t.rich` with code-supplied tags; this component remains the designated renderer for the legacy HTML-bearing catalog corpus, and can be deleted outright if those messages are ever normalized.
- This clears the last unblocked importer of `@/hooks/useTranslation`. Still importing it: `FindWalletProductTable/**` (waits #18702), `ui/swiper.tsx` (waits #18698), `ReleaseCarousel.tsx` (waits #18734). The hook deletion + lint-rule flip to `error` lands once those are migrated.

## Test plan

- [x] `pnpm type-check`, `pnpm lint` clean locally
- [ ] Deploy-preview sweep across Translation-heavy pages (staking + roadmap/merge + docs suites, 25 locales) — posted as comment

— Fable

### DIFF
--- a/src/components/Translation.tsx
+++ b/src/components/Translation.tsx
@@ -4,12 +4,10 @@ import parse, {
   Element,
   type HTMLReactParserOptions,
 } from "html-react-parser"
-import type { TranslationValues } from "next-intl"
+import { type TranslationValues, useTranslations } from "next-intl"
 import type { ComponentType } from "react"
 
 import TooltipLink from "./TooltipLink"
-
-import useTranslation from "@/hooks/useTranslation"
 
 type TransformMap = Record<string, ComponentType<Record<string, unknown>>>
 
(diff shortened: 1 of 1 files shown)


myelinated-wackerow:

Against [deploy-preview-18771.ethereum.it](https://deploy-preview-18771.ethereum.it): **225/225 pages clean** across the two heaviest `<Translation>` consumer suites — the staking set (6 pages: withdrawals list items, launchpad, comparisons) and the docs/history/roadmap-merge set (side-nav, upgrade summaries, MergeInfographic ids) — × 25 locales. Same assertions as prior batches: locale-resolved sentinels, no un-interpolated ICU args, no escaped markup, no missing-key fallbacks. Both suites were production-calibrated in earlier batches.

Also pushed a JSDoc clarification from review discussion: glossary tooltips do **not** require this component (`GlossaryTooltip` directly, or `TooltipLink` as a `t.rich` tag handler with the href in code) — only messages whose href lives inside the translated string do.

— Fable

pettinarip (approved):
LGTM

PR #18774, opened 2026-07-11
fix(ui/intl): overflow handling for long intl strings

myelinated-wackerow:
## Description
- Handles eth price card layout for longer intl strings and number formatting
- Wraps "Last 24 hour" tagline under percent change on overflow; center aligns text

image

## Related Issue
image


pettinarip (approved):
💯

PR #17742, opened 2026-03-09
revert: remove homepage A/B test setup

pettinarip:
## Summary
- The HomepageRedesign2026 A/B test has concluded. This removes the test infrastructure from #17261 and #17294 while preserving the Homepage2026 component code for future use.
- Reverts `app/[locale]/page.tsx` to its pre-AB-test state, with the `ENTERPRISE_ETHEREUM_URL` constant fix from `d117fc78` applied on top.

### What was reverted
- `ABTestWrapper` wrapping the entire homepage with 3 variants
- `export const dynamic = "force-dynamic"` (was needed for A/B cookie-less assignment)
- `.catch(() => null)` graceful degradation on all data fetches (restored `throw` on failure)
- `ScrollDepthTracker` and `TrackedSection` wrappers around homepage sections
- `getAccountHolders` fetch (only used by redesign variants)
- Event name changes (`"find_wallet"` → `"find wallet"`, `"cta_click"` → `"Top 4 CTAs"`, etc.)

### What was kept
- `Homepage2026` component and all sub-components (`ScrollDepthTracker`, `TrackedSection`, etc.)
- A/B testing infrastructure improvements (`ab-testing/server.ts` bot detection, `constants.ts` SITE_URL)
- `app/api/ab-config/route.ts` (shared A/B test API)

## Test plan
- [x] Verify homepage renders correctly (original variant)
- [x] Verify no type errors (`npx tsc --noEmit`)
- [x] Verify lint passes (`pnpm lint`)
- [x] Verify Homepage2026 component still exists in codebase

PR #17581, opened 2026-02-16
Remove trailingSlash config

pettinarip:
## Summary
- Remove `trailingSlash: true` from next.config.js
- Add redirect rule to 301 trailing slash URLs to non-trailing slash for SEO preservation
- Update middleware A/B testing URL rewriting to not append trailing slashes
- Update URL utility functions (`addSlashes`, `getFullUrl`) to normalize without trailing slashes
- Update all redirect destinations in redirects.config.js to remove trailing slashes

## Test plan
- [ ] Verify pages load without trailing slashes
- [ ] Verify trailing slash URLs redirect properly (301)
- [ ] Verify A/B testing still works on homepage
- [ ] Verify sitemap URLs don't have trailing slashes

PR #17294, opened 2026-02-12
feat: A/B/C test for homepage 2026 redesign

pettinarip:
## Summary

- Adds A/B/C testing infrastructure for the homepage 2026 redesign
- Variant A: Original homepage (control)
- Variant B: Redesigned homepage with persona modal flow
- Variant C: Redesigned homepage with direct action buttons (no modal)
- Uses dynamic site URL detection for proper tracking in branch/preview deploys
- Polishes hero CTA buttons to match Figma design specifications

## Test plan

- [x] Verify A/B/C test panel appears in development/preview environments
- [x] Test variant switching works correctly via debug panel
- [x] Confirm Matomo tracking fires for each variant
- [x] Check dynamic URL detection works on Netlify preview deploys
- [x] Verify hero CTA button styling matches Figma design

PR #17265, opened 2026-02-10
feat(ab-testing): add Vercel Flags SDK integration with Matomo

pettinarip:
## Summary

- Adds GDPR-compliant, cookie-less A/B testing using Vercel Flags SDK
- Integrates with Matomo for experiment configuration and analytics tracking
- Uses middleware to precompute flag variants and rewrite URLs (e.g., `/en/` → `/en/abc123`)
- Implements deterministic variant assignment via IP + User-Agent fingerprinting
- Adds comprehensive skill documentation for adding new experiments

## Key Changes

- **New flags system**: `src/lib/ab-testing/flags.ts` with Matomo adapter
- **Middleware integration**: Precomputes variants, handles URL rewrites
- **Precomputed route**: `app/[locale]/[code]/page.tsx` for A/B tested homepage
- **Components**: `ABTest`, `TestDebugPanel`, `TestTracker` for variant rendering and tracking
- **Documentation**: Skill file at `.claude/skills/ab-testing/SKILL.md`

## Test plan

- [ ] Set `USE_MOCK_EXPERIMENTS=true` and verify mock experiments work locally
- [ ] Verify homepage renders correctly with precomputed routes
- [ ] Check debug panel shows correct variant assignments (when enabled)
- [ ] Confirm Matomo events fire correctly for variant tracking
- [ ] Test that different IP/UA combinations get different variants

PR #18777, opened 2026-07-13
spike(ab-testing): revive Flags SDK precompute approach (#17265 v2)

pettinarip:
## Summary

Revival of #17265 on current `dev`, to verify on a deploy preview that the Netlify blocker is gone. A fresh minimal repro confirmed that middleware `NextResponse.rewrite()` + next-intl no longer gets converted to a 308 on the current Netlify Next.js runtime (v5.15.12) — in both `trailingSlash` modes — so this approach should now work without removing `trailingSlash` (#17581).

**Why:** the current `ABTestWrapper` infra reads `headers()` at render time, forcing any A/B-tested page into dynamic rendering (the homepage needed `force-dynamic` in #17294, reverted in #17742). The Flags SDK precompute pattern instead assigns the variant in the proxy and rewrites to a statically generated per-permutation route — intended for the upcoming /find-wallets redesign test without giving up static rendering.

## How it works

1. `proxy.ts` precomputes all flags for routes in `AB_TEST_ROUTES` (currently `/`) and rewrites to `/en/ab-code/<signed-code>/` — internal-only, users keep seeing `/`
2. `app/[locale]/ab-code/[code]/page.tsx` statically generates one page per variant permutation (`generatePermutations`) and renders the homepage with the precomputed variant index
3. `ABTest` renders the variant + Matomo tracking; the debug panel (dev/preview) forces variants via `flag_override_*` cookies, which the proxy honors when re-precomputing

## Changes vs #17265

- Coded route under a literal `ab-code` segment — `app/[locale]/[code]` would have shadowed every single-segment content page served by the `[...slug]` catch-all
- Adapted to `localePrefix: "as-needed"` (English URLs are unprefixed) and Next 16 `proxy.ts`
- Fingerprint takes only the client IP from `x-forwarded-for` (#17612 fix)
- `generateStaticParams` tolerates missing `FLAGS_SECRET` (CI); invalid codes 404
- Legacy `ABTestWrapper` infra untouched — consolidation after the preview verifies

## Verification

Locally verified with `USE_MOCK_EXPERIMENTS=true`:
- `/` serves the coded route silently (200, URL unchanged)
- Override cookie `0`/`1` switches original ↔ variant marker
- `/about/`, `/en/` → `/` 301, garbage codes 404, content routes unshadowed

**On this preview** (no Matomo experiment named `HomepageHero` is running, so everyone gets Original by default):
- [ ] `curl -sI <preview>/` → 200, no 308, homepage content
- [ ] Set cookie `flag_override_HomepageHero=1` → yellow *A/B VARIANT A (Flags SDK spike)* banner above the hero
- [ ] Check cache headers on `/` (does the rewritten response still hit Netlify durable cache?)
- [ ] Content pages / other locales unaffected

## Open questions before productionizing

- Cacheability of rewritten responses on Netlify (the actual TTFB win)
- Bot handling: bots currently get fingerprint-assigned like #17605 reverted state
- Remove legacy `ABTestWrapper` + `/api/ab-config` once this lands

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -31,6 +31,14 @@ NEXT_PUBLIC_MATOMO_SITE_ID=
 # API token for A/B testing experiments
 # MATOMO_API_TOKEN=your-matomo-api-token
 
+# Flags SDK secret (required for precomputed A/B testing)
+# Generate with: openssl rand -base64 32
+# FLAGS_SECRET=
+
+# Use mock A/B experiments instead of fetching from Matomo
+# Useful for local development without Matomo credentials
+# USE_MOCK_EXPERIMENTS=true
+
 # Used to avoid loading Matomo in our preview deploys
 NEXT_PUBLIC_IS_PREVIEW_DEPLOY=false
 
--- a/app/[locale]/ab-code/[code]/page.tsx
+++ b/app/[locale]/ab-code/[code]/page.tsx
@@ -1,0 +1,67 @@
+import { generatePermutations, getPrecomputed } from "flags/next"
+import { notFound } from "next/navigation"
+import { setRequestLocale } from "next-intl/server"
+
+import type { Lang } from "@/lib/types"
+
+import { DEFAULT_LOCALE } from "@/lib/constants"
+
+import OriginalHomePage from "../../page"
+
+import { abTestFlags, homepageHeroFlag } from "@/lib/ab-testing/flags"
+
+export { generateMetadata } from "../../page"
+
+/**
+ * Generate a static page for each possible combination of flag values.
+ * Falls back to on-demand rendering (dynamicParams) when permutations
+ * can't be generated at build time (e.g. FLAGS_SECRET missing in CI).
+ */
+export async function generateStaticParams() {
+  try {
+    const codes = await generatePermutations(abTestFlags)
+    // Only the default locale is A/B tested
+    return codes.map((code) => ({ locale: DEFAULT_LOCALE, code }))
+  } catch (error) {
+    console.warn(
+      "[A/B Testing] generatePermutations failed (missing FLAGS_SECRET?):",
+      error instanceof Error ? error.message : error
+    )
+    return []
+  }
+}
+
+interface PageProps {
+  params: Promise<{ locale: string; code: string }>
+}
+
+/**
+ * Precomputed homepage with A/B test variants.
+ * The proxy rewrites eligible requests to include the signed flags code,
+ * which is used here to retrieve the precomputed flag values.
+ * This route is internal-only: it is never linked and only reached via rewrite.
+ */
+export default async function PrecomputedHomePage({ params }: PageProps) {
+  const { locale, code } = await params
+
+  // Only the default locale is A/B tested
+  if (locale !== DEFAULT_LOCALE) notFound()
+
+  // Enable static rendering
+  setRequestLocale(locale)
+
+  let heroVariant: number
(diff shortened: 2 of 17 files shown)


pettinarip:
## ✅ Preview verification results (commit 792da668)

The Flags SDK approach **works on Netlify** — including the case that killed #17265.

| Check | Result |
|---|---|
| `/` (no cookie) | 200, URL unchanged, silently serves precomputed route |
| `/` with `flag_override_HomepageHero=1` (origin-reaching request — the case that 308'd in #17265) | **200, variant banner rendered, no redirect leak** |
| `/` with override `0` | 200, original |
| Cache | `Netlify Edge: hit` on repeat requests — variant pages are **prerendered + edge-cached**, no per-request SSR (this is the force-dynamic fix) |
| `/about/`, `/es/`, `/wallets/find-wallet/` | 200, unaffected |
| `/en/` | 301 → `/` (unchanged) |
| Garbage codes | 404 |

### Root cause of the original #17265 failure, finally pinned down

Not a Netlify runtime bug per se, and not `trailingSlash` sitewide: the **signed flags code contains dots**, so Next.js treats the rewrite target as a *file path*. With `trailingSlash: true`, rewrite normalization re-appends a trailing slash, then the origin 308-redirects to strip it — exposing the internal URL. Cache hits masked it, so it looked intermittent/platform-dependent.

**Fix:** encode the code's dots as `~` in the URL (`encodeABCode`/`decodeABCode`), making the segment directory-like. No sitewide `trailingSlash` change needed (#17581 can stay closed).

### Before productionizing
- [ ] Decide bot handling (bots currently get fingerprint-assigned)
- [ ] Real variant component instead of the spike banner
- [ ] Point `AB_TEST_ROUTES` at the find-wallets redesign page when ready
- [ ] Remove legacy `ABTestWrapper` + `/api/ab-config` infra

pettinarip:
## ✅ Second test verified: `FindWalletHero` on `/wallets/find-wallet/` (commit 3eda06fd)

Validates the **nested-path case** (code is not the last URL segment: `/en/ab-code/<code>/wallets/find-wallet/`).

| Check | Result |
|---|---|
| `/wallets/find-wallet/` (no cookie) | 200, silent rewrite, `Next.js: hit` cached |
| Override cookie `1` (origin-reaching) | **200, variant banner, no 308** |
| Override cookie `0` | 200, original |
| Homepage test | still working |
| Unslashed URL | 200 (same as production today — pre-existing) |

Also restructured flags into **per-route groups** (`abTestRoutes` map): the proxy precomputes only the route's own flags, so permutations don't multiply across pages (2 pages × 2 variants = 4 prerendered pages, not a combined 2×2 per page). Adding a test to a new page = one flag + one map entry + one thin coded page + the `ABTest` wrapper.

Cross-flag isolation verified locally: `flag_override_FindWalletHero=0` wins independently of `flag_override_HomepageHero=1`.

pettinarip:
Superseded by the split pair:
- **#18807** — Flags SDK infrastructure (ready for review; merges inert, legacy infra removed, docs rewritten)
- **#18808** — demo experiments stacked on #18807 (deploy-preview playground, not for merge as-is)

All verification done on this spike's preview (silent rewrites, edge-cached variant pages, the dot-encoding fix for the #17265 308 bug) carries over — the implementation in #18807/#18808 is the same code, restructured with a `defineABFlag` helper and an empty-by-default `abTestRoutes`.

PR #18729, opened 2026-07-09
Docs: mention Vyper's module system on the languages page

FDongFDong:
## Description

The "Vyper does not support:" list on the smart contract languages page includes "Inheritance" as a bare bullet, which can read as if Vyper has no mechanism for code reuse at all. This adds one sentence after the list noting that, since v0.4.0, Vyper supports a module system and achieves code reuse through composition rather than class inheritance.

## Related Issue

Closes #18613

PR #18614, opened 2026-06-29
docs: note Vyper's module system on the languages page

lufa23:
## Description

Adds one sentence to the Vyper section of the smart contract languages page, noting that Vyper supports code reuse through its module system (added in v0.4.0). The existing "Vyper does not support: … Inheritance" bullet is left unchanged; the addition corrects the impression that Vyper has no code-reuse mechanism.

## Related Issue
Closes #18613

PR #18780, opened 2026-07-13
content: mention Vyper module system on the languages page

### LINKED ISSUE #18613: Mention Vyper's module system on the languages page
### Affected page URL or path

/developers/docs/smart-contracts/languages/

### Type of suggestion

Update existing content

### Current content or behaviour

Under ## Vyper, the "Vyper does not support:" list includes a bare bullet "Inheritance". Listed alone among missing features, it reads as if Vyper has no mechanism for code reuse at all.

### Proposed change

Keep the list unchanged (Vyper genuinely has no class inheritance), and add one sentence after the list:

"Since v0.4.0, Vyper supports a [module system](https://docs.vyperlang.org/en/stable/using-modules.html). Code reuse is achieved through composition, rather than class inheritance."

### Motivation and context

The "Inheritance" entry is true but incomplete. Vyper 0.4.0 added a module system (`import` / `initializes` / `uses` / `exports`) that is the way Vyper allows for code reuse. Currently, readers comparing languages on this page may incorrectly conclude that Vyper doesn't allow for code reuse (which is no longer accurate). The addition corrects that impression.

### References or related resources

[Vyper docs: Modules](https://docs.vyperlang.org/en/stable/using-modules.html)
[Vyper 0.4.0 release notes](https://docs.vyperlang.org/en/stable/release-notes.html#v0-4-0-nagini)

Semak12345:
## Description
Added a short note to the Vyper section clarifying that while Vyper does not support class inheritance, it does support code reuse through its module system since v0.4.0.

## Related Issue
Fixes #18613

### DIFF
--- a/public/content/developers/docs/smart-contracts/languages/index.md
+++ b/public/content/developers/docs/smart-contracts/languages/index.md
@@ -101,6 +101,8 @@ This example should give you a sense of what Solidity contract syntax is like. F
   - Infinite-length loops
   - Binary fixed points
 
+Since v0.4.0, Vyper supports a [module system](https://docs.vyperlang.org/en/stable/using-modules.html). Code reuse is achieved through composition, rather than class inheritance.
+
 For more information, [read the Vyper rationale](https://vyper.readthedocs.io/en/latest/index.html).
 
 ### Important links {#important-links-1}


wackerow:
Fixed with #18729 -- odd that the wording is byte-identical 🤔

wackerow:
Yeah, and byte-identical to the original contributor that I missed who actually posted this in #18614

Plagiarizing like this is not tolerated.

wackerow:
Cerberus, block Semak12345

PR #18783, opened 2026-07-13
[recovery] Return 404 for non-existent story slugs

pettinarip:
## Error context

Production error captured from Netlify logs (Grafana → Keep):

```
[ERROR] Error: ENOENT: no such file or directory, open '/var/task/public/content/stories/-6/index.md'
    at async Object.readFileFallbackBlobStore [as readFile] (.netlify/dist/run/next.cjs:554:16)
    ...
    at async Module.t [as generateMetadata] (...)
```

- **Path:** `/var/task/public/content/stories/-6/index.md` (request to `/stories/-6/`)
- **Hit count:** 2 (first seen 2026-06-10)

## Root cause

The stories `[slug]` route (`app/[locale]/stories/[slug]/page.tsx`) reads the markdown file directly — in both the page body (`getPageData`) and `generateMetadata` (`getMdMetadata` → `importMd` → `fs.readFile`) — without guarding against non-existent slugs. `/stories/` has `generateStaticParams`, but with default `dynamicParams`, unknown slugs like `-6` (a crawler/bot probe) are rendered on demand. `generateMetadata` then throws an unhandled `ENOENT`, surfacing as a 500 in the Netlify server handler instead of a 404.

The shared catch-all `[...slug]` route already handles this exact case: it validates the path and calls `notFound()`, and wraps `getMdMetadata` in a try/catch returning basic not-found metadata. The stories route was written without that guard.

## Fix

Mirror the catch-all route's handling:
- Page body: validate `slug` against `getStorySlugs()` and call `notFound()` for unknown slugs.
- `generateMetadata`: wrap `getMdMetadata` in a try/catch that returns basic not-found metadata.

Single-file change. `pnpm type-check` passes and ESLint is clean.

## Affected files

- `app/[locale]/stories/[slug]/page.tsx`


wackerow (approved):
👍 Lgtm

PR #18788, opened 2026-07-14
test: verify Chromatic required check on dev (do not merge)

pettinarip:
Test PR to verify the new required check setup after #18690:

- Touches `src/components/**` (comment-only change) → `chromatic-gate` should output `run: true`
- `Visual regression (Chromatic)` job should run and report as a **required** check
- Will be closed after verification — **do not merge**

PR #18789, opened 2026-07-14
test: verify Chromatic required check skip path (do not merge)

pettinarip:
Companion to #18788 — verifies the skip path of the new required check:

- Touches only `docs/**` (non-UI) → `chromatic-gate` should output `run: false`
- `Visual regression (Chromatic)` should report **skipped**, which must still satisfy the required check (not hang as "Expected")
- Will be closed after verification — **do not merge**

### DIFF
--- a/docs/e2e-testing.md
+++ b/docs/e2e-testing.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
-This project uses [Playwright](https://playwright.dev/) for end-to-end testing.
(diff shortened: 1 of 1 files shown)


pettinarip:
✅ Verified: non-UI PR → `chromatic-gate` output `run: false`, `Visual regression (Chromatic)` reported **skipped**, which satisfies the required check (no merge block from the check). Closing this test PR.

PR #18794, opened 2026-07-14
Collapse mobile category filters on dev tools page

pettinarip:
## Description

On mobile, the `/developers/tools` category filter rendered the full top-level category list expanded above the results, pushing content far down the page.

This wraps the mobile filter box in the existing `Collapsible` primitive, closed by default:

- The trigger row shows the active selection and count ("Resource categories · 285" on the main page, or e.g. "Contract tooling · 73" on a category page) with a chevron that flips when open.
- Inside the expanded panel, the "Resource categories" (all) link is hidden on the all-resources page where it duplicated the trigger label and linked to the current page; it remains on category pages as the way back to the unfiltered view (via new `showAllLink` prop on `CategorySidebar`).
- Desktop sidebar is unchanged.

Verified at 390px viewport: collapsed by default, expands to the category tree, category navigation and subcategory filtering work, back-to-all reset works; desktop layout untouched. `pnpm type-check` and ESLint pass.

**Preview:** https://deploy-preview-18794.ethereum.it/developers/tools


wackerow:
Nice.. overall looks good, couple comments though:

- Should the menu expanding take up space, or should it act more like a floating dropdown? Right now it shifts the content below it downward
- We're using chevrons, but clicking on them simultaneously 1- navigates the URL to the category selected, 2- closes the category menu, 3- (hidden) expands the category itself (noticeable if you re-open the menu)

Perhaps this second one is a limitation of the way we're handing the routing for this, but feels like the chevron's action should be to expand that category in-place.

cc: @konopkja any thoughts on the desired behavior here?

konopkja:
> Nice.. overall looks good, couple comments though:
> 
> * Should the menu expanding take up space, or should it act more like a floating dropdown? Right now it shifts the content below it downward
> * We're using chevrons, but clicking on them simultaneously 1- navigates the URL to the category selected, 2- closes the category menu, 3- (hidden) expands the category itself (noticeable if you re-open the menu)
> 
> Perhaps this second one is a limitation of the way we're handing the routing for this, but feels like the chevron's action should be to expand that category in-place.
> 
> cc: @konopkja any thoughts on the desired behavior here?

this was discussed with pablo lgtm

wackerow (approved):
@konopkja roger that

PR #18801, opened 2026-07-14
[recovery] fix: guard IssuesList against GitHub issues with null user

pettinarip:
## Production error

```
TypeError: Cannot read properties of undefined (reading 'login')
  at x (.next/server/chunks/[root-of-the-server]__0jap1v0._.js:2:7061)
  at async l (…:2:10929)
  at async s (…:2:11970)
  at async Module._ [as handler] (…:2:13040)
```
- Function: `___netlify-server-handler` · hit_count: 1 · request_id: `01KTAFRXH3YJ525MSC7HT2VPTW`
- URL: none captured

## Root cause

`src/components/IssuesList/index.tsx` is an async server component (rendered via MDX on the public `/contributing` page). It reads `issue.user.login`/`issue.user.avatar_url` unguarded inside its synchronous `.map()` callback:

```tsx
{issues.map((issue) => (
  <Avatar name={issue.user.login} href={`https://github.com/${issue.user.login}`} … />
```

GitHub's REST issues API returns `user: null` for issues authored by deleted ("ghost") accounts. The `GHIssue` type declares `user` as non-null, so there is no compile-time guard. When the cached "good first issues" list contains such an issue, the SSR render dereferences `undefined.login` and the whole page render throws.

The synchronous error frame `x` (no `async`) nested under the async render frames and the top-level `___netlify-server-handler` matches this synchronous `.map()` callback rather than the webhook's async POST body (`app/api/gfi-issues-webhook/route.ts`), which shares the same field access but reads a webhook payload where `issue.user` is reliably present.

## Fix

Filter out issues without a `user` before rendering — a one-line null check that drops only malformed (ghost-authored) entries and leaves the existing card markup untouched.

## Verification

- `pnpm type-check` — passes
- `pnpm exec eslint src/components/IssuesList/index.tsx` — clean

### DIFF
--- a/app/api/gfi-issues-webhook/route.ts
+++ b/app/api/gfi-issues-webhook/route.ts
@@ -44,11 +44,14 @@ export async function POST(req: Request) {
       timestamp: issue.created_at,
       description: issue.labels.map((label) => label.name).join(" • "),
       color: 10181046, // purple
-      author: {
-        name: issue.user.login,
-        url: issue.user.html_url,
-        icon_url: issue.user.avatar_url,
-      },
+      // `issue.user` is null for issues authored by deleted ("ghost") accounts.
+      ...(issue.user && {
+        author: {
+          name: issue.user.login,
+          url: issue.user.html_url,
+          icon_url: issue.user.avatar_url,
+        },
+      }),
     },
   ]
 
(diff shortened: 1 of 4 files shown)


wackerow (approved):
Looks clean

PR #18807, opened 2026-07-15
feat(ab-testing): Flags SDK precompute infrastructure

pettinarip:
## Summary

Replaces the legacy A/B testing infrastructure with the [Flags SDK](https://flags-sdk.dev) precompute pattern. **This PR merges inert** — `abTestRoutes` is empty, so the proxy no-ops until an experiment is registered. A companion PR stacked on this branch demos two working experiments on its deploy preview.

**Why:** the legacy `ABTestWrapper` read `headers()` at render time, forcing any A/B-tested page into dynamic rendering — running a test on the homepage meant losing edge caching on our highest-traffic page (#17294 → reverted in #17742). With this setup, variants are assigned in the proxy and served as **statically prerendered, edge-cached pages**. Verified on the #18777 spike preview: variant responses are `Netlify Edge: hit`.

## How it works

1. `proxy.ts` matches the path against `abTestRoutes`, fingerprints request headers (cookie-less, GDPR-compliant), evaluates the route's flags against Matomo experiment config, and packs the result into an HMAC-signed code
2. It **rewrites** to `/en/ab-code/<code>/<path>` — internal-only; the visitor's URL never changes
3. Each tested route has a thin coded page that prerenders one static page per flag permutation (`generateStaticParams`)
4. `ABTest` renders the variant; `ABTestTracker` pushes `AbTesting::enter` + custom dimension 1 to Matomo

Experiments stay Matomo-driven: status, traffic weights, and scheduling are dashboard changes, no deploys. On any failure (Matomo down, no experiment, tampered code) everything falls through to the original page.

## Scope & safety

- **Default locale only** — route keys are locale-less canonical paths; non-EN visitors get the original page with no tracker (verified: no rewrite header on `/es/`, `/de/`)
- **Internal URLs unreachable** — direct `/ab-code/…` visits get case-mangled by the sitewide lowercase 301, fail signature verification, and 404; coded pages re-export the original page's canonical metadata as a second layer
- **Signed codes are dot-encoded** (`.` → `~`): dotted URL segments + `trailingSlash: true` trigger an origin 308 that leaks the internal URL — this was the root cause that killed #17265, finally pinned down and fixed
- **Bots**: fingerprint-assigned like any visitor (stable UA/IP → consistent variant); Matomo filters known bots from reports. No proxy-level bot exclusion — documented as the accepted behavior

## Changes

- `src/lib/ab-testing/`: `defineABFlag` helper, `identify` fingerprinting (client-IP-only `x-forwarded-for`, #17612), Matomo adapter (FNV-1a weighted assignment, mock experiments for local dev), `encodeABCode`/`decodeABCode`
- `src/components/AB/`: new `ABTest`, debug panel now drives proxy re-evaluation via `flag_override_*` cookies (dev/preview only), tracker hardened
- **Removed legacy infra**: `ABTestWrapper`, `ClientABTestWrapper`, `src/lib/ab-testing/server.ts`, `app/api/ab-config/` (unused since #17742)
- Rewrote `docs/ab-testing.md` (architecture + step-by-step recipe); updated `CLAUDE.md`/`AGENTS.md`
- `.env.example`: `FLAGS_SECRET`, `USE_MOCK_EXPERIMENTS` (both already configured on Netlify)

## Test plan

- [x] Type-check, lint, unit tests pass
- [x] Full flow verified on the #18777 spike preview: silent rewrites (200, URL unchanged), edge cache hits on variant pages, override cookies, invalid codes 404, non-EN locales and content routes untouched
- [ ] Examples PR (stacked on this branch) demos two live experiments on its deploy preview

## History

Supersedes #17265 (closed: blocked by what turned out to be the dot/trailingSlash 308) and the #18777 spike, where this implementation was developed and verified. #17581 (remove trailingSlash) is not needed.


pettinarip:
Addressed in 9e01d52742:

- ✅ **Version regression restored to `11.14.0`** — confirmed root cause: `package.json` was checked out wholesale from the spike branch (cut from pre-11.14.0 dev) when assembling this branch. Also audited `pnpm-lock.yaml` for the same staleness: its diff vs `dev` is only the `flags@4.2.0` package and its `@edge-runtime/cookies` dependency — no drift reverted.
- ✅ **`defineABFlag` JSDoc completed** (`@param description`).
- ⏭️ **`force-cache` kept as-is.** In the proxy's edge runtime (the only place this fetch runs) Next's data cache doesn't operate, so the hint is best-effort and config propagates in practice; keeping it avoids implying cache semantics the runtime doesn't provide.
- ⏭️ **`Secure` cookie kept as-is** — dev-tooling nit, acceptable (Chrome/Firefox accept Secure on http localhost; Safari devs can use the cookie manually if needed).

Fixes are also propagated to the stacked examples branch (#18808).

pettinarip:
Dropped the `X-Robots-Tag: noindex` rules for `/ab-code/*` (3c96149d5f). While verifying the demo preview I noticed Netlify sends a blanket `noindex` on all deploy previews, which means the rule's production behavior can't be verified before merge — and in the worst case (CDN matching header rules against the middleware-rewrite target) it would noindex the *public* pages under test. It was redundant defense anyway: direct `/ab-code/…` visits already 404 via signature verification, and coded pages declare the public URL as canonical. Documented the pitfall in `docs/ab-testing.md` so it doesn't get re-added.

myelinated-wackerow (commented):
@pettinarip looks clean! A few things before the companion PR turns on a real experiment:

**1. Matomo config is fetched live in the proxy, uncached and untimed.** `fetchMatomoExperiments()` runs inside `decide()` → `precompute()` → the proxy (edge middleware), on every request to a registered route (it's what selects the cached variant page, so it can't be cached itself). Two problems:
- `cache: "force-cache"` is a no-op in middleware — the Next Data Cache isn't available there, so there's no cross-request reuse. Every hit to a tested route (incl. the homepage) adds a blocking Matomo round-trip on the edge critical path. The page render is cached; the variant lookup isn't.
- No timeout on the `fetch`. A refusal is handled, but a slow/hanging Matomo would stall `await precompute` up to the function timeout on every request to that route.

Can we add a module-level TTL cache (persists within a warm edge instance), an `AbortSignal.timeout(...)`, and `dedupe` the fetch so multi-flag routes don't multiply calls? Did the #18777 spike give any read on the added edge latency per request?

**2. `isExperimentActive` counts `"created"` as active.** An experiment created but not launched (no `start_date`) will bucket and track users before it's started. Intended, or should it be `"running"` only?

**Minor:**
- `assignedAt: Date.now()` is never read by the tracker, and in the static coded page it bakes in build time anyway — probably drop it from `ABTestAssignment`.
- Debug override value isn't bounded by `numVariants`, but it's dev/preview-only and `ABTest` clamps at render — just noting.

Approving once `#1` is handled here or explicitly slated as the first task in the companion PR, with a word on `#2`.

_- Opus 4.8 / @wackerow_

pettinarip:
Thanks @wackerow — all addressed in 1b4efcbcba.

## 1. Uncached Matomo fetch in the proxy — fixed

You're right on both counts, and I measured the baseline to be sure. From a Netlify **edge isolate** to Matomo Cloud, `AbTesting.getAllExperiments` costs **~154 ms cold / ~45–65 ms warm** — paid on every request to a tested route, on the critical path before the rewrite. Confirmed `force-cache` is inert in the edge runtime (no Data Cache there).

Fix — a **module-level TTL cache** in the adapter (`getExperimentConfig`):
- **1-hour TTL**, cached per warm isolate. Measured after: **139.9 ms once**, then **0.0–0.3 ms** for every subsequent lookup. Crucially, `totalMatomoFetches` stayed at **1** across separate HTTP invocations spanning ~40 s — module state does persist across requests in a warm isolate, so the cache is real cross-request reuse.
- **In-flight dedupe** (`inflight` promise) — concurrent requests and multi-flag routes share one fetch.
- **`AbortSignal.timeout(2 s)` + stale-if-error** — a slow/hanging Matomo can never stall page requests; the isolate serves the last known config and retries after 60 s (decoupled from the 1 h TTL so an outage doesn't pin stale config for an hour).

Net vs the old system: the legacy `/api/ab-config` cached config for 1–2 h too, so propagation is unchanged, but we keep static edge-cached pages instead of per-visit SSR. Propagation trade-off (dashboard change visible within ≤1 h) is documented in `docs/ab-testing.md`.

_Methodology note: measured via a standalone edge function on a throwaway site, tokenless (auth adds only ~5–10 ms, calibrated separately) — the real API path hasn't run on a deployed preview yet since previews use `USE_MOCK_EXPERIMENTS`. Planning a real integration test on Matomo dev site 43 before we turn on a live experiment._

## 2. `isExperimentActive` counting `created` — fixed

Agreed. This was inherited verbatim from the legacy `/api/ab-config`, but your scenario is right: a created-but-not-launched experiment would bucket users. Now `running` is always active, and `created` only counts if it has an explicit `start_date` (so scheduled launches still work, but merely preparing an experiment in the dashboard doesn't go live).

## Minor
- **`assignedAt`** — dropped from `ABTestAssignment`. Good catch that it baked in build time in the prerendered pages.
- **Unbounded override** — left as-is (dev/preview-only, clamped at render), as you flagged.

Same commit is already on the companion PR (#18808) so its preview runs the cached adapter.

pettinarip:
Integration-tested end-to-end on Matomo staging (site 43, real running `HomepageHero`, `USE_MOCK_EXPERIMENTS=false`) — the live path (fetch → auth → parse → weighted assignment) had never run on a preview until now:

- Natural bucketing, 20 distinct fingerprints: **8 Original / 12 VariantA** (mocks off, real config applied)
- Same fingerprint ×5: deterministic
- Cookie-less request: silent rewrite, `Netlify Edge: hit`
- Route with no matching Matomo experiment: 0/6 → all Original (disabled→Original fallback confirmed)

wackerow (approved):

PR #18810, opened 2026-07-15
feat(dev tools): move "Suggest a resource" CTA into catalog sidebar

phipsae:
## Description

On `/developers/tools`, the "Suggest a resource" CTA sat at the very bottom of the page, below the entire catalog, so with ~285 resources it was effectively never seen.

This moves it into the category sidebar as a compact, full-width button:
- **Desktop**: pinned below the category filters and stays visible while the catalog scrolls (the sticky column is now a flex layout, so the list scrolls in the remaining space and the button never gets pushed off-screen).
- **Mobile**: sits directly under the filter box, above the results.

The old bottom block also used a non-existent `bg-radial-a` class (rendered with no background); that block is removed. No new copy or translation keys, the existing `page-developers-tools-suggest-resource-button` string is reused. Both `/developers/tools` and the category pages share `ToolsPageBody`, so both get the change.

## Screenshots

**Desktop (button stays pinned while scrolling)**

Desktop

**Mobile (under the filters)**

Mobile


wackerow:
I like the idea, but think we can get away with a simple link here instead of a button, and would hide this on mobile since we're already limited with space there and also less likely the user-base who'd be suggesting resources. I can make those adjustments

phipsae:
@wackerow thanks for taking a look and making the adjustments, agree the plain link works better and hiding it on mobile makes sense.

phipsae:
Thanks for taking a look and making the adjustments, agree the plain link works better and hiding it on mobile makes sense. Let me know if anything else is needed from my side.

Issue #17718, opened 2026-03-05
fix(seo): /wallets/find-wallet/ missing from sitemap.xml


konopkja:
## Description

The `/wallets/find-wallet/` page (and the entire `/wallets/` section) is completely absent from the site's `sitemap.xml`. This reduces crawl priority and discovery speed for one of the site's most visited pages (~35-80k monthly visits).

### Verification

```bash
curl -s https://ethereum.org/sitemap.xml | grep -i "wallet"
```

The only wallet-related URL in the sitemap is `/contributing/adding-wallets/` (the contributor guide). The actual wallet finder page and `/wallets/` landing page are missing.

### Context

- `robots.txt` is fully permissive (no Disallow rules), so crawling isn't blocked
- The page is discoverable via internal links (main nav, homepage hero CTA), so it's not completely invisible
- However, sitemap inclusion signals crawl priority to search engines and helps with faster re-indexing after content changes
- Three open PRs address broader sitemap generation issues: #17563, #17567, #17639

### Impact

- Search engines don't get an explicit signal to prioritize crawling this page
- After content updates (new wallets added, data changes), re-indexing is slower without sitemap signals
- The page's `lastmod` date isn't communicated to search engines

### Suggested fix

Ensure the sitemap generation logic includes dynamic route pages like `/wallets/find-wallet/` and its locale variants. This may be resolved by one of the existing sitemap PRs (#17567 or #17639), but should be verified explicitly for this page.

wackerow:
This command will search for and count the instances of the `/wallets/` and `/wallets/find-wallet/` paths for both production and dev environments:

```sh
for domain in ethereum.org dev.ethereum.org; do echo "=== $domain ===" && xml=$(curl -s --max-time 30 "https://$domain/sitemap.xml") && echo "/wallets/: $(echo "$xml" | grep '<loc>' | grep -c '/wallets/<')" && echo "/wallets/find-wallet/: $(echo "$xml" | grep '<loc>' | grep -c '/wallets/find-wallet/')"; done
```

(Searches through https://ethereum.org/sitemap.xml and https://dev.ethereum.org/sitemap.xml)

Which returns the expected 25 version of each for all 25 locales:

```sh
=== ethereum.org ===
/wallets/: 25
/wallets/find-wallet/: 25
=== dev.ethereum.org ===
/wallets/: 25
/wallets/find-wallet/: 25
```

[...]

PR #18812, opened 2026-07-15
fix: shard sitemap.xml per locale

### LINKED ISSUE #18014: [Recovery Agent] Sitemap.xml runs at runtime with empty mdSlugs — public/content/ excluded from Netlify lambda

myelinated-wackerow:
## Summary

Part 2 of #18014 (Part 1 was #18802, now merged). Splits the sitemap into one shard per locale so it clears Google's per-file limits, and adds a `<sitemapindex>` at `/sitemap.xml` that routes crawlers to those shards.

## Why

Even after #18802 makes the sitemap build statically (complete, no truncation), it's still a **single ~51 MB document with ~17.5k URLs and ~455k hreflang alternates** — over Google's **50 MB / 50k-URL per-file limits**, so crawlers reject it outright. Sharding is required for the sitemap to be accepted at all.

## Changes

- `app/sitemaps/sitemap.ts`: `generateSitemaps()` emits one shard per locale (`id = locale`), served at `/sitemaps/sitemap/<locale>.xml`. The default `sitemap({ id })` filters pages to that locale and keeps the **full hreflang alternates block**, so cross-language reciprocity is preserved. A memoized traversal lets all shards share a single content walk. Nested under `app/sitemaps/` so the reserved `/sitemap.xml` path is free for the index.
- `app/sitemap.xml/route.ts`: a route handler emitting a `<sitemapindex>` at the stable `/sitemap.xml`, enumerating every per-locale shard. Next's `generateSitemaps` never emits an index, so this supplies it — keeping `/sitemap.xml` (the URL already registered in Google Search Console) as the single canonical entry point instead of dropping it.
- `app/robots.ts`: points crawlers at the single `/sitemap.xml` index (which enumerates the shards), replacing the per-shard listing.
- `.gitignore`: anchors the `sitemap*.xml` rule to root (mirroring `/feed.xml`) so the `app/sitemap.xml/` route-handler directory stays tracked.

## Verification

- No route conflict; `/sitemap.xml` (index) and `/sitemaps/sitemap/<locale>.xml` (shards) both register and prerender **static** (no ISR).
- `/sitemap.xml` returns a well-formed `<sitemapindex>` with 25 `<sitemap>` entries, one per locale, each pointing at its shard (verified on the deploy preview).
- Every shard closes with `</urlset>`, carries ~700 URLs, and includes the sentinel `/wallets/find-wallet/` (the page that was invisible to crawlers in prod).
- Total across shards: 17,087 URLs. Largest shard ~2.01 MB — well under 50 MB, ~700 URLs each vs the 50k cap.

## Post-merge follow-up (not code)

`/sitemap.xml` is preserved as the index, so the URL registered in Google Search Console keeps resolving — **no GSC reconfiguration needed**.

A CI regression guard (parse each shard: valid XML, closing `</urlset>`, URL-count floor, sentinel late-alphabet URL) will follow as a separate PR — this is the check that was missing when #17718 was closed on a single lucky fetch.

-- Opus 4.8

## Related issue

- Closes #18014


pettinarip (commented):
Nice. Looks good on the codeside. No preview deploy for this PR. Maybe once the base PR gets merged, we can double check the generated sitemaps.

I'm fine with current approach, but was wondering about using sitemap index <https://developers.google.com/search/docs/crawling-indexing/sitemaps/large-sitemaps>. Have you seen that? is there any difference vs current implementation that uses the robots as the index.

wackerow:
@pettinarip Site preview is showing the `/sitemap/{locale}.xml` paths correctly... but to your point, we lose `/sitemap.xml` with this approach. Since the shards are generated by `generateSitemaps`, a route handler can't also claim `/sitemap.xml`...

So, going to push a commit that reclaims the `/sitemap.xml` path, acting as a pointer to the others (so GSC stays unaffected), but will nest the locale-specific sitemap shards further, inside a `/sitemaps` path. The `generateSitemaps` forces placing all of these in a `/sitemap` directory, so the locale shards will end up at `/sitemaps/sitemap/{locale}.xml`

wackerow:
https://deploy-preview-18812.ethereum.it/sitemap.xml

@pettinarip This is looking better to me. Let me know if you have any other thoughts, else can get this pulled in soon.

wackerow:
- Fixes #18854

pettinarip (approved):
LGTM

PR #18814, opened 2026-07-15
refactor: remove legacy useTranslation hook (closes #18742)

### LINKED ISSUE #18742: Migrate legacy useTranslation hook call sites to next-intl useTranslations

myelinated-wackerow:
The last 9 consumers migrate to namespace-bound `useTranslations`, and `src/hooks/useTranslation.ts` is deleted. This completes the migration tracked in #18742.

## Changes

- **`FindWalletProductTable` (8 files)** — all bind `page-wallets-find-wallet`. `useWalletFilters` gains a `tTable` binding for its three `table:` cross-namespace keys; `PersonaTags` / `useWalletPersonaPresets` dynamic keys resolve to plain labels (verified — no ICU/markup).
- **`ReleaseCarousel`** (rewritten in #18734, still on the legacy hook) — binds `page-roadmap`. The `t` passed into `release.content(t)` satisfies that function's existing `(key: string) => string` signature unchanged; all keys plain.
- **Deletes `src/hooks/useTranslation.ts`.**
- **Removes the `no-restricted-imports` guard** on the hook — redundant now that the module is gone (any new import fails type-check and the build on its own).
- **`AGENTS.md`** — drops the deprecation note that pointed at the removed hook.

## Verification

- [x] Repo-wide audit: **0** files import `@/hooks/useTranslation`
- [x] `pnpm type-check` clean; `pnpm lint` clean
- [ ] Deploy-preview sweep: `/wallets/find-wallet/` + `/roadmap/` × 25 locales (posted as comment)

With this merged, #18742 is fully resolved — all 103 original call sites migrated, both translation APIs collapsed to one.

-- Opus 4.8

## Related issue
- Closes #18742


myelinated-wackerow:

Against [deploy-preview-18814.ethereum.it](https://deploy-preview-18814.ethereum.it):

**1. Fetch sweep — 50/50 pages clean** (25 locales × `/wallets/find-wallet/`, `/roadmap/`): locale-resolved sentinels render (`page-find-wallet-persona-legend`, `page-roadmap-release-main-features`), no un-interpolated ICU args, no escaped markup, no dotted missing-key fallbacks. Production-calibrated first (also 50/50), so a failure here would be a true regression.

**2. Playwright interactions (en, ar, zh) — 0 failures**: find-wallet persona presets + the client-gated table filter strings (`useWalletFilters` `tTable` keys) and the sr-only persona count render correctly; roadmap release-status labels (`release.content(t)` path) resolve in all three.

Repo-wide audit confirms **0** importers of `@/hooks/useTranslation`; the module, the lint guard, and the AGENTS.md note are all removed. This closes #18742 — all original call sites migrated to a single translation API.

-- Opus 4.8

pettinarip (approved):
💪🏼

PR #18822, opened 2026-07-16
Deduplicate CoinGecko Ethereum metrics fetch

### LINKED ISSUE #18790: [recovery] CoinGecko 429 on ETH price simple/price fetch (demo-tier rate limit, concurrent hourly fetchers)
## Error Summary

- **Source:** Netlify logs (Grafana → Keep)
- **Fingerprint:** `grafana-fn-error-fetch-response-not-ok-from-https-api-coingecko-com-api-vn-simple-pric`
- **Status:** 429 (Too Many Requests)
- **Hit count:** 3 (first seen 2026-05-29 02:00:04.999)
- **Request ID:** `01KWRVF54HT017F0ZKRA3N5DDT`

**Message:**
```
[ERROR] ⨯ Error: Fetch response not OK from https://api.coingecko.com/api/v3/simple/price?ids=ethereum&vs_currencies=usd: 429 Too Many Requests
  digest: '1806199069'
```
(The logged URL is normalized/redacted — the live requests append `&include_24hr_change=true` / `&include_market_cap=true` and `&x_cg_demo_api_key=…`.)

## Analysis

The CoinGecko `simple/price` endpoint is fetched by **two** data-layer tasks, both scheduled **hourly** and both fired **concurrently** via `Promise.all` in the hourly orchestrator (`src/data-layer/tasks.ts:151`):

- `fetchEthPrice` → `simple/price?ids=ethereum&vs_currencies=usd&include_24hr_change=true` (`src/data-layer/fetchers/fetchEthPrice.ts:13`)
- `fetchEthereumMarketcap` → `simple/price?ids=ethereum&vs_currencies=usd&include_market_cap=true` (`src/data-layer/fetchers/fetchEthereumMarketcap.ts:13`)

A third hourly task (`fetchStablecoinsData`) also hits `api.coingecko.com` (`coins/markets`). All three fire in the same minute, producing a burst against CoinGecko's **demo/free tier**, which is the most likely trigger for the intermittent 429s.

Mitigations already in place — a demo API key (`x_cg_demo_api_key`, added in 120072002b) and retry-on-429 with backoff (`fetchRetry.ts`, up to 3 attempts) — reduce but don't eliminate the problem: the two `simple/price` calls are near-identical and redundant, and the concurrent burst can still exhaust the per-minute limit before retries succeed. When retries are exhausted the fetcher throws, the task fails, and the stored blob goes stale.

Hit count is low (3), so this is intermittent rather than a hard outage — but it is a recurring resilience gap, not a one-off.

## Affected Files

- `src/data-layer/fetchers/fetchEthPrice.ts` — ETH price `simple/price` fetch
- `src/data-layer/fetchers/fetchEthereumMarketcap.ts` — second, near-identical `simple/price` fetch
- `src/data-layer/tasks.ts` — hourly orchestrator fires all CoinGecko fetchers concurrently (`Promise.all`, line 151)
- `src/data-layer/fetchers/fetchRetry.ts` — shared retry/backoff config

## Suggested Approach

Several options, needing team input on tradeoffs:

1. **Deduplicate the two `simple/price` calls** — a single request with `include_24hr_change=true&include_market_cap=true` returns price, 24h change, and market cap together, halving the calls to that endpoint.
2. **Stagger / serialize CoinGecko fetchers** — avoid firing them in the same `Promise.all` tick; run sequentially or offset their schedules.
3. **Graceful fallback on 429** — return the last stored blob value instead of throwing, so stale data is served rather than the task failing.
4. **Tune retry backoff for 429** — respect `Retry-After` and use a longer initial delay than the current shared config.
5. **Verify the demo key's tier** — confirm `COINGECKO_API_KEY` is valid and whether a higher tier is warranted.

## Confidence Assessment

Low confidence — root cause is external (CoinGecko demo-tier rate limiting), not a single-file code defect. Multiple valid fixes exist with different tradeoffs (dedup, staggering, fallback, tier upgrade), several of which are architectural and need team input. Related precedent: #17825 (Beaconcha.in 429) — same class, different resource.

---

alerodriargui:
## Description
Combines the hourly Ethereum price and market cap requests into a single CoinGecko /simple/price call, reducing concurrent API usage and the likelihood of 429 rate-limit errors.
The combined response is stored under the existing price and market cap keys, so getters and consumers remain unchanged. The Demo API key is also moved from the query string to the official x-cg-demo-api-key header.
Adds tests covering the combined response mapping, optional 24-hour change, and invalid responses.

## Related Issue
Closes #18790


myelinated-wackerow:
Thanks @alerodriargui

A few structural notes before this can go in:

**Register through the task factory, not a bespoke task.** `ethereumMetricsFetchTask` re-implements `createDataTask`'s `retry` / `catchError` / `logger` boilerplate, which will drift the first time we change retry behavior. Widening `TaskDef` to accept multiple keys keeps every hourly task on one code path:

```ts
type TaskDef = [string | string[], () => Promise<unknown>]
```

**The task id should describe what the task does.** It's registered as `KEYS.ETH_PRICE` but writes both keys, so a market-cap failure alerts as `[fetch-eth-price] failed`, and `fetch-ethereum-marketcap` disappears from the trigger.dev dashboard along with its run history. Please add a `KEYS.ETHEREUM_METRICS` entry and use that as the id.

**Drop the `Extract<>` gymnastics.** `SuccessfulEthPriceData` / `SuccessfulMetricReturnData` narrow away an error variant neither fetcher ever returned. Declaring the return shape directly is clearer.

**Coupling regression.** An invalid or missing `usd_market_cap` now throws before the price is stored, where the two tasks used to fail independently. One call instead of two lowers the failure rate, but doubles the blast radius when it does fail. Worth either splitting the validation so a usable price still gets written, or adding the fallback-to-stale behavior discussed in #18790.

**`fetchStablecoinsData` is the third hourly CoinGecko caller** and is untouched: still key-in-query-string, still logs the full URL (so the key leak persists there), still fires in the same `Promise.all` tick. The header/log fix should be applied there too, otherwise the concurrent-burst cause in the issue is only half addressed.

**Formatting/lint.** CI is currently red on `simple-import-sort/imports` at `tasks.ts:9`, and several new lines aren't Prettier-formatted. Please run `pnpm lint:fix && pnpm format` and push -- we can't merge with the lint job failing.

-- Opus 5

PR #18825, opened 2026-07-16
fix: stop root not-found from poisoning request locale in dev

pettinarip:
## Description

Fixes the dev-only bug where non-English pages render with mixed languages — server components in English, client components in the page's locale (e.g. `/es/developers/tutorials` with an English hero/nav and Spanish filters).

## Root cause

In dev, Next.js renders the global `app/not-found.tsx` boundary alongside **every** full-document request, sharing the request-scoped React cache with the actual page. Because it reused `LocaleLayout` hardcoded to `DEFAULT_LOCALE`, every request fired `setRequestLocale("en")` plus an implicit `getMessages()` — resolving and caching next-intl's implicit request config as English before the real page's `setRequestLocale(locale)` could take effect (next-intl memoizes the first implicit resolution per request).

- Server components (`getTranslations()` / `getLocale()`) → poisoned English
- Client components → correct locale, because pages pass an **explicit** `getMessages({ locale })` catalog to the provider (separate cache key)

Production is unaffected: `/_not-found` is prerendered in isolation at build time. The bug surfaced when the language picker moved from soft nav to a hard refresh, since RSC-payload requests don't co-render the not-found boundary the way full document loads do.

## Fix

Make the root not-found self-contained: its own `<html>` document using only explicit-locale APIs and no `setRequestLocale`, so it can never write to or resolve the shared request-locale store. Localized 404s keep the full site chrome via `app/[locale]/not-found.tsx`; the root 404 (only reachable for extension-style paths the proxy matcher excludes) drops the nav/footer chrome, since `Nav`/`Footer` are server components with implicit i18n calls that would reintroduce the race. Fonts move to a shared `app/fonts.ts` used by both documents.

## Verified (dev server)

- `/es/developers/tutorials/`: fully Spanish server + client output (was mixed)
- `/en/developers/tutorials/`: fully English
- Root 404 path (`/foo.txt`): renders the 404 page correctly
- `tsc --noEmit` + ESLint pass

### DIFF
--- a/app/[locale]/layout.tsx
+++ b/app/[locale]/layout.tsx
@@ -1,6 +1,5 @@
 import { Suspense } from "react"
 import { pick } from "lodash"
-import { IBM_Plex_Mono, Inter } from "next/font/google"
 import { notFound } from "next/navigation"
 import { getMessages, setRequestLocale } from "next-intl/server"
 
@@ -11,6 +10,8 @@ import Matomo from "@/components/Matomo"
 import { getLastDeployDate } from "@/lib/utils/getLastDeployDate"
 import { getLocaleTimestamp } from "@/lib/utils/time"
 import { toLanguageTag } from "@/lib/utils/url"
+
+import { ibmPlexMono, inter } from "../fonts"
 
 import Providers from "./providers"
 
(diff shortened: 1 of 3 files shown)


wackerow (approved):
Nice, thanks @pettinarip.. this is working for me locally with `pnpm dev`. The menu/footer/heros all loading correct intl strings now. "Not Found" page also loading correctly in non-English locales.

PR #18832, opened 2026-07-17
Fix: EIP and ERC omission fix for better understanding

HiroyukiNaito:
There are several omissions for EIP-7702 and ERC-4337, like 7702, EIP 7702, 4337, ERC 4337.
For better readability, not only Etherian but also all, I think these are integrated into one format. 

## Description

All parts 7702 and 4337 into EIP-7702 and ERC-4337.
If you disagree, please feel free to close this PR.

## Related Issue

None.


wackerow (approved):
Clean, thanks @HiroyukiNaito!

PR #18845, opened 2026-07-20
refactor: extract config-driven FilterableCatalog from ToolsCatalog (find-wallet revamp PR 1)

pettinarip:
**PR 1 of the find-wallet revamp.** Extracts a reusable catalog component from `ToolsCatalog` and migrates `/developers/tools` onto it with zero behavior change. PR 2 rebuilds `/wallets/find-wallet` on top of it.

## What

- `src/components/FilterableCatalog/` — the core owns the shared machinery (layout shell, search, filter state, the `useDeferredValue` pipeline, count header, no-results). Everything page-specific is composed in: consumers draw filters (`renderSidebar`), decide membership (`filterFn`), and own results markup (`renderResults`).
- Filters ship as presentational, controlled building blocks — `CatalogNavGroup` and `CatalogCheckboxGroup` (value in, event out; no knowledge of how state is stored). Consumers compose them or drop in bespoke controls; a new filter UI needs no core change.
- `ToolsCatalog` shrinks from a 435-line monolith to a thin consumer.

## Why composition over a config union

An earlier revision drove filters from a discriminated union the core interpreted via a `switch`. The two members had disjoint shapes and each consumer used exactly one — every new affordance would have meant growing the union *and* the core. Composition keeps the core small and closed over what's genuinely shared, and leaves filter UI open-ended for PR 2.

## Verification

- Catalog DOM for both dev-tools pages is **byte-identical** to before the refactor.
- Visual-test coverage for `/developers/tools` added in a **separate first commit** so Chromatic's baseline predates the change.
- `type-check` / lint / Prettier clean; both building blocks covered by a Storybook story.

## Notes

- Matomo event-naming-via-config is deferred to PR 2 (dev tools emits no events today — it would be a dark code path here).

### DIFF
--- a/app/[locale]/developers/tools/_components/ToolsCatalog.tsx
+++ b/app/[locale]/developers/tools/_components/ToolsCatalog.tsx
@@ -3,9 +3,10 @@
 import { memo, useMemo } from "react"
 
 import FilterableCatalog from "@/components/FilterableCatalog"
+import CatalogNavGroup from "@/components/FilterableCatalog/CatalogNavGroup"
 import type {
   CatalogFilterState,
-  CatalogNavGroup,
+  CatalogNavGroupConfig,
 } from "@/components/FilterableCatalog/types"
 
 import type {
@@ -18,6 +19,12 @@ import { numberFormat } from "@/lib/utils/numbers"
 import ToolCard from "./ToolCard"
 
 const SUBCATEGORY_FILTER_KEY = "subcategory"
+
+const PATH_SEPARATOR = "\u00A0\u00A0/\u00A0\u00A0"
+
+function formatPathSegment(value: string): string {
+  return value.toLocaleUpperCase()
+}
 
 type ToolsCatalogProps = {
   locale: string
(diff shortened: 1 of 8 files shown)


app/[locale]/developers/tools/_components/ToolsCatalog.tsx
```diff
@@ -311,124 +150,164 @@ export default function ToolsCatalog({
...
+      tool.name,
+      tool.description,
+      getCategoryLabel(tool.categoryId, categoryLabels),
+      getSubcategoryLabel(tool.subcategory_id, subcategoryLabels),
+      ...tool.tags,
+    ]
+      .join(" ")
+      .toLowerCase()
+    return searchableText.includes(normalizedQuery)
+  }
```
Comment (wackerow) [inline app/[locale]/developers/tools/_components/ToolsCatalog.tsx:258]: Small note, this would violate the `filterFn` doc it ships with. `types.ts`/`index.tsx` warn to pass a stable `filterFn` reference, but this `filterTool` is defined inline, unmemoized.

Benign here, no state post-mount so stable, but for PR 2's `find-wallet` consumer would be worth using `useCallback`. If we do that here, it'd be more copy-paste-safe for the next usage of it.

Same would apply to inline navConfig/renderSidebar/renderResultsHeader, though those only feed cheap renders.

wackerow (approved):
Looks good! Thanks @pettinarip... approving with a small non-blocking note.

app/[locale]/developers/tools/_components/ToolsCatalog.tsx
Comment (pettinarip) [reply app/[locale]/developers/tools/_components/ToolsCatalog.tsx:258]: Yeah, agreed. ToolsCatalog is stable in this case but should be wrapped in usecallback otherwise

PR #18846, opened 2026-07-20
feat: rebuild find-wallet on the shared catalog (revamp PR 2)

pettinarip:
## Find Wallet revamp — PR 2 of 2

Rebuilds `/wallets/find-wallet` on the shared `FilterableCatalog` from PR 1 (#18845, merged), mirroring the `/developers/tools` architecture: static server shell + one client island + parallel/intercepting routes. Driver: **INP on slow mobile devices** — the old expandable-card table was the bottleneck.

Covers the index, persona pages (`personas/[persona]/`), and wallet detail as either a modal or a standalone page. Deletes `FindWalletProductTable/` and its now-dead helpers (**−2,395 lines**).

## Deliberate decisions

- **Narrower filter set at launch** — the 16 "Advanced" filters are out of the sidebar (blocked on design). They stay in the data and on detail pages.
- **AND semantics** across Devices / Buy-Sell / Networks; **Language is OR**.
- **Detail routes are deliberately not prebuilt** — `generateStaticParams` returns `[]` so a build-time prerender can't permanently shadow the `@modal` interceptor (reasoning in the comment there). `revalidate = false` throughout, since wallet data is repo-checked-in.
- **Matomo events are new series, not continuations.** Only the event *category* is preserved: filter actions moved to locale-independent data keys, wallet-open is now `find-wallet`/`open_wallet_modal`, social links dropped their row-index suffix, and catalog search is new. The A/B should get its own goals per arm rather than reusing the old ones.
- **Wallet descriptions are optional.** No wallet ships one today, so detail meta and JSON-LD fall back to a sentence synthesized from device/network/language data; a real description replaces it with no code change.

## Verification

`type-check`, `eslint`, `prettier` and a production build all clean. INP measured on Pixel 5 emulation at 6× CPU throttle: every main interaction stays inside the "Good" band. e2e green on Chromium desktop + mobile.

## Follow-ups

- Backfill real wallet descriptions — the data field and the submission form already ask for them — plus a descriptions i18n namespace like apps and dev tools have.
- Persona slug `hardware` is provisional; content may rename it before launch, which fossilizes the slug.
- Cold modal content ~1.7 s @6× throttle. Worth confirming on the preview whether prefetch or a cacheable interceptor is the answer.
- `E2E tests`, `Build website (mock data)` and `Page visual snapshots` all gate on `master`/`staging`/`test/*`, so they skip on this `dev`-targeted PR — the rewritten specs and the new visual pages get no CI run here and surface on the release PR.


konopkja:
Hey @pettinarip 👋 — we want to run a **causal A/B test** (old table vs this new catalog) on `/wallets/find-wallet` to validate the redesign, mirroring how we measured the homepage revamp.

I've created the Matomo experiment as a **draft**: `FindWalletCatalog2026` (id 15) — 50/50 **Original vs NewCatalog**, co-primary metrics **bounce rate + wallet-visit conversion** (Goal 10 = `event_action = "Tap main button"`, the main "Visit website" button), MDE 5%, EN-only, enrollment on the index URL only.

For that test to work, the redesign needs a few hooks. Flagging what's on the dev side — happy to pair or take some of these as a follow-up PR.

### Blocking (test can't run without these)

1. **Keep the old find-wallet alive for the test.** This PR deletes `FindWalletProductTable/` (−2,395 lines). For a real old-vs-new split we need the old tree still reachable at an internal shadow route (e.g. `/wallets/find-wallet-classic`, `noindex`). Options: don't delete it in this PR, or restore it behind that route in a small companion PR.
2. **Add an Edge `middleware.ts`** (the repo has none today). It should: assign 50/50 using the existing FNV-1a fingerprint in `src/lib/ab-testing/server.ts`, **rewrite the Original cohort** to the shadow route while keeping the public URL `/wallets/find-wallet`, and pass the assigned variation to the page (request header) so tracking can fire. The current `ABTestWrapper` only swaps a component subtree on one route — it can't split two route architectures, which is why middleware is needed here.
3. **Fire `AbTesting::enter` on BOTH variations** (reuse `ABTestTracker`). Experiment name must be exactly **`FindWalletCatalog2026`** and variation names exactly **`Original`** / **`NewCatalog`**.
4. **Main-button conversion event — verified, no action needed.** The wallet click-through is tracked via `WalletExternalLinkList` / `eventAction: "Tap main button"` (eventName = wallet name). This PR **already preserves it** identically to the current live design (`WalletInfo.tsx` → the new detail + `@modal` pages), so Goal 10 measures both variations correctly. *(Correcting an earlier version of this comment: the primary conversion string is `Tap main button`, not `Go to wallet` — the latter is only the rarely-clicked Discord/Twitter links, ~349/mo vs ~14.6k/mo. Nothing to change on your side.)*

### Important (data quality)

5. **Cache correctness.** Per-user rewrite means the index can't be one shared CDN entry or we get cross-variation leakage (~1.5% last time). Vary cache on the variation, or dynamic-render the index for the test window only. This conflicts with `revalidate = false` — scope it to the experiment duration.
6. **Locale handling.** Enrollment is default-locale (EN) only. Please decide non-EN behavior — recommendation: non-EN → NewCatalog (unenrolled), so the INP win isn't withheld from the international majority.
7. **Do NOT start the experiment from the code side.** Deploy first, we verify assignment + event firing in prod, then the analytics side starts Experiment 15. (On the homepage test the experiment ran ~5 days before the code deployed and contaminated the data — avoiding that this time.)

### Out of scope for this test

- The **SEO benefit** of the persona/`[wallet]` subpages can't be A/B tested (per-user split can't test Google indexing). We'll measure that separately in Search Console before/after over 4–8 weeks — no dev action needed.

If the team would rather **not** do the full causal split (it does mean keeping the old code alive + an ISR trade-off for the test window), the fallback is a before/after comparison on the preserved event categories — lighter on eng but not causal. Your call on which way to go; let me know and I'll adjust the Matomo setup accordingly. 🙏

app/[locale]/wallets/find-wallet/[wallet]/page.tsx
```diff
@@ -0,0 +1,268 @@
...
+  setRequestLocale(locale)
+
+  const wallet = getWalletBySlug(walletSlug, locale)
+  if (!wallet) return {}
+
+  return await getMetadata({
+    locale,
+    slug: ["wallets", "find-wallet", walletSlug],
+    title: wallet.name,
+    description: wallet.descriptionStripped?.slice(0, 160) ?? wallet.name,
```
Comment (myelinated-wackerow) [inline app/[locale]/wallets/find-wallet/[wallet]/page.tsx]: **Ask:** worth a synthesized fallback here.

No wallet has a `description` today — verified, zero `description:` fields in `wallet-data.ts` — so every one of the 49 detail pages ships `<meta description>` = the bare wallet name. And `translationRegistry` now registers those 49 slugs across every translated locale, so that is ~1,200 sitemap entries carrying the same one-word English string. The detail JSON-LD omits `description` for the same reason.

Everything a fallback needs is already computed on this page — device buckets, chain count, language count — so something like `MetaMask — Ethereum wallet for browser, mobile and desktop. Supports N networks and M languages.` fixes all of them at once, and real descriptions later just replace it.

app/[locale]/wallets/find-wallet/[wallet]/page.tsx
```diff
@@ -0,0 +1,268 @@
...
+  )
+}
+
+export function generateStaticParams() {
+  // Deliberately empty: these pages must render on demand so the response
+  // carries `Vary: Next-Url`. Build-time prerenders have no response to read,
+  // so Next marks them "cannot be intercepted" and the client then caches this
+  // URL ignoring Next-Url, which permanently shadows `@modal/(.)[wallet]`.
+  return []
+}
```
Comment (myelinated-wackerow) [inline app/[locale]/wallets/find-wallet/[wallet]/page.tsx:252]: Doc drift: the PR body says `generateStaticParams` covers "49 English wallet pages". It deliberately returns `[]` now — correctly, per this comment. Worth fixing the body so nobody reads those pages as prebuilt.

app/[locale]/wallets/find-wallet/@modal/(.)[wallet]/page.tsx
```diff
@@ -0,0 +1,216 @@
...
+    <div className="min-w-0 text-end text-sm">{children}</div>
+  </div>
+)
+
+/**
+ * Shown as a modal when navigated to from inside the find-wallet subtree; a
+ * direct load or refresh renders the standalone `[wallet]` page instead. Body
+ * is server-rendered and passed to the client shell as children.
+ */
+export default async function InterceptedWalletModal(props: {
```
Comment (myelinated-wackerow) [inline app/[locale]/wallets/find-wallet/@modal/(.)[wallet]/page.tsx:77]: Hypothesis for the ~1.7s cold modal content you flagged: this route exports no `revalidate`, so unlike the standalone `[wallet]` page (`revalidate = false`) the intercepted render is not cacheable at all — every modal open is a fresh dynamic render. Dev-tools' interceptor has the identical gap, so if that is the cause, one export fixes both pages.

Counter-risk, honestly: this may be exactly what the `Vary: Next-Url` reasoning in your `generateStaticParams` comment warns against. A cached on-demand interceptor response is a different failure mode from a build-time prerender, but I cannot assert from reading that it is safe — worth measuring on the preview rather than changing blind.

app/[locale]/wallets/find-wallet/@modal/(.)[wallet]/page.tsx
```diff
@@ -0,0 +1,216 @@
...
+        <div className="flex flex-col gap-2">
+          {wallet.supported_chains.length > 0 && (
+            <DetailRow label={t("page-find-wallet-network-support")}>
+              <ChainImages
+                chains={wallet.supported_chains as ChainName[]}
+                className="flex-wrap justify-end gap-1"
+                nested
+              />
+            </DetailRow>
+          )}
```
Comment (myelinated-wackerow) [inline app/[locale]/wallets/find-wallet/@modal/(.)[wallet]/page.tsx]: `ChainImages` filters out any chain missing from `networkData`, so this guard can pass and still render an empty row — same on the standalone page (`[wallet]/page.tsx:130`).

Same root cause lets the sidebar's Networks group, built from raw `supported_chains` in `getWalletNetworks`, offer options with no visual counterpart.

app/[locale]/wallets/find-wallet/_components/WalletFilters.tsx
```diff
@@ -0,0 +1,249 @@
...
+  ADVANCED_KEY,
+] as const
+
+// Category kept from the old sidebar for trend comparability.
+const trackFilterToggle = (action: string, name: string) =>
+  trackCustomEvent({
+    eventCategory: "WalletFilterSidebar",
+    eventAction: action,
+    eventName: name,
+  })
```
Comment (myelinated-wackerow) [inline app/[locale]/wallets/find-wallet/_components/WalletFilters.tsx:37]: **Ask (docs only):** the PR body says event strings were "preserved for trend comparability where the interaction is genuinely unchanged (`WalletFilterSidebar`, `WalletExternalLinkList`)". Only the *category* survived here — the old action was `` `${locale} - ${t(label)}` `` (e.g. `en - Desktop`), this is the bare data key. Same story for `WalletExternalLinkList`, where the social-link `eventName` lost its row-index suffix.

Both changes are improvements: locale-independent keys, and names that actually aggregate. It is just that you were precise about `expanded` → `open_wallet_modal` and this sentence is not, and it is the sentence goal definitions get written against. Full before/after in the review summary for @konopkja.

src/components/FilterableCatalog/index.tsx
```diff
@@ -148,8 +275,17 @@ export default function FilterableCatalog<TItem>({
...
+              {labels.noResultsDesc && (
+                <p className="m-0 max-w-prose">{labels.noResultsDesc}</p>
+              )}
+              {labels.resetLabel && (
+                <Button variant="outline" onClick={resetAll} className="mt-1">
+                  <RotateCcw className="size-4" />
+                  {labels.resetLabel}
+                </Button>
+              )}
             </div>
```
myelinated-wackerow, line 289: The old `FindWalletsNoResults` fired `Wallet_empty_state` / `reset` / `triggered` whenever this state rendered. Nothing fires here now, so "how often do users hit zero results" goes dark — only `reset_button_clicked` survives, and that counts just the people who click the button.

Worth knowing the old one was a render-phase side effect, so it multi-fired per re-render and was never a usable baseline either. So if zero-result rate matters for the A/B, this is the moment to add it properly — a `useEffect` firing once when `filteredItems.length` hits 0 — rather than restore what was there. @konopkja's call.

src/components/FilterableCatalog/index.tsx
```diff
@@ -69,71 +101,166 @@ export default function FilterableCatalog<TItem>({
...
+      () =>
+        trackCustomEvent({
+          eventCategory: "CatalogSearch",
+          eventAction: "search",
+          eventName: query,
+        }),
+      SEARCH_TRACKING_DEBOUNCE_MS
+    )
+    return () => clearTimeout(timeout)
+  }, [search])
```
myelinated-wackerow, line 153: This lives in the shared component, so `/developers/tools` starts emitting `CatalogSearch` too — a page outside this PR's stated scope. Reads intentional to me, given the component is shared by design.

One small thing worth a conscious yes: raw user queries land in `eventName`, so worth confirming Matomo's anonymization covers that field.

src/components/FilterableCatalog/utils.ts
```diff
@@ -1,6 +1,11 @@
...
   return ids.includes(id)
     ? ids.filter((existing) => existing !== id)
     : [...ids, id]
 }
+
+/** Normalize a filter-state slot to an array; a single-select string widens. */
+export const asArray = (value: string | string[] | undefined): string[] => {
+  if (Array.isArray(value)) return value
+  return value === undefined ? [] : [value]
+}
```
myelinated-wackerow, line 11: `asArray("")` returns `[""]`, which downstream reads as one active filter. Unreachable today since every writer passes an array or `undefined`, but `value ? [value] : []` closes it before the next single-select consumer finds it.

app/[locale]/wallets/find-wallet/_components/FindWalletBreadcrumbs.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </BreadcrumbItem>
+        {separator}
+        <BreadcrumbItem>
+          <BreadcrumbLink href="/wallets">{t("wallets")}</BreadcrumbLink>
+        </BreadcrumbItem>
+        {separator}
+        <BreadcrumbItem>
+          <BreadcrumbLink href="/wallets/find-wallet">
+            {t("find-wallet")}
+          </BreadcrumbLink>
```
Comment (myelinated-wackerow) [inline app/[locale]/wallets/find-wallet/_components/FindWalletBreadcrumbs.tsx:51]: These two hrefs lack trailing slashes under `trailingSlash: true`, so each click takes a 308. The dev-tools breadcrumbs you mirrored use `/developers/` and `/developers/tools/`.

Also, `ms-[0.625rem] me-[0.625rem]` on the separator (line 33) is just `mx-[0.625rem]`.

app/[locale]/wallets/find-wallet/_components/WalletPersonaCards.tsx
```diff
@@ -0,0 +1,121 @@
...
+}: WalletPersonaCardsProps) => {
+  const t = await getTranslations({
+    locale,
+    namespace: "page-wallets-find-wallet",
+  })
+
+  return (
+    <nav
+      className="relative min-w-0 overflow-x-clip"
+      aria-label={t("page-find-wallet-persona-legend")}
```
Comment (myelinated-wackerow) [inline app/[locale]/wallets/find-wallet/_components/WalletPersonaCards.tsx:36]: `page-find-wallet-persona-legend` is "Filter wallets by user type", but these are real page links now, not filters — the label describes the old behaviour.

src/data/wallets/personas.ts
```diff
@@ -0,0 +1,93 @@
...
+  | "tag-yellow"
+
+/** Keep in sync with `PERSONA_COLORS` in WalletPersonaCards. */
+export const PERSONA_TAG_STATUS: Record<WalletPersonaId, PersonaTagStatus> = {
+  "new-to-crypto": "primary",
+  developer: "accent-b",
+  finance: "accent-c",
+  hardware: "accent-a",
+  nfts: "tag-yellow",
+}
```
myelinated-wackerow: Persona colour lives in two places — this map and `PERSONA_COLORS` in `WalletPersonaCards` — held together by a comment. Since both are keyed by the same `WalletPersonaId`, one map carrying `{ tag, text, border, bgTint }` would make desync impossible rather than merely commented against.

app/[locale]/wallets/find-wallet/page-jsonld.tsx
```diff
@@ -97,8 +88,8 @@ export default async function FindWalletPageJsonLD({
             name: wallet.name,
             url: wallet.url,
             applicationCategory: "Cryptocurrency Wallet",
-            ...(platforms(wallet).length > 0 && {
-              operatingSystem: platforms(wallet).join(", "),
+            ...(getWalletPlatforms(wallet).length > 0 && {
+              operatingSystem: getWalletPlatforms(wallet).join(", "),
```
Comment (myelinated-wackerow) [inline app/[locale]/wallets/find-wallet/page-jsonld.tsx]: `getWalletPlatforms(wallet)` runs twice per wallet — once to guard, once for the value. A local hoists it, as the detail JSON-LD already does with `const os`.

Unrelated, same file: line 22 uses `getTranslations("page-wallets-find-wallet")` while both sibling JSON-LD files use the explicit `{ locale, namespace }` form. Works via `setRequestLocale`; just inconsistent.

tests/e2e/find-wallet.spec.ts
```diff
@@ -15,42 +15,74 @@ test.describe("Find Wallet Page", () => {
...
   })
 
-  test("personas filter updates counter and list", async () => {
-    const personaCount =
-      await findWalletPage.selectPersonaFilter("New to crypto")
+  test("persona card navigates to a persona page", async () => {
+    await findWalletPage.openPersona("nfts")
+    const count = await findWalletPage.getResultsCount()
+    expect(count).toBeGreaterThan(0)
+    expect(count).toBeLessThanOrEqual(49)
```
myelinated-wackerow: `49` is today's wallet count, so this assertion silently tracks the data file — it will fail the day the 50th wallet lands, for a reason unrelated to what the test checks. Asserting against the `/ total` half of the counter instead would keep it self-referential.

tests/e2e/pages/FindWalletPage.ts
```diff
@@ -4,245 +4,96 @@ import { testData } from "../fixtures/testData"
...
-    await this.mobileFiltersSubmitButton.click()
-
-    return { initialCount, osOptions }
+  private async clickFilterOption(label: string) {
+    await this.page
+      .locator("label:visible")
+      .filter({ hasText: new RegExp(`^${label}`) })
+      .first()
+      .click()
   }
```
myelinated-wackerow: `^${label}` prefix-matches, so `toggleDeviceFilter("Hardware")` also matches "Hardware wallet support" in the Advanced group — it works only because the Device group precedes it in the DOM and `.first()` wins. Scoping to the calling group, or anchoring the regex at both ends, would make it robust to sidebar reordering.

myelinated-wackerow (commented):
Reviewed the full diff — **no correctness blockers.** The architecture mirrors `/developers/tools` faithfully, the `getSupportedLanguages` splice-mutation trap is correctly defended with a copy in `enrichWallet`, and `seededShuffle`-then-stable-`sort` in `getRelatedWallets` really does preserve the shuffle within persona-overlap ties. Everything specific is inline; two of those are asks, the rest are FYI.

**Asks:** a synthesized fallback for the detail-page meta description (`[wallet]/page.tsx`), and one PR-body sentence tightened (`WalletFilters.tsx`).

### Analytics delta — cc @konopkja

Every change below is a defensible improvement. The point is only that these are new series rather than continuations, so the A/B wants its own goal per arm rather than reusing the old ones.

| interaction | before | after |
| --- | --- | --- |
| filter toggle (device, buy/sell, advanced) | action `en - Desktop`, locale-prefixed translated label. Two strays: `privacy` had no locale prefix, `gas_fee_customization` used `advanced_features` | action `desktop`, bare data key, locale-independent. eventName format unchanged (`desktop true`) |
| network toggle | `network` / `<chain>`, fired on add only | identical strings, now fires on add **and** remove |
| wallet open | `find-wallet` / `expanded` / `<name>` | `find-wallet` / `open_wallet_modal` / `<name>` — already documented in the body |
| social links | `Go to wallet` / `Discord: MetaMask 3` | `Go to wallet` / `Discord: MetaMask` — row-index suffix gone, so these finally aggregate |
| website link | `Go to wallet` / `Website: MetaMask 3` | no equivalent; the main CTA fires `Tap main button` / `<name>` on both detail and modal |
| zero results reached | `Wallet_empty_state` / `reset` / `triggered` | **nothing** — see the inline note on `FilterableCatalog` |
| search | — | `CatalogSearch` / `search` / `<query>`, 1.2s debounce, and now also on `/developers/tools` |

### Two notes with no line to attach to

**Orphaned English keys** now the table is gone — zero references anywhere: `page-find-wallet-showing-all-wallets`, `-search-languages`, `-active`, `-wallets`, `-check-out`, `-social-links`, `-choose-features`, `-see-wallets`, `page-stake-eth`, and `-try-removing` (its one remaining hit is a same-named key in the tutorials namespace, not this one).

**This lands with the tests unexercised.** `E2E tests`, `Build website (mock data)` and `Page visual snapshots` all gate on `base_ref == master | staging | test/*`, so they skip on a `dev`-targeted PR. The rewritten spec (old suite deleted, 7 new tests) and the two new visual pages get no CI run here — which also means your own open item, "the 2 desktop client-nav e2e tests were flaky locally, validate on CI's Playwright image", cannot close on this PR. Not your doing; just worth knowing the release PR is where it surfaces.

— Opus 5

app/[locale]/wallets/find-wallet/_components/FindWalletBreadcrumbs.tsx
Comment (pettinarip) [reply app/[locale]/wallets/find-wallet/_components/FindWalletBreadcrumbs.tsx:51]: Done — both hrefs slashed, and the separator is `mx-[0.625rem]` now.

src/components/FilterableCatalog/utils.ts
pettinarip: Done.

app/[locale]/wallets/find-wallet/page-jsonld.tsx
Comment (pettinarip) [reply app/[locale]/wallets/find-wallet/page-jsonld.tsx]: Done — hoisted to `const os`, and switched to the explicit `{ locale, namespace }` form.

app/[locale]/wallets/find-wallet/_components/WalletPersonaCards.tsx
Comment (pettinarip) [reply app/[locale]/wallets/find-wallet/_components/WalletPersonaCards.tsx:36]: Done — now "Browse wallets by user type".

src/data/wallets/personas.ts
pettinarip: Done — one `PERSONA_STYLES` map carrying the chip status too; `PERSONA_TAG_STATUS` is gone.

tests/e2e/find-wallet.spec.ts
pettinarip: Done — asserts against the counter's own total instead.

tests/e2e/pages/FindWalletPage.ts
pettinarip: Done — scoped to the group's Collapsible content via Radix' `aria-controls`.

app/[locale]/wallets/find-wallet/@modal/(.)[wallet]/page.tsx
Comment (pettinarip) [reply app/[locale]/wallets/find-wallet/@modal/(.)[wallet]/page.tsx]: Guard done — `getRenderableChains` is exported from `ChainImages` and both detail views gate on it. Left the sidebar alone: those options are text + count with no logo, so they still filter correctly, and non-EVM chains like Starknet could never render one.

app/[locale]/wallets/find-wallet/[wallet]/page.tsx
Comment (pettinarip) [reply app/[locale]/wallets/find-wallet/[wallet]/page.tsx]: Done — synthesized fallback in `buildWalletMetaDescription`, behind one ICU key, wired into `generateMetadata` and both detail JSON-LD nodes. `descriptionStripped` still wins, so real descriptions replace it with no code change.

`MetaMask — Ethereum wallet for Mobile and Browser. Supports 11 networks and 43 languages.`
`Ready Wallet — Ethereum wallet for Mobile and Browser. Supports 1 network and 1 language.`

`(devices, networks, languages)` is unique for 44 of the 48 wallets, so these are near-unique rather than boilerplate. Counts go through `numberFormat()` and are passed next to the raw numbers, since ICU `#` would bypass the numbering system `ur` needs.

Separately: `WalletData.description` is fully plumbed already (modal, detail, JSON-LD, meta) and `suggest_wallet.yaml` requires a description from every submitter — we just never land them. Worth a follow-up to backfill + a `page-wallet-descriptions` namespace like apps/tools have.

app/[locale]/wallets/find-wallet/[wallet]/page.tsx
Comment (pettinarip) [reply app/[locale]/wallets/find-wallet/[wallet]/page.tsx:252]: Done — PR body rewritten; it now says the detail routes are deliberately not prebuilt.

app/[locale]/wallets/find-wallet/_components/WalletFilters.tsx
Comment (pettinarip) [reply app/[locale]/wallets/find-wallet/_components/WalletFilters.tsx:37]: Done — PR body now says only the event *category* is preserved and that these are new series, with the A/B needing its own goals per arm.

src/components/FilterableCatalog/index.tsx
pettinarip: Checked: Matomo has no setting that covers this. `PrivacyManager` anonymizes visit-level data (IP, location, User ID, referrer) — custom event names are stored verbatim, so raw queries would land in the database as typed. Leaving this open for @konopkja: either accept that, or send something coarser (query length, or just that a search happened). Happy to change it either way — the `/developers/tools` blast radius makes it worth deciding before merge rather than after.

pettinarip:
Picked up the two notes with no line to attach to:

- **Orphaned keys** — removed 9 of the 10 from `page-wallets-find-wallet` (`page-find-wallet-choose-features` doesn't exist in `src/intl` at all, so nothing to drop). English only; translations follow the usual sync.
- **CI gating** — acknowledged, noted in the PR body. The rewritten e2e specs and the two new visual pages get no run on a `dev`-targeted PR, so the flaky client-nav check can only close on the release PR.

app/[locale]/wallets/find-wallet/@modal/(.)[wallet]/page.tsx
Comment (pettinarip) [reply app/[locale]/wallets/find-wallet/@modal/(.)[wallet]/page.tsx:77]: Done — RSC responses are cache HITs now, and interception still holds on both the Full details and cold-load paths.

konopkja (commented):
Reviewed only the three `page-jsonld.tsx` files here, as part of a wider structured-data audit (#18977–#18983, #18995, #18996). Nothing below blocks the INP work this PR is actually about.

**Two things this PR gets right that the older files don't**, worth calling out because both have been recurring bug sources:

- All three files use `getTranslations({ locale, namespace: "page-wallets-find-wallet" })` — the correct namespace, with the locale pinned explicitly. The equivalent bug shipped twice on adjacent files (#17719, and still live on `/wallets/` and `/quizzes/` per #18979), so avoiding it here matters.
- URLs are built with trailing slashes, matching the canonical convention. Most markdown-route pages currently mismatch their own canonical (noted in #18981), so these new pages start clean.
- `buildWalletMetaDescription` composes from factual data through a translated template with correct plural selection and locale-aware numbering, and returns `undefined` rather than a half-formed string when data is missing. That's a synthesized *fact*, not generated prose — good call.

Three suggestions:

---

**1. Give the `ItemList` items a stable `@id` (highest value, cheapest)**

The `SoftwareApplication` items inside both `ItemList`s have no `@id`, while the detail page's node is `<url>#wallet`. So one wallet becomes three unlinked entities: an anonymous node on the index, another anonymous node on each persona page it matches, and an identified node on its detail page.

Setting `"@id": ${detailUrl}#wallet` on the list items would collapse them into one entity referenced from several pages, which is the `@graph` pattern the rest of the site uses (see `reports/page-jsonld.tsx`, which does this well). Since the detail URL is already computed for the persona list, this is nearly free there.

**2. `applicationCategory` should be a schema.org enum value, in one shared helper**

All three files hardcode `applicationCategory: "Cryptocurrency Wallet"`. Google's spec enumerates 25 supported values and states the value "must be a supported app type", so free text isn't accepted.

`apps/[application]/page-jsonld.tsx` already solved this with `APPLICATION_CATEGORY_MAP` (from #17863) — `FinanceApplication` or `UtilitiesApplication` would be the wallet equivalent. Worth a single shared helper rather than three literals, since this PR moves the string from one file to three.

While you're there: `offers: { price: "0" }` can't produce a rich result on its own. Google requires `name` + `offers.price` + **either `aggregateRating` or `review`**, and there's no rating here. I'm **not** suggesting adding one — ethereum.org doesn't rate wallets and inventing a rating would be worse than no rich result — just flagging that the `offers` block is currently inert, so nobody optimizes it expecting a SERP feature. (Minor drift: the detail page omits `availability` while `/apps/` includes it.)

**3. `SoftwareApplication.url` means two different things**

On the index, `url: wallet.url` (external vendor site). On persona pages, `url:` the ethereum.org detail page. Same entity type in the same catalog with two different URL semantics depending on which page a crawler lands on. Worth picking one — vendor site as `url` with the detail page reachable via the item's `@id` would be my read, but either is fine as long as it's consistent.

---

**One non-blocking modelling question for after this lands:** the PR adds a `hardware` persona, so hardware devices are now an explicit first-class category, but Trezor/Ledger/imKey are still typed `SoftwareApplication`. `Product` is arguably right for physical devices. No rich result is reachable either way, so this is a taxonomy decision rather than a bug — happy to open it separately once the revamp is in.

Nice reduction, by the way: −2,395 lines from deleting `FindWalletProductTable/` is the kind of change that makes the schema easier to reason about too.

pettinarip:
@konopkja thanks! applied all three suggestions.

Agreed on taking the hardware->product change separately.

wackerow (approved):
Small note, a tooltip for the "Fees" is shown in the modal, but not on the dedicated "wallet details" page. Non-blocking, but would consider a follow-up to add that in (not for the /wallets/find-wallet/ page itself, but the /wallets/find-wallet/[wallet] page)

@konopkja Heads up that I don't see this AB test set up for the production Matomo instance—only for the preview deploys (ethereum.it). Assuming we want to add that for the launch of this PR

pettinarip:
@wackerow thanks, just to confirm that we do have the fees in the standalone pages 
https://deploy-preview-18846.ethereum.it/wallets/find-wallet/coin98-super-wallet/
image

I'm going to check the ab test setup for prod before merging.

wackerow:
@pettinarip 

> @wackerow thanks, just to confirm that we do have the fees in the standalone pages 

No, I know that -- I was referring to the tooltip specifically, which gives more information about what the fees are, and their distinction to network fees

PR #18868, opened 2026-07-23
i18n: translation pipeline (all languages)

wackerow:
## Automated Translations

This PR contains translations managed by the intl pipeline.
Each run appends a summary below.

---
### Run: 2026-07-23 01:01:45 UTC
- Languages: ar, bn, cs, de, es, fr, hi, id, it, ja, ko, mr, pl, pt-br, ru, sw, ta, te, tr, uk, ur, vi, zh-tw, zh
- Files: 1638 (1158 MD, 480 JSON)
- Mode: auto
- [View workflow run](https://github.com/ethereum/ethereum-org-website/actions/runs/29962972384)

**8 task(s) failed:**

- `public/content/developers/docs/consensus-mechanisms/pos/index.md` (fr): Gemini returned no content (finishReason=RECITATION). This file/language combination may be triggering content filters.
- `public/content/developers/docs/nodes-and-clients/index.md` (es): Gemini returned no content (finishReason=RECITATION). This file/language combination may be triggering content filters.
- `public/content/developers/docs/scaling/plasma/index.md` (ru): Gemini returned no content (finishReason=PROHIBITED_CONTENT). This file/language combination may be triggering content filters.
- `public/content/developers/docs/scaling/plasma/index.md` (vi): Gemini returned no content (finishReason=RECITATION). This file/language combination may be triggering content filters.
- `public/content/developers/docs/smart-contracts/testing/index.md` (es): Gemini returned no content (finishReason=RECITATION). This file/language combination may be triggering content filters.
- `public/content/whitepaper/index.md` (pt-br): Gemini returned no content (finishReason=RECITATION). This file/language combination may be triggering content filters.
- `public/content/whitepaper/index.md` (zh-tw): Output validation failed after 3 attempts: Suspiciously short: 386 chars vs 1571 English chars
- `public/content/whitepaper/index.md` (zh): Output validation failed after 3 attempts: Suspiciously short: 378 chars vs 1571 English chars

Rerun the failed combinations:

```
gh workflow run "Intl Pipeline" -f target_path="public/content/developers/docs/consensus-mechanisms/pos/index.md" -f target_languages="fr"
gh workflow run "Intl Pipeline" -f target_path="public/content/developers/docs/nodes-and-clients/index.md" -f target_languages="es"
gh workflow run "Intl Pipeline" -f target_path="public/content/developers/docs/scaling/plasma/index.md" -f target_languages="ru"
[...]

PR #18869, opened 2026-07-23
fix(roadmap): make SSF card title translatable

myelinated-wackerow:
## Summary

The `single-slot-finality` `ExpandableCard` used `title= "..."` with a space between `=` and the opening quote. The intl-pipeline's JSX attribute-value extractor doesn't match that form, so the card title (`"Why can't we have SSF today?"`) was never sent for translation and shipped verbatim in English across **all 24 locales** (the card body translated fine). Removing the space lets the next pipeline run translate the title.

Surfaced during the PR #18868 translation review.

Note: the same line's `eventName` reads `"clicked Why can't we hear SSF today?"` ("hear" vs "have") -- a pre-existing analytics-label typo left untouched here, since renaming a Matomo event affects tracking continuity. Flagging for a separate decision.

### DIFF
--- a/public/content/roadmap/single-slot-finality/index.md
+++ b/public/content/roadmap/single-slot-finality/index.md
@@ -32,7 +32,7 @@ With the current mechanism design, in order to reduce the time to finality, it i
 
 ## Routes to SSF {#routes-to-ssf}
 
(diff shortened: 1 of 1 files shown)


myelinated-wackerow:
Folding into #18868 -- the English `title=` fix and the manual per-locale title translations all belong in the reviewed translation PR. -- Opus 4.8

PR #18874, opened 2026-07-23
Fix: Card hover-lift causes text to blur (#18364)

### LINKED ISSUE #18364: Card hover-lift causes text to blur then re-sharpen (sub-pixel scale transform)

shriyashsoni:
Fixes #18364. Replaced \scale-*\ with \-translate-y-1\ in \hover-lift\ utilities to prevent sub-pixel rendering blur.

### DIFF
--- a/src/styles/utilities.css
+++ b/src/styles/utilities.css
@@ -161,19 +161,19 @@
 }
 
 @utility hover-lift-xs {
-  @apply scale-100 shadow-none duration-300 hover:scale-[1.005] hover:shadow hover:duration-300;
+  @apply shadow-none duration-300 hover:-translate-y-1 hover:shadow hover:duration-300;
 }
 
 @utility hover-lift-base {
-  @apply scale-100 shadow-none duration-300 hover:scale-101 hover:shadow-md hover:duration-300;
+  @apply shadow-none duration-300 hover:-translate-y-1 hover:shadow-md hover:duration-300;
 }
 
 @utility hover-lift-sm {
-  @apply scale-100 shadow-sm duration-300 hover:scale-101 hover:shadow-lg hover:duration-300;
+  @apply shadow-sm duration-300 hover:-translate-y-1 hover:shadow-lg hover:duration-300;
 }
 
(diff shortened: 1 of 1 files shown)


wackerow:
Replica of #18365

PR #18880, opened 2026-07-23
update hegotá release year to 2027

nixorokish:
hegotá will not ship in 2026, 2027 is most likely


wackerow (approved):
☹️ Bummer, but also.. Soon™ memories 😆 Thanks @nixorokish!

PR #18884, opened 2026-07-23
chore(i18n): remove 56 dead keys from common namespace

pettinarip:
Removes 56 dead keys (~4.2KB) from the `common` namespace — all zero-reference (removed nav items, old translation-banner copy, `kraken-logo`). `common` ships on every page, so these are dead weight everywhere.

Excludes `language-*`/`region-*` keys: they look unused to a literal grep but are live via dynamically-composed keys (language picker, exchange regions).

English source only; the intl pipeline propagates deletions.


wackerow (approved):
Love it, thanks @pettinarip!

PR #18886, opened 2026-07-23
ui/card: opt-in hover affordances, CardLinkFake

myelinated-wackerow:
## Summary

Refines the `ui/card` interaction model — how a card signals that it (or something inside it) is actionable — and rolls the new conventions out across all consumers.

### Primitive (`ui/card`, new `ui/arrow`, `utilities.css`)

- **New `ui/arrow` primitive** (`ArrowNext` / `ArrowPrev`): RTL-aware Lucide arrow wrappers that bake in the `rtl:-scale-x-100` flip. `ui/card` and `ui/Link` now use it, so the flip logic lives in one place.
- **`hoverLift` is automatic on any `href` card** (a link *is* an action) — you no longer pass it by hand on link cards. It stays a manual opt-in for non-link cards that carry an action (multiple buttons, or an inline link).
- **The card title no longer underlines on hover** — the outline/fill + lift (+ cursor) already signal "clickable"; the underline was redundant.
- **New `CardLinkFake`** — presentational text-link mirror for the single CTA of an `href` card (sibling of `CardButtonFake`), with `withForwardArrow` / `hideArrow` and an automatic external-link arrow.
- **`CardBanner` `zoom` is now opt-in** (off by default, no `zoom={false}`). With the card carrying its own hover affordances, the image zoom stacked into a "double scale"; it's now applied explicitly only where the media tile should react.
- **`hover-lift-*` reverted from translate back to a +1% scale over 250ms** — keeps the lift subtle and fixes the mid-transition text blur without the translate jump.

### Consumer sweep

- Single-CTA cards now hoist the destination onto the `Card` `href` and render the CTA as `CardButtonFake` / `CardLinkFake` instead of nesting a real `ButtonLink` / `Button` / `LinkWithArrow` inside the card's anchor: `BuilderCard`, `SpeedRunCard`, `LearnCard`, `ReportCard`, `UseCaseCard`, developers "Resources" cards, roadmap, community/events, run-a-node, `BuildYourOwnAIAgent`, get-eth, and others.
- Removed redundant `hoverLift` from every `href` card (now automatic).
- Applied explicit `zoom` to the media/thumbnail tiles that should keep the image zoom.
- Genuinely multi-link / multi-CTA cards were intentionally left as real `ButtonLink`s (they can't collapse to one card-level link).

### Docs & stories

Design-system skill (`card-walkthrough`, `components`, `gotchas`, `i18n-rtl`) and the Card Storybook stories updated to the new conventions (auto-`hoverLift`, `CardLinkFake`, `ui/arrow`, opt-in `zoom`, `CardFooter buttons="responsive"` default).

## Test plan

- [ ] Storybook `UI / Primitives / Card`: `InteractionPatterns` (single-CTA button + link, multi-CTA, inline-link), `LinkHoverByVariant`, and the Linkable zoom opt-in cell
- [ ] Hover a link card: single scale (lift only), no title underline; `ghost` fills, others show the ring
- [ ] Spot-check converted pages: /developers, /learn, /use-cases, /get-eth, /run-a-node, /roadmap, /community/events, /reports/trillion-dollar-security, /stablecoins
- [ ] RTL (Arabic): forward/back arrows flip
- [ ] Light + dark mode


src/components/ui/card.tsx
```diff
@@ -96,7 +97,8 @@ const Card = React.forwardRef<HTMLDivElement | HTMLAnchorElement, CardProps>(
       cardVariants({
         variant,
         size,
-        hoverLift,
+        // Link cards (href) always lift; non-link cards opt in via the prop.
+        hoverLift: !!href || hoverLift,
```
pettinarip: ` <Card href hoverLift={false}>` doesn't disable the hover effect

```suggestion
        hoverLift: hoverLift ?? !!href,
```

src/styles/utilities.css
pettinarip, line 1: nice to have: would be interesting to add a `motion-safe` behind the scaling for those users that pref reduced motion

pettinarip (approved):
@myelinated-wackerow Looks great! left a minor finding, but the rest looks clean

src/components/ui/card.tsx
wackerow: Nice, good call

PR #18890, opened 2026-07-23
fix(i18n): propagate template key to locales

myelinated-wackerow:
## Summary

Commit 26bdbe8e69 added a `template:` frontmatter key (`roadmap`/`upgrade`) to 12 English roadmap pages, but the intl-pipeline's deterministic propagation only handles frontmatter **value changes**, not **key additions** — so no locale ever received it, and all 24 locale versions of those 12 pages render the old Static layout instead of TopicLayout (`frontmatter.template || getLayoutFromSlug(slug)` falls back to `"static"` for roadmap slugs). This inserts the inert key after the `lang:` line in all 288 locale files, mirroring English key order. Manifests catch up via a scoped `stamp_only` pipeline run against this branch.


pettinarip (approved):
LGTM

PR #18284, opened 2026-05-26
feat: port Matomo AI tracker to Netlify Edge Functions

kafeelraza:
## Description

Ports the upstream Matomo Cloudflare Worker implementation to a dormant Netlify Edge Function for ethereum.org.

Adds the initial TypeScript implementation under `netlify/edge-functions/matomo-ai-tracker` along with local development instructions and curl-based verification examples in the README.

This PR intentionally does **not** wire the function into `netlify.toml` yet, following the staged rollout plan discussed in the issue.

## Related Issue

Closes #18219

Issue #18219, opened 2026-05-19
Capture user-driven AI agent traffic in Matomo via Netlify Edge Function


konopkja:
## Summary

ethereum.org's Matomo cannot currently see traffic from user-driven AI agents fetching pages to answer users (`ChatGPT-User`, `Claude-User`, `Perplexity-User`, etc.) because Matomo's JS tracker requires JavaScript execution, and these agents don't run JS.

In March 2026, Matomo 5.8.0 introduced a first-party "AI Chatbots" report fed by server-side ingestion. Matomo ships reference implementations for Cloudflare Workers ([matomo-org/tracker-cloudflare](https://github.com/matomo-org/tracker-cloudflare)) and AWS CloudFront, but not for Netlify, where ethereum.org runs.

This issue proposes porting the Cloudflare Worker to a Netlify Edge Function. **Phase 1 (this PR) ships chatbot UAs to populate the AI Chatbots report. Phase 2 (immediate follow-up) extends the same Edge Function's UA allowlist to also cover crawlers, giving us data for the whole AI stack.**

## The four AI signals on ethereum.org

Matomo distinguishes four distinct AI traffic signals. Three are live today; this PR covers the fourth, and a follow-up extends coverage to crawlers.

| Signal | What it tells us | Status (site 4, last 30d) | Where it lands in Matomo |
|---|---|---|---|
| Humans clicking through from AI answers | How often AI tools refer real users to us | **Live: 3,191 visits** (1.89% of total) | `Acquisition > AI Assistants` |
| AI agents acting on the site solo | How often autonomous agents (`NovaAct`, `GoogleAgent-Mariner`, etc.) visit | **Live: 253 visits** | `AI Assistants > AI Agents` |
| AI chatbots fetching to answer | How often AI fetches our content to generate a response | **Phase 1 (this PR)** | `AI Assistants > AI Chatbots` |
| Crawlers ingesting for training/RAG | How often LLMs index us for training (`GPTBot`, `ClaudeBot`, `Google-Extended`, `OAI-SearchBot`, etc.) | **Phase 2 (follow-up)** | Bot reports (with "Track Bots" enabled) or a custom dimension |

## Phasing rationale

Phase 1 and Phase 2 use the **same Edge Function**, differing only in the UA allowlist regex. They're sequenced rather than combined because:

[...]

PR #18893, opened 2026-07-24
[hot-fix] feat: enable Matomo AI tracker site-wide (flag-gated)

myelinated-wackerow:
## Summary

PR 2 of the staged rollout from #18219, on top of the merged #18284. Wires the Matomo AI tracker edge function into `netlify.toml` for all page routes (`/*` with an `excludedPath` list so asset requests never invoke the function), gated per deploy context by a `MATOMO_AI_TRACKER_ENABLED` kill switch (default off, silent pass-through). Matomo URL/site ID fall back to the existing context-scoped `NEXT_PUBLIC_MATOMO_*` vars, so non-production deploys report to the non-production Matomo site; `onError: bypass` ensures a function failure never breaks the page. Kill switch and env fallback covered in the unit suite.

Verified end-to-end on this PR's deploy preview: requests with allowlisted chatbot UAs landed in the non-production Matomo site's AI Assistants > AI Chatbots report (server-side Measurement Protocol hits, no JS). Production stays inert until its `MATOMO_AI_TRACKER_ENABLED` is flipped to `true` -- planned after the currently-launching A/B test completes.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -47,11 +47,33 @@
   included_files = ["i18n.config.json", "src/intl/**/*", "src/data-layer/mocks/**/*"]
 
 # Matomo AI tracker (see netlify/edge-functions/matomo-ai-tracker/README.md).
-# Gated by MATOMO_AI_TRACKER_ENABLED; canary scope while the Measurement
-# Protocol pipeline is verified, broaden after clean production data.
+# Gated per deploy context by MATOMO_AI_TRACKER_ENABLED. excludedPath only
+# trims edge invocation cost on asset routes; the function's own
+# URL_EXCLUDE_REGEX is the tracking-side backstop for anything that slips
+# through.
 [[edge_functions]]
   function = "matomo-ai-tracker"
-  path = "/learn/*"
+  path = "/*"
+  excludedPath = [
+    "/_next/*",
+    "/api/*",
+    "/.netlify/*",
+    "/.well-known/*",
+    "/images/*",
+    "/content/*",
+    "/videos/*",
+    "/audio/*",
+    "/code-examples/*",
+    "/reports/*",
+    "/*.css",
+    "/*.js",
+    "/*.map",
+    "/*.json",
+    "/*.ico",
+    "/*.txt",
+    "/*.xml",
+    "/*.webmanifest"
+  ]
 
 # Override SITE_URL for named branches with custom subdomains so canonical
 # URLs and OG metadata use the custom domain instead of *.netlify.app.
--- a/netlify.toml
+++ b/netlify.toml
@@ -46,6 +46,13 @@
 [functions]
   included_files = ["i18n.config.json", "src/intl/**/*", "src/data-layer/mocks/**/*"]
 
+# Matomo AI tracker (see netlify/edge-functions/matomo-ai-tracker/README.md).
+# Gated by MATOMO_AI_TRACKER_ENABLED; canary scope while the Measurement
+# Protocol pipeline is verified, broaden after clean production data.
+[[edge_functions]]
+  function = "matomo-ai-tracker"
+  path = "/learn/*"
+
 # Override SITE_URL for named branches with custom subdomains so canonical
 # URLs and OG metadata use the custom domain instead of *.netlify.app.
 [context.dev.environment]
--- a/netlify.toml
+++ b/netlify.toml
@@ -65,12 +65,12 @@
     "/audio/*",
     "/code-examples/*",
     "/reports/*",
+    "/robots.txt",
     "/*.css",
     "/*.js",
     "/*.map",
     "/*.json",
     "/*.ico",
-    "/*.txt",
     "/*.xml",
     "/*.webmanifest"
   ]
--- a/netlify/edge-functions/matomo-ai-tracker/index.ts
+++ b/netlify/edge-functions/matomo-ai-tracker/index.ts
@@ -10,6 +10,10 @@ declare const Netlify: {
   }
 }
 
+// If the function itself errors at the platform level, skip it and continue
+// to origin rather than failing the user request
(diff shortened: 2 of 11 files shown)


wackerow (dismissed):

netlify/edge-functions/matomo-ai-tracker/index.ts
pettinarip, line 88: cool

netlify/edge-functions/matomo-ai-tracker/matomo.ts
```diff
@@ -0,0 +1,65 @@
...
+  const method = request.method.toUpperCase()
+  if (
+    Array.isArray(config.httpMethodAllowlist) &&
+    config.httpMethodAllowlist.length > 0 &&
+    !config.httpMethodAllowlist.includes(method)
+  ) {
+    return null
+  }
+
+  const url = request.url
```
pettinarip: The URL we record here is the **post-middleware rewritten** URL, not the one the bot actually requested.

Per Netlify's documented execution order, framework-generated edge functions declared in a configuration file run in tier 1, ahead of every user-declared function — and `@netlify/plugin-nextjs` declares its middleware edge function via a generated `manifest.json`. It continues the chain with `context.next(newRequest)`, where `newRequest` carries the rewrite target. So by the time we read `request.url`, it's already rewritten. Two concrete consequences:

- `localePrefix: "as-needed"` rewrites `/learn` → `/en/learn`, so AI-chatbot pageviews land on `/en/learn` while client-side pageviews land on `/learn` — same page, two rows in the report.
- `proxy.ts` rewrites A/B routes to `/en/ab-code/<signed-code>/<path>`. Those internal variant URLs would go straight into the AI Chatbots report, fragmenting each tested page into N rows. We have an A/B test launching, so this isn't hypothetical.

Worth noting this **isn't fixable by reordering** — no declaration placement puts a user edge function ahead of the framework's tier-1 one. So the fix has to happen here:

- Normalize before sending to Matomo: strip the `/en` prefix that `as-needed` adds, and map `/en/ab-code/<code>/<path>` back to `/<path>`. Deterministic and fully under our control.
- Or, if Netlify exposes the original pre-rewrite URL on the request at this point, read that instead — I haven't verified that it does, so it'd need checking on the deploy preview first.

netlify/edge-functions/matomo-ai-tracker/index.ts
```diff
@@ -0,0 +1,108 @@
...
+
+declare const Netlify: {
+  env: {
+    toObject: () => Record<string, string | undefined>
+  }
+}
+
+// If the function itself errors at the platform level, skip it and continue
+// to origin rather than failing the user request
+export const config = { onError: "bypass" }
```
pettinarip, line 15: nit: consider dropping this and moving it to `on_error = "bypass"` in the `[[edge_functions]]` block in `netlify.toml`, so the function has one declaration site instead of two and sits with the rest of the config.

netlify/edge-functions/matomo-ai-tracker/index.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    return response
+  } catch (err) {
+    const message = err instanceof Error ? err.message : String(err)
+
+    log.error("Origin request failed", {
+      error: message,
+    })
+
+    const fallback = new Response("Bad Gateway", {
```
pettinarip: This hand-rolled 502 defeats the `onError: "bypass"` we just added.

When `context.next()` rejects on a transient origin blip, we swallow the error and return a bare `Bad Gateway` — no `content-type`, no cache headers, no Netlify error page. `onError: "bypass"` only fires on *uncaught* throws, so it can never engage here. Once the flag is on, an analytics wrapper is deciding what users see when the origin hiccups.

Prefer re-throwing and letting the platform bypass.

netlify.toml
```diff
@@ -46,6 +46,35 @@
...
+    "/videos/*",
+    "/audio/*",
+    "/code-examples/*",
+    "/reports/*",
+    "/*.css",
+    "/*.js",
+    "/*.map",
+    "/*.json",
+    "/*.ico",
+    "/*.txt",
```
pettinarip: This excludes `/llms.txt` and `/developers/docs/llms.txt` (both exist as routes) — arguably the single clearest AI-traffic signal on the site, and it's excluded twice over, since the default `URL_EXCLUDE_REGEX` also lists `txt`.

If the concern is `robots.txt` noise, could we exclude that specifically rather than the whole extension?

pettinarip (commented):
Looking good @myelinated-wackerow. Left something to double check in terms of the URL that we record to send to matomo, it might not be the correct one

wackerow:
> nit: consider dropping this and moving it to `on_error = "bypass"` in the `[[edge_functions]]` block in `netlify.toml`, so the function has one declaration site instead of two and sits with the rest of the config.

Yeah, good instinct, but the toml schema doesn't actually support it — `[[edge_functions]]` declarations only accept `function`, `path`/`excludedPath`, `pattern`/`excludedPattern`, `cache`, `header`, and `method` ([edge-bundler `declaration.ts`](https://github.com/netlify/build/blob/main/packages/edge-bundler/node/declaration.ts)), and the docs likewise only document `onError` for in-source config. So the inline export should stay; added a README note explaining this for later reference.

The rest of the issues notes in your review have fixes in (URL normalization, re-throw for platform bypass, `llms.txt` un-excluded)... Each verified live against the non-prod Matomo site from the deploy preview, including a fun extra find: hits carrying the `download` param never surface in the AI Chatbots report, so `txt` came out of `DOCUMENT_REGEX` too and `llms.txt` records as a plain pageview.

-- @wackerow / Fable 5

pettinarip (approved):
LGTM

PR #18904, opened 2026-07-24
refactor(quiz): self-provision standalone quiz i18n messages

### LINKED ISSUE #18903: Standalone quiz widget over-ships i18n messages and requires provider boilerplate on every page
### Problem

Every page that embeds `StandaloneQuizWidget` has to load all locale messages with `getMessages`, `pick` namespaces via `getRequiredNamespacesForPage`, and wrap the widget in `I18nProvider` (see `app/[locale]/gas/page.tsx`, `app/[locale]/privacy/page.tsx`, etc.). This has two costs:

1. **Boilerplate**: 6+ pages repeat the same `getMessages` + `pick` + `I18nProvider` dance just to render one quiz, and `QUIZZES_PREFIXES` in `src/lib/utils/translations.ts` must be kept in sync with every path that hosts a quiz.
2. **Client payload bloat**: the widget only consumes its own quiz's question strings, ~15 shared UI keys from `learn-quizzes`, and a single title key from `common`. But because `pick` works at namespace granularity, each page ships the entire `learn-quizzes` namespace (~80 KB in English -- the question bank for every quiz on the site) plus all of `common` (~26 KB) to the client, and in some cases the page's own namespace too, which is dead weight on fully server-rendered pages.

### Proposed solution

Convert `StandaloneQuizWidget` into a self-provisioning async server component that wraps the client widget in its own scoped `I18nProvider`. Since `quizKey` is known server-side and question ids map 1:1 to message-key prefixes, the wrapper can subset `learn-quizzes` down to the quiz's own question keys plus a fixed UI-key list, and `common` down to the quiz-title key.

- Call sites collapse to a single line: `<StandaloneQuizWidget quizKey="..." />` -- no `getMessages`/`pick`/`I18nProvider` in pages
- Client-shipped messages drop from ~110 KB to a few KB per quiz page
- `QUIZZES_PREFIXES` (and the `learn-quizzes` entry in the `use-cases` layout namespaces) can be removed; `/quizzes/` keeps the full namespace as its primary namespace since the hub page legitimately needs the full question bank for client-side quiz switching

Splitting `learn-quizzes.json` into per-quiz namespace files was considered and rejected: it would require a coordinated migration across the intl pipeline, manifests, and 25 locales for the same wire result that server-side key filtering achieves at zero migration cost.

-- Fable 5

myelinated-wackerow:
## Summary

`StandaloneQuizWidget` is now an async server component that provisions its own scoped client i18n context, so pages render `<StandaloneQuizWidget quizKey="..." />` with no `getMessages`/`pick`/`I18nProvider` boilerplate. Since `quizKey` is known server-side, the wrapper subsets messages to the quiz's own question strings, the ~14 widget UI keys, and the quiz-title key.

**Performance:** client-shipped messages drop from ~106 KB to ~5 KB per quiz page. Every page hosting a standalone quiz previously serialized the entire `learn-quizzes` namespace (~80 KB -- the question bank for every quiz on the site) plus all of `common` (~26 KB) into its RSC payload for a single quiz; the six standalone-quiz pages and the MDX topic pages (use-cases layout) all shed that weight. `QUIZZES_PREFIXES` is removed entirely; `/quizzes/` still receives the full namespace as its primary namespace since the hub legitimately needs every quiz's strings for client-side quiz switching.

The interactive widget moves to `QuizWidget/QuizWidgetClient.tsx` behind the `"use client"` boundary; only server modules import the directory index, so an accidental client-side import of the barrel fails the build loudly via `next-intl/server`.

## Related issue

- Closes #18903


pettinarip (approved):
Niiiice! gj

PR #18917, opened 2026-07-27
chore(chromatic): per-project tokens for local runs; fix pages config clash

pettinarip:
Follow-up to #18898.

**Local publishes were 37% of our snapshot spend.** Over Jul 20–24, 30 of 96 Chromatic builds match no GitHub Actions run — local `pnpm chromatic` runs against the shared project, mostly full rebuilds. The script carried a hardcoded token, so an accidental run cost ~252 billed snapshots.

Each local script now requires its own token and fails immediately without it:

| Script | Token |
| --- | --- |
| `pnpm chromatic` | `CHROMATIC_STORYBOOK_TOKEN` |
| `pnpm chromatic:pages` | `CHROMATIC_PAGES_TOKEN` |

Per-project rather than one shared variable because Chromatic falls back to `CHROMATIC_PROJECT_TOKEN` — which would let a pages token publish a Storybook build into the wrong project.

**Also fixes a latent break from #18898.** `chromatic.config.json` is auto-loaded by *every* Chromatic invocation, and its `untraced` option is a hard error without `onlyChanged` — so the pages job would have failed on its next run (it's release-gated, so it hasn't run since). Renamed to `chromatic.storybook.json`, loaded explicitly by the Storybook job only.

CI is otherwise untouched: both jobs pass tokens to `chromaui/action` directly, so **no new GitHub secrets are needed**.

**Migration:** if you have `CHROMATIC_PROJECT_TOKEN` exported for local pages runs, rename it to `CHROMATIC_PAGES_TOKEN`. Tokens must be exported in your shell — `.env.local` won't work.

This gets us from ~2.2× to ~1.4× over the monthly cap. The rest is the 64% TurboSnap bail rate, tracked separately.


package.json
```diff
@@ -17,8 +17,8 @@
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
     "build-storybook:chromatic": "storybook build --test",
-    "chromatic": "chromatic --project-token fee8e66c9916",
-    "chromatic:pages": "chromatic --playwright --zip --exit-zero-on-changes",
+    "chromatic": "chromatic --project-token=${CHROMATIC_STORYBOOK_TOKEN:?}",
+    "chromatic:pages": "chromatic --playwright --zip --exit-zero-on-changes --project-token=${CHROMATIC_PAGES_TOKEN:?}",
```
wackerow, line 21: Guard appears to be convention-only -- this Storybook token remains hardcoded at ci.yml:93 and in git history. If I'm correct, anyone can still run chromatic with this project token and bill the quota. Bit of asymmetry too where pages token is a secret, Storybook token isn't. Consider rotating it into secrets.CHROMATIC_STORYBOOK_TOKEN

(follow-up, not a merge blocker)

wackerow (approved):
@pettinarip Lgtm, no blockers, but left a note below

package.json
pettinarip: yeah, I want to see how this goes and rotate if necessary

PR #18920, opened 2026-07-27
fix(apps): rehost app screenshots to S3

myelinated-wackerow:
## Summary

App screenshots were never rehosted to S3. `uploadAppImages` only covered the logo and banner, so screenshot URLs passed through from the sheet verbatim and any host outside `next.config.js` `remotePatterns` 400s in `next/image` -- which is what broke the [Session](https://ethereum.org/apps/session/) screenshot. It went unnoticed because every other app's screenshots were hand-uploaded to the bucket, with the S3 URL pasted back into the sheet.

Screenshots now go through `uploadManyToS3` under an `apps/screenshots` prefix. Two supporting changes in `s3.ts`: `uploadToS3` passes through URLs already on the `S3_ENDPOINT` host, so the existing hand-uploaded cells aren't refetched from our own bucket and re-keyed under a content hash; and `uploadManyToS3` chunks instead of fanning out unbounded, since every upload buffers the full image before the 5MB size check.

`fetch-apps` is a daily task, so this takes effect on the next run after deploy.

## Related issue
Fixes:
image
(https://ethereum.org/apps/session/)

### DIFF
--- a/src/data-layer/s3.ts
+++ b/src/data-layer/s3.ts
@@ -127,6 +127,21 @@ function buildS3Url(bucket: string, key: string): string {
 }
 
 /**
+ * True when the URL already points at our own S3 host. Sources are often
+ * hand-uploaded to the bucket and pasted back into a sheet, so re-uploading
+ * would re-key the object under a content hash and orphan the original.
+ */
+function isAlreadyOnS3(url: string): boolean {
+  const endpoint = process.env.S3_ENDPOINT
+  if (!endpoint) return false
+  try {
+    return new URL(url).hostname === new URL(endpoint).hostname
+  } catch {
+    return false
+  }
+}
+
+/**
  * Upload an image from an external URL to S3.
  *
  * @param sourceUrl - The external image URL to fetch and upload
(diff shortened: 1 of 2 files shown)


src/data-layer/s3.ts
```diff
@@ -126,6 +126,21 @@ function buildS3Url(bucket: string, key: string): string {
   return `${process.env.S3_ENDPOINT}/${bucket}/${key}`
 }
 
+/**
+ * True when the URL already points at our own S3 host. Sources are often
+ * hand-uploaded to the bucket and pasted back into a sheet, so re-uploading
+ * would re-key the object under a content hash and orphan the original.
+ */
+function isAlreadyOnS3(url: string): boolean {
```
pettinarip, line 134: is this ever going to happen for external images? we also checking later in the logic if the image already exists in S3

pettinarip (approved):

src/data-layer/s3.ts
wackerow: The HEAD check later dedupes by a hash of the source URL, so it only catches re-runs of the same external URL. It can't catch S3 URLs pasted into the sheet (the current convention for most screenshots, e.g. `apps/aave/screenshot-1.png`): hashing that URL produces a key under `apps/screenshots/` that doesn't exist, so it misses and we'd refetch our own object and duplicate it under a hash key.
The fetch also happens before the HEAD check, so even on a HEAD hit we'd re-download ~180 of our own screenshots every daily run. The guard short-circuits both and it never matches genuinely external images (only the `S3_ENDPOINT` host)

Issue #18792, opened 2026-07-14
[recovery] Error: Invalid locale: static — uncaught 500 from /static catch-all in [locale] segment


pettinarip:
> Filed by the automated recovery agent from a production Netlify error captured via Grafana → Keep.

## Error context

- **Message:** `[ERROR] ⨯ Error: Invalid locale: static`
- **Function:** `___netlify-server-handler` (SSR)
- **Digest:** `2164734634`
- **Fingerprint:** `grafana-fn-error-invalid-locale-static-at-j-next-server-chunks-ssr-root-of-the-s`
- **Hit count:** 24 (first seen 2026-05-26)
- **URL / resource:** not captured in the alert

Minified stack:
```
at j (.next/server/chunks/ssr/[root-of-the-server]__0537cz0._.js:2:185878)
at <unknown> (.next/server/chunks/ssr/[root-of-the-server]__0nxmwjg._.js:4:15972)
at ad (.next/server/chunks/ssr/_0vy~fo1._.js:2:37449)
```

## Root-cause analysis

`static` is being interpreted as a `[locale]` path segment. With `localePrefix: "as-needed"` and no `middleware.ts`, the `app/[locale]/...` segment acts as a catch-all for unknown top-level paths (documented in next-intl's `getRequestConfig` types: *"the `[locale]` segment effectively acts like a catch-all for unknown routes"*). A crawler/scanner hitting `/static/...` (a very common bot probe path for static-asset dirs) therefore lands with `params.locale === "static"`.

`app/[locale]/layout.tsx` guards this with `if (!routing.locales.includes(locale)) notFound()`, **but** `generateMetadata` on the page runs *concurrently* with the layout render (same concurrency issue called out in #18013 / PR #18733), so a page-level metadata path can touch the raw `"static"` locale before the layout's `notFound()` fires — surfacing as an uncaught `Error` → 500 instead of a clean 404.

[...]

PR #18932, opened 2026-07-28
[recovery] fix(bug-bounty): set request locale before next-intl APIs (static→dynamic bailout on /.well-known/bug-bounty)

pettinarip:
> Opened by the automated recovery agent from a production Netlify error captured via Grafana → Keep.

## Production error

- **Message:**

  ```
  [ERROR] ⨯ Error: Page changed from static to dynamic at runtime /.well-known/bug-bounty, reason: headers
      at y (.next/server/chunks/ssr/0cdz_next_dist_0dhscug._.js:2:12452)
  ```

- **Function:** `___netlify-server-handler` (SSR)
- **URL / resource:** `/.well-known/bug-bounty`
- **Occurrences:** 1 in this alert window, `hit_count: 4` cumulative; last seen `2026-07-28T15:43:02.421Z`
- **Request ID:** `01KYMP9BWX1A3DNT23QXVDV1C4`

## Root cause

`/.well-known/bug-bounty` is matched by `app/[locale]/bug-bounty/page.tsx` with `params.locale === ".well-known"`. The proxy matcher excludes `.well-known` (`proxy.ts:128`), so next-intl's routing never normalizes the path, and the invalid locale segment reaches the route directly — the same `[locale]`-acts-as-a-catch-all class of bug as #18013 / #18792.

`app/[locale]/bug-bounty/page.tsx` was the only page that called next-intl's *implicit* server APIs without first pinning the request locale:

```ts
const { locale } = params

const t = await getTranslations("page-bug-bounty")   // implicit — no locale
const tCommon = await getTranslations("common")
```

With no `setRequestLocale(locale)`, next-intl falls back to `getRequestLocale()`, which reads the `X-NEXT-INTL-LOCALE` request header via `next/headers` (`next-intl/dist/.../server/react-server/RequestLocale.js`). Calling `headers()` inside a prerendered route is exactly what trips Next.js' static→dynamic bailout, producing `reason: headers`. next-intl is the only dependency in the render path that imports `next/headers`, and there are no direct `next/headers` calls anywhere in the repo — so this is the only possible source of that reason string.

Normally `app/[locale]/layout.tsx:42` populates the request-locale cache before the page renders, which is why valid locales don't log this. For `.well-known` the layout hits `notFound()` on line 39 *before* `setRequestLocale`, leaving the cache empty; page-level code runs concurrently with the layout guard (documented in #18792 and relied on by #18733), so the page's `getTranslations` calls hit `headers()` and the prerendered route bails to dynamic.

The page's own `generateMetadata` already does this correctly (`setRequestLocale(locale)` at line 838) — only the page component was missing it. 49 of the 58 pages under `app/[locale]/**` already call `setRequestLocale(locale)` in the page body, and `app/[locale]/[...slug]/page.tsx:40` carries the comment this fix reuses verbatim.

## Fix

One line in `app/[locale]/bug-bounty/page.tsx`, matching the existing repo pattern:

```ts
const { locale } = params

// Set locale before next-intl APIs so on-demand renders stay static
setRequestLocale(locale)

const t = await getTranslations("page-bug-bounty")
```

`setRequestLocale` performs no validation, so `.well-known` is simply cached as the request locale; `src/i18n/request.ts` already falls back to `routing.defaultLocale` for any value not in `routing.locales`. Behaviour for real locales is unchanged, and the 404 render no longer touches request headers.

No refactoring or drive-by cleanup — `setRequestLocale` was already imported.

## Verification

```
$ pnpm type-check
> next typegen && tsc --noEmit && tsc --noEmit -p netlify/edge-functions
Generating route types...
✓ Types generated successfully
```

```
$ pnpm exec eslint 'app/[locale]/bug-bounty/page.tsx'
(no output — clean)
```

## Follow-up (not in this PR)

Eight other pages under `app/[locale]/**` still call implicit next-intl APIs in the page body without `setRequestLocale`, so they can bail to dynamic the same way when hit with a non-locale first segment:

- `app/[locale]/learn/page.tsx`
- `app/[locale]/use-cases/page.tsx`
- `app/[locale]/what-is-ethereum/page.tsx`
- `app/[locale]/what-is-the-ethereum-network/page.tsx`
- `app/[locale]/founders/page.tsx`
- `app/[locale]/community/events/page.tsx`
- `app/[locale]/community/events/meetups/page.tsx`
- `app/[locale]/community/events/conferences/page.tsx`

Kept out of scope to keep this PR to the one alerting route. The systemic fix suggested in #18792 (reject unknown top-level segments once, before any page or metadata code runs) would cover all of them plus #18013 and #17967.

### DIFF
--- a/app/[locale]/bug-bounty/page.tsx
+++ b/app/[locale]/bug-bounty/page.tsx
@@ -77,6 +77,9 @@ export default async function Page(props: { params: Promise<Params> }) {
   const params = await props.params
   const { locale } = params
 
+  // Set locale before next-intl APIs so on-demand renders stay static
+  setRequestLocale(locale)
+
   const t = await getTranslations("page-bug-bounty")
   const tCommon = await getTranslations("common")
 


pettinarip:

Blondie0123 (commented):
Commit

PR #18933, opened 2026-07-28
fix(apps): rebuild gallery thumbnails as buttons

myelinated-wackerow:
## Summary

The Gallery thumbnails on `/apps/<application>` had their `onClick` on a bare ``, so the lightbox was unreachable by keyboard and the thumbnail carried no role. Each thumbnail is now a real `<button>` (accessible name from the image alt) with `rounded-base` corners, a 1px resting border, and a 2px `primary-hover` outline on hover and focus. The old hover scale is gone -- it was being clipped, because Swiper's `overflow: hidden` sits flush to the slides, so the Swiper root now takes a small padding cancelled by a negative margin to move the clip boundary clear.

Each thumbnail is capped at 16:9 via a max width derived from the tile height, so a wide banner can't hog the carousel, with a blurred low-res copy of the image behind the contained one to fill the letterbox.

Two loading glitches fixed along the way: the slides touched on first paint because Swiper only applies `spaceBetween` as an inline margin once initialized, and the pre-load box was far narrower than the image it resolved into because the declared dimensions were a 1:2 placeholder.

Hard-coded English alt text and the modal title are now in the `page-apps` namespace.

### DIFF
--- a/app/[locale]/apps/[application]/_components/ScreenshotSwiper.tsx
+++ b/app/[locale]/apps/[application]/_components/ScreenshotSwiper.tsx
@@ -1,6 +1,7 @@
 "use client"
 
 import { useState } from "react"
+import { useTranslations } from "next-intl"
 
 import { Image } from "@/components/Image"
 import Modal from "@/components/ui/dialog-modal"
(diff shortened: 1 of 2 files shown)


pettinarip (approved):
Nice, gj

PR #18939, opened 2026-07-29
feat(ab-testing): A/B the rebuilt find-wallet catalog vs the legacy table

pettinarip:
## Find Wallet revamp — A/B test

Sets up the **`FindWalletCatalog2026`** experiment (Matomo experiment 15, currently a draft) so the revamp in #18846 can be validated causally rather than before/after. The catalog is the shipped design; the old expandable table returns as the **Original** arm only for the duration of the test.

> **Stacked on #18846.** Targets `feat/find-wallet-catalog`; will retarget to `dev` once that merges. Split out so the experiment can be reviewed and exercised on its own preview.

## How it's wired

- Flag `FindWalletCatalog2026` registered for `/wallets/find-wallet/`; index 0 = `Original` (legacy table), index 1 = `NewCatalog`.
- Coded page at **`find-wallet/ab-code/[code]/`** — deliberately *inside* the route, not hoisted to the app root. A root-level coded page sits outside `find-wallet/layout.tsx` and therefore loses the `@modal` slot, so the catalog arm would hard-navigate to standalone detail pages instead of intercepting into a modal — i.e. we'd be measuring a degraded version of the design under test. `proxy.ts` appends the coded segment to the matched path to suit.
- `page.tsx` swaps only the page body. Hero and JSON-LD stay shared, so both arms emit identical structured data at the same URL.
- Restores `FindWalletProductTable/`, `wallet-filters.ts` and four `wallets.ts` helpers deleted in #18846, all marked delete-when-the-test-ends. **No translation work needed** — that PR added keys without removing any.

## Three failure modes fixed here

1. **The proxy only rewrites while an experiment is actually running.** Rewriting unconditionally pins every visitor to variant 0. That's only harmless when index 0 is the shipped design — here it would have quietly served the design the revamp replaced, between merge and experiment start, permanently on pause, and would have flipped CI's e2e/visual suites onto the old table. Debug-override cookies still force a rewrite on dev/preview so variants stay previewable pre-launch.
2. **`dynamicParams` on the coded page.** Without it the route serves only the permutations `generateStaticParams` enumerated, so a build that couldn't sign them (no `FLAGS_SECRET` → empty list) would **404 every request** rather than falling back to on-demand rendering as the docs claimed. This was reproducible locally with a clean `.next`.
3. **Legacy JS kept out of the shipped bundle.** The coded page injects the legacy body rather than the real page importing it, keeping ~150KB of old-table client JS off the shipped route (**1730.7 → 1580.5 KB** of referenced JS). Otherwise it would undercut the INP win the revamp exists for.

## Verified on a production build

| Check | Result |
| --- | --- |
| Both arms prerendered, served from cache | `x-nextjs-cache: HIT`, `x-nextjs-prerender: 1` |
| Modal interception in the catalog arm under rewrite | dialog opens, catalog stays behind it, URL `/find-wallet/metamask/` — Chromium + Firefox |
| Modal close | returns to `/wallets/find-wallet/` |
| Public URL | never changes; `ab-code` appears only in the RSC flight payload |
| Canonical / `og:url` | both the public URL |
| Direct coded-URL visit | 301 lowercase → 404 |
| Non-EN (`/es/`) | no rewrite, gets the catalog, unenrolled |
| No active experiment | no `ab-code` rewrite at all |
| Debug panel | lists `Original` / `NewCatalog`, switches arms via override cookie |
| find-wallet e2e | 8/8 |

## Measurement notes for the analytics side

- Primary conversion `WalletExternalLinkList` / `Tap main button` / `<wallet name>` fires **identically in both arms** — Goal 10 measures both correctly, no change needed.
- **Bounce rate is safe as a co-primary.** The redesign turns personas and wallet details into real links, so they emit pageviews where the old design emitted events. I verified against the raw visit log that events count as actions on this Matomo instance and that bounce derives from `actions` (visits with `actions <= 1` matched `bounce_count` exactly, 136/136), so both arms un-bounce equally.
- **Pageview-derived metrics are not comparable across arms** (pages/visit, time on page, index exit rate) — modal open *and* close each fire a pageview, so the catalog arm inflates them for free.
- `percentage_participants` is **ignored** by the adapter — it only reads variation weights and derives Original as `100 − Σ`. Setting participants below 100% in the dashboard would not reduce enrollment.
- Ad-hoc `segment=` API params return all-zero on this instance (no on-demand segment archiving), so any `dimension1==<variation>` segments must be **created as stored segments and archived** before analysis.

## Docs

`docs/ab-testing.md` updated: new rewrite convention and the layout rationale, the running-experiment gate, and a third silent-failure rule — variant element `key`s become Matomo variation names verbatim (hyphens → spaces, title-cased), so `key="new-catalog"` reports `"New Catalog"` and silently matches nothing.


konopkja:
One structured-data question from a wider schema audit (#18977–#18983, #18995, #18996), not a blocker:

During the test, does the legacy arm serve **different JSON-LD at the same canonical URL** as the rebuilt catalog?

#18846 gives the rebuilt catalog a materially different graph from the current page — `CollectionPage` + `ItemList` with per-wallet `SoftwareApplication` items, plus new `@id`s, breadcrumbs and `sameAs` on the detail routes. If both arms render at `/wallets/find-wallet/` behind `ab-code/[code]/`, then Googlebot may see one of two different structured-data shapes for one URL depending on which arm it lands in.

Google generally tolerates A/B tests, so this probably isn't harmful — I mainly want to know whether it's a known and accepted property of the setup, or an accident. If it is accidental, serving the control arm's schema consistently (or keeping the schema identical across arms and varying only the UI) would remove the ambiguity.

Adjacent, and the reason I'm asking rather than assuming: the last find-wallet test's effective date was confounded by riding another deploy, so I'd rather have the measurement conditions written down up front this time. The PR description already notes Matomo events are new series rather than continuations and that the A/B should get its own goals per arm — this is the same category of question.

myelinated-wackerow (approved):
Approving. Three notes, none blocking:

1. **INP was measured on the direct page, not the coded copy real users get.** Both designs sit in that copy's module graph. The browser only fetches what renders, so I'd expect no difference — but nesting the coded page was a decision not to measure a hobbled version of the design under test, and that applies to its own bundle too. Quick re-check on the preview with the override cookie.

2. **"No wallets matched your filters" is counted in one arm only.** cc @konopkja — the restored table fires `Wallet_empty_state`; the catalog fires nothing. Side by side, the old design shows empty-result hits and the new one zero, because nobody's counting. The old number overcounts anyway: it fires during render, so it repeats on every redraw. Either the catalog gets a once-per-empty-state event, or zero-result rate is agreed to be outside the comparison. Better settled before the experiment starts.

3. **Persona cards report their label in the visitor's language.** cc @konopkja — one button spreads across 24 names in the reports. Right for this test, since it mirrors what the old design emitted, but worth switching to a language-independent name at teardown. It isn't on the delete-when-the-test-ends list yet.

Also worth filing: the running-gate has no test. "Nothing running means no rewrite" is the description's scariest failure mode and it's guarded by one manual preview check.

-- Opus 5

pettinarip:
> One structured-data question from a wider schema audit (#18977–#18983, #18995, #18996), not a blocker:
> 
> During the test, does the legacy arm serve **different JSON-LD at the same canonical URL** as the rebuilt catalog?
> 
> #18846 gives the rebuilt catalog a materially different graph from the current page — `CollectionPage` + `ItemList` with per-wallet `SoftwareApplication` items, plus new `@id`s, breadcrumbs and `sameAs` on the detail routes. If both arms render at `/wallets/find-wallet/` behind `ab-code/[code]/`, then Googlebot may see one of two different structured-data shapes for one URL depending on which arm it lands in.
> 
> Google generally tolerates A/B tests, so this probably isn't harmful — I mainly want to know whether it's a known and accepted property of the setup, or an accident. If it is accidental, serving the control arm's schema consistently (or keeping the schema identical across arms and varying only the UI) would remove the ambiguity.
> 
> Adjacent, and the reason I'm asking rather than assuming: the last find-wallet test's effective date was confounded by riding another deploy, so I'd rather have the measurement conditions written down up front this time. The PR description already notes Matomo events are new series rather than continuations and that the A/B should get its own goals per arm — this is the same category of question.

Hmm not an accident. The jsonld sits outside the variant swap. The coded page delegates to the same `page.tsx` and only replaces the body, so both arms emit the identical graph, with public `@id`s and no `ab-code`. So it's your second option: same schema, only the UI varies.

PR #18941, opened 2026-07-29
fix(i18n): make incremental propagation structurally safe

### LINKED ISSUE #18940: intl-pipeline: incremental mode silently corrupts locale files on structural edits
## Summary

The intl-pipeline's incremental (`auto`) mode silently corrupts locale files when an English edit touches structural markup. Across three consecutive runs on PRs #18935 and #18937 it deleted whole sections, dropped section headings, and rewrote link URLs onto the wrong anchors — in up to 24 of 24 locales at a time, with nothing in CI to catch it.

The workaround so far has been `mode=full`, which costs 5-10x per run and discards any human fix in the regenerated file. **`auto` should be the default and `mode=full` the last-resort fallback for problem cases**, which means fixing the propagation layer rather than routing around it.

**The diff/manifest layer is not at fault.** That lives in the external `intl-content-tree` package (`^0.3.2`) and correctly reported every removal, rename and drift in each failure below. Every defect is in `src/scripts/intl-pipeline/pipeline.ts`, which applies the diff to the locale file.

## Observed failures

| Run | English edit | Result |
|---|---|---|
| #18935 run 1 (`auto`) | `translation-program/index.md`: heading replaced, 5-step list + `<ButtonLink>` deleted, new paragraphs added | All 24 locales: anchor renamed `{#help-us-translate}` → `{#program-status}` but the old Crowdin signup list and "Start translating" button **kept underneath it**. Both `discord.gg/ethereum-org` hrefs **overwritten with the new GitHub-issue URL**, anchor text still reading "Discord". In `de`, step 3's links became `/acknowledgements/` and `/contributors/` — the two links from the *new* gratitude paragraph. |
| #18935 run 1 (`auto`) | `contributing/index.md`: one bullet replaced, one trailing sentence removed from a paragraph | 19/24 locales **lost the entire `## How to work on ethereum.org {#how-to-update-content}` section** (heading + 5 paragraphs, 18 lines). 20/24 never received the new bullet. |
| #18937 run 2 (`auto`) | `translators-guide/index.md`: **a single 1:1 sentence replacement** | 17/24 locales **lost the `## Using Crowdin {#using-crowdin}` heading outright** while correctly translating the paragraph beneath it. Blank line before the following heading also eaten. |

Note the third case: a lone sentence swap, no block deletion at all. Any theory that scopes this bug to "deletions only" is too narrow.

### 1. Deleted content is never removed — `pipeline.ts:290-295`

`dr.removed` is consumed in exactly one place, and the only action is a hardcoded `<Divider />` regex:

```js
for (const removed of dr.removed) {
  if (renamedOldIds.has(removed.id)) continue
  if (removed.id.startsWith("component:")) {
    result = result.replace(/\n*<Divider\s*\/>\s*\n*/g, "\n\n")
  }
}
```

Every other removal the differ reports — a section, a paragraph, a list item, a `<ButtonLink>` — is left in the locale file. Deterministic, not model variance.

### 2. The section heading is delegated to the LLM and never validated — `pipeline.ts:78`, `594-600`

`findSection` sets `start` at the heading line:

```js
const lineStart = text.lastIndexOf("\n", match.index) + 1
```

and the writeback replaces the whole range, trusting the model's output to contain a translated heading:

```js
const localeSec = findSection(result, sectionId)
if (localeSec) {
  result = result.slice(0, localeSec.start) + translated + "\n" + result.slice(localeSec.end)
}
```

When the model returns only the prose, the heading disappears. This is why the same edit lost the heading in 17 locales and preserved it in 7.

### 3. Value substitutions are global, not scoped to the changed section — `pipeline.ts:365-372` and siblings

```js
if (change.elementType === "link" && change.key === "href") {
  result = result.replace(
    new RegExp(`\\]\\(${escapeRegex(change.oldValue)}\\)`, "g"),
    `](${change.newValue})`
  )
}
```

A file-wide `/g` replace keyed only on the old URL, so one changed link rewrites **every** occurrence of that URL in the document. The same shape repeats for `html-tag` href, `image` src, and `inline-code`. This is the mechanism behind both `discord.gg` links becoming the GitHub URL.

The *pairing* that mapped old-link-N to new-link-N appears to be positional and lives in the bundled `intl-content-tree` differ — inferred from the observed href mapping, not read from source. Scoping the substitution (below) contains the damage regardless of how the pairing behaves.

### 4. No post-assembly invariant check

`pipeline()` returns whatever the merge produced. Nothing verifies the result still structurally matches English, so all three defects reach a commit silently. Note that `main.ts:553-565` **already** falls back to `runFullTranslation` when the old English blob can't be retrieved — the recovery mechanism exists and simply isn't wired to structural failure.

## Why this shipped

`tests/unit/intl-pipeline/incremental-pipeline.spec.ts` asserts the differ *detects* removals (`dr.removed` has 2 entries) but the only propagation test for a removal is `M3: <Divider /> removed` — the one hardcoded case. There is no test that a removed section disappears from locale output, that a heading survives a section rewrite, or that an href substitution stays within its section.

## Proposed fixes

1. **Scope value substitutions to the changed section's byte range** instead of the whole file (defect 3). Smallest change, removes the worst symptom — silently wrong links.
2. **Preserve the locale's own heading line deterministically** (defect 2): reuse the existing heading, apply anchor renames to it, and prepend it when the model's output doesn't begin with a heading of the expected level. Never rely on the model to reproduce structure.
3. **Implement general removal** (defect 1) for sections, blocks and list items, replacing the `<Divider />` special case. Needs a design call: what to do when the locale counterpart of a removed English block has drifted and can't be matched confidently — safest is to fall back per-file (see 4) rather than guess.
4. **Post-assembly invariants, with automatic fallback** (defect 4). After `pipeline()` returns, assert against English: heading count, `{#anchor}` set, and link/href multiset. On mismatch, log and fall back to `runFullTranslation` for that file. This converts silent corruption into self-repair and is what makes `auto` trustworthy enough to be the default.

## Acceptance criteria

- [ ] Regression test: English deletes a section → locale output no longer contains it.
- [ ] Regression test: English deletes a list item / `<ButtonLink>` inside a section → removed from locale, siblings intact.
- [ ] Regression test: English replaces one sentence in a section → locale keeps its heading, and heading count matches English.
- [ ] Regression test: an href change where the same URL appears twice in the locale → only the occurrence inside the changed section is rewritten.
- [ ] Regression test: post-assembly invariant violation triggers the full-translation fallback rather than committing.
- [ ] The three real-world cases above reproduce as fixtures and pass.
- [ ] `pnpm test:unit` green.

## Non-goals

- Changing `intl-content-tree`. If the pairing logic there proves to be the root of defect 3, file upstream separately; the scoping fix should hold regardless.
- Touching the sanitizer.
- Any change to manifest format or the `intl/pending-*` orchestration model.

## References

- Corruption and recovery evidence: PR #18935 review, PR #18937 review
- Patterns 32-35 in `.claude/translation-review/known-patterns.md` (detection methods, including the anchor-set diff that caught the dropped heading when the link-set diff did not)
- Spec: `tests/specs/PIPELINE-SPEC.md`

myelinated-wackerow:
## Summary

Incremental (`auto`) mode corrupted locale files whenever an English edit touched structural markup — deleted sections, dropped headings, links repointed onto the wrong anchors, up to 24 locales per run. The diff layer reported everything correctly; all four defects were in how `pipeline.ts` applied that diff. This fixes them and adds a post-assembly check that falls a file back to full translation instead of committing a bad merge, so `auto` can be the default and `mode=full` the last resort.

Notably, `PIPELINE-SPEC.md` already specified scoped substitution and section removal — the implementation had drifted from its own spec.

**All four fixes landed:**

1. **Scoped substitutions** — value replaces are confined to the changed section instead of a document-wide `/g` replace keyed on the old value (the mechanism behind both `discord.gg` links becoming a GitHub issue URL).
2. **Deterministic heading line** — the LLM contract is body-only (`buildSectionList` passes the heading as a prompt attribute), so writeback keeps the locale's own heading, uses english-B's for new sections, forces the expected anchor onto a model-echoed heading, and discards an empty response rather than wiping the section. The blank line before the next heading survives too.
3. **General removal** — sections by anchor (stopping at any heading whose anchor english-B still has), components by tag + inert attribute values with a per-identity budget, replacing the hardcoded `<Divider />` regex. Prose and list items are deliberately *not* matched by identifier: they change their section's translatable content, so the LLM retranslates the section, and leftovers are caught by (4). That is the per-file fallback the issue asked for rather than a guess at which translated paragraph to delete.
4. **Post-assembly invariants + fallback** — heading count, `{#anchor}` set and href multiset vs english-B, minus the same comparison for english-A vs locale-A so pre-existing drift isn't counted (~7% of locale files differ on anchors before any run). Any regression this run introduced logs and returns `runFullTranslation` for that file.

Also: a renamed section whose translatable content changed is now retranslated — it lands in neither `added` nor `translatableDrift`, which is why the `{#help-us-translate}` → `{#program-status}` rename left the deleted Crowdin list under the new heading in all 24 locales.

## Acceptance criteria

All seven covered by `tests/unit/intl-pipeline/incremental-structural.spec.ts` (22 cases; the three real-world runs use the actual English diffs and actual `de` files from 968c366354 / 7a72aa50b4 / 2e88449b8d). 16 of the 22 fail against the previous implementation — verified by running the same spec against `origin/dev`'s `pipeline.ts`.

One caveat on the fallback criterion: the test covers `findStructuralRegressions`, the decision the fallback is keyed on, not the call site in `runIncremental`. `main.ts` self-executes on import, so it can't be exercised from a test without extracting that function behind git/LLM/committer mocks.

## Test fixtures regenerated

The E2E mock in `incremental-pipeline.spec.ts` was fence-unaware and truncated locale-B sections at `#` comments inside code fences. The three `locale-expected/*/fixture-1.md` files had frozen that truncation: an unbalanced code fence, and only 7 of 16 headings parseable. With the mock fixed they regenerate with english-B's exact heading structure and balanced fences.

Two residual gaps are tracked in `FUTURE.md` (1b), both from headings never reaching the model: a newly added section keeps english-B's heading line, and a rename that also relabels its heading keeps the old translated label.

Closes #18940


pillowtalk-Qy:
I completed the requested structural-safety review and opened #18952 as a companion PR targeting this branch (I do not have write permission on the upstream branch).

It adds fail-closed preflight for heading/title/move/reorder cases, strict post-assembly and full-translation invariants, the actual fallback call-chain tests, minimal move/duplicate-link fixtures, and a no-write canary over all 24 configured locales. Unique link redirects and safe section deletion remain incremental; hazardous files fall back to full translation before any write.

Local evidence: targeted spec 33/33, full unit suite 996 passed / 1 skipped, ESLint passed, Markdown lint 345 files / 0 errors, and an en/es/zh/ar production build generated 1617 pages. No real translations or dependencies changed.

Recommendation: merge #18952 into this branch before merging #18941.

pillowtalk-Qy:
The companion safety PR #18952 has been updated with aa7985a4a0 to close the link-syntax gap found in follow-up review. It now handles single-quoted HTML href redirects, detects stale/repeated single-quoted hrefs, and preserves localized Markdown link titles while changing only the destination.

Final local evidence: 37 targeted structural tests, 1000 full unit tests with 1 skipped, lint/type/Markdown checks, en/es/zh/ar build with 1617 pages, and a repository-real single-quoted redirect canary passing all 24 locales without writes.

Please merge #18952 into this branch before merging #18941, then rerun CI on the combined SHA.

pettinarip (approved):
LGTM

PR #18952, opened 2026-08-01
fix(i18n): fail closed on structural changes

pillowtalk-Qy:
## Purpose

Complements #18941 with a fail-closed boundary for #18940. Incremental translation is accepted only when the English change fits the body-only contract and the assembled locale preserves the pre-run structure. Unsafe files fall back to full translation before any content or manifest write.

## What changed

- Reject heading additions, relabels, level changes, moves, reordering, unanchored heading changes, and translatable frontmatter changes during preflight.
- Reject ambiguous partial replacements of repeated links while retaining deterministic unique redirects and section deletion.
- Validate anchor multiplicity, heading count/level/parent/order, and href multiplicity after assembly.
- Apply the same strict structural validation to per-file full translation before commit.
- Exercise the actual preflight -> incremental -> invariant -> commit/fallback boundary.
- Add minimal heading-move and duplicate-link fixtures plus an in-memory no-write canary across all 24 configured locales.

No real translation files, dependencies, or lockfiles are changed. No full-language translation run was performed.

## Risk

This is a high-impact path because one English edit can affect 24 locales. The intentional behavior change is that structural and translatable-frontmatter edits now use per-file full translation, increasing LLM cost and latency for those files in exchange for preventing silent corruption. Ordinary body edits, unique link redirects, and safe section deletions remain incremental.

## Test evidence

- Target structural spec: 37 passed.
- Full unit suite: 1000 passed, 1 skipped.
- ESLint: passed.
- Markdown lint: 345 English files, 0 errors.
- Prettier check for changed code/docs/fixtures: passed.
- Production build with USE_MOCK_DATA=true and en, es, zh, ar: passed; 1617 static pages generated.
- 24-locale redirect/deletion canary: passed in memory with no writes; a real single-quoted HTML href redirect also passed across all 24 locales.
- git diff --check: passed.

The production build still emits the same 9 existing Codeblock children is not valid warnings; this PR does not mix in the separate release-reliability task.

## Rollback

Revert the #18952 merge commit, or revert `e81d158f23`, `aa7985a4a0`, and `66f082aff4` in reverse chronological order. If any structural uncertainty remains after rollback, keep risky incremental translation disabled and use targeted per-file full translation; do not overwrite all locales.

## Unresolved

- Existing historical locale drift is baseline-exempt and requires targeted repair; this PR prevents new drift.
- Deep nested JSON structural validation remains future work outside #18940.
- A live Gemini translation was not run because it would require credentials and external translation writes; the orchestration and no-write boundaries are covered by unit tests.
- The 9 Codeblock build warnings belong to the next release-reliability task.

Parent PR: #18941
Issue: #18940


pillowtalk-Qy:
Follow-up review fix pushed in aa7985a4a0.

The structural safety net now reads hrefs from the parsed content tree instead of double-quote-only regexes. Deterministic propagation supports both HTML quote styles and replaces only a Markdown destination, preserving any localized optional title. This closes the reviewed silent-stale-link cases:

- `<a href='/old/'>` now becomes `<a href='/new/'>`
- a stale single-quoted href is rejected by post-assembly invariants
- partial changes to repeated single-quoted hrefs fail closed
- `[label](old "localized title")` redirects the destination without overwriting the locale title

Evidence on the final branch:
- new regression cases: 4 failed before the implementation and pass after it
- targeted structural spec: 37 passed, including the 24-locale no-write canary
- full unit suite: 1000 passed, 1 skipped
- type-check, ESLint, Prettier, git diff check: passed
- Markdown lint: 345 English files, 0 errors
- en/es/zh/ar production build: passed, 1617 static pages
- repository-real single-quoted href redirect: 24/24 locales passed in memory with no writes

The build still emits the same 9 Codeblock warnings tracked by the separate release-reliability task. This companion PR still needs to be merged into #18941 so CI can run against the combined final SHA.

pillowtalk-Qy:
Workflow validation follow-up `e81d158f23` is pushed. The prior failure was reproduced locally with actionlint (`unexpected character #` in the `if: |` expression) and fixed by moving the disabled-trigger comments outside that block scalar without changing the active trigger expression.

Evidence on the latest SHA:
- actionlint: failed before, passes after
- Ruby YAML, Prettier, and `git diff --check`: passed
- full unit suite: 1000 passed, 1 skipped
- lint and type-check: passed
- no new `claude-review-translations.yml` failure run was created for `e81d158f23`; the visible failed run belongs to old SHA `aa7985a`

This companion PR is now conflict-free. A maintainer still needs to review and merge it into `fix/intl-incremental-propagation`, then re-run the parent #18941 checks on the combined final SHA.

PR #18963, opened 2026-08-04
Mirror hot-linked meetup and community-pick images to S3

### LINKED ISSUE #18796: [recovery] Next.js image optimizer 400: "The requested resource isn't a valid image" on remote images

pettinarip:
## Description

Alternative to #18821 for the same production error (#18796).

The image optimizer 400s (`"The requested resource isn't a valid image"`) whenever a third-party image rots upstream. #18821 mitigates by marking ~13 hosts `unoptimized`. This PR removes the hot-links instead, using the pattern already in the repo: `uploadToS3` at ingestion, as `fetchEvents`, `fetchApps` and `fetchRSS` already do ("so /latest never hot-links third-party hosts").

Two surfaces were still hot-linking:

- **`community-meetups.json`** — 78 unique images across Meetup, Luma, Eventbrite, Twitter, Framer
- **`unavatar.io` avatars** on /apps

## Changes

- **`fetchMeetupImages`** — new daily task returning a `source URL → S3 URL` map. A map rather than a JSON rewrite keeps adding a meetup a pure data change: new entries render from their original URL until the next sync, and contributors need no S3 credentials.
- **`fetchCommunityPicks`** — mirrors each avatar into `CommunityPick.avatarImage` (optional, so existing blob records keep working). Falls back to the member's initial instead of hot-linking.
- **`events/meetup-groups.ts`** — new `server-only` module so pages just call `await getMeetupGroups(locale)`. It can't live in `events/utils.ts`, which client components import: pulling `src/lib/data` in there drags node built-ins into the browser bundle.
- **Replaced a dead ETHBhopal avatar** that 404s — a live instance of #18796 found while testing.

`remotePatterns` is unchanged, since the fallback path still resolves original URLs.

## Why not `unoptimized`

- **Nothing changes for users.** `event-card.tsx`, `ui/avatar.tsx` and `CardImage.tsx` already have `onError` fallbacks — a dead image renders the same placeholder either way. Only the log line moves. Mirroring stops images dying in the first place.
- **Privacy.** `/_next/image` proxies; `unoptimized` makes visitors' browsers hit Meetup/Twitter/Eventbrite directly. `CardImage` already sets `referrerPolicy="no-referrer"`, so avoiding that is existing intent.
- **Performance.** `unoptimized` drops srcset and format conversion — the event banner's tuned `sizes` would stop doing anything and mobile would pull full-resolution originals.

## Testing

Verified end-to-end against the dev environment (`trigger dev` + `data-layer-dev` blob):

| | `/events/meetups` | `/community/events` |
|---|---|---|
| `s3-dcl1` URLs | 491 | 345 |
| meetupstatic / lumacdn | **0** | **0** |

- `fetch-meetup-images` run: **78/78** images upload cleanly (was 77/78 before the ETHBhopal fix).
- `pnpm lint`, `pnpm type-check` clean. Unit suite 948 passing — the 20 `data-layer/getters.spec.ts` failures are pre-existing and env-dependent (confirmed identical on a clean tree).

⚠️ **For anyone re-running this:** `trigger dev` isolates Netlify Blobs via `BLOB_STORE_NAME`, but **not S3** — S3 credentials come from the Trigger.dev dev environment and point at the shared bucket, so a dev run writes real objects (additive, content-hashed, under the new `community/meetups/` prefix).

## Not covered

`WhatAreAppsStoriesList.tsx` still builds one `unavatar.io` URL. Its data comes from content files rather than a fetcher, so it needs a different mechanism — follow-up.

## Related

- Closes #18796
- Alternative to #18821


myelinated-wackerow (changes requested):
Approach is right (mirroring over `unoptimized`), but two things need addressing before this lands.

---

**1. `revalidate: false` means the map is never picked up without a second deploy**

The deploy preview still emits 320 `meetupstatic` + 25 `lumacdn` + 15 `pbs.twimg` + 10 `evbuc` hot-links on `/community/events/meetups/`, and every community pick on `/apps` renders as a letter tile. That's expected before the task's first run, but with `revalidate: false` the daily task filling the blob won't reach the pages until the *next* deploy after that — so merging leaves both surfaces unchanged (and `/apps` visibly degraded) for a day or more.

The comment's justification doesn't hold either: these pages aren't static. `getEventsData` at `CACHE_REVALIDATE_DAY` already puts them on day-granularity ISR, so `false` buys no staticness here.

```diff
 export const getMeetupImages = createCachedGetter(
   dataLayer.getMeetupImages,
   ["meetup-images"],
-  false
+  CACHE_REVALIDATE_DAY
 )
```

And worth triggering `fetch-meetup-images` + `fetch-community-picks` manually at deploy so the first render is already mirrored.

---

**2. A rotted source drops an image that's already mirrored**

`uploadToS3` fetches the source *before* the `HeadObjectCommand` existence check (`src/data-layer/s3.ts:150` returns `null` on `!response.ok`, never reaching `:193`). So the day a meetup logo 404s upstream, the next sync drops it from the map, `resolveMeetupImage` falls back to the dead original URL, and the optimizer 400s again — #18796 recurring on exactly the images this PR protects. For picks it's worse: the avatar reverts to a letter permanently.

The empty-map guard only catches total failure, not the partial case (40 of 78).

Cheapest durable fix, in `fetchMeetupImages`: seed from the stored map rather than rebuilding it.

```ts
const previous =
  (await get<Record<string, string>>(FETCH_MEETUP_IMAGES_TASK_ID)) ?? {}
// ...
return { ...previous, ...imageMap }
```

Same idea for `fetchCommunityPicks` (`avatarImage: uploaded ?? previousAvatar`). That also makes the empty-map throw redundant.

---

**3. Avatar fallback fails contrast**

`text-white` on `bg-accent-c`: dark mode resolves to `--teal-400` `#58F6C9` at **1.36:1**; light mode `--teal-700` `#0F9971` at **3.6:1**, still short of 4.5:1 (18px bold sits below WCAG's 18.66px large-text threshold).

`ui/avatar.tsx` already has `AvatarFallback` with `bg-body text-body-inverse`, which inverts per theme — reuse it instead of hand-rolling the tile.

---

**Nits**

- The `""` (sync failed) vs `undefined` (never synced) distinction is documented in `types.ts` but unused — the component treats both the same, which is what makes issue 1 user-visible. Falling back to the unavatar URL only when `avatarImage === undefined` would cover the gap.
- `fetch-community-picks.json` mock never got `avatarImage`, so `USE_MOCK_DATA=true` shows letters locally.
- `MeetupGroup` (`meetup-groups.ts`) and `MeetupImageSource` (`fetchMeetupImages.ts`) redeclare the same JSON shape — one exported type keeps writer and reader in sync.
- `import "server-only"` sits mid-file between value imports.

-- Opus 5

PR #18754, opened 2026-07-10
refactor: migrate Layer2NetworksTable + ProductTable to useTranslations (#18742)

myelinated-wackerow:
# Tables batch for #18742

Fourth batch per the #18749 recipe: **Layer2NetworksTable** (6 files → `page-layer-2-networks`) and **ProductTable** (`Filters`/`MobileFilters` → `table`; `PresetFilters` → `page-wallets-find-wallet`).

## Scope note

**`FindWalletProductTable` is deliberately excluded**: #18702 (find-wallet INP rebuild) rewrites 10 files in that subtree and #18715 touches it as well. Migrating it now would hand both PRs conflicts; it gets its own small batch after they land.

## Notes for review

- All 60+ literal keys pre-verified plain text (no ICU, no markup) in the catalogs; the wrapper's raw-return → cooked `t()` switch is behavior-identical for them.
- `PresetFilters`'s sr-only persona count is the one ICU call: colon form with values (already interpolated pre-migration), prefix now redundant and dropped — `t("page-find-wallet-persona-count-available", { count })`.
- `NetworkMaturityTooltip`'s template key `` t(`page-layer-2-networks-${maturity}-label`) `` resolves only to the five verified maturity-label keys.
- The two client components (`Layer2NetworksTable/index`, `PresetFilters`) keep `"use client"` as the first statement (regression check after the #18753 transform bug).

## Test plan

- [x] `pnpm type-check`, `pnpm lint` clean locally
- [x] Deploy-preview sweep: `/layer-2/networks/` × 25 locales incl. filter/tooltip strings (posted as comment)
- [x] Playwright: expand a network row + open maturity tooltip + wallet filter on the preview (en, ar, zh)

— Fable

PR #18973, opened 2026-08-05
revert(find-wallet): restore NFTs to persona slot 2

konopkja:
## What

Reverts the persona half of the July 2026 CTR test (95058c9033), restoring the order **New to crypto, NFTs, Hardware, Finance, Developer**.

The filter sidebar reorder from that same commit is deliberately **left in place** (see below).

This is a pure block reorder of `useWalletPersonaPresets.tsx`. No presets, copy, or translation keys change, and the later `useTranslations` refactor (969ebd3f01) is preserved.

## Why

Measured 24 days before vs 24 days after the change actually took effect. Note the effective date is **2026-07-10**, not the 07-07 merge: it shipped in v11.14.0 late on 07-09, which the daily event series confirms (mobile `Developer true` sits at ~50/day through 07-09, then steps to ~110/day).

All tests below are **day-level Welch t-tests on per-visit rates**, not pooled z-tests. Pooling overstates significance here because daily traffic is overdispersed.

| Metric (per 100 find-wallet visits) | Before | After | Change | |
|---|---|---|---|---|
| Mobile persona selections | 104.5 | 96.3 | **-7.8%** | p=0.0007 |
| Desktop persona selections | 52.2 | 59.8 | +14.6% | p=0.078, ns |
| All devices | 93.8 | 88.9 | **-5.2%** | |

Key points:

- **Find-wallet is 74% mobile**, the inverse of the site-wide 53/43 desktop split, so mobile drives the aggregate. The page is also only ~1% bot even on peak bot days, so these denominators need no correction.
- **NFTs is the more wanted chip at equal position on both devices.** In slot 5 it harvested 1.53x (desktop) and 1.76x (mobile) what Developer harvested in that same slot. In desktop slot 2 the two are a tie (0.96x, z=-0.6).
- That tie is why this reverts **globally** rather than per-device: desktop loses nothing measurable, and we avoid device detection on a statically generated page.
- Developer's apparent +114% desktop / +118% mobile was a **position effect** from moving slot 5 to slot 2, not evidence of demand.

## Caveats, please read before treating this as settled

1. **Wallet click-through never moved.** `WalletExternalLinkList` per 100 find-wallet visits went 59.6 to 63.0 on mobile, p=0.109 (ns); excluding the 3-day post-deploy burst it is +2.9%, p=0.40. Desktop was flat. So persona ordering changes which chip gets pressed without changing whether anyone reaches a wallet. This PR restores a better intermediate metric with no demonstrated conversion benefit.
2. **The original test was never cleanly isolated.** Four find-wallet-affecting changes shipped within the same 24 hours: the reorder, #18715 (filter sidebar UI bug fixes), #18754 (table i18n refactor), and the developer-tools redesign. Position-relative comparisons survive that; window totals do not.
3. **Ordering should be settled by an A/B test on Goal 10** (`Tap main button`, the real wallet click-through), with device as a pre-registered segment, using the Flags SDK harness from #18807. There is also a live untested hypothesis that presets may *cost* conversions, since the Developer preset switches on six filters at once and narrows the catalog hard.

## Sidebar left unchanged

The sidebar half of the original commit tested as null and is not touched here. Only Language support (moved up one slot, +36%, p=0.0001) survived Bonferroni correction across 8 sections. Network support lost just 8% (ns) despite dropping two slots, and two sections that did **not** move showed -13% and -22%, which puts section-level drift at roughly 10 to 20%.

## Verification

- `eslint` clean on the changed file; prettier applied via lint-staged.
- `pnpm type-check` returns 2 errors, both **identical on `dev` without this change** (stale `.next/dev/types/validator.ts` references to route files absent from source). Confirmed by typechecking `origin/dev` directly.
- E2E `selectPersonaFilter` locates personas by **name** regex and targets "New to crypto" (still slot 1), so the reorder does not affect it.
- The reorder is assertion-verified as content-preserving (sorted line multiset unchanged) and semantically identical to the pre-test configuration.


wackerow:
Hey @konopkja, is this reordering still needed?

image

Seeing this order for both A/B cases in production right now

konopkja:
yes its needed, developer should not be second

Issue #18967, opened 2026-08-04
Reorganize Storybook: repo-mirrored taxonomy, dead-code cleanup, coverage gaps


myelinated-wackerow:
## Summary

Our Storybook has grown into a patchwork of at least six competing organization schemes (`Atoms/Molecules/Organisms` leftovers, `UI /` vs `UI / Primitives /`, ad-hoc `Components / Widgets`, bare titles, one story literally titled "The early Web", and six stories with no title at all). This issue reorganizes it into a taxonomy that mirrors the repo's actual component architecture, removes dead components along the way, and fills coverage gaps so Storybook works as a design-system showcase: designers can see what exists, and engineers can see where it lives.

**Decision: we mirror the repo, not atomic design.** Atomic design already half-failed here because "molecule vs organism" is a judgment call and entropy wins. The repo layout (`ui` primitives / shared `src/components` / page-colocated `_components` / `layouts`) is a taxonomy with an objective test — the file path.

### Target taxonomy

Top level is derived from file path (deterministic, enforced by test). Second level is a small functional grouping for browsability.

```
Design System /      ← src/styles/__stories__ + foundations
  Colors · Gradients · Shadows · Typography (Heading, Text, FlowRhythm) · Icons
UI /                 ← src/components/ui, 1:1
  Actions /          Button, ButtonLink, Tag, CopyToClipboard
  Forms /            Input, Checkbox, Switch, Select, Field, Label
  Layout /           Flex, Grid, Section, EdgeScrollContainer, Skeleton
  Navigation /       Link, LinkBox, Breadcrumb, Tabs, TabNav
  Overlays /         Dialog, Modal, Sheet, Popover, DropdownMenu, Command, Tooltip
  Data Display /     Card, Table, List, Accordion, Avatar, Callout, Chart, Progress, Spinner, TagFilter, TruncatedText, TerminalTypewriter, PersistentPanel
Components /         ← src/components (shared, non-primitive)
  Cards /            AppCard, CommentCard, HighlightCard, HorizontalCard, SubpageCard, PathwayCard, MarkdownCard
  Heroes /           HomeHero, HubHero, PageHero
  Navigation /       Breadcrumbs, DocsNav, DocLink, TableOfContents, FilterBar, DeveloperDocsLinks
  Site Chrome /      Nav, Footer, FeedbackWidget, TranslationBanner, LanguagePicker, Search, CopyPageButton
[...]

PR #18989, opened 2026-08-05
chore: rename story files to kebab-case (Storybook reorg phase 6)

myelinated-wackerow:
## Summary

Phase 6 of #18967. Renames 119 story files to kebab-case, matching the repo's file convention and the direction components are already migrating -- `AppCard.stories.tsx` -> `app-card.stories.tsx`, including inside PascalCase component directories. **Component filenames are untouched**; that's a separate migration.

**Zero Chromatic impact, verified rather than assumed.** Story IDs derive from `meta.title`, not the filename, and every file has an explicit title (enforced since Phase 2). I captured the built `index.json` IDs before and after: the same 656 IDs, none added, none removed.

Each rename goes through a temporary name so the case-only ones (`Button` -> `button`) survive a case-insensitive checkout.

Adds a filename assertion to the taxonomy test so this can't re-drift -- negative-checked by renaming one file back to PascalCase and confirming it fails -- and replaces the note in #18988 that documented the inconsistency as acceptable.

Base: `docs/storybook-conventions` (phase 5, #18988).


pettinarip (approved):
🚀

PR #18993, opened 2026-08-06
Consolidate roadmap status, EIP chips, and Hegota page

nloureiro:
## Summary

- render Forkcast-backed upgrade status, mainnet targets, milestones, and EIP confidence on roadmap pages
- preserve actual, proposed, and projected testnet dates from Forkcast and fail loudly on date-schema drift
- make content-review reporting retryable by classifying and opening the human-review issue before advancing the sync branch
- add the Hegotá roadmap page and navigation entry

## Test plan

- `pnpm test:unit`
- `pnpm type-check`
- `pnpm lint`
- `pnpm lint:md`
- regenerated the store against live `ethereum/forkcast` and verified Glamsterdam/Hegotá roadmap pages render successfully

### DIFF
--- a/.github/workflows/sync-upgrades.yml
+++ b/.github/workflows/sync-upgrades.yml
@@ -73,12 +73,11 @@ jobs:
           pnpm type-check
           pnpm test:unit
 
-      # Empty output means nothing changed that any page can contradict, and a
-      # failure here costs a notification rather than the week's sync.
+      # Empty output means nothing changed that any page can contradict.
+      # A classifier failure stops the sync so it cannot suppress human review.
       - name: Classify the change
         id: classify
         if: steps.diff.outputs.changed == 'true'
-        continue-on-error: true
         env:
           SYNC_BASE_REF: ${{ steps.base.outputs.ref }}
         run: |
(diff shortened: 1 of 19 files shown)


nloureiro:
## Feasibility detail: layout-rendered vs MDX-embedded

Assessment behind the open question in the description.

### It would work, and it's small

- **All 24 translated Glamsterdam pages already carry `template: upgrade`**, so they already route through `TopicLayout`. Nothing needs to reach the translated markdown for them to pick this up.
- **`slug` is locale-independent.** `[...slug]/page.tsx` builds it as `slugArray.join("/")` → `"roadmap/glamsterdam"`, with the locale stripped by the route. `slug.split("/").pop()` is a reliable key into `src/data/upgrades`.
- **Precedent exists.** `[...slug]/page.tsx` already injects a component at layout level based on layout type (`StakingCommunityCallout` as `afterContent` for `layout === "staking"`), and `TopicLayout` already injects `<PageActions />` above `{children}`. This would sit in the same place.
- **Size:** roughly a lookup plus a conditional render in `TopicLayout`, and deleting the tag from the English markdown.

### The switch has to be data presence, not layout type

Six English pages use `template: upgrade` — `beacon-chain`, `dencun`, `fusaka`, `glamsterdam`, `merge`, `pectra` — and only Glamsterdam has a data file. So the condition is "does `upgrades[slug]` exist", not "is this the upgrade topic". That's self-limiting in a useful way: PR 4's Hegotá page gets the block by existing in the data, with no layout change.

Worth noting a marker in frontmatter (e.g. `upgrade: glamsterdam`) does **not** solve the problem — frontmatter propagates through the pipeline on the same latency as the MDX tag. Deriving from slug is the only approach with zero per-file markers, and therefore the only one that reaches all locales immediately.

### Costs

1. **Temporary duplication.** Translated pages keep their existing `<Alert>` until the pipeline propagates the English removal, so they'd show both the block and the old alert for one cycle. Not contradictory — most say "H2 2026" against the block's "2026" — but redundant.
2. **Authors lose placement control.** With a tag, the block can be moved or omitted per page. Layout-rendered, it's always in one spot. Given it's a status header, that's arguably correct, but it is a real loss of flexibility and the reason I didn't just assume it.
3. **Slightly more magic.** "This page renders a status block because a data file shares its slug" is less greppable than a tag in the markdown. Mitigable with a comment in `TopicLayout` and a note in the data README.

### Recommendation

Worth doing, but it's a design call about who owns page composition, so I'd rather have a maintainer view than switch unilaterally. If the answer is yes, the cleanest sequencing is to fold it into this PR before it leaves draft rather than adding a PR 7 that undoes the tag.

pettinarip (approved):

PR #18998, opened 2026-08-06
Add EIP inclusion status chips to upgrade pages

nloureiro:
> **Stacked on #18993**, and based on its branch so the diff here shows only this PR's own changes.
>
> A Netlify preview was verified while this PR briefly targeted `dev`: [deploy-preview-18998.ethereum.it](https://deploy-preview-18998.ethereum.it). Because CI and Netlify only run for PRs into `dev`/`master`/`staging`, further pushes here won't rebuild it until #18993 merges and this retargets to `dev`.

## Description

Adds an inclusion-status chip to each EIP section on the Glamsterdam page, driven by `eips[].status` from the generated data layer.

In July the page described already-scheduled EIPs as "being considered for inclusion". Driving the chip from data means that framing can't drift again.

## Three statuses render

| status | chip | tone |
| --- | --- | --- |
| `scheduled` | Scheduled for inclusion | success |
| `included` | Included | success |
| `considered` | Considered for inclusion | neutral |

`considered` reads as neutral rather than a success on purpose: an upgrade whose scope hasn't frozen may still drop it.

Anything else — `proposed`, `declined`, `withdrawn` — has **no entry in the store at all**, so a missing entry renders nothing. That's also the right answer for a descoped EIP: it should be removed from the page, not labelled.

## Why this PR touches the data layer

One line in `src/scripts/sync-upgrades/normalize.ts`, adding `considered` to the stored statuses, plus the regenerated store (**+1 row**: EIP-8141, considered for Hegotá at ACDE #233).

#19059 filtered `considered` out on the grounds that nothing consumed it. That was circular — this component is the consumer, and it only stopped consuming it because the data had stopped carrying it. `considered` means ACD has formally weighed the EIP, which is exactly the signal a planning-phase page wants before scope freezes.

`proposed` stays excluded, and that distinction is the point: Hegotá alone carries **64** proposals, churning at every ACD call, which would mean weekly diffs against data no page reads. `considered` costs one row.

A test now pins both directions so this can't quietly regress.

## What the test checks, and what it doesn't

`tests/unit/roadmap/eip-tags.spec.ts` catches one **mechanical** failure: a chip that points at nothing. `<EipTag id={7955} />` with a typo renders `null`, and a missing chip looks identical to a page that never had one. Verified the assertions bite — flipping one chip id by a single digit fails two of them.

It deliberately does **not** assert that every scheduled EIP has a section. The page covers 11 of Forkcast's 23, which is a pre-existing editorial state rather than anything this PR changes. Gating CI on it would only teach people to add exemptions, leaving a suite that reports green while coverage falls further behind.

Anyone wanting the current number can compare `eips[]` in `src/data/upgrades/generated.ts` against the `**Resources**` links on the page — a comparison that only became possible once #19059 made the data derive from Forkcast rather than be hand-fitted to the page.

## i18n

3 new strings (`scheduled`, `included`, `considered`).

### DIFF
--- a/public/content/roadmap/glamsterdam/index.md
+++ b/public/content/roadmap/glamsterdam/index.md
@@ -5,11 +5,10 @@ lang: en
 template: upgrade
 ---
 
+<UpgradeSummary slug="glamsterdam" />
+
 <Alert variant="update">
 <AlertContent>
-<AlertTitle>
-Glamsterdam is an upcoming Ethereum upgrade planned for Q4 2026
-</AlertTitle>
 <AlertDescription>
 The Glamsterdam upgrade is only a single step in Ethereum's long-term development goals. Learn more about [the protocol roadmap](/roadmap/) and [previous upgrades](/ethereum-forks/).
 </AlertDescription>
--- a/public/content/roadmap/glamsterdam/index.md
+++ b/public/content/roadmap/glamsterdam/index.md
@@ -5,8 +5,6 @@ lang: en
 template: upgrade
 ---
 
-<UpgradeSummary slug="glamsterdam" />
-
 <Alert variant="update">
 <AlertContent>
 <AlertDescription>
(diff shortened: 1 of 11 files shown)


nloureiro:
Superseded by the consolidated roadmap PR #18993, which now contains these EIP chip changes.

PR #19000, opened 2026-08-06
Add Hegota roadmap page in planning phase

nloureiro:
> **Stacked on #18993**, and based on its branch so the diff here shows only this PR's own changes.
>
> A Netlify preview was verified while this PR briefly targeted `dev`: [deploy-preview-19000.ethereum.it](https://deploy-preview-19000.ethereum.it). Because CI and Netlify only run for PRs into `dev`/`master`/`staging`, further pushes here won't rebuild it until #18993 merges and this retargets to `dev`.

## Description

Adds a `/roadmap/hegota/` page for the upgrade after Glamsterdam, in its planning phase, and registers it in the upgrades nav.

## No hand-written data

The original version of this PR added `src/data/upgrades/hegota.ts` by hand. **That file is gone** — Hegotá is generated from Forkcast like every other upgrade, so the page needs no data of its own:

```
UPGRADE STATUS
[ IN PLANNING ]
Expected on mainnet   Q2 2027 · Date not yet confirmed
Next milestone        Sepolia fork, 2027
→ Track live progress on Forkcast
```

`Q2 2027` comes from Forkcast's phase timeline. The testnet forks are known only to the year, which is why they render as `2027` — the data carries only the precision its source has.

Hegotá has no running milestone yet, so the stage tag falls back to its `status`, giving "In planning". That fallback is added in #18993.

## Scope

A thin page by design: the upgrade has one scheduled EIP so far, and anything more would be inventing content. It grows as ACD decides.

## i18n

2 new strings (planning-phase label, nav entry).

### DIFF
--- a/public/content/roadmap/glamsterdam/index.md
+++ b/public/content/roadmap/glamsterdam/index.md
@@ -5,11 +5,10 @@ lang: en
 template: upgrade
 ---
 
+<UpgradeSummary slug="glamsterdam" />
+
 <Alert variant="update">
 <AlertContent>
-<AlertTitle>
-Glamsterdam is an upcoming Ethereum upgrade planned for Q4 2026
-</AlertTitle>
 <AlertDescription>
 The Glamsterdam upgrade is only a single step in Ethereum's long-term development goals. Learn more about [the protocol roadmap](/roadmap/) and [previous upgrades](/ethereum-forks/).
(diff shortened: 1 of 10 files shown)


nloureiro:
Superseded by the consolidated roadmap PR #18993, which now contains the Hegota page changes.

PR #19002, opened 2026-08-06
feat: track top-level nav menu interactions

myelinated-wackerow:
Adds Matomo events for the top-level navigation. Desktop section opens were never tracked at all; the mobile top-level sections lost their coverage in an earlier refactor when they moved to a plain `Collapsible`. This establishes a baseline for both.

## Nav menu tracking - how to read the numbers

**Where it lands in Matomo** (Behaviour -> Events)

| Category | Action | Name |
|---|---|---|
| `Desktop navigation menu` | `Section changed` | `Open section: en - learn` |
| `Mobile navigation menu` | `Section changed` | `Open/Close section: en - learn` |
| `Mobile navigation menu` | `Menu toggled` | `Open/Close menu: en` |

Section names are always English (`learn`, `use`, `build`, `participate`, `research`) regardless of the visitor's language. The locale is the `en` part of the name, so you can split by language.

**What one event means**

One event is one interaction. Every row gives you both numbers: `nb_events` is how many times the thing happened, `nb_visits` is how many distinct sessions did it at least once. So "how many people looked in Learn" and "how often did they look" are both available, and open counts divide cleanly into leaf-link clicks to give open-to-click conversion.

**Desktop opens require a 300ms pause**

The desktop menu opens on hover, so a mouse moving left to right across the bar would otherwise register an open for all five sections. An open is only recorded once the pointer has rested on a section for 300ms. A section brushed past in transit is not counted; a section someone actually stopped on is.

**Two caveats that matter for interpretation**

1. **Desktop opens on hover, mobile opens on tap.** Even with the dwell filter, a deliberate 300ms hover is a weaker signal of intent than a tap. Expect desktop numbers to run somewhat higher than mobile for the same real interest. Compare desktop-to-desktop and mobile-to-mobile.
2. **Desktop records no close event.** A desktop menu closes when the mouse leaves, which is not an interaction. Mobile records both open and close, since both are taps. Do not compare open/close ratios across the two.

**What this is good for:** relative demand between the five sections, open-to-click conversion, how both differ by device and language, and as a baseline for A/B testing a nav redesign.

**What it can't tell you:** how long someone browsed the menu, or whether a hover past 300ms was genuinely intentional.

## Bug fixed along the way

`MenuContent` called `useNavMenu` a second time, forking the hook state and leaving `activeSection` permanently `null`. Existing desktop submenu link events have therefore been reporting `Menu - null - en`. Threading the parent state down as props additionally makes `onClose` work (clicking a nav link now closes the menu) and cuts the duplicated number-key focus listener from six registrations to one.

**Continuity heads-up:** on deploy, the desktop leaf-click action changes from `Menu - null - en` to `Menu - learn - en` and so on. Every desktop section series splits at that date, and any saved segment keyed on the old string stops matching. The historical data is not lost - per @konopkja, the section is recoverable from the href in `eventName` for 98.4% of the 26,816 affected events using the current nav config, and 99.996% using the union of nav versions live during the window.

## Not included

Kept out to hold the diff scoped, all worth separate PRs:

- **Two orphaned files**, both zero references: `Nav/MobileMenu/MenuAccordion.tsx` (carries a full mobile section-tracking implementation that has never run - this is what made mobile look instrumented) and `Nav/MobileMenu/TrackedCollapsible.tsx` (a duplicate of `ui/collapsible.tsx`'s `CollapsibleTracked`).
- **Desktop leaf clicks double-fire.** `SubMenu`'s `BaseLink` passes no `customEventOptions`, so `Link.tsx` falls back to `eventCategory: "Link"` on top of the menu event, inflating `Link` by roughly 26.5k events.
- **Search** records the open but neither the query typed nor the result clicked.

## Test plan

- [ ] Desktop: resting on a section reports one open; sweeping across the bar reports only where the pointer stops
- [ ] Desktop: clicking a submenu link reports the real section name, not `null`, and closes the menu
- [ ] Mobile: every section open and close reports, including repeat toggles
- [ ] Mobile: level-2+ sub-sections still fire on every toggle (unchanged)

Events only fire when `NODE_ENV=production`, so verify on the deploy preview rather than locally.

-- Opus 5

### DIFF
--- a/src/components/Nav/Menu/MenuContent.tsx
+++ b/src/components/Nav/Menu/MenuContent.tsx
@@ -1,13 +1,12 @@
-import { motion } from "motion/react"
+import { motion, type MotionProps } from "motion/react"
 import { tv } from "tailwind-variants"
 import { Content } from "@radix-ui/react-navigation-menu"
 
 import { cn } from "@/lib/utils/cn"
 
-import { NavItem, NavSections } from "../types"
+import { NavItem, NavSectionKey } from "../types"
 
 import SubMenu from "./SubMenu"
-import { useNavMenu } from "./useNavMenu"
 
 export const navMenuVariants = tv({
   slots: {
@@ -50,12 +49,21 @@ export const navMenuVariants = tv({
 type MenuContentProps = {
   items: NavItem[]
   isOpen: boolean
-  sections: NavSections
+  activeSection: NavSectionKey | null
+  containerVariants: MotionProps["variants"]
+  onClose: () => void
 }
 
 // Desktop Menu content
-const MenuContent = ({ items, isOpen, sections }: MenuContentProps) => {
-  const { activeSection, containerVariants, onClose } = useNavMenu(sections)
+// State comes from the single useNavMenu instance in the parent; calling the
+// hook here would fork it and leave activeSection permanently null
+const MenuContent = ({
+  items,
+  isOpen,
+  activeSection,
+  containerVariants,
+  onClose,
+}: MenuContentProps) => {
   const { base } = navMenuVariants()
 
   return (
--- a/src/components/Nav/Menu/index.tsx
+++ b/src/components/Nav/Menu/index.tsx
@@ -24,8 +24,14 @@ type NavMenuProps = BaseHTMLAttributes<HTMLDivElement>
 
 const Menu = ({ ...props }: NavMenuProps) => {
   const { linkSections } = useNavigation()
-  const { activeSection, direction, handleSectionChange, isOpen } =
-    useNavMenu(linkSections)
+  const {
+    activeSection,
+    containerVariants,
+    direction,
+    handleSectionChange,
+    isOpen,
+    onClose,
+  } = useNavMenu(linkSections)
 
   return (
     <div {...props}>
@@ -65,7 +71,9 @@ const Menu = ({ ...props }: NavMenuProps) => {
                 <MenuContent
                   items={items}
                   isOpen={isOpen}
-                  sections={linkSections}
+                  activeSection={activeSection}
+                  containerVariants={containerVariants}
+                  onClose={onClose}
                 />
               </Item>
             )
--- a/src/components/Nav/Menu/useNavMenu.ts
+++ b/src/components/Nav/Menu/useNavMenu.ts
@@ -1,7 +1,9 @@
-import { useState } from "react"
+import { useRef, useState } from "react"
 import type { MotionProps } from "motion/react"
+import { useLocale } from "next-intl"
 
 import { isModified } from "@/lib/utils/keyboard"
+import { trackCustomEvent } from "@/lib/utils/matomo"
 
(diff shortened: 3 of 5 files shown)


konopkja:
Thanks for this. The diagnosis matches what we measured independently on the Matomo side, and threading the state down from the single `useNavMenu` instance is the right root-cause fix rather than a patch.

One thing I would want changed before merge, and three notes.

## Please drop the once-per-session cap

Matomo already gives us this server side. Every event row returns both `nb_events` and `nb_visits`, where `nb_visits` is distinct visits that fired that exact category + action + name. The existing mobile event `Open section: Start with Ethereum...` currently reports 3,142 events across 2,737 visits, so "sessions that opened Learn at least once" is available today without capping anything, and we get the interaction count on top of it.

The cap costs us the thing we added this tracking for. As the PR body says, opens and link clicks will not form a clean funnel. That funnel is why this was prioritised: without it we cannot measure open-to-click conversion, and we cannot A/B test a nav redesign, because a change in leaf clicks cannot be split into "more people opened the menu" versus "more of the people who opened it converted".

It also does not fix the hover problem it targets. A single left-to-right mouse sweep fires `onValueChange` for all five sections, and the cap makes that sweep permanent and indistinguishable from real intent. Left uncapped, a sweep is trivially detectable in analysis (five opens inside 500ms) and filterable.

Suggestion: fire every open, and add a dwell threshold in `handleSectionChange`, a `setTimeout` of roughly 300ms cancelled if the value changes again. That removes sweeps at the source and preserves both metrics. Mobile has no hover, so I would drop the cap there regardless of what we decide for desktop.

## Notes

**Capped and uncapped events in the same category.** `Menu toggled` is uncapped while `Section changed` is capped, both under `Mobile navigation menu`. Anyone computing sections-per-menu-open will divide a capped numerator by an uncapped denominator. Whichever way we go, applying it uniformly matters more than the choice itself.

**The historical desktop data is recoverable, so please do not write it off.** We reconstructed the section for 98.4% of the 26,816 `Menu - null - en` events from the href in `eventName` using the current nav config, and 99.996% using the union of the nav versions that were live during the window, with zero cross-section href moves. That is three months of usable section data.

**Continuity heads-up.** `SubMenu` consumes `activeSection`, so the desktop leaf-click action changes from `Menu - null - en` to `Menu - learn - en` on deploy. That is the correct outcome, but every desktop section series splits at that date and any saved segment keyed on the old string stops matching. Worth a line in the PR description.

## Out of scope here

Two adjacent gaps I would keep out of this PR rather than grow it: desktop leaf clicks double-fire into the site-wide `Link` category, because `SubMenu`'s `BaseLink` has no `customEventOptions` (it inflates `Link` by roughly 26.5k events and skews any menu versus in-content comparison), and search records the open but neither the query typed nor the result clicked.

wackerow:
Thanks @konopkja, suggestions applied and PR description updated

pettinarip (approved):

PR #19004, opened 2026-08-06
refactor(quiz): regroup hub sections, rebuild layout, live community stats

myelinated-wackerow:
## Summary

Reworks the /quizzes hub in three parts.

**Sections.** Regroups from 3 sections into 5, defined once as an ordered `quizzesSections` array that both the hub render and `getNextQuiz` derive from, so a new section can't be forgotten in a second place. `wallets` sits in Ethereum basics, since holding your own keys is foundational, which leaves the second section as "Security and privacy". Also fixes the "completed X/Y" denominator, which summed only two of the three section arrays and undercounted by 2.

**Layout.** The two stat boxes move from a right-hand column to sit side by side below the hero. Each section is now a full-width row with its heading and description on the left and its quiz list on the right, collapsing to a single column on narrow screens. The columns are grid tracks rather than flex: under `border-box` a flex item's `flex-basis: 0` is floored by its own padding/border, so the bordered quiz lists and the padded contribute callout resolved to different 1/3 splits and sat ~20px out of line, while `minmax(0, 1fr)` tracks ignore item padding. Section headings stick alongside their list on desktop, the whole quiz row is clickable via `LinkBox`/`LinkOverlay` so it keeps one tab stop, the hero moves to the server page, spacing comes from the flow rhythm and the page/space tokens rather than hand-rolled margins, and difficulty tags pick up the green/yellow/red statuses.

**Community stats.** Replaces three constants committed in June 2023 and never updated with a daily Matomo fetch, read through a cached getter in `src/lib/data` so the hub revalidates daily rather than freezing the panel at build time. Average score and retry rate derive from events already tracked. The fetcher throws rather than falling back, so failures surface through the existing Sentry/Discord handler, and the panel hides entirely when there's no data instead of showing stale numbers. That panel renders on the server; only the user's own score box stays client-side, since it reads `localStorage`.

Along the way: the `daos` and `defi` quizzes were carrying literal strings rather than catalog keys as titles, so they had never been translatable in any locale - they now point at the existing, already-translated `common.dao-page` / `common.defi-page`; the `energy-consumption` quiz takes new copy, English-only until the next pipeline run; the deprecated `Translation` component is gone from these components; `setRequestLocale` now precedes the next-intl calls it guards; and `addNextQuiz` moves to `src/data/quizzes` to break an import cycle with `components/Quiz/utils`.

The Matomo fetcher has been run against live credentials via `trigger:dev` and returns sane figures.

## Preview link
https://deploy-preview-19004.ethereum.it/quizzes


app/[locale]/quizzes/page.tsx
```diff
@@ -7,21 +7,41 @@ import {
...
+import { Section } from "@/components/ui/section"
 
 import { getAppPageContributorInfo } from "@/lib/utils/contributors"
+import { formatDate } from "@/lib/utils/date"
 import { getMetadata } from "@/lib/utils/metadata"
 import { getRequiredNamespacesForPage } from "@/lib/utils/translations"
 
-import QuizzesPage from "./_components/quizzes"
-import QuizzesPageJsonLD from "./page-jsonld"
+import { getQuizStats } from "@/data-layer"
```
Comment (pettinarip) [inline app/[locale]/quizzes/page.tsx]: **Blocker — the "daily" community stats will never refresh.**

`getQuizStats` is imported straight from `@/data-layer`, bypassing the `createCachedGetter` layer in `src/lib/data/index.ts` that every other page uses (~20 call sites; `IssuesList` is the only other exception). Without that wrapper there's no `unstable_cache`, and therefore no `revalidate` on this route — and the page is fully prerendered (`setRequestLocale`, no `revalidate` export).

Net effect: the community panel and its `last-updated` stamp freeze at build time, so the daily Trigger.dev fetch only reaches visitors on the next deploy — the staleness this PR sets out to fix. Secondary: 25 uncached Netlify Blobs reads per build instead of one deduped read.

Suggest adding a cached getter in `src/lib/data/index.ts` and importing from there.

src/data/quizzes/index.ts
```diff
@@ -112,7 +119,7 @@ const quizzes = {
     ],
   },
   "energy-consumption": {
-    title: "energy-consumption",
+    title: "learn-quizzes:page-energy-consumption",
```
pettinarip, line 122: **This one is a translation regression, not a fix.**

The PR description groups `energy-consumption` with `daos`/`defi` as "carrying literal strings rather than catalog keys, so they had never been translatable" — but `"energy-consumption"` was a real `common` key, translated in every locale (`es: "Consumo de energía de Ethereum"`, `ja: "イーサリアムのエネルギー消費量"`, …).

Swapping it for the new English-only `learn-quizzes:page-energy-consumption` means all 24 non-English locales render English for this row until Crowdin round-trips.

Related, lower impact: `common.dao-page` and `common.defi-page` already exist and are fully translated (same `*-page` convention as the neighbouring `nft-page`/`payments-page`), so lines 68 and 109 add untranslated keys where translated ones were available.

src/data-layer/fetchers/fetchQuizStats.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  // Never default a missing row to zero: that is indistinguishable from a real
+  // zero, which is how the previous hard-coded figures went stale unnoticed.
+  if (!retried) {
+    throw new Error(`No "${RETRY_NAME}" row returned by Matomo`)
+  }
+
+  const questionsAnswered = answered.nb_events
+  const averageScore =
+    ((answered.sum_event_value ?? 0) / questionsAnswered) * 100
```
pettinarip: `?? 0` here contradicts the guard eight lines above.

Lines 80–89 deliberately throw when the `Question answered` or `Retry question` row is missing, with the comment *"Never default a missing row to zero: that is indistinguishable from a real zero"*. This line then does exactly that for the value sum — if Matomo returns the action row without `sum_event_value` (event values dropped, archive rebuilt without them), `averageScore` becomes `0` and the panel advertises "Average score 0%" instead of surfacing the failure to Sentry/Discord.

Suggest throwing on a missing `sum_event_value` the same way.

pettinarip (commented):
Automated code review (`/code-review`). 5 findings below as inline comments — one blocker, one i18n regression, three minor.

Not commented inline because it has no diff anchor: the branch is behind `dev`'s Quiz story rename, so merging produces a rename/delete conflict on `src/components/Quiz/stories/quizzes-stats.stories.tsx`. `dev`'s version is kept in the tree and imports `../QuizzesStats`, which this PR deletes — resolve by deleting the renamed story file, not by keeping `dev`'s. `quizzes-list.stories.tsx` merges cleanly, no action needed there.

app/[locale]/quizzes/page.tsx
```diff
@@ -32,11 +52,106 @@ const Page = async (props: { params: Promise<PageParams> }) => {
...
   )
 
+  // Null until the daily Matomo fetch lands; the community panel hides in that case
+  const communityStats = await getQuizStats()
+
+  // Pinned to UTC: there is no viewer timezone to honour at render time, and the
+  // page is statically rendered, so the stamp must be timezone-independent.
+  const lastUpdated =
+    communityStats &&
+    formatDate(new Date(communityStats.timestamp).toISOString(), locale, {
```
Comment (pettinarip) [inline app/[locale]/quizzes/page.tsx]: Pre-converting with `.toISOString()` defeats `formatDate`'s own invalid-date guard.

`formatDate` checks `isValidDate` and returns `""` for malformed input, but it never gets the chance here: if a blob ever lands without a numeric `timestamp` (fetcher shape change, hand-written blob, partial write), `new Date(undefined).toISOString()` throws `RangeError: Invalid time value` during prerender and fails the build for all 25 locales, rather than degrading to a hidden stamp.

Passing `communityStats.timestamp` through a validity check — or handing `formatDate` a value it can reject — keeps the intended graceful failure.

app/[locale]/quizzes/page.tsx
```diff
@@ -32,11 +52,106 @@ const Page = async (props: { params: Promise<PageParams> }) => {
...
+        header={t("test-your-knowledge")}
+        description={t("quizzes-subtitle")}
+      />
+
+      {/* Wraps the whole body: the stats panel and the quiz lists are client
+          components and need the messages. */}
+      <I18nProvider locale={locale} messages={messages}>
+        <main className="px-page pt-space-3x pb-page">
+          <MainArticle className="flow space-y-space-3x">
+            <Grid columns={2} size="wider">
```
Comment (pettinarip) [inline app/[locale]/quizzes/page.tsx]: `columns={2}` folds by content width, so the sibling's `md:col-span-2` is keyed to the wrong breakpoint.

With `size="wider"`, `grid-cols-auto-2` resolves to `repeat(auto-fill, minmax(min(100%, max(26rem, (100% - 1rem)/2)), 1fr))` — two tracks only exist once the container reaches ~848px, roughly a 912px viewport given `px-page`. Between `md` (768px) and that point the grid is one column, so with `communityStats === null` the `md:col-span-2` item spans into an implicit second track, and with stats present the two panels stack despite the `md:` naming.

`Grid`'s docs point at `balanced={2}` for a fixed pair of items ("a breakpoint-driven alternative to `columns={2}`, which folds by content width") — that would make the breakpoint and the span agree.

pettinarip (approved):

PR #19009, opened 2026-08-06
[recovery] fix(founders): set request locale before next-intl APIs (static→dynamic bailout on /api/founders)

pettinarip:
## Production error

Captured from Netlify logs (`___netlify-server-handler`) via Grafana → Keep.

```
[ERROR] Error: Page changed from static to dynamic at runtime /api/founders, reason: headers
see more here https://nextjs.org/docs/messages/app-static-to-dynamic-error
    at y (.next/server/chunks/ssr/1itz_next_dist_1j1_2py._.js:2:12450)
```

- **URL / resource:** `/api/founders`
- **Occurrences:** 1 in this alert window (`hit_count: 3`), last seen 2026-08-06T19:36:45.080Z
- **Request ID:** `01KZC971DSR32TKR67WNBDMFZW`

## Root cause

`/api/founders` is not a route handler — `app/api/` only contains `gas-eth-price`, `revalidate`, `torch-webhook` and `gfi-issues-webhook`. The request therefore falls through to the `app/[locale]/…` tree with `"api"` captured as the `[locale]` param. `"api"` is not in `generateStaticParams`, so the page renders **on-demand** rather than being served from the prerender.

`app/[locale]/founders/page.tsx` called `getTranslations("page-founders")` as the first next-intl API in the page component — it never called `setRequestLocale` at all (only `generateMetadata` primed it). With the per-request locale cache unprimed, next-intl's `requestLocale` falls back to reading the request **headers**, which opts the route into dynamic rendering. Next.js aborts with `Page changed from static to dynamic at runtime` instead of rendering statically or returning a clean 404.

At build time there are no request headers, so every prerendered locale route looks fine — which is why this stayed latent and only surfaces on bot/broken-link probes to on-demand paths.

This is the documented failure mode in [`docs/solutions/architecture/setrequestlocale-static-to-dynamic-rendering.md`](../blob/dev/docs/solutions/architecture/setrequestlocale-static-to-dynamic-rendering.md), same class as #18785, #18797, #18798, #18800, #18811 and the recent recovery PRs #19003, #19005, #19006, #19007, #19008.

## Fix

One file, four added lines: call `setRequestLocale(locale)` immediately after resolving `locale` from `params`, before the `getTranslations` call.

```diff
 const Page = async (props: { params: Promise<PageParams> }) => {
   const params = await props.params
   const { locale } = params
+
+  // Set locale before next-intl APIs so on-demand renders stay static
+  setRequestLocale(locale)
+
[...]

PR #19008, opened 2026-08-06
[recovery] fix(start): set request locale before next-intl APIs (static→dynamic bailout on /api/start)

pettinarip:
## Production error

```
[ERROR] Error: Page changed from static to dynamic at runtime /api/start, reason: headers
see more here https://nextjs.org/docs/messages/app-static-to-dynamic-error
    at y (.next/server/chunks/ssr/1itz_next_dist_1j1_2py._.js:2:12450)
```

- **URL / resource:** `/api/start`
- **Function:** `___netlify-server-handler`
- **Occurrences:** 1 (last seen 2026-08-06T19:36:23Z, `hit_count: 4`)
- **Fingerprint:** `grafana-fn-error-page-changed-from-static-to-dynamic-at-runtime-api-start-reason-header`

## Root cause

`app/[locale]/start/page.tsx` called `getTranslations("page-start")` **before** `setRequestLocale(locale)`.

`proxy.ts`'s matcher excludes `/api`, so a request to `/api/start` reaches the App Router without i18n rewriting and matches `app/[locale]/start/page.tsx` with `locale = "api"`. That param is not in `generateStaticParams`, so Next.js renders it on-demand. Because `setRequestLocale` had not yet primed the per-request cache, next-intl's `requestLocale` (`src/i18n/request.ts`) fell back to reading the request **headers**, which opts the route into dynamic rendering — and Next.js aborts the on-demand static render with the error above instead of rendering statically or returning a clean 404.

This is the same class of bug documented in `docs/solutions/architecture/setrequestlocale-static-to-dynamic-rendering.md` and previously fixed for `/videos/[slug]` (#18785), `/wallets` (#18797), `/resources` (#18798), the `[...slug]` catch-all (#18800), and the systemic `generateMetadata` sweep (#18811). The `/start` page's `generateMetadata` was already correct; only the page component had the wrong ordering.

## Fix

One file, three lines: hoist `setRequestLocale(locale)` above the `getTranslations` call in the page component.

```diff
   const { locale } = params
-  const t = await getTranslations("page-start")
 
+  // Set locale before next-intl APIs so on-demand renders stay static
   setRequestLocale(locale)
 
+  const t = await getTranslations("page-start")
```

No behavior change for prerendered locales (build-time renders have no headers to read); on-demand renders of `/api/start` and friends now stay static and fall through to the normal 404 path.

## Verification

[...]

PR #19007, opened 2026-08-06
[recovery] fix(use-cases): set request locale before next-intl APIs (static→dynamic bailout on /api/use-cases)

pettinarip:
## Production error

Captured from Netlify (`___netlify-server-handler`) via Grafana → Keep.

```
[ERROR] Error: Page changed from static to dynamic at runtime /api/use-cases, reason: headers
see more here https://nextjs.org/docs/messages/app-static-to-dynamic-error
    at y (.next/server/chunks/ssr/1itz_next_dist_1j1_2py._.js:2:12450)
```

- **URL / resource:** `/api/use-cases`
- **Occurrences:** 1 alert (`hit_count: 3`), last seen `2026-08-06T19:36:20.347Z`
- **Request id:** `01KZC97007D62AP13940YC4K3G`

## Root cause

`/api/use-cases` has no route handler (`proxy.ts`'s matcher excludes `/api`), so it falls through to the App Router and is captured by `app/[locale]/…` with `locale = "api"`, `slug = ["use-cases"]`. That param is not in `generateStaticParams`, so it renders on demand.

`app/[locale]/use-cases/page.tsx` called `getTranslations("page-use-cases")` **before** priming the next-intl locale cache. Without `setRequestLocale(locale)`, next-intl's `requestLocale` falls back to reading the request **headers** (`src/i18n/request.ts`), which opts the route into dynamic rendering. On an on-demand render Next.js aborts with the static-to-dynamic error instead of rendering statically or returning a clean 404.

`generateMetadata` in the same file already called `setRequestLocale(locale)` (from the systemic sweep in #18811); the page component was missed — the same gap fixed for `/api/learn` in #19006, `/api/get-eth` in #19003, `/api/what-is-ethereum` in #19005 and `/api/stablecoins` in #18994.

This is the documented failure mode in `docs/solutions/architecture/setrequestlocale-static-to-dynamic-rendering.md`.

## Fix

One file, three lines — call `setRequestLocale(locale)` as the first next-intl-touching statement of the page component, matching the rule in the solutions doc and the existing `generateMetadata` in the same file.

```diff
 export default async function Page(props: { params: Promise<PageParams> }) {
   const params = await props.params
   const { locale } = params
+
+  setRequestLocale(locale)
+
   const t = await getTranslations("page-use-cases")
```

`setRequestLocale` was already imported in this file, so no import change is needed.

## Verification

[...]

PR #19006, opened 2026-08-06
[recovery] fix(learn): set request locale before next-intl APIs (static→dynamic bailout on /api/learn)

pettinarip:
## Production error

Captured from Netlify logs (`___netlify-server-handler`) via Grafana → Keep:

```
[ERROR] ⨯ Error: Page changed from static to dynamic at runtime /api/learn, reason: headers
see more here https://nextjs.org/docs/messages/app-static-to-dynamic-error
    at y (.next/server/chunks/ssr/1itz_next_dist_1j1_2py._.js:2:12450)
```

- **URL:** `/api/learn` — `api` is captured as `[locale]`, so this resolves to the learn hub page rendered on-demand (no `generateStaticParams` entry for `api`).
- **Occurrences:** 1 in the alert window, `hit_count: 6`; last seen 2026-08-06T19:36:16Z.
- **Request id:** `01KZC96EBBAK9P389NWKRX0SGJ`

## Root cause

next-intl resolves the request locale lazily: `requestLocale` in `src/i18n/request.ts` reads the request **headers** unless `setRequestLocale(locale)` has already primed the per-request cache. Any next-intl API called before priming forces that header read and opts the whole route into dynamic rendering.

`app/[locale]/learn/page.tsx` called `getTranslations("page-learn")` / `getTranslations("common")` in the page component without priming first. At build time there are no headers, so the prerendered locale routes are fine and the bug stays latent — but an on-demand render (a `[locale]` value outside `generateStaticParams`, such as `api`) performs the header read at request time and Next.js aborts with the static-to-dynamic error instead of rendering statically or returning a clean 404.

`generateMetadata` in this same file already had `setRequestLocale` (added by the systemic sweep in #18811, which covered `generateMetadata` only); the page component was missed.

This is the documented failure mode in `docs/solutions/architecture/setrequestlocale-static-to-dynamic-rendering.md`, and matches prior recovery PRs #18785, #18797, #18798, #18800, #18994, #19003, #19005.

## Fix

One file, three lines — call `setRequestLocale(locale)` as the first statement of the page component, before any next-intl API:

```diff
 export default async function Page(props: { params: Promise<PageParams> }) {
   const params = await props.params
   const { locale } = params
+
+  setRequestLocale(locale)
+
   const t = await getTranslations("page-learn")
   const tCommon = await getTranslations("common")
```

`setRequestLocale` was already imported in this file. No other changes.

## Verification

[...]

PR #19010, opened 2026-08-06
[recovery] fix(collectibles): set request locale before next-intl APIs (static→dynamic bailout on /api/collectibles)

pettinarip:
## Production error

```
[ERROR] Error: Page changed from static to dynamic at runtime /api/collectibles, reason: headers
```

- **URL / resource:** `/api/collectibles`
- **Occurrences:** 1 (`hit_count: 4`), last seen `2026-08-06T19:42:38.315Z`
- **Request id:** `01KZC9772XP8DGGDNJ06FQZQXG`

## Root cause

`proxy.ts` excludes `api` from its matcher (`"/((?!api|_next|_vercel|.well-known|.*\\.[^/]*$).*)"`), so a probe for `/api/collectibles` never gets locale-normalized. There is no `app/api/collectibles` route handler, so Next.js matches the `[locale]` segment instead, with `locale = "api"` and the static segment `collectibles` — i.e. it renders `app/[locale]/collectibles/page.tsx`.

In that page, `getTranslations()` ran **before** `setRequestLocale(locale)`:

```ts
const { locale } = params
const t = await getTranslations("page-collectibles")
setRequestLocale(locale)
```

With no cached request locale yet, next-intl's `getRequestLocale()` falls back to `getLocaleFromHeader()`, which calls `headers()` from `next/headers` (`next-intl/dist/.../server/react-server/RequestLocale.js`). Reading `headers()` marks the render dynamic, so the prerendered page reports `revalidate: 0` and Next.js throws the static→dynamic bailout from `next/dist/build/templates/app-page.js`.

`generateMetadata` in the same file already ordered these calls correctly; only the page component was affected.

This is the same defect already fixed for sibling pages in #18994 (stablecoins), #19003 (get-eth), #19005 (what-is-ethereum), #19006 (learn), #19007 (use-cases), #19008 (start) and #19009 (founders). `/api/collectibles` is a distinct route not covered by any of those.

## Fix

Move `setRequestLocale(locale)` above the first next-intl call in `app/[locale]/collectibles/page.tsx`, matching the ordering used by the sibling fixes and by this file's own `generateMetadata`.

## Verification

- `pnpm type-check` — passed (`next typegen && tsc --noEmit && tsc --noEmit -p netlify/edge-functions`, no errors)
- `pnpm exec eslint "app/[locale]/collectibles/page.tsx"` — clean, no output

### DIFF
--- a/app/[locale]/collectibles/page.tsx
+++ b/app/[locale]/collectibles/page.tsx
@@ -44,8 +44,11 @@ async function fetchStats() {
 export default async function Page(props: { params: Promise<PageParams> }) {
   const params = await props.params
   const { locale } = params
-  const t = await getTranslations("page-collectibles")
+
+  // Set locale before next-intl APIs so on-demand renders stay static
   setRequestLocale(locale)
+
+  const t = await getTranslations("page-collectibles")
 
   // Fetch data
   const [badges, stats]: [Badge[], Stats] = await Promise.all([


pettinarip:

PR #19016, opened 2026-08-07
fix(i18n): set request locale before next-intl APIs in all [locale] pages

pettinarip:
Replaces the 11 per-route recovery PRs for this bug class with one AST-verified sweep, then hardens the follow-up so a primed `generateMetadata` cannot hide an unsafe `Page`.

## Problem

`proxy.ts`'s matcher excludes `api`, `_next`, `_vercel`, `.well-known` and any dotted path, so next-intl's middleware never normalizes those segments and they land in `[locale]` as the locale. When the second segment collides with a real page slug, the page renders on-demand with an invalid locale — and if it touches a locale-resolving next-intl API before `setRequestLocale`, next-intl falls back to `headers()` and Next.js aborts:

```
Error: Page changed from static to dynamic at runtime /api/<slug>, reason: headers
```

`/api/<garbage>` is already a clean 404; only the collisions 500.

## Audit

Walked every entry point under `app/` with the TypeScript AST — default exports, `generateMetadata`, `generateViewport` — across `page`/`layout`/`not-found`/`default`/`template`/`loading`/`opengraph-image` files, checking call ordering per function rather than per file. **A file-level grep is unreliable here**: most affected pages already call `setRequestLocale` in `generateMetadata`, so the file matches while the `Page` body is still broken. That is why the earlier sweep (#18811) reported ~54 pages clean and left these 14.

Only forms without an explicit `locale` property resolve the request locale. `getConfig` short-circuits it when one is passed:

```js
get requestLocale() {
  return localeOverride ? Promise.resolve(localeOverride) : getRequestLocale()
}
```

So `getTranslations({ locale, namespace })`, `getMessages({ locale })`, `getFormatter({ locale })`, `getNow({ locale })` and `getTimeZone({ locale })` never read headers. A spread such as `{ ...options }` does not prove a locale is present. `getLocale()` has no override and always resolves the request locale.

`getMetadata` now forwards its required `locale` into `getTranslations({ locale, namespace: "common" })`. `getMdMetadata` already passed locale through, so those helpers are safe without priming. A separate bare next-intl call in the same entry is not.

## Fixed (14 pages)

Missing entirely in the `Page` body — `bug-bounty`, `learn`, `what-is-ethereum`, `what-is-the-ethereum-network`, `use-cases`, `founders`, `community/events`, `community/events/meetups`, `community/events/conferences`

Called after the first next-intl API — `stablecoins`, `collectibles`, `start`, `get-eth`, `wallets/find-wallet`

The three `community/events` pages are latent rather than live: they already render dynamic (they fetch event data), so the header read is legal and they 404 today. No per-route PR covered them because they never alerted.

## Lint rule

`local/set-request-locale-first` (`.eslint-rules/`) is scoped to those route files in `.eslintrc.json`. It analyzes the default export and exported `generateMetadata` / `generateViewport` separately, follows same-file helpers and known eager callbacks/IIFEs, requires a top-level `setRequestLocale` with a real argument, and only treats an explicit (normal or quoted) `locale` property as safe. It is not a cross-module call graph.

## Not changed

`app/not-found.tsx` deliberately stays on explicit-locale APIs. In dev, Next.js renders it alongside every page request in the same request-scoped cache, so calling `setRequestLocale` there poisons the locale of the page actually being rendered. Its existing comment says so; the doc now records it too. The glob still covers localized `not-found.tsx` files; the root file stays valid because it already uses `{ locale: DEFAULT_LOCALE }`.

`page-jsonld.tsx` siblings and `_components/*` trees call bare `getTranslations` by design — they are descendants and resolve from the cache their page primes.

## Relationship to #18999

#18999 moves `setRequestLocale` above the `notFound()` bail in `app/[locale]/layout.tsx`. It is **complementary, not redundant**. next-intl is explicit:

> You need to call this function in every page and every layout that you intend to enable static rendering for since Next.js can render layouts and pages independently.

The layout cannot discharge the per-page obligation, so this PR is the contractual fix and #18999 is defence-in-depth for pages authored without their own call. Both should land.

## Supersedes

#18932, #18994, #19003, #19005, #19006, #19007, #19008, #19009, #19010, #19011, #19012

## Verification

- Re-ran the AST audit against the page sweep — zero findings
- `local/set-request-locale-first` unit tests plus an integration test that loads the actual ESLint config
- `pnpm type-check` passes
- `pnpm lint` clean
- Doc updated: `docs/solutions/architecture/setrequestlocale-static-to-dynamic-rendering.md`


.eslint-rules/set-request-locale-first.js
```diff
@@ -0,0 +1,191 @@
...
+
+// No override parameter, so it always resolves the request locale.
+const ALWAYS = new Set(["getLocale"])
+
+// Project helpers that reach next-intl internally with no way to pass a locale
+// through, so the caller must still prime the request locale. `getMetadata`
+// accepts a `locale` but never forwards it to its own `getTranslations` call.
+const RISKY_HELPERS = new Map([
+  ["@/lib/utils/metadata", new Set(["getMetadata"])],
+])
```
wackerow:
```suggestion
// Project helpers that reach next-intl internally with no way to pass a locale
// through, so the caller must still prime the request locale. Both accept a
// `locale` but neither forwards it to the `getTranslations` call underneath
// (`getMdMetadata` delegates to `getMetadata`).
const RISKY_HELPERS = new Map([
  ["@/lib/utils/metadata", new Set(["getMetadata"])],
  ["@/lib/md/metadata", new Set(["getMdMetadata"])],
])
```

.eslintrc.json
```diff
@@ -78,6 +83,20 @@
...
+    },
+    {
+      // Next.js route entry points only. Descendants (page-jsonld, _components)
+      // inherit the locale the entry primes, so the rule must not apply there.
+      "files": [
+        "app/**/page.tsx",
+        "app/**/layout.tsx",
+        "app/**/template.tsx",
+        "app/**/default.tsx",
+        "app/**/opengraph-image.tsx"
```
wackerow, line 97:
```suggestion
        "app/**/opengraph-image.tsx",
        "app/**/not-found.tsx"
```

.eslint-rules/set-request-locale-first.js
```diff
@@ -0,0 +1,191 @@
...
+
+// Safe when passed an explicit `locale`: next-intl's getConfig then skips the
+// header read entirely.
+const OPTIONAL_LOCALE = new Set([
+  "getTranslations",
+  "getMessages",
+  "getFormatter",
+  "getNow",
+  "getTimeZone",
+])
```
wackerow, line 20: Appears to be a mirror of next-intl's export surface? Would drift silently, consider flipping to an exemption list:

```suggestion
// Treat every next-intl/server export as locale-resolving, exempting only
// these. Opt-out rather than opt-in: an API next-intl adds later is flagged
// (loud, one-line fix) instead of silently unchecked.
const EXEMPT = new Set(["setRequestLocale", "getRequestConfig"])
```

This would be in tandem with the suggestion below to the `isRisky` function

.eslint-rules/set-request-locale-first.js
```diff
@@ -0,0 +1,191 @@
...
+
+      "Program:exit"() {
+        if (defaultExport && byName.has(defaultExport))
+          entries.add(byName.get(defaultExport))
+
+        const isRisky = (c) =>
+          ALWAYS.has(c.api) ||
+          RISKY_HELPER_NAMES.has(c.api) ||
+          (OPTIONAL_LOCALE.has(c.api) &&
+            !hasExplicitLocale(c.node.arguments[0]))
```
wackerow: With `EXEMPT` suggestion above:

```suggestion
        const isRisky = (c) =>
          RISKY_HELPER_NAMES.has(c.api) ||
          (!EXEMPT.has(c.api) && !hasExplicitLocale(c.node.arguments[0]))
```

wackerow (changes requested):
@pettinarip looks good! just a couple small suggestions to round this out, then good to pull in

.eslint-rules/set-request-locale-first.js
pettinarip: Good catch on the old helper. `getMetadata` now forwards `locale` into `getTranslations`, and `getMdMetadata` already passed it through, so I dropped the special-case list instead of growing it.

.eslintrc.json
pettinarip: Added, plus `loading.tsx`. Root `app/not-found.tsx` still uses an explicit locale, so it stays valid.

.eslint-rules/set-request-locale-first.js
pettinarip: I kept the two lists so `getLocale` stays always flagged — a `{ locale }` argument does not skip the header read. To cover the drift concern I added `getExtracted`, which is the same shape as `getTranslations` and was missing.

.eslint-rules/set-request-locale-first.js
pettinarip: That helper is gone now, and bringing `RISKY_HELPER_NAMES` back would undo the `getMetadata` change.

wackerow (approved):

amannn:
Hey @pettinarip and @wackerow,

not sure if you've already seen it, but I've set up a PR that switches to the recently-released `next/root-params` API: https://github.com/ethereum/ethereum-org-website/pull/19135

That avoids `setRequestLocale` in the first place and could be a useful change in your setup!

PR #19017, opened 2026-08-07
fix(seo): drop transcript from VideoObject

### LINKED ISSUE #18996: fix(seo): drop transcript from VideoObject (unsupported by Google, values corrupted, contradicts inLanguage)
## Summary

Remove the `transcript` property from `VideoObject` in `app/[locale]/videos/[slug]/page-jsonld.tsx`. It is unsupported by Google, its values are currently corrupted, and it creates a language contradiction. Deleting it resolves all three at once.

The transcript stays where it belongs: rendered in the page body, which is what actually gets indexed.

## Reason 1 — Google does not support it

`transcript` appears in neither Google's required nor recommended property list for `VideoObject`:

- Required: `name`, `thumbnailUrl`, `uploadDate` (all three correctly emitted today)
- Recommended: `contentUrl`, `description`, `duration`, `embedUrl`, `expires`, `hasPart`, `ineligibleRegion`, `interactionStatistic`, `publication`, `regionsAllowed`

So the property earns nothing in Google Search. It is 88% of the JSON-LD block:

| Page | transcript chars | JSON-LD block | transcript share |
|---|---|---|---|
| `/videos/ethereum-privacy-stack-andy-guzman/` | 23,558 | 26,853 B | 88% |
| `/es/videos/…` | 25,892 | 29,315 B | 88% |
| `/pt-br/videos/…` | 25,637 | 29,056 B | 88% |
| `/zh/videos/…` | 8,960 | 11,940 B | 75% |

Roughly 23KB of redundant bytes per page, across 67 videos in 25 locales.

## Reason 2 — the values are corrupted

```js
transcript: transcript
  ? stripMarkdown(transcript, true)
      .replace(/</g, "\\u003c")
      .replace(/\//g, "\\u002f")
```

In a JS string literal, `"\\u002f"` is the six literal characters `/`, not `/`. `JSON.stringify` then escapes the backslash again, so a consumer parsing the JSON reads literal `/` text rather than a slash. Verified across four locales:

```
transcript chars=23558   literal '/' occurrences=1   real '/' characters=0
```

Example: `...you can go to pse.dev/research.` Every slash in every transcript is mangled, so URLs, dates and fractions in spoken content all break.

Not a security hole, incidentally: since no real `<` survives either, a `</script>` breakout is impossible. The XSS intent is met by accident. This is content corruption only.

## Reason 3 — the language declaration contradicts itself

Transcripts **are** translated per locale, but `VideoObject.inLanguage` reads `frontmatter.lang`, which is `en` on every locale:

| URL | `WebPage.inLanguage` | `VideoObject.inLanguage` | transcript language |
|---|---|---|---|
| `/videos/…` | en | en | English |
| `/es/videos/…` | es | **en** | **Spanish** |
| `/zh/videos/…` | zh | **en** | **Chinese** |
| `/pt-br/videos/…` | pt-br | **en** | **Portuguese** |

Declaring `en` is defensible for the audio track, but it contradicts the translated transcript attached to the same node.

There is no clean way to fix this in place: schema.org's `transcript` expects **Text** only, is used on `AudioObject`/`VideoObject`, and provides **no mechanism to declare a transcript's language**. Language is already correctly declared everywhere it can be, via `WebPage.inLanguage` (which varies per locale) and hreflang.

## Not a duplicate-content concern

Worth recording, since this comes up: the transcript appearing both in the rendered body and in the JSON-LD is not a penalty. Markup is *supposed* to reflect visible content, and I verified the transcript is fully server-rendered (probes at 25%, 50% and 75% of the transcript all found in the body, in both English and Spanish). Translated transcripts across locales are not duplicate content either, and hreflang is correct. The only real cost is bytes.

## Suggested change

Delete the `transcript` property and its two `.replace()` calls. Keep everything else on the node.

If a transcript signal is wanted later, `subtitleLanguage` is the schema.org property that can carry language information, though Google does not list it as supported either.

## Acceptance criteria

- [ ] `VideoObject` JSON-LD contains no `transcript` property
- [ ] Transcript still renders in the page body in every locale
- [ ] Required `VideoObject` properties (`name`, `thumbnailUrl`, `uploadDate`) unchanged
- [ ] JSON-LD block size on video pages drops by roughly 75 to 88%

patrick-star-10:
## Description

Closes #18996.

This PR removes the `transcript` property from the `VideoObject` JSON-LD emitted by video detail pages.

Google does not list `transcript` as a supported property for video rich results. In addition, the existing escaping converted every slash in transcript content into literal `\u002f` text after JSON serialization, corrupting URLs, dates, and fractions. On localized pages, attaching a translated transcript to a `VideoObject` whose `inLanguage` describes the original audio also made the node internally ambiguous.

The transcript remains in the server-rendered page body, where it is visible to users and available for indexing.

## Changes

- Removed `transcript` from the video detail page's `VideoObject` JSON-LD node.
- Removed the now-unused `transcript` prop from `VideoPageJsonLD`.
- Removed the now-unused `stripMarkdown` import and transcript escaping logic.
- Stopped passing transcript Markdown into the JSON-LD component.
- Added Playwright regression coverage for English and Spanish versions of the example video from the issue.

The regression test parses the emitted JSON-LD and verifies that:

- the `VideoObject` has no `transcript` property;
- `name`, `thumbnailUrl`, and `uploadDate` remain populated; and
- localized transcript text still renders in the page body.

No transcript content, translations, accordion behavior, or other `VideoObject` fields were changed.

## JSON-LD size measurements

Measured from the `application/ld+json` script containing the `VideoObject` while running the site locally:

| Page | Before | After | Reduction |
| --- | ---: | ---: | ---: |
| `/videos/ethereum-privacy-stack-andy-guzman/` | 26,853 bytes | 3,124 bytes | 88.4% |
| `/es/videos/ethereum-privacy-stack-andy-guzman/` | 29,820 bytes | 3,258 bytes | 89.1% |

## Testing

- TDD regression check before the implementation: `corepack pnpm exec playwright test tests/e2e/videos-jsonld.spec.ts --project=e2e --workers=1` — failed in both locale cases specifically because `VideoObject.transcript` was still present.
- `corepack pnpm lint` — passed with no ESLint errors.
- `corepack pnpm type-check` — passed Next.js route type generation, application TypeScript checks, and Netlify edge-function TypeScript checks.
- `corepack pnpm test:unit` — 990 passed, 1 skipped, 0 failed.
- `corepack pnpm exec playwright test tests/e2e/videos-jsonld.spec.ts --project=e2e --workers=1` — 2 passed, 0 failed.

### DIFF
--- a/app/[locale]/videos/[slug]/page-jsonld.tsx
+++ b/app/[locale]/videos/[slug]/page-jsonld.tsx
@@ -4,7 +4,6 @@ import type { VideoFrontmatter } from "@/lib/interfaces"
 
 import PageJsonLD from "@/components/PageJsonLD"
 
-import { stripMarkdown } from "@/lib/utils/md"
 import { toIsoDuration } from "@/lib/utils/time"
 import { normalizeUrlForJsonLd } from "@/lib/utils/url"
 
@@ -16,12 +15,10 @@ export default async function VideoPageJsonLD({
   locale,
   slug,
   frontmatter,
-  transcript,
 }: {
   locale: string
   slug: string
   frontmatter: VideoFrontmatter
-  transcript: string | null
 }) {
   const url = normalizeUrlForJsonLd(locale, `/videos/${slug}/`)
   const videoGalleryUrl = normalizeUrlForJsonLd(locale, "/videos/")
@@ -101,13 +98,6 @@ export default async function VideoPageJsonLD({
         publisher: REFERENCE.ETHEREUM_FOUNDATION,
         isAccessibleForFree: true,
         isFamilyFriendly: true,
-        // Add transcript as plain text if available
-        transcript: transcript
-          ? stripMarkdown(transcript, true)
-              // Escape < and / to prevent script injection (XSS protection)
-              .replace(/</g, "\\u003c")
-              .replace(/\//g, "\\u002f")
-          : undefined,
       },
     ],
   }
--- a/app/[locale]/videos/[slug]/page.tsx
+++ b/app/[locale]/videos/[slug]/page.tsx
@@ -49,12 +49,7 @@ const VideoLandingPage = async (props: {
 
   return (
     <>
-      <VideoPageJsonLD
-        locale={locale}
-        slug={slug}
-        frontmatter={frontmatter}
-        transcript={transcriptMdx}
-      />
+      <VideoPageJsonLD locale={locale} slug={slug} frontmatter={frontmatter} />
 
       <main className="max-w-4xl p-page pt-hero lg:pt-hero-2x">
         <MainArticle className="flow">
--- a/tests/e2e/videos-jsonld.spec.ts
+++ b/tests/e2e/videos-jsonld.spec.ts
@@ -1,0 +1,63 @@
+import { expect, Page, test } from "@playwright/test"
+
+const VIDEO_SLUG = "ethereum-privacy-stack-andy-guzman"
+
+interface JsonLdNode {
+  "@type"?: string
+  "@graph"?: JsonLdNode[]
+  [key: string]: unknown
+}
+
+const getVideoObject = async (page: Page) => {
+  const documents = await page
+    .locator('script[type="application/ld+json"]')
+    .evaluateAll((scripts) =>
+      scripts.map((script) => JSON.parse(script.textContent || "{}"))
+    )
+
+  const nodes = (documents as JsonLdNode[]).flatMap(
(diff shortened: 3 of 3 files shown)


patrick-star-10:
Confirmed — removing `transcript` from the `VideoObject` JSON-LD is intentional. Google does not use this property for video rich results, and the existing values add substantial payload size while creating content and language inconsistencies. The transcript remains server-rendered in the page body, with E2E coverage verifying that behavior in both English and Spanish.

PR #19022, opened 2026-08-10
fix: localize wallets and quizzes JSON-LD

### LINKED ISSUE #18979: fix(seo): raw i18n keys leaking into JSON-LD on /wallets/ and /quizzes/ (recurrence of #17719)
## Describe the bug

Two pages emit untranslated i18n message keys as JSON-LD text, live in production:

```
/wallets/   WebPage.name        = "page-wallets-meta-title"
/wallets/   Article.headline    = "page-wallets-title"
/quizzes/   WebPage.description = "quizzes-subtitle"
```

`Article.headline` is the single most important text node in the graph, and on `/wallets/` it currently reads `page-wallets-title` to every crawler.

**This is a recurrence of #17719**, fixed on `/wallets/find-wallet/` by #17721. That fix corrected a wrong `getTranslations()` namespace. The same wrong namespace string was left in the sibling file and has been live since.

## To reproduce

```bash
curl -s https://ethereum.org/wallets/ | grep -o '"headline":"[^"]*"'
# -> "headline":"page-wallets-title"

curl -s https://ethereum.org/quizzes/ | grep -o '"description":"quizzes-subtitle"'
# -> "description":"quizzes-subtitle"
```

Verified against production on 2026-08-05.

## Expected behavior

JSON-LD contains translated strings, in every locale.

## Root cause

`getTranslations()` is called with a namespace that has no corresponding file in `src/intl/en/`, so `next-intl` echoes the key back.

I scanned all 54 `page-jsonld.tsx` files against the real namespace files. **Exactly two are broken:**

| File | Line | Uses | Should be | Keys actually live in |
|---|---|---|---|---|
| `app/[locale]/wallets/page-jsonld.tsx` | 15 | `"page-find-wallet"` | `"page-wallets"` | `src/intl/en/page-wallets.json` |
| `app/[locale]/quizzes/page-jsonld.tsx` | 19 | `"page-quizzes"` | `"learn-quizzes"` | `src/intl/en/learn-quizzes.json` |

Note that `"page-find-wallet"` on line 15 is the *same wrong string* that #17721 removed from `wallets/find-wallet/page-jsonld.tsx`.

Reproduce the scan:

```bash
find app -name "page-jsonld.tsx" -print0 | while IFS= read -r -d '' f; do
  grep -oE 'getTranslations\("[^"]+"' "$f" | sed 's/getTranslations("//; s/"$//' | sort -u \
  | while IFS= read -r n; do
      [ -f "src/intl/en/$n.json" ] || echo "BAD | ${f#app/} | \"$n\""
    done
done
```

## Suggested fix

Two one-word changes. Correct the namespace in each file.

## Why this bug class keeps surviving

It is invisible in the browser. `<title>` and `<meta>` are produced separately and are correct, so the page looks fine; only the JSON-LD block is broken. That is exactly why #17719 went unnoticed, and why two instances survived its fix.

A guard is proposed separately: a CI assertion that no JSON-LD string value matches an i18n-key shape would make this class of bug impossible to ship a third time.

## Acceptance criteria

- [ ] No JSON-LD value on `/wallets/` or `/quizzes/` matches a `page-*` / `*-subtitle` key shape
- [ ] Localized variants (e.g. `/es/wallets/`, `/ja/quizzes/`) emit translated text
- [ ] The namespace scan above reports no BAD entries

pillowtalk-Qy:
## Description

Fix the translation namespaces used by the wallets and quizzes JSON-LD components so structured data contains localized copy instead of raw translation keys.

- use `page-wallets` for `/wallets/` JSON-LD
- use `learn-quizzes` for `/quizzes/` JSON-LD
- add browser-level regression coverage for English and non-English pages

Closes #18979.

## Risk

Low. This changes only the message namespaces used to populate existing JSON-LD fields. It does not change visible page content, routes, schemas, or translation files. The regression test parses the rendered `application/ld+json` and compares it with the source locale messages, including Spanish wallets and Japanese quizzes to detect accidental English fallback.

## Test evidence

- Before the fix, the new test fails against production in all four cases:
  - wallet `name`: `page-wallets-meta-title` (English and Spanish)
  - quiz `description`: `quizzes-subtitle` (English and Japanese)
- `pnpm lint`
- `pnpm type-check`
- `pnpm test:unit` (990 passed, 1 skipped)
- `USE_MOCK_DATA=true IS_VISUAL_TEST=true NEXT_PUBLIC_BUILD_LOCALES=en,es,ja,ar pnpm build` (1605 pages generated; 9 pre-existing Codeblock warnings)
- JSON-LD E2E against the local production build across desktop/mobile Chromium and WebKit (16 passed)
- Netlify deploy preview for final SHA `1a86a971fd` passed, including Header rules, Redirect rules, GitGuardian, and Labeler
- JSON-LD E2E against `https://deploy-preview-19022.ethereum.it` across desktop/mobile Chromium and WebKit (16 passed)

## Rollback

Revert this PR to restore the previous namespace lookups. No data migration or content rollback is required.

## Unresolved items

- Full repository CI still requires maintainer approval for workflows from this fork.
- The 9 existing Codeblock build warnings are unrelated and intentionally not changed here.

### DIFF
--- a/tests/e2e/jsonld.spec.ts
+++ b/tests/e2e/jsonld.spec.ts
@@ -1,0 +1,62 @@
+import { expect, Page, test } from "@playwright/test"
+
+import enQuizzes from "@/intl/en/learn-quizzes.json"
+import enWallets from "@/intl/en/page-wallets.json"
+import esWallets from "@/intl/es/page-wallets.json"
+import jaQuizzes from "@/intl/ja/learn-quizzes.json"
+
+type JsonLdNode = {
+  "@type"?: string
+  name?: string
+  headline?: string
+  description?: string
+}
+
+async function getJsonLdNode(page: Page, path: string, type: string) {
+  await page.goto(path)
+
+  const scripts = await page
+    .locator('script[type="application/ld+json"]')
+    .allTextContents()
+  const nodes = scripts.flatMap((script) => {
+    const data = JSON.parse(script)
+    return Array.isArray(data["@graph"]) ? data["@graph"] : []
+  }) as JsonLdNode[]
+
+  const node = nodes.find(({ "@type": nodeType }) => nodeType === type)
+  expect(node, `${type} JSON-LD node on ${path}`).toBeDefined()
+
+  return node!
+}
+
+test.describe("Localized JSON-LD", () => {
+  for (const { locale, path, messages } of [
+    { locale: "English", path: "/wallets/", messages: enWallets },
+    { locale: "Spanish", path: "/es/wallets/", messages: esWallets },
+  ]) {
+    test(`uses ${locale} wallet translations`, async ({ page }) => {
+      const webPage = await getJsonLdNode(page, path, "WebPage")
+      const article = await getJsonLdNode(page, path, "Article")
+
+      expect(webPage.name).toBe(messages["page-wallets-meta-title"])
+      expect(webPage.description).toBe(
+        messages["page-wallets-meta-description"]
+      )
+      expect(article.headline).toBe(messages["page-wallets-title"])
+      expect(article.description).toBe(
+        messages["page-wallets-meta-description"]
+      )
+    })
+  }
+
+  for (const { locale, path, messages } of [
+    { locale: "English", path: "/quizzes/", messages: enQuizzes },
+    { locale: "Japanese", path: "/ja/quizzes/", messages: jaQuizzes },
+  ]) {
+    test(`uses ${locale} quiz translations`, async ({ page }) => {
+      const webPage = await getJsonLdNode(page, path, "WebPage")
+
+      expect(webPage.description).toBe(messages["quizzes-subtitle"])
+    })
+  }
+})
--- a/tests/e2e/jsonld.spec.ts
+++ b/tests/e2e/jsonld.spec.ts
@@ -1,15 +1,17 @@
 import { expect, Page, test } from "@playwright/test"
 
-import enQuizzes from "@/intl/en/learn-quizzes.json"
-import enWallets from "@/intl/en/page-wallets.json"
-import esWallets from "@/intl/es/page-wallets.json"
-import jaQuizzes from "@/intl/ja/learn-quizzes.json"
-
 type JsonLdNode = {
   "@type"?: string
   name?: string
   headline?: string
   description?: string
+}
+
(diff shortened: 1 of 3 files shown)


tests/e2e/jsonld.spec.ts
```diff
@@ -0,0 +1,62 @@
+import { expect, Page, test } from "@playwright/test"
+
+import enQuizzes from "@/intl/en/learn-quizzes.json"
+import enWallets from "@/intl/en/page-wallets.json"
+import esWallets from "@/intl/es/page-wallets.json"
+import jaQuizzes from "@/intl/ja/learn-quizzes.json"
```
wackerow: These locale JSON imports pull ~324K to read one key each, and since `src/i18n/request.ts` merges en under every locale, a missing `es`/`ja` key compares rendered English against `undefined`... e2e goes red for a non-code reason. Would assert shape instead:

```ts
const I18N_KEY = /^[a-z0-9]+(-[a-z0-9]+){2,}$/
expect(webPage.name).not.toMatch(I18N_KEY)
```

Drops the imports and the coupling, and generalizes past these two pages

wackerow (changes requested):
Thanks @pillowtalk-Qy! Looks good overall but left one request

tests/e2e/jsonld.spec.ts
pillowtalk-Qy: Addressed in `360caad4bb`:

- removed all four locale JSON imports and the exact-copy comparisons
- kept the EN/ES wallet and EN/JA quiz scenarios
- added a shared shape assertion that requires a string and rejects raw kebab-case i18n keys

I used `^[a-z0-9]+(?:-[a-z0-9]+)+$` rather than `{2,}` for the repeated segment because the production regression value `quizzes-subtitle` has only one hyphen and must still be caught.

Verification on the updated test:

- PR Preview across desktop/mobile Chromium and WebKit: 16/16 passed
- current production, Desktop Chromium: all four cases failed as expected on `page-wallets-meta-title` / `quizzes-subtitle`
- `pnpm lint`: passed
- `pnpm type-check`: passed

@wackerow, please re-review when convenient.

wackerow (approved):
Lgtm, thanks @pillowtalk-Qy!

wackerow:
@all-contributors please add @pillowtalk-Qy for bug fix

PR #19028, opened 2026-08-10
fix(seo): add accurate sitemap modification dates

### LINKED ISSUE #18982: fix(seo): sitemap has zero <lastmod> across ~17.5k URLs and publishes an internal @modal route (25 x 404)
## Describe the bug

Two separate sitemap problems.

### 1. No `<lastmod>` on any URL

```bash
curl -s https://ethereum.org/sitemaps/sitemap/en.xml | grep -c "<loc>"      # -> 897
curl -s https://ethereum.org/sitemaps/sitemap/en.xml | grep -c "<lastmod>"  # -> 0
```

Zero `<lastmod>` across all 25 locale shards (~17.5k URLs). There is also no `Last-Modified` response header (`cf-cache-status: DYNAMIC`), so the site emits no declarative recrawl hint at all.

Google uses sitemap `lastmod` for crawl scheduling when it is consistently accurate. For a site this size across 25 locales, that is the main lever available for crawl efficiency.

The data is already in process: `getStaticGitHubContributors()` / `getAppPageLastCommitDate()` supply a real git commit date, currently used only for the "last edited" display string.

### 2. A Next.js internal route is published as a public URL

```bash
curl -s https://ethereum.org/sitemaps/sitemap/en.xml | grep "@modal"
# <loc>https://ethereum.org/developers/tools/@modal/(.)[tool]/</loc>
```

Present in all **25** locale shards, and it 404s:

```bash
curl -s -o /dev/null -w "%{http_code}\n" "https://ethereum.org/developers/tools/%40modal/(.)%5Btool%5D/"
# -> 404
```

Also one genuinely dead content URL: `/apps/scout-game/` returns 404. (A full sweep of all 897 EN URLs found 895×200, 1×404, 0 redirects, so coverage is otherwise healthy.)

## Expected behavior

- Sitemap entries carry an accurate `lastmod`
- Only routable public URLs appear in the sitemap

## Root cause

**Missing `lastmod`:** `app/sitemaps/sitemap.ts:67` pushes `entries.push({ url, alternates })` and never sets `lastModified`.

**Invalid route:** `src/lib/utils/staticPages.ts` skips directories starting with `[` (dynamic routes) and `_` (private folders), but not Next.js parallel-route slots (`@modal`) or intercepting routes (`(.)`):

```js
if (entry.name.startsWith("[")) continue
if (entry.name.startsWith("_")) continue
// no guard for "@" or "("
```

## Suggested fix

1. Set `lastModified` on sitemap entries from the existing commit-date source.
2. Add `@` and `(` to the skipped-prefix list in `staticPages.ts`. One line, removes 25 invalid URLs.
3. Remove `/apps/scout-game/` from the apps data feeding the sitemap.

## Related, same data source

Thirteen `page-jsonld.tsx` files pass `lastEditLocaleTimestamp` into `Article.dateModified`, which is human display output from `Intl.DateTimeFormat` rather than ISO 8601:

```
/what-is-ethereum/         "March 31, 2025"
/pt-br/what-is-ethereum/   "31 de março de 2025"
/es/what-is-ether/         "15 de julio de 2026"
```

Schema.org requires ISO 8601, so Google discards these. The ISO value (`latestCommitDate`) exists one line above the formatting call in `src/lib/utils/contributors.ts:67`. Separately, of 21 files emitting `Article`, zero emit `datePublished` on the page-level node.

Worth fixing in the same PR since it is the same underlying date. Impact is low on its own: `dateModified` is a recommended not required property, and the visible date is already in the rendered body where Google reads it. Flagging so nobody scopes it as a ranking fix.

## Acceptance criteria

- [ ] `<lastmod>` present and accurate on sitemap entries
- [ ] `grep "@modal"` on every locale shard returns nothing
- [ ] No 404s in a full sitemap sweep
- [ ] `Article.dateModified` values match `^\d{4}-\d{2}-\d{2}`

Suede-AI:
## Description

Fixes #18982.

This adds accurate ISO 8601 `lastModified` values to sitemap entries using the existing GitHub contributor data, removes internal Next.js route segments and the dead Scout Game app URL from sitemap discovery, and passes the ISO commit date—not the localized display date—to the 13 affected Article JSON-LD components.

## Changes

- add commit-derived `lastModified` to content and app-page sitemap records
- exclude `[`, `_`, `@`, and `(` route segments from static route discovery
- exclude `/apps/scout-game/` from dynamic app routes
- expose `latestCommitDate` alongside the localized visible timestamp
- use ISO dates for Article `dateModified`
- add unit coverage for route exclusion and latest-date selection

## Validation

- `USE_MOCK_DATA=true pnpm playwright test --project=unit tests/unit/utils/static-pages.spec.ts tests/unit/utils/gh.spec.ts` (3 passed)
- ESLint on all changed TypeScript files
- Prettier on all changed TypeScript files

The full workspace typecheck was not usable in the local partial clone because unrelated modules and image assets are intentionally absent; CI has the complete tree.

### DIFF
--- a/app/[locale]/ethereum-history-founder-and-ownership/page-jsonld.tsx
+++ b/app/[locale]/ethereum-history-founder-and-ownership/page-jsonld.tsx
@@ -11,11 +11,11 @@ import { REFERENCE } from "@/lib/jsonld/references"
 
 export default async function EthereumHistoryFounderAndOwnershipPageJsonLD({
   locale,
-  lastEditLocaleTimestamp,
+  latestCommitDate,
   contributors,
 }: {
   locale: Lang | undefined
-  lastEditLocaleTimestamp: string
+  latestCommitDate: string
   contributors: FileContributor[]
 }) {
   const t = await getTranslations("page-ethereum-history-founder-and-ownership")
@@ -104,7 +104,7 @@ export default async function EthereumHistoryFounderAndOwnershipPageJsonLD({
               "The historical development and launch of the Ethereum blockchain network",
           },
         ],
-        dateModified: lastEditLocaleTimestamp,
+        dateModified: latestCommitDate,
       },
     ],
   }
--- a/app/[locale]/ethereum-history-founder-and-ownership/page.tsx
+++ b/app/[locale]/ethereum-history-founder-and-ownership/page.tsx
@@ -27,7 +27,7 @@ const Page = async ({ params }: { params: Promise<{ locale: Lang }> }) => {
 
   const t = await getTranslations("page-ethereum-history-founder-and-ownership")
 
-  const { contributors, lastEditLocaleTimestamp } =
+  const { contributors, lastEditLocaleTimestamp, latestCommitDate } =
     await getAppPageContributorInfo(
       "ethereum-history-founder-and-ownership",
       locale as Lang
@@ -66,7 +66,7 @@ const Page = async ({ params }: { params: Promise<{ locale: Lang }> }) => {
     <>
       <PageJsonLD
         locale={locale}
-        lastEditLocaleTimestamp={lastEditLocaleTimestamp}
+        latestCommitDate={latestCommitDate}
         contributors={contributors}
       />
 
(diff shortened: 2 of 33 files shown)


Suede-AI:
Superseded by #19029 from the requested contributor account.

PR #18741, opened 2026-07-09
Update/staking content revamp

mnelsonBT:
### Staking pillar content updates

- Updated /staking/ hub per the outline: 
-- Add role-based benefit cards
-- Add trust-spectrum framing on the staking-options hierarchy
-- Add a new comparison table covering all five staking approaches (solo / bonded node operation / delegated / pools / CEX)
-- Post-Pectra FAQ refresh (page's FAQ structured data updated to match)

- Updated /staking/solo:
-- Frame home staking as direct protocol participation
-- Add Pectra compounding/2048 ETH economics and post-Fusaka hardware guidance
-- Add new sections on DVT squad staking, bonded node operation for protocols (Lido CSM, Rocket Pool), restaking clarification, and non-staking node use

- Updated /staking/saas: 
-- Reframe /staking/saas as "Delegated staking" (URL unchanged)
-- Adds delegation-spectrum structure from non-custodial SaaS through custodial exchange staking
-- Adds a new trust-model evaluation section built around withdrawal-key custody and post-Pectra exit rights

- Updated /staking/pools: 
-- Restructured /staking/pools as "Liquid & pooled staking" (URL unchanged) around a transparency axis risk framework

- Other:
-- Brought /staking/dvt into the staking pillar (staking template/hero/nav), updated the technology explanation, and added the production landscape (Obol, SSV) with per-audience guidance and a new FAQ
-- Made minor updates to /staking/withdrawals terminology plus a new callout on pool/LST redemption vs. protocol withdrawals

### Supporting updates:

- Aligned hardware guidance on /run-a-node/ and the developer docs node-running page with current EIP-7870 recommendations, resolving disagreement with the new staking content
- Updated staking quiz answers for post-Pectra accuracy (compounding credentials, partial withdrawals)
- Added the missing "liquid staking tokens" glossary entry and tooltip (fixing a broken glossary link/tooltip)
- Corrected navigation, cross-links, and topic-dropdown labels to match the renamed pages

PR #19030, opened 2026-08-10
feat: staking pages redesign

myelinated-wackerow:
## Summary

Rebuilds the /staking landing page and its subpages (solo, saas, pools, withdrawals, dvt) from the Figma redesign: four colored staking-option boxes replacing the old hierarchy diagram, a new at-a-glance comparison table covering bonded node operation and custodial exchanges, and rewritten copy framing the options from protocol-native to most abstracted. Considerations move into a TabNav, alert and card glyphs move from emoji to lucide icons throughout, StakingCommunityCallout renders from the markdown rather than being injected by the topic layout, and /staking/dvt/ declares `template: staking` so it picks up the shared staking dropdown instead of the docs-style ToC.

Supersedes #18741, which will close in favor of this branch — its content pass is incorporated here: FAQ questions and answers across every staking page, the rewritten dvt page (DVT in production, DKG and middleware detail, four FAQ entries plus a new one clarifying that cluster sizes are configurable rather than fixed at 3 of 4), and current hardware and bandwidth guidance on run-a-node. Review feedback from that PR is applied too: EIP references now lead with a name and cite the number parenthetically, each page's `[Ethereum](/)` link sits on its first plain mention, and the `emoji` and `alt` frontmatter keys are gone (nothing reads either; the hero images are decorative and dvt has none).

Three silent bugs turned up along the way and are fixed here: glossary tooltips on the landing page rendered raw message keys (an app-router page needs its own `I18nProvider`, and a nested one replaces rather than merges the namespaces it is given), `cn()` was dropping `text-h*` sizes because tailwind-merge classified them as text colors, and a horizontally scrollable tab row shoved the ToC off-screen just above `lg` on every staking page. The FAQ structured data was also incomplete, declaring five questions where the page renders eight.

Two things to know before merging. The `page-staking` namespace is fully migrated off the deprecated `Translation` component to `t`/`t.rich`, with link hrefs moved out of the catalog strings into tag handlers — so **the intl pipeline needs to re-translate this namespace before launch**, or the retagged strings render key tails in the other locales. And the shared-primitive changes are isolated in their own commits for easier review: an `AlertIcon` `size` variant (with `AlertEmoji` deprecated), the tailwind-merge fix, callout banner padding, and a one-shade darker `--warning` in light mode.

## Preview link
https://deploy-preview-19030.ethereum.it/staking


wackerow:
pinging @konopkja and @mnelsonBT for review

Noting the hero/Leslie images are not updated... not sure if we wanted to include that in this PR or not. cc: @nloureiro

src/components/Staking/StakingLaunchpadWidget.tsx
```diff
@@ -1,98 +1,88 @@
 "use client"
```
pettinarip, line 1: we could probably turn this into a sc now
```suggestion
```

src/lib/utils/cn.ts
```diff
@@ -17,6 +20,7 @@ const twMerge = extendTailwindMerge({
           ],
         },
       ],
+      "font-size": [{ text: [(value: string) => /^h[1-6]$/.test(value)] }],
```
pettinarip, line 23: 👍🏼

src/intl/en/page-staking.json
```diff
@@ -165,60 +171,101 @@
...
   "page-staking-faq-5-answer-p2": "Stakers will also earn rewards in the form of fees and MEV when proposing blocks, which are made available immediately via the set fee recipient address.",
   "page-staking-faq-5-answer-link": "More on staking withdrawals",
+  "page-staking-faq-6-question": "How much ETH do I need to stake?",
+  "page-staking-faq-6-answer-p1": "As little as 0.01 ETH through a staking pool, or at least 32 ETH to run your own validator. A single validator can hold up to 2048 ETH: with compounding (0x02) withdrawal credentials, rewards are automatically added to your stake and you earn rewards on every whole ETH above the 32 ETH minimum.",
+  "page-staking-faq-6-answer-p2": "You can participate with less. Bonded node operation lets you run validators on your own hardware for a staking protocol with a bond of roughly 1.5–4 ETH, and staking pools accept almost any amount, some as little as 0.01 ETH.",
```
pettinarip, line 251: `p1` already gives the 0.01 ETH pool floor, so "You can participate with less" reads as less than that, and the closing clause repeats it. The only new info here is the 1.5–4 ETH bond.

app/[locale]/staking/page.tsx
```diff
@@ -65,35 +102,333 @@ const Page = async (props: { params: Promise<PageParams> }) => {
...
+          ),
+          link2: (chunks) => (
+            <TooltipLink href="/glossary/#consensus-client">
+              {chunks}
+            </TooltipLink>
+          ),
+        }),
+        t.rich("page-staking-section-comparison-solo-requirements-li3", {
+          a: (chunks) => (
+            <InlineLink href="https://hoodi.launchpad.ethereum.org">
```
Comment (pettinarip) [inline app/[locale]/staking/page.tsx:298]: This one's hardcoded, but `StakingLaunchpadWidget` builds the same URL from `CANONICAL_STAKING_TESTNET`. Next testnet rotation will update the widget and leave this link on a dead host.

.claude/skills/design-system/references/layouts.md
```diff
@@ -87,7 +101,7 @@ When the topic genuinely needs something extra:
 
 - **`config.hubHero`** — Swap `PageHero` for `HubHero` on a specific slug (used by Roadmap on `/roadmap/`). Declarative; lives in the topic config.
 - **`config.editBanner`** — Render the top-of-page "edit this page" banner on every page in the topic. Used by UseCases and AiAgents. Per-page opt-out via frontmatter `hideEditBanner: true` if a specific page needs to suppress.
-- **`afterContent` prop** — Render arbitrary JSX after the markdown content. Used by Staking for its community callout. Passed by the slug router for the one or two topics that need it. If you find yourself wanting a *third* `afterContent` consumer, consider promoting it to `config.afterContent` (still keyed by topic data).
+- **`afterContent` prop** — Render arbitrary JSX after the markdown content. Used by Staking for its community callout. Passed by the slug router for the one or two topics that need it. If you find yourself wanting a _third_ `afterContent` consumer, consider promoting it to `config.afterContent` (still keyed by topic data).
```
pettinarip, line 104: This PR removes the only `afterContent` implementation (the callout moved into the markdown), so this is now describing code that's gone. Same for the slot list up at line 28 and `cleanup-playbook.md:464`. While in here, `a11y.md:36` still lists `StakingHierarchy`, which this PR deletes.

pettinarip (approved):

PR #19044, opened 2026-08-11
fix(seo): server-render the tutorials list and homepage savings carousel

### LINKED ISSUE #18977: fix(seo): tutorials list and homepage carousel are invisible to crawlers (ssr: false)
## Describe the bug

Two content areas are loaded with `dynamic(..., { ssr: false })`, so they are absent from server-rendered HTML. Search engine crawlers and AI crawlers (GPTBot, ClaudeBot, PerplexityBot, OAI-SearchBot) largely do not execute JavaScript, so this content does not exist for them.

The tutorials hub is the severe case: crawlers receive **loading skeletons and zero tutorial links**.

### `/developers/tutorials/`

| Measure | Value |
|---|---|
| Tutorial links in server DOM (scripts stripped) | **0** |
| Tutorial paths present only in the RSC script payload | 57 |
| Loading skeleton markers in HTML | 83 |
| Words in raw HTML vs rendered | 426 vs 5,512 (**92% client-only**) |

### `/` (homepage)

| Measure | Value |
|---|---|
| `swiper-slide` in server DOM | **0** |
| `swiper-slide` on `/apps/` (control, same library) | 14 |
| Words in raw HTML vs rendered | 849 vs 1,098 (23% client-only) |

The `/apps/` control matters: Swiper server-renders fine in this codebase, so the library is not the cause. `ssr: false` is.

## To reproduce

```bash
# 0 tutorial links reach a crawler
curl -s https://ethereum.org/developers/tutorials/ \
  | perl -0pe 's/<script.*?<\/script>//gs' \
  | grep -c 'href="/developers/tutorials/'
# -> 0

# homepage carousel absent, but /apps/ carousel present
curl -s https://ethereum.org/ | perl -0pe 's/<script.*?<\/script>//gs' | grep -c swiper-slide      # -> 0
curl -s https://ethereum.org/apps/ | perl -0pe 's/<script.*?<\/script>//gs' | grep -c swiper-slide # -> 14
```

Verified against production on 2026-08-05.

## Expected behavior

Editorial content is present in server-rendered HTML and reachable without JavaScript.

## Root cause

- `app/[locale]/developers/tutorials/_components/TutorialsLazy.tsx:8` — `ssr: false` on the `dynamic()` import of `./tutorials`
- `app/[locale]/_components/HomepageLazy.tsx:34` — `ssr: false` on `SavingsCarousel`

## Suggested fix

Remove `ssr: false` from the tutorials list and from `SavingsCarousel`.

**Leave `SimulatorSection` (`HomepageLazy.tsx:44`) as-is.** The wallet simulator is a genuinely interactive widget and client-only rendering is defensible there.

This is not a blanket change to every `ssr: false` in the repo. `/developers/`, `/get-eth/` and `/resources/` measure a ~0% raw-vs-rendered gap despite also using `dynamic()`, so only these two need changing.

## Why this matters

The tutorials hub is a content hub plus 57 internal links that no non-JS crawler can see. Internal links also drive crawl discovery, so this plausibly affects indexing of the individual tutorials, not just the hub.

Of everything in a broader structured-data/crawlability audit, this is the largest single item: it is the difference between content existing and not existing for a crawler, which ranks well above any schema.org refinement.

## Acceptance criteria

- [ ] `curl` on `/developers/tutorials/` returns > 0 `href="/developers/tutorials/…"` with scripts stripped
- [ ] No loading-skeleton markers in server HTML for either page
- [ ] Homepage carousel copy greppable in `curl` output
- [ ] Wallet simulator still functions

SayedZahur786:
## Description

`/developers/tutorials/` and the homepage savings carousel were loaded with `dynamic(..., { ssr: false })`, so neither reached server-rendered HTML. Crawlers that don't execute JavaScript — Googlebot on its first pass, and GPTBot / ClaudeBot / PerplexityBot / OAI-SearchBot at all — received loading skeletons. The tutorials hub was the severe case: **zero links to individual tutorials**, which also removes the crawl-discovery path to the tutorials themselves.

### `ssr: false` was not the whole bug

Removing the flag is the obvious fix and it is **not sufficient**. `next/dynamic` wraps the component in `Suspense` either way, so the server emits the fallback skeleton inline and streams the real markup into a `<div hidden id="S:0">` that only an inline `$RC()` script swaps in.

Measured on a production build:

| `/developers/tutorials/` | before | drop `ssr:false` only | **direct import (this PR)** |
| --- | --- | --- | --- |
| tutorial links in server DOM | 0 | 57 | **57** |
| loading skeleton present | yes | **still yes** | **gone** |
| hidden streaming container | – | **1** | **0** |

The middle column satisfies "links present" but still hands a crawler a skeleton with the content in a `hidden` div. So both components are now **imported directly** rather than via `dynamic()`. This matches `app/[locale]/apps/_components/TopApps.tsx`, which imports the same Swiper component directly and already server-renders 13 slides inline — the control that showed the library was never the problem.

`TutorialsLazy.tsx` becomes unnecessary and is deleted, restoring the direct import that existed before #15616.

**`SimulatorSection` keeps `ssr: false`** — it is a genuinely interactive widget.

### Publish dates had to be pinned to UTC

Tutorial publish dates are date-only strings, so they parse as UTC midnight, and `getLocaleTimestamp` was formatting them in the *runtime's* timezone. The UTC server and any browser behind UTC disagree by a day.

This was already wrong for readers in the Americas — a tutorial published Sept 15 displayed as Sept 14 — and server-rendering the list would have converted it into a hydration mismatch. Verified by removing the pin: **58 cards shift by exactly one day** (`server=February 27, 2026 browser=February 26, 2026`).

Fixed at the call site rather than in the shared `dateTimeFormat` util, to avoid changing date formatting site-wide.

### Results (production build, `NEXT_PUBLIC_BUILD_LOCALES=en`)

| Page | Before | After |
| --- | --- | --- |
| `/developers/tutorials/` links | 0 | **57** |
| `/developers/tutorials/` skeleton | 37 markers | **0** |
| `/` `swiper-slide` | 0 | **3** |
| `/` carousel skeleton | present | **gone** |
| `/apps/` (control) | 13 slides | **13 slides, unchanged** |

### Performance

Server-rendering 109 cards (57 internal + 52 external) grows raw HTML, but the cost that ships is compressed:

| | raw | brotli |
| --- | --- | --- |
| `/developers/tutorials/` | 228 KB → 543 KB | **26.8 KB → 36.6 KB** |
| `/` | 298 KB → 328 KB | **33.4 KB → 35.7 KB** |

543 KB raw is above the `rsc.md` 500 KB figure, so flagging it explicitly. In context, **53 of 399 prerendered pages already exceed it**, including every comparable catalog page — `/developers/tools` is 949 KB raw / 77 KB brotli and `/wallets/find-wallet` is 944 KB / 71 KB. At 37 KB brotli the tutorials hub is now the **lightest catalog page on the site by transfer**.

- **CLS** should improve on both pages: the skeleton→content swap disappears.
- **LCP** unchanged — both areas are below the fold and the hero's `fetchPriority` is untouched.
- **INP/TBT** is the real trade: this JS now hydrates on initial load instead of just after. It was always downloaded, so total bytes are unchanged. Happy to defer to the Lighthouse CI numbers here — Next 16 + Turbopack no longer emits per-route First Load JS, so I could not measure it locally.

### Perf skill

`bundle.md` named `HomepageLazy.tsx` as the reference for the `ssr: false` pattern without distinguishing what the component holds, which is how this happened. Second commit splits that guidance (`ssr: false` for interactive widgets, direct import for editorial content), notes the `Suspense`/`<div hidden>` trap, and reframes the 500 KB HTML target as a compressed-bytes smoke signal compared against a peer page. **Kept as a separate commit so it can be dropped without affecting the fix.**

### Tests

`tests/e2e/lazy-sections-server-html.spec.ts` — fetches raw HTML over HTTP and strips `<script>` first, since the RSC flight payload contains the same hrefs and would make a naive search pass while the bug is present. Follows the shape of `tabs-server-html.spec.ts`. Covers the tutorials list, the carousel, the `/apps/` control, and timezone stability. Each assertion was confirmed to fail before the fix.

Verified locally: `eslint` · `prettier` · `tsc --noEmit` · 997 unit tests · 4 new e2e tests. Also checked in-browser: filtering (57→28→57), carousel navigation, simulator loads, **0 console errors and no hydration warnings**.

### Noticed, not changed

- `KPISection` still uses `dynamic()` without `ssr: false`, so its content also lands in a hidden streaming container — same mechanism, out of scope here.
- `dateTimeFormat` has no site-wide `timeZone` pin, so the same off-by-one exists anywhere a date-only string renders in a client component.
- The carousel's comparison cards are `motion.div`s with `initial={{ opacity: 0 }}`, so those stats sit at `opacity: 0` in server HTML. The slide copy and CTAs render normally. Changing it would mean changing the animation.

## Related Issue

Fixes #18977


SayedZahur786:
Hi @wackerow Can you please merge this

PR #19059, opened 2026-08-13
Derive network upgrade facts from Forkcast

pettinarip:
## Description

Upgrade facts on the roadmap pages are maintained by hand, so they drift. The Glamsterdam alert has been edited three times — H1 2026 → H2 2026 → Q4 2026 — while [Forkcast](https://forkcast.org/upgrade/glamsterdam) has published `Q4 2026` since 2025-11-15. The page has never matched the source.

This generates those facts from [`ethereum/forkcast`](https://github.com/ethereum/forkcast) instead.

**No visual change. Nothing imports this yet — the Netlify preview should be identical to `dev`.** That's the review check.

```
src/data/upgrades/types.ts       the contract
src/data/upgrades/generated.ts   machine-written, DO NOT EDIT
src/scripts/sync-upgrades/       fetch, normalize, emit
```

`pnpm sync-upgrades` rewrites the store from four upstream files: the upgrade records, the phase timeline, devnet launches, and per-EIP status.

## Supersedes #18991

Same schema — its `types.ts` is the origin of this contract and survives largely intact. What changed is the author: records are derived rather than transcribed, which also scales past one hand-authored file per fork.

**#18993, #18998 and #19000 stay open and are still Nuno's.** They need a rebase and a few field renames once this lands — [details here](https://github.com/ethereum/ethereum-org-website/pull/18991#issuecomment-5279556805).

## Design notes

- **Confidence lives in the data.** `mainnetTarget.confirmed` is true only once an activation epoch exists, so a projected quarter can't render as settled.
- **Both upstream sources are read.** `upgrades.ts` flattens the target to a bare year; `timeline-phases.ts` has `Q4 2026`. Reading only the first made our store less precise than its own source.
- **Dates carry only the precision they have** — year, quarter, month, or exact day. Quarter ranges (`Q3-Q4 2026`) degrade to the year rather than picking a bound.
- **Status is confidence, kind is kind.** Forkcast's `Networking` grouping becomes `scheduled` plus a flag, not a status of its own.
- **Only EIPs expected to ship are stored** — 50 rows, not 197. The full vocabulary is still mapped so an unknown status throws. Descoping stays visible: a declined EIP leaves the store, and a deletion reads clearly in a diff.
- **Nothing is hand-maintained.** Devnet-8's window, which Forkcast publishes only as prose, is left out rather than transcribed.

## Why committed rather than fetched at build time

Pages reading `public/content/*.md` can't use ISR on Netlify, so a data-layer getter would need `revalidate: false` and the value would change at an arbitrary later deploy, unreviewed. Committed means **merging is what makes a value live**. Think `pnpm-lock.yaml`. No commit sha is stamped in the file, so the staleness guard is just:

```bash
pnpm sync-upgrades
git diff --exit-code src/data/upgrades/
```

## Out of scope

The scheduled job (it opens PRs — deserves its own review), checking page prose against the data, any EIP directory, and consolidating with `networkUpgradeSummaryData.ts` (append-once history; this store is bot-writable, and keeping them apart bounds the blast radius of a bad automated write).

## Tests

14 unit tests over fixtures, no network — date parsing, range degradation, phase precedence, testnet extraction, milestone ordering, EIP filtering, and that an unmapped upstream status throws instead of silently emitting a gap.

## Open questions

1. **Is `src/data/upgrades/` the right home**, or should it live under `src/data/roadmap/`? Cheap to change now.
2. **Ask Forkcast for an `api/upgrades.json` endpoint?** They already ship `api/eip-stage-changes.json`, so it's ~12 lines of their own pattern — and it would replace regex-parsing their TypeScript with a real contract. Cross-team, needs an owner.
3. **AGPL-3.0 vs our MIT.** We ingest data rather than link code, but worth a sanity check.
4. **Fail-soft.** Upstream is another team's project with no API contract. The eventual job must fail into an issue and keep the last good store serving, and the CI check must be path-filtered so upstream downtime can't fail unrelated builds.


myelinated-wackerow (commented):
**Verdict: approve in principle, two things before merge.** Generating these facts is the right call and the mapping is the reviewable artifact it claims to be. I ran the sync end-to-end against live upstream rather than reading it: the parsers hold on the real files and the committed store reproduces exactly, except Devnet-8, which launched upstream yesterday (Aug 13). So the store is already stale. 19 tests pass (the body says 14). ESLint clean, no type errors in these files, no consumers, so the "preview identical to `dev`" claim holds.

### Before merge

**1. Re-sync.** Devnet-8 landed in `devnet-launches.json` on Aug 13 and isn't in the committed store. Nice accidental proof that the guard works and a preview of how often it will fire.

**2. `Milestone.name` bakes English into generated data.** `"Mainnet activation"` (7 rows), `"Sepolia fork"`/`"Hoodi fork"` (4), `"Devnet-8"` (9). Those render untranslated in all 24 non-English locales, and `"Mainnet activation"` is fully redundant with `kind: "mainnet"`. House convention splits the other way: `src/data/networks/networks.ts` carries `name: "Base"` because it's a proper noun, while every label on `/roadmap` comes out of `page-roadmap.json`. Suggest dropping `name` for mainnet and storing `network: "Sepolia"` / `version: 8`, so the UI does `t("testnet-fork", { network })`. Cheapest now, while nothing imports it, and it's a field rename #18993/#18998/#19000 are already taking.

Everything else is set up well for i18n: statuses and kinds are enum keys, `" Upgrade"` is stripped so the suffix stays translatable, and `PartialDate` is structured rather than pre-formatted. Two follow-ons worth naming so the UI PRs don't hand-roll them: a `PartialDate` formatter over `dateTimeFormat()`/`numberFormat()` (quarters have no Intl primitive, so `Q{n} {year}` needs an intl string, and Arabic/Urdu need the right numbering system), and `EipDecision.date` being a raw ISO string while every other date is a `PartialDate`.

### Worth fixing, especially before the scheduled job

**Silent gap on key mismatch.** EIPs join on `upgrade.id === forkName.toLowerCase()`, and phases and devnets key the same way. Nothing checks the join. It works today only because upstream writes `forkName: 'Hegota'` while the display name is `Hegotá`; if that ever gains its accent, an upgrade quietly drops to zero EIPs and loses its quarter precision with no throw. `normalizeEips` can't detect it (the filter just yields nothing), so the guard belongs in `normalize()`: collect upstream fork names and phase keys, throw on any that match no upgrade id. Same doctrine as the unmapped-status throw, and it's the one place the fail-loud design has a hole. The CI diff would show the deletion, but "descoping shows up as a deletion" is exactly what makes a mass deletion easy to wave through.

**`resolveMainnetTarget` is asymmetric in the direction that bit us.** `actualEndDate` outranks the headline across years, but a *projected* year disagreement silently prefers `upgrades.ts` — the file whose drift motivated this PR. A timeline of Q1 2027 against a headline of 2026 publishes 2026. Two upstream files disagreeing on the year is drift worth surfacing rather than resolving. No test either way.

**`disabled` is parsed, tested, and never read.** The signpost is excluded by a hardcoded `id === "previous-upgrades"` instead, correctly, since Dencun and Shapella are also `disabled: true` upstream. Drop the field or the next reader will assume it's load-bearing. Same block: the `!activationDate && !path?.startsWith("/upgrade/")` guard above it matches nothing in current data.

**Minor.** `emit.write` resolves `GENERATED_PATH` against cwd and shells out to `pnpm exec prettier`, so running it anywhere but the repo root writes elsewhere. `detail()`'s `[\d\s*]+` reads `23_935_694` as `23`. `{ when: null, confirmed: true }` is representable but meaningless.

### Suggestions for your open questions

1. Keep `src/data/upgrades/`. "Roadmap" is a page; this is fork data `/history` and the upgrade pages want too.
2. Worth asking, and it's the fragility everything above orbits, but don't block on another team's roadmap.
3. Reasoning looks right to me (facts ingested, no upstream code in the output, `sourceUrl` attributes), but that's a legal call, not a review one.
4. Agreed, and note nothing enforces staleness today since the check isn't in this PR.

— Opus 5

pettinarip:
@myelinated-wackerow pushed fixes for everything: re-synced the store (Devnet-8), restructured `Milestone` to drop the baked-in English names (`network`/`version` instead), added the join-key validation for the Hegotá-style drift, made `resolveMainnetTarget` throw on projected-year disagreement instead of trusting the headline, dropped the dead `disabled` field, and fixed the underscore-number parsing bug. Left the two UI follow-ons for whichever PR needs them first.

wackerow (approved):
Looks good! 🔥

pettinarip:
Force-merging — the Netlify checks are stuck on `pending`/`in_progress`, not failing. Netlify's own bot comment updated at 17:27 UTC with "Deploy Preview ready" for this exact commit, and the preview URL returns a clean 200. This lines up with today's GitHub incident (degraded API requests/webhooks), which is stopping the check-run status from updating, not the build itself. All other checks (unit tests, lint/type-check, Chromatic, GitGuardian) already passed.

PR #19072, opened 2026-08-14
chore(gh-aw): right-size Issue Triager schedule, model and effort

pettinarip:
Reduces the Issue Triager's scheduled run volume by ~80% and right-sizes its model and effort, with no change to triage behaviour.

## Changes

**`schedule: every 6 hours` → `daily around 05:00 on weekdays`** — 124 → ~22 runs/month.

The cron sweep is a backstop that drains `needs triage 📥` for issues the event path missed, not the primary triage route. Over the last month it ran 124 times at a median of 522 output tokens — roughly one label decision each — while the 30 event-driven runs did the real work.

Deliberately reduced rather than removed: the Intake Digest reads the queue at 07:00 weekdays and needs triage current, so the sweep still has to land ahead of it. 05:00 preserves the ordering (sweeper 06:00, digest 07:00). Real-time triage is untouched — the `issues` trigger still fires on every open/reopen, and `workflow_dispatch` stays as a manual escape hatch.

**`engine.model: claude-sonnet-5?effort=medium`** — previously unpinned, resolving to `claude-opus-4-8` at the provider default effort.

Why not Haiku, since this looks like cheap classification: the prompt also has to treat issue bodies as untrusted input and resist injection, spot wallet-support scams, and abstain rather than guess (*"Never apply `invalid` to a plausible real submission"*). A false `invalid` labels a real contributor as spam.

Why `medium` and not `low`: Sonnet 5 respects effort strictly at the low end and scopes work down to exactly what was asked — the wrong trade for those same abstention rules.

## Reviewer notes

- Setting `engine.model` also pins the threat-detection job and removes the `GH_AW_MODEL_AGENT_CLAUDE` / `GH_AW_MODEL_DETECTION_CLAUDE` repo-variable override *for this workflow*. That var is global across all four workflows, so per-workflow frontmatter is the right tool here — flagging in case you'd rather keep the var lever.
- Effort rides as a query param on the model string and is resolved by gh-aw's API proxy (same mechanism as its built-in `opus?effort=high` alias). Accepted values are `low` / `medium` / `high`.
- Lock file regenerated with `gh aw compile issue-triager` (v0.81.6, same compiler as the existing locks). Cron resolves to `54 4 * * 1-5`.

## Not in this PR

- **Effort on the other three workflows** — all still run at the default `high`. PR Reviewer and Backlog Sweeper do the real judgement work, so their effort is a quality decision that wants its own evaluation. Intake Digest is the best `low` candidate of the three (rigid template fill from pre-fetched JSON).
- **PR Backlog Sweeper model swap** — hardest of the four: long-horizon, multi-PR, and it must gather evidence across base-branch history to justify `recommend close`. Wants a shadow comparison, not a straight swap.
- **PR Reviewer stays on Opus** — best value per run of the four.


Blondie0123 (commented):
Commit

PR #19075, opened 2026-08-14
intl: bound translation-pipeline spend

myelinated-wackerow:
## Summary

Fixes the batching bug behind [run 31149083965](https://github.com/ethereum/ethereum-org-website/actions/runs/31149083965) ($1,108 to translate 42KB of new text) and bounds the pipeline so a repeat is arithmetically impossible.

`batchSections` subtracted the replicated CONTEXT payload from the per-batch budget, which floored the budget at 1 byte for any file with more than 32KB of existing translation -- one full-context call per changed section, 488 sections x 24 locales = 11,712 calls. TRANSLATE content now gets the whole chunk budget, per-batch context is capped and chosen from the sections nearest the batch, and batching splits on wire bytes (content plus the `<SECTION>` envelope, which outweighs the content for small JSON keys).

Bounds are input-side, because prompt bytes are assembled locally and knowable before sending: 64KB per call, `ceil(changedWireBytes / 32KB) x 64KB` of planned prompt bytes per file+locale checked before the first request, 32 batches per file, and a $100 whole-run fuse. The per-file bound is the primary one -- a run's total legitimately scales with how many files and locales are in scope, work per file does not.

Adds an OpenRouter transport (`LLM_PROVIDER=openrouter`, the workflow default) whose key carries a server-side daily credit limit the pipeline cannot exceed, sends `max_price` at our expected rate, and meters the provider's own reported cost. Also reveals that thinking tokens are billed as output and were invisible on the Gemini-direct path: `thoughtsTokenCount` is now counted, so historical cost lines in old logs are floors rather than actuals.

`MODE=estimate` (`pnpm intl:estimate`, or `mode: estimate` on the workflow) assembles every prompt a run would send, sends none, and reports projected bytes/requests/tokens/cost and the budget verdict per file+locale, exiting non-zero if anything would be refused. It shares the pipeline's planner rather than reimplementing it. Validated against the first supervised live run: $2.08 projected, $2.1552 actual.

Verified by `tests/unit/intl-pipeline/cost-incident.spec.ts`, which replays the incident against the real files from commit 3c76f545 -- 488 changed strings still detected, now 3 batches per locale within budget, and the same budget refuses the 62MB that run actually sent.


src/scripts/intl-pipeline/lib/llm/incremental-translate.ts
```diff
@@ -551,69 +551,134 @@ export type {
...
+}
+
+/**
+ * Bytes a section costs on the wire, not just its content: buildIncrementalPrompt
+ * wraps each one in a <SECTION id=".." action=".." heading=".."> envelope that
+ * repeats the id twice. JSON leaves are small and numerous with long key paths
+ * (~134 bytes of envelope each for quiz keys), so counting content alone
+ * understates a batch by more than the content itself.
+ */
+export function sectionWireBytes(section: {
```
pettinarip, line 583: `sectionWireBytes()` doesn't count the `heading="..."` attribute added for markdown sections — just content + `id`×2 + a flat pad. For files with long headings this can undercount enough that a batch that looks within budget actually serializes over the 64KB per-call ceiling.

src/scripts/intl-pipeline/estimate.ts
```diff
@@ -0,0 +1,364 @@
...
+      (inputTokens / 1_000_000) * INPUT_RATE_USD_PER_1M +
+      (outputTokens / 1_000_000) * OUTPUT_RATE_USD_PER_1M,
+    verdict: "ok",
+  }
+}
+
+// Belt to the braces: nothing in this file's import graph reaches an LLM call
+// site, and a zero fuse means that if one is ever added, the first request
+// throws instead of spending. Set before ./constants is read.
+process.env.INTL_MAX_COST_USD = "0"
```
pettinarip: This `INTL_MAX_COST_USD = "0"` safety net runs after the imports that already read the real env value, so it has no effect — the "fail closed at $0" comment isn't actually true.

src/scripts/intl-pipeline/lib/llm/cost-meter.ts
```diff
@@ -0,0 +1,165 @@
...
+
+export function runFuseUsd(): number {
+  return fuseUsd
+}
+
+/**
+ * Throw if the run has already spent its fuse. Called before each request, so
+ * the guard trips one call late at worst.
+ */
+export function assertRunFuse(context: string): void {
```
pettinarip, line 126: `assertRunFuse` is check-then-act — spend only gets recorded after each call resolves, so with concurrent calls several can pass the fuse check before any of them record cost. Could overshoot the $100 cap by a few calls' worth under concurrency.

src/scripts/intl-pipeline/lib/llm/gemini.ts
```diff
@@ -1412,7 +1460,13 @@ export async function callGeminiRaw(
           `[${ts()}] [gemini] RESPONSE model=${modelId} ${ctx} ` +
             `duration=${duration}s ` +
             `tokens_in=${usage?.promptTokenCount || 0} ` +
-            `tokens_out=${usage?.candidatesTokenCount || 0}` +
+            `tokens_out=${visibleOutput}` +
```
pettinarip: `tokens_out=X reasoning=Y` reads as additive here, but for OpenRouter `tokens_out` already includes reasoning tokens while on Gemini-direct it doesn't — worth double-checking before using this line for cost triage.

pettinarip (approved):
LGTM. Left a few inline notes from an AI review pass on the new cost guards, in case they're useful. Nothing blocking.

pettinarip:
This blocker is resolved — `614911d136` ("intl: vendor incident fixtures so CI can run them"), currently at HEAD, vendors the incident fixtures so `cost-incident.spec.ts` no longer relies on `git show` against a commit the shallow CI checkout can't see.

PR #19102, opened 2026-08-18
fix(ui): allow `rel` on Card, restore referrer on Speedrun links

### LINKED ISSUE #19101: Speedrun Ethereum links do not send referrer data
### Describe the bug

We found a possible referral attribution problem in the links from ethereum.org to Speedrun Ethereum.

Plausible stopped identifying `ethereum.org` as the referral source on July 30, 2026. The reported traffic decreased from approximately 20–35 visits per day to almost zero.

Image

## Cause

This change can be the cause:

https://github.com/ethereum/ethereum-org-website/commit/87fe92ff195d68c92f17e51e936d2d3438a1fd7f

Before this change, `SpeedRunCard` and `BuilderCard` used `ButtonLink` with this property:

```tsx
rel="noopener"
```

The change moved each URL from `ButtonLink` to `Card`. It also removed the explicit `rel="noopener"` property.

When `Card` has an `href`, it renders `BaseLink`. For external URLs, `BaseLink` uses:

```tsx
target="_blank"
rel="noopener noreferrer"
```

The `noreferrer` value prevents the browser from sending the `Referer` header.

The commit does not add `noreferrer` directly. It removes the previous `rel="noopener"` override. This causes the default value from `BaseLink` to become active.

## Affected links

The current `/developers/` page uses `rel="noopener noreferrer"` for these Speedrun Ethereum links:

- `https://speedrunethereum.com/`
- `https://speedrunethereum.com/challenge/tokenization`
- `https://speedrunethereum.com/challenge/dex`
- `https://speedrunethereum.com/challenge/stablecoins`

If this behavior is not intentional, I would be happy to open a PR with a small fix., restoring the `rel="noopener"` for these links

### To reproduce

1. Go to https://ethereum.org/developers/.
2. Find a link to Speedrun Ethereum.
3. Inspect the link in the browser developer tools.
4. See that the link uses `rel="noopener noreferrer"`.
5. Open the link.
6. See that the browser does not send `ethereum.org` as the referrer.

This behavior affects the main Speedrun Ethereum link and the three challenge links.

### Expected behavior

The links should send `ethereum.org` as the referrer, unless the removal of this data is intentional.

Plausible should identify these visits as referral traffic from `ethereum.org`. An explicit source parameter can also provide this attribution.

### Screenshots

_No response_

_No response_

_No response_

### Additional context

_No response_

- [x] Yes
- [ ] No

pettinarip:
`87fe92f` moved the Speedrun hrefs from an inner `ButtonLink` onto `Card` and dropped `rel="noopener"` — `CardProps` couldn't accept it, since `rel` lives on `AnchorHTMLAttributes` not `HTMLAttributes`. `BaseLink`'s external default (`noopener noreferrer`) then took over and stripped the `Referer` header, so Speedrun Ethereum lost all referral attribution from ethereum.org.

Widens the `Pick` to include `rel` and restores the override on the 4 affected links.

Fixes #19101

### DIFF
--- a/app/[locale]/developers/_components/BuilderCard.tsx
+++ b/app/[locale]/developers/_components/BuilderCard.tsx
@@ -20,8 +20,7 @@ type BuildCardProps = {
 const BuilderCard = ({ path }: BuildCardProps) => (
   <Card
     href={path.href}
-    // Preserve referrer for partner attribution (#19101)
-    rel="noopener"
+    sendReferrer
     variant="ghost"
     className="border"
     customEventOptions={{
--- a/app/[locale]/developers/_components/BuilderCard.tsx
+++ b/app/[locale]/developers/_components/BuilderCard.tsx
@@ -20,6 +20,8 @@ type BuildCardProps = {
 const BuilderCard = ({ path }: BuildCardProps) => (
   <Card
     href={path.href}
+    // Preserve referrer for partner attribution (#19101)
+    rel="noopener"
     variant="ghost"
     className="border"
     customEventOptions={{
--- a/app/[locale]/developers/_components/SpeedRunCard.tsx
+++ b/app/[locale]/developers/_components/SpeedRunCard.tsx
@@ -25,6 +25,8 @@ const SpeedRunCard = ({
 }: SpeedRunCardProps) => (
   <Card
     href="https://speedrunethereum.com/"
+    // Preserve referrer for partner attribution (#19101)
+    rel="noopener"
     variant="ghost"
     className={cn(
       "relative min-h-112 overflow-hidden rounded-b-none hover:bg-inherit hover:shadow-none",
--- a/app/[locale]/developers/_components/SpeedRunCard.tsx
+++ b/app/[locale]/developers/_components/SpeedRunCard.tsx
@@ -25,8 +25,7 @@ const SpeedRunCard = ({
 }: SpeedRunCardProps) => (
   <Card
     href="https://speedrunethereum.com/"
-    // Preserve referrer for partner attribution (#19101)
-    rel="noopener"
+    sendReferrer
     variant="ghost"
     className={cn(
       "relative min-h-112 overflow-hidden rounded-b-none hover:bg-inherit hover:shadow-none",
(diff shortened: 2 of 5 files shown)


myelinated-wackerow:
Diagnosis is spot on -- thanks for tracking this down to the commit.

One suggestion on the shape. `rel="noopener"` is redundant at the call site (browsers imply it for `target="_blank"`), so its real job here is *deleting* `noreferrer` from `BaseLink`'s hardcoded string. That makes the attribute name say the opposite of its intent, and it means `CardProps` has to expose an anchor-only prop that also lands on the `<div>` branch when there's no `href`.

An intent-named prop on `BaseLink` is a few lines and reads correctly:

```tsx
// BaseProps
/** Opt out of `noreferrer` on external links so the destination gets attribution */
sendReferrer?: boolean

// isExternal branch
rel={sendReferrer ? "noopener" : "noopener noreferrer"}
```

Then `CardProps` picks `sendReferrer` instead of `rel`, `Card` destructures and forwards it to `BaseLink`, and the two call sites become `sendReferrer` with no comment needed.

Separately, worth a follow-up: Speedrun still gets no referrer from ~12 other links -- 11 challenge URLs in `src/data/externalTutorials.json` (`/developers/tutorials`) and the `speedrunethereum.com` entry in the `/developers/tools` listing. Happy to open an issue for a partner-host allowlist in `BaseLink` so this can't regress the next time a component is refactored.

-- Opus 5/@wackerow

pettinarip:
Good call — `sendReferrer` is in as of a516ba3. You're right that `rel="noopener"` was describing its mechanism rather than its intent, and the `<div>`-branch leak was a real wart.

Implemented as you sketched. One detail worth noting: both `BaseLink` and `Card` destructure `sendReferrer`, so it can't reach the DOM as an unknown attribute on either branch.

I also migrated the one remaining raw `rel="noopener"` (the Scaffold-ETH docs link in `developers/page.tsx`) to `sendReferrer` — same intent, and leaving it behind would have meant two idioms for the same thing right after we picked one. Easy to drop if you'd rather keep this PR narrower.

On the follow-up: your count is exact — 11 challenge URLs in `externalTutorials.json`. Two additions to the scope:

- There's a **third** surface you didn't mention: markdown content links, e.g. `public/content/developers/docs/oracles/index.md:435` plus its translations.
- That surface actually strengthens the allowlist argument. Markdown links can't take a prop at all, so no call-site solution can ever cover them — a host allowlist in `BaseLink` is the *only* shape that fixes all three surfaces at once.

Please do open the issue. Happy to take it.

wackerow (approved):

Issue #17146, opened 2026-01-22
fix: Filter browser extension stack overflow errors in Sentry


minimalsm:
## Problem

Sentry receives ~1,700+ errors/month with message:
```
Maximum call stack size exceeded RangeError src/preload/document.js HTMLMediaElement.canPlayType
```

This is **NOT a codebase bug**. The path `src/preload/document.js` doesn't exist in our codebase—it's injected by browser extensions that wrap `HTMLMediaElement.canPlayType` with improper recursion.

## Solution

Add filtering to `instrumentation-client.ts` using Sentry's built-in `denyUrls` plus a minimal `beforeSend` hook for the specific injection path.

```typescript
Sentry.init({
  // ... existing config ...

  denyUrls: [
    /chrome-extension:\/\//,
    /moz-extension:\/\//,
  ],

  beforeSend(event) {
    // Filter extension injection script errors not caught by denyUrls
    const frames = event.exception?.values?.[0]?.stacktrace?.frames ?? []
    const isExtension = frames.some(f =>
      f.filename?.includes('preload/document.js') ||
      f.abs_path?.includes('preload/document.js')
    )
    return isExtension ? null : event
  },
})
```

## Acceptance Criteria

- [ ] Add `denyUrls` for Chrome/Firefox extension protocols
- [ ] Add `beforeSend` filter for `preload/document.js` path
- [ ] Check both `filename` and `abs_path` in stack frames
- [ ] Verify "Maximum call stack" errors drop by >90% after deploy

## Files to Modify

- `instrumentation-client.ts` - Add `denyUrls` and update `beforeSend`

## References

- [Sentry denyUrls docs](https://docs.sentry.io/platforms/javascript/configuration/filtering/#decluttering-sentry)
- [Sentry beforeSend docs](https://docs.sentry.io/platforms/javascript/configuration/filtering/#using-beforesend)

MarcorpAI:
tried bugmail for error tracking ?? its lightweight and doesnt have any bloat or overwhelming dashboard , been working on this myself too

[bugmail](https://www.bugmail.site)

PR #19119, opened 2026-08-20
[recovery] fix(sentry): drop frameless stack-overflow noise (ETHORG-9Q)

pettinarip:
## Production error

- **Message:** `RangeError: Maximum call stack size exceeded.`
- **Sentry:** [ETHORG-9Q](https://ethereumorg-ow.sentry.io/issues/6813688416/)
- **Occurrences:** 31 (cumulative, first seen 2026-08-13, last seen 2026-08-20)
- **Culprit:** `/:locale/what-is-ethereum` · unhandled · 0 users affected

## Root cause

This is the residue of an error class the team already diagnosed. Issue #17146
established that "Maximum call stack size exceeded" on this site is **not a
codebase bug** — a browser extension injects `src/preload/document.js` and wraps
`HTMLMediaElement.canPlayType` with improper recursion. At the time it was
generating ~1,700 events/month.

The fix from #17146 landed in `instrumentation-client.ts` and works: it walks the
event's stack frames and drops anything referencing `preload/document.js`. Volume
is down roughly 90%+.

What it can't catch is the case where the stack overflows *inside* the injected
frame, so Sentry receives the event with **no stack frames at all**. Three
signals in ETHORG-9Q line up with that:

- the alert carries no stack/detail
- `culprit` is the transaction name (`/:locale/what-is-ethereum`) rather than a
  function or module — Sentry only falls back to the transaction when there is no
  usable in-app frame
- 0 users across 31 events

With no frames, the `frames.some(...)` filter has nothing to match on and the
event passes through. I also audited the server and client render path for
`/what-is-ethereum` (page, `page-jsonld`, `ContentLayout`, `TableOfContents`,
`Breadcrumbs`, `CrawlableNav`) and swept `src/` and `app/` for self-recursive
functions; every recursion reachable from this route is depth-bounded or bounded
by static data. There is no unbounded recursion of our own on this page — it just
happens to be the site's highest-traffic content page, which is why the residual
noise groups here.

## Fix

`instrumentation-client.ts` already has a no-stacktrace branch in `beforeSend`,
added for the same reason (frameless wallet-extension JSON-RPC errors,
ETHORG-7Q). This adds one more message check to that existing branch:

```ts
if (hasNoStacktrace) {
  const message = values[0]?.value ?? ""
  if (/Internal JSON-RPC error/i.test(message)) return null
  if (/Maximum call stack size exceeded/i.test(message)) return null
}
```

The `hasNoStacktrace` guard is what keeps this narrow, and it's the reason this
went in `beforeSend` rather than `ignoreErrors`. An `ignoreErrors` entry would
match on message alone and permanently blind us to genuine infinite-recursion
regressions in our own code — exactly the kind of severe bug you want paged on.
Anything thrown from our bundle carries frames and is still reported; only the
frameless third-party variant is dropped.

## Verification

```
$ pnpm type-check
> next typegen && tsc --noEmit && tsc --noEmit -p netlify/edge-functions
Generating route types...
✓ Types generated successfully
```

```
$ pnpm exec eslint instrumentation-client.ts
(no output — clean)
```

Behavioural verification requires production traffic: after deploy, ETHORG-9Q
should stop accruing events. If it keeps growing, the events do carry frames and
this diagnosis is wrong — revert and pull a raw event off the Sentry issue.

### DIFF
--- a/instrumentation-client.ts
+++ b/instrumentation-client.ts
@@ -87,6 +87,13 @@ Sentry.init({
     if (hasNoStacktrace) {
       const message = values[0]?.value ?? ""
       if (/Internal JSON-RPC error/i.test(message)) return null
+      // Extension-injected recursion (see #17146: wrappers around
+      // HTMLMediaElement.canPlayType) overflows the stack inside the injected
+      // frame, so the event arrives with no frames at all and the
+      // `preload/document.js` filter below never sees it. Real recursion bugs
+      // in our own bundle always carry frames, so they still get reported.
+      // (ETHORG-9Q)
+      if (/Maximum call stack size exceeded/i.test(message)) return null
     }
 
     // Filter extension injection script errors not caught by denyUrls


pettinarip:
Closing in favour of #19133. Checking this against the live Sentry data turned up three problems.

**The change is inert.** The raw event for ETHORG-9Q is not frameless — it carries one synthetic `window.onerror` frame:

```
mechanism: auto.browser.global_handlers.onerror
stacktrace: 1 frame
  filename: "undefined"  absPath: "undefined"  lineNo: 189  colNo: 70
```

`hasNoStacktrace` is `values.every((v) => !v.stacktrace?.frames?.length)`, so with `frames.length === 1` the guard is `false` and the new message check never runs. `rawStacktrace` carries the same frame, so this is the SDK synthesising a frame from the `onerror` `url/lineno/colno` args (where `url` arrived as the literal string `"undefined"`), not something the server added — the frame is present at `beforeSend` time.

**The root cause does not hold.** All 242 events in 30d are iOS: 231 Chrome Mobile iOS, 11 Google app. iOS has no browser extensions — Chrome there is a WKWebView shell — so no extension can be injecting `preload/document.js`. Mobile Safari has 90 error events in the same window and contributes zero to this issue, which also rules out WebKit's stack limit and our own bundle; it points at Google's injected `WKUserScript`.

**The figures are off.** ETHORG-9Q is 2,666 occurrences first seen 2025-08-16, not 31 first seen 2026-08-13. Culprit is `/:locale`, not `/:locale/what-is-ethereum`, and it is spread over 20+ URLs (homepage 83, find-wallet 27, whitepaper 21, what-is-ethereum 16) — so the render-path audit of `/what-is-ethereum` was aimed at the wrong page.

Worth noting the verification criterion here would have misfired: ETHORG-9Q would have kept accruing, and the conclusion would have been "diagnosis wrong, revert" when the diagnosis was merely unverified and the implementation inert.

The reasoning that an `ignoreErrors` entry is too blunt was right, and #19133 keeps it — it drops the event on evidence the frame has no usable location, so a real recursion bug in our bundle still reports. That PR also picks up ETHORG-73, a bigger neighbour at 3,794 occurrences that no frame-based rule can match.

pettinarip:
Superseded by #19133.

PR #19122, opened 2026-08-20
upgrades: weekly Forkcast data sync, with a content-review issue

pettinarip:
Keeps the upgrade store shipped in #19059 current, and notifies a human only when a change has prose behind it. Two outputs with different lifetimes: a disposable data PR, and one long-lived issue.

## What's changing

- **`sync-upgrades.yml`** — runs `pnpm sync-upgrades` every Friday at 09:00 UTC (after Thursday's ACD call). Exits silently when the store hasn't moved. Otherwise it type-checks and unit-tests the regenerated store, then opens or refreshes a data PR on one long-lived branch, so an unmerged PR gets updated rather than stacked behind a new one. The checks run in the workflow because PRs opened with `GITHUB_TOKEN` don't trigger CI.
- **`classify.ts`** — decides whether a human needs to look. Flags the mainnet target, an upgrade's status, an EIP arriving or being descoped, and a new upgrade. Stays silent on devnet dates and ACD call references: Glamsterdam went through 9 devnets in 4 months and no page states a devnet number, so flagging them would train everyone to ignore the issue. It compares the two store versions as objects rather than grepping diff text, which is what makes "an EIP arrived" distinguishable from "a devnet slipped a day". When something notable moved, the run appends a comment to a single issue titled *Upgrade data: changes needing content review* — a comment rather than a body edit, so subscribers get notified and a year of appends can't outgrow the body size limit.
- **`coverage.ts`** — rides along in that issue when an EIP moved, listing scheduled EIPs on unshipped upgrades that no English page names. Today: EIP-8070, EIP-8136 and EIP-8189 on Glamsterdam, plus Hegotá having no page yet.

## Notes for review

- The store's header claimed *"CI fails on any diff this script would produce"*. It never did, and that guard is deliberately out of scope here, so the comment now says what actually happens. The same 2-line change is applied to the committed `generated.ts` so the first real sync PR shows data only.
- Simulated end to end against live Forkcast. This week's run produces `glamsterdam · +EIP-8189 · scheduled` — EIP-8189 is `snap/2 - BAL-Based State Healing`, which arrived after the store was generated on 2026-08-13.
- Both directions of the gate are covered by tests, including the negative cases — a devnet date moving asserts an empty result, so a future change can't quietly start flagging them.


wackerow (approved):
@pettinarip Looks good overall, no blockers on my end.

Claude/Opus notes two non-blocking suggestions that I'll drop here:

---

**1. Issue dedup ages out (`sync-upgrades.yml:151`).** `gh issue list --limit 100` only sees the 100 newest open issues; the repo has 93 open today. Once it crosses 100, the long-lived issue drops out of that window and each Friday opens a duplicate instead of appending. One flag fixes it: `--search "in:title \"$ISSUE_TITLE\""`.

**2. Weekly no-op churn while a data PR is unmerged.** `changed` is computed against `dev`, not the sync branch, so a week where the store didn't move still commits, force-pushes an identical-content commit and re-titles the PR - a notification for nothing. This is the normal case rather than an edge one: three `update-chains` PRs are open right now (#19018, #19073, #19138). `SYNC_BASE_REF` already solved this noise for the issue side; the PR side still has it.

The one-branch design is a real fix for the stacking those three PRs demonstrate.
---

Approving as-is if we want to pull in and iterate separately

pettinarip:
Thanks! Leaving 1 as-is — we're at 93 open issues against a 100 window, and we'll clear that backlog long before adding 100 more. Worst case is a duplicate issue, and we'll be watching it run.

Fixed 2 though, good catch. The baseline is now resolved before the sync, so it diffs against the sync branch instead of dev — a week with no new data does nothing at all now, no push and no re-dated title.

Small side effect: the diff in the issue comment is now just that week's change rather than everything since dev. Reads better to me anyway, and the full diff is still on the PR.

PR #19124, opened 2026-08-20
feat: make site content agent-discoverable

myelinated-wackerow:
## Summary

Every file under `public/content` is already served as `text/markdown` in production -- just at unguessable URLs (`/content/<slug>/index.md`) that nothing advertises. This makes that content addressable and discoverable, and adds Content Signals to `robots.txt`.

`/<page>.md` now 307s to the same source file the page renders, for English and all 24 translations, mirroring `importMd()`'s slug resolution. Rules live in `md-redirects.config.js` next to the existing `redirects.config.js`. Redirects rather than rewrites: on Netlify's Next runtime a rewrite resolves inside the server function, which cannot reach `public/` assets (they live only on the CDN) -- the first deploy preview proved this with 404s, which doubles as proof the content tree is not bundled into the function. Markdown-backed pages advertise the URL via `<link rel="alternate" type="text/markdown">`, added in `getMdMetadata` rather than the shared `getMetadata` -- React pages have no `.md` counterpart and advertising a 404 is worse than silence.

`robots.txt` moves from the `robots.ts` metadata route to a route handler, because `MetadataRoute.Robots` has no field for `Content-Signal`. Output is byte-identical to today's apart from the directives; the non-production `Disallow: /` branch is unchanged.

Prompted by [isitagentready.com](https://isitagentready.com)'s scan. A follow-up commit adds the static HTTP `Link:` header its `linkHeaders` check wants and advertises the `.md` convention in both llms.txt indexes. The `markdownNegotiation` check stays unimplemented on purpose -- `Accept`-header negotiation needs `Vary: Accept`, which would fragment the CDN cache on every page. Not worth it for a checkbox.

Verified on the deploy preview: the `Link` header, both llms.txt updates, the per-page alternate link, direct `/content/.../index.md` serving as `text/markdown`, and the non-production robots branch; the `.md` redirect chain re-verified after the redirect fix. The Content-Signal directives render only on production. Known gap: `importMd` falls back to English when a translation is missing, but `/es/<page>.md` will 404 in that case.


pettinarip (changes requested):
Tested the rewrites on the deploy preview — they 404 for everything except `/latest/*`:

| URL | Result |
|---|---|
| `/smart-contracts.md` | 404 (HTML error page) |
| `/developers/docs/intro-to-ethereum.md` | 404 |
| `/latest/building-on-ethereum-in-2026.md` | 200 `text/markdown` |
| `/content/smart-contracts/index.md` (direct) | 200 `text/markdown` |

`next.config.js:275` excludes all of `public/content` from the SSR function, and `next.config.js:307-316` re-includes only `./public/content/latest/**/*.md`. An `afterFiles` rewrite resolves *inside* the Next server, so outside that one subtree the function has no markdown on disk and there's nothing to serve — Netlify's CDN only serves the literal `/content/...` path, which is why the direct URL still works. That's exactly the risk called out in the description, and it did fail.

Re-including the content isn't an option: `public/content` is 319MB (229MB of it translations), and that exclude is what keeps the function under the 250MB limit. So either the rewrite moves to a layer that runs before the function (netlify.toml `[[redirects]]` or an edge function), or the `<link rel="alternate">` points at the already-working `/content/<slug>/index.md` URLs and the rewrite is dropped.

Consequence worth flagging separately: `src/lib/md/metadata.ts:50` ships `<link rel="alternate" type="text/markdown">` on every markdown page, so right now we'd advertise a 404 to crawlers and agents on everything but the four `/latest` posts — the inverse of the "advertising a 404 is worse than silence" principle in the description. This shouldn't land ahead of a working rewrite.

Smaller things:

- **`.md` responses are uncacheable.** The working `/latest/*.md` response carries `cache-control: public,max-age=0` plus the full `netlify-vary` RSC header list, so every agent fetch is a function invocation, while the identical bytes at `/content/.../index.md` are a plain CDN asset (`etag`, `must-revalidate`).
- **`rewrites.config.js:26` duplicates content.** With `localePrefix: as-needed` the HTML route `/en/smart-contracts/` 301s to `/smart-contracts/`, but rule 2 gives `/en/latest/building-on-ethereum-in-2026.md` a real 200 — byte-identical to the unprefixed URL, no canonical between them.
- **`app/robots.txt/route.ts:16`** — `Content-Signal` is separated from its `User-Agent: *` group by a blank line. RFC 9309's grammar tolerates that, but record-based parsers (e.g. Python's `urllib.robotparser`) reset group state on a blank line and drop the directive. Cloudflare's own managed output keeps it adjacent. Free to fix.

Two notes rather than findings: the new unit spec passes (5/5) but structurally can't catch the 404 — it asserts pattern matching and bucket assignment, not file availability in the deployed function; one preview smoke check on a non-`latest` URL would have. And the preview zone has Cloudflare Managed robots.txt enabled, prepending its own `Content-Signal` group, so the preview can't be used to verify that part — production has no such block.

app/robots.txt/route.ts
```diff
@@ -0,0 +1,31 @@
...
+//
+// The sitemap is sharded per locale via generateSitemaps() (app/sitemaps/sitemap.ts)
+// and indexed by a <sitemapindex> at /sitemap.xml (app/sitemap.xml/route.ts). Point
+// crawlers at that single index; it enumerates every per-locale shard.
+const PRODUCTION = `User-Agent: *
+Allow: /
+
+# Content Signals (https://contentsignals.org). Site content is MIT-licensed,
+# so every declared use is permitted.
+Content-Signal: ai-train=yes, search=yes, ai-input=yes
```
wackerow, line 17: Btw, this entry is the reason we cannot use the type-restricted `robots.ts` file that is removed in this PR

myelinated-wackerow:
Point-by-point after the follow-up commits:

- **Rewrites 404 outside `/latest/*`**: fixed in ef5dd98 by replacing the rewrites with 307 redirects (`md-redirects.config.js`) -- the function never needs the files on disk; the client's second request fetches the CDN asset directly. Verified on the current preview: `/smart-contracts.md`, `/developers/docs/intro-to-ethereum.md`, and `/es/smart-contracts.md` all 307 -> 200 `text/markdown`.
- **Advertising 404s**: same fix -- the alternate-link URLs now resolve wherever the page is markdown-backed.
- **Uncacheable `.md` responses**: the bytes are now the plain CDN asset (`etag`, `must-revalidate`); what remains per fetch is a zero-byte 307. Pointing the alternate straight at `/content/<slug>/index.md` would drop that hop too, but the `.md` URLs stay guessable from page URLs and decoupled from the content-dir layout, which is what llms.txt documents.
- **`/en/` duplicate 200s**: dissolved by the same change -- `/en/<page>.md` and `/<page>.md` both 307 to the same single `/content/...` URL, so one canonical resource.
- **`Content-Signal` adjacency**: fixed in de2ca8b -- directive now sits inside the `User-Agent: *` group; comment-only lines don't reset `urllib.robotparser` state (verified against it).
- **Spec blind spot**: agreed -- this round was verified with preview curls; the spec still only pins the mapping logic.

-- Fable 5

pettinarip (approved):
Nice work on this

One small nice-to-have, not blocking: hitting `/nonexistent.md` lands on an HTML 404, so an agent asking for markdown gets an error page. A non-forced Netlify rule on `/content/*` could return a markdown 404 instead, pointing at `llms.txt`. Next's docs do something similar, though they return 200, which we probably don't want to copy.

Bit of a shame we can't just rewrite and skip the redirect hop entirely.

PR #19127, opened 2026-08-21
fix(i18n): use CLDR country names for zh-tw and te

### LINKED ISSUE #19126: Event locations render in Simplified Chinese on the zh-tw site
On `/zh-tw/community/events`, country names in event locations appear in Simplified Chinese, and Taiwan is shown as "中国台湾省" — e.g. "Taipei, Taiwan" renders as `台北, 中国台湾省` rather than `台北, 台灣`.

`getCountryTranslation` in `src/lib/utils/geography.ts` strips the region subtag before the lookup, so `zh-tw` resolves through `i18n-iso-countries`' `zh` data. That works for `pt-br` → `pt`, but the package only ships Simplified Chinese, so for `zh-tw` the fallback also changes the script.

Two related things I noticed while looking at this:

- `zh` shows "中国台湾省" as well; CLDR uses "台湾"
- `te` has no data in the package, so country names stay in English there

I'd be glad to open a PR. Resolving just these locales through `Intl.DisplayNames` (CLDR) fixes them without changing the other 22.

Toby1009:
Closes #19126.

`getCountryTranslation` strips the region subtag before the lookup, because `i18n-iso-countries` is keyed by base language. That's right for `pt-br` → `pt`, but the package ships no Traditional Chinese data, so `zh-tw` resolved through `zh` and event locations rendered in Simplified script — with Taiwan as "中国台湾省".

This routes the two locales the package can't serve through `Intl.DisplayNames` (CLDR) instead. Node ships full ICU, so there's no new dependency.

| locale | before | after |
| --- | --- | --- |
| `zh-tw` | `Taipei, 中国台湾省` | `Taipei, 台灣` |
| `zh-tw` | `美国` / `南韩` / `德国` | `美國` / `南韓` / `德國` |
| `te` | `Taiwan` (untranslated) | `తైవాన్` |

**Scope.** Only `zh-tw` and `te` change — `zh` and the other 22 locales keep reading from the package, unchanged. Switching everything to ICU would move ~350 country names, and not all of them for the better (ICU renders Italy as "Italy" in `vi` and Croatia as "Croatia" in `sw`), so that's better left to someone who can review those languages.

The CLDR long form for the SARs is a descriptor ("中國香港特別行政區"), so `HK`/`MO` read the short form; elsewhere the short form is an abbreviation, so long is the default.

**Testing.** Verified on `/zh-tw/community/events/meetups` — the `Taipei, Taiwan` and `Hong Kong SAR` entries in `community-meetups.json` render as `Taipei, 台灣` and `香港`, and `/zh/` is unchanged. `pnpm test:unit`, `pnpm type-check` and `eslint` pass.

### DIFF
--- a/src/lib/utils/geography.ts
+++ b/src/lib/utils/geography.ts
@@ -132,6 +132,28 @@ const COUNTRY_NAME_ALIASES: Record<string, string> = {
   // Add more here as needed
 }
 
+// i18n-iso-countries is keyed by base language and ships no Traditional
+// Chinese or Telugu data, so zh-tw fell back to Simplified zh and te to English.
+const ICU_LOCALES = new Set(["zh-tw", "te"])
+
+// CLDR's long form for the SARs is a descriptor ("中國香港特別行政區"); its short
+// form is the plain city name. Elsewhere the short form is an abbreviation.
+const SHORT_FORM_REGIONS = new Set(["HK", "MO"])
+
+const regionNamesCache = new Map<string, Intl.DisplayNames>()
+
+function getRegionName(code: string, locale: string): string | undefined {
+  const style = SHORT_FORM_REGIONS.has(code) ? "short" : "long"
+  const cacheKey = `${locale}:${style}`
+
+  let displayNames = regionNamesCache.get(cacheKey)
+  if (!displayNames) {
+    displayNames = new Intl.DisplayNames([locale], { type: "region", style })
+    regionNamesCache.set(cacheKey, displayNames)
+  }
+  return displayNames.of(code)
+}
+
 /**
  * Translate an English country name into the target locale.
  *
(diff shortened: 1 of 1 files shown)


Toby1009:
@Xeift you've contributed here before and read Traditional Chinese — would you mind sanity-checking the zh-tw country names in this one? The Taiwan label is the part I'd most like a second pair of eyes on.

Xeift:
IMO, `台灣` looks correct and natural for zh-TW.
To be clear, I haven’t checked the code since I'm not a reviewer

moven0831 (changes requested):
Nice work! some thoughts open to discussion:

1. The comments on the new lines in `geography.ts` feel a bit heavy to me. Could you drop the ones that aren't adding much and tighten up the rest? Some prompt that might be helpful: https://github.com/diegopetrucci/ai-custom-prompts/blob/main/remove-ai-code-slop.md

2. Want to check If you've already run the web app locally and confirmed the fix? If so, then `tests/unit/i18n/geography.spec.ts` feels a little over-engineered to me, and it only covers zh-TW. Happy to keep it if you think it earns its place.

Toby1009:
> Nice work! some thoughts open to discussion:
>
> 1. The comments on the new lines in geography.ts feel a bit heavy to me. Could you drop the ones that aren't adding much and tighten up the rest? Some prompt that might be helpful: https://github.com/diegopetrucci/ai-custom-prompts/blob/main/remove-ai-code-slop.md
>
> 2. Want to check If you've already run the web app locally and confirmed the fix? If so, then tests/unit/i18n/geography.spec.ts feels a little over-engineered to me, and it only covers zh-TW. Happy to keep it if you think it earns its place.

Sorry about that, and thanks for the pointers! Both are fixed now.

moven0831 (approved):
lgtm! nice work

pettinarip (approved):
LGTM! checked it in the preview, `zh-tw` and `te` pick the new names, and nothing else moves. Thanks!

PR #19132, opened 2026-08-21
Add SwiftNodes to the node services list

swiftnodes:
Adds SwiftNodes to the Popular node services list (the page invites missing providers). Inserted alphabetically between SettleMint and Tenderly, matching the existing entry format. All listed features are accurate: free tier without KYC, flat-rate plans, 75+ networks under one key, HTTP + WebSocket, archive on paid plans.

### DIFF
--- a/public/content/developers/docs/nodes-and-clients/nodes-as-a-service/index.md
+++ b/public/content/developers/docs/nodes-and-clients/nodes-as-a-service/index.md
@@ -352,6 +352,16 @@ Here is a list of some of the most popular Ethereum node providers, feel free to
     - Pay-per-hour pricing
     - Direct support
 
+- [**SwiftNodes**](https://swiftnodes.io/)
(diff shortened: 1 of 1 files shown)


Blondie0123 (approved):

swiftnodes:
Following up — this adds SwiftNodes to the node services list alongside the other public-RPC entries.

Two concrete things I can answer up front, since they're usually what stalls this kind of PR: the listed URL `https://rpc.swiftnodes.io/rpc/eth` answers without an API key (verified: `eth_chainId` → `0x1`), and there is no account requirement to use it, so it behaves like the other public endpoints already in the file rather than a trial link.

If this needs a second reviewer, an accessibility/content check, or a change in where it sits in the list, tell me which and I'll do it rather than re-bump.

Blondie0123 (approved):

PR #19139, opened 2026-08-21
Trim and dedupe in-repo Claude skills

myelinated-wackerow:
## Summary

- Audit-driven cleanup of `.claude/skills/` and `.claude/commands/` to cut agent context bloat: -76KB (13%), with each fact given one canonical home and pointers elsewhere
- Biggest per-use wins:
  - `review-translations.md` 33KB -> 25KB per invocation
  - design-system SKILL.md -32%
  - intl-pipeline SKILL.md -20%
  - one mandatory reference read removed from every performance-skill use
- Also fixes a dozen stale claims that actively misled agents:
  - forwardRef guidance contradicting the React 19 ref-as-prop rule
  - a phantom `Card decoration` API
  - recovery commands targeting the pre-`.manifests/` layout (silent no-op)
  - the MarkdownCard emoji-only claim
  - an image `quality` value not in the allow-list
  - dead doc links
  - a repo command writing into `$HOME`

🤖 _Generated by Fable 5_


pettinarip (approved):
LGTM

PR #19153, opened 2026-08-25
chore(ci): upgrade Chromatic to v18

pettinarip:
Chromatic now bills **zero** for builds that don't touch any story ([Aug 2026 changelog](https://www.chromatic.com/blog/chromatic-changelog-august-2026/)). Before, those still cost ~50 snapshots each. Needs CLI >= 17.7.0; we're on 16.0.0.

**44 of the last 97 merged PRs (45%) touch no UI code**, so they should become free.

## Changes

- `chromaui/action@v16` -> `@v18.5.0` in both Chromatic jobs (exact pin — `@v18` is mutable and bundles its own CLI)
- devDep `chromatic` `16.0.0` -> `18.5.0`
- `uploadMetadata: false` — it silently defaults to **on** when `onlyChanged` is set (CLI 17.6+), which would publish an unredacted debug log next to our public Storybook
- Fixed `ci.yml` comments + `docs/applying-storybook.md`, which claimed every non-draft PR publishes a build

Heads up: the lockfile regen also bumps semver 7.7.4 -> 7.8.5 on the webpack path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,10 +19,11 @@ env:
 # gating automatically via `needs:` (skipped jobs propagate).
 #
 # Exception: Storybook Chromatic (`visual-tests`) runs on every non-draft PR.
-# TurboSnap (`onlyChanged`) keeps snapshot usage low, so PRs with no affected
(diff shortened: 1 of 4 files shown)


wackerow (approved):
Looks good!

PR #19152, opened 2026-08-25
Deploy v11.22.0

pettinarip:
***
### ⚡️ Changes
- docs(defi): clarify streaming payments explanation (#18911) @BonganiN-KhulaIQ
- fix(ui): graceful image fallback for card thumbnails (#17936) @pettinarip
- fix: localize wallets and quizzes JSON-LD (#19022) @pillowtalk-Qy
- Update public/content/privacy/index.md (#18693) @jayproof
- patch(ui): full width card footer alert (#19053) @myelinated-wackerow
- Fix contradictory ToC card width caps (#19032) @myelinated-wackerow
- fix(content): correct slashing penalty figure (#19040) @myelinated-wackerow
- fix(layer-2): sub-cent tx cost rounding (#18966) @myelinated-wackerow
- Update Lido staking data: audits and client diversity numbers (#14825) @Sven-NOM

## 🌐 Translations

- feat: add Devcon 8 India banners (#19123) @myelinated-wackerow
- i18n: translation pipeline, devcon banner (all languages) (#19142) @wackerow
- fix(bug-bounty): repair dead deposit contract link, refresh stale links and copy (#19143) @nloureiro
- i18n: sync external descriptions - August 24 (#19145) @github-actions
- i18n: translation pipeline (all languages) (#19034) @wackerow
- i18n: sync external descriptions - August 10 (#19021) @github-actions

## 🔧 Tooling

- feat: add Devcon 8 India banners (#19123) @myelinated-wackerow
***
### 🦄 Contributors
Thank you @BonganiN-KhulaIQ, @Sven-NOM, @jayproof, @nloureiro, @pettinarip, @pillowtalk-Qy and @wackerow for the contributions! 🏆
***

PR #19164, opened 2026-08-27
fix(prepare-release): never leave a deploy PR without a published release

pettinarip:
## Why

v11.22.0 shipped to production on 2026-08-25 with **no GitHub release**. It went unnoticed for two days.

What happened:

1. The scheduled `weekly-release.yml` run died 1.1s into its first API call (`num_turns: 1`, `$0` spent). Discord got the failure alarm.
2. The release was then prepared locally. Tag `v11.22.0` was pushed, `staging` merged.
3. The `publish` step was blocked by a permission prompt.
4. The flow **continued to `create-pr` anyway** — the command doc orders publish (Step 7) before create-pr (Step 8) but never says "stop if publish fails".
5. The deploy PR (#19152) merged. Version live in prod, release notes never published, draft still sitting at `v11.21.1`.

A deploy PR without a published release is the one failure state that ships silently — the PR is mergeable by anyone and the gap is invisible unless someone opens the releases page.

## What this changes

- **`verify <version>`** — new script command asserting a published, non-draft release exists for a version. Exits non-zero with a clear message otherwise.
- **`weekly-release.yml`** — runs `verify` right after the deploy PR is captured, so a missing release fails the `prepare` job and trips the existing Discord alarm before the review job ever starts. Strips the `v` prefix, since the script adds its own.
- **Command doc** — Step 7 is now an explicit hard gate: if publish fails for any reason, stop, do not create the deploy PR, and report the exact state.
- **Stale-dev fix** (first commit, written 2026-08-20, never pushed) — `setup_worktree` checks out the *local* `dev`, which on a dev machine is usually stale; back-merging into it then rewinds the branch and the push is rejected as non-fast-forward. Both the Aug 25 run and a run today hit this. Now the release branches are fast-forwarded to their origin counterparts first, and it hard-errors if a local release branch has commits origin doesn't.

## Verification

- `bash -n` on the script; `yaml.safe_load` on the workflow
- `verify 11.21.0` → passes; `verify 11.22.0` → failed correctly while that release was missing, passes now that it is published
- `--dry-run verify` → no-ops without calling `gh`
- `--dry-run preflight` → confirms the branch now resets to `origin/dev` before back-merging

The missing v11.22.0 release has been published separately: https://github.com/ethereum/ethereum-org-website/releases/tag/v11.22.0

### DIFF
--- a/.claude/commands/prepare-release.md
+++ b/.claude/commands/prepare-release.md
@@ -123,11 +123,30 @@ EOF
 RELEASE_URL=$(./src/scripts/prepare-release.sh $DRY_RUN_FLAG publish "$VERSION" "$DRAFT_TAG" /tmp/claude/release-notes.md)
 ```
 
+**This step is a hard gate. If it fails for ANY reason — a `gh` error, or a
+permission prompt you cannot get approved — STOP HERE.** Do not run Step 8, and
+do not work around the block. Report to the user that the version was bumped and
+`staging` was merged, that the release is NOT published, and that the deploy PR
+was deliberately not created. A deploy PR without a published release is the one
+state that ships silently: the PR can be merged by anyone, and the missing
+release notes are invisible until someone checks the releases page.
+
+(This is exactly how v11.22.0 shipped on 2026-08-25 with no release.)
+
 ### Step 8: Create Deploy PR
 
 ```bash
 PR_URL=$(./src/scripts/prepare-release.sh $DRY_RUN_FLAG create-pr "$VERSION" /tmp/claude/release-notes.md)
 ```
+
+### Step 8b: Verify the Release Published
+
+```bash
+./src/scripts/prepare-release.sh $DRY_RUN_FLAG verify "$VERSION"
+```
+
+If this fails, say so loudly in the final report — the deploy PR must not be
+merged until the release exists.
 
 ### Step 9: Cleanup Worktree
 
--- a/src/scripts/prepare-release.sh
+++ b/src/scripts/prepare-release.sh
@@ -11,6 +11,7 @@ set -euo pipefail
 #   ./src/scripts/prepare-release.sh [--dry-run] fetch-draft        # Fetch draft release body
 #   ./src/scripts/prepare-release.sh [--dry-run] publish <version> <draft_tag> <body_file>  # Publish release
 #   ./src/scripts/prepare-release.sh [--dry-run] create-pr <version> <body_file>            # Create deploy PR
+#   ./src/scripts/prepare-release.sh [--dry-run] verify <version>   # Assert the release was published
 #   ./src/scripts/prepare-release.sh cleanup            # Remove worktree if created
 #   ./src/scripts/prepare-release.sh reset              # Reset worktree to clean state
 #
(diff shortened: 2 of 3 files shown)


src/scripts/prepare-release.sh
```diff
@@ -153,6 +154,21 @@ cmd_reset() {
...
+  local branch="$1"
+  local ahead
+  ahead=$(run_in_workdir git rev-list --count "origin/$branch..$branch")
+  if [[ "$ahead" -gt 0 ]]; then
+    log_error "Local $branch has $ahead commit(s) not on origin/$branch."
+    log_error "Release branches must mirror origin. Resolve manually."
+    exit 1
+  fi
+  run_in_workdir_or_dry git reset --hard "origin/$branch"
+}
```
myelinated-wackerow, line 170: Two bugs here, both fixed by the suggestion below.

**`git reset --hard` moves HEAD's branch, not `$branch`.** The guard reads the *branch ref* (`origin/$branch..$branch`) but the action targets *HEAD*, so they disagree about which branch is being protected:

```
before: staging=a877489  origin/dev=9d0dd03  HEAD-branch=staging
guard:  rev-list --count origin/dev..dev = 0   -> passes, proceeds
after:  staging=9d0dd03                         # staging clobbered to origin/dev
```

This is reachable in exactly the scenario this PR fixes. `git push origin staging` failing under `set -e` strands the worktree on `staging`, and `setup_worktree` only runs `git checkout dev` when the tree is *dirty* (L106-112) -- so the next run's `sync_branch_to_origin dev` at L195 silently rewrites local `staging` to `origin/dev`. Preflight then reports clean, `cmd_version` commits the bump and tags on `staging`, `git push origin dev --follow-tags` no-ops on both (the tag isn't reachable from `dev`), and `cmd_merge_staging` dies on the now-inverted guard. Nothing bad reaches origin, but it wedges the clone with a stranded local tag that blocks the next `pnpm version`.

**The `rev-list` fatals when the local branch doesn't exist** -- see the separate note on L286.

```suggestion
sync_branch_to_origin() {
  local branch="$1"
  local current ahead

  # No local ref yet (fresh clone, CI checkout): create it and stop.
  if ! run_in_workdir git show-ref --verify --quiet "refs/heads/$branch"; then
    run_in_workdir_or_dry git branch --force "$branch" "origin/$branch"
    return 0
  fi

  ahead=$(run_in_workdir git rev-list --count "origin/$branch..$branch")
  if [[ "$ahead" -gt 0 ]]; then
    log_error "Local $branch has $ahead commit(s) not on origin/$branch."
    log_error "Release branches must mirror origin. Resolve manually."
    exit 1
  fi

  # reset --hard moves HEAD's branch, so only reach for it when $branch IS HEAD.
  current=$(run_in_workdir git branch --show-current)
  if [[ "$current" == "$branch" ]]; then
    run_in_workdir_or_dry git reset --hard "origin/$branch"
  else
    run_in_workdir_or_dry git branch --force "$branch" "origin/$branch"
  fi
}
```

Checked against three cases: fresh clone with no local `staging` (dry-run completes, exit 0), HEAD stranded on `staging` while syncing `dev` (`staging` preserved, `dev` correctly fast-forwarded), and a genuinely-ahead local branch (guard still fires, exit 1).

src/scripts/prepare-release.sh
```diff
@@ -263,7 +281,9 @@ cmd_merge_staging() {
 
   log_info "Merging dev into staging..."
 
+  run_in_workdir git fetch origin
   run_in_workdir_or_dry git checkout staging
+  sync_branch_to_origin staging
```
myelinated-wackerow, line 286: `checkout staging` on L285 is dry-run-skipped while the `rev-list` inside `sync_branch_to_origin` is unconditionally real, so `--dry-run merge-staging` fatals on any clone without a local `staging` branch:

```
[DRY-RUN] Would run: git checkout staging
fatal: ambiguous argument 'origin/staging..staging': unknown revision or path not in the working tree
EXIT=128
```

Same failure at L203 whenever `master` is ahead of `staging`. Reproduced against this branch's script in a `--branch dev` clone -- it passes locally only if you already happen to have a local `staging`, which is likely why the `--dry-run preflight` check in the description didn't catch it. The suggestion on the function above covers this.

.claude/commands/prepare-release.md
```diff
@@ -123,12 +123,31 @@ EOF
...
+
+(This is exactly how v11.22.0 shipped on 2026-08-25 with no release.)
+
 ### Step 8: Create Deploy PR
 
 ```bash
 PR_URL=$(./src/scripts/prepare-release.sh $DRY_RUN_FLAG create-pr "$VERSION" /tmp/claude/release-notes.md)
 ```
 
+### Step 8b: Verify the Release Published
```
myelinated-wackerow, line 142: Non-blocking: Step 8b verifies *after* Step 8 creates the PR, so at the doc level the title is enforced by prose -- and prose is what failed on Aug 25. Moving this block ahead of Step 8 (as "Step 7b") turns it from detection into prevention, and costs nothing since `verify` is a no-op under `$DRY_RUN_FLAG`. The workflow's post-`capture_pr` placement is fine as intentional -- it fails the job and trips the Discord alarm before `wait_and_review`.

myelinated-wackerow (changes requested):
@pettinarip: the new "did we actually publish the release?" safety check works fine, so merging this does fix the silent-ship problem it was written for.

But the bundled cleanup has a bug that only bites when a release run has already failed partway: the script says "put branch X back where the server has it" while actually doing that to whatever branch happens to be checked out, so it can quietly overwrite your local `staging` with `dev`'s commits and leave the next release wedged -- plus `--dry-run` stops working on any fresh clone.

Details below: both bugs are in `sync_branch_to_origin`, and one suggestion fixes both. Nothing wrong with the verify work itself -- `cmd_publish` retags the draft to `v$VERSION` so `gh release view "v$VERSION"` is the right assertion, the workflow step inherits job-level `GH_TOKEN`, and an empty `version` out of `capture_pr` still fails the job loudly.

-- Opus 5

PR #19167, opened 2026-08-27
intl: restore daily cron, fix CJK length floor

myelinated-wackerow:
## Summary

Re-enables the daily 04:20 UTC cron on the intl pipeline, and fixes the one thing that would otherwise have made it fail most mornings.

The cost bug that took the cron off (run 31149083965, $1,108 unattended) is closed -- the input-side bounds landed in 0390ed18d9 / c14ff63804 / 6e519b4a3f, OpenRouter's server-side credit limit backs them, and five supervised dispatch runs have gone through since. Stacked runs don't re-translate: when `intl/pending-dev` already exists, the pipeline merges `dev` into it, syncs the working tree from it, and branches off pending, so drift detection reads the manifests that unmerged work already stamped and only genuinely changed sections are sent. Manifest coverage on `dev` is 9,838 of 9,840 pairs.

The remaining 2 are `whitepaper/index.md` for zh and zh-tw, which fail output validation on a flat 0.3 character-count floor that Chinese legitimately sits below -- measured min 0.34 whole-file, and the failing chunk is 0.245. Six recorded attempts across the two locales, from 2026-07-30 through the 2026-08-20 OpenRouter run, all landed in a 381-393 band against the same 1571-char chunk: stable correct output read as truncation, no manifest stamped, and the next run retranslating the whole 90KB file. CJK locales now use a 0.2 floor, everything else keeps 0.3.

RECITATION, the other recurring scheduled-run failure, needed nothing -- it's a Gemini-direct finish reason and went away with the move to OpenRouter.

One caveat: the fix clears the known blocker for those two pairs, but they have never completed end-to-end. A dispatch run with `base_branch: intl/restore-cron`, `target_path: public/content/whitepaper/index.md`, `target_languages: zh,zh-tw` would confirm before merge.

### DIFF
--- a/.claude/skills/intl-pipeline/SKILL.md
+++ b/.claude/skills/intl-pipeline/SKILL.md
@@ -5,7 +5,7 @@ description: Use when working on the translation pipeline (`src/scripts/intl-pip
 
 # intl-pipeline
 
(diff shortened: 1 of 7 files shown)


wackerow:
Marking draft while I attempt some intl-pipeline cleanup

PR #19182, opened 2026-08-31
Report unnamed EIPs on data-only upgrade syncs

pettinarip:
## Summary

- run EIP coverage on every store change, not only when classify finds a notable key
- comment on the standing content-review issue when there are unnamed EIPs, even if the week was testnet-date noise
- stay quiet when classify is empty and coverage has no gaps

Networking EIPs are printed but never trigger a comment on their own: a number-only match cannot see their wire names (`eth/72`), so they would sit in the list and comment every Friday.

Today the gap list is Glamsterdam EIP-8070 / 8136 / 8189 (networking, reported only) and Hegotá EIP-8141 (the one that raises the flag).

## Test plan

- `pnpm exec playwright test tests/unit/upgrades/coverage.spec.ts --project=unit`
- `GITHUB_OUTPUT` from `pnpm exec tsx src/scripts/sync-upgrades/coverage.ts` prints `gaps=true` against the current store

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/sync-upgrades.yml
+++ b/.github/workflows/sync-upgrades.yml
@@ -89,17 +89,23 @@ jobs:
             echo "Data-only change — notable keys unchanged."
           fi
 
-      # Snapshot of unnamed EIPs, not a diff. Runs on every store change so a
+      # Snapshot of the gap list, not a diff. Runs on every store change so a
       # data-only week still surfaces gaps classify ignores (testnet dates).
+      # Advisory only — a crash here must not cost the week's sync. Printed as
+      # well as saved, so a quiet week still leaves the list in the log.
       - name: Measure content coverage
         id: coverage
         if: steps.diff.outputs.changed == 'true'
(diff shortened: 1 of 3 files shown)


myelinated-wackerow (approved):
LGTM. Verified locally: 12/12 unit tests pass, lint clean, and the script reproduces the reported gap set (`gaps=true`, Glamsterdam 8070/8136/8189 networking + Hegotá 8141).

Nice-to-haves, none blocking:

- `coverage.txt` is discarded on a quiet week (`notable=false` + `gaps=false`), so the networking list doesn't actually print anywhere. `cat coverage.txt` after the redirect would fix it. Avoid a bare `| tee` — the default step shell is `bash -e` without `pipefail`, so it'd swallow a script failure.
- The comment at lines 101-102 says "unnamed EIPs" — it's unnamed *non-networking* EIPs, or a missing page.
- Snapshot, not a diff, so the same list re-posts every store-change week until 8141 is named. Reading that as intent, just flagging the cadence.

-- Opus 5/@wackerow

PR #19184, opened 2026-08-31
feat(roadmap): drive the release carousel from Forkcast data

nloureiro:
## Summary

The `/roadmap` release carousel described every upgrade from hardcoded values in `src/data/roadmap/releases.tsx` — a release date, a planned year, a display string and a Forkcast URL per entry — all of which the Forkcast-backed store from #18993 already knows. This moves the carousel onto that store.

Three things the user asked for, plus what the data made worth adding:

**The status tag now comes from Forkcast.** It was derived from whether a hardcoded date had gone past, which gave every unshipped fork "In development" in **error red**. Meanwhile that same fork's own page said "Testing on devnets" from the data. Stage derivation now lives in `src/lib/utils/upgrades.ts` and *both* surfaces read from it, so `/roadmap` and `/roadmap/glamsterdam/` can no longer describe the same upgrade two different ways.

**Dates come from the store**, at whatever precision Forkcast actually has, through `formatPartialDate`. This fixes live drift: Hegotá read `2027` against Forkcast's **Q2 2027**.

**What else the data supports**, kept to facts a general reader can act on:

- **Mainnet target with an explicit confidence clause.** Glamsterdam's "Q4 2026" was presented as settled; the store says `confirmed: false`, so it now reads `Q4 2026 · Date not yet confirmed`. Rendered as its own clause so translation word order can't turn it into a softer adjective.
- **Next milestone** — `Sepolia fork, September 28, 2026`. The most concrete "what happens next" fact available.
- **Scope** — `23 changes scheduled` / `13 changes included`. Nowhere on `/roadmap` did the size of an upgrade appear before.
- **The Forkcast link** is driven by `sourceUrl` instead of a duplicated hardcoded field.

Deliberately left out: ACD call references and per-EIP lists (`decidedAt`, `eips[]` detail). That is Forkcast's and the upgrade pages' job, not a roadmap summary's.

## Data-layer bug this surfaced

Driving the Forkcast button off `sourceUrl` exposed that **three of the seven URLs 404**. The normalizer built `sourceUrl` from upstream's `path`, but Forkcast populates `path` for every fork while rendering a page only for those it has not `disabled`:

| fork | upstream | `forkcast.org/upgrade/…` |
| --- | --- | --- |
| the-merge, shapella, dencun | `disabled: true` | **404** |
| pectra, fusaka, glamsterdam, hegota | `disabled: false` | 200 |

Fixed at the cause: the parser captures `disabled`, the normalizer emits `sourceUrl: null` for a fork with no page, and `UpgradeData.sourceUrl` is now `string | null` with both consumers guarding on it. The three values in `generated.ts` are patched to match what the fixed generator emits, so Friday's sync reproduces them rather than reverting them.

Nothing linked those URLs before this PR, so no dead link ever shipped — but `UpgradeSummary` would have hit them as soon as one of those forks were unshipped.

## Translation cost

Two new labels and two plural strings. The other five stage labels reuse keys that are already translated in all 25 locales, so switching the carousel onto a complete stage map costs 4 keys rather than 7.

## Notes for review

- `Release` collapses from a three-way union to one interface: with dates gone there is nothing left to make mutually exclusive. `releaseName` stays hardcoded on purpose — the store says "The Merge", the carousel says "Paris (The Merge)", and that is editorial, not volatile.
- `startIndex` and the timeline gradient are ported off date arithmetic onto stage, preserving the two-month post-activation grace period. Verified it still lands on Glamsterdam today.
- The store is imported into a client component (~1.5KB gzipped). Kept there rather than threaded as props from `page.tsx` so the carousel shares one formatting path with `UpgradeSummary`.
- The tag keeps its existing visual treatment (rounded rect, `text-sm font-bold`) rather than switching to `<Tag>`. Colour is now three tones by *confidence* — shipped / date locked / still moving — which is what removes the wrong red. Unifying the two surfaces on `<Tag>` is a design call worth making separately.

## Test plan

- `pnpm test:unit` — 1103 passing, including a new `tests/unit/roadmap/release-carousel.spec.ts` (every `upgradeSlug` resolves, order holds, no hand-maintained date fields creep back, every stage has a label that exists in English) and two new normalizer tests
- `pnpm type-check`
- `pnpm lint`
- Checked in a browser against the production build: all 7 slides, light + dark, 390px + 1400px. Tag contrast 14.9:1 and 17.1:1 in light mode, no horizontal overflow. Confirmed the rendered HTML links only the four Forkcast pages that return 200.


app/[locale]/roadmap/_components/ReleaseCarousel.tsx
```diff
@@ -15,58 +15,115 @@ import {
...
+          tone: UPGRADE_STAGE_TONES[stage],
+          stageLabel: t(UPGRADE_STAGE_LABEL_KEYS[stage]),
+          targetDate: upgrade.mainnetTarget.when
+            ? formatUpgradeDate(upgrade.mainnetTarget.when, locale, t)
+            : null,
+          targetConfirmed: upgrade.mainnetTarget.confirmed,
+          nextMilestone: next
+            ? `${formatMilestoneLabel(next, t)}, ${formatUpgradeDate(next.when, locale, t)}`
+            : null,
+          eipCount: upgrade.eips.length,
```
Comment (pettinarip) [inline app/[locale]/roadmap/_components/ReleaseCarousel.tsx:82]: `eips.length` is taken at face value here, but Forkcast only really tracks EIPs from Pectra onward — the older entries are backfilled stubs. Dencun's store entry holds just EIP-4844, so this slide will read "1 change included" for a fork that shipped nine. The Merge and Shapella have `eips: []` and render nothing, which makes the gap more visible rather than less: 12 changes on Pectra, 1 on Dencun, silence on Shapella.

`sourceUrl` is a clean discriminator — it's `null` for exactly the three forks Forkcast doesn't track (the ones upstream marks `disabled`, whose `/upgrade/*` pages 404):

```ts
eipCount: upgrade.sourceUrl ? upgrade.eips.length : 0,
```

Separately: `normalizeEips` deliberately keeps `considered` entries, so once Hegotá's EIPs are marked `considered` this line will announce them as "N changes scheduled". Probably want to count only `included`/`scheduled` statuses instead of the whole array.

app/[locale]/roadmap/_components/ReleaseCarousel.tsx
```diff
@@ -241,52 +232,112 @@ const ReleaseCarousel = () => {
...
-                      className="w-full lg:w-fit"
-                      variant={release.href ? "outline" : "solid"}
-                    >
-                      {t("page-roadmap-release-forkcast")}
-                    </ButtonLink>
-                  )}
+                    {/* Scale of the upgrade, in plain words rather than an EIP
+                        list — the upgrade page owns the per-EIP detail. */}
+                    {eipCount ? (
+                      <p className="mt-3 text-sm text-body-medium">
```
Comment (pettinarip) [inline app/[locale]/roadmap/_components/ReleaseCarousel.tsx:299]: Small UI thought, take it or leave it — I wondered whether "13 changes included" is clear enough on its own. Changes to what? I'm not sure a reader can tell whether that means EIPs, features, or something else, and the bare number doesn't give them an obvious next step.

A couple of ideas:

- Either say EIPs outright, or keep the plain-language "changes" and add a small info icon next to it explaining that a change is an EIP. Avoiding the jargon feels like the right instinct, it may just need a line of explanation to land.
- Would it be worth surfacing the changes themselves somewhere? Probably not in the carousel — the slide is a summary and already fairly dense — but maybe in the upgrade page content, so the count becomes a way in rather than a dead end.

Nothing blocking here, just curious what you think.

src/lib/utils/upgrades.ts
```diff
@@ -0,0 +1,164 @@
...
+    (milestone) => milestone.kind === "mainnet"
+  )
+  if (mainnet?.status === "complete" || upgrade.status === "live") return "live"
+
+  if (upgrade.mainnetTarget.confirmed) return "scheduled"
+
+  const running = upgrade.milestones.find(
+    (milestone) => milestone.status === "live"
+  )
+  if (running?.kind === "testnet") return "testnet"
```
pettinarip, line 47: I think this branch is unreachable, so the `testnet` stage never renders.

Testnet milestones can't carry `status: "live"` — the sync maps Forkcast's date-confidence wording through `TESTNET_STATUS` (`actual → complete`, `proposed → anticipated`, `projected → projected`), so `"live"` isn't in the vocabulary. Only devnets get it, via `isLatest && status !== "live"` in `normalizeMilestones`. So this `.find()` can only ever return a devnet, and `page-roadmap-upgrade-status-phase-testnet` is dead as written.

The effect on today's data: Glamsterdam's public testnet already forked (Platåberget, 2026-08-13, `complete`) but devnet-9 is still `live`, so both `/roadmap` and the upgrade page say "Testing on devnets" — and will keep saying it through Sepolia (Sep 28) and Hoodi (Oct 26), right up to activation.

Underneath it the two files seem to disagree about what a testnet is: the sync treats the fork as an event that either happened or hasn't (`complete` is fair enough), while this function assumes it's a phase you sit in for weeks. If that's right, the fix belongs here rather than in the sync — take the furthest-advanced *completed* milestone instead of the live one. Platåberget being `complete` is exactly the signal that Glamsterdam has reached public testnets.

Minor, same lookup: `.find()` returns array order and `KIND_ORDER` sorts devnet → testnet → mainnet, so even if a testnet could be `live`, a live devnet would still win — not the "strongest claim first" the docstring promises.

src/lib/utils/upgrades.ts
```diff
@@ -0,0 +1,164 @@
...
+  live: "shipped",
+  scheduled: "locked",
+  testnet: "moving",
+  devnet: "moving",
+  development: "moving",
+  planning: "moving",
+  research: "moving",
+}
+
+export const isUpgradeShipped = (upgrade: UpgradeData) =>
```
pettinarip, line 91: `isUpgradeShipped` has no callers. Meanwhile `UpgradeSummary.tsx:47-50` still does its own `mainnet?.status === "complete"` check and the carousel writes `stage === "live"` inline four times — same question, three spellings.

In `UpgradeSummary` the helper is also the more correct one (it catches `upgrade.status === "live"` too), and `mainnet` isn't used anywhere else in the file, so `if (isUpgradeShipped(upgrade)) return null` drops four lines. Worth using it there, or dropping it.

src/lib/utils/upgrades.ts
```diff
@@ -0,0 +1,164 @@
...
+  getUpgradeStage(upgrade) === "live"
+
+/**
+ * The next milestone a reader has not seen yet, excluding mainnet activation.
+ *
+ * `live` is happening now rather than next, so a running devnet is not the
+ * answer to "what comes next". Mainnet is excluded because both surfaces state
+ * the target separately, and repeating it as the next milestone is noise.
+ */
+export const getNextMilestoneBeforeMainnet = (
```
pettinarip, line 101: Heads up: `src/data/upgrades/index.ts` already exports `getNextMilestone(slug)` — the same logic minus the `kind !== "mainnet"` clause — plus an unused `getUpgrade(slug)`. Both have zero callers on dev too, so this is pre-existing dead code, not something the PR broke.

No objection to the split itself — data access vs presentation derivations reads right. Just seems a good moment to delete those two so there's one of each.

app/[locale]/roadmap/_components/ReleaseCarousel.tsx
Comment (pettinarip) [inline app/[locale]/roadmap/_components/ReleaseCarousel.tsx:82]: Related to the count — Forkcast's own Upgrade Overview for Glamsterdam shows **18 Scheduled for Inclusion**, where we'd show 23. The gap is exactly the 5 networking EIPs (7975, 8070, 8136, 8159, 8189).

Upstream treats `Networking` as a status of its own, separate from `Scheduled for Inclusion`; our normalizer folds it into `"scheduled"` and keeps the distinction in the `networking` flag. So Forkcast's 18 is the scheduled bucket alone.

Should we match it? I'd lean yes — the flag is already there for exactly this:

```ts
upgrade.eips.filter((e) => e.status === "scheduled" && !e.networking).length
```

Glamsterdam is the only fork with networking EIPs today, so it's the only number that changes.

Comment (pettinarip) [reply app/[locale]/roadmap/_components/ReleaseCarousel.tsx:82]: Image

pettinarip (changes requested):
Left a few notes inline — two look like real bugs (Dencun's EIP count and the unreachable `testnet` stage), the rest are minor. Overall a good simplification.

PR #18103, opened 2026-05-01
content(apps): remove orphan Scout Game description key

nloureiro:
## Summary

Removes the orphan `app-scout-game-description` translation key from `src/intl/en/page-app-descriptions.json`.

The Scout Game listing on `/apps/` is broken — `scoutgame.xyz` no longer resolves to any host (verified in #18049 with `dig` / `curl`). Users clicking the card hit a connection failure.

## Change

```diff
-  "app-scout-game-description": "The Scout Game is designed to reward individuals for identifying and supporting emerging developer talent within onchain ecosystems...",
```

One line removed, between `app-sandbox-description` and `app-self-description`. JSON validity confirmed.

## ⚠️ Out-of-repo follow-up still required

The apps page is fed by a Google Sheet (the source of truth — see `GOOGLE_SHEET_ID_DAPPS` in `src/data-layer/fetchers/fetchApps.ts`). The Scout Game row in that sheet must also be removed, otherwise the next data-layer fetch may surface the entry again with a missing description.

Refs #18049.

PR #19185, opened 2026-08-31
content(roadmap): correct the roadmap slides against Forkcast

nloureiro:
Audited every slide on `/roadmap` against Forkcast's EIP records — and, where a claim was about live network state, against mainnet itself. Three slides were wrong. This fixes those and writes the missing Hegotá content.

| slide | verdict |
| --- | --- |
| Paris (The Merge) | ✅ accurate |
| Shapella | ✅ accurate |
| Dencun | ✅ accurate |
| **Pectra** | ❌ future tense 15 months after shipping; one overstatement |
| **Fusaka** | ❌ a factually wrong figure; 4 orphan keys |
| Glamsterdam | ✅ accurate |
| **Hegotá** | ❌ said only "Proposals are currently under discussion" |

## Fusaka: a wrong figure, not just a stale one

> "Default gas limit increase to ~60M (from current 45M)"

`git log -S` dates this to **2025-08-28** — three months before Fusaka activated, so it was written as a forecast and was true then. Mainnet now runs **exactly 60,000,000** (checked at block 25,877,090), so the increase already happened, and 45M is two revisions behind: EIP-7935's own baseline was 36M, mainnet passed through 45M, and Fusaka shipped 60M.

Now: *"Raised the default gas limit to 60M, increasing L1 execution capacity"* — what the fork set, which cannot drift. Phrasing a validator-chosen parameter as "current" in a static string guarantees this recurs.

The other two Fusaka bullets are correct (EIP-7825's cap is 16,777,216 = 2²⁴; PeerDAS ✓).

## Fusaka: 4 orphan keys

`page-roadmap-fusaka-additional-*` — a "Potential Additional Features" heading and three bullets — is referenced by **no component**. Verified against every `t()` call in `releases.tsx` and by grepping `src/` and `app/`. That is **100 translated strings across 25 locales for copy no reader has ever seen**.

Worse, one bullet — "Support for secure enclaves on mobile devices" — is EIP-7951, which is `Included` in Fusaka. It was describing a delivered feature as potential.

Same class as the orphan key removed in #18103. The English keys are deleted here; unused-key detection is a site-wide problem and is left for a later process (no `slide-copy.spec.ts`).

## Pectra: tense and one overstatement

> "The blob count **will be** increased from 3 to 6 targets, with a maximum of 9"

Numbers check out against EIP-7691 (target 6, max 9, up from 3/6) — but Pectra activated **2025-05-07**.

> "Stakers can now choose an **arbitrary** amount of ETH to stake"

Not arbitrary: EIP-7251 caps `MAX_EFFECTIVE_BALANCE` at **2048 ETH**, up from exactly 32.

## Hegotá: the content was a placeholder

The slide said "Proposals are currently under discussion" and the page said "One further proposal is under consideration and around twenty more have been proposed, but only the headliner is scheduled". I counted Hegotá fork relationships across all 618 upstream EIP records:

| status | count | |
| --- | --- | --- |
| **Scheduled** | **2** | EIP-7805 (FOCIL, headliner) · EIP-8141 (Frame Transaction) |
| Proposed | 62 | |
| Declined | 3 | |
| Withdrawn | 1 | |

EIP-8141 moved `Considered → Scheduled` at **acde/244 on 2026-08-27**. So "only the headliner is scheduled" is false, "under consideration" is stale, and "around twenty" is off by 3x. Our store already had both at `status: "scheduled"`; only the prose lagged.

The slide now covers both, three bullets each, plus an explicit note that scope is not final since Hegotá is still `planning` upstream. The page gains a `## Frame transactions` section whose EIP chip renders `SCHEDULED FOR INCLUSION` from the store.

Copy is written **from** Forkcast's `laymanDescription` / `northStarAlignment` / `stakeholderImpacts`, not lifted. Upstream targets the ACD audience — 7805's reads "FOCIL empowers multiple validators to mandate the inclusion of specific transactions in each block, thereby improving the network's censorship resistance properties" — so each bullet is rephrased in the voice of the existing Glamsterdam copy.

No hardcoded proposal count in the prose: the store only carries `considered` / `scheduled` / `included`, so a "62" here could never be test-verified and would rot. "Dozens more have been proposed" stays true.

## Page content

The Pectra and Fusaka errors appeared **verbatim** on `/roadmap/pectra/` and `/roadmap/fusaka/` too ("Currently, network targets… will be increased to 6"; "above today's 45M"). Those sentences are corrected — fixing only the slides would have left the pages contradicting them.

## Guard

**`eip-tags.spec.ts`** extended to resolve *every* `<EipTag />` in roadmap content, not just pages using `**Resources**` blocks (the Hegotá page deliberately doesn't use one). Verified it fails on a typo'd id.

## What I did not change

Glamsterdam shows 2 of 23 scheduled EIPs with no "scope can change" note, even though `mainnetTarget.confirmed` is false and its own page carries such a caveat. Defensible for a summary, so I left it — flagging it as a judgment call for you.

`/roadmap/fusaka/` still has other sections written in pre-fork planning tense. Fixing the whole page is a larger editorial pass than this PR; only the factually wrong sentence is corrected here.

## Translation

9 new English keys, 3 corrected, 6 removed — English only. The obsolete translations stay in the other 24 locales until the intl-pipeline removes them; no code references them, so they are inert. Per `CLAUDE.md` I did not hand-propagate.

## Relationship to #19184

Independent — verified with a trial merge, **no conflicts either order**. Once #19184 lands you get a free cross-check: its panel prints "2 changes scheduled" for Hegotá straight from the store, beside this PR's prose.

## Test plan

- `pnpm test:unit` — including the EipTag resolve check (verified to fail on a typo'd id)
- `pnpm type-check`, `pnpm lint`, `pnpm lint:md` (0 errors, 343 files)
- Checked all seven slides and both pages in a browser against the production build; asserted the corrected strings render and none of the four stale ones survive anywhere in the HTML
- Verified every internal link target exists (`/glossary/#account`, `/roadmap/account-abstraction/`, `/roadmap/security/#quantum-resistance`, `/roadmap/fusaka/`)

**Content review wanted** — the facts are Forkcast's and mainnet's, but the wording is mine.


pettinarip (approved):
LGTM. Pulled the slide-copy unused-key test — that’s a site-wide problem, not something to special-case on the carousel. We’ll deal with orphan keys separately. The English key deletions and the EipTag check can stay.

PR #19194, opened 2026-09-01
feat(search): migrate site search to self-hosted Typesense

myelinated-wackerow:
## Summary

Replaces Algolia DocSearch with self-hosted Typesense, one collection per locale, plus the pipeline that builds and publishes the index.

Measured against 45 labelled queries drawn from 30 days of real Algolia search logs:

| | hit@1 | hit@3 | hit@10 | MRR |
| --- | --- | --- | --- | --- |
| Algolia (production today) | 33% | 40% | 60% | 0.39 |
| Typesense, untuned | 42% | 58% | 70% | 0.51 |
| Typesense as shipped | **67%** | **72%** | **84%** | **0.71** |

Most of Algolia's misses are video transcripts outranking canonical pages -- `privacy`, `smart contracts`, `defi` and `nonce` all return `/videos/...` today rather than the topic's own page.

All 25 locales have already been built, promoted and curated through this pipeline: 51,727-52,310 documents each, 63/63 curation rules resolved in every language, zero promotion refusals.

## How it works

Three steps per locale, run as independent matrix jobs five at a time, so one locale can never block another:

1. **Scrape** -- `typesense/docsearch-scraper` crawls production into `ethereumorg-staging-<locale>`. Ranking hints come from the `docsearch:*` meta tags added in #19166.
2. **Promote** -- swaps the locale's alias onto the new collection, but only if it holds >=90% of the live document count, has a sortable `pagerank` field, returns documents for its own language, and clears a hit@1 floor on the labelled query set. On refusal the alias is untouched, so search serves slightly stale results rather than none.
3. **Curate** -- applies the pinned results in `typesense/curation.json`.

A single combined index took ~16h to crawl, past the Actions job limit. Per-locale collections take ~25 min each; a full 25-locale rebuild measured 2h08m.

## Notes for review

- **`typesense/curation.json` must live in the repo**, not a dashboard. Typesense pins by document id, and the scraper assigns ids as sequential counters that change on every crawl, so dashboard pins would silently stop matching after the next run. Storing URLs and resolving them at promote time is what keeps them working.
- **The scraper is a version behind, and it bites on the second run.** Typesense v30 replaced per-collection synonyms and overrides with synonym sets and curation sets; scraper 0.11.0 still calls the removed v29 endpoints -- but only when the staging alias resolves a previous collection. A locale therefore crawls cleanly the first time and then dies with a 404 at commit on every run after, having indexed everything and published nothing. The workflow clears that alias before each crawl, which makes every run look like a first run. Found by running all 25 locales twice; a single-locale smoke test would have shipped it and the first weekly cron would have failed every language.
- **Ranking is `_text_match(buckets: 100):desc,pagerank:desc`.** The bucket granularity was measured rather than guessed: 100 gives 67% hit@1, 50 gives 65%, 25 gives 60%, and 4 collapses to 21%.
- The workflow matrix and the scripts both derive their locale list from `i18n.config.json`.

## Before merging

- [ ] Point Netlify's deploy-succeeded notification at a `repository_dispatch` of type `deploy-succeeded`. Deferred rather than blocking: Netlify's basic webhook cannot set GitHub's auth header, so this needs a build plugin and a PAT. The weekly Friday cron covers freshness in the meantime.
- [ ] Remove the Algolia env vars from Netlify (after merge, once this is confirmed live).

Actions secrets, Netlify browser env, and CORS on the Typesense server are all confirmed working.

## Operational gaps worth knowing

Search no longer has a managed fallback. Algolia was a hosted service; if `search.buidl.org` is unreachable the modal returns nothing and there is no second source. The indexing workflow reports to Sentry Crons with one monitor environment per locale -- verified against both real successes and real failures -- but nothing yet watches the search endpoint itself.

Relevance is gated only at promote time, and only in English: the labelled query set has no non-English equivalent, so the other 24 locales are checked for structure and size but not for quality.

## Deferred

`pagerank` weighting so root-level pages outrank tutorials and videos; page-level results appearing above their own `#hash` sub-items; stemming, to retire the pinned singular/plural pairs; and the 0.12.x scraper image, which may fix the v29 endpoint calls properly but would invalidate the relevance baseline above and so belongs in its own change.

Separately, a `0x` query should offer a block-explorer result rather than searching the corpus for it. That one is not curation -- an address is an unbounded query, so there is nothing to pin -- but a pattern match in the modal that injects a synthetic hit, which needs no index change.


nloureiro:
Screen Shot 2026-09-10 10 14 35 PM

very small details, do want to center the icons on the window?

nloureiro:
too much space on safari

Screen Shot 2026-09-10 10 15 58 PM

nloureiro:
did a few serches but no history on the search window
Screen Shot 2026-09-10 10 17 40 PM

nloureiro:
search for "llm" and got these results, and the one highlighted is a 404 link
Screen Shot 2026-09-10 10 19 36 PM

nloureiro:
search for Vitalik and got this big transcript from the video. is it expected

Screen Shot 2026-09-10 10 21 56 PM

nloureiro:
now the AI Q&A
===

<head></head><h2 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">QA review — #19194 (Typesense migration)</h2><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Tested<span class="Apple-converted-space"> </span><code>deploy-preview-19194</code><span class="Apple-converted-space"> </span>against production (Algolia), plus direct queries to<span class="Apple-converted-space"> </span><code>search.buidl.org</code><span class="Apple-converted-space"> </span>using the exact parameters the modal sends. Everything below is reproducible from the preview.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><strong>Params observed in flight (for reference):</strong></p><pre style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code>collection: ethereumorg-en
group_by:   url          group_limit: 3
sort_by:    _text_match(buckets: 100):desc,pagerank:desc
query_by:   hierarchy.lvl0..lvl6,content
</code></pre><hr style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">What works</h3><ul style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><li>Modal,<span class="Apple-converted-space"> </span><code>⌘K</code>, arrow-key navigation,<span class="Apple-converted-space"> </span><code>Enter</code>,<span class="Apple-converted-space"> </span><code>Esc</code>, recent searches, and the "no results / try searching for" state all behave the same as production. The vendor footer is gone as intended.</li><li>CORS is live, the browser key is correctly scoped read-only (<code>GET /collections</code><span class="Apple-converted-space"> </span>with it returns<span class="Apple-converted-space"> </span><code>401</code>), and<span class="Apple-converted-space"> </span><code>GET /health</code>returns<span class="Apple-converted-space"> </span><code>{"ok":true}</code>.</li><li>Latency measured over 30 queries:<span class="Apple-converted-space"> </span><strong>68–143 ms</strong>, one outlier at 372 ms. Comparable to Algolia (47–78 ms) and well inside typeahead budget.</li><li>Locale collections resolve and return locale-correct URLs — I confirmed 22 (<code>en ar bn cs de es fr hi id it ja ko pl pt-br ru sw tr uk ur vi zh zh-tw</code>), each with its own prefixed results.</li><li><code>ethereumorg-en</code><span class="Apple-converted-space"> </span>reports<span class="Apple-converted-space"> </span><strong>52,310 documents</strong>, matching the PR.</li></ul><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">The headline win is real</h3><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">The video-transcript problem is gone. Same query, production vs preview:</p>
Query | Production (Algolia) #1 | Preview (Typesense) #1
-- | -- | --
privacy | /videos/why-privacy-matters/ | /privacy/
smart contracts | /videos/smart-contracts-code-is-law/ | /smart-contracts/
defi | /videos/defi-history-inception-to-2021/ | /defi/
nonce | /videos/ethereum-everywhere-santiago-palladino/ | /developers/docs/transactions/
gas | /apps/gaszip/ | /gas/
nft | /videos/learn-nfts-and-defi/ | /nft/
layer 2 | /videos/layer-2-scaling-explained/ | /layer-2/
dao | /videos/dao-hack-ethereum-classic/ | /apps/categories/dao/

<p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Typo handling is the one I would prioritise.<span class="Apple-converted-space"> </span><code>typesenseSearchParameters</code><span class="Apple-converted-space"> </span>being empty is a deliberate choice in this PR and I agree with the reasoning, but<span class="Apple-converted-space"> </span><code>walet</code><span class="Apple-converted-space"> </span>→ not-the-wallets-page is the kind of miss users notice.<span class="Apple-converted-space"> </span><code>num_typos</code><span class="Apple-converted-space"> </span>/<span class="Apple-converted-space"> </span><code>typo_tokens_threshold</code><span class="Apple-converted-space"> </span>/<span class="Apple-converted-space"> </span><code>drop_tokens_threshold</code><span class="Apple-converted-space"> </span>are the obvious first knobs, and the labelled query set is already there to measure them.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Also:<span class="Apple-converted-space"> </span><code>/contributing/style-guide/…</code><span class="Apple-converted-space"> </span>ranks #1 for<span class="Apple-converted-space"> </span><code>zk rollup</code><span class="Apple-converted-space"> </span>and #2 for<span class="Apple-converted-space"> </span><code>the merge</code>. Contributor documentation probably should not compete with reader content — either exclude<span class="Apple-converted-space"> </span><code>/contributing/</code><span class="Apple-converted-space"> </span>from the crawl or give it a low<span class="Apple-converted-space"> </span><code>pagerank</code>.</p><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Smaller things</h3><ul style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><li><strong>Recent searches are heavy.</strong><span class="Apple-converted-space"> </span>Each entry stores the full<span class="Apple-converted-space"> </span><code>_rawTypesenseHit</code>, including page content and highlight markup — I measured<span class="Apple-converted-space"> </span><strong>9,354 bytes for a single entry</strong>. DocSearch keeps up to 7, so that is ~65 KB of<span class="Apple-converted-space"> </span><code>localStorage</code><span class="Apple-converted-space"> </span>per locale, most of it duplicated page text. Trimming in<span class="Apple-converted-space"> </span><code>transformItems</code><span class="Apple-converted-space"> </span>would fix it.</li><li><strong>Row labels use the page description, not the page name.</strong><span class="Apple-converted-space"> </span><code>staking</code><span class="Apple-converted-space"> </span>returns the right page but the clickable row reads "Earn rewards while securing Ethereum" under a group header of "Ethereum staking: How does it work?". Correct behaviour given the<span class="Apple-converted-space"> </span><code>docsearch:*</code><span class="Apple-converted-space"> </span>meta tags from #19166, but worth confirming it is the intended reading experience.</li><li><strong>Monitoring.</strong><span class="Apple-converted-space"> </span><code>https://search.buidl.org/health</code><span class="Apple-converted-space"> </span>is publicly reachable and returns<span class="Apple-converted-space"> </span><code>{"ok":true}</code><span class="Apple-converted-space"> </span>— that closes the "nothing watches the search endpoint" gap in the PR description with an uptime check rather than a new PR.</li><li><strong>Locale list.</strong><span class="Apple-converted-space"> </span>My probe found 22 collections; the PR says 25. Worth confirming the three I did not guess exist and that the set matches<span class="Apple-converted-space"> </span><code>i18n.config.json</code><span class="Apple-converted-space"> </span>exactly, since a missing collection fails closed with<span class="Apple-converted-space"> </span><code>Collection not found</code><span class="Apple-converted-space"> </span>and an empty modal.</li></ul><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Not covered</h3><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Dark mode rendering, RTL locales (<code>ar</code>,<span class="Apple-converted-space"> </span><code>ur</code>) in the modal, and behaviour when<span class="Apple-converted-space"> </span><code>search.buidl.org</code><span class="Apple-converted-space"> </span>is unreachable.</p>

pettinarip (commented):
Nice, left a few comments. The ones I'd fix before merging: the CJK fields need `locale` set or zh/ja/ko barely search at all, and prune can delete the index we'd want to roll back to.

Rest is minor, nothing blocking.

src/scripts/typesense/client.ts
```diff
@@ -0,0 +1,99 @@
...
+ * cannot search, and the search key is the reverse. Passing the wrong one produces a 401
+ * that reads like a credentials problem rather than a scoping one.
+ */
+
+import i18nConfig from "../../../i18n.config.json"
+
+const trimSlash = (s: string) => s.replace(/\/+$/, "")
+
+export const TYPESENSE_URL = trimSlash(process.env.TYPESENSE_URL ?? "")
+export const ADMIN_KEY = process.env.TYPESENSE_API_KEY ?? ""
```
pettinarip, line 14: Small naming mismatch: this reads `TYPESENSE_API_KEY` / `TYPESENSE_API_SEARCH_KEY`, but `src/global.d.ts`, `.env.example` and `typesense/README.md` all use `TYPESENSE_ADMIN_KEY` / `TYPESENSE_SEARCH_KEY`. CI is unaffected since the workflow remaps the secret, but following `.env.example` locally fails with `Missing environment variables: TYPESENSE_API_KEY`, which also makes the manual recovery steps in the README not work first time. Could we settle on one pair of names?

src/scripts/typesense/client.ts
```diff
@@ -0,0 +1,99 @@
...
+export const SEARCH_KEY = process.env.TYPESENSE_API_SEARCH_KEY || ADMIN_KEY
```
pettinarip, line 15: The fallback to `ADMIN_KEY` works against the note just above about the admin key not being usable for search — if `TYPESENSE_API_SEARCH_KEY` is unset we get a 401 at query time instead of a clear startup error. Dropping `|| ADMIN_KEY` and adding the search key to `requireEnv` would surface it earlier.

typesense/scraper-config.json
```diff
@@ -0,0 +1,126 @@
...
+  "scrape_start_urls": true,
+  "custom_settings": {
+    "field_definitions": [
+      {
+        "name": "anchor",
+        "type": "string",
+        "optional": true
+      },
+      {
+        "name": "content",
```
pettinarip, line 39: I don't think the CJK locales will tokenize correctly here. These custom `field_definitions` don't set `locale`, and the workflow doesn't pass `TYPESENSE_COLLECTION_LOCALE`, so zh/ja/ko `content` and `hierarchy.lvl*` get split on whitespace only — a search on /zh/ for something like 智能合约 comes back near-empty.

What makes this one worth catching before merge is that the promote gates all still pass (the relevance gate is English-only and the size gate is happy), so the index would ship looking green.

Setting `.locale` on `content` and the `hierarchy.lvl*` fields in the jq step should do it, mapping `zh-tw` -> `zh` and `pt-br` -> `pt`.

src/scripts/typesense/promote.ts
```diff
@@ -0,0 +1,251 @@
...
+    const i = argv.indexOf(flag)
+    return i === -1 ? undefined : argv[i + 1]
+  }
+  const locale = get("--locale")
+  return {
+    locales: argv.includes("--all") ? [...LOCALES] : locale ? [locale] : ["en"],
+    dryRun: argv.includes("--dry-run"),
+    force: argv.includes("--force"),
+    minRatio: Number(get("--min-ratio") ?? MIN_SIZE_RATIO),
+    keep: Number(get("--keep") ?? KEEP_PER_LOCALE),
```
pettinarip, line 75: These go through `Number()` unchecked, so a typo or an empty value becomes `NaN` and fails quietly in opposite directions: `NaN` for `keep` makes `mine.slice(NaN)` return the whole array, so prune deletes every predecessor, and `NaN` for `minRatio` makes the `ratio < args.minRatio` comparison always false, disabling the shrink gate. A `Number.isFinite` check on both would be enough.

src/scripts/typesense/promote.ts
```diff
@@ -0,0 +1,251 @@
...
+const promoteLocale = async (
+  locale: string,
+  collections: CollectionInfo[],
+  args: Args
+): Promise<boolean> => {
+  const alias = `ethereumorg-${locale}`
+  const source = newestStaged(collections, locale)
+  if (!source) {
+    console.log(`  ${locale}: no staged collection -- skipped`)
+    return true
```
pettinarip, line 97: Returning `true` here means a locale whose crawl produced nothing is counted as promoted in the summary, so a silently failed scrape reads as a successful run. Worth returning `false` (or tracking skips separately) so it shows up.

src/scripts/typesense/promote.ts
```diff
@@ -0,0 +1,251 @@
...
+  let hits = 0
+  for (const { q, correct } of queries) {
+    const params = new URLSearchParams({
+      q,
+      query_by:
+        "hierarchy.lvl0,hierarchy.lvl1,hierarchy.lvl2,hierarchy.lvl3,hierarchy.lvl4,content",
+      group_by: "url_without_anchor",
+      group_limit: "1",
+      per_page: "1",
+    })
```
pettinarip, line 194: These search params don't include `sort_by`, but the site sends `_text_match(buckets: 100):desc,pagerank:desc` (`src/components/Search/index.tsx`). So the gate is scoring a ranking configuration we never actually serve, which also means the 35% floor isn't comparable to the numbers measured with the sort applied. Passing the same `sort_by` the modal uses would make the gate measure production.

src/scripts/typesense/promote.ts
```diff
@@ -0,0 +1,251 @@
...
+      (n) =>
+        n.startsWith(`ethereumorg-${locale}_`) ||
+        n.startsWith(`ethereumorg-staging-${locale}_`)
+    )
+    .filter((n) => n !== live)
+    .sort()
+    .reverse()
+
+  // `live` occupies one of the kept slots, so only keep-1 predecessors survive.
+  const doomed = mine.slice(Math.max(keep - 1, 0))
```
pettinarip, line 225: I think this can delete the collection we'd want to roll back to. `mine` is newest-first over the leftovers, so after a run that was refused by a gate the newest leftover is the rejected staging collection — it survives, and the previous known-good index is what gets pruned. The live one is correctly never touched, but the README's "live plus one predecessor, rollback is a single alias flip" wouldn't hold after any refusal.

Keeping whichever collection the alias pointed at before the swap, and pruning the rest, would be more predictable.

src/components/Search/index.tsx
```diff
@@ -46,29 +87,49 @@ const Search = ({ asChild = false, children }: SearchProps) => {
...
+  const collectionName = `${collectionPrefix}-${locale}`
 
   const searchModalProps = {
-    apiKey,
-    appId,
-    indexName,
-    onClose,
-    searchParameters: {
-      facetFilters: [`lang:${locale}`],
+    typesenseCollectionName: collectionName,
```
pettinarip, line 107: For a locale that hasn't been promoted yet this collection name resolves to a missing alias, and Typesense answers 200 with the 404 nested inside `results[0]` rather than failing the request. That slips past the adapter's error guard, so the modal stays in its idle state with no error UI and throws on each keystroke. A check on the inner result status would let us show the normal error state.

.github/workflows/typesense-index.yml
```diff
@@ -0,0 +1,179 @@
...
+        run: |
+          # Guarded here rather than in `if:`, which is evaluated before step env exists.
+          [ -n "$SENTRY_DSN" ] || { echo "no SENTRY_DSN; monitoring disabled"; exit 0; }
+          # DSN is https://<public_key>@<host>/<project_id>
+          KEY="${SENTRY_DSN#*//}"; KEY="${KEY%%@*}"
+          REST="${SENTRY_DSN#*@}"; HOST="${REST%%/*}"; PROJECT="${REST##*/}"
+          echo "SENTRY_CRONS=https://$HOST/api/$PROJECT/cron/typesense-index/$KEY/" >> "$GITHUB_ENV"
+          curl -sS -m 10 -X POST \
+            "https://$HOST/api/$PROJECT/cron/typesense-index/$KEY/?environment=${{ matrix.locale }}" \
+            --data-raw '{"status":"in_progress","monitor_config":{"schedule":{"type":"crontab","value":"0 22 * * 5"},"checkin_margin":120,"max_runtime":60}}' \
```
pettinarip, line 87: This POST doesn't capture the returned check-in id, and the final step (~L178) calls the same base URL with `?status=`, which registers a *second* check-in rather than closing this one. The `in_progress` one is then left to hit its `checkin_margin` and alert. Saving the id from this response and appending it to the URL in the result step should link them up.

wackerow:
@nloureiro 

> very small details, do want to center the icons on the window?

Addressed in #19231 (this PR targets base migration from Algolia to Typesense.. the other does more polishes)

> too much space on safari

Seeing same, will patch

> did a few serches but no history on the search window

Unable to replicate this. Can you clarify what browser you're using? Are you following through to links, or just typing a search and then clearing? It only holds the links you follow through (existing behavior). Is this working for you in production?

> search for "llm" and got these results, and the one highlighted is a 404 link

This is because of the preview deployment environment. The index is of production, but the preview deploy uses mock data for the actual apps we load on /apps/*. Shouldn't be an issue in production.

> search for Vitalik and got this big transcript from the video. is it expected

Looking into this. Yes, it should catch video transcripts, but they shouldn't display that much text. 19231 attempts to de-prioritize the "video" results as well

myelinated-wackerow:
Thanks both -- this round addresses all of @pettinarip's inline comments and the @nloureiro items that belong to this PR. Eight of the nine land here; one was already fixed in #19231, and the rest are queued there.

**Fixed in this PR**

| Comment | Change |
| --- | --- |
| Key naming mismatch | The scripts now read the documented `TYPESENSE_ADMIN_KEY` / `TYPESENSE_SEARCH_KEY`, so `.env.example` works as written and the workflow's remap is gone. Nothing changes in Netlify or Actions. The scraper container keeps `TYPESENSE_API_KEY` -- that's its own contract. |
| `\|\| ADMIN_KEY` fallback | Dropped, and the search key is in `requireEnv`. It only turned a missing variable into a 401 per keystroke. |
| CJK tokenization | `locale` is now set per crawl on `content` and `hierarchy.lvl*` -- `ja`, `ko`, `zh` (`zh-tw` maps onto it), plus `ru` and `uk`, which Typesense also has customizations for. Verified, numbers below. |
| `Number()` unchecked | Both flags refuse a non-finite value. You were right that it failed in whichever direction was worse. |
| Skipped locale counted as promoted | Now its own outcome, reported separately and exiting non-zero. Promote runs straight after that locale's crawl, so nothing staged means the crawl produced nothing. |
| `prune` deleting the rollback target | The collection the alias pointed at before the swap is now held explicitly. |
| Missing alias, no error UI | Failed queries reach the error screen the translations were already wired for. Covers the unreachable-server and revoked-key cases too, which were equally silent. |
| Sentry second check-in | The start step captures the check-in id and the result step closes that one, falling back to the old behaviour if no id comes back. |
| Safari row heights (@nloureiro) | `--docsearch-hit-height: fit-content` was reaching the vendor's `.DocSearch-Hit-Tree`, which is an `<svg>` -- `fit-content` on a replaced element is undefined, so Safari resolved it enormous. Only child rows render that connector, which is why the parent row looked fine. |
| Eight-line transcript rows (@nloureiro) | Titles clamp to two lines, paths to one. See the note below on why the row was that tall. |
| Stale Algolia references | `docs/stack.md` named Algolia as the search stack; two comments pointed at the Algolia package and its styling docs. |

**On the long transcript rows**

For a content hit, DocSearch renders the matching text *as* the title line and the page heading beneath it -- so the transcript paragraph landing on the title line is by design, not a mix-up. What made it eight lines is that we override the vendor's one-line truncation with `whitespace-normal`, so a wrapped title isn't cut mid-word, and unbounded that lets a whole paragraph through.

Worth flagging what turned up while checking: Typesense already returns a short snippet for these hits -- around 50 characters against a 300-character full value -- and the renderer uses it for some hits and falls back to the full value for others. That is why rows now sit at one or two lines rather than a uniform height: the short ones are the snippet path, the two-line ones are the fallback being clamped. The clamp bounds the symptom; always using the snippet is the real fix, and it lives in the search adapter rather than in CSS. Queued in #19231.

**CJK, measured**

Re-crawled `ja ko zh zh-tw ru uk` through the pipeline. Same query, against the collection the alias pointed at before versus after:

| Query | Before | After |
| --- | --- | --- |
| 智能合约 (zh) | 108 | 492 |
| 以太坊 (zh) | 132 | 652 |
| イーサリアム (ja) | 118 | 633 |

One correction: "near-empty" was a little strong -- the old index did return matches. But roughly four fifths of the matching content was unreachable, so the defect is real. That comparison was only possible because the prune fix in the same round kept the previous collection; under the old logic it would already have been deleted.

Worth underlining the point about why this was catchable only by hand: every gate passed green. The size and field checks pass, and the relevance gate is English-only, so an index that barely searches in four locales promotes without complaint. That gap is listed in the PR description and just stopped being theoretical.

**Already fixed in #19231**

- `sort_by` missing from the relevance gate. You're right that the floor wasn't comparable to numbers measured with the sort applied.
- Centering the modal icons.
- `/contributing/` outranking reader content, and video transcripts dominating -- the merged `pagerank` change puts both in the lowest tier. Still reproducible on this preview until a full re-crawl runs.

**Queued in #19231**

- Typo tolerance. Agreed it's the most user-visible gap. One caveat on measuring it: all 43 labelled queries are pinned by curation, so the query set can't currently tell us whether a change helped. That needs fixing first or we would be tuning blind.
- Recent searches storing the full raw hit (~9.4 KB each).
- Row labels reading the page description rather than the page name.
- Using the snippet Typesense already returns, per the note above.

**Not defects**

- The `llm` 404 is the preview's mock app data; production serves the real pages.
- 22 vs 25 collections -- all 25 exist; the three missing from the probe were guessed locale codes.
- `search.buidl.org/health` being public is the right answer to the monitoring gap, and is an uptime-check task rather than code.

**Not verified here:** the prune and Sentry paths need a refused run and a live check-in to exercise. Still open from your list, @nloureiro -- which browser were you on for the missing search history, and were you following results through to the page? It only records hits you open.

-- Opus 5

myelinated-wackerow:
**Correction on the snippet note above.**

I said Typesense returns a short snippet and the renderer displays the full field instead, and queued "use the snippet" for #19231. That was wrong, and the queued item doesn't exist.

The snippet _is_ what renders. `Snippet.js` prefers `_snippetResult[attr].value`, and the adapter fills that from Typesense's `highlight.<field>.snippet` whenever the field matched -- the full value is only reached when a field has no highlight at all.

That case doesn't occur in practice. Across 30 queries, 0 of 297 group leaders were a `content` row without a content highlight, because a content record only leads its group by matching on content. And no rendered row exceeded 115 characters under #19231's parameters (317 rows) or 81 under the ones this PR ships (320 rows), across 32 queries.

So the one-vs-two-line variation I described above is snippet length, not two renderer paths. The eight-line row you saw was real, but I can't reproduce it against the current `ethereumorg-en` index under either parameter set -- and the index has been re-crawled since -- so I'm not going to attribute it to a cause I can't demonstrate. @nloureiro, if you can still hit it, the query would settle it.

The two-line clamp stays as the bound. It's doing less work than I claimed, but it's what holds a long word or a CJK title where a character count can't.

-- Opus 5

myelinated-wackerow:
## Changes since review, and where each comment landed

**Ranking and parity.** The relevance table in the original body was measured on the labelled query set, and every query in that set is also a curation rule, so it scored pins rather than ranking. Re-measured on the live collections against production Algolia, with the parameters the modal sends (details now in the body): pages searched by their own title, 25 locales, hit@1 233/248 vs Algolia 222; unpinned glossary terms, 8 locales, top result agrees with Algolia on 83/200; the 653 localized forms of the curated head terms land on the curated page 650 times vs Algolia's 268. Non-English head terms were the regression this fixes: a pin matches its string exactly and every rule was English, so 钱包 on /zh/ ranked plainly and landed on /zh/security/. Rules now name an ETHGlossary term, `pnpm typesense:localize-curation` writes each term's forms in every locale to `typesense/curation-aliases.json` (generated, committed, run by hand when rules change), and curate pins them beside the English query. Ranking is `max_weight` at 100 buckets; the coarser buckets tried in between were reverted after measuring, see 17c04d9d14. Search params are shared between the modal and the promote gate in `src/lib/utils/searchParams.ts`.

**Curate fixes found while applying the pins live.** Item ids were a `[a-z0-9-]` slug, so every non-Latin query collapsed to one id (0e9b9f80ae). Typesense stores an exact rule's text with punctuation removed but splits the typed query on it, so rules written `erc-20` or `crypto.com` matched nothing anyone types; queries are now stored with punctuation collapsed to spaces (4e01480b64, 0ec474508b). The scripts read `.env.local` (52e7f6e943). A final review pass found two manual-run cases in promote: a re-run without a new scrape pruned the rollback predecessor, and one locale throwing under `--all` cut the others off (bba1910ed2).

**Review comments**

| thread | resolution |
| --- | --- |
| client.ts:14, key names | One pair everywhere, `TYPESENSE_ADMIN_KEY` / `TYPESENSE_SEARCH_KEY`, matching the Actions secrets: 0651bcb01a |
| client.ts:15, admin-key fallback | Dropped; search key added to `requireEnv`: 0651bcb01a |
| scraper-config.json:39, CJK tokenization | Workflow sets `locale` on `content` and `hierarchy.lvl*` for ja, ko, ru, uk, zh (zh-tw onto zh): 3b7481a454, d9a4eff20e. Live check: 智能合约 on /zh/ returns 487 hits, /zh/smart-contracts/ first |
| promote.ts:75, NaN flags | `Number.isFinite` on both, exits naming the flag: b64bafe3ba |
| promote.ts:97, empty crawl counted as promoted | Distinct `skipped` outcome, fails the run: b64bafe3ba |
| promote.ts:216, gate scores a ranking we never serve | Gate sends the modal's `query_by`, `sort_by`, `text_match_type` from the shared module: 608aa0cad6 |
| promote.ts:225, prune deletes the rollback target | Previously-live collection held explicitly and reserved: b64bafe3ba |
| index.tsx:130, nested 404 leaves the modal idle | `transformSearchClient` rethrows as `RetryError`, the one name the modal shows an error screen for: b0be7cb64e |
| typesense-index.yml:87, second Sentry check-in | Start step captures the check-in id, result step closes it: b64bafe3ba |

CI is green and the deploy preview is up. All 25 collections carry the new curation sets as of today.

-- Fable 5.1

myelinated-wackerow:
Since the comment above: a spot check of "developers" landed on the Elixir guide, because the landing page's og:title ("Ethereum Developer Resources") lacks the word and og:title outweighs h1. Measured over all 66 root-level pages searched by their slug words, the shipped ranking placed 57 first (Algolia 25); re-weighting h1 above og:title fixed this one and broke four others, so the remaining misses are pinned instead, with singulars and glossary terms (9e53f67ba3, 80 rules). Live now: 65/66, the miss being the `10years` slug. Curation sets on all 25 collections are current.

-- Fable 5.1

wackerow (approved):
@pettinarip Moving forward with this to unblock #19231 and #19300. Per usual, please feel free to ping me with any additional feedback or issues if you spot anything, happy to iterate and bug squash.

PR #19210, opened 2026-09-02
fix(sentry): drop unlocatable stacks

pettinarip:
Generalises one rule in `getDropReason()` so third-party client noise stops needing a new `ignoreErrors` regex per error message.

## Why

`instrumentation-client.ts` is, by design, the noise gate for the Sentry pipeline — `feeds/sentry-feed.mjs` on the recovery box says so explicitly:

> Client-side extension/adware noise is already filtered upstream in instrumentation-client.ts (denyUrls / ignoreErrors / beforeSend); anything reaching Sentry has passed that gate, so this list stays short by design.

But the gate's only general-purpose tool was a message list, so every new wording of the same third-party failure produced another PR. There were **13 open `[recovery]` PRs doing exactly that**, all touching this one file, so merging any one conflicted the rest. Sentry's server-side *Error Message* filter would be the alternative home, but it's Business-plan gated and unavailable on this account.

## What changed

**`unattributable-overflow` → `unattributable`.** The rule already detected the right shape — frames present, none carrying a location — but gated it behind `Maximum call stack size exceeded|too much recursion`. Removing the message gate makes it general.

Safe because `nextjsClientStackFrameNormalization` rewrites our own frames to `app:///` before `beforeSend` runs, and `app:///` passes `hasUsableLocation`. A stack where *no* frame has any location therefore cannot be ours — browsers withhold URLs for cross-origin scripts. This keeps the file's "positive evidence it isn't ours" invariant.

A genuinely frameless event is still **kept**: absence of frames is not evidence, and that residue is what a message rule would still be needed for.

That is the whole diff. The docstring's carve-out is gone too, so the invariant is unconditional again.

## Two rules were removed after review

The first version of this PR also added an SDK-frame discount and an `unsupported-browser` engine rule. Both were **removed** — neither could ever fire.

Credit to @wackerow for catching it. Root cause was shared: the fixtures were transcribed from payloads read *out* of Sentry, which are post-enrichment and post-symbolication, so they didn't reflect what `beforeSend` actually receives. The unit tests couldn't catch it because they hand-built the event objects to match those same downstream shapes.

**`isSdkFrame`** looked for `/@sentry/` in a frame path. That path is produced by Sentry resolving the sourcemaps we upload (`widenClientFileUpload: true`); in the browser the SDK is bundled into our own chunks, and `eventbuilder.js` parses frames off a synthetic `Error`'s stack, so the frame arrives as `app:///_next/static/chunks/<hash>.js`. Filename cannot separate the SDK's own frame from ours once `rewriteFrames` has run — `mechanism.synthetic` is the only real discriminator, and using it needs a captured fixture first.

**`unsupported-browser`** read `contexts.browser.name`. No `@sentry/*` package writes that field client-side — Relay derives it server-side from the User-Agent — so it is `undefined` at `beforeSend` and the rule never fired. ETHORG-1AY is already archived forever in Sentry, so nothing regresses by dropping it. If an engine rule is ever wanted, the UA *is* available: `httpContextIntegration` (default, `sdk.js:34`) sets `event.request.headers["User-Agent"]` in `preprocessEvent`, which runs at `client.js:708` inside `_prepareEvent` — before `beforeSend` at `client.js:827`.

Waterfox was also in that engine list and shouldn't have been: it tracks Firefox ESR, is actively maintained, and would have suppressed real errors.

## Tests

`tests/unit/sentry/filter-event.spec.ts` — 16 pass. Existing overflow cases retargeted to `unattributable`, plus new coverage for:

- an unlocatable stack with an arbitrary (non-overflow) message
- one locatable frame among unlocatable ones → kept
- a frameless event with a noisy message → kept (documents the deliberate residue)

`pnpm type-check` and `pnpm exec eslint` clean.

## Follow-ups (not in this PR)

- Once this ships, the `ignoreErrors` **message array** can likely be trimmed hard — but only after checking Sentry for what still arrives, since the rule can't see zero-frame events.
- Any future rule that reads event fields needs a real fixture first: `console.log(JSON.stringify(event))` in `beforeSend` on a deploy preview, or Sentry's raw stack-trace view. That one missing fixture caused both removals above.
- ETHORG-1AV is not handled by this PR (`isSdkFrame` was its intended fix). It needs the `mechanism.synthetic` approach and a captured payload.
- The agent still opens one PR per Sentry issue; that's an agent-side change, tracked separately.

### DIFF
--- a/src/lib/sentry/filter-event.ts
+++ b/src/lib/sentry/filter-event.ts
@@ -41,6 +41,12 @@ const NO_LOCATION = new Set(["", "undefined", "null", "<anonymous>"])
 const hasUsableLocation = (filename?: string, absPath?: string): boolean =>
   [filename, absPath].some((v) => v !== undefined && !NO_LOCATION.has(v))
 
+// For a non-Error throw the SDK synthesizes an exception and prepends its own
+// eventbuilder frame. It resolves to app:///node_modules/@sentry/..., so left
+// in it would count as a location and hide that nothing else can be placed.
+const isSdkFrame = (filename = "", absPath = ""): boolean =>
+  [filename, absPath].some((v) => v.includes("/@sentry/"))
+
 // Not app:/// -- nextjsClientStackFrameNormalization rewrites our own frames to
 // that scheme before beforeSend runs, so it marks first-party code too.
 const isThirdPartyLocation = (filename = "", absPath = ""): boolean =>
@@ -99,7 +105,9 @@ export function getDropReason(event: ErrorEvent): DropReason {
   // ETHORG-1AC's `AutoScroll` scraper). Our own frames are always rewritten to
   // app:///, so an event where *no* frame has a location cannot be ours. A
   // genuinely frameless event is kept: absence of frames is not evidence.
-  const frames = values.flatMap((v) => v.stacktrace?.frames ?? [])
+  const frames = values
+    .flatMap((v) => v.stacktrace?.frames ?? [])
+    .filter((f) => !isSdkFrame(f.filename, f.abs_path))
   if (
     frames.length > 0 &&
     !frames.some((f) => hasUsableLocation(f.filename, f.abs_path))
--- a/src/lib/sentry/filter-event.ts
+++ b/src/lib/sentry/filter-event.ts
@@ -4,7 +4,8 @@ import type { ErrorEvent } from "@sentry/nextjs"
 export type DropReason =
   | "extension-payload"
   | "extension-throw"
-  | "unattributable-overflow"
+  | "unattributable"
+  | "unsupported-browser"
   | "wallet-provider"
   | null
 
@@ -27,7 +28,11 @@ const EIP_1193_CODES = new Set([4001, 4100, 4200, 4900, 4901])
 const isProviderErrorCode = (code: number): boolean =>
   EIP_1193_CODES.has(code) || (code >= -32768 && code <= -32000)
 
-const OVERFLOW_MESSAGE = /Maximum call stack size exceeded|too much recursion/i
+// Abandoned Firefox/Pale Moon forks that fail on modern baseline features, so
+// nothing they report is actionable here. Sentry's legacy-browser inbound
+// filter only knows the mainstream engines and never matches these.
+const UNSUPPORTED_BROWSER =
+  /^(?:Basilisk|Pale ?Moon|Waterfox|SeaMonkey|K-Meleon)$/i
 
 // window.onerror reports cross-origin scripts with no URL, which the SDK
 // stringifies into a frame filename.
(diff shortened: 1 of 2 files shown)


myelinated-wackerow (approved):
pettinarip looks good for merge. nothing blocking but a few things appear to be inert though, so just wanted to flag:

**1. `isSdkFrame` won't match at `beforeSend` time.**

It looks for `/@sentry/` in `filename`/`abs_path`, but `app:///node_modules/@sentry/core/...` is a *symbolication* artifact -- it only exists after Sentry resolves the sourcemaps we upload (`widenClientFileUpload: true`, `next.config.js:322`). In the browser the SDK is bundled into our own chunks, so the frame the client actually sends is `app:///_next/static/chunks/<hash>.js`.

So ETHORG-1AV should still be arriving. Fails safe (keeps the event), just doesn't drop it.

**2. `contexts.browser` is never populated client-side.**

No `@sentry/*` package writes it -- the default integration list is inboundFilters / functionToString / conversationId / browserApiErrors / breadcrumbs / globalHandlers / linkedErrors / dedupe / httpContext / cultureContext / browserSession, and the only context writers in the browser SDK are `culture`, `trace`, `response` and webWorker. `contexts.browser` is derived server-side by Relay from the User-Agent, so `browserName` is `undefined` in `beforeSend` and `unsupported-browser` never fires.

The UA *is* there though: `httpContextIntegration.preprocessEvent` sets `event.request.headers["User-Agent"]` from `navigator.userAgent` before `beforeSend` runs. An unanchored match against that header (`/Basilisk\/|Pale ?Moon\/|SeaMonkey\//`) should get ETHORG-1AY.

Same root cause for both: the payloads pulled out of Sentry are post-enrichment and post-symbolication, so they don't reflect what `beforeSend` sees. The unit tests can't catch it since they hand-build the event. Easiest way to get a real fixture is a temporary `console.log(JSON.stringify(event))` in `beforeSend` on a preview, or Sentry's raw stack-trace view.

**3. Minor:** Waterfox is still actively maintained (Firefox 115 base), so it doesn't really fit "abandoned forks that fail on modern baseline features" -- that entry would drop real, actionable errors.

The `unattributable` generalization itself checks out, and that's the part that unblocks the other `[recovery]` PRs. `rewriteFrames` does run before `beforeSend` (`prepareEvent` -> `beforeSend` in `core/client.js`), and with no `assetPrefix`/`basePath`/`crossOrigin` in `next.config.js` our chunks are same-origin, so "no frame has a location => not ours" holds. ETHORG-1AC's shape is exactly what `_enhanceEventWithInitialFrame` produces for a cross-origin `onerror`, so that one will fire live.

-- Opus 5/@wackerow

wackerow (approved):

pettinarip:
Good catches, both rules are out.

You were right about the payloads being post-symbolication. Checked the SDK: `eventbuilder` parses frames off a synthetic Error's stack, so the SDK frame arrives as `app:///_next/static/chunks/<hash>.js` and `isSdkFrame` could never match. And `contexts.browser` has no client-side writer in any `@sentry/*` package, so `unsupported-browser` never fired either. Waterfox was just wrong of me.

Kept only the `unattributable` generalization, which is the part that unblocks the recovery PRs. ETHORG-1AV goes back on the list — `mechanism.synthetic` looks like the right lever, but I want a real fixture off a preview before writing that one.

PR #19211, opened 2026-09-02
fix(intl): normalize Intl whitespace through one shared helper

pettinarip:
Consolidates #19191, #19195 and #19201 into one change with a shared helper.

## Why

Node's ICU and the browser's ICU disagree on which space character to emit around formatted parts — U+202F narrow no-break space vs U+00A0 vs a plain U+0020, varying by locale, engine and ICU version. The results render identically, so nothing looks wrong, but React's hydration check compares strings and bails out into a full client re-render.

`formatDateRange` and `formatDateRangeToParts` already worked around this inline. Three separate recovery-agent PRs then proposed adding the same `.replace(/\s+/g, " ")` plus a six-line comment to three more formatters — which would have left five inline copies of one rule.

## What changed

`normalizeIntlSpaces` added to the existing `src/lib/utils/intl.ts`, and all five call sites routed through it:

| formatter | file | was |
| --- | --- | --- |
| `formatDate` | `date.ts` | unnormalized (#19191) |
| `formatDateRange` | `date.ts` | inline regex |
| `formatDateRangeToParts` | `date.ts` | inline regex |
| `getLocaleTimestamp` | `time.ts` | unnormalized (#19195) |
| `formatCompactNumber` | `numbers.ts` | unnormalized (#19201) |

## Which of these were actually broken

I measured the raw `Intl` output under Node 22 rather than assuming:

```
compact 123M (en-US)   "123M"        clean
compact 123M (ar)      "123 مليون"   EXOTIC U+00A0
compact 123M (ur)      "۱۲٫۳ کروڑ"   EXOTIC U+00A0
compact 123M (fr)      "123 M"       EXOTIC U+00A0
compact 123M (de)      "123 Mio."    EXOTIC U+00A0
compact 123M (ru)      "123 млн"     EXOTIC U+00A0
date long  (all)                     clean
date+time  (all)                     clean
```

So **`formatCompactNumber` is a real mismatch today** in 5 of 8 locales tested. The date formatters come out clean on Node's side — which is precisely the ICU-72 hazard rather than a reason to skip them: Node emits U+0020 where a browser on newer ICU emits U+202F before AM/PM, and only normalizing *both* sides makes them agree. A one-sided probe can't see the browser half.

## Tests

New `tests/unit/utils/intl.spec.ts` — 14 pass. Covers each space character ICU actually varies (U+202F, U+00A0, U+2009, U+3000, U+205F), run collapsing, and pass-through of already-clean strings. It also asserts every one of the five formatters emits no exotic whitespace, so a formatter added later without normalization fails the suite rather than shipping a silent hydration bailout.

`pnpm type-check` and `pnpm exec eslint` clean. Full `pnpm test:unit` shows 1103 passing and 20 failures, all in `tests/unit/data-layer/getters.spec.ts` — I verified those 20 fail identically on unmodified `dev`, so they are pre-existing and unrelated.

## Note on `wallets.ts`

`src/lib/utils/wallets.ts:103` normalizes with an explicit `[ -   　]` class. `\s` is a strict superset of that (it also covers U+00A0), so that call site should move onto the helper too — but #19202 is currently in review against the same lines, so it is left alone here to avoid a conflict. Worth a one-line follow-up once that lands.

Closes #19191, closes #19195, closes #19201.

### DIFF
--- a/src/lib/utils/date.ts
+++ b/src/lib/utils/date.ts
@@ -3,6 +3,8 @@ import type { PartialDate } from "@/data/upgrades/types"
 import { DEFAULT_LOCALE } from "../constants"
 import type { Lang } from "../types"
 
+import { normalizeIntlSpaces } from "./intl"
+
 /**
  * A wrapper for Intl.DateTimeFormat that enforces Web3 date standards.
  * - Forces the Gregorian calendar universally.
@@ -65,12 +67,14 @@ export const formatDate = (
   if (!isValidDate(date)) {
     return ""
   }
-  return dateTimeFormat(locale, {
-    month: "long",
-    day: "numeric",
-    year: "numeric",
-    ...options,
-  }).format(new Date(date))
+  return normalizeIntlSpaces(
+    dateTimeFormat(locale, {
+      month: "long",
+      day: "numeric",
+      year: "numeric",
+      ...options,
+    }).format(new Date(date))
+  )
 }
 
 /**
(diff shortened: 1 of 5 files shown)


wackerow (approved):
@pettinarip lgtm, nice clean up thanks!

PR #19218, opened 2026-09-03
chore(find-wallet): conclude the FindWalletCatalog2026 A/B test

pettinarip:
The rebuilt catalog won `FindWalletCatalog2026`, so it ships to everyone. This removes the losing arm and the per-experiment wiring, and keeps the generic A/B infrastructure for the next test.

- deletes the coded route, `LegacyWalletsBody` and `FindWalletProductTable`
- renders `WalletsPageBody` directly from the find-wallet page
- drops the flag, the `abTestRoutes` entry and the mock experiment
- drops the legacy-only helpers in `lib/utils/wallets.ts` and 10 orphaned `en` intl keys

No user-visible change: everyone already lands on the catalog once the experiment is stopped in Matomo. Emptying `abTestRoutes` stops the proxy rewriting this path, so this is safe to merge in either order.

**After deploy:** purge the Cloudflare cache for `/wallets/find-wallet/` — variant HTML was edge-cached.

Verified: `type-check`, `lint`, 1109 unit tests, and a prod build that emits `/wallets/find-wallet` as SSG with no `ab-code` routes.


wackerow (approved):
@pettinarip Lgtm, couple non-blocking notes on some orphaned code:

**`src/lib/utils/wallets.ts:151-202`** — `filterFn` and `getActiveFilterKeys` outlived their only consumer (`FindWalletProductTable/index.tsx`). Dropping them also orphans the `ChainName`, `FilterOption`, and `WalletRow` imports at `:14-19` (lint will flag those), plus the `WalletRow` type itself at `src/lib/types.ts:905`. Worth noting since the PR body claims the legacy-only helpers here were dropped.

**`src/components/ProductTable/FilterInputs/CheckboxFilterInput.tsx`** — whole file; consumers were `useWalletFilters.tsx` and `Layer2SelectInput.tsx`. `SwitchFilterInput` beside it is still live via `Layer2NetworksTable`.

**`src/lib/ab-testing/constants.ts:15`** — `decodeABCode` has no caller now. Keeping it as the inverse of `encodeABCode` (`proxy.ts:125`) is fine; flagging so it's a choice.

**Follow-up issue, not this PR:** presets are now unused — `Layer2NetworksTable/index.tsx:58` passes `presetFilters={[]}`, stranding `ProductTable/PresetFilters.tsx`, the preset branch in `src/lib/product-table/index.ts`, and `TPresetFilters`/`WalletPersonas`.

-- Opus 5/@wackerow

pettinarip:
Thanks @wackerow — applied note 1: dropped `filterFn` + `getActiveFilterKeys`, the orphaned `ChainName`/`FilterOption`/`WalletRow` imports, and the `WalletRow` type.

Leaving the rest for now: `CheckboxFilterInput` and the preset machinery are likely reusable in the follow-up work on the new catalog, and `decodeABCode` is the inverse of `encodeABCode` that the next coded page will need. If they turn out not to be, I'll clean them up then.

PR #19231, opened 2026-09-04
feat(search): block explorer results and page-level grouping

myelinated-wackerow:
## Summary

Search v2, stacked on #19194. Pasting an address or a hash now offers block explorers instead of nothing, and results lead with the page rather than a section anchor.

**Block explorers.** A `0x` value is never answerable from site content. Ethereum Mainnet and the eight featured L2s Blockscout covers get a row each, then Starkscan and Zircuit as their own sections. An EIP-3770 prefix (`base:0x…`) narrows it to one row.

We never probe explorers to work out which chain a value belongs to -- that would hand the user's address to every one of them on each keystroke. All networks are offered and the user picks, which they can do because they know where the value came from.

**Result grouping.** The library grouped by anchor-scoped `url`, so one page could occupy several rows without its own record among them. Grouping by page, restoring the heading-depth signal that overriding `sort_by` had dropped, and sorting the page record to the front of its group together mean a page leads with its sections nested beneath.

That also settles a disagreement worth flagging: the promote gate already scored against `url_without_anchor` while the app grouped by `url`, so the two did not agree on what a result was. The relevance numbers in #19194 described something users never saw. Worth re-running the scorer against the shipped parameters once v1 is live.

## Notes for review

- **Two patches to `typesense-docsearch-react`**, both for caps with no prop to reach them. `MAX_QUERY_SIZE` of 64 is applied twice -- as the input's `maxLength` and as a `slice` on every keystroke -- so a 66-character transaction hash was silently truncated and never matched; raised to autocomplete-core's own default of 512. `groupBy` caps every section at five rows as a stand-in for `distinct`, which is right for content results but drops real destinations here, so items may opt out. A unit test fails loudly if the first patch stops applying.
- **`src/data/explorers.ts` is generated** by joining chainid.network (EIP-3770 short names) with Blockscout's registry (URLs). Neither alone is enough. The chain list stays hand-curated in `constants.ts` -- deriving it from the registry's 745 entries would start endorsing explorers for chains the site says nothing about.
- **Starknet gets its own section** because a field element below 2^251 is the same shape as an Ethereum hash. Explorers disagree on padding, so 56 hex digits and up is accepted; leading zeros cascade at one in sixteen, putting the miss rate near one in 130 million. Starkscan has no unified search route and a felt gives no hint which it is, so contract and transaction are both offered. It is closed source, which we would rather avoid -- the one open-source Starknet explorer is archived and stuck on 0.13.2.
- **Rows keep the library's own markup** so the save and remove buttons on Recent entries survive, and are decorated from CSS. Results rows show the network with the external arrow at the row end; Recent rows also show the truncated value, since the input is empty there and "Base" alone would not say what was looked up.

## Known gaps

A page's sections can still render without their parent if the page's own `lvl1` record is not among the rows returned for it. Guaranteeing it needs `min_indexed_level: 0` in the scraper config and a re-crawl, so it waits on #19194 shipping.

The four new `docsearch-explorer-*` strings are English-only until the intl pipeline runs. They render in English everywhere meanwhile rather than as raw keys -- `request.ts` merges the default-locale messages underneath every locale.

Clicking an explorer row writes the full address into `localStorage` as part of the Recent entry. That was already true before this change, since the URL carries the value, but the feature makes it routine.

## Deferred

`pagerank` weighting is a separate PR off `dev`, since it only touches `metadata.ts` and needs a deploy before a re-crawl can pick it up. Stemming and the 0.12.x scraper image both need a re-crawl and wait on #19194 too.


src/data/chains.ts
wackerow, line 1: Note here, the update-chains.ts workflow script was used to generate these. The info is included in the same source that these networks are pulled from. `shortName` can be used as a prefix for an address or hash in search to sepcify the exact network

nloureiro:
AI Q&A review
=====

<head></head><h2 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">QA review — #19231 (block explorer results + page-level grouping)</h2><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Tested<span class="Apple-converted-space"> </span><code>deploy-preview-19231</code><span class="Apple-converted-space"> </span>against<span class="Apple-converted-space"> </span><code>deploy-preview-19194</code><span class="Apple-converted-space"> </span>and production. Both previews query the same live<span class="Apple-converted-space"> </span><code>ethereumorg-en</code>collection, so every difference below is client-side and attributable to this PR.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><strong>Params observed in flight, diffed against #19194:</strong></p><pre style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code>group_by:  url_without_anchor      (was: url)
sort_by:   _text_match(buckets: 100):desc,pagerank:desc,item_priority:desc
input maxLength: 512                (was: 64)
</code></pre><hr style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Grouping: confirmed, and it is the bigger of the two features</h3><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code>privacy</code><span class="Apple-converted-space"> </span>on #19194 returns five rows that are all the same page (<code>/privacy/</code>,<span class="Apple-converted-space"> </span><code>#getting-started</code>,<span class="Apple-converted-space"> </span><code>#privacy-can-be-the-default</code>, …). Here it returns five distinct pages with their sections nested underneath. Across a 30-query set, every result list became distinct pages instead of anchor spam.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">This is what makes v1 shippable in my view — I would not ship #19194 without it.</p><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Block explorers: works, across every shape I threw at it</h3>
Input | Result
-- | --
40-hex address | "Search this address on Blockscout" — 9 network rows + Zircuit section
66-char tx hash | "Search this hash on Blockscout" — correctly not truncated
base:0x… | single Base row
oeth:0x… | single Optimism row
vitalik.eth | "Search this name on Blockscout" plus site content results
57 / 62 hex felt | Starkscan only, contract + tx rows
64 hex felt | Blockscout + Starkscan + Zircuit (12 rows)
leading/trailing whitespace | trimmed, recognised
0x alone, foo:0x… | falls through to content search / no-results, correct

<p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Links verified: correct per-network hosts,<span class="Apple-converted-space"> </span><code>target="_blank"</code>,<span class="Apple-converted-space"> </span><code>rel="noopener noreferrer"</code>, and an "(opens in a new tab)" label for screen readers.<span class="Apple-converted-space"> </span><code>⌘K</code><span class="Apple-converted-space"> </span>→ type →<span class="Apple-converted-space"> </span><code>ArrowDown</code><span class="Apple-converted-space"> </span>→<span class="Apple-converted-space"> </span><code>Enter</code><span class="Apple-converted-space"> </span>reaches them, and<span class="Apple-converted-space"> </span><code>Esc</code><span class="Apple-converted-space"> </span>closes. On a 375 px viewport the modal goes full-screen with 47 px rows and no clipping.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">The<span class="Apple-converted-space"> </span><code>MAX_QUERY_SIZE</code><span class="Apple-converted-space"> </span>patch is confirmed applied — the input reports<span class="Apple-converted-space"> </span><code>maxLength: 512</code><span class="Apple-converted-space"> </span>and a 66-character hash survives intact. Good call adding a test that fails loudly if the patch stops applying; that is exactly the failure mode that would otherwise reappear silently on a dependency bump.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Not probing explorers to identify the chain is the right call, and worth keeping in the PR description as a decision rather than an omission.</p><hr style="font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Two small input-detection gaps</h3><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><strong>Uppercase<span class="Apple-converted-space"> </span><code>0X</code><span class="Apple-converted-space"> </span>is not recognised.</strong><span class="Apple-converted-space"> </span><code>0XD8DA6BF26964AF9D7EED9E03E53415D37AA96045</code><span class="Apple-converted-space"> </span>falls through to "No results". The hex body is matched case-insensitively but the prefix apparently is not. Some tooling and a lot of copy-paste from spreadsheets produce<span class="Apple-converted-space"> </span><code>0X</code>.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><strong>A bare 40-hex value with no prefix is not recognised.</strong><span class="Apple-converted-space"> </span><code>d8dA6BF2…96045</code><span class="Apple-converted-space"> </span>gives no results. Arguably intentional — ambiguity — but both Etherscan and Blockscout accept it, and someone pasting from a log will hit it. Your call; I would take the<span class="Apple-converted-space"> </span><code>0X</code><span class="Apple-converted-space"> </span>fix and leave this one.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Neither is a blocker.</p><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">The known gap is visible in practice</h3><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">The PR flags that a page's sections can render without their parent when the page's own<span class="Apple-converted-space"> </span><code>lvl1</code><span class="Apple-converted-space"> </span>record is not in the returned rows. That is what users see today on a majority of queries — the row that leads a group is frequently a<span class="Apple-converted-space"> </span><code>content</code><span class="Apple-converted-space"> </span>record, so the clickable label is a snippet or the page description rather than the page name.<span class="Apple-converted-space"> </span><code>staking</code><span class="Apple-converted-space"> </span>leads with "Earn rewards while securing Ethereum" under a header of "Ethereum staking: How does it work?", where the word "Staking" appears nowhere.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">It is not wrong, and I understand it waits on<span class="Apple-converted-space"> </span><code>min_indexed_level: 0</code><span class="Apple-converted-space"> </span>and a re-crawl. Just noting that it is the most noticeable rough edge in the shipped experience, so it is worth being first in the queue after #19194 lands rather than drifting.</p><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Result tail is still weak on some queries</h3><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Grouping fixed duplicates, but it exposes that positions 2–4 are often unrelated.<span class="Apple-converted-space"> </span><code>staking</code><span class="Apple-converted-space"> </span>returns:</p><pre style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code>1. Ethereum staking: How does it work?   → /staking/
2. Coin Wallet                            → /wallets/find-wallet/coin-wallet/
3. Timeline of all Ethereum forks         → /ethereum-forks/
4. Trezor                                 → /wallets/find-wallet/trezor/
</code></pre><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><code>/staking/solo/</code>,<span class="Apple-converted-space"> </span><code>/staking/pools/</code><span class="Apple-converted-space"> </span>and<span class="Apple-converted-space"> </span><code>/staking/saas/</code><span class="Apple-converted-space"> </span>are all absent. Similar on<span class="Apple-converted-space"> </span><code>erc-20</code><span class="Apple-converted-space"> </span>(glossary + erc-223 + erc-1363, the ERC-20 page missing) and<span class="Apple-converted-space"> </span><code>run a node</code><span class="Apple-converted-space"> </span>(<code>/run-a-node/</code><span class="Apple-converted-space"> </span>outside the top 3). These are ranking issues that belong to the index rather than this PR — I have detailed them on #19194 — but they are much easier to see now that the list is one page per row, so expect them to come back as feedback once this is live.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Re-running the scorer against the shipped parameters, as you suggest in the description, seems worth doing before the pagerank PR rather than after; otherwise there is no baseline to attribute the change to.</p><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Privacy notes</h3><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Both are called out in the PR and I agree with the framing, recording them so they are on the record:</p><ul style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><li>Clicking an explorer row puts the full address in a third-party URL query string.<span class="Apple-converted-space"> </span><code>rel="noreferrer"</code><span class="Apple-converted-space"> </span>keeps ethereum.org out of the referrer, which is the part we control.</li><li>The address lands in<span class="Apple-converted-space"> </span><code>localStorage</code><span class="Apple-converted-space"> </span>as a Recent entry. Separately, I measured a single recent entry at<span class="Apple-converted-space"> </span><strong>9.4 KB</strong><span class="Apple-converted-space"> </span>because it carries the whole raw Typesense hit — so a recent address search persists a lot more than the address. Trimming what<span class="Apple-converted-space"> </span><code>transformItems</code><span class="Apple-converted-space"> </span>keeps would shrink both the footprint and the amount of user-typed data at rest.</li></ul><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Given the address is now routinely stored, it may be worth not saving explorer rows to Recents at all — the value of "search this again on Base" is low compared to keeping pasted addresses out of persistent storage.</p><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Smaller things</h3><ul style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><li>Group headers use the page's meta title, so they read as full sentences ("Ethereum privacy apps: tornado cash and others") above a one-word row. Consistent with the<span class="Apple-converted-space"> </span><code>docsearch:*</code><span class="Apple-converted-space"> </span>tags, just visually heavy in a narrow modal.</li><li>A 64-hex value produces 12 explorer rows and no content results — correct, but it is a long scroll on mobile. Worth considering whether Blockscout's nine networks need to be listed in full before the user has expressed any interest in a chain.</li><li>The four<span class="Apple-converted-space"> </span><code>docsearch-explorer-*</code><span class="Apple-converted-space"> </span>strings render in English in every locale until the intl pipeline runs. Fine as a temporary state given the fallback merge in<span class="Apple-converted-space"> </span><code>request.ts</code>.</li></ul><h3 style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Not covered</h3><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Dark mode rendering of the explorer rows, RTL locales, and clicking through to each explorer to confirm the destination pages resolve.</p>

myelinated-wackerow:
Brings in #19194's review round and the four items queued here. Both branches are now merged up from `dev`.

| Queued item | Outcome |
| --- | --- |
| Recent searches at ~9.4 KB/entry | Fixed. `_rawTypesenseHit` and `_grouped_hits` are stripped in `transformItems`; nothing reads them. Measured 9,899 -> 515 bytes against live data. |
| Row labels using the page description | Fixed. A page row now takes its title from the page rather than its `h1`, which on a landing page is a slogan -- `/staking/` read "Earn rewards while securing Ethereum". Only the rendered copies change; the raw `hierarchy.lvl1` is what a child row matches to find its parent. |
| Typo tolerance | Partly. The knobs named don't move anything: typo correction is already on, `walet` returns 390 results and all of them are wallet pages, so `num_typos` and `typo_tokens_threshold` measure identical at every value. A misspelling costs rank, not recall. `drop_tokens_threshold: 5` is the one that helped -- hit@5 65% -> 71% on the labeled queries misspelled one character each, hit@1 on the correctly spelled set unchanged. |
| Use the snippet instead of the full value | Dropped -- the snippet already is what renders. [Correction posted on #19194](https://github.com/ethereum/ethereum-org-website/pull/19194#issuecomment-5671247580). |

**`group_limit: 3` -> `1`**, found while measuring the first item. The adapter collapses each group to its best-scoring member and staples the rest on under a key nothing reads, so the library's 3 bought three times the response for the same rows -- 55 KB against 18 KB for `wallet`, on every keystroke. Verified identical leaders and `found` across six queries.

**Result tracking**, which didn't exist: the only search event was the modal opening, so nothing recorded follow-through, and the pageview a click produces can't answer it either -- `Matomo.tsx` tracks `usePathname()`, which has no fragment, so a page and a section inside it arrive as the same URL. Every selection now emits `search / <kind> / <path>`, where kind separates a page, a section, an explorer row, and a row taken from history. Both selection paths are covered and neither double-counts.

**Also:** result spacing tightened, and the rules inside a page block lightened to the colour of the elbow connector so only the rule closing a block reads at full strength.

Two parts of the description above are now stale -- the `min_indexed_level` known gap (`withPageRow` synthesizes the parent row) and the deferred `pagerank` note (#19232 merged). Happy to update the body.

-- Opus 5

myelinated-wackerow:
Since review:
- Explorer click events name the network only; the explorer URL carrying the pasted address or hash is no longer sent to Matomo (621b6ff4c1).
- The group header is hidden when a group leads with its page row, since the two read the same title twice; Recent and explorer groups keep theirs (e25bcba871).
- Uppercase `0X` is recognized (c9ba2dc59e), `dev` is merged in, and the Known gaps section reflects the synthesized page row.

-- Fable 5.1

wackerow (approved):
@pettinarip Bringing this one in as well to complete the initial Typesense v1/v2 pair of migration PRs. Will chat with you, but as with #19194, please feel free to add further review here and I can continue to iterate if you spot issues or have any change requests.

PR #19251, opened 2026-09-09
feat(find-wallet): target ranking phrases in page titles

pettinarip:
## Description

First of two follow-ups from the FindWalletCatalog2026 A/B readout ([analytics hub, "Do this" → 4](https://ethorg-analytics-hub.vercel.app/matomo/ab-test-find-wallet-catalog.html)): rename the find-wallet pages after the searches they already rank for.

- Split each persona's meta title/description from its hero copy (`metaTitleKey` / `metaDescKey` on `WALLET_PERSONAS`), so one can move without the other. `generateMetadata` and the persona JSON-LD read the meta keys; the hero H1 keeps `heroTitleKey`.
- New phrases, applied to both the `<title>` and the H1:
  - Index: **List of Ethereum wallets: compare by feature**
  - NFTs: **NFT wallets for Ethereum**
  - Finance: **DeFi wallets for Ethereum**
  - New to crypto: **Ethereum wallets for beginners**
  - Hardware and Developer unchanged
- Card titles and breadcrumb leaves keep the short `titleKey`.
- English only; other locales fall back until the translation pipeline picks up the new keys.
- e2e: the page-heading fixture read the nav label ("Choose your wallet"), which no longer matches the H1, so it now reads `page-find-wallet-title`.

Companion PR (interaction follow-ups, independent files): #19252

## Related Issue

A/B test wrap-up, no issue.

PR #19252, opened 2026-09-09
feat(find-wallet): multi-select persona presets, URL-addressable filters, client-side modal

pettinarip:
## Description

Follow-ups to the find-wallet catalog after the A/B test, plus a simpler modal.

**Personas are multi-select client-side presets**

- Checkbox cards (styling ported from the old preset cards). Several can be selected and AND-combine on `wallet.personas`; counts are live over the currently visible wallets.
- Switching personas no longer resets the page: persona routes render their own wallets and mount the rest on the visitor's first interaction, so switching stays client-side. JSON-LD stays subset-only.
- Reset buttons clear sidebar filters only. The "See all wallets" link is gone.

**URL-addressable state** via `history.replaceState`

- One persona → `/wallets/find-wallet/personas/{id}/`; several → `?personas=a,b` off the index.
- Sidebar groups sync to comma-joined query params (`?devices=ios,android`), read once from `window.location` on mount so the page stays static.

**Client-side modal**

- The intercepting `@modal` route is removed; `WalletDetailModal` renders from the card payload (`url`, `discord`, `twitter` added). No navigation, no RSC fetch, no per-card detail prefetch.
- Cards keep a real anchor to the standalone `[wallet]` page; a plain click opens the modal.
- Back closes the modal (one same-URL history entry, added a frame after the dialog paints).
- Matomo event triples unchanged; the per-wallet modal pageview is gone by design.

`FilterableCatalog` gains optional controlled `selection` / `onSelectionChange` and `onFilteredChange`; `ToolsCatalog` is untouched.

**Perf** (local prod builds vs `dev`, 6× CPU throttle, mobile, INP p75 steady-state): modal 80 → 56 ms, filter sheet 112 → 104 ms, device checkbox 160 → 136 ms, persona toggle 280 ms (was a route navigation). Second commit fixes a 904 ms modal regression the bench caught. Payload: index HTML gzip +1 KB, JS +10.6 KB gzip; persona pages render their subset only, −27.5% raw / −19.9% gzip against the first pass in this PR.

**Follow-ups from a quality audit** (last five commits, one per finding)

- Persona pages were shipping the full 48-card markup with 43 `hidden`, so each one linked to every wallet and the persona-to-subset signal survived only in the JSON-LD. They now render their own wallets and mount the rest on the first pointer/key event, which crawlers never fire. Index HTML byte-identical, persona toggle INP unchanged.
- Selecting two personas dropped both from the URL, so a reload or a shared link came back unfiltered while the page still said it was filtered — hence `?personas=a,b`. The e2e assertion for the old bare URL is updated and now covers the reload.
- Back used to leave the catalog (and the user's filters) instead of dismissing the modal.
- Closing the modal dropped focus on `<body>`, so keyboard users restarted from the top of the page (WCAG 2.4.3); focus now returns to the card. Pre-existing — `dev` does the same.
- `onFilteredChange` hands the filtered list over in a transition. Behaviour-neutral and unmeasurable in INP here, since `useDeferredValue` already keeps that pass off the critical path — happy to drop it.

Worth knowing if you touch the modal: adding a history entry inside the click handler makes the browser wait for the dialog's ~430 ms render before the interaction's paint (modal open 32 → ~900 ms). Not Next's patched `pushState` — the native method behaves the same — so the entry goes in on the next frame.

**Tests**: e2e rewritten for the new behaviour. 19/20 pass locally; the one failure is a local image-optimizer stall unrelated to this change, expect green on the preview.

Companion PR (titles): #19251

## Related Issue

A/B test wrap-up, no issue.


myelinated-wackerow (changes requested):
@pettinarip overall looks good, just a few things to double-check and tidy up before we pull this in:

**1. The persona pages are now orphaned.** Nothing on the site links to `/wallets/find-wallet/personas/{id}/` anymore — the cards became checkboxes and the "See all wallets" link is gone, so grepping `app/` + `src/` for `/personas/` only turns up the JSON-LD `@id` and the sitemap generator (`src/lib/i18n/translationRegistry.ts:145`). Sitemap-only discovery with zero internal inbound links is a weaker signal than the paths themselves buy us.

Could the cards go back to being real anchors to their persona page, using the same pattern you already use on the wallet card (`WalletCard.tsx:49-68`) — real `href` so crawlers and modifier-clicks get a genuine link, `preventDefault()` on a plain left click so it still filters in place? Would that satisfy the design here, or does the multi-select checkbox affordance rule out wrapping each card in a link?

**2. Every persona toggle fires a Matomo pageview.** `buildUrl` (`WalletsCatalog.tsx:179-198`) rewrites the pathname, and Next's patched `replaceState` dispatches `ACTION_RESTORE` (`next/dist/client/components/app-router.js:267-278`), which moves `canonicalUrl` → the `usePathname` memo (`app-router.js:113-125`) → `src/components/Matomo.tsx:50-77` pushes `trackPageView`. So every check/uncheck inflates pageviews on both the index and the persona pages — on the PR we're relying on to read persona engagement. The same dispatch also re-renders every `usePathname`/`useSearchParams` consumer (header, nav, footer) on each toggle, which the INP numbers won't attribute because it lands inside `startTransition`.

Smallest fix is the native method, mirroring what you already do for the modal's `pushState`:

```js
History.prototype.replaceState.call(window.history, window.history.state, "", url)
```

One trap: pass `window.history.state`, not `null`. The current `null` is only safe because Next's patch re-copies `__NA` and the router tree; the native call doesn't, and a popstate onto an entry without `__NA` triggers `window.location.reload()` (`app-router.js:288-291`). Trade-off is that `usePathname()` then stays on the pre-toggle route, so the language picker (`src/components/LanguagePicker/index.tsx:28`) builds its hrefs from the original path. If you'd rather keep that truthful, the alternative is suppressing on the Matomo side — a `markSilentPathChange(pathname)` helper in `src/lib/utils/matomo.ts`, called before the `replaceState`, with `Matomo.tsx` skipping the `trackPageView` (but still advancing `previousPath`) on a match. ~10 lines, and it'd cover any future replaceState URL sync.

**3. Campaign params get stripped before Matomo's tracker reads them.** This one only bites external inbound links that land directly on the catalog carrying tracking params — a newsletter, social post, or ad pointing at `/wallets/find-wallet/?utm_source=newsletter`. `WalletsCatalog.tsx:288-292` rebuilds the query from scratch and keeps only the six recognized filter keys, so the param is erased a few ms after load — before the `async defer` `matomo.js` has finished loading and read `location.href`. `init()` only queues `['trackPageView']` into `_paq`, so the URL is read late and the visit logs as if it came from nowhere. Same for `mtm_campaign`, `pk_campaign`, `gclid`, and referral params. Mutating the live `URLSearchParams` (set/delete only the six keys we own) instead of rebuilding would keep them.

**4. "Reset filters" can't clear a persona-driven result set.** Check all five personas and you're down to one wallet (OneKey); "Reset filters" clears the sidebar but leaves the personas checked, so OneKey stays the only result and the button reads as broken. `onReset` (`WalletsCatalog.tsx:326-333`) only deletes `ALL_FILTER_KEYS`, and the sidebar-header reset is `disabled` at `activeCount === 0` (`WalletFilters.tsx:241`), so there's no other affordance in view — the checkboxes are scrolled off above. Either clear personas too, or relabel so it's clear what it covers.

**5. Worth confirming the build cost.** `generateStaticParams` for `[wallet]` now returns every slug (`[wallet]/page.tsx:256-259`), so ~1,200 prerendered pages (48 × 25 locales), each running `getRelatedWallets` — a 48-wallet scan plus a shuffle and six `enrichWallet` calls. Reasonable trade now that the interception hack is gone, just worth eyeballing the Netlify build time and memory delta. Also `getAllWalletSlugs()` already does exactly this without running `stripMarkdown` over all 48 records.

**Minor**

- Persona card counts are live (relative to the currently filtered set) while the sidebar counts are static server-side totals. Two conventions on one page reads as a bug even though both are deliberate.
- `WalletPersonaCards.tsx:81,95` use `text-left` where they want `text-start` — on `/ar/` at desktop width the short last line of each card description sits on the left inside an otherwise right-to-left card. Carried over from the old preset cards, but the component's been rewritten here anyway.

-- Opus 5/@wackerow

pettinarip:
Thanks, good catches. Ended up rethinking this instead of patching it: personas were basically a second filter system, which is where 1, 2 and 4 all came from. They're now just shortcuts that check the underlying features in the sidebar. Same results, one system.

- Card titles are real links to the persona pages again, same trick as the wallet cards.
- The pathname doesn't move anymore, only the query, so the Matomo pageview thing goes away on its own.
- Query is mutated in place, campaign params survive. Added an e2e for it.
- Reset clears everything now.
- Wallet detail pages are on demand + cached until next deploy. Only index and persona pages get prerendered.

Also fixed `text-start`. Counts (live vs static) are still different, no strong opinion there — happy to align them either way.

app/[locale]/wallets/find-wallet/_components/WalletsCatalog.tsx
```diff
@@ -57,47 +66,199 @@ export type WalletCatalogLabels = {
...
+    ).flatMap((persona) => persona.features)
+  )
+  const next = { ...state }
+  for (const feature of PERSONA_FEATURES[id]) {
+    if (keep.has(feature)) continue
+    const slot = slotOf(feature)
+    next[slot] = asArray(next[slot]).filter((existing) => existing !== feature)
+  }
+  return next
+}
```
Comment (wackerow) [inline app/[locale]/wallets/find-wallet/_components/WalletsCatalog.tsx:186]: This seems to work for filters that are not part of a persona, but for those that are it can drop a users selection.

Example: Filter "Connect to dapps" -> Check "NFT" persona -> Uncheck "NFT" persona
Since "Connect to dapps" is part of that persona, it gets unchecked.

If the filter is not part of the persona, it stays checked (Try "Desktop" instead in the example above... stays checked after toggling "NFT" persona)

---

If we want it to keep filters that the user had intentionally checked, consider having `addPersona` return the features it actually added and stash them in a ref keyed by persona id, then `removePersona` drops only those.

wackerow (approved):
@pettinarip Nice! Working really well on my end. Noted one semi-bug with a suggestion below, otherwise this looks good. Approving as-is though.

wackerow:
@pettinarip I'm going to pull this in since the above is non-blocking, but @konopkja confirmed we should address the above bug in a separate issue.

PR #19274, opened 2026-09-14
refactor(tools): share the find-wallet filter drawer and make category filters expandable

pettinarip:
Consolidates the mobile filter UI on `/developers/tools/` with `/wallets/find-wallet/`.

- Tools now uses the find-wallet drawer on mobile (search inside it); the unused `collapsible` variant is deleted.
- Category rows expand in place instead of navigating; subcategories filter client-side, each group keeps a link to its listing page.
- Subcategory selection is now a URL (`?sub=<id>`): shareable, survives refresh, undone by the back button.
- Sidebar links from other categories land pre-filtered.
- Following an "all resources" link clears the filter.
- The drawer closes once a subcategory or link is picked; expanding a group or searching leaves it open.
- Dropped the redundant category line above the result count.


myelinated-wackerow (approved):
Nice consolidation. Checked the behavior on the preview: cross-category `?sub=` deep links apply on both client nav and hard reload, "All resources" clears the filter, one history entry per pick with Back restoring the filtered state, the mobile drawer closes on a pick or link but stays open on expand/search, and the collapsed `keepMounted` rows stay out of the tab order.

@pettinarip Looks good overall, no blockers, approving with a couple suggestions:

**1. Give the mobile bar a reset affordance.** A `?sub=` value that matches nothing (a shared link after a taxonomy rename -- subcategory ids have no redirect layer the way category slugs do) lands on "Resources found: 0 / 73" with nothing highlighted in the sidebar and no way out except opening the drawer. A `RotateCcw` / "Reset" control next to **Filters** would clear search + filter without opening the menu, and it earns its keep in the normal case too. Ignoring an unknown `sub` on read would cover the dead-link case as well.

**2. Distinguish the chevron's accessible name.** The `sr-only` text is the group label verbatim, so the toggle and the label link announce identically back to back ("Contract tooling, button" / "Contract tooling 73, link"). `Expand {label}` reads better.

Nits, take or leave:

- On `/developers/tools/`, "Resource categories" keeps its active highlight while a subcategory is selected, so two rows look active at once -- `!hasCurrentItem` could also check `selectedChildId`.
- On a category page, the other categories' subcategory rows render without counts while the current category's have them; correct (counts are page-scoped) but reads as missing data.
- `resetAll`'s `pushState` skips the dead-entry guard `setFilter` has. Unreachable until tools passes a `resetLabel` -- relevant if you add one for #1.
- `filterTool` matches `sub` globally, so a subcategory id reused in two categories would cross-filter on the root page. All 24 are unique today, just not enforced.

-- Opus 5/@wackerow

PR #19275, opened 2026-09-14
refactor(modals): share one detail modal across tools and find-wallet

pettinarip:
`/developers/tools/` and `/wallets/find-wallet/` are both product listings with an intercepted detail modal, but each had its own implementation. They now share one.

### Before / after

Same tool (ethPM), same viewport, modal content only.

| Before | After |
|---|---|
|  |  |

---

**New `src/components/CatalogDetailModal/`** — the shell (logo + title + close in a row, pinned header, scrolling body), `DetailRow`, and `FullDetailsLink`. `WalletDetailModal` and `ToolDetailModal` are gone.

**Fixes a close-button bug.** The tools modal used `dialog-modal`'s `media` variant, whose header is `absolute end-0 top-0`. `DialogContent` isn't a positioning context, so the close ✕ rendered in the viewport corner instead of the modal — overlapping the site nav on mobile. That variant had no other consumer and is removed. (It's absent from the "before" shot above for exactly this reason: it wasn't inside the dialog.)

**Tool modal rebuilt on the shared pieces.** Category / type / tags become spec rows. Repos and packages were an unbounded grid of outline buttons; most tools have one, but Hats Protocol lists 31, which stretched the modal far past the rest of the catalog. `ToolLinkRows` shows the top 3 by stars/downloads with `+N more →`, one line each. The standalone tool page is unchanged and keeps the full list.

The tool modal no longer shows its banner image; the standalone page still does.

### Testing

`pnpm lint` and `pnpm type-check` clean. Checked by hand on a local dev build at 1280px — tools now matches find-wallet, find-wallet itself is unchanged.

Not verified: mobile widths. No Storybook story yet; whoever adds one will need `parameters.nextjs.appDirectory: true`, since the shell calls `useRouter`.


src/components/ui/dialog-modal.tsx
```diff
@@ -41,16 +41,6 @@ const dialogVariant = tv({
       unstyled: {
         content: "block p-0 rounded-none bg-none gap",
       },
-      // Edge-to-edge media modal: no padding, a single scrolling body, and a
-      // close button that floats over the content (e.g. over a banner image).
-      media: {
```
pettinarip, line 46: This was dead code after I removed the old tool modal.

myelinated-wackerow (approved):
@pettinarip Lgtm.. Couple non-blocking nits/suggestions below, with a question... approving

**1. Mobile link labels truncate to nothing.** The most substantive one. At 390px the card is 358px, the shell's inherited `p-8` leaves 294px inside, `roomyLabel`'s 35% takes 103px, icon+gap another 22 — so the label gets ~153px and ethPM renders as `ethpm/et…`, `ethpm/eth…`, `ethpm/py-…`: three rows you can't tell apart. Hats is the same, and `title` doesn't exist on touch.

Suggestion: below `sm`, stack `LinkList` under the row label and align it left, which gives the values the full 272px and clears the longest real slug (`Hats-Protocol/farcaster-delegator`, ~224px). Worth making it an opt-in `DetailRow` prop — find-wallet's values are short and read better two-column, and stacking all eight rows would stretch that modal. While in there, `p-4 sm:p-8` on the shell's `DialogContent` buys another 32px on mobile and helps find-wallet too.

**2. The description undersells the find-wallet delta.** The shell now caps at `max-h-[calc(100dvh-2rem)]` with an inner scroller (title pinned instead of the whole card scrolling) and the wallet logo gains `rounded-lg` + explicit 56px. Both are improvements — `rounded-lg` matches `WalletCard` — but "find-wallet itself is unchanged" would lead a reviewer to skip re-checking a modal whose scroll behavior changed.

**3. The block most likely to drift is still duplicated.** The footer actions row (`flex flex-wrap items-center justify-between gap-4`, solid CTA, `size-10 rounded-lg border` icon-social link) is copy-pasted in both modals with identical classes. Not a blocker, but it's the natural third extraction. Related: the tools side diverges — with no website, "Social" becomes a full solid button, where find-wallet always uses the icon. Consistent icon treatment would be nicer.

**4. Pointless plural.** `{count, plural, one {+# more} other {+# more}}` has identical branches, so `"+{count} more"` says the same thing with one fewer ICU structure for 24 locales and MT to mangle. Also `labels.more(repos.length - SHOWN)` is evaluated unconditionally, so a one-repo tool computes `+-2 more` (unrendered) — worth gating on `hidden > 0`.

**5. a11y.** Two identically-worded `+N more` links in one dialog are ambiguous out of context; an `aria-label` naming the row fixes it. Separately, since this PR correctly gives the new tools social link an `aria-label`, it'd be a good moment to close the gap on find-wallet's Discord/Twitter icon links, which still have no accessible name.

**6. No story for the new `src/components/` component.** The `parameters.nextjs.appDirectory: true` caveat only applies to the shell — `DetailRow` and `FullDetailsLink` are pure markup, and `roomyLabel`/`alignTop` (plus whatever #1 adds) is exactly the contract that needs showing.

**7. `getRankedRepos` and `getToolRepoHrefs` now carry the same comparator.** Collapses to one line:

```ts
export const getToolRepoHrefs = (tool: DeveloperTool) =>
  getRankedRepos(tool).map((repo) => repo.href)
```

The only consumers are the two JSON-LD paths (`sameAs`, and `getToolPrimaryUrl`'s canonical URL), which shouldn't emit a non-external href anyway, and `isExternal` is loose enough (`includes("http")` || mailto || ipfs || discord) that the filter only drops malformed entries.

**Optional:** `DetailRow`'s `tooltip?: string` puts Radix Popover+Tooltip in the client graph for every consumer, since the import edge alone is enough and a conditionally-rendered default export can't be shaken. There's no Tooltip under the tools route today, so it's bytes fetched when a tool modal opens (not initial-load — the `@modal` slot only streams on interception). `tooltip?: ReactNode` with the wallet page passing its own `<Tooltip nested>` would avoid it. Modest payoff; fine to skip.

**Question:** dropping the banner image from the tool modal — intentional product call, or a side effect of moving off the `media` variant? The result is consistent with find-wallet, so I suspect the former, just want it on the record.

-- Opus 5

myelinated-wackerow:
@pettinarip Merged `dev` in and resolved the conflicts — #19252 landed in between and moved find-wallet off the intercepting route, which collided head-on with this PR's extraction.

- **`@modal/(.)[wallet]/` deleted.** dev wins: the wallet modal is a client component now, not a route.
- **`WalletDetailModal` rebuilt on the shared pieces.** It stays dev's client modal fed from the card payload, but composes `CatalogDetailModal` + `DetailRow` instead of its own copies, so the consolidation this PR is about still holds on both sides.
- **Shell gained two optional props** to make that work: `onClose` (defaulting to the `router.back()` the tools interception still needs) and a pass-through `onCloseAutoFocus`, so find-wallet keeps returning focus to the card.

Net diff vs `dev` is the same 11 files as before; `lint` and `type-check` clean. The two find-wallet deltas I flagged in the review (pinned header + inner scroller, `rounded-lg` logo) now apply to the client modal.

-- Opus 5

wackerow (approved):
@pettinarip Pulling this in, since above suggestions are non-blocking. Can place these in a separate PR if we want to move forward with any of them.

PR #19279, opened 2026-09-15
feat(a11y): keyboard shortcut registry and ? help dialog

myelinated-wackerow:
## Summary

Every keyboard shortcut on the site now lives in one place (`src/lib/keyboard-shortcuts.ts`), and pressing <kbd>?</kbd> opens a dialog listing all of them. The dialog renders from the registry, so the two cannot drift.

The theme toggle, desktop language picker and nav section digits were each hand-rolling their own `keydown` listener; they now bind against a registry entry via `useKeyboardShortcut`. That also fixes the theme handler swallowing a bare <kbd>\\</kbd> typed into an input.

There is no visible entry point for the dialog yet, pending a design pass.

-- Opus 5


wackerow:
- [ ] Add matomo tracking to keyboard shortcut events

Issue #19284, opened 2026-09-16
Developer tools page serves 299 tools, the catalog has 440


phipsae:
## Description

The developer tools page shows 299 tools. The builder-resources catalog on `main` has 440.

The page render itself is current, so what is stale is the `fetch-developer-tools` blob. It was last written between 11 and 13 Sep. It has the three subcategory ids added to the catalog on 11 Sep, none of the entries added on 13 Sep, and its stargazer counts trail GitHub by about three days.

A possible cause, in case it helps. The task seems to reach its last stage and then stop. Thumbnails for entries added on 13 Sep are already in the S3 bucket, so images get uploaded and `set()` is never reached. Running `fetchDeveloperTools()` against the current catalog peaks at about 394 MB of RSS, mostly in `uploadToolImages`, which fans out over all 440 resources at once. The default `small-1x` machine has 512 MB. Your Trigger.dev run logs would confirm or rule that out.

## Steps to reproduce

1. Open https://ethereum.org/developers/tools/ and read "Resources found", 299.
2. Compare with `catalog/resources.json` on `ethereum/builder-resources` at `main`, 440 entries.

phipsae:
Folding this into the pull request, the full diagnosis now lives in #19285.

PR #19285, opened 2026-09-16
fix(data-layer): own maxDuration for the developer tools fetch

phipsae:
## Description

The developer tools page has been serving a 299-entry snapshot since 13 Sep, while the builder-resources catalog is at 440.

The page render is current, so this is not an ISR problem. It regenerates daily and still reads 299 out of the data layer. What is stale is the `fetch-developer-tools` blob that the daily Trigger.dev task writes.

**The task has not written since 13 Sep 2026.**

- The embedded payload reports `totalCount: 299`. The catalog reached 439 entries on 13 Sep and 440 on 16 Sep.
- That payload does contain the three subcategory ids added to the catalog on 11 Sep, so the last good write sits between 11 and 13 Sep.
- Stargazer counts in the payload trail GitHub by about three days (`foundry-rs/foundry` is 10593 on the page and 10606 now).

**It reaches its last stage and then dies.**

`uploadToolImages` is the final step before the blob write. 55 of 60 sampled thumbnails belonging to entries added on 13 Sep are already in the S3 bucket, under exactly the key `uploadToS3` derives. So the run enriches, ranks, uploads images, and then never gets to `set()`.

**Cause.**

`trigger.config.ts` sets `maxDuration: 300` for every task. Per the Trigger.dev docs, `maxDuration` excludes only `wait.for`, `triggerAndWait` and `batchTriggerAndWait`, so the fetcher's own `sleep()` pacing counts against the limit.

Running `fetchDeveloperTools()` unchanged against the current catalog takes 162 seconds locally, 94 of which are hard-coded sleeps (15s between scoped npm batches in `fetchNpmJs`, 1s between GitHub batches in `fetchGitHub`). The same run against the pre-13-Sep catalog takes 135 seconds. Production runs on the default `small-1x` machine (0.5 vCPU) and does the real S3 uploads that my local run skipped, so the 300 second budget is thin, and it became thin exactly when the catalog grew by 47 percent. Your Trigger.dev run history should confirm the timeout, I cannot see it from outside.

**The change.**

Task definitions are built by one factory from a `[key, fetchFn]` tuple, so every task inherits the config default. This adds an optional third slot for per-task overrides and gives the developer tools fetch 900 seconds. Every other task is untouched.

Two things I left alone, happy to follow up if you want them.

- `fetchNpmJs` waits 15 seconds between scoped-package batches, which is 75 of those 94 seconds. npm documents no request rate limit for the downloads API that would tell me a safe lower value, so I did not guess at one.
- `uploadToolImages` fans out over every resource at once with `Promise.all`, while `uploadManyToS3` right next to it chunks for exactly that reason. On `small-1x` (0.5 GB) that is around 93 MB of image bodies in flight on a cold run.

## Related Issue

None.


phipsae:
Closing this. The 300s maxDuration was my first theory, but the run only uses about 6 seconds of CPU in 139 seconds of wall clock, so the cap is not what it is hitting. Leaving the report in #19284 instead.

PR #19292, opened 2026-09-17
fix(data-layer): unblock developer tools fetch

### LINKED ISSUE #19284: Developer tools page serves 299 tools, the catalog has 440

myelinated-wackerow:
## Summary

`fetch-developer-tools` has been timing out since 13 Sep, so the page serves a 299-entry snapshot while the catalog is at 440. Nothing in the code changed -- the catalog grew 47% and the run stopped fitting the shared 300s budget.

- **Bounded the image fan-out.** `uploadToolImages` fired ~440 source fetches through one `Promise.all`, all sharing a single `AbortSignal.timeout(10s)` deadline created before most were dispatched. Replayed against the live catalog: 292 timeouts, 195 `EAI_AGAIN`, zero successes. Bounded at 10: 491/491 clean. `uploadManyToS3` right below it already chunked for this exact reason, so the helper is now shared as `mapWithConcurrency`.
- **Check S3 before downloading.** 488 of 491 catalog URLs carry a usable extension, so the key is derived and HEAD-checked first. Where the extension lies about the served type -- 235 of those URLs are `github.com/<org>.png` avatars, which GitHub returns in whatever format the org uploaded -- a HEAD on the source settles the real key instead of a full download. Measured over the real catalog, a steady-state pass resolves 435 of 491 images on the S3 pre-check alone with no source contact; the rest cost a source HEAD, and only genuinely new images pull a body.
- **Gave the S3 client a socket timeout.** The SDK ships none, so a stalled endpoint hung the run until `maxDuration` killed it. (`requestTimeout` is the wrong knob here: in `@smithy/node-http-handler` 4.12.0 it only warns unless `throwOnRequestTimeout` is set, and it caps total wall clock, which would abort a legitimately slow multi-MB `fetchApps` screenshot PUT. `socketTimeout` fires on inactivity instead.)
- **Parallelized GitHub and npm enrichment.** They touch disjoint fields (`repos` / `packages`) but ran in series, spending both pacing budgets back to back: ~159s to ~83s. npm's own 15s spacing is unchanged -- probing the downloads API confirms it rate limits hard.
- **Handled GitHub 403s.** Secondary rate limits come back as 403, which `fetchRetry` doesn't cover, and the error kept only the status. Batches now honor `Retry-After`, and failures carry the body and rate-limit headers.
- **Stopped publishing degraded rankings.** `ranking.ts` reads a missing stargazer count as zero, so a dropped batch quietly demoted every repo it missed. Below 75% repo coverage the run now fails instead, keeping yesterday's blob.
- **Fixed positional GraphQL alias misalignment.** `buildGraphQLQuery` filtered unqueryable owner/name pairs and then aliased over the filtered list, while the result loop walked the unfiltered one -- so one rejected name shifted every later alias in that batch onto the wrong repo. Pre-existing and latent today (all 479 catalog repos pass the pattern), but it writes plausible-but-wrong stats that the coverage gate cannot see, because the map comes back full. `tests/unit/data-layer/fetch-github.spec.ts` covers it and fails on the pre-fix code.
- **Bounded the aggregate rate-limit wait.** Per-batch retries were unbounded across the run: 20 batches honoring a `Retry-After: 60` twice each is ~40 minutes against a 900s budget, so a bad GitHub day got the task killed rather than failing into the readable coverage error. There is now a run-level ceiling shared across batches.

Measured end to end against the real catalog with a stubbed bucket: **115s cold, 86s warm, 188MB peak** (was timing out at 300s). Adds `tests/unit/data-layer/s3.spec.ts` covering the pre-check and the concurrency bound.

Per-task `maxDuration`/`machine` overrides are now supported; this task gets 900s. Every other task is untouched.

Closes #19284.

## Open calls

- `MIN_REPO_COVERAGE = 0.75` converts a bad GitHub day into a failed task rather than a quietly wrong page. Currently 477/479 repos resolve, so there is wide headroom.
- `maxDuration: 900` against a measured 115s is deliberate headroom.
- No machine bump. An earlier revision raised this task to `small-2x`; that was dropped on review, since the measured run fits `small-1x` and the image stage is now 10-wide rather than 440-wide. The override plumbing is still there if a production run says otherwise.

Three catalog images fail on every run and are untouched here: `speedrunethereum.com/favicon.ico` and `w3pk.w3hc.org/favicon.ico` (`.ico` is not in `IMAGE_TYPES`), and `solx.zksync.io/favicon.png` (serves `text/html`). All pre-existing; they degrade to the original URL.

-- Opus 5


wackerow:
Running this in Trigger's development mode, getting 440 results without errors

myelinated-wackerow (commented):
Reviewed at `7bbc46e`. Type-check, lint, and the unit suite are green locally.

The diff holds up against the other S3/trigger consumers. `mapWithConcurrency` preserves `uploadManyToS3`'s chunking and default of 5; the `TaskDef` third element leaves the other 30 tasks spreading `{}`; `fetchGitHub` is developer-tools-only, so the coverage gate's hard-fail semantics don't reach anything else; and the GitHub/npm index merge is safe since both fetchers are `appData.map(...)`. `fetchVideoThumbnails`' sd→hq fallback survives the S3 pre-check, since both attempts key to the same `slug.ext`.

Two things I dug into:

**`socketTimeout` is the right knob.** Traced through `@smithy/node-http-handler@4.12.0` — it resolves to `socket.setTimeout(19000, onTimeout)` after the 1s defer, which destroys the request and rejects. Worth noting for anyone reading this later: `requestTimeout` would *not* have worked, since it only `console.warn`s unless `throwOnRequestTimeout` is set, and `socketTimeout` has no fallback from it.

**The alias regression test is real.** I checked out the pre-fix `fetchGitHub.ts` and ran `fetch-github.spec.ts` against it: fails with `Expected 222, Received 333`, exactly the one-slot shift the fix describes. The empty-batch early return is a good catch too — an all-rejected batch would otherwise have posted `query { }` and 400'd.

One non-blocking note: the coverage denominator is `allRepos.length`, which includes repos `VALID_REPO_PATTERN` rejects and that therefore can never resolve. Moot at 479/479 today, but the gate and the filter are working off different populations, so a malformed catalog URL would eat headroom rather than being excluded from the count.

Would approve as-is.

-- Opus 5

PR #19294, opened 2026-09-17
fix(trigger): move off the Node 21 runtime

myelinated-wackerow:
## Summary

`trigger.config.ts` left `runtime` at the unpinned default, which Trigger.dev resolves to **Node 21.7.3**. The dashboard now warns that deployments on that runtime fail from **5 October 2026**, so the data-layer tasks stop being deployable on that date.

`node-22` resolves to 22.16.0, matching the 22.22.0 in `.nvmrc` that the rest of the repo builds against. The installed SDK (4.3.3) accepts `"node"`, `"node-22"` and `"bun"` only -- the docs list `node-24` and `node-26`, but those need an SDK bump.

One line, plus a comment recording why the default isn't safe to fall back to.

-- Opus 5


wackerow (approved):
@pettinarip bringing this in

PR #19308, opened 2026-09-21
feat(scope): Add banner of current Gloas scope + FCR to default BBP

0xMushow:
Adding a banner specifying what stage is currently in scope for Gloas-only and FCR as well to the default Bug Bounty.
As these are "temporary" / "new", I highlighted them using the alert component.

Capture d’écran 2026-09-21 à 12 04 23

+

Capture d’écran 2026-09-21 à 09 52 59

The link to the client code is still missing some, but will hopefully soon be updated when we get it by client teams.

> https://notes.ethereum.org/@mushow/glamsterdam_clients_scope

cc / @0xFredrik @moodmosaic


0xMushow:
Waiting decision from FCR channels on what we would consider Low,Medium,High.

0xFredrik:
> Waiting decision from FCR channels on what we would consider Low,Medium,High.

We should treat it the same as our other bounties imo, basically what the overall impact of the vulnerability in FCR is for Ethereum as a whole.

0xMushow:
> > Waiting decision from FCR channels on what we would consider Low,Medium,High.
> 
> We should treat it the same as our other bounties imo, basically what the overall impact of the vulnerability in FCR is for Ethereum as a whole.

Yes, we're just trying to decide what falls in High, they also are pushing adopters to use 2 CL clients, so that could be a rule to define a High. A bug in both clients, so a bug in the algorithm itself (would be a High), considering the adoption is mainly by bridge, an algorithm bug would be quite bad.

Capture d’écran 2026-09-21 à 15 59 55

Just trying to decide what falls in which category, but it's good to go otherwise, if you agree we can merge it and decide severity when we get bounty reports (if any).

cc @Ma-Julian

0xFredrik:
I think it's good to go without making any specific mentions of what's a
Medium, High or so and that we make the judgement call once we get a report

Ma-Julian:
Hey @0xFredrik , it could make sense to me to define it later but just want to check your reasoning on what you would think of e.g. using the classification as proposed above in the screenshot. Do you perhaps not expect many bug reports and so that it would be faster to deal with cases on an ad-hoc basis or is it that you think it is very hard to define a proper classification ex-ante?

Ma-Julian:
@0xFredrik to clarify my thinking: In speaking with potential users of FCR, it would help bring confidence in FCR if we could make a statement like: there has been a bug bounty of X USD on it for Y months and these are the exploits that were found. 

I think it is fine not to have assigned what types of bugs exactly fall under which level of severity as long as there is a credible statement of what the max bug bounty would be. If we don't assign classifications exactly, would it be reasonable for me to say to people that there is a max bug bounty of 50k USD on FCR?

0xFredrik:
Yeah I'm at least hoping we won't see a lot of valid reports, and if we do we should be able to make a fair assessment at the time on the potential impact it has on the Ethereum protocol. I would prefer if we don't have a lot of different ways to measure impact (e.g. one for kzg, one for specifications, one for clients, one for solidity, one for the deposit contract and one for FCR) but rather keep the severity measurement we have now which is how big of an overall impact a vulnerability in anything that's in scope would have on Ethereum as a whole.

We generally publish reported vulnerabilities after each hardfork, but clients could also disclose it sooner if the network is viewed as secure and that disclosing it won't have any negative impact on the network.

In my view, if a vulnerability in FCR could have created a black swan event for Ethereum (what is what we define as a Critical) then we should reward it as a critical vulnerability for Ethereum, which means up to $1M usd as long as a case can be made that the specific vulnerability would have caused a black swan event on Ethereum had it been exploited by a malicious party.

app/[locale]/bug-bounty/page.tsx
```diff
@@ -281,6 +281,18 @@ export default async function Page(props: { params: Promise<Params> }) {
...
           </p>
           <p>{t("glamsterdam-scope")}</p>
+          {/* Boxed rather than a plain paragraph: clients are not in scope for
+              Glamsterdam yet, only specifications, so a reporter who skims past this
+              wastes their time and ours on an ineligible client report. */}
+          {/* max-w-3xl by hand: MainArticle constrains the text column with
+              `**:[:is(p,ul,ol)]:max-w-3xl`, and an Alert is a div, so without this it
+              runs the full page width and reads as a different section rather than an
+              aside to the paragraph above it. */}
+          <Alert variant="update" className="max-w-3xl">
```
Comment (wackerow) [inline app/[locale]/bug-bounty/page.tsx]: Soft suggestion, could consider using "warning" on this one which will render it as a yellow alert box instead of purple

```suggestion
          <Alert variant="warning" className="max-w-3xl">
```

wackerow (dismissed):
@0xMushow Code side looks good to me! Small non-blocking suggestion below for color styling.

@0xFredrik @Ma-Julian Say the word and I can bring this in if it looks good to you.

0xMushow:
I picked this color to match the purple theme but the yellow is also fine if you want to make the diff. I don't mind)

Ma-Julian:
@wackerow good to go from my side!

@0xFredrik makes sense you don't want to have many different ways to measure impact!

wackerow (approved):
Updated to use yellow alert color on the time-sensitive notice at the top. Retargeted PR to `staging` branch to pull into today's deploy.

